PR #332, opened 2018-09-07
skip sync if p2p flag is set and throw error if fail to allocate mem object

chienwei-lan:
Hi all,

Thanks to Min's test case, help us to find out some corner cases.

1. It will attempt to allocate buffer once fail to do so, so just throw error explicitly.
2. sync a p2p BO doesn't make sense, no host buffer, direct read/write p2p BO by mmap, should skip sync any p2p BO both direction in runtime instead of handle by driver(sync_bo_ioctl will capture the attempt), it should fail earlier.

Change request are very welcome...

### DIFF
--- a/src/runtime_src/xocl/core/device.cpp
+++ b/src/runtime_src/xocl/core/device.cpp
@@ -784,7 +784,7 @@ unmap_buffer(memory* buffer, void* mapped_ptr)
   if (flags & (CL_MAP_WRITE | CL_MAP_WRITE_INVALIDATE_REGION)) {
     if (auto ubuf = static_cast<char*>(buffer->get_host_ptr()))
       xdevice->write(boh,ubuf+offset,size,offset,false);
-    if (buffer->is_resident(this))
+    if (buffer->is_resident(this) && !((buffer->get_ext_flags() >> 30) & 0x1))
       xdevice->sync(boh,size,offset,xrt::hal::device::direction::HOST2DEVICE,false);
   }
 }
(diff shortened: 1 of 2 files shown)


src/runtime_src/xocl/core/device.cpp
```diff
@@ -798,7 +784,7 @@ unmap_buffer(memory* buffer, void* mapped_ptr)
   if (flags & (CL_MAP_WRITE | CL_MAP_WRITE_INVALIDATE_REGION)) {
     if (auto ubuf = static_cast<char*>(buffer->get_host_ptr()))
       xdevice->write(boh,ubuf+offset,size,offset,false);
-    if (buffer->is_resident(this))
+    if (buffer->is_resident(this) && !((buffer->get_ext_flags() >> 30) & 0x1))
```
mamin506: The condition (buffer->get_ext_flags() >> 30) & 0x1) is not readable. Could we add a new method in memory class? For example, is_p2p_memory(). This line could become 'if (buffer->is_resident(this) && !buffer->is_p2p_memory())'.

Definition:
bool is_p2p_memory()
{
    return (m_ext_flags >> 30) & 0x1;
}

src/runtime_src/xocl/core/device.cpp
```diff
@@ -812,8 +798,10 @@ migrate_buffer(memory* buffer,cl_mem_migration_flags flags)
     buffer_resident_or_error(buffer,this);
     auto boh = buffer->get_buffer_object_or_error(this);
     auto xdevice = get_xrt_device();
-    xdevice->sync(boh,buffer->get_size(),0,xrt::hal::device::direction::DEVICE2HOST,false);
-    sync_to_ubuf(buffer,0,buffer->get_size(),xdevice,boh);
+    if(!((buffer->get_ext_flags() >> 30) & 0x1)){
```
mamin506: See above review.

src/runtime_src/xocl/core/device.cpp
```diff
@@ -822,10 +810,11 @@ migrate_buffer(memory* buffer,cl_mem_migration_flags flags)
   auto xdevice = get_xrt_device();
   xrt::device::BufferObjectHandle boh = buffer->get_buffer_object(this);
 
-  // Sync from host to device to make make buffer resident of this device
-  sync_to_hbuf(buffer,0,buffer->get_size(),xdevice,boh);
-  xdevice->sync(boh,buffer->get_size(), 0, xrt::hal::device::direction::HOST2DEVICE,false);
-
+  if(!((buffer->get_ext_flags() >> 30) & 0x1)){
```
mamin506: See above review.

PR #341, opened 2018-09-10
enable ext tag in XDMA, origin change is from Keren Xie

mamin506:
This change is from Karen Xie. It explicitly set Extended Tag bit in XDMA driver.

From Puneet email:
-Hi Ravinder,

By default the Extended Tag is always set for us. But seems that on some of Samsung’s systems the BIOS inexplicably comes and clears it. We should request Samsung to get a Lecroy capture for that system. 

Karen, Can we explicitly set this Extended Tag bit in the driver?

Thanks
Puneet_


abalasa:
retest this please

Issue #438, opened 2018-09-25
xbutil top returns ERROR: No card found


hcneema:
need to call scan

maxzhen:
This has been fixed in 2018.2 XDF

PR #439, opened 2018-09-25
xbutil top found no device

maxzhen:
This should fix issue #438.


maxzhen:
are we going to merge this to XDF branch?

PR #467, opened 2018-10-02
Added support to auto add sections when reading from a compiled JSON file

rozumx:
Work Done
    + Removed add-validation command
    + Added hidden option to skip adding a validation image
    + Added ability to set the name of a section.
    + When an xclbin header is created, it is now initialized with the correct values (instead of all zeros).
    + When sections are added, the size of the section is now reported.
    + Refactor code to have only one constructor for a section.
    + Added DRCs to verify that only one section exists when adding sections via JSON
    + Add key-value pair support
    + Added support to add sections via a multi-formatted JSON file.
    + Added multiplicity for the following options: --add-section, --remove-section, --dump-section, and --replace-section.


rozumx:
please re-test

rozumx:
Closing request for a later submission.

PR #472, opened 2018-10-03
Phase 3: xclbinutil changes

rozumx:
+ Cleans up some messaging things
+ Adds the ability to parse the build RTD into the header
+ Fixed a bug where the stringstream was not being initialized correctly via an in-memory buffer.
+ Added support to update a header section when replacing a section.
+ Refactored code to easily extract JSON section data
+ Added hooks to enable extraction of JSON data when a new section is added.
+ Updated BUILD_METADATA JSON section name to match what is used by xocc
+ Updated addSection to add the payload instead of trying to "build" a section object - bug fix.
+ Created the Section::readPayload() method to read a section's payload
+ Removed add-validation command
+ Added hidden option to skip adding a validation image
+ Added ability to set the name of a section.
+ When an xclbin header is created, it is now initialized with the correct values (instead of all zeros).
+ When sections are added, the size of the section is now reported.
+ Refactor code to have only one constructure for a section.
+ Added DRCs to verify that only one section exists when adding sections via JSON
+ Add key-value pair support
+ Added support to add sections via a multiformatted JSON file.
+ Added multiplicty for the following options: --add-section, --remove-section, --dump-section, and --replace-section.
+ Removed add-validation command
+ Added hidden option to skip adding a validation image
+ Added ability to set the name of a section.
+ When an xclbin header is created, it is now initialized with the correct values (instead of all zeros).
+ When sections are added, the size of the section is now reported.
+ Refactor code to have only one constructure for a section.
+ Added DRCs to verify that only one section exists when adding sections via JSON
+ Add key-value pair support
+ Added support to add sections via a multiformatted JSON file.
+ Added multiplicty for the following options: --add-section, --remove-section, --dump-section, and --replace-section.

### DIFF
--- a/src/runtime_src/tools/xclbin/ParameterSectionData.cxx
+++ b/src/runtime_src/tools/xclbin/ParameterSectionData.cxx
@@ -41,7 +41,7 @@ ParameterSectionData::transformFormattedString(const std::string _formattedStrin
 // String being parsed:  <section>:<formatType>:<filename>
 // Example String:       BUILD_METADATA:JSON:MY_FILE.JSON
 {
-  const std::string& delimiters = ":";      // Our delimiter
+  const std::string delimiters = ":";      // Our delimiter
 
   // Working variables
   std::string::size_type pos = 0;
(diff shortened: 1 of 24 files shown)


abalasa:
retest this please

PR #475, opened 2018-10-04
 Show DNA if available

chienwei-lan:
DNA PR

### DIFF
--- a/src/runtime_src/driver/xclng/tools/xbutil/xbutil.h
+++ b/src/runtime_src/driver/xclng/tools/xbutil/xbutil.h
@@ -276,6 +276,9 @@ public:
     {
         std::stringstream ss, subss;
         subss << std::left;
+        std::string errmsg;
+        std::string dna_info;
+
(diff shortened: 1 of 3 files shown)


chienwei-lan:
http://jira.xilinx.com/browse/PR-8479

src/runtime_src/driver/include/xclhal2.h
```diff
@@ -138,6 +138,7 @@ struct xclDeviceInfo2 {
   unsigned short mVccIntVol;
   unsigned short mVccIntCurr;
   unsigned short mNumCDMA;
+  char mDNA[128];
```
sonals: We cannot change this struct since it breaks binary compatibility.

chienwei-lan:
get value with sysfs_get(subdev, sysfs, errmsg, &string) for short term, will implement key-value pair API in the coming week.

chienwei-lan:
image

chienwei-lan:
It will look like...
image

PR #479, opened 2018-10-04
Validate execBO against compute units specified in context

stsoe:
If context has no CUs, validation is successful, otherwise the CUs specified in context must be a superset of those specfied in the command.

Update XRT command launcher to check return code of exec_buf and throw an exception if validation failed.


stsoe:
retest this please

nbalagopal:
restest this please

stsoe:
Close for cleanup

PR #486, opened 2018-10-08
QDMA update pidx twice, Ctrl+C during large MM transaction, mm queue …

houlz0507:
…counters, hw counters


src/runtime_src/driver/xclng/drm/xocl/lib/libqdma/qdma_wq.h
```diff
@@ -75,6 +75,25 @@ struct qdma_wq_stat {
         u64			total_complete_bytes;
         u32			total_req_num;
         u32			total_complete_num;
+
+	u32			descq_rngsz;
+	u32			descq_pidx;
+	u32			descq_cidx;
+	u32			descq_avail;
+	u32			desc_wb_cidx;
```
keryell: Weird indentations...

houlz0507: Those are all QDMA details for advanced debugging and from low level qdma driver. I think descq_pidx means descriptor queue producer index. :) Anyway, these are not for users to look at.

PR #488, opened 2018-10-09
fix missing interrupt for mailbox subdevice

maxzhen:
This change fixed a race in mailbox's ISR which causes interrupt to be dropped. It should fix CR 1010797.


maxzhen:
Need more work.

PR #490, opened 2018-10-09
Adds printing of the DSA, XRT, Kernel, etc. information to the info mode

ncimino:
Adds printing of the DSA, XRT, Kernel, etc. information to the info mode of xclbinutil


ncimino:
Please retry, looks like an infra issue:  sqlite3.OperationalError: database is locked

PR #491, opened 2018-10-10
Shared memory banks

mamin506:
Add Samsung FPGA-SSD DSA id.
Modify Chien-Wei's 'shared memory region across DDRs' commit and resolved conflict with master branch.

After this, we can remove the 'Storage' branch.

checkpatch.pl passed.


mamin506:
This PR pass 'board.sh' test on VCU1525 5.1 DSA

PR #492, opened 2018-10-10
use xocl_pages_to_sg instead of drm_prime_pages_to_sg

chienwei-lan:
We may not get correct sg table while create a bo >= 4G and call drm_prime_pages_to_sg

It tries to shift an **unsigned int** variable with <<**PAGE_SHIFT**, haven't been fixed in 4.18.

so we better come up with our own solution.

code:
struct sg_table *drm_prime_pages_to_sg(struct page **pages, unsigned int **nr_pages**)
{
	 ret = sg_alloc_table_from_pages(sg, pages, nr_pages, 0,
				**nr_pages << PAGE_SHIFT**, GFP_KERNEL);


sonals:
Should we instead just error out if someone is attempting allocate buffers larger than 4GB?

PR #508, opened 2018-10-15
disable ERT for u200

houlz0507:
root@XSJBRD1:~# lspci -vd 10ee:
b3:00.0 Processing accelerators: Xilinx Corporation Device 5000
	Subsystem: Xilinx Corporation Device 000e
	Flags: bus master, fast devsel, latency 0
	Memory at f8000000 (32-bit, non-prefetchable) [size=32M]
	Memory at fa000000 (32-bit, non-prefetchable) [size=128K]
	Capabilities: [40] Power Management version 3
	Capabilities: [60] MSI-X: Enable- Count=33 Masked-
	Capabilities: [70] Express Endpoint, MSI 00
	Capabilities: [100] Advanced Error Reporting
	Capabilities: [1c0] #19
	Capabilities: [400] Access Control Services
	Kernel driver in use: xclmgmt
	Kernel modules: xclmgmt

b3:00.1 Processing accelerators: Xilinx Corporation Device 5001
	Subsystem: Xilinx Corporation Device 000e
	Flags: bus master, fast devsel, latency 0, IRQ 125
	Memory at f6000000 (32-bit, non-prefetchable) [size=32M]
	Memory at fa020000 (32-bit, non-prefetchable) [size=64K]
	Capabilities: [40] Power Management version 3
	Capabilities: [60] MSI-X: Enable+ Count=33 Masked-
	Capabilities: [70] Express Endpoint, MSI 00
	Capabilities: [100] Advanced Error Reporting
	Capabilities: [400] Access Control Services
	Kernel driver in use: xocl_xdma
	Kernel modules: xocl

root@XSJBRD1:~/temp/opt/xilinx/dsa/xilinx_u200_xdma_201830_1/test# ./validate.exe verify.xclbin 

Platform information
Platform name:       Xilinx
Platform version:    OpenCL 1.0
Platform profile:    EMBEDDED_PROFILE
Platform extensions: cl_khr_icd

Found 1 compute devices!:
loading verify.xclbin

RESULT:
Hello World
root@XSJBRD1:~/temp/opt/xilinx/dsa/xilinx_u200_xdma_201830_1/test# 
root@XSJBRD1:~/temp/opt/xilinx/dsa/xilinx_u200_xdma_201830_1/test# 
root@XSJBRD1:~/temp/opt/xilinx/dsa/xilinx_u200_xdma_201830_1/test# ls
bandwidth.xclbin  kernel_bw.exe  validate.exe  verify.xclbin
[...]

PR #510, opened 2018-10-15
xocl driver does not load automatically for DSA6.0

houlz0507:


rradjabi:
Same comment as PR #508, what was the impetus for this change?

PR #511, opened 2018-10-16
Added XRT versioning support to both the xclbin archives and xclbinutil.

rozumx:
Work Done

1. Replaced axlf_header m_version (uint32_t) with m_versionPatch (uint16_t), m_versionMajor (uint8_t), and m_versionMinor (uint8_t).  Note: The order in the structure is import so that the old version of 2017, will be used for the patch version.
2. Updated xclbinutil, xclbincat, and xclbinsplit to support this new versioning schema.
3. Removed the CRC checksum dead code.


hcneema:
lizhi can you please review. the xclbin.h has the changes.

PR #512, opened 2018-10-16
Cross compile libxilinxopencl.so and libxclzynqdrv.so

mamin506:
Add a script and a Makefile to compile XRT library for Zynq platforms.

1. Run xrtdeps.sh to install necessary header file
2. Depend on Boost library

How to use this script:
$./make_arm.sh --boost_inc <Boost include> --boost_lib <Boost library> --target aarch64

### DIFF
--- a/src/runtime_src/Makefile
+++ b/src/runtime_src/Makefile
@@ -1,0 +1,88 @@
+# Build xilinxopencl.so for ARM
+
+CROSS_COMPILE := aarch64-linux-gnu-
+CXX := $(CROSS_COMPILE)g++
+
+ifeq ($(CROSS_COMPILE), aarch64-linux-gnu-)
+	ARCH := aarch64
+else
+	ARCH := arm
+endif
+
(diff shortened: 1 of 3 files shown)


hcneema:
please check if xrtdeps.sh needs to be committed too.

NaAl-Xlnx:
retest this please

hcneema:
please retest

hcneema:
please retest

abalasa:
retest this please

src/runtime_src/Makefile
```diff
@@ -0,0 +1,88 @@
+# Build xilinxopencl.so for ARM
```
stsoe: Let's rename to arm.mk or something else.  Point is this is not a generic Makefile from build runtime_src which you might otherwise be led to believe.

sonals:
Now that we have support for Yocto we decided to keep these Makefiles in a separate branch.

PR #515, opened 2018-10-16
Enable build for Pegasus 

rradjabi:
'setenv AWS_PEGASUS_TARGET 1' will build the rpm for Pegasus and enables building awsmgmt.


rradjabi:
After discussion with @stsoe , I think I need to modify how this flow works. The workflow for the builder/tester in automation should be to run build.sh and generate two packages -> xrt-<version>.<pkg_type> and xrt-aws-<version>.<pkg_type>. We do not want to kick off two separate builds, but we do want to generate two packages, assuming the environment variable is set.

abalasa:
retest this please

rradjabi:
Closing in favor of #621

PR #516, opened 2018-10-16
Updated hot reset to first stop XMC/ERT before going through reset

cellery:
Added stop xmc command which is copy pasted from load_xmc command. Maybe could optimize code and just have load_xmc call stop_xmc function instead of duplicated in code.

Created a stub for legacy mgmt microblaze which doesn't need to do anything for this

### DIFF
--- a/src/runtime_src/driver/xclng/drm/xocl/mgmtpf/mgmt-utils.c
+++ b/src/runtime_src/driver/xclng/drm/xocl/mgmtpf/mgmt-utils.c
@@ -237,6 +237,9 @@ long reset_hot_ioctl(struct xclmgmt_dev *lro)
 		xocl_reset(lro, true);
 	}
 
+	/* request XMC/ERT to stop */
+	xocl_mb_stop(lro);
+
 	if (!lro->reset_firewall) {
 		/*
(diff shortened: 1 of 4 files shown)


NaAl-Xlnx:
retest this please

cellery:
@maxzhen I cleaned up the code and removed the duplicate part

cellery:
Do we even need a lock on xmc? Can multiple threads be calling the same stop/load xmc commands?

You'd have to have nested locks if you want a lock in both load and stop.

maxzhen:
> 
> 
> Do we even need a lock on xmc? Can multiple threads be calling the same stop/load xmc commands?
> 
> You'd have to have nested locks if you want a lock in both load and stop.

Inside OS kernel, it's always going to be multi-threaded. So, yes. If you are not sure if mutex can be dropped, better not. It does not mean you will have nested lock if you want a lock in both stop and load. You could create a xmc_stop_nolock() for both xmc_load and xmc_stop to call. This seems to be very common practice to me.

cellery:
@maxzhen updated load_xmc to not use 2 locks

PR #528, opened 2018-10-18
Bug Fixes

rozumx:
+ The info option now report both the Feature ROM's VBNV and timestamp
    + Added support to remove the key for a key-value pair
    + Added positional argument detection that produces an error if no option was specified with it.
    + Removed obsolete help example
    + Added warning if user didn't specify and output file
    + Added initial support for sub-sections


stsoe:
That's a lot of bugs :-)

PR #551, opened 2018-10-22
DNA AXI with integrated xclbin and cert

chienwei-lan:
don't need dnas_cert.bin under /lib/firmware/xilinx/ anymore

get certificate section from xclbin, copy the section, parse and write message and signature in driver


chienwei-lan:
XRT/build/board.sh -board u200 -sync ALL PASS

PR #553, opened 2018-10-22
Clean up documentation

keryell:
Fix documentation generation.

Connect the documentation about testing to the main page.

Clarify GTest usage for old and recent Ubuntu.

Do not assume that the repository is actually clone in XRT directory.

Clarify a few points while testing compilation.

Fix some typos.

This is mainly fixing a lot of ReST issues.

Use real em-dash instead of en-dash

### DIFF
--- a/src/runtime_src/doc/toc/build.rst
+++ b/src/runtime_src/doc/toc/build.rst
@@ -19,7 +19,7 @@ Build the runtime
 
 ::
 
-   cd XRT/build
+   cd build
    ./build.sh
 
 ``build.sh`` script builds for both Debug and Release profiles.  On RHEL/CentOS, if ``build.sh`` was accidentally run prior to enabling the devtoolset, then it is necessary to clean stale files makefiles by running ``build.sh clean`` prior to the next build.
@@ -29,7 +29,7 @@ Build RPM package on RHEL/CentOS or DEB package on Ubuntu
 
 ::
 
-   cd XRT/build/Release
+   cd build/Release
    make package
 
 Install the XRT RPM package
@@ -53,6 +53,8 @@ XRT Documentation can be built automatically using Sphinx doc builder together w
 
 ::
 
-   cd XRT/src/runtime_src/doc/
-   make
-   firefox html/index.html
+   cd src/runtime_src/doc
+   cmake .
+   make xrt_doc
+   # To look at the generated local documentation with a web browser:
+   xdg-open html/index.html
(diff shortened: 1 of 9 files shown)


src/runtime_src/doc/toc/build.rst
```diff
@@ -19,40 +19,47 @@ Build the runtime
...
    yum reinstall ./XRT-2.1.0-Linux.rpm
 
 Install the XRT DEB package
 ...........................
 
-::
+Install with (the actual package name might differ) ::
 
-   apt install --reinstall ./XRT-2.1.0-Linux.deb
+   sudo apt install --reinstall ./xrt_201830.2.1.0_18.10.deb
```
stsoe: Add sudo for the yum command also.  I'd prefer to  leave out the 201830 and 18.10 from package name; these change depending on release and OS

src/runtime_src/doc/toc/build.rst
```diff
@@ -19,40 +19,47 @@ Build the runtime
...
 
-::
+To compile and install the documentation into the ``doc`` directory at
+the top of the repository::
 
-   cd XRT/src/runtime_src/doc/
-   make
-   firefox html/index.html
+   cd src/runtime_src/doc
+   cmake .
```
stsoe: .I would recommend running cmake only in temporary directories, e.g. 
% mkdir build
% cd build
% cmake ..
Makes it much easier to clean up.

src/runtime_src/doc/toc/build.rst
keryell: Yes this is why I have added `(the actual package name might differ)` in the line above, so I think it can remain like that. Before it was wrong in the sense that almost none of the characters would happen like that. So it is at least a plausible example. :-)

I can add `sudo` for the `yum` command too. I did not tried the RedHat style, so I did not update this part, but yes I can add it.

The philosophy behind the changes was to make only minor changes since I do not have the big picture of the project. I am expecting some other newcomers to do improvements on the RedHat side for example, now it is open-source ! :-)

src/runtime_src/doc/toc/build.rst
keryell: Yes, it would be great, even if `git clean` can help.
But it would require changing `src/runtime_src/doc/CMakeLists.txt` to make it work...
So just keep this PR low-profile with only non-functional change and leave some work for the future generations. :-)

stsoe: Works for me in a separate build directory.

Issue #544, opened 2018-10-19
The compiled documentation should not be in the source repository


keryell:
There should be no https://github.com/Xilinx/XRT/tree/master/docs directory in the Git repository, since it is a by-product of the documentation compilation.

I hope that this is not directly used by https://xilinx.github.io/XRT/ ...

keryell:
To make progress on this, one can look for example at https://github.com/triSYCL/triSYCL/blob/master/doc/architecture.rst#source-code-documentation about how to move the generated documentation into a separate branch.

mamin506:
@keryell Hi, Ronan. The docs/ folder was removed from master branch. If you run build.sh, the document will be built in Release/runtime_src/doc/ folder. Only Sonal and Soren can publish documents to gh-pages branch. See #765.

keryell:
That seems like a reasonable solution.
At least the git repository is quite cleaner now!
Thanks.

PR #555, opened 2018-10-22
Remove top docs directory

keryell:
This directory is actually generated when compiling the documentation,
so do not keep it in the repository.
Furthermore it messes up the ``git status``.

This should fix https://github.com/Xilinx/XRT/issues/544


stsoe:
I verified that the build pipelined failed after successfully building and running all tests, so presumably infrastructure.

PR #561, opened 2018-10-23
fix missing interrupt issue for mailbox

maxzhen:
This change fixed the missing interrupt issue for mailbox. It also disables mailbox timer when no msg is in the queue (2nd try).
It passed Soren's board test.


stsoe:
Already verified with local board tests.

PR #564, opened 2018-10-23
Enhance the DNA Certificate Section to support JSON output.

rozumx:
Work Done
    + DNA_CERTIFICATE Section now supports dumping the contents out in a JSON format
    + Updated the MIRROR DATA section to only insert JSON data if only the section being inserted supports both readering and wriing JSON
    + Updated xclbinsplit to add the section names to the output file for DNA_CERTIFICATE, BUILD_METADATA, KEYVALUE_PAIRS, and USER_METADATA sections.
    + Updated xclbinsplit to add 'unknown(#)' if the section name is not know (future proofing).

### DIFF
--- a/src/runtime_src/tools/xclbin/xclbindata.cxx
+++ b/src/runtime_src/tools/xclbin/xclbindata.cxx
@@ -341,6 +341,22 @@ XclBinData::extractSectionData( int sectionNum, const char* name )
     type = "debug";
     ext = ".bin";
   }
+  else if ( header.m_sectionKind == DNA_CERTIFICATE ) {
+    type = "dna_certificate";
+    ext = ".bin";
+  }
+  else if ( header.m_sectionKind == BUILD_METADATA ) {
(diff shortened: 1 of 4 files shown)


src/runtime_src/tools/xclbin/xclbindata.cxx
```diff
@@ -373,8 +389,13 @@ XclBinData::extractSectionData( int sectionNum, const char* name )
   else if ( header.m_sectionKind == BMC ) {
     extractAndWriteBMCImages((char*) data.get(), sectionSize);
     return true;
+  } else {
+    static unsigned int uniqueCount = 1;
+    type = "unknown(" + std::to_string(uniqueCount) + ")";
```
ncimino: Should this get incremented?

rozumx: Great catch.  Changes made to increment this value.

PR #566, opened 2018-10-23
Trace support for streaming platform

hackwa:


src/runtime_src/xrt/device/halops2.cpp
```diff
@@ -161,6 +162,10 @@ operations(const std::string &fileName, void *fileHandle, unsigned int count)
   if (!mGetProfilingSlotName)
     return;
 
+  mGetProfilingSlotProperties = (getSlotPropertiesFuncType)dlsym(const_cast<void *>(mDriverHandle), "xclGetProfilingSlotProperties");
+  if (!mGetProfilingSlotProperties)
+    return;
```
stsoe: I am not liking the "return" out of the constructor when it fails to find a symbol.  I know this is a general problem in this file.  A return half way through means the object is partially constructed.   It seems that the return is a indication something is wrong, so maybe an exception should be thrown instead and the object should never exist?   

Some functions don't have to exist in the hal implementation so client code handles this.  For example emulation hal drivers don't need to have all functions implemented.  But when a function is required in the hal driver, then it should be an exception when it is not there.   

Have returns in the middle of the constructor means that one has to be very careful where new data members are initialized.   

I would like to clean this up; therefore I am asking you what the consequence would be if all returns were removed for the profiling / stats functions?

hackwa: Hey Soren,
Good idea. Profiling shouldn't be affected by removal of return. if the function is absent then a default value is returned for most of the calls. I will double check this.

hackwa:
hey @stsoe I have removed the return statements from profiling calls.

PR #567, opened 2018-10-23
Multi-process support feature for 2018.3

sonals:
The feature can be enabled by setting env variable XCL_MULTIPROCESS_MODE to 1


src/runtime_src/driver/xclng/drm/xocl/subdev/mb_scheduler.c
```diff
@@ -48,6 +48,8 @@
 		DRM_INFO("packet(0x%p) data[%d] = 0x%x\n",data,i,data[i]); \
 })
 
+//#define SCHED_VERBOSE
```
stsoe: Let's not have this macro in two places.   I don't care if it is on line 26 or 51, but remove one, please.

src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
```diff
@@ -73,12 +73,14 @@ int xocl_execbuf_ioctl(struct drm_device *dev,
 		return -EINVAL;
 	}
 
+#if 0
+	//xocl_exec_validate() should do the validation instead
```
stsoe: Why is validate commented out?

src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
```diff
@@ -173,10 +175,12 @@ int xocl_ctx_ioctl(struct drm_device *dev, void *data,
 			goto out;
 
 		xdev->ip_reference[args->cu_index]--;
-		if (bitmap_empty(client->cu_bitmap, MAX_CUS))
-                        // We jsut gave up the last context, give up the xclbin lock
+		if (bitmap_empty(client->cu_bitmap, MAX_CUS)) {
+                        // We just gave up the last context, give up the xclbin lock
+			uuid_copy(&client->xclbin_id, &uuid_null);
```
stsoe: clearing of xclbin_id is also done by out:

src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
```diff
@@ -545,13 +556,16 @@ xocl_read_axlf_helper(struct xocl_dev *xdev, struct drm_xocl_axlf *axlf_ptr)
 			err = -EBUSY;
 			goto done;
 		}
+#if 0
```
stsoe: remove ?

src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
```diff
@@ -678,10 +692,14 @@ xocl_read_axlf_helper(struct xocl_dev *xdev, struct drm_xocl_axlf *axlf_ptr)
 done:
 	if (size < 0)
 		err = size;
-	if (err != 0) {
+	/*
+	 * Always give up ownership for multi process use case; the real locking
+	 * is done by context creation API
+	 */
+//	if (err != 0) {
```
stsoe: remove?

src/runtime_src/xrt/device/hal2.cpp
```diff
@@ -123,7 +123,7 @@ void
 device::
 acquire_cu_context(const uuid& uuid,size_t cuidx,bool shared)
 {
-#if 0
+#if 1
```
stsoe: Enabling this implies we are confident mailbox works.

src/runtime_src/xrt/device/hal2.cpp
```diff
@@ -139,7 +139,7 @@ void
 device::
 release_cu_context(const uuid& uuid,size_t cuidx)
 {
-#if 0
+#if 1
```
stsoe: Enabling this implies we are confident mailbox works.

src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
```diff
@@ -678,10 +692,14 @@ xocl_read_axlf_helper(struct xocl_dev *xdev, struct drm_xocl_axlf *axlf_ptr)
 done:
 	if (size < 0)
 		err = size;
-	if (err != 0) {
+	/*
+	 * Always give up ownership for multi process use case; the real locking
```
maxzhen: If we always unlock xclbin, why do we lock it in the first place?

src/runtime_src/driver/xclng/drm/xocl/subdev/mb_scheduler.c
sonals: Resolved in my updated PR

src/runtime_src/xrt/device/hal2.cpp
sonals: Yes it does, but we are still protected by XCL_MULTIPROCESS_MODE which we should remove once MailBox is solid.

src/runtime_src/xrt/device/hal2.cpp
sonals: Yes it does, but we are still protected by XCL_MULTIPROCESS_MODE which we should remove once MailBox is solid.

src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
sonals: This is to keep the userpf population of xclbin metadata in sync with image running on FPGA (actually loaded by mgmt). Once the user gives up the lock, another process can change the image on FPGA via mgmt and will be then responsible to sync the view in userpf -- in the current architecture.

maxzhen: Why do we need to keep them in sync? Does userpf interact with the xclbin on the FPGA while populating metadata somehow? It seems not...

src/runtime_src/xrt/device/hal2.cpp
stsoe: No, we are not protected by the env var for acquire/release_cu_context.  Enabling this code implies that xclOpen/CloseContext will be called always and reference counting of xclbin better be working.  I am okay enabling it after Max's changes yesterday.

src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
sonals: This is incremental change from before where this lock would be given up only at process exit. Now we want to give up the lock right after we have updated the metadata. 2019.1 let's revisit this.

maxzhen: > This is incremental change from before where this lock would be given up only at process exit. Now we want to give up the lock right after we have updated the metadata. 2019.1 let's revisit this.

Fair enough, yes.

sonals:
To test this do the following in every shell you want to leverage this feature --
setenv XCL_MULTIPROCESS_MODE 1

Then in each of shell1 and shell2 run one application and attempt to run a different application in shell3. 1 and 2 should pass and 3 should fail.
shell1>./kernel_bw.exe bandwidth.xclbin

shell2>./kernel_bw.exe bandwidth.xclbin

shell3>./verify.exe verify.xclbin

PR #569, opened 2018-10-24
mailbox request processing should not be done in RX thread

maxzhen:
Fixes in this push:
+ mailbox request processing should not be done in RX thread
+ reenable reset scheduler via mailbox since the mailbox interrupt issue has been fixed
+ add mailbox_no_intr module param to enable timer-driven msg exchange mode for debugging


maxzhen:
Passed mailbox message stress test and all board test. Load/unload drivers worked fine.

PR #574, opened 2018-10-25
For Sprite machine

sarab96:
For Sprite machine
@hcneema Please do the needful


sarab96:
Changed the include and lib paths to match.
@hcneema  @stsoe  Please check

PR #575, opened 2018-10-25
fixed getting arguments from port:memory strings

pgschuey:


src/runtime_src/xdp/profile/rt_profile.cpp
```diff
@@ -1704,14 +1704,20 @@ else if (functionName.find("clEnqueueMigrateMemObjects") != std::string::npos)
     argNames = "All";
     memoryName = "N/A";
 
+    std::string portName2 = portName;
+    size_t index = portName.find_last_of(":");
+    if (index != std::string::npos) {
+      portName2 = portName.substr(0, index);
+    }
+
```
stsoe: Above 5 lines  could be simply
portName = portName.substr(0,portName.find_last_of(":"));

pgschuey: good suggestion. much better now...

PR #579, opened 2018-10-25
Update testcases with memory bank error message

haeseung81:
Updated all the utils.h files and host codes.
But, 100_ert is not updated because it's using utils.hpp, which has a different structure.
100_ert and utils.hpp will be handled separately.

### DIFF
--- a/tests/xrt/00_hello/main.cpp
+++ b/tests/xrt/00_hello/main.cpp
@@ -145,11 +145,15 @@ int main(int argc, char** argv)
     {
 	xclDeviceHandle handle;
 	uint64_t cu_base_addr = 0;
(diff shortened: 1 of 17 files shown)


hcneema:
please retest. I will file new issues for the following
1. Makefile is not working currently
2. every test has a copy of utils.h. We should pick up from host_src area for for nightly and for 1) above.

hcneema:
please retest

haeseung81:
I was able to compile/build the xrt on the local machine and my change does not include the XRT portion.
I also checked the build log, and it looks like something went wrong during driver installation, which is not related to my change.

abalasa:
retest this please

PR #599, opened 2018-10-31
profile summary support in emulation on aws platform

hackwa:


hackwa:
This is on 2018.2_XDF branch.

PR #602, opened 2018-10-31
get frequency from clock frequency counter(actual value)

chienwei-lan:
http://jira.xilinx.com/browse/CR-1015217?filter=-1

http://jira.xilinx.com/browse/CR-1011272?filter=-1

1. check the clock frequency counter(in KHz) and compare to the request frequence value.
2. tolerance is 5%, if delta is large than 5%, consider as fail.
3. get frequency from clock counter instead of clockwiz(shown in xbutil query)


abalasa:
retest this please

abalasa:
retest this please

PR #606, opened 2018-11-01
AWS Package 

rradjabi:
Create an AWS package that is a plugin to XRT. This pulls awssak and libxrt_aws out of the main installer. 'xrt-aws' also contains the dkms installer for awsmgmt. 

So when build.sh is called, two packages are generated. awssak and libxrt_aws are part of the 'aws' component, thus are packaged in xrt-<version>-aws.rpm (along with awsmgmt via dkms). All other target objects are part of an unspecified component and are unfortunately packaged as "-Unspecified" component, so the package is xrt-<version>-Unspecified.rpm. We need to figure out how to remove the -Unspecified name from not only the file, but also the package, because you install/remove by name 'xrt-Unspecified'.


rradjabi:
@sonals & @stsoe , can you please suggest a solution for the default package name? I have not seen a solution either in the documentation or in forums online.

I still need to add support for Debian builds.

rradjabi:
Debian support added, but this needs to be tested.

rradjabi:
One solution for maintaining the the 'xrt' package name without appending '-core' or something is to run 'make package' an additional time and switch component building on. I've tested this in build.sh by adding these lines:

build.sh:
```
131 cd $BUILDDIR/Release
132 export AWS_PEGASUS_TARGET=1
133 time cmake -DRDI_CCACHE=$ccache -DCMAKE_BUILD_TYPE=Release -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ../../src
134 make package
135 cd $here
```

src/CMakeLists.txt:
```
139 if( $ENV{AWS_PEGASUS_TARGET} )
140     set(CPACK_RPM_COMPONENT_INSTALL ON)
141     set(CPACK_DEB_COMPONENT_INSTALL ON)
142     set(CMAKE_INSTALL_DEFAULT_COMPONENT_NAME "core")
143 endif()
```

The first run of make package will produce the original package containing all components in one deliverable 'xrt'. The second run of make package will turn on component packaging, and will produce 'xrt-aws' and 'xrt-core', which we can ignore 'xrt-core'. 

I'm going to see if I can disable packaging the 'aws' component in the first make package.

PR #608, opened 2018-11-01
ECC mem controller support

maxzhen:
This change is to support ECC controllers. The xbutil mem --query-ecc is added to report ECC errors. The xbutil mem --reset-ecc is added to reset ECC error status.


abalasa:
retest this please

PR #609, opened 2018-11-02
various bug fixes

maxzhen:
Fixed a few bugs in driver xocl:
1. icap->icap_regs is NULL in xocl, can't be dereferenced in icap_probe_chip().
2. count in kds_custat_show() should be initialized before comparing to 20.
3. xdev->topology maybe NULL when no xclbin is ever downloaded.
4. i in kdsstat_show() is not used, causing compilation warning.


abalasa:
retest this please

PR #612, opened 2018-11-05
Drop crude uuid hack in xclbin.h in favor of standard uuid/uuid.h

sonals:
uuid package has been a requirement for some time now. Remove our UUID hack from xclbin.


abalasa:
retest this please

PR #615, opened 2018-11-05
Added macronix sector size #s to xspi.cpp

cellery:
Fixes bug w/ Macronix parts


cellery:
Hem can you review this and approve? This is a bug fix for the Vega board since it uses Macronix and not Micron flash parts. I tested it on a Vega board to verify it works.

PR #619, opened 2018-11-06
icap_download_boot_firmware() may release unheld mutex

maxzhen:
During error handling, icap->icap_lock may be unlocked without first being locked. In this change, the code is moved around so that lock is grabbed only after dsabin is loaded into kernel memory and will always be unlocked on error.


maxzhen:
fixed with another commit.

PR #625, opened 2018-11-06
Fixes the Ubuntu 18.04 xclbin{util,cat,split} linking issue where a d…

ncimino:
…ynamic executable was being produced even though a static executable was specified in CMake


ncimino:
Fixes: http://jira.xilinx.com/browse/CR-1015915 
Needed for both HEAD and 2018.3

PR #631, opened 2018-11-07
Store sensor info in boost::ptree and output to JSON

rradjabi:
This is a work in progress and it needs some feedback.


rradjabi:
Refs #632

src/runtime_src/driver/xclng/tools/xbutil/xbutil.h
```diff
@@ -207,32 +210,53 @@ class device {
         return 0;
     }
 
+    int parseComputeUnits(std::vector<ip_data> &computeUnits) const
```
sonals: **const** std::vector<ip_data> &computeUnits)

src/runtime_src/driver/xclng/xrt/user_common/sensor.cpp
```diff
@@ -0,0 +1,28 @@
...
+//    root.put( "board.firewall", "" );
+//    root.put( "board.xclbin", "" );
+//    root.put( "board.memory", "" );
+//    root.put( "board.stream", "" );
+//    root.put( "board.compute_unit", "" );
+    
+    return 0;
+}
+
+int writeTree( boost::property_tree::ptree &root )
```
sonals: **const** boost::property_tree::ptree &root

src/runtime_src/driver/xclng/tools/xbutil/xbutil.h
```diff
@@ -207,32 +211,53 @@ class device {
...
```
maxzhen: Maybe we should stop adding implementation code in header file...

src/runtime_src/driver/xclng/tools/xbutil/xbutil.h
```diff
@@ -722,6 +775,239 @@ class device {
 
         lines.push_back(ss.str());
     }
+    
+    int readSensors( void ) const
```
maxzhen: We are not reading just sensor output here. Maybe we should split the function into two: one is for reading static info and the other is for dynamic info. We may need to call dynamic one in xbutil top repeatedly to get up-to-date info.

rradjabi:
retest this please

src/runtime_src/driver/xclng/tools/xbutil/xbutil.h
rradjabi: I agree, I will try to move this into the sensor files but that's beyond the scope of this feature.

nbalagopal:
retest this please

src/runtime_src/driver/xclng/xrt/user_common/sensor.h
```diff
@@ -0,0 +1,12 @@
+#ifndef SENSOR_H
+#define SENSOR_H
+
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/json_parser.hpp>
+#include <boost/foreach.hpp>
+
+static boost::property_tree::ptree gSensorTree;
```
stsoe: No static variables in header files.  If the header is include multiple times, each compilation unit will have an instance of the static variable.  Move to sensor.cpp

rradjabi: @stsoe 
Please review my latest changes. I made this a singleton class where the ptree object is a private member variable. I've written wrappers around ptree::get and ptree::put.

stsoe: I am confused, where is the global used?  Who calls Instance()?

rradjabi: The property tree (previously gSensorTree) is no-longer global. Instance() provides a reference to the singleton class where sensorTree::put() and sensorTree::get() are wrappers around boost::property_tree::ptree::put() and ::get().

stsoe: I still don't see where Instance() is called?

rradjabi: In xbutil.h (See line 239) wherever sensorTree::put() is called it must be called like so:


```
sensorTree &tree = sensorTree::Instance();
tree.put( val );
```

stsoe: ok. thanks.

PR #636, opened 2018-11-07
better check xmc error and show power string

maxzhen:
This change fixes:
CR-1015457: Improved XMC error messaging and checking
CR-1015461: Show max power available in Watts

### DIFF
--- a/src/runtime_src/driver/xclng/tools/xbflash/flasher.cpp
+++ b/src/runtime_src/driver/xclng/tools/xbflash/flasher.cpp
@@ -138,6 +138,16 @@ std::string charVec2String(std::vector<char>& v)
     return ss.str();
 }
 
+std::string int2PowerString(unsigned lvl)
+{
+    std::vector<std::string> powers{ "75W", "150W", "225W" };
+
+    if (lvl < powers.size())
+        return powers[lvl];
+
+    return std::to_string(lvl);
+}
+
 int Flasher::getBoardInfo(BoardInfo& board)
 {
     std::map<char, std::vector<char>> info;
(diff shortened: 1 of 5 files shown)


maxzhen:
@cellery please take a look at this change.

maxzhen:
You can upgrade BMC FW when the MSP432 is not working? Basically, you don't require a working MSP432 FW as a base to be able to upgrade the FW?

cellery:
If the FW fails in the middle of updating it (i.e. power loss) the next time the system boots up MSP432 will go into BSL mode by default since there is no valid FW. In this scenario the XMC will boot and not be able to get board information until we've loaded FW into MSP432. BSL mode does not support reading from EEPROM to retrieve board info.

Do you see any issues with this? The only thing you rely on from board info is the MSP342 FW version right? In this scenario it doesn't matter that we can't read it because we have to update the FW anyways.

maxzhen:
I'd thought we always have a working MSP432 FW as base just like the golden image. I think that if MSP432 is in BSL mode, xbutil flash -p may work, but xbutil flash -a won't since it needs to compare the FW version, which requires a working FW. I'll make the change and see how it goes.

cellery:
For -a option then I would check MSP432 mode. If it reports BSL mode then you can just skip the version check and program it anyways. I've already tested the XMC for this scenario so it should already handle this condition.

The MSP432 only stores 1 FW image and if its corrupted or unprogrammed it reverts back to BSL mode. Its not ideal and I've asked for long term to support rollback/recovery for MSP432

maxzhen:
OK. That sounds like a bigger change not suitable for 2018.3. I'll close this PR for now and resubmit it after we have a branch for 2019.1 next week.

PR #640, opened 2018-11-09
Force install packages with xrtdeps.sh

rradjabi:
All install commands take the -y switch now. This simplifies scripted calls of xrtdeps.sh.


stsoe:
I am closing this PR, which is targeted wrong branch.
I am in the process of adding cmake3 to centos and will make the -y changes also.

Issue #529, opened 2018-10-18
Add option to force install in xrtdeps.sh


rradjabi:
Non-interactive shells need the option to force install of packages in xrtdeps.sh. This is an issue for AWS F1 developer flow, since xrtdeps.sh will be called from sdaccel_setup.sh, and needs to confirm install for all packages by default.

rradjabi:
I have added a '-force' option to xrtdeps.sh and confirmed it worked in these steps:

1) in a setup with all packages installed, remove one package 'sudo yum remove devtoolset-6'
2) call xrtdeps.sh to confirm it prompts for y/n to install devtoolset-6, reject with 'n'
3) call 'xrtdeps.sh -force' and see that it forces install of devtoolset-6.

PR #641, opened 2018-11-09
Add cmake3 to xrtdeps.sh for CentOS

stsoe:
Also incorporate changes in #640


rradjabi:
Soren, would you please merge this commit to 2018.2_XDF as well? We need that for a new AWS 2018.2 revision.

09b9363 closes #529

Issue #645, opened 2018-11-09
xclbinsplit fails to link when CMAKE_MINIMUM_REQUIRED is upped


stsoe:
Please change CMAKE_MINIMUM_REQUIRED in src/CMakeLists.txt to:
CMAKE_MINIMUM_REQUIRED(VERSION **3.5.1**)
then run ./build.sh.

The build will fail to link xclbinsplit on both Ubuntu and CentOS.

We need to use minimum 3.5 of cmake for some changes we are making.  PR #644 updates build.sh to use cmake3 on CentOS, Ubuntu already has version 3.5.x.  

While we can still use the later version of cmake without upping the minimum required version in CMakeLists.txt,  it'd be better if the required version could be encoded in the CMakeLists.txt file.

PR #646, opened 2018-11-09
Updated CMakeLists.txt to find only static boost libraries.

rozumx:
Note: Prior to this change both the dynamic and static boost libraries were
being found.  Unfortunately, this led to compilation issues when CMAKE_MINIMUM_REQUIRED was updated to 3.5.1.

Note: This pull request addresses issue #645 .


stsoe:
Resolves #645

PR #651, opened 2018-11-10
Break scheduler loop when timing out

stsoe:


stsoe:
This change has been verified by @houlz0507

PR #652, opened 2018-11-12
Kick start the effort for the Python library and possibly other community contributions 

tianhaoz95:
It is working in progress, but good to give it a kick start.

##### Standalone & directory independent
* The project can be moved from `/contrib/python-xrt` to anywhere and it will work fine.
* The project is completely separate from XRT project structure and build system.

##### Designed to achieve zero extra C++ when XRT becomes ready
* Top level APIs are made with ctypes (both value and pointer) conversion which eventually can be replaced by native XRT API when it becomes more high level language friendly.
* Switching between static linking / compiling with XRT source / dynamic loading is easy.


maxzhen:
close per discussion with Tianhao/Sonal/Shivangi

PR #658, opened 2018-11-13
Add interface for integrated logging with multi-media frameworks

jnicholsXilinx:
This is a small update to XMA for support of integrated logging messages by registering a callback function that gets invoked when a log message of the requested severity level or greater is generated.


sarab96:
Looks Good
@sonals  Please go ahead

PR #680, opened 2018-11-15
CR-1016758 XRT driver version are still 2018.2.X

houlz0507:
Verified on vcu1525
# xbutil scan
INFO: Found total 1 card(s), 1 are usable
[0]mgmt:[45:00.0]:0x6a9f:0x4360:[xclmgmt:2.1.0:83200]
[0]user:[45:00.1]:0x6aa0:0x4360:[xocl_qdma:2.1.0:129]


houlz0507:
With this change, driver always has the same version with XRT.  We do not have to manually change driver version any more.

PR #685, opened 2018-11-16
mailbox's wait for reponse should not be interruptible

maxzhen:
This change fixes kernel panic when user hits Ctrl-C to quit the application.

In mailbox_request(), we want to wait for response uninterruptibly. We will rely on mailbox's timeout mechanism to return from waiting, if anything goes wrong. This will ensure that even when user application is killed, mailbox_request() works as expected, instead of returning prematurely.


maxzhen:
I'm merging it in after reviewed with Sonal.

PR #687, opened 2018-11-16
ignore bad ECC controller entry in IP_LAYOUT section

maxzhen:


maxzhen:
pipeline is getting stuck again? could someone check it?

maxzhen:
I verified on Ellery's U200 and Anapayan also verified it on his boards to make sure this is working. I'm merging it in since this is urgent and pipeline does not work.

PR #689, opened 2018-11-16
Xbutil query format improvements

### LINKED ISSUE #681: xbutil query output improvement after json format change
+ A few bits of information are missing. For example part name/IDCODE, etc.
+ It seems that xbutil query now report vendor/device IDs in DEC. Could you fix them so that they are still reported in HEX since it’s in HEX everywhere else…

rradjabi:
* Adds idcode, xclbin uuid
* formats some output in hex as they were before

Fixes #681


sonals:
Merging after manual validation on my host. Build pipeline seems to be stuck.

PR #693, opened 2018-11-17
Fixed device info for VCU1525 2018.3 DSA

cellery:
Fixed VCU1525 2018.3 device to point to non XBB DSA. Was accidentally putting in xmc subdev for VCU1525 which it does not support.


maxzhen:
This is urgent, should not impact build. Merging without waiting for pipeline fix.

PR #694, opened 2018-11-17
Lock xclbin prior to using KDS

stsoe:
Fixes problem related to xclbin locking and icap reset of kds as part of AXI reset.  The AXI reset is performed when a process tries to download an xclbin (icap_download_bitstream_axlf) *and* that xclbin is unused (first download or already downloaded but unlocked).

- P1 downloads xclbin calling xocl_read_axlf_helper
-- not locked so resets AXI => resets KDS
-- aquires xclbin lock (ref=1)
-- releases xclbin lock (ref=0)
- P1 configures scheduler
-- This does not mess with xclbin locking, which remains ref=0
-- Configure command enters KDS command queue
-- P1 waits in execWait for configure command to complete
- P2 downloads xclbin
-- not locked so resets AXI => resets KDS
-- The pending P1 configure command is ‘stale’ from KDSs perspective, this is the point of reset, so the command is removed
-- acquires xclbin lock (ref=1)
-- releases xclbin lock (ref=0)

If P1 is slow getting serviced by KDS, P1 will never complete, it hangs in xclExecWait for configure command.

The fix in this PR is to lock the xclbin for a client context prior to submitting a command to KDS.  The xclbin is locked only if necessary, no repeated attempts to lock an already locked xclbin.

Error check that context locks the expected xclbin, not some xclbin swapped in by another process.

A process must now acquire a context on the device (lock the xclbin) before submitting any commands to the scheduler.  In this PR the lock on the xclbin lock is acquired implicitly if necessary, but if and only if the current process did indeed download the xclbin.

### DIFF
--- a/src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
+++ b/src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
@@ -33,6 +33,37 @@
 xuid_t uuid_null = NULL_UUID_LE;
 #endif
 
+static int
+xocl_client_lock_bitstream(struct xocl_dev *xdev, struct client_ctx *client)
+{
+	if (atomic_read(&client->xclbin_locked))
+		return 0;
+
+	if (xocl_icap_lock_bitstream(xdev, &xdev->xclbin_id,pid_nr(task_tgid(current)))) {
+		userpf_err(xdev,"could not lock bitstream for process %d",pid_nr(task_tgid(current)));
+		return 1;
+	}
+
+//      Allow second process to use current xclbin without downloading
+//	if (uuid_is_null(&client->xclbin_id))
+//		uuid_copy(&client->xclbin_id,&xdev->xclbin_id);
+//	else
+	if (!uuid_equal(&xdev->xclbin_id,&client->xclbin_id)) {
+		userpf_err(xdev,"device xclbin does not match context xclbin, cannot obtain lock for process %d",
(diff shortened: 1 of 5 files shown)


stsoe:
This is not the ultimate solution, but a work-around for hang and panic.

src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
```diff
@@ -188,7 +188,7 @@ int xocl_ctx_ioctl(struct drm_device *dev, void *data,
 		goto out;
 	}
 
-	if (bitmap_empty(client->cu_bitmap, MAX_CUS))
+	if (bitmap_empty(client->cu_bitmap, MAX_CUS) && uuid_is_null(&client->xclbin_id))
```
maxzhen: i think it's better to add a dedicated boolean to keep track of contexts. it's quite complicated to infer...

stsoe:
Max please review again.   I am thinking this is a permanent solution based on last email exchange.

src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
stsoe: Yup. Thanks

sonals:
Let's document the steps then and merge this PR.

sonals:
Retest this please

stsoe:
Merging as this PR has passed pipeline tests earlier and pipeline now appear stuck

PR #695, opened 2018-11-18
better error msg for non-supported ECC controller entry in IP_LAYOUT …

maxzhen:
…section


src/runtime_src/driver/xclng/drm/xocl/subdev/icap.c
```diff
@@ -1826,12 +1826,20 @@ static int icap_download_bitstream_axlf(struct platform_device *pdev,
 
 		if(ip->m_type == IP_DDR4_CONTROLLER) {
 			uint32_t memidx = ip->properties;
+
 			if (!memtopo || ip->properties >= memtopo->m_count ||
-				!memtopo->m_mem_data[memidx].m_used) {
```
sonals: We dropped check for if mem topology exists (!memtop). Is it redundant now?

sonals: Actually the check is still there :-)

PR #697, opened 2018-11-19
CR-1016834 [Redhat only] QDMA C2H AIO pending request causes closing …

houlz0507:
…queue fail silently


NaAl-Xlnx:
retest this please

PR #699, opened 2018-11-19
fix temp overflow

chienwei-lan:


chienwei-lan:
http://jira.xilinx.com/browse/CR-1017061?filter=-1

PR #704, opened 2018-11-20
Xbutil fix

AShivangi:
xbutil top -s gives the stream


rradjabi:
Adding "do not merge" label until Shivangi can build and test.

PR #713, opened 2018-11-21
keep dna with blank certificate

chienwei-lan:
http://jira.xilinx.com/browse/CR-1017137?filter=-1

We should wipe out everything if xclbin download fails, 

however, if we remove dna device, there is no way for DNA IP users to retrieve dna.

We treat it as invalid dna xclbin but keep dna sub device if and only if there is a blank certificate for development purpose.

and add dna entry to json


chienwei-lan:
please retest this

hcneema:
please retest

PR #720, opened 2018-11-26
stop checking memory type for ECC controller

maxzhen:
The memory type in memory topology section is incorrect on U280. The HW team requires driver not to check the memory type for ECC controller.


maxzhen:
close based on management decision.

PR #733, opened 2018-11-29
support Twitch MFG board

maxzhen:


cellery:
Max what does the "twitch" string for in the table? Is this just for printouts with xbutil flash scan?

maxzhen:
The "twitch" is the board name as in the VBNV DSA string. So that xbutil can pick the right DSA to flash if -a is used.

cellery:
Okay Jeff confirmed "twitch" is the name of the board in the dsa so looks good to me

PR #738, opened 2018-11-30
Create buffer object in drm_zocl_bo size instead of "drm_gem_cma_object" 

mamin506:
It is not safe to alloc a cma object then cast to drm_zocl_bo object.

This change was in SDx 2018.2. Seems when XRT moved to github, this change was lost.  (For fixed CR-1004460)


mamin506:
http://jira.xilinx.com/browse/CR-1017885 
This has been approved. Please merge it.

PR #740, opened 2018-11-30
Support P2P for 416+ kernel and RHEL7.6

chienwei-lan:
solved compilation error on linux kernel version 4.16+ and RHEL/CentOS 7.6


src/runtime_src/driver/xclng/drm/xocl/xocl_drv.h
```diff
@@ -111,15 +111,22 @@ static inline bool uuid_is_null(const xuid_t *uuid)
...
+#define RHEL_P2P_SUPPORT_76  0
 #endif
 #else
-#define RHEL_P2P_SUPPORT  0
+#define RHEL_P2P_SUPPORT_74  0
+#define RHEL_P2P_SUPPORT_76  0
 #endif
 
+
+#define RHEL_P2P_SUPPORT (RHEL_P2P_SUPPORT_74 | RHEL_P2P_SUPPORT_76)
```
keryell: Just curious: what is the motivation for testing for some distribution instead of just looking at the kernel version or a kernel feature?

maxzhen: Instead of upgrading the whole kernel (and bump up kernel version), RedHat/CentOS keep backporting individual fixes from latest kernel without bumping up their kernel version. Ubuntu will just upgrade the entire kernel (and bump up kernel version).

PR #742, opened 2018-12-03
Updated documentation for 2018.3 and cleaned up header files

sonals:


keryell:
Please use ReST syntax when you type the Sphinx documentation...

PR #749, opened 2018-12-04
get temp order by mem_topology

chienwei-lan:
get temperature order by mem_topology,

2019.1 CR-1017537


stsoe:
This PR broke "xbutil mem".  mem_access.h is not aware of mem_topology moved to icap.

PR #759, opened 2018-12-07
fix the bug where profile ignores the size argument in clEnqueueReadBuffer and always use the size of the buffer

tianhaoz95:
A CR was filed on profile timeline trace always report the same size no matter how much is actually migrated. 

Cause: the profiling plugin inside clEnqueueReadBuffer is only capturing the buffer object and getting the size from the buffer object.

Fix: also pass in the user-defined size and offset, and use the actual size in the report.


sonals:
ok to test

PR #763, opened 2018-12-10
qdma 2018.3 support

karenx-xilinx:
This patch is to upgrade the libqdma the latest qdma 2018.3 code base. And updated xocl driver to use the libqdma exported APIs.


maxzhen:
ok to test.

src/runtime_src/driver/xclng/drm/xocl/subdev/mm_qdma.c
```diff
@@ -253,18 +245,41 @@ static ssize_t qdma_migrate_bo(struct platform_device *pdev,
 		return -EIO;
 	}
 	sgt->nents = nents;
+	
+	ret = 0;
+	for_each_sg(sgt->sgl, sg, sgt->nents, i)
```
houlz0507: Why do we need to go through sgl to compare the length here?  The sgl could be really long. I do not think we need to re-calculate the length.

karenx-xilinx: Yes, we can remove it. I put it there for cross-check.

src/runtime_src/driver/xclng/drm/xocl/subdev/str_qdma.c
```diff
@@ -339,29 +342,58 @@ static ssize_t stream_post_bo(struct str_device *sdev,
 	drm_gem_object_reference(gem_obj);
 	xobj = to_xocl_bo(gem_obj);
 
-	memset(&wr, 0, sizeof (wr));
-	wr.write = write;
-	wr.len = len;
-	wr.sgt = xobj->sgt;
-	wr.eot = (header->flags & XOCL_QDMA_REQ_FLAG_EOT) ? true : false;
+	io_req = kmem_cache_alloc(req_cache, GFP_KERNEL);
```
houlz0507: I am not sure how efficient kmem_cache_alloc is. And this is critical path. Should we avoid kmem_cache_alloc() on critical path? The stream_async_req structure should be able to be pre-allocated during queue creation.

src/runtime_src/driver/xclng/drm/xocl/subdev/str_qdma.c
```diff
@@ -428,65 +459,101 @@ static ssize_t queue_rw(struct str_device *sdev, struct stream_queue *queue,
...
+		return -EFAULT;
 	}
 
-	memset(&wr, 0, sizeof (wr));
-	wr.write = write;
-	wr.len = sz;
-	wr.sgt = unmgd.sgt;
-	wr.eot = (header.flags & XOCL_QDMA_REQ_FLAG_EOT) ? true : false;
-
+	io_req = kmem_cache_alloc(req_cache, GFP_KERNEL);
```
houlz0507: Same as previous comment, I think we should avoid kmem_cache_alloc on critical path if possible.

karenx-xilinx: Yes, we can pre-alloc 2x of the queue depth at the Q creation time.

src/runtime_src/driver/xclng/drm/xocl/userpf/qdma.c
```diff
@@ -123,10 +123,10 @@ static int xocl_user_qdma_probe(struct pci_dev *pdev,
 
 	conf = &qd->dev_conf;
 	memset(conf, 0, sizeof(*conf));
-	conf->poll_mode = 0;
 	conf->pdev = pdev;
 	conf->intr_rngsz = QDMA_INTR_COAL_RING_SIZE;
-	conf->master_pf =  PCI_FUNC(pdev->devfn);
+	conf->master_pf = 1;
```
houlz0507: Why master_pf is forced to 1? Using PCI_FUNC(pdev->devfn) should be better.

karenx-xilinx: It is a bit field, to indicate to libqdma if this dev is the qdma master pf (i.e., controls the qdma global csr space). Since there is only 1 qdma pf, so it is hard coded to 1.

PR #764, opened 2018-12-10
perf mon using XDP & minor fixes

hasansayani-xilinx:
1. Added performance monitoring using XDP. Tested for host side. New architecture under implementation. APIs used will not change (ref: PaulS)
2. Added a fix for correct resource labelling


hasansayani-xilinx:
Hey guys (@sonals, @sarab96), please can we have this pulled in?
Thanks!

sarab96:
@hcneema  Please merge this

src/xma/src/xmaapi/CMakeLists.txt
```diff
@@ -38,6 +38,7 @@ target_link_libraries(xmaapi
   stdc++
   xml2
   yaml
+  xdp
```
stsoe: I don't think it is a good idea to introduce a link time dependency on xdp.
Dependencies should be other way around, xdp can depend in xrt, but not vice versa.
Look at runtime loading and symbol look-up similar to what is done for OCL.

hasansayani-xilinx: Hey Soren, this is change is part of XMA, and does not belong to core XRT. Do you suggest we do not do this for XMA as well?

hasansayani-xilinx:
closing this pull request and will create a new one to clear out stale refs.

PR #768, opened 2018-12-11
Remove atomics that are not needed or flat out wrong

### LINKED ISSUE #766: Needs to revisit calls to atomic_xxx APIs in driver
In our drivers, we call atomic_xxx() APIs in multiple places. After examination of a few cases, I can see that some of the usage of these APIs are problematic. I'd like to list some good and not-so-good use cases that I know of. We want to double check all cases in our driver to make sure we don't have lurking bugs by calling atomic kernel APIs.

1. In our code, we sometimes use mutex together with atomic API. This is OK, but not efficient. We want to use one or the other, not both. We prefer atomic API to mutex, if we *have to* use one.

2. We don't want to call atomic API, if not necessary. Atomic API is to be used when the operand may be accessed at the same time by more than one thread at any point of time. This is to avoid half updated variable. We need to use atomic API in the above case *and* if
a) the address of the variable is NOT naturally aligned.
b) the operation cannot be done atomically to a memory location.
Case a) usually is resulted from casting one data structure to a different one. E.g., you cast a char array to a data structure. In this case, the data structure has higher alignment requirement than char array and compiler does not see that cast coming. Case b) may be a inc/dec operation, which needs to perform read-modify-write action. E.g., you may need to call atomic_inc() to increment a counter. The compiler should be able to properly align an explicitly defined (vs cast one) data structure, so there is no need to call atomic API to access it's members. The kernel memory allocator should hand out 64-bit aligned address so that the buffer allocated is suitable for any data structure. So, again, no need to use atomic API for dynamically allocated data structures.

3. In below code, you should NOT use atomic API, but need a mutex:
if (atomic_read(xxx))
    do_something();
The atomic API does not make the check and do_something an atomic operation automatically. A mutex is needed to wrap around them to make this multi-thread safe. It maybe that the race may never happen based on where these code are used. But, instead of spending time analyzing each individual case, we should employ a mutex and be done with it. This is going to be safer and the code will be easy to maintain in the future.

stsoe:
Fixes #766

### DIFF
--- a/src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
+++ b/src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
@@ -35,9 +35,9 @@ xuid_t uuid_null = NULL_UUID_LE;
 #endif
 
 static int
-xocl_client_lock_bitstream(struct xocl_dev *xdev, struct client_ctx *client)
-{
-	if (atomic_read(&client->xclbin_locked))
+xocl_client_lock_bitstream_nolock(struct xocl_dev *xdev, struct client_ctx *client) {
+
+	if (client->xclbin_locked)
 		return 0;
 
 	if (xocl_icap_lock_bitstream(xdev, &xdev->xclbin_id,pid_nr(task_tgid(current)))) {
@@ -45,23 +45,29 @@ xocl_client_lock_bitstream(struct xocl_dev *xdev, struct client_ctx *client)
 		return 1;
 	}
 
-//      Allow second process to use current xclbin without downloading
-//	if (uuid_is_null(&client->xclbin_id))
-//		uuid_copy(&client->xclbin_id,&xdev->xclbin_id);
-//	else
 	if (!uuid_equal(&xdev->xclbin_id,&client->xclbin_id)) {
 		userpf_err(xdev,"device xclbin does not match context xclbin, cannot obtain lock for process %d",
 			   pid_nr(task_tgid(current)));
 		goto out;
 	}
 
-	atomic_set(&client->xclbin_locked,true);
+	client->xclbin_locked=true;
 	userpf_info(xdev,"process %d successfully locked xcblin",pid_nr(task_tgid(current)));
 	return 0;
 
 out:
 	(void) xocl_icap_unlock_bitstream(xdev, &xdev->xclbin_id,pid_nr(task_tgid(current)));
 	return 1;
+}
+
+static int
+xocl_client_lock_bitstream(struct xocl_dev *xdev, struct client_ctx *client)
+{
+	int ret = 0;
+	mutex_lock(&client->lock);
+	ret = xocl_client_lock_bitstream_nolock(xdev,client);
+	mutex_unlock(&client->lock);
+	return ret;
 }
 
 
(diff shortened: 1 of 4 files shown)


src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
```diff
@@ -214,11 +220,6 @@ int xocl_ctx_ioctl(struct drm_device *dev, void *data,
 		// CU unlocked explicitly
 		--client->num_cus;
 		--xdev->ip_reference[args->cu_index];
-		if (!client->num_cus) {
```
maxzhen: Why not unlock here if you removed the last context?

src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
```diff
@@ -35,27 +35,23 @@ xuid_t uuid_null = NULL_UUID_LE;
 #endif
 
 static int
-xocl_client_lock_bitstream(struct xocl_dev *xdev, struct client_ctx *client)
-{
-	if (atomic_read(&client->xclbin_locked))
+xocl_client_lock_bitstream_helper(struct xocl_dev *xdev, struct client_ctx *client) {
```
maxzhen: _helper() -> _nolock() to better align the intention of this function?

src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
```diff
@@ -98,8 +104,8 @@ int xocl_execbuf_ioctl(struct drm_device *dev,
 	int numdeps;
 	int ret = 0;
 
-	if (atomic_read(&xdev->needs_reset)) {
-		userpf_err(xdev, "device needs reset, use 'xbsak reset -h'");
+	if (xdev->needs_reset) {
```
maxzhen: Don't need a lock?

stsoe: You mentioned that if the data is aligned it by default atomic.  This is a Boolean only.

src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
stsoe: Yes.

src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
maxzhen: This is the 3rd case mentioned in the issue: behavior changed based on value of a variable shared between multiple threads. This usually indicates that a lock is needed. It was not clear to me why we don't use a lock here. Let's have some more discussion when I'm in office this afternoon.

src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
stsoe: yes

maxzhen:
chat with Soren, changes look good.

stsoe:
Replacing with another PR

PR #776, opened 2018-12-12
Fix for 2kernelglobal bandwidth checking

haeseung81:
2kernelglobal has incorrect bandwidth threshold.
So, I fixed the bandwidth threshold.


sonals:
Should we merge this into 2018.3?

haeseung81:
If it's possible, yes.
The current bandwidth threshold is 40GB, which is not correct value.
So, without this fix, this test will always say fail.

sonals:
Can you close this PR and recreate another PR for 2018.3?
2018.3 -> master merging is automatically handled by XRT team.

haeseung81:
Ok. I'll do that.

PR #777, opened 2018-12-12
Remove atomics that are not needed or flat out wrong

### LINKED ISSUE #766: Needs to revisit calls to atomic_xxx APIs in driver

stsoe:
Fixes #766

Also fix xocl_client_release to properly call xocl_exec_destroy_client and let it reset the scheduler if necessary.

### DIFF
--- a/src/runtime_src/driver/xclng/drm/xocl/subdev/icap.c
+++ b/src/runtime_src/driver/xclng/drm/xocl/subdev/icap.c
@@ -2035,6 +2035,8 @@ static int icap_unlock_bitstream(struct platform_device *pdev, const xuid_t *id,
 			err = icap->icap_bitstream_ref;
 	} else {
 		err = icap_lock_unlock_peer_bitstream(icap, id, pid, false);
+		if (err==0)
+			xocl_exec_reset(xocl_get_xdev(pdev));
 	}
 
 	ICAP_INFO(icap, "proc %d try to unlock bitstream %pUb, ref=%d, err=%d",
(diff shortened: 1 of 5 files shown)


stsoe:
Hi Max.  Just a quick sanity check that my merge didn't mess up anything.  Passes unit tests, hanging test, and multi test, so probably ok.

PR #778, opened 2018-12-13
NULL string of drm driver date causes crash with drm_version()

houlz0507:


sonals:
ok to test

src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_drm.c
```diff
@@ -49,6 +49,8 @@
 #define DRM_DBG(fmt, args...)
 #endif
 
+char driver_date[9];
```
sonals: static qualifier?

sonals:
Also can we add a call to ioctl(fd, DRM_IOCTL_VERSION, &version) in xclOpen()?
Reference code is src/runtime_src/driver/zynq/user/shim.cpp

src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_drm.c
houlz0507: added static

houlz0507:
> Also can we add a call to ioctl(fd, DRM_IOCTL_VERSION, &version) in xclOpen()?
> Reference code is src/runtime_src/driver/zynq/user/shim.cpp

Added ioctl

PR #779, opened 2018-12-13
Yocto recipes and document update for 2018.3

mamin506:
1. Add XRT yocto recipes in src/platform/recipes-xrt
2. Fix reference issue in create_platforms.rst
3. Update yocto.rst


sonals:
ok to test

PR #782, opened 2018-12-14
Make the hal comments great again

tianhaoz95-xilinx:
👍


maxzhen:
ok to test

sonals:
Should we merge this into 2018.3?

maxzhen:
I don't think we need this for 2018.3.

PR #788, opened 2018-12-15
not allow access to microblaze before reste

maxzhen:
The logic in safe_read32/write32() is broken. It does not protect microblaze from being accessed before we load image on it. This will result in firewall trip during early boot when driver is loaded.


maxzhen:
retest this please

PR #789, opened 2018-12-16
not allow access to microblaze before reste

maxzhen:
cherry pick pr#788 into 2018.3.


sonals:
Awesome! You nailed it :-)

PR #794, opened 2018-12-17
Fix to xclLoadXclBin for AWS

rradjabi:
Submit IOCTL even when the xclbin-to-load is already present in the device.

This PR will need to be merged into 2018.3 and Master.


stsoe:
ok to test

stsoe:
Fixes part 2 of http://jira.xilinx.com/browse/CR-1018821

PR #797, opened 2018-12-18
the request was not cleaned up properly when it is synchronous

karenx-xilinx:
Fixed a bug where the request struct was not cleaned up properly when the request is synchronous.


maxzhen:
ok to test

PR #799, opened 2018-12-18
fixed cleaning up of the synchronous qdma streaming mode requests

karenx-xilinx:
Fixed a bug for the QDMA streaming mode, when the request is submitted synchronously the request was not cleaned out of the pending list.


maxzhen:
ok to test

PR #802, opened 2018-12-18
Run launch_hw_emu.sh to support versal platforms for hw_emu

chvamshi-xilinx:
This change is needed for 2019.1/HEAD. 
Calling launch_hw_emu.sh instead of simulate.sh. Calling simulate.sh if launch_hw_emu.sh is not there


chvamshi-xilinx:
I am assuming master is now 2019.1. This change shouldn't go to 2018.3

PR #806, opened 2018-12-18
Check return value of exec_buf

stsoe:
Throw on error at lowest xrt::hal level, that way upstream doesn't have to check.


stsoe:
Good point, no.   But then again, exec_wait should only be called if exec_buf return success.  poll() ends up calling poll_client (in mb_scheduler.c), there is no check in this function that would create a negative return code.    I guess it could check the client context to see if it has locked this xclbin?

stsoe:
Close and update exec_wait also

PR #807, opened 2018-12-18
xocl_read_axlf_helper() may unlock xclbin before locking it

maxzhen:
goto done will just blindly perform a xclbin unlock, which is not good for these two failure cases.


maxzhen:
This fix itself does not address any bug that we know today. Soren's debugging a bigger issue with multiple processes support. Let's wait for his debug result and see if this PR is needed for 2018.3.

PR #808, opened 2018-12-19
Check return value of exec_buf and exec_wait

stsoe:
Throw on error.


stsoe:
This is needed for MP support.  I will cherry-pick to 2018.3 also along with PR #807

PR #810, opened 2018-12-19
Qdma 2018.3: force sync'ing of per Q irq_en with the device-wide interrupt mode settings and add context dump when i/o fails

karenx-xilinx:


stsoe:
ok to test

karenx-xilinx:
Hi, Lizhi.
Yes, I understand your concern.  I am planning to reduce the debug messages after more extensive testing is done. Right now, The dumping of contexts would greatly help & speed up the debugging.
Thanks,
Karen

PR #816, opened 2018-12-19
Add sdx_fast tag for the sprite

haeseung81:
Add sdx_fast tag in the sdainfo.yml files for the sprite run


sonals:
ok to test

sonals:
@haeseung81 should we merge into 2018.3 and auto merge into master?

PR #817, opened 2018-12-19
libqdma: fixed return value when the stm eot was received before requ…

karenx-xilinx:
…est was posted.

The return value didn't take eot into account.


maxzhen:
ok to test

PR #820, opened 2018-12-20
Copying sdx_emulator log to run directory

chvamshi-xilinx:
Hi All,

This change contains a fix to copy sdx_emulator log to current run directory in emulation.

This change is needed only in 2019.1/HEAD

Thanks
Vamshi Krishna


maxzhen:
ok to test

PR #827, opened 2018-12-21
Fix multi-process handling in kernel driver

stsoe:
Don't let icap_lock_bitstream override the icap bitstream uuid.  The icap bitstream is the current downloaded bitstream, it can be changed only if a new bitstream is downloaded.

Scenario was P1 (process 1) downloads bitstream through mgmt, P2 changes the bitstream before P1 starts using it; icap bitstream is now that of P2.  When P1 next wants to lock its bitstream it should fail even when P2's bitstream is currently unused.

Fixed few unrelated unlock / free errors

### DIFF
--- a/src/runtime_src/driver/xclng/drm/xocl/subdev/icap.c
+++ b/src/runtime_src/driver/xclng/drm/xocl/subdev/icap.c
@@ -872,7 +872,7 @@ static uint64_t icap_get_section_size(struct icap *icap, enum axlf_section_kind 
 			size = sizeof_sect(icap->connectivity, m_connection);
 			break;
 		default:
-			break;		
+			break;
 	}
 
 	return size;
@@ -1611,6 +1611,7 @@ static int icap_download_bitstream_axlf(struct platform_device *pdev,
 	struct ip_layout* layout = NULL;
 	struct mem_topology* memtopo = NULL;
 	bool dna_check = false;
+	int pid = pid_nr(task_tgid(current));
 	int i;
 
 	/* Can only be done from mgmt pf. */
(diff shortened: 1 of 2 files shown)


src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
```diff
@@ -62,9 +62,11 @@ static int
 xocl_client_lock_bitstream(struct xocl_dev *xdev, struct client_ctx *client)
 {
 	int ret = 0;
-	mutex_lock(&client->lock);
+	mutex_lock(&client->lock);         // protect current client
+	mutex_lock(&xdev->ctx_list_lock);  // protect xdev->xclbin_id
 	ret = xocl_client_lock_bitstream_nolock(xdev,client);
 	mutex_unlock(&client->lock);
+	mutex_unlock(&xdev->ctx_list_lock);
```
maxzhen: mutex unlock should be done in the reverse order of locking for nested mutex.

src/runtime_src/driver/xclng/drm/xocl/subdev/icap.c
```diff
@@ -2021,21 +2024,19 @@ static int icap_lock_bitstream(struct platform_device *pdev, const xuid_t *id,
 	mutex_lock(&icap->icap_lock);
 
 	if (ICAP_PRIVILEGED(icap)) {
-		if (!icap_bitstream_in_use(icap, pid) ||
-			uuid_equal(id, &icap->icap_bitstream_uuid)) {
-			uuid_copy(&icap->icap_bitstream_uuid, id);
+		if (uuid_equal(id, &icap->icap_bitstream_uuid))
```
maxzhen: Yeah, you don't want to change ID while locking/unlocking it. The original code looks very strange. Can't remember what I was thinking back then, but thanks for fixing it :-).

src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_ioctl.c
stsoe: Yes, of course; my bad.

PR #835, opened 2019-01-01
kernel arg connectivity deduction

stsoe:
Multiple changes to support clSetKernelArg deducing device buffer
memory connectitivity.

Kernel is constructed with all device CUs based on loaded program.
The kernel CUs are dynamically filtered based on connectivity
requirements of kernel arguments.

The kernel CUs that remain after filtering are sticky in the sense
that when reusing a kernel object, only the prior filtered resulting
CUs are avaiable, if other CUs are required, a new kernel object should
be constructed.

This commit also does away with the memory extention where
{kernel,argidx} is specified, Though still supported, clCreateBuffer
simply calls clSetKernelArg under the hood when the the buffer carries
the now obsolete extension.

The deduced memory connectivity of a cl_mem buffer is stored as a data
member on the memory object.  This is used for the cases where the
targeted device is unknown until first clEnqueue operation involving
the buffer.

### DIFF
--- a/src/runtime_src/xocl/api/api.h
+++ b/src/runtime_src/xocl/api/api.h
@@ -20,7 +20,7 @@
 #include <CL/opencl.h>
 
 /**
- * Extern declration of xocl::api function that are used 
+ * Extern declration of xocl::api function that are used
  * in implementation of OCL api functions.
  *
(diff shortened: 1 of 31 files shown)


stsoe:
Needs additional fixes

PR #838, opened 2019-01-03
Remove python-cffi from xrtdeps.sh

rradjabi:
Sonal's PR Xilinx/XRT#832 removes the dependency on cffi, so this can be removed from xrtdeps.sh.


sonals:
ok to test

abalasa:
retest this please

PR #837, opened 2019-01-03
add hal API for getting debug and profile deivce information

tianhaoz95:
This API will get the part of device information that debugger and profile need from `scan.h`.

I also plan to add it to xbutil dump so that GUI can make use of it when doing profiling or debugging.

The information will look like this in a host code:
```bash
2019-01-03 09:03:24,941 INFO [default] 	 Showing device[0] info:
2019-01-03 09:03:24,941 INFO [default] 		 Device Index: 0
2019-01-03 09:03:24,941 INFO [default] 		 User Instance: 129
2019-01-03 09:03:24,941 INFO [default] 		 NIFD Instance: 0
2019-01-03 09:03:24,941 INFO [default] 		 Management Instance: 1024
2019-01-03 09:03:24,941 INFO [default] 		 Device Name: /dev/dri/renderD129
2019-01-03 09:03:24,941 INFO [default] 		 Sysfs Name: /sys/bus/pci/devices/0000:04:00.1
2019-01-03 09:03:24,941 INFO [default] 		 NIFD Name: /dev/nifd0
```

PR #840, opened 2019-01-03
Add hal API to give generic access to sysfs for each device

tianhaoz95:
In debug and profile refactoring, the knowledge about the debug ip layout will be moved to a higher level and leave hal layer cleaner, so access to sysfs will be needed.

This API takes subdevice name, entry name, either read_all or a certain size and read the content from sysfs.

example usage:
```cpp
xclSysfsQuery query;
strcpy(query.subdev, subdev.c_str());
strcpy(query.entry, entry.c_str());
query.read_all = read_all;
query.size = size;
int err = xclReadSysfs(device_handle, query, data);
```

>note: if size is non-zero and read_all is true the code will return -1 to prevent potential uninitialized query.


maxzhen:
ok to test

abalasa:
retest this please

tianhaoz95:
just resolved a conflict introduced by enum clean up. please build again

tianhaoz95:
I think that will also work for our purpose. The full path of sysfs is exposed in my other PR #837. If that is the way we want to give access to sysfs. This PR can be ignored.

maxzhen:
Are you sure about that? It seems PR#837 only retrieves path up to the device, not up to the entry, right? I'm talking about an API to return full abs path up to the entry...

PR #847, opened 2019-01-05
Add Context Creation to Github Tests

rradjabi:
xclOpenContext and xclCloseContext are needed in the test cases. This makes most, but not all, the tests pass using the xclbins from the TA. 

Removed duplicate copies of utils.h, and all use the common utils.h now.


maxzhen:
ok to test

tests/xrt/02_simple/main.cpp
```diff
@@ -72,14 +72,17 @@ static void printHelp()
 
 
 
-static int runKernel(xclDeviceHandle &handle, uint64_t cu_base_addr, size_t alignment, bool ert, bool verbose, int first_mem)
+static int runKernel(xclDeviceHandle &handle, uint64_t cu_base_addr, size_t alignment, bool ert, bool verbose, int first_mem, uuid_t xclbinId)
 {
+    if(xclOpenContext(handle, xclbinId, (first_mem-1), true))
```
sonals: xclOpenContext() requires IP index as the third argument. first_mem-1 does not make sense here.

rradjabi: Thanks, I'll make the change to pass cu_index into runKernel. Also needs to be used as an argument in xclCloseContext().

PR #853, opened 2019-01-08
move aio cancellation work to libqdma context.

karenx-xilinx:
handle the actual work of cancelling in the libqdma context.

### DIFF
--- a/src/runtime_src/driver/xclng/drm/xocl/lib/libqdma/qdma_descq.c
+++ b/src/runtime_src/driver/xclng/drm/xocl/lib/libqdma/qdma_descq.c
@@ -1111,9 +1111,60 @@ int qdma_descq_prog_stm(struct qdma_descq *descq, bool clear)
 }
 #endif
 
+static void descq_proc_request_cancel(struct qdma_descq *descq)
+{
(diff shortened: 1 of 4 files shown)


src/runtime_src/driver/xclng/drm/xocl/lib/libqdma/qdma_descq.c
```diff
@@ -1111,9 +1111,60 @@ int qdma_descq_prog_stm(struct qdma_descq *descq, bool clear)
...
+	}
+
+unlock:
+	unlock_descq(descq);
+}
+
 void qdma_descq_service_cmpl_update(struct qdma_descq *descq, int budget,
 				bool c2h_upd_cmpl)
 {
+	descq_proc_request_cancel(descq);
```
houlz0507: when will qdma_descq_service_cmpl_update() be called? if there is not any write back generated, is this still called?

karenx-xilinx: 1. when a cancel is requested
2. When interrupt/polling comes

PR #859, opened 2019-01-09
Xrt log msg API implementation (2018.3)

sarab96:
@sonals Please check

### DIFF
--- a/src/runtime_src/driver/xclng/xrt/user_gem/shim.cpp
+++ b/src/runtime_src/driver/xclng/xrt/user_gem/shim.cpp
@@ -37,8 +37,6 @@
 #include "driver/include/xclbin.h"
 #include "scan.h"
 #include "driver/xclng/xrt/util/message.h"
-#include <cstdio>
-#include <cstdarg>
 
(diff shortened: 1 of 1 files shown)


src/runtime_src/driver/xclng/xrt/user_gem/shim.cpp
```diff
@@ -331,10 +333,36 @@ int xocl::XOCLShim::pcieBarWrite(unsigned int pf_bar, unsigned long long offset,
...
-    //This is TODO
-    xrt_core::message::send((xrt_core::message::severity_level)level, format);
+    va_list args1;
+    va_start(args1, format);
+    int len = std::vsnprintf(nullptr, 0, format, args1);
+    va_end(args1);
+    
+    {
+        if (len < 0) {
+            goto handle_errors;
```
sonals: No goto in C++ code. Please move the code inside handle_errors label here.

sarab96: Done

PR #863, opened 2019-01-10
Fix for 100_ert_ncu: host code and utils.hpp

haeseung81:
Fix for 100_ert_ncu: update the host code and utils.hpp to check the used memory bank.


maxzhen:
ok to test

PR #865, opened 2019-01-10
enhance xbutil dump with debug and profile device information

tianhaoz95:
> Note: for some reason I can't edit reviewers and milestone. The first admin that read this please add jvillar and other people that may have concern about this change into the reviewers and set the milestone to 2019.1

Some external system like sdx_server or GUI needs to manage or pass around device information, for which an entry is added into `xbutil dump` so that other processes can access it.

The new entry in `xbutil dump`:

```bash
    "debug_profile": {
        "device_info": {
            "error": "0",
            "device_index": "0",
            "user_instance": "129",
            "mgmt_instance": "1024",
            "nifd_instance": "0",
            "device_name": "\/dev\/dri\/renderD129",
            "nifd_name": "\/dev\/nifd0"
        }
    }
```


stsoe:
ok to test

src/runtime_src/driver/xclng/tools/xbutil/xbutil.h
```diff
@@ -700,6 +701,30 @@ class device {
...
+         * e.g. sdx_server, hardware_sercer, GUI, etc
+         */
+        xclDebugProfileDeviceInfo info;
+        int err = xclGetDebugProfileDeviceInfo(m_handle, &info);
+        unsigned int device_index = (unsigned int)info.device_index;
+        unsigned int user_instance = (unsigned int)info.user_instance;
+        unsigned int mgmt_instance = (unsigned int)info.mgmt_instance;
+        unsigned int nifd_instance = (unsigned int)info.nifd_instance;
+        std::string device_name = (std::string)info.device_name;
+        std::string nifd_name = (std::string)info.nifd_name;
```
stsoe: Let's get rid of all explicit casts.  In this case they are not necessary.  If a cast is necessary, the use proper C++ conversion casts, e.g. static_cast.  Is there any reason to save the info data in a variable?

tianhaoz95-xilinx: Just pushed a new one to remove the complexity and resolve conflict with newly added p2p

PR #866, opened 2019-01-10
Adding the python binding testcase for 04_swizzle

haeseung81:
@rradjabi This is the python binding testcase for 04_swizzle


maxzhen:
ok to test

PR #868, opened 2019-01-11
p2p force reassign iomem, driver remove, close protection

houlz0507:


sonals:
ok to test

houlz0507:
retest this please

NaAl-Xlnx:
retest this please

houlz0507:
Hi Max,
I thought about that. It seems not working. the ioctl for example can happen concurrently with remove. so considering this scenario
remove()
     -> cv wait passed
    -> remove success
ioctl comes in

And yes, we can discuss better way to handle this.

maxzhen:
Had a lengthy discussion with Lizhi about this PR. We concluded that the current implementation for protecting driver from crashing during removal is not efficient and does not cover all race cases. We have found a better solution and Lizhi will implement it in a separate PR. I'm merging this one so that the resizing BAR functionality is complete in master.

PR #869, opened 2019-01-11
Reset sector select before trying to flash mcs files.

cellery:


sonals:
ok to test

sonals:
Already applied the change to 2018.3. Will auto merge this to master and 2018.2_XDF from 2018.3. Do not need separate PR for each branch.

PR #870, opened 2019-01-12
Reset sector select before trying to flash mcs files.

cellery:


sonals:
Already merged to 2018.2_XDF with cherry-pick from 2018.3.

PR #873, opened 2019-01-12
kernel arg connectivity deduction

stsoe:
Multiple changes to support clSetKernelArg deducing device buffer
memory connectivity.

Kernel is constructed with all device CUs based on loaded program.
The kernel CUs are dynamically filtered based on connectivity
requirements of kernel arguments.

The kernel CUs that remain after filtering are sticky in the sense
that when reusing a kernel object, only the prior filtered resulting
CUs are available, if other CUs are required, a new kernel object should
be constructed.

This commit also does away with the memory extension where
{kernel,argidx} is specified. Though still supported, clCreateBuffer
simply calls clSetKernelArg under the hood when the the buffer carries
the now obsolete extension.

### DIFF
--- a/src/runtime_src/xocl/api/api.h
+++ b/src/runtime_src/xocl/api/api.h
@@ -20,7 +20,7 @@
 #include <CL/opencl.h>
 
 /**
- * Extern declration of xocl::api function that are used 
+ * Extern declration of xocl::api function that are used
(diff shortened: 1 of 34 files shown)


stsoe:
Merging this.  Verified with 300+ board tests.

PR #877, opened 2019-01-15
Support kernel version 4.19 in zocl and fix a cmake issue in cmake 3.12

mamin506:
1. PetaLinux will update to kernel version 4.19 in 19.1 release. The release_page() API changed since kernel 4.15.
2. Library impl has header files only. Cmake 3.12 complains not able to determine linker language. Update CMakeList.txt in impl/, issue fixed.
3. Update ocl-icd license in the recipe to avoid warning.


sonals:
ok to test

PR #881, opened 2019-01-18
Rename xclbinid to xclbinuuid for AWS F1

rradjabi:
Userspace fix for AWS F1: 

'xclbinid' on sysfs has moved to 'xclbinuuid' in 2018.3.


sonals:
ok to test

PR #891, opened 2019-01-23
Added support to allocate cacheable CMA

larry9523:
This PR adds support to allocate cacheable BO backed by CMA buffer. It includes 3 main changes/enhancement

1) Add a new interface file xclhal2_mem.h. Currently, move the legacy BO flags to this file and also add new BO flags. Eventually, those new flags will be unified by both x86 and embedded code. But this PR only defines the flags and modify all zocl driver to use them.

2) Modify zocl mmap() entry point zocl_mmap() interface to map the BO as cacheable to user application if user sets CACHEABLE flag when creating BO.

3) Enhance the zocl_sycn_bo_ioctl to actually invalidate/flush read/write cache. It used to be NOP on embedded system because our BO memory used to be always coherent.

### DIFF
--- a/src/runtime_src/driver/include/CMakeLists.txt
+++ b/src/runtime_src/driver/include/CMakeLists.txt
@@ -9,7 +9,8 @@ set(XRT_HEADER_SRC
   xcl_app_debug.h
   xcl_macros.h
   xcl_app_debug.h
-  xclperf.h)
+  xclperf.h
+  xclhal2_mem.h)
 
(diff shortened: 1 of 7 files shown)


larry9523:
@mamin506, @maxzhen , @sonals please take a close look at the changes and check if this could possibly break anything. Thanks!

larry9523:
@stsoe please take a look too. Thanks!

PR #893, opened 2019-01-23
Provide functionality to override XMA default configuration file path through ENV variable.

abhisawa:
Currently, we need the ability to target particular FPGA device with any transcoding application which calls `xma_initialize` without touching default configuration file `/var/tmp/xma_cfg.yaml` which will be reserved to use while serving production traffic. We need the ability to pass the configuration file path to `xma_initialize` through environment variable to override defaults in `XMA_CFG_DEFAULT`. Priority of config file path lookup will be in following descending order

- path provided by `xma_initialize(cfgfile)`
- the path provided by environment variable `XMA_CFG_OVERRIDE`
- path defined by `XMA_CFG_DEFAULT` which is `/var/tmp/xma_cfg.yaml`

This way I can run ffmpeg or any equivalent transcoding in two different ways 
```
# this will target FPGA devices according to device_id_map[]  defined in  /var/tmp/xma_cfg.yaml

~$ ffmpeg <transcoding options>


# to schlep transcoding to 2nd FPGA device, 
# create config with device_id_map [2] in /tmp/test_2nd_FPGA_device.yaml and 
# execute ffmpeg in following manner.
 
~$ XMA_CFG_OVERRIDE="/tmp/test_2nd_FPGA_device.yaml" ffmpeg <transcoding options>

```

If we need thread safety in `xma_initialize` then we can replace `get_env` with `secure_getenv`

### DIFF
--- a/src/xma/src/xmaapi/xmaapi.c
+++ b/src/xma/src/xmaapi/xmaapi.c
@@ -36,7 +36,8 @@ int32_t xma_initialize(char *cfgfile)
     bool    rc;
 
     if (!cfgfile)
-        cfgfile = XMA_CFG_DEFAULT;
+        if (!(cfgfile = getenv("XMA_CFG_OVERRIDE")))
+            cfgfile = XMA_CFG_DEFAULT;
 
     g_xma_singleton = malloc(sizeof(*g_xma_singleton));
     memset(g_xma_singleton, 0, sizeof(*g_xma_singleton));


maxzhen:
ok to test

jnicholsXilinx:
I don't believe this commit will work because two or more configurations files cannot be used at the same time.  The reason for this is because only one shared memory database is present that maintains a common view of all available resources.  As a result, only one of the configuration files will take affect, most likely the last one provided, causing unexpected results.

I believe that adding the requested behavior is more complex than providing separate configuration files.  The configuration file mechanism will have to be replaced with a daemon that can manage the devices available on a per user or group basis.  In addition, changes are needed in the resource manager database to support this capability.

abhisawa:
@jnicholsXilinx made me realize that the approach is not feasible to get the desired outcome.

PR #895, opened 2019-01-23
fix XRT reporting temp

chienwei-lan:
it's going to fix the xbutil query temperature report.

before this change, both FPGA and fan temperature are 0.(2 digits divided by 1000)


maxzhen:
ok to test

PR #898, opened 2019-01-24
sdainfo.yml update to add uuid library option & fix host code for 15_buffer_size

haeseung81:
sdainfo.yml update to add uuid library option & fix host code for 15_buffer_size


stsoe:
ok to test

PR #901, opened 2019-01-25
master branch XMA changes - 2019-01-24

hasansayani-xilinx:
-Check for existence of default directory path for cfg file
-Add flag to indicate resources have been programmed
-Close possible race condition in init of shm db and mutex


maxzhen:
ok to test

PR #903, opened 2019-01-25
Merging master to 2018.3

haeseung81:
Merging xrt testcases from the master branch

### DIFF
--- a/tests/xrt/00_hello/main.cpp
+++ b/tests/xrt/00_hello/main.cpp
@@ -14,9 +14,6 @@
  * under the License.
  */
 
-// Copyright 2017 Xilinx, Inc. All rights reserved.
-
-
 #include <getopt.h>
 #include <iostream>
 #include <stdexcept>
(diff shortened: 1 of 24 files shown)


maxzhen:
ok to test

tests/xrt/00_hello/main.cpp
```diff
@@ -14,9 +14,6 @@
  * under the License.
```
rradjabi: Move copyright date to -2018, then it matches the file in master/2019.1

tests/xrt/02_simple/main.cpp
```diff
@@ -14,8 +14,6 @@
  * under the License.
```
rradjabi: Bump copyright to -2018

rradjabi:
@haeseung81 

1) Can you merge the changes from #847 so that these test files match more closely? 
2) In #847, I removed utils.h from inside the individual test folders, you should do the same and update the common utils.h, and have each test case include this version. I'm not familiar with the YML file, but it looks like that's what's going on.

You can discus with me before making those changes if you like.

PR #904, opened 2019-01-25
cloud refactoring

chienwei-lan:
decouple user func and mgmt func. They only communicate with each other by mailbox request.
For example:

xbutil reclock no longer goes to mgmt directly, instead it comes with an ioctl that down to userpf driver , sending a mailbox request to mgmt to perform the reclock operation in mgmt driver.

same as xbutil program -p a.xclbin.


maxzhen:
ok to test

maxzhen:
Chien-Wei, do these failed cases also fail in base line test?

stsoe:
These are the same tests that fail for me; and always have.  I should remove them from my cached suite.

stsoe:
Chien-Wei, you should run the board-all+.sh  (with the plus :-) ) it has some 50 extra tests from board.dsv_onbrd.

chienwei-lan:
Forget to copy board.unit. and all 60 tests are PASSED after re-run board-all+.sh

PR #905, opened 2019-01-25
Kernel Upgrade Fix for F1

rradjabi:
Due to issues found by AWS F1 customers who upgraded the Linux kernel to 3.10.0-957, we have added driver support. This has been sanity tested on F1, but needs a full regression suite.


maxzhen:
ok to test

PR #906, opened 2019-01-26
Mark destination buffer resident when KDMA fills it.

stsoe:
KDMA fills dst buffer same as migrate_buffer does, hence the dst buffer is resident after KDMA is done even if host does explicitly migrate. CR1020795


stsoe:
test this please

stsoe:
retest this please

PR #907, opened 2019-01-27
fixing subbuffer issue in emulation

chvamshi-xilinx:
Hi,

This change contains a fix in emulation for subbuffer sync.

Thanks
Vamshi Krishna


stsoe:
ok to test

PR #911, opened 2019-01-28
Cage temp

chienwei-lan:
To solve:CR-1018702 No Temp value in XRT for QSFP0 & 1 Temps


maxzhen:
ok to test

maxzhen:
retest this please

maxzhen:
Exporting one temperature in one sysfs entry does not sound ideal. We probably should export an array of temperatures in one sysfs entry. We will clean up the code later for all temperature exported through sysfs.

PR #913, opened 2019-01-28
Bug Fixes

rozumx:
Bug Fixes

1) Updated xclbinutil error working regarding creating an output file.
2) Added support (via a CR and -deploy option) to support creating in both a development and deployment package in one session.
3) Fixed issue in the flow where the dsabin package directory (e.g., firmware) may not have been created.
4) Added more descriptive messages when creating both a deployment and development package.
5) Error out of the pkgdsa.sh script if xclbinutil tool fails.
6) For debian packages, added a revision value to the package name.  Note: This is already done for CentOS packages.

### DIFF
--- a/src/runtime_src/driver/include/xclbin.h
+++ b/src/runtime_src/driver/include/xclbin.h
@@ -127,7 +127,7 @@ extern "C" {
         BUILD_METADATA,
         KEYVALUE_METADATA,
         USER_METADATA,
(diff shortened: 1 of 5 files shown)


rozumx:
I have chatted with Eric regarding his code reviews.  At the moment, these changes are not required.  Going forward and in the next few weeks we will start incorporating them into the overall code flow.

PR #924, opened 2019-01-31
2018.3xrt

IshitaGhosh:
Fixes for Coverity issues


src/runtime_src/xdp/profile/rt_profile_writers.cpp
```diff
@@ -80,7 +83,7 @@ namespace XCL {
     const int maxLength = 1024;
     char buf[maxLength];
     ssize_t len;
-    if ((len=readlink("/proc/self/exe", buf, maxLength-1)) != -1) {
+    if (((len=readlink("/proc/self/exe", buf, maxLength-1)) != -1) && (len < maxLength)) {
```
stsoe: This does not seem right.   What if len>=maxLength, e.g. the input is truncated, where is the buffer null terminated?

src/runtime_src/xdp/profile/rt_profile_writers.cpp
```diff
@@ -1095,14 +1098,23 @@ namespace XCL {
     if(SummaryFileName != "") {
       assert(!Summary_ofs.is_open());
       SummaryFileName += FileExtension;
-      openStream(Summary_ofs, SummaryFileName);
+      try {
+        openStream(Summary_ofs, SummaryFileName);
+      } catch(std::runtime_error &err) {
```
stsoe: Is it really safe to continue to writeDocumentHeader when an exception was thrown?

src/runtime_src/xdp/profile/rt_profile_writers.cpp
```diff
@@ -1095,14 +1098,23 @@ namespace XCL {
...
 
     if (TimelineFileName != "") {
       assert(!Timeline_ofs.is_open());
       TimelineFileName += FileExtension;
-      openStream(Timeline_ofs, TimelineFileName);
+      try {
+        openStream(Timeline_ofs, TimelineFileName);
+      } catch(std::runtime_error &err) {
+        std::cout << " ERROR : " << err.what() << " SDAccel Timeline Trace." << std::endl;
+      }
```
stsoe: Is it really safe to continue to writeDocumentHeader when an exception was thrown?

src/runtime_src/xdp/profile/rt_profile_writers.cpp
stsoe: Ok, I see.  That's a false positive from Coverity then.

src/runtime_src/xdp/profile/rt_profile_writers.cpp
IshitaGhosh: Hi Soren,
writeDocumentHeader itself checks whether the passed ofstream is open. WriterI::openStream throws the exception when the ofstream could not be opened. So, handling the exception and generating error message should be sufficient. All the other methods using the ofsteam checks whether it is open.
Thanks!

src/runtime_src/xdp/profile/rt_profile_writers.cpp
IshitaGhosh: Hi Soren,
writeDocumentHeader itself checks whether the passed ofstream is open. WriterI::openStream throws the exception when the ofstream could not be opened. So, handling the exception and generating error message should be sufficient. All the other methods using the ofsteam checks whether it is open.
Thanks!

src/runtime_src/xdp/profile/rt_profile_writers.cpp
stsoe: Ok, but this seems a bit odd.   Catching is supposed to handle the exception, but it really doesn't, why not just rethrow?

src/runtime_src/xdp/profile/rt_profile_writers.cpp
stsoe: Ok, but this seems a bit odd.   Catching is supposed to handle the exception, but it really doesn't, why not just rethrow?

sonals:
ok to test

src/runtime_src/xdp/profile/rt_profile_writers.cpp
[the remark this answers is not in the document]
IshitaGhosh: This change is on 2018.3 branch.

PR #929, opened 2019-02-01
Opencl headers append

mamin506:
The OpenCL header recipe does not support install to the rootfs. This is because it contains header files only. Create a .bbappend recipe to fix this issue.


maxzhen:
ok to test

PR #932, opened 2019-02-01
Enhanced the "--info" option to take an optional file argument. 

rozumx:
Enhanced the "--info" option to take an optional file argument.  This argument if set will re-direct the info output to a file instead of the console.


stsoe:
ok to test

PR #956, opened 2019-02-07
support P2P mem bank auto deduce thru cl_mem_ext_ptr

maxzhen:
this also fixes a bug causing legacy XCL_MEM_DDR_BANKx flag not to work.


maxzhen:
@stsoe, please take a look at the 2nd commit.

maxzhen:
> 
> 
> Looks good to me. Did you run all cached HW board tests?

Yes, all passed (with the same 5 failed cases).

PR #957, opened 2019-02-08
Interrupt Tcl hook

mamin506:
This Tcl hook wire CU interrupt to PS.
It will automatically insert AXI Interrupt Controller IP to PL base on the CU numbers.
Each AXI interrupt controller can support up to 32 interrupt inputs.


maxzhen:
ok to test

PR #964, opened 2019-02-11
get mem_topology ptr after acquired ctx_list_lock

chienwei-lan:
1. get mem_topology & mm_stat[bank] after acquired ctx_list_lock.
http://jira.xilinx.com/browse/CR-1021924

2. remove meaningless sections (banks without m_size assigned in xclbin) from xbutil top

3. get bank memory usage directly from sysfs memstat_raw instead of calling xclGetUsageInfo
http://jira.xilinx.com/browse/CR-1019452


maxzhen:
ok to test

PR #969, opened 2019-02-13
region can't less than zero

chienwei-lan:


chienwei-lan:
CR-1009425

maxzhen:
ok to test

PR #974, opened 2019-02-14
Pulling xdp IP register reading from hal layer to xdp layer

tianhaoz95:
So far we have been reading xdp IPs within the hal and get the values through hal APIs. This makes it hard to manage the xdp implementation and hal api header file. This PR is the most basic structures to support register reading at xdp level.

> note: The code is not used anywhere now. The purpose of this PR is to provide the basic component for xdp refactoring for @pgschuey @hackwa

### DIFF
--- a/src/runtime_src/xdp/profile/device/profile_ip_access.cpp
+++ b/src/runtime_src/xdp/profile/device/profile_ip_access.cpp
@@ -47,11 +47,24 @@ ProfileIP::~ProfileIP() {
 } 
 
 void ProfileIP::request_exclusive_ip_access(xclDeviceHandle handle, int index) {
+    /**
+     * TODO: when the XRT implements the exclusive context hal API, this
+     * method should try to open a exclusive context here and set the
+     * exclusive flag to be true if successful and save the context object
+     * into a member variable. If it fails, it should show a proper warning
+     * and return gracefully.
+     */
     exclusive = true;
     return;
 }
 
 void ProfileIP::release_exclusive_ip_access(xclDeviceHandle handle, int index) {
+    /**
+     * TODO: when the XRT implements the exclusive context hal API, this
+     * method should close the previously requested exclusive context if
+     * one was request and then set the exclusive flag back to false. If
+     * it fails, it should show a proper warning and return gracefully.
+     */
     exclusive = false;
     return;
 }
--- a/src/runtime_src/xdp/profile/device/profile_ip_access.cpp
+++ b/src/runtime_src/xdp/profile/device/profile_ip_access.cpp
@@ -70,6 +70,11 @@ void ProfileIP::release_exclusive_ip_access(xclDeviceHandle handle, int index) {
 }
 
 void ProfileIP::map() {
+    /**
+     * TODO: so far we are asking the debug_ip_layout where the IP is. Once the XRT hal implements
+     * the function that maps the IP registers to user memory space, this method should be simplified
+     * to one function call to the hal API and saves the result in a mapped_address and set mapped flag.
+     */
     if (!exclusive) {
         return;
     }
@@ -111,6 +116,10 @@ void ProfileIP::map() {
 }
 
 void ProfileIP::unmap() {
+    /**
+     * TODO: This should use the unmapping API provided by XRT hal in
+     * the future. Now the API is not in place
+     */
     if (!exclusive || !mapped) {
         return;
     }
@@ -120,6 +129,11 @@ void ProfileIP::unmap() {
 }
 
 int ProfileIP::read(uint64_t offset, size_t size, void* data) {
+    /**
+     * TODO: so far we are using xclRead under the hood because the hal API that maps
+     * the IP is not ready yet. Once the API is ready, xclRead should be replaced by a
+     * memcpy from the mapped address with exception handling.
+     */
     if (!exclusive || !mapped) {
         return -1;
     }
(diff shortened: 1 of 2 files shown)


src/runtime_src/xdp/profile/device/profile_ip_access.cpp
```diff
@@ -0,0 +1,179 @@
...
+        return;
+    }
+    std::string debug_ip_layout_path(raw_debug_ip_layout_path);
+    std::ifstream debug_ip_layout_fs(debug_ip_layout_path.c_str(), std::ifstream::binary);
+    size_t max_sysfs_size = 65536;
+    char buffer[max_sysfs_size];
+    if (debug_ip_layout_fs) {
+        debug_ip_layout_fs.read(buffer, max_sysfs_size);
+        if (debug_ip_layout_fs.gcount() > 0) {
+            debug_ip_layout* layout = (debug_ip_layout*)(buffer);
```
stsoe: don't use c-style casts in c++

src/runtime_src/xdp/profile/device/profile_ip_access.cpp
```diff
@@ -0,0 +1,179 @@
...
+            if (ip_index >= layout->m_count) {
+                show_warning("ip_index out of bound");
+                return;
+            }
+            debug_ip_data ip_data = layout->m_debug_ip_data[ip_index];
+            ip_name = std::string((char*)(&ip_data.m_name));
```
stsoe: This looks like a Coverity defect?   I suggest
ip_name.assign(reinterpret_cast<const char*>(&ip_data.m_name),64);
this will ensure that max 64 characters are read from a possible non-terminated char array.

src/runtime_src/xdp/profile/device/profile_ip_access.cpp
```diff
@@ -0,0 +1,179 @@
...
+    char raw_debug_ip_layout_path[max_path_size];
+    int get_sysfs_ret = xclGetSysfsPath(device_handle, subdev.c_str(), entry.c_str(), raw_debug_ip_layout_path, max_path_size);
+    if (get_sysfs_ret < 0) {
+        show_warning("Get debug_ip_layout path failed");
+        return;
+    }
+    std::string debug_ip_layout_path(raw_debug_ip_layout_path);
+    std::ifstream debug_ip_layout_fs(debug_ip_layout_path.c_str(), std::ifstream::binary);
+    size_t max_sysfs_size = 65536;
+    char buffer[max_sysfs_size];
```
stsoe: It's always a good idea to initialize variables when declared. Consider adding ' = {0}'

src/runtime_src/xdp/profile/device/profile_ip_access.cpp
```diff
@@ -0,0 +1,179 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "profile_ip_access.h"
+
+namespace xdp {
+
+ProfileIP::ProfileIP(xclDeviceHandle handle, int index) {
+    // initialize member variables
```
stsoe: Use constructor initializer list.

src/runtime_src/xdp/profile/device/profile_ip_access.cpp
```diff
@@ -0,0 +1,179 @@
...
+     * the function that maps the IP registers to user memory space, this method should be simplified
+     * to one function call to the hal API and saves the result in a mapped_address and set mapped flag.
+     */
+    if (!exclusive) {
+        return;
+    }
+    std::string subdev = "icap";
+    std::string entry = "debug_ip_layout";
+    size_t max_path_size = 256;
+    char raw_debug_ip_layout_path[max_path_size];
```
stsoe: It's always a good idea to initialize variables when declared. Consider adding ' = {0}'  In this example the variable is assigned to a string after being populated by xclGetSysfsPath, can you count on the content being terminated?

src/runtime_src/xdp/profile/device/profile_ip_access.cpp
tianhaoz95: In the new push, this has been changed to reinterpret casting.

src/runtime_src/xdp/profile/device/profile_ip_access.cpp
tianhaoz95: Changed to suggested approach in the new push.

src/runtime_src/xdp/profile/device/profile_ip_access.cpp
tianhaoz95: Changed to suggested approach in the new push.

src/runtime_src/xdp/profile/device/profile_ip_access.cpp
tianhaoz95: Changed to suggested approach in the new push.

src/runtime_src/xdp/profile/device/profile_ip_access.cpp
tianhaoz95: xclGetSysfsPath does make sure within the range it is null terminated, but I added null initialization and null termination at the end to make sure.

stsoe:
ok to test

tianhaoz95:
I checked the build console output but didn't find any code related failure. Can someone check what is failing and rebuild if necessary?

stsoe:
retest this please

PR #976, opened 2019-02-15
Fix for Coverity_fortify issues

IshitaGhosh:
Add the fix for Coverity_Fortify issues initially reported in 2018.3 XRT.
Type of m_name member of debug_ip_data changed from uint8_t array to char array. Coverity fails to determine whether "m_name" is null-terminated when the type is "uint8_t" array.  So, changed the type to char array and removed "char*" type-casting at different place.


src/runtime_src/driver/hw_em/generic_pcie_hal2/debug.cxx
```diff
@@ -142,7 +142,7 @@ namespace xclhwemhal2 {
           if (count >= size) break;
           if (map->m_debug_ip_data[i].m_type == type) {
             if (baseAddress) baseAddress[count] = map->m_debug_ip_data[i].m_base_address;
-            if (portNames)   portNames[count]   = (char*)map->m_debug_ip_data[i].m_name;
+            if (portNames)   portNames[count]   = map->m_debug_ip_data[i].m_name;
```
stsoe: is m_name guaranteed to be null-terminated?

src/runtime_src/driver/xclng/tools/xbutil/xbutil_debug.cpp
```diff
@@ -60,7 +60,7 @@ uint32_t xcldev::device::getIPCountAddrNames(int type,
             if (baseAddress)
                 baseAddress->push_back(map->m_debug_ip_data[i].m_base_address);
             if(portNames)
-                portNames->push_back((char*)map->m_debug_ip_data[i].m_name);
+                portNames->push_back(map->m_debug_ip_data[i].m_name);
```
stsoe: is m_name guaranteed to be null-terminated?

src/runtime_src/xdp/profile/writer/util.cpp
```diff
@@ -31,10 +31,14 @@ namespace xdp {
...
     //ss << std::put_time(std::localtime(&time), "%Y-%m-%d %X");
     //return ss.str();
-    struct tm tstruct = *(std::localtime(&time));
-    char buf[80];
-    strftime(buf, sizeof(buf), "%Y-%m-%d %X", &tstruct);
-    return std::string(buf);
+
+    struct tm *p_tstruct = std::localtime(&time);
+    if(p_tstruct) {
+        char buf[80];
```
stsoe: I'd recommend intializing buf to '= {0}'

src/runtime_src/driver/hw_em/generic_pcie_hal2/debug.cxx
IshitaGhosh: When the binary image of debug_ip_data is created in "XclBinData::createDebugIPLayoutBinaryImage", the entire debug_ip_data is first initialized with 0.
Then while writing the content of "m_name", the input string length is confirmed to be less than "m_name" size and then the input string along with its terminating null character is copied into m_name like this 
    memcpy( debugIpDataHdr.m_name, sm_name.c_str(), sm_name.length() + 1);

Also, while reading the debug_ip_data, the entire structure is read altogether and so, the terminating null will also be read in.

stsoe: I guess my point is that this is a lot of assumptions, none of which can be deduced from looking at this code :-) I think Fortify will keep complaining, and I'd say rightfully so.  You should consider using the std::string::assign function to copy the char array content.

src/runtime_src/xdp/profile/writer/util.cpp
IshitaGhosh: Added it. Thanks for pointing it out!

src/runtime_src/driver/hw_em/generic_pcie_hal2/debug.cxx
IshitaGhosh: Okay, then. I will update it.
Thanks!

stsoe:
ok to test

PR #978, opened 2019-02-16
fix/wrong-appdebug-debug-ip-layout-location

tianhaoz95:
The location of debug_ip_layout used in AppDebug was wrong. This PR contains propagating the HAL API for getting SYSFS path up to XRT device and use that XRT device API to get debug_ip_layout in AppDebug.

### DIFF
--- a/src/runtime_src/xrt/device/device.h
+++ b/src/runtime_src/xrt/device/device.h
@@ -810,10 +810,10 @@ public:
     return m_hal->stopTrace(type);
   }
 
-  hal::operations_result<size_t>
-  getSysfsPath(std::string& subdev, std::string& entry, char* path, int size)
+  hal::operations_result<std::string>
+  getSysfsPath(std::string& subdev, std::string& entry)
   {
-    return m_hal->getSysfsPath(subdev, entry, path, size);
+    return m_hal->getSysfsPath(subdev, entry);
   }
 
   /**
--- a/src/runtime_src/xrt/device/device.h
+++ b/src/runtime_src/xrt/device/device.h
@@ -811,7 +811,7 @@ public:
   }
 
   hal::operations_result<std::string>
-  getSysfsPath(std::string& subdev, std::string& entry)
+  getSysfsPath(const std::string& subdev, const std::string& entry)
   {
     return m_hal->getSysfsPath(subdev, entry);
   }
--- a/src/runtime_src/xrt/device/device.h
+++ b/src/runtime_src/xrt/device/device.h
@@ -810,6 +810,12 @@ public:
     return m_hal->stopTrace(type);
   }
 
+  hal::operations_result<size_t>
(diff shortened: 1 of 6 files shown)


src/runtime_src/xrt/device/device.h
```diff
@@ -810,6 +810,12 @@ class device
     return m_hal->stopTrace(type);
   }
 
+  hal::operations_result<size_t>
```
stsoe: This is not an appropriate C++ interface.  Please use const appropriately and use std::array for wrapping fixed length char* arrays.  But I think this function should return a std::string with the path, not size_t.  So signature should be hal_operations<std::string>
getSysfsPath(const std::string& subdev, const std::stirng& entry);

src/runtime_src/xrt/device/hal.h
```diff
@@ -603,6 +603,12 @@ class device
     return operations_result<size_t>();
   }
 
+  virtual operations_result<size_t>
+  getSysfsPath(std::string subdev, std::string entry, char* path, int size)
```
stsoe: Fix per device.h comments

src/runtime_src/xrt/device/device.h
tianhaoz95: changed to use `hal::operations_result<std::string>` in the new push

src/runtime_src/xrt/device/hal.h
tianhaoz95: changed to use `hal::operations_result<std::string>` in the new push

src/runtime_src/xrt/device/hal2.h
```diff
@@ -574,6 +574,21 @@ class device : public xrt::hal::device
       return hal::operations_result<size_t>();
     return m_ops->mStopTrace(m_handle,type);
   }
+
+  virtual hal::operations_result<std::string>
+  getSysfsPath(const std::string subdev, const std::string entry)
```
stsoe: Please pass by reference

src/runtime_src/xrt/device/hal.h
```diff
@@ -603,6 +603,12 @@ class device
     return operations_result<size_t>();
   }
 
+  virtual operations_result<std::string>
+  getSysfsPath(const std::string subdev, const std::string entry)
```
stsoe: Please pass by reference, e.g. like you do in device.h

src/runtime_src/xrt/device/hal2.h
tianhaoz95: fixed in new push

src/runtime_src/xrt/device/hal.h
tianhaoz95: fixed in new push

stsoe:
ok to test

PR #979, opened 2019-02-17
kds dataflow support

stsoe:
Dataflow is supported in KDS only, no ERT support yet.


stsoe:
Pipeline hung.  Redo for more debug

PR #981, opened 2019-02-18
Reset scheduler on first reference on xclbin

stsoe:
Move reset from release of last reference to where first reference is acquired.
This change is needed for xbutil top to show execution status after application finishes.


stsoe:
Sounds right.  Can we blindly call xocl_exec_reset after hot reset?  Where should this call be added?

maxzhen:
> 
> 
> Sounds right. Can we blindly call xocl_exec_reset after hot reset? Where should this call be added?

It seems fine if we just blindly reset scheduler. No one is using it at this point anyway since all clients will be killed at the beginning of hot reset process.
I think we can reset scheduler in xocl_reset_notify(prepare = false), which indicate that the mgmt pf has performed the reset on the device and it has just recovered from reset successfully.
Does it make sense?

stsoe:
Yep.  Thanks.  Added the call as suggested.

PR #983, opened 2019-02-18
Disable CUDMA for 5.1 DSAs when regmap exceeds 127 words

stsoe:
For 5.1 DSAs CUDMA has a bug and can't copy more than 127 words to
kernel control registr (regmap).

Rather than relying on upstream (OCL) to conditionally disable CUDMA
by parsing xclbin XML meta-data, let the scheduler firmware revert to
manual CU configuration if and only if regmap size exceeds 127 words.

This change is needed for auto scheduler configuration where xclbin
XML meta data isn't easily parsed.

This change has to go first (before autocfg) as DSABINs must be
updated with new firmware.


stsoe:
retest this please

stsoe:
Looks like infra problem, sqlite something

stsoe:
retest this please

PR #984, opened 2019-02-19
Scheduler configuration as part of xclLoadXclbin

stsoe:


stsoe:
Sarab.  This change move scheduler configuration to xclLoadXclBin so presumably xmahw_hal.cpp can be simplified to no longer exec the configuration command.   This is just a heads up that XMA changes may be needed when I merge this PR.

Vamshi.  Minor changes to hw_emu shim to call the function that configures the scheduler.  This affects HW emu only, SW emu remains unchanged, or continues to be handled by OCL.

PR #985, opened 2019-02-19
whitelist CentOS version 7.6 in setup.csh and setup.sh

rradjabi:
Add support for CentOS 7.6 in setup scripts. This should be merged into 2018.3 and then I will cherrypick this into 2018.3.RC3 branch.


stsoe:
ok to test

PR #990, opened 2019-02-19
whitelist CentOS version 7.6 in setup.csh and setup.sh (#985)

rradjabi:
Cherry-pick ba3378b from branch:2018.3 into branch 2018.3_RC3. After merging this PR, we will tag as 2018.3_RC3.P1.


maxzhen:
ok to test

PR #991, opened 2019-02-19
[WIP] hal level plugin for profiling (PR mainly for discussion not submission)

tianhaoz95:
This PR shows the structure for hal level plugins.

The work is incomplete (most likely going to close this one and open a new PR) and will need input for the detailed implementation:
* what do we want in the config for turning on the hal level plugins?
* xdp library plugin needs to change to make use of hal level plugin because hal can exist together with OpenCL and xma.
* Suggetions on the overall implementation


tianhaoz95:
@stsoe the objective is to capture host activities for hal applications the similar way we do for OpenCL applications. I will call a meeting later to go over the code.

tianhaoz95:
There are many conflicts. Pending this work for now.

PR #992, opened 2019-02-20
Should unmap buffer in XRT test cases

mamin506:
Some test cases didn't call unmap. This is no good coding style for test cases that everyone can see.


maxzhen:
ok to test

PR #1001, opened 2019-02-21
Add FeatureROM UUID support to dsabin packaging.

emenchen:
Also removed dodsabin_xclbincat from pkgdsa.sh as it is no longer used.


emenchen:
Stephen Rozum has reviewed these changes.

sonals:
ok to test

PR #1003, opened 2019-02-22
Initial hard-coded implementation of Zynq function to retreive the pa…

jvillarre:
…th to a file in the sysfs directory (specifically debug_ip_layout).  This is necessary to test out profile and debug features and should be updated when we have a programmatic way to find this directory on Zynq platforms

### DIFF
--- a/src/runtime_src/driver/zynq/user/shim.cpp
+++ b/src/runtime_src/driver/zynq/user/shim.cpp
@@ -449,6 +449,22 @@ int ZYNQShim::xclExecWait(int timeoutMilliSec)
   return poll(&uifdVector[0], uifdVector.size(), timeoutMilliSec);
 }
 
+int ZYNQShim::xclGetSysfsPath(const char* subdev, const char* entry, char* sysfsPath, size_t size)
+{
+  // Until we have a programmatic way to determine what this directory
+  //  is on Zynq platforms, this is hard-coded so we can test out 
+  //  debug and profile features.
+  std::string path = "/sys/devices/platform/amba/a0000000.zyxclmm_drm/";
+  path += entry ;
+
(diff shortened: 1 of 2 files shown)


src/runtime_src/driver/zynq/user/shim.cpp
```diff
@@ -449,6 +449,22 @@ int ZYNQShim::xclExecWait(int timeoutMilliSec)
   return poll(&uifdVector[0], uifdVector.size(), timeoutMilliSec);
 }
 
+int ZYNQShim::xclGetSysfsPath(const char* subdev, const char* entry, char* sysfsPath, size_t size)
+{
+  // Until we have a programmatic way to determine what this directory
+  //  is on Zynq platforms, this is hard-coded so we can test out 
+  //  debug and profile features.
+  std::string path = "/sys/devices/platform/amba/a0000000.zyxclmm_drm/";
```
mamin506: The linux will name the device <address>.<device tree node name>. This will change once we remove the CU hard coding address.
For right now, this is good for me.

maxzhen:
ok to test

jvillarre:
The changes are Zynq specific and the failure appears to be on x86 architecture.  Can this be rebuilt and rechecked?

larry9523:
While this is OK for now, a better way might be using libudev to get the sysfs path. For example, we can use the driver name "zocl-drm" property, which should be more stable than the CU address, to match the sysfs path.
udev_enumerate_add_match_property(enumerate, "DRIVER", "zocl-drm");
path = udev_list_entry_get_name(dev_list_entry);

Though, this requires libudev to be added into our platform's image.ub.

maxzhen:
retest this please

PR #1006, opened 2019-02-24
KDMA support in XRT

maxzhen:
@sonals @stsoe Please take a look.

Application should provide an exec buf BO and call ert_fill_kdma_cmd() to initialize KDMA exec buf, then submit it as a normal exec buf to xocl.

When copy size is 64 bytes, I see bandwidth is 1.30MB/s (maximum overhead seen from exec buf and exec wait)
When copy size is 3G bytes, I see bandwidth is 11GB/s (overhead from exec buf and wait can be ignored, probably saturates DDR bandwidth?)

So, comparing with XDMA (DDR0 -> HOST -> DDR1), KDMA is about 2X faster.

Let me know what you think.

### DIFF
--- a/src/runtime_src/driver/include/ert.h
+++ b/src/runtime_src/driver/include/ert.h
@@ -109,6 +109,25 @@ struct ert_start_kernel_cmd {
   uint32_t data[1];          /* count-1 number of words */
 };
 
+#define COPYBO_UNIT   64     /* Limited by KDMA CU */
+struct ert_start_copybo_cmd {
+  uint32_t state:4;          /* [3-0], must be ERT_CMD_STATE_NEW */
+  uint32_t unused:6;         /* [9-4] */
+  uint32_t extra_cu_masks:2; /* [11-10], = 3 */
+  uint32_t count:11;         /* [22-12], = sizeof(ert_start_copybo_cmd)-1 */
+  uint32_t opcode:5;         /* [27-23], = ERT_START_COPYBO */
+  uint32_t type:4;           /* [31-27], = ERT_DEFAULT */
+  uint32_t cu_mask[4];       /* mandatory cu masks */
+  uint32_t reserved[4];      /* for scheduler use */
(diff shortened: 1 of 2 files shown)


maxzhen:
retest this please

src/runtime_src/driver/include/ert.h
```diff
@@ -109,6 +109,25 @@ struct ert_start_kernel_cmd {
   uint32_t data[1];          /* count-1 number of words */
 };
 
+#define KDMA_UNIT   64
```
stsoe: My preference would be to keep a much simpler user exposed ert_kdma_cmd, which just allows opcode, srcbo, dstbo, offsets and size.   The scheduler can repopulate this command in place with a new opcode and proper bo addresses and cumasks.

src/runtime_src/driver/include/ert.h
```diff
@@ -372,4 +394,26 @@ enum ert_cmd_type {
 #define ERT_INTC_IAR_ADDR                 (ERT_INTC_ADDR + 0x0C) /* acknowledge */
 #define ERT_INTC_MER_ADDR                 (ERT_INTC_ADDR + 0x1C) /* master enable */
 
+static inline void
```
stsoe: Do we have to expose ert_fill_kdma_cmd in ert.h?

maxzhen:
The idea is that user should manage the life cycle of exec buf BO used for kdma (also for p2p copy later) and keep the allocation and free outside of critical path. User does not need to care about the internal members of ert_start_kdma_cmd. They should always just call ert_fill_kdma_cmd() and provide what you mentioned in your comment. User would need to know the size of struct ert_start_kdma_cmd so that the allocated exec buf BO has enough space to fill in all necessary data inside driver after conversion. So, user is exposed to an API (ert_fill_kdma_cmd()) and a data structure that only size matters. Note that the API is used in both userland and driver, so ert.h sounds a good place to place it.

PR #1011, opened 2019-02-26
add API into the XRT device to retrieve hal device handle

tianhaoz95:
To refactor XDP, we want xma, OpenCL, and HAL (XfDNN and Deephi on the way) profiling to share as much code as much, so we want to contain the device related data and depend only on HAL. To do that, we will need the raw HAL device handle from the OpenCL device object. Therefore, this API is added to prepare the refactoring.


maxzhen:
ok to test

PR #1013, opened 2019-02-27
auto config scheduler when loadXclbin success

mamin506:
1. move common code from driver/xclng/xrt/util/ to driver/common/
2. support auto config scheduler when load XCLBIN success

After fix the conflict:
message.cpp add version.h in the include files. The second commit is for version.h on MPSoC build.


mamin506:
Because version.h is included in message.cpp. I need to add version.h for MPSoC platform, which today is not support.

sonals:
ok to test

mamin506:
retest this please

PR #1015, opened 2019-02-27
mailbox msg timer on

chienwei-lan:
1.should check size before vmalloc and check pointer after vmalloc.

2. turn on rx enqueued msg  and outstanding msg timer on


maxzhen:
ok to test

PR #1020, opened 2019-02-28
fixed xdma config bar lookup failure

houlz0507:


sonals:
#1021 includes this commit

sonals:
ok to test

PR #1021, opened 2019-02-28
Resolve xbutil validate issue

sonals:
This PR also contains @houlz0507 fix for XDMA BAR


maxzhen:
This PR should fix CR-1023154

PR #1023, opened 2019-03-01
KDMA support in XRT (2018.3)

maxzhen:
Porting KDMA support from master to 2018.3


maxzhen:
@stsoe please take a look. I have to create a new API for scheduler subdevice so that I can call convert_execbuf from outside of mb_scheduler. This is the only difference from the change in master.

maxzhen:
Close it for now until we really decide to merge

PR #1027, opened 2019-03-03
no init mem, if no mem topo

maxzhen:
This change should fix kernel panic when downloading xclbin and xclmgmt is not fully initialized.


NaAl-Xlnx:
retest this please

PR #1033, opened 2019-03-04
Do not require P2P buffer to be resident on device

stsoe:
The P2P exported buffer must be allocated (xclAlloc) but is not
required to be resident (xclSync) on device.

CR1023406


stsoe:
Max.  Mostly FYI.  This is what we talked about, the buffer is reserved, but that's all that's required.  I don't know if we should require the buffer to have the P2P flag?

maxzhen:
Soren, I think this change applies to all BOs, not just P2P ones. For non-P2P BOs (default BOs), we also support BO export operation and we will export s/g table pointing to the host mem anyway. It seems irrelevant whether the device mem has been sync'ed or not since that the dev mem is not exported at all.

PR #1041, opened 2019-03-06
remove device id check in libqdma

houlz0507:


houlz0507:
ok to test

houlz0507:
@karenx-xilinx could you review?

PR #1045, opened 2019-03-06
Support imported BO (P2P) copy via ert_start_copybo_cmd

maxzhen:
This is to support (P2P) copy of imported BO via ert_start_copybo_cmd, which is also used for KDMA copy. After this change, we should be able to start working on the support for KDMA and P2P BO copy through clEnqueueCopyBuffer() and remove xclCopyBO() API.

### DIFF
--- a/src/runtime_src/driver/include/ert.h
+++ b/src/runtime_src/driver/include/ert.h
@@ -423,18 +423,12 @@ ert_fill_copybo_cmd(struct ert_start_copybo_cmd *pkt, uint32_t src_bo,
 static inline uint64_t
 ert_copybo_src_offset(struct ert_start_copybo_cmd *pkt)
 {
-  uint64_t off = pkt->src_addr_hi;
-  off <<= 32;
-  off |= pkt->src_addr_lo;
-  return off;
+  return (uint64_t)pkt->src_addr_hi << 32 | pkt->src_addr_lo;
 }
 static inline uint64_t
 ert_copybo_dst_offset(struct ert_start_copybo_cmd *pkt)
 {
-  uint64_t off = pkt->dst_addr_hi;
-  off <<= 32;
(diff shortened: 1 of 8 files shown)


src/runtime_src/driver/include/ert.h
```diff
@@ -414,6 +418,29 @@ ert_fill_copybo_cmd(struct ert_start_copybo_cmd *pkt, uint32_t src_bo,
   pkt->dst_addr_hi = (dst_offset >> 32) & 0xFFFFFFFF;
   pkt->dst_bo_hdl = dst_bo;
   pkt->size = size / COPYBO_UNIT;
+  pkt->arg = 0;
+}
+static inline uint64_t
+ert_copybo_src_offset(struct ert_start_copybo_cmd *pkt)
+{
+  uint64_t off = pkt->src_addr_hi;
```
stsoe: return (uint64_t)pkt->src_addr_hi << 32 | pkt->src_addr_low;
???

src/runtime_src/driver/xclng/drm/xocl/subdev/mb_scheduler.c
```diff
@@ -1493,7 +1493,16 @@ exec_execute_write_cmd(struct exec_core *exec, struct xocl_cmd* xcmd)
 static int
 exec_execute_copybo_cmd(struct exec_core *exec, struct xocl_cmd* xcmd)
 {
-	return 1; // error for now
+	int ret;
+	struct ert_start_copybo_cmd *ecmd =
```
stsoe: Do you want to add ert_start_copybo_cmd to the union under xocl_cmd?  I think consolidating the various supported command formats under xocl_cmd is good

maxzhen:
@stsoe , I just pushed a change to address your comments. Please take another look.

PR #1048, opened 2019-03-06
fix for correctly picking sdaccel.ini

AShivangi:
This is a fix to correctly catch the file does not exist error. Without this, "xbutil flash scan" gives files not found error if the user does not have all permissions to the directory


maxzhen:
ok to test

PR #1052, opened 2019-03-06
Adding support in xbutil and xclbinutil to handle newly inserted Streaming Protocol Checkers

jvillarre:


sonals:
ok to test

sonals:
@jvillarre , please rebase it with upstream as there is a merge conflict.

PR #1053, opened 2019-03-07
Changes on embedded XRT builds to meet petalinux requirement

mamin506:
1. Move header and library file to /usr, this is the standard approach to install library in embedded system.
2. Yocto recipe team do not accept install libraries into /opt/xilinx/xrt. Breaking those rules might lead to yocto QA issue or PetaLinux regression test failed.
3. Do not install setup.sh and version.json for embedded system
4. Do not install appdebug python scripts for now. The profiling team has hard coding path /opt/xilinx/xrt/share/appdebug . Should file a CR to fix it.


mamin506:
ok to test

maxzhen:
ok to test

Issue #836, opened 2019-01-02
build issue with cmake 3.13.2


tianhaoz95-xilinx:
OS: Ubuntu 16.04
System Spec:
```bash
uname -a
Linux 4.4.0-138-generic #164-Ubuntu SMP Tue Oct 2 17:16:02 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```
cmake version:
```bash
cmake --version
cmake version 3.13.2
CMake suite maintained and supported by Kitware (kitware.com/cmake).
```
Error message:
```bash
/bin/sh: ./kernel-doc: No such file or directory
runtime_src/doc/CMakeFiles/xrt_docs.dir/build.make:79: recipe for target 'runtime_src/doc/core/ert.rst' failed
make[2]: *** [runtime_src/doc/core/ert.rst] Error 127
make[2]: *** Deleting file 'runtime_src/doc/core/ert.rst'
make[2]: *** Waiting for unfinished jobs....
[  9%] Generating core/mgmt-ioctl.rst
/bin/sh: ./kernel-doc: No such file or directory
runtime_src/doc/CMakeFiles/xrt_docs.dir/build.make:67: recipe for target 'runtime_src/doc/core/mgmt-ioctl.rst' failed
make[2]: *** [runtime_src/doc/core/mgmt-ioctl.rst] Error 127
make[2]: *** Deleting file 'runtime_src/doc/core/mgmt-ioctl.rst'
Scanning dependencies of target xclbinutil
CMakeFiles/Makefile2:537: recipe for target 'runtime_src/doc/CMakeFiles/xrt_docs.dir/all' failed
make[1]: *** [runtime_src/doc/CMakeFiles/xrt_docs.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

mamin506:
@tianhaoz95-xilinx Do you still see this issue in the latest code?
Could you check below items:
1. internet access on that build machine
2. "./kernel-doc downloading" message in the cmake log
3. build/Release/kernel-doc exist or not

tianhaoz95-xilinx:
I haven't been using that machine for a while. Will give it a try.

tianhaoz95:
@mamin506 Yes I still see this issue in the latest code.

error message:

```
/bin/sh: ./kernel-doc: No such file or directory
[...]

PR #1055, opened 2019-03-07
fix CMake 3.13 incompatible issue

tianhaoz95:
In CMake 3.13 the building of the documentation fails (refer to issue #836 ) because "./" no longer translate to root binary directory, and in general "./" is not preferable. I changed it to use the relative binary build directory which doesn't change the actual location of the kernel-doc executable.


maxzhen:
ok to test

maxzhen:
retest this please

sonals:
@tianhaoz95 , please rebase with upstream as there is a conflict

tianhaoz95:
@sonals @mamin506 Both of the changes seem to aim at fixing the same thing. I will see if that helps to resolve the issue. If so, this can be closed.

tianhaoz95-xilinx:
I confirm that the incompatibility is fixed by the conflicting commit, so this one can be closed.

PR #1060, opened 2019-03-08
Changes to xbutil log

AShivangi:
**CR-1019848**: Replace references to Board and DSA with Card and Shell in xbutil log output

**CR-1020709**: Alveo card xbutil flashing information displayed during shell deployment package install is incorrect

### DIFF
--- a/src/runtime_src/driver/xclng/tools/xbmgmt/xbmgmt.cpp
+++ b/src/runtime_src/driver/xclng/tools/xbmgmt/xbmgmt.cpp
@@ -199,12 +199,12 @@ unsigned selectDSA(unsigned idx, std::string& dsa, uint64_t ts)
     {
         if (installedDSA.empty())
         {
-            std::cout << "no DSA installed" << std::endl;
+            std::cout << "no shell installed" << std::endl;
             return candidateDSAIndex;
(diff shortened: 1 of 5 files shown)


maxzhen:
ok to test

AShivangi:
I believe it is part of xbutil dump:

```
{                                                                                            
    "runtime": {                                                                             
        "build": {                                                                           
            "version": "2.2.0",                                                              
            "hash": "5a5bd8a9079434db82fdb649fa758bebc54f852c",                              
            "date": "2019-03-01 11:38:16",                                                   
            "branch": "flash-scan-fix"                                                       
        }                                                                                    
    },                                                                                       
    "board": {                                                                               
        "info": {                                                                            
            "dsa_name": "xilinx_u250_xdma_201830_2",
...
```

"board" comes from the tree. 
I changed the sensor_tree.put() as well as get() in xbutil.h, but not in xbmgmt.h. There it still uses board. I didn't find any other dependency on the name board, but I might be missing something.

maxzhen:
Yes, you can dump it, but "board" is just a variable name. It does not refer to any HW entity on the system. So, I don't think we need to change them.

PR #1061, opened 2019-03-08
xbutil validate warning

AShivangi:
**CR-1022897**: Update xbutil output when 'warning' encountered
In case of a warning during xbutil validate:
```
...
INFO: Card[0] validated with warnings.             

INFO: All cards validated successfully but with warnings.
```


maxzhen:
ok to test

maxzhen:
retest this please

PR #1063, opened 2019-03-09
Virtual CU context

maxzhen:
This change introduce virtual CU (index number is -1). Application can reserve virtual CU if it does not need to use any real CUs, but would like to lock down xclbin for performing other activities, such as KDMA.


maxzhen:
> 
> 
> Looks good to me. You have tested all board tests and multiprocess tests right?

Tried multiple processes test, full xbutil validate, ert copybo cmd. will try all board test and see if there is anything missed.

maxzhen:
board test ran fine. Merging now.

PR #1064, opened 2019-03-09
fix zocl sysfs not releasing lock

larry9523:
When reading beyond the size of some sysfs file, need to release lock before bailing out


maxzhen:
ok to test

PR #1069, opened 2019-03-11
Various build fixes

u1f35c:
This is a collection of 3 fixes I needed to get XRT building on a Debian Buster system - in particular 2 cleanups for warnings in GCC 8.3.0, plus a fixup to correctly use the local kernel-doc copy. I understand Debian is probably not a supported distro, but these fixes seem relevant to other systems.

### DIFF
--- a/src/runtime_src/xocl/core/memory.h
+++ b/src/runtime_src/xocl/core/memory.h
@@ -488,7 +488,7 @@ public:
       // allocate sufficiently aligned memory and reassign m_host_ptr
       if (posix_memalign(&m_host_ptr,alignment,sz))
         throw error(CL_MEM_OBJECT_ALLOCATION_FAILURE);
-    if (flags & CL_MEM_COPY_HOST_PTR)
+    if (flags & CL_MEM_COPY_HOST_PTR && host_ptr)
       std::memcpy(m_host_ptr,host_ptr,sz);
 
     m_aligned = (reinterpret_cast<uintptr_t>(m_host_ptr) % alignment)==0;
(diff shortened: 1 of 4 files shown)


maxzhen:
ok to test

src/runtime_src/xocl/core/memory.h
```diff
@@ -488,7 +488,7 @@ class buffer : public memory
       // allocate sufficiently aligned memory and reassign m_host_ptr
       if (posix_memalign(&m_host_ptr,alignment,sz))
         throw error(CL_MEM_OBJECT_ALLOCATION_FAILURE);
-    if (flags & CL_MEM_COPY_HOST_PTR)
+    if (flags & CL_MEM_COPY_HOST_PTR && host_ptr)
```
maxzhen: I'm not sure about this fix. Why check host_ptr == NULL here? It does not make sense for user to set CL_MEM_COPY_HOST_PTR and pass in NULL host_ptr, right?

u1f35c: It doesn't, and I think the callers do the right thing, but GCC 8.3 isn't smart enough to realise this and so bombs out with "-Werror" due to not knowing host_ptr is definitely null in this case. Adding the explicit check seemed to be the simplest way to fix the build without losing the check generally.

maxzhen: Before the fix, if host_ptr is null, the program will crash, which is good. After the fix, we'll just silently skip copying? That does not sound right?

PR #1072, opened 2019-03-11
NIFD driver

tianhaoz95:
> Opened this one because last one had a few conflicts and I left that as a backup

Fixed the following:
* removed installation for vcu1525 platforms
* added installation for xbb qdma platforms
* fixed format
* replaced free with xocl_drvinst_free
* replaced printk with info and error loggers
* in probe, changed to early return when NIFD not found in feature ROM


src/runtime_src/driver/xclng/drm/xocl/subdev/nifd.c
```diff
@@ -0,0 +1,680 @@
...
+    void *__iomem nifd_base;
+    void *__iomem icap_base;
+    unsigned int instance;
+    struct cdev sys_cdev;
+    struct device *sys_device;
+};
+
+static dev_t nifd_dev;
+
+struct xocl_nifd *nifd_global;
```
maxzhen: How do we support multiple boards with a system-wide global pointer?

src/runtime_src/driver/xclng/drm/xocl/subdev/nifd.c
```diff
@@ -0,0 +1,680 @@
...
+    if (copy_from_user(kernel_memory, arg, total_data_size))
+    {
+        vfree(kernel_memory);
+        return -EFAULT;
+    }
+
+    result = add_breakpoints_core(kernel_memory);
+    vfree(kernel_memory);
+
+    if (result)
```
maxzhen: can we just return result?

src/runtime_src/driver/xclng/drm/xocl/subdev/nifd.c
```diff
@@ -0,0 +1,680 @@
...
+        break;
+    default:
+        status = -ENOIOCTLCMD;
+        break;
+    }
+
+    return status;
+}
+
+static int char_open(struct inode *inode, struct file *file)
```
maxzhen: Do you want to follow what is done in xvc's open/close routines to avoid race that Lizhi has fixed?

tianhaoz95-xilinx: I changed it in one of the following commits. @houlz0507 please take a look and see if that is the correct way to do it. Thanks!

src/runtime_src/driver/xclng/drm/xocl/subdev/nifd.c
tianhaoz95-xilinx: Yes. I changed it in the following commit.

tianhaoz95-xilinx:
I ran the checkpatch.pl and fixed all the warnings except:

```
ERROR: Does not appear to be a unified-diff format patch
```

Which is caused by this file being newly added.

src/runtime_src/driver/xclng/drm/xocl/subdev/nifd.c
houlz0507: so does nifd_valid blow.

src/runtime_src/driver/xclng/drm/xocl/subdev/xvc.c
```diff
@@ -341,8 +341,7 @@ static int xvc_probe(struct platform_device *pdev)
 	struct resource *res;
 	struct xocl_dev_core *core;
 	int err;
-
-	xvc = xocl_drvinst_alloc(&pdev->dev, sizeof(*xvc));
+	xvc = devm_kzalloc(&pdev->dev, sizeof(*xvc), GFP_KERNEL);
```
houlz0507: Is this miss merge?

tianhaoz95: yes, this is a wrong merge. I have reverted `xvc.c` to the version on Xilinx master

src/runtime_src/driver/xclng/drm/xocl/subdev/nifd.c
tianhaoz95: @maxzhen in the new push, I changed it to pass through ioctl instead of using a global variable.

tianhaoz95: @houlz0507 in the new push, I changed it to do early return in the probe and got rid of the nifd_valid global flag.

src/runtime_src/driver/xclng/drm/xocl/subdev/nifd.c
```diff
@@ -0,0 +1,717 @@
...
+        xocl_err(&pdev->dev, "core is NULL in NIFD remove");
+    }
+
+    nifd = platform_get_drvdata(pdev);
+    if (!nifd)
+    {
+        xocl_err(&pdev->dev, "driver data is NULL");
+        return -EINVAL;
+    }
+    device_destroy(xrt_class, nifd->sys_cdev->dev);
```
houlz0507: This might be crashing the system. Please Handle the probe early return case. One possible way is checking the resource availability before trying to destroy or free.

tianhaoz95: @houlz0507 if the char device was not created, shouldn't it return at line 662? Is there an  API that I can use to check if create_device was called from pdev?

houlz0507: No, it will not return on 662.  I think you may use nifd->sys_device != NULL to check if create_device() is success or not.

tianhaoz95: @houlz0507 In the most recent commit, I changed it to only destroy the device if there is one. I will soon try that out on a few different systems. At the same time, please take a look if the change make sense.

src/runtime_src/driver/xclng/drm/xocl/subdev/nifd.c
houlz0507: It is clean. Thanks

src/runtime_src/driver/xclng/drm/xocl/subdev/nifd.c
houlz0507: looks good to me

tianhaoz95:
conflict resolved. retest this please.

houlz0507:
retest this please

PR #1075, opened 2019-03-13
Revert "Virtual CU context (#1063)"

maxzhen:
This reverts commit cea5d3e1c4bd484331c0ac4faab8a19ff6bc641d, which breaks emulation.


maxzhen:
Close this PR since the fix is in PR1067 now.

PR #1078, opened 2019-03-14
Locking

maxzhen:
This PR is for discussion. @stsoe @sonals, please take a look at the change.
Things tested:
xbutil validate, board test, multi-process with same xclbin, multi-process with different xclbin, single process with multi-thread share the same xocl fd. Any other test?

### DIFF
--- a/src/runtime_src/driver/xclng/drm/xocl/subdev/mb_scheduler.c
+++ b/src/runtime_src/driver/xclng/drm/xocl/subdev/mb_scheduler.c
@@ -1183,7 +1183,6 @@ static int
 exec_cfg_cmd(struct exec_core *exec, struct xocl_cmd *xcmd)
 {
 	struct xocl_dev *xdev = exec_get_xdev(exec);
-	struct client_ctx *client = xcmd->client;
 	uint32_t *cdma = xocl_cdma_addr(xdev);
 	unsigned int dsa = exec->ert_cfg_priv;
 	struct ert_configure_cmd *cfg = xcmd->ert_cfg;
@@ -1237,7 +1236,6 @@ exec_cfg_cmd(struct exec_core *exec, struct xocl_cmd *xcmd)
 	if (cdma) {
 		uint32_t *addr = 0;
 
-		mutex_lock(&client->lock); /* for modification to client cu_bitmap */
 		for (addr = cdma; addr < cdma+4; ++addr) { /* 4 is from xclfeatures.h */
 			if (*addr) {
 				struct xocl_cu *xcu = exec->cus[cuidx];
@@ -1259,7 +1257,6 @@ exec_cfg_cmd(struct exec_core *exec, struct xocl_cmd *xcmd)
 				++cuidx;
 			}
 		}
-		mutex_unlock(&client->lock);
 	}
 
 	if ((ert_full || ert_poll) && !exec->ert)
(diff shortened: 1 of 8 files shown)


src/runtime_src/driver/xclng/drm/xocl/subdev/mb_scheduler.c
```diff
@@ -2459,46 +2445,31 @@ static void destroy_client(struct platform_device *pdev, void **priv)
...
 {
-	struct client_ctx	*client = (struct client_ctx *)priv;
-	struct exec_core	*exec;
-	int			counter;
-	uint			ret = 0;
-
-	exec = platform_get_drvdata(pdev);
+	struct exec_core *exec = platform_get_drvdata(pdev);
 
 	poll_wait(filp, &exec->poll_wait_queue, wait);
```
stsoe: What happens on time out?  The client will be woken up with POLLIN, but that's a change in behavior.

src/runtime_src/driver/xclng/drm/xocl/subdev/mb_scheduler.c
```diff
@@ -1526,19 +1523,7 @@ exec_execute_copybo_cmd(struct exec_core *exec, struct xocl_cmd *xcmd)
...
 	SCHED_DEBUGF("-> %s(%d)\n", __func__, exec->uid);
-
-	/* now for each client update the trigger counter in the context */
-	mutex_lock(&xdev->ctx_list_lock);
-	list_for_each(ptr, &xdev->ctx_list) {
-		entry = list_entry(ptr, struct client_ctx, link);
-		atomic_inc(&entry->trigger);
-	}
-	mutex_unlock(&xdev->ctx_list_lock);
 	/* wake up all the clients */
```
stsoe: Should we really wake up all clients as are doing today and have always done in exec_notify_host?   Part of the reason for having the client trigger was maybe to know which specific client should be awoken, though that part of the logic was never used.

src/runtime_src/driver/xclng/drm/xocl/subdev/mb_scheduler.c
maxzhen: Hmm.... it does not work, then. I'll have to think about it.

src/runtime_src/driver/xclng/drm/xocl/subdev/mb_scheduler.c
maxzhen: why do we wake up all clients today? did we run into any issue if we just wake up the one who sent the cmd?

stsoe: No.  We just never implemented the necessary hooks to just wake up the right client.

sonals: xocl_cmd has access to client so we can improve this by waking up only the client which submitted this request. Need to enhance exec_notify_host() to take a second argument, client pointer--
exec_notify_host(exec, xcmd->client);

However, let's limit the current cleanup to bug fixes only. The wakeup improvement can be done in a subsequent patch?

maxzhen: Yes, I think the wake up improvement should be a separate PR and needs more thoughts.

Maybe we should just get fd from exec BO, and exec wait on that fd, instead of waiting on a global fd to xocl.

maxzhen:
@stsoe, just pushed a commit to fix ^C. I have to move the block of code waiting for outstanding cmds outside of dev_lock so that scheduler can finish outstanding cmds successfully. Please take another look.

sonals:
While we are at it, we need to fix this scenario
Thread0 -> open device -> submit execs -> CTRL-C -> close (destroy client in driver)
Thread1 -> execWait

Will Thread1 be waiting for stale client object that has been freed?

stsoe:
No, client1 can continue while client0 is waiting for its commands to complete in destroy_client, or has this changed?

sonals:
> No, client1 can continue while client0 is waiting for its commands to complete in destroy_client, or has this changed?

Both threads share the client object in the driver because they are both sharing the file descriptor.

maxzhen:
> 
> 
> > No, client1 can continue while client0 is waiting for its commands to complete in destroy_client, or has this changed?
> 
> Both threads share the client object in the driver because they are both sharing the file descriptor.

Oh, I haven't tested this case yet. I briefly mentioned this in my email that we may need a ref counter for client context data structure. But since close of the client and exec wait/buf are all going through DRM framework, I was hoping that DRM will protect us from the race. Otherwise, we'll have to come up with something...

maxzhen:
force pushed a commit to resolve conflict and added comment in xclhal2.h to require separate device handle for different threads since they can't share the same client context.

PR #1079, opened 2019-03-14
Use ERT for CU polling in dataflow mode

stsoe:
In dataflow mode KDS configures, starts, and checks CU status
directly.

If KDS is configured with ert=false, then CUs are polled for
completion in a busy loop.

If KDS is configured with ert=true (and KDS is in dataflow mode), then
ERT is used to poll CUs for completion and notify KDS when one
completes.  If KDS is not busy attempting to start commands, then it
will be waiting for ERT notification.  The ERT command queue is
divided into CU slots, where each slot corresponds to a CU.  KDS
requests ERT polling of a CU by writing AP_START or AP_CONTINUE into
the CQ slot corresponding to the CU.

Until xclbin meta data advertises dataflow kernels, sdaccel.ini can be
used to trigger dataflow mode by setting Runtime.dataflow=true.  Note
that together with ert=true (the default) the requires updated ERT
firmware part of the PR that supports polling of CUS.

### DIFF
--- a/src/runtime_src/driver/common/config_reader.cpp
+++ b/src/runtime_src/driver/common/config_reader.cpp
@@ -31,9 +31,15 @@
 namespace {
 
 static const char*
-valueOrEmpty(const char* cstr)
+value_or_empty(const char* cstr)
 {
   return cstr ? cstr : "";
+}
+
+static bool
+is_true(const std::string& str)
+{
+  return str=="true";
 }
 
 static std::string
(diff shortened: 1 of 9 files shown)


stsoe:
Max. to your comment.  Until we have ERT redesigned to support dataflow, KDS is responsible for starting CUs.  The ERT polling is simply to allow KDS to sleep while waiting for CUs to complete or accept another start, otherwise the flow is the same KDS with no ERT polling.   ERT polling will degrade performance ever so slightly since there is one additional AXI write to the CQ in poll mode and interrupt is used to wake up KDS again.  This theoretically slight degradation has not been measured, but since KDS can be configured with or without ERT polling, once we get some interesting dataflow examples we can check.

PR #1080, opened 2019-03-15
Tell CMake to build .debs for Debian as well as Ubuntu

u1f35c:
RPMS are built for RHEL + CentOS, but .debs are only enabled for Ubuntu. Correct this to build them for Debian proper as well.


stsoe:
ok to test

PR #1082, opened 2019-03-15
Support CU selection with clCreateKernel

stsoe:
cl_kernel clCreateKernel(program, kernel_name, err) now supports an
extension to the kernel_name such that it can include the compute unit
that should be used.

The format of the kernel_name argument is "kernel:{cu1,cu2,...}" where
"kernel" is the name of the kernel as normal and "cu1" through "cu_n"
are the names of the compute units to use.  It is strictly enforced
that the CU list is enclosed in a comma separated list without any
spaces between the CU names, commas, and curly braces.  It is also
strictly enforced that the list must follow a ":" separating a legal
kernel name.  Invalid CU names are currently ignored without warning,
but error is thrown if no CUs matched.

### DIFF
--- a/src/runtime_src/xocl/core/kernel.cpp
+++ b/src/runtime_src/xocl/core/kernel.cpp
@@ -26,10 +26,6 @@
 #include <algorithm>
 #include <regex>
 
-#include <boost/algorithm/string.hpp>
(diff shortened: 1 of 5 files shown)


stsoe:
Lizhi.  This an FYI, I don't expect you to review code unless you like to.

PR #1089, opened 2019-03-18
clean uuid if verify xclbin fail

chienwei-lan:
Should not cache uuid if verify xclbin fail(download bitstream fails),  
CR: http://jira.xilinx.com/browse/CR-1022451

and checkpatch.pl warning fixes


maxzhen:
ok to test

PR #1092, opened 2019-03-19
Added JSON output support to bit slice the properties value of IP_KERNELs

rozumx:
When a xclbin archive's IP_LAYOUT JSON section is dumped.  For IP_KERNELs, the m_int_enable, m_interrupt_id, and m_ip_control values will be inserted into the schema instead of properties.


maxzhen:
ok to test

PR #1100, opened 2019-03-20
[For 2018.3 branch] auto config scheduler when loadXclbin success #1013

mamin506:
The 2018.3 branch has auto config scheduler feature for xclng. This is to support MPSoC DSA.


maxzhen:
ok to test

PR #1101, opened 2019-03-20
board.sh should handle more test cases

maxzhen:
Add support for test command line in more test cases to board.sh


maxzhen:
ok to test

PR #1102, opened 2019-03-20
 Add exec_write command for non-consecutive CU layout

stsoe:
Expose C++ API level command interface that allows host code to construct, execute, and wait for commands to complete.

Introduce new command type (ERT_CU) to support multiple opcodes for CU commands, one new opcode being ERT_EXEC_WRITE.


stsoe:
Max. Minor change to mb_scheduler.cpp to facilitate new CU style command.  This new command will be  used by XDNN to schedule CUs that have non-consecutive regmap layout.  It supersedes the old ERT_WRITE command which is no longer needed.

PR #1106, opened 2019-03-21
Zynq profiling

jvillarre:
Adding initial implementation of HAL level profiling functions for Zynq platforms in the shim.


maxzhen:
ok to test

src/runtime_src/driver/zynq/user/shim.cpp
```diff
@@ -151,7 +155,7 @@ size_t ZYNQShim::xclWrite(xclAddressSpace space, uint64_t offset, const void *ho
     return -1;
   }
 
-  if(XCL_ADDR_KERNEL_CTRL == space) {
+  if(XCL_ADDR_KERNEL_CTRL == space || XCL_ADDR_SPACE_DEVICE_PERFMON == space) {
```
mamin506: Hi @jvillarre. To support floating CU address, we changed the implementation of xclRead() and xclWrite(). This should be refine. I can work together with you to figure out how to do that.

mamin506: Just talk to Jason, let this pull request merge first. Then I can update my pull request.

PR #1108, opened 2019-03-21
xclbincat and xclbinsplit are now deprecated

rozumx:
Created deprecation message for xclbincat and xclbinsplit.  
Fixed issue where the SectionHeader variable m_eType wasn't being initialized in the class constructor.


maxzhen:
ok to test

PR #1112, opened 2019-03-21
Acquire context on KDMA units before using them

stsoe:
OCL needs to open a virtual context on the device to get access to
the implicit CU resources.

CR1025263


maxzhen:
retest this please

PR #1114, opened 2019-03-22
[MPSoC] Floating cu address

mamin506:
1. Get rid of 'reg' in device tree, instead, use CU address from configure command.
2. Map CU to user space right after configure command. CU address and virtual address is recorded by a map.
3. Re-implemented xclRead() and xclWrite. Get rid of hard coding address

### DIFF
--- a/src/runtime_src/driver/common/scheduler.cpp
+++ b/src/runtime_src/driver/common/scheduler.cpp
@@ -212,7 +212,7 @@ get_dbg_ips_pair(const axlf* top)
  * Check sdaccel.ini for default overrides.
  */
 int
-init(xclDeviceHandle handle, const axlf* top, int encode)
+init(xclDeviceHandle handle, const axlf* top, bool encode)
 {
   uuid_t uuid;
   auto execbo = create_exec_bo(handle,0x1000);
--- a/src/runtime_src/driver/common/scheduler.cpp
+++ b/src/runtime_src/driver/common/scheduler.cpp
@@ -22,7 +22,6 @@
 #include <sys/mman.h>
 #include <memory>
 #include <string>
-#include <vector>
 #include <cstring>
 #include <uuid/uuid.h>
 
@@ -85,7 +84,7 @@ get_axlf_section(const axlf* top, axlf_section_kind kind)
 }
 
 static std::vector<uint64_t>
-get_cus(const axlf* top)
+get_cus(const axlf* top, bool encoding)
 {
   std::vector<uint64_t> cus;
   auto ip_layout = get_axlf_section<const ::ip_layout>(top,axlf_section_kind::IP_LAYOUT);
@@ -95,14 +94,43 @@ get_cus(const axlf* top)
   for (int32_t count=0; count <ip_layout->m_count; ++count) {
     const auto& ip_data = ip_layout->m_ip_data[count];
     if (ip_data.m_type == IP_TYPE::IP_KERNEL) {
-      // encode handshaking control in lower unused address bits
       uint64_t addr = ip_data.m_base_address;
-      addr |= ((ip_data.properties & IP_CONTROL_MASK) >> IP_CONTROL_SHIFT);
+      if (encoding) {
+          // encode handshaking control in lower unused address bits
+          addr |= ((ip_data.properties & IP_CONTROL_MASK) >> IP_CONTROL_SHIFT);
+      }
       cus.push_back(addr);
     }
   }
   std::sort(cus.begin(),cus.end());
   return cus;
+}
+
+static std::vector<std::pair<uint64_t, size_t> >
+get_debug_ips(const axlf* top)
+{
+  std::vector<std::pair<uint64_t, size_t> > ips;
+  auto debug_ip_layout = get_axlf_section<const ::debug_ip_layout>(top,
+                         axlf_section_kind::DEBUG_IP_LAYOUT);
+  if (!debug_ip_layout)
+    return ips;
+
+  for (int32_t count=0; count < debug_ip_layout->m_count; ++count) {
+    const auto& debug_ip_data = debug_ip_layout->m_debug_ip_data[count];
+    uint64_t addr = debug_ip_data.m_base_address;
+    // There is no size for each debug IP in the xclbin. Use hardcoding size now.
+    // The default size is 64KB.
+    size_t size = 0x10000;
+    if (debug_ip_data.m_type == AXI_MONITOR_FIFO_LITE
+        || debug_ip_data.m_type == AXI_MONITOR_FIFO_FULL)
+       // The size of these two type of IPs is 8KB
+       size = 0x2000;
+
+    ips.push_back(std::make_pair(addr, size));
+  }
+
+  std::sort(ips.begin(), ips.end());
+  return ips;
 }
 
 static uint64_t
@@ -153,10 +181,28 @@ get_dataflow(const axlf* top)
   return false;
 }
 
-
 } // unnamed
 
 namespace xrt_core { namespace scheduler {
+
+std::vector<std::pair<uint64_t, size_t> >
+get_cus_pair(const axlf* top)
+{
+  std::vector<uint64_t> cus;
+  std::vector<std::pair<uint64_t, size_t> > ret;
+  cus = get_cus(top, false);
+
+  for (auto it = cus.begin(); it != cus.end(); ++it)
+    ret.push_back(std::make_pair(*it, 0x10000));
+
+  return ret;
+}
+
+std::vector<std::pair<uint64_t, size_t> >
+get_dbg_ips_pair(const axlf* top)
+{
+  return get_debug_ips(top);
+}
 
 /**
  * init() - Initialize scheduler
@@ -166,7 +212,7 @@ namespace xrt_core { namespace scheduler {
  * Check sdaccel.ini for default overrides.
  */
 int
-init(xclDeviceHandle handle, const axlf* top)
+init(xclDeviceHandle handle, const axlf* top, int encode)
 {
   uuid_t uuid;
   auto execbo = create_exec_bo(handle,0x1000);
@@ -174,7 +220,7 @@ init(xclDeviceHandle handle, const axlf* top)
   ecmd->state = ERT_CMD_STATE_NEW;
   ecmd->opcode = ERT_CONFIGURE;
 
-  auto cus = get_cus(top);
+  auto cus = get_cus(top, encode);
 
   ecmd->slot_size = config::get_ert_slotsize();
   ecmd->num_cus = cus.size();
--- a/src/runtime_src/driver/common/scheduler.cpp
+++ b/src/runtime_src/driver/common/scheduler.cpp
@@ -95,10 +95,9 @@ get_cus(const axlf* top, bool encoding)
     const auto& ip_data = ip_layout->m_ip_data[count];
     if (ip_data.m_type == IP_TYPE::IP_KERNEL) {
       uint64_t addr = ip_data.m_base_address;
-      if (encoding) {
+      if (encoding)
           // encode handshaking control in lower unused address bits
           addr |= ((ip_data.properties & IP_CONTROL_MASK) >> IP_CONTROL_SHIFT);
-      }
       cus.push_back(addr);
     }
   }
--- a/src/runtime_src/driver/common/scheduler.cpp
+++ b/src/runtime_src/driver/common/scheduler.cpp
@@ -85,7 +85,7 @@ namespace xrt_core { namespace scheduler {
  * Check sdaccel.ini for default overrides.
  */
 int
-init(xclDeviceHandle handle, const axlf* top, bool encode)
+init(xclDeviceHandle handle, const axlf* top)
 {
   uuid_t uuid;
   auto execbo = create_exec_bo(handle,0x1000);
@@ -93,7 +93,7 @@ init(xclDeviceHandle handle, const axlf* top, bool encode)
   ecmd->state = ERT_CMD_STATE_NEW;
   ecmd->opcode = ERT_CONFIGURE;
 
-  auto cus = xclbin::get_cus(top, encode);
+  auto cus = xclbin::get_cus(top, true);
 
   ecmd->slot_size = config::get_ert_slotsize();
   ecmd->num_cus = cus.size();
--- a/src/runtime_src/driver/common/scheduler.cpp
+++ b/src/runtime_src/driver/common/scheduler.cpp
@@ -16,6 +16,7 @@
 
 #include "scheduler.h"
 #include "config_reader.h"
+#include "xclbin_parser.h"
 #include "message.h"
 #include "driver/include/ert.h"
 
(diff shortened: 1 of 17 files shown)


maxzhen:
ok to test

mamin506:
This change has conflict with #1106. Please let that PR merge first.

src/runtime_src/driver/zynq/drm/zocl/sched_exec.c
```diff
@@ -546,8 +549,9 @@ configure(struct sched_cmd *cmd)
 
 	for (i  = 0; i < exec->num_cus; i++) {
 		exec->cu_addr_phy[i] = cfg->data[i];
-		SCHED_DEBUG("++ configure cu(%d) at 0x%x\n", i,
-		    exec->cu_addr_phy[i]);
+		exec->cu_addr_virt[i] = ioremap(exec->cu_addr_phy[i], 0x80000);
```
larry9523: While this should be 0x10000, it seems better we create a micro than use instant number (if the CU size is not configurable any more)
Another thought is we'd better check the return value of ioremap to avoid bad CU physical address which can be anything by customer ert configuration command

src/runtime_src/driver/zynq/user/shim.cpp
```diff
@@ -143,144 +126,160 @@ ZYNQShim::~ZYNQShim()
 }
 #endif
 
+void ZYNQShim::mapKernelControl(std::vector<uint64_t> cus_offset) {
```
sonals: const std::vector<uint64_t>& cus_offset

src/runtime_src/driver/zynq/user/shim.cpp
```diff
@@ -143,144 +126,160 @@ ZYNQShim::~ZYNQShim()
...
+    return;
+  }
+
+  auto cu = cus_offset.begin();
 
-//TODO: UKP: return definition is not defined.
-size_t ZYNQShim::xclWrite(xclAddressSpace space, uint64_t offset, const void *hostBuf, size_t size)
-{
-  if(!hostBuf) {
+  while (cu != cus_offset.end()) {
```
sonals: Capture the end outside of while loop or else end() is reevaluated in every iteration.
cuEnd = cus_offset.end()
while (cu != cuEnd) {

src/runtime_src/driver/zynq/user/shim.cpp
```diff
@@ -143,144 +126,160 @@ ZYNQShim::~ZYNQShim()
...
+    auto it = mKernelControl.find(*cu);
+    if (it == mKernelControl.end()) {
+      ptr =
+          mmap(0, 0x10000, PROT_READ | PROT_WRITE, MAP_SHARED, mKernelFD, *cu);
+      mKernelControl[*cu] = (uint32_t *)ptr;
```
sonals: You can reuse the iterator from the lookup which makes the insertion more efficient:
mKernelControl.insert(it, std::pair<char,int>(*cu, (uint32_t *)ptr));

mamin506: Got it. This has been done after resolved conflict.

src/runtime_src/driver/zynq/user/shim.cpp
mamin506: Got it.

src/runtime_src/driver/zynq/drm/zocl/sched_exec.c
```diff
@@ -545,9 +548,18 @@ configure(struct sched_cmd *cmd)
 	}
 
 	for (i  = 0; i < exec->num_cus; i++) {
+		if (is_valid_apts(zdev, cfg->data[i]) == -EINVAL) {
```
mamin506: The aperture list contain CUs and Debug IPs for now. This is not the best solution. Let's refine this once we have a general solution to support variety apertures.

src/runtime_src/driver/zynq/drm/zocl/sched_exec.c
```diff
@@ -546,9 +529,30 @@ configure(struct sched_cmd *cmd)
...
+		}
+		/* For MPSoC as PCIe device, the CU address for PS = base
+		 * address + PCIe offset.
+		 *
+		 * For Pure MPSoC device, the base address is always 0
+		 */
+		exec->cu_addr_phy[i] = zdev->res_start + cfg->data[i];
+		exec->cu_addr_virt[i] = ioremap(exec->cu_addr_phy[i], CU_SIZE);
+		if (!exec->cu_addr_virt[i]) {
+			write_unlock(&zdev->attr_rwlock);
```
larry9523: We may want to add DRM_ERR to indicate why the configuration fails

src/runtime_src/driver/zynq/drm/zocl/zocl_drv.c
```diff
@@ -262,17 +282,25 @@ zocl_gem_cma_mmap(struct file *filp, struct vm_area_struct *vma)
 	return rc;
 }
 
+/* This function map two types of kernel address to user space.
+ * The first type is pysical registers of a hardware IP, like CUs.
+ * The seconde type is GEM buffer.
```
larry9523: second

src/runtime_src/driver/zynq/drm/zocl/zocl_drv.c
```diff
@@ -297,20 +325,43 @@ static int zocl_mmap(struct file *filp, struct vm_area_struct *vma)
 		return 0;
 	}
 
-	if (vma->vm_pgoff != 0)
+	/* Hardware component physical address mapping. Typically, this is used
+	 * to map the regsiters of a compute unit to user space.
```
larry9523: registers

src/runtime_src/driver/zynq/drm/zocl/zocl_drv.c
```diff
@@ -297,20 +325,43 @@ static int zocl_mmap(struct file *filp, struct vm_area_struct *vma)
...
+	 *
+	 * For the most of the time, the hardware is at 0 to 4GB address range.
+	 * *NOTE* Base on MPSoC TRM, it is possible to assign hardware to higher
+	 * address than 4GB. But for now, no one use those higher address range
+	 * for IPs. The RPU will has problem when access outside of 4GB memory.
```
larry9523: The RPU will have

src/runtime_src/driver/zynq/drm/zocl/zocl_drv.c
```diff
@@ -297,20 +325,43 @@ static int zocl_mmap(struct file *filp, struct vm_area_struct *vma)
...
+	 */
+	if (!zdev->exec->configured) {
+		DRM_ERROR("Schduler is not configured\n");
+		return -EINVAL;
+	}
+
+	phy_addr = vma->vm_pgoff << PAGE_SHIFT;
+
+	/* Only allow user to map register ranges in apertures list.
+	 * Could not map from the middle of a aperture.
```
larry9523: of an aperture

src/runtime_src/driver/zynq/drm/zocl/zocl_ioctl.c
```diff
@@ -385,6 +385,46 @@ zocl_read_sect(enum axlf_section_kind kind, void *sect,
...
+				/* FIFO_LITE has 4KB and FIFO FULL has 8KB
+				 * address range. Use both 8K is okay.
+				 */
+				zdev->apertures[zdev->num_apts].size = _8KB;
+			else
+				/* Others debug IPs have 64KB address range*/
+				zdev->apertures[zdev->num_apts].size = _64KB;
+			zdev->num_apts++;
+		}
+	}
```
larry9523: We'd better check the value of zdev->num_apts is smaller than MAX_APERTURES. Otherwise, a bad xclbin can cause kernel memory corruption

src/runtime_src/driver/zynq/drm/zocl/zocl_drv.c
```diff
@@ -134,6 +134,26 @@ static int get_reserved_mem_region(struct device *dev, struct resource *res)
...
 }
 
+/**
+ * is_valid_apts - Check the geiven phys address is the start of a aperture
+ *
+ * @dev: DRM device struct
+ * @size: This size was not use, just to match function prototype.
+ *
+ */
+int is_valid_apts(struct drm_zocl_dev *zdev, phys_addr_t addr)
```
larry9523: Nit: I suggest using a function name something like get_apt_num(). Normally is_valid() returns true or false and can be used directly in if statement.

src/runtime_src/driver/zynq/drm/zocl/zocl_ioctl.c
mamin506: Thanks for point this out. I will add a check.
I'm thinking once we are going to support more apertures for embedded customers. we should add aperture numbers in xclbin and dynamically allocate addr_aperture.

src/runtime_src/driver/zynq/drm/zocl/zocl_drv.c
mamin506: Good suggestion, will do.

src/runtime_src/driver/zynq/drm/zocl/zocl_ioctl.c
larry9523: I agree. Even for now, we don't have to define a MAX_APERTURES. we can dynamically allocate space for zdev->apertures based on the count in ip and debug_ip section.

mamin506: Good point. Let me update this soon.

src/runtime_src/driver/zynq/drm/zocl/sched_exec.c
mamin506: done

src/runtime_src/driver/common/scheduler.cpp
```diff
@@ -95,16 +94,45 @@ get_cus(const axlf* top)
   for (int32_t count=0; count <ip_layout->m_count; ++count) {
     const auto& ip_data = ip_layout->m_ip_data[count];
     if (ip_data.m_type == IP_TYPE::IP_KERNEL) {
-      // encode handshaking control in lower unused address bits
       uint64_t addr = ip_data.m_base_address;
-      addr |= ((ip_data.properties & IP_CONTROL_MASK) >> IP_CONTROL_SHIFT);
+      if (encoding) {
```
stsoe: coding style in general says no curly brace for one-line statements

src/runtime_src/driver/common/scheduler.h
```diff
@@ -33,7 +47,7 @@ namespace xrt_core { namespace scheduler {
  * Check sdaccel.ini for default overrides.
  */
 int
-init(xclDeviceHandle handle, const axlf* top);
+init(xclDeviceHandle handle, const axlf* top, bool encode);
```
stsoe: Why is there an encode argument here?   Is 'false' ever passed?

stsoe: In general this file is no longer exclusively about scheduler initialization.  I think I would have preferred leaving scheduler init alone and simply add another file and API for get_cus and get_dbg_ips_pair.   For sure the file name and namespace is now confusing.

src/runtime_src/driver/common/scheduler.h
```diff
@@ -19,12 +19,26 @@
 
 #include "driver/include/xclhal2.h"
 #include "driver/include/xclbin.h"
+#include <map>
```
stsoe: Why include map?

src/runtime_src/driver/zynq/user/shim.h
```diff
@@ -26,6 +26,8 @@
 //#include "driver/zynq/include/zynq_perfmon_params.h"
 #include <cstdint>
 #include <fstream>
+#include <map>
+#include <vector>
```
stsoe: I didn't see vector used, but maybe I didn't look hard enough?

src/runtime_src/driver/zynq/drm/zocl/zocl_ioctl.c
```diff
@@ -391,38 +391,67 @@ zocl_read_sect(enum axlf_section_kind kind, void *sect,
...
 {
 	struct ip_data *ip;
 	struct debug_ip_data *dbg_ip;
+	struct addr_aperture *apt;
+	int total = 0;
 	int i;
 
+	/* Update aperture should only happen when loading xclbin */
+	DRM_INFO("minm clean apertures\n");
+	kfree(zdev->apertures);
```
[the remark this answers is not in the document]
mamin506: This function is used to update zdev->apertures. If it existed, we should free the old one. This is useful when we can reconfigure xclbin. The zocl_drm_platform_remove() do free zdev->apertures.

src/runtime_src/driver/zynq/user/shim.h
mamin506: vector is used in:
int mapKernelControl(const std::vector<std::pair<uint64_t, size_t>>& offsets);

src/runtime_src/driver/common/scheduler.h
mamin506: The xclLoadXclbin in zynq/user/shim.cpp pass 'false' to the scheduler init.
We could have another file which provides XCLBIN related APIs and namesapce.

src/runtime_src/driver/common/scheduler.h
mamin506: removed

src/runtime_src/driver/common/scheduler.cpp
mamin506: done

src/runtime_src/driver/common/scheduler.h
stsoe: I still don't understand the encode argument.  Why will zynq pass 'false'?  How will it know that CUs are configured for dataflow?    Shouldn't it at least be passed the information and filter it out like x86 does?

mamin506: In 2019.1, we are not going to support dataflow for MPSoC. That is why I need to disable the encoding here. The zocl driver could not handle CU physical address with encode handshaking control in the lower unused bits.
We would remove the encode argument after we have dataflow support on MPSoC. Do you think this make sense?

stsoe: No. It still doesn't make sense.   The change should not have been at this level, but instead zocl should have adjusted the addresses in the same way xocl does before it actually uses them.   xocl simply checks the encoding, the clears it.  The checking is zocl is no-op since dataflow is not supported, but the clearing is not.   Is there maybe a reason this cannot be done in zocl?

mamin506: OK. I have made that change in zocl driver. I saw in the mb_scheduler.c. It has "addr = addr & ~(0xFF)" to clear encoding. I do the same in zocl for now.

stsoe:
retest this please

stsoe:
retest this please

PR #1117, opened 2019-03-22
Fixing issue of creating execBuf in MemorySpace in emulation

chvamshi-xilinx:
Hi, 
This change contains a fix which allocates execBuf memory in DATA space


maxzhen:
ok to test

PR #1120, opened 2019-03-23
fix zynq lib xrt_core's dependency on lib uuid

larry9523:
Since driver/common/scheduler.cpp now depends on libuuid, xrt_core depends on uuid as well.


maxzhen:
ok to test

PR #1135, opened 2019-03-26
Adding timeout to connect to device process in emulation

chvamshi-xilinx:
Hi, 

Adding a timeout to connect to device process in emulation. Exiting from the application If connection not established in 300 seconds.

Thanks


maxzhen:
ok to test

PR #1140, opened 2019-03-27
Coverity fixes in emulation

chvamshi-xilinx:
Hi,
This change contains coverity fixes of emulation. This change also contains xclLogMsg changes for emulation
Thanks


maxzhen:
ok to test

PR #1142, opened 2019-03-27
better profile/trace reporting of copy buffers (KDMA and P2P)

pgschuey:


maxzhen:
ok to test

pgschuey:
retest this please

src/runtime_src/xocl/api/plugin/xdp/profile.cpp
```diff
@@ -215,6 +220,22 @@ void get_address_bank(cl_mem buffer, uint64_t &address, std::string &bank)
   }
 }
 
+// Attempt to find if two buffers are resident on the same device
```
pgschuey: @stsoe - We need to differentiate between CDMA and P2P buffer transfers. Is this a good method to do that? I realize clEnqueueCopyBuffer does not yet support P2P transfers, but this is in anticipation of that.

pgschuey:
Good question, @jvillarre. Profile summary and xbutil have always had different formats. While xbutil shows the raw counters (e.g., tranx count, total bytes, etc.), the profile summary shows processed information (e.g., avg bytes = total bytes / tranx count).

maxzhen:
retest this please

PR #1143, opened 2019-03-27
show dataflow in profile summary

hackwa:
parse cu trace with queue to catch parallel tx
simplify trace string


maxzhen:
ok to test

hackwa:
@jvillarre @pgschuey  we currently add major/minor IP versions in debug IP Layout. Do you think I should increment SAM IP version to 1.2? Currently it's 1.1

PR #1147, opened 2019-03-28
CR-1026218 Need PIDs of processes running on device

maxzhen:
@sonals , please take a look and see if this is what you have in mind.

### DIFF
--- a/src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_sysfs.c
+++ b/src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_sysfs.c
@@ -51,21 +51,28 @@ static ssize_t user_pf_show(struct device *dev,
 }
 static DEVICE_ATTR_RO(user_pf);
 
-/* -live client contects-- */
+/* -live client contexts-- */
 static ssize_t kdsstat_show(struct device *dev,
 	struct device_attribute *attr, char *buf)
 {
 	struct xocl_dev *xdev = dev_get_drvdata(dev);
-	int size;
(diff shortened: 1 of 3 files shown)


src/runtime_src/driver/xclng/drm/xocl/userpf/xocl_sysfs.c
```diff
@@ -51,21 +51,28 @@ static ssize_t user_pf_show(struct device *dev,
...
-			   xclbin_id ? xclbin_id : 0,
-			   atomic_read(&xdev->outstanding_execs),
-			   atomic64_read(&xdev->total_execs),
-			   get_live_client_size(xdev));
+	size += sprintf(buf + size, "xclbin:\t\t\t%pUb\n",
+		xclbin_id ? xclbin_id : 0);
+	size += sprintf(buf + size, "outstanding execs:\t%d\n",
+		atomic_read(&xdev->outstanding_execs));
+	size += sprintf(buf + size, "total execs:\t\t%ld\n",
+		atomic64_read(&xdev->total_execs));
```
sonals: I wonder if should keep the original contexts and then add the pids in the following line?

maxzhen:
@sonals , pushed a change to fix this. please take a look.

keryell:
Just curious about the feature implemented here. Is it something similar to running `lsof` on the device driver controlling the card?

maxzhen:
> Just curious about the feature implemented here. Is it something similar to running `lsof` on the device driver controlling the card?

Yes.

PR #1150, opened 2019-03-29
coverity fixes-6

sarab96:
coverity fixes-6
@jmouroux Since you are most familiar with this code so please review the changes

### DIFF
--- a/src/xma/src/xmaapi/xmares.c
+++ b/src/xma/src/xmaapi/xmares.c
@@ -298,8 +298,13 @@ int32_t xma_res_alloc_dec_kernel(XmaResources shm_cfg, XmaDecoderType type,
 
     kern_props =
         xma_res_create_kern_req(xma_res_decoder, vendor, dev_excl);
-    if (!kern_props || !shm_cfg)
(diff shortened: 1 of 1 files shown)


maxzhen:
ok to test

src/xma/src/xmaapi/xmares.c
```diff
@@ -519,8 +544,11 @@ static XmaResConfig *xma_shm_open(char *shm_filename, XmaSystemCfg *config)
     /* Ensure other processes will fail to open properly until initialzied */
     fchmod(fd, 0200);
     ret = ftruncate(fd, sizeof(XmaResConfig));
-    if (ret)
+    if (ret) {
+        fchmod(fd, 0666);
```
jmouroux: Hi @sarab96 .   We don't want' to change the file permissions at this point.  The file must be initialized by the process in control of this file and the file permission bits can only be set to rw-rw-rw when this has completed successfully.   If there is a problem here, it is fatal and we do not wish other processes to read a file that may be in an inconsistent state (i.e. contain an uninitialized mutex).

PR #1151, opened 2019-03-29
CR-1016423 XBUTIL flash to restore board back to golden

maxzhen:
xbutil flash -r will revert board back to golden image.


maxzhen:
retest this please

PR #1152, opened 2019-03-29
Fixing uuid undefined symbol in emulation 

chvamshi-xilinx:
Hi, 
This fix is for CR-1026507 where loading hw_emu library is throwing an error of undefined symbol.
Thanks


maxzhen:
ok to test

PR #1156, opened 2019-03-30
pass MSG_FLAG from tx side and wait tx completion

chienwei-lan:
1. Pass MSG_FLAG_(RESPONSE/REQUEST) from TX side:

2. Wait for TX completion

3. PASS ttl while alloc_msg


maxzhen:
ok to test

PR #1162, opened 2019-04-02
Fix xclLogMsg va_list processing -- cannot call va_start() twice on t…

sonals:
…he same va_list

Renamed xclhal2.h to xrt.h with the former being legacy wrapper. 
Moved experimental XRT APIs to xrt-next.h under experimental directory
Do not document now deprecated register read/write APIs in XMA


src/runtime_src/driver/common/config_reader.cpp
```diff
@@ -54,27 +54,52 @@ get_self_path()
...
   return "";
 }
 
+static std::string
+get_ini_path()
+{
+  std::string full_path;
+  try {
+    auto ini_path = boost::filesystem::path(value_or_empty(std::getenv("XRT_INI_PATH")));
+    full_path = verify_ini_path(ini_path);
```
maxzhen: this seems to assume XRT_INI_PATH contains a directory only. But the original code seems to indicate that it may contain path to the ini file. Intentional behavior change here?

src/runtime_src/driver/include/experimental/xrt-next.h
```diff
@@ -0,0 +1,105 @@
...
+  * Returns 0 on success and nonzero on failure. errno will be set to EMSGSIZE when the passed userspace
+  * buffer is too small for the outbound message. This should only happen in the is_tx=true condition.
+  */
+XCL_DRIVER_DLLESPEC int xclMSD(xclDeviceHandle handle, struct drm_xocl_sw_mailbox *args);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
```
maxzhen: want to also move xclP2pEnable(), xclMapMgmt() and xclOpenMgmt() from xrt.h into this file?

src/runtime_src/driver/common/config_reader.cpp
```diff
@@ -54,27 +54,52 @@ get_self_path()
...
 }
 
+/*
+ * Look for xrt.ini and i not found look for legacy sdaccel.ini
+ */
 static std::string
-get_ini_path()
+verify_ini_path(const boost::filesystem::path& dir)
 {
   try {
```
stsoe: Leave try/catch to calling function, which already is in try/catch block.

src/runtime_src/driver/common/config_reader.cpp
stsoe: yes, I do think you should add the check from old line 63/64.   I know I have scripts that specify full path to ini file, which seems more intuitive that just specifying the directory with the ini file.

src/runtime_src/driver/xclng/xrt/user_gem/shim.cpp
```diff
@@ -382,7 +385,7 @@ int xocl::XOCLShim::xclLogMsg(xclDeviceHandle handle, xclLogMsgLevel level, cons
     len++; //To include null terminator
```
stsoe: Never use post decrement unless the previous value is needed.

src/runtime_src/driver/include/experimental/xrt-next.h
sonals: Done

src/runtime_src/driver/common/config_reader.cpp
```diff
@@ -54,27 +54,52 @@ get_self_path()
 #endif
 }
 
+/*
+ * Look for xrt.ini and i not found look for legacy sdaccel.ini
```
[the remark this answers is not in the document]
sonals: done

src/runtime_src/driver/common/config_reader.cpp
sonals: If ENV is set (XRT or SDACCEL) we now require the file to be present or else we give up. No fall back to other directories any more. Note this change is currently limited to config_reader in XRT. I still need to update OCL version.

src/runtime_src/driver/common/config_reader.cpp
sonals: Done

src/runtime_src/driver/xclng/xrt/user_gem/shim.cpp
sonals: Done

PR #1164, opened 2019-04-02
fixing remaining coverity issues

chvamshi-xilinx:
Hi ,

This change contains fixes of remaining coverity issues. There are still two coverity issues are remaining which are risky at this stage.

Thanks


maxzhen:
ok to test

PR #1167, opened 2019-04-02
Fix for monitoring OpenCL APIs from multiple threads

jvillarre:
This portion of XDP supported single threaded applications, but OpenCL APIs called from different threads were previously overwriting their values, resulting in a profile summary that sometimes contained negative values.


maxzhen:
ok to test

PR #1174, opened 2019-04-03
ignore dataflow for hw emulation in profile summary

hackwa:


maxzhen:
ok to test

hackwa:
hey @pgschuey this table is only generated from monitor IP data. In emulation, we call trace logger three times. one for every monitor type. In that case this value could be zero if IP type isn't SAM. Actually good idea, I can make it more clear in code.  Let me do that

PR #1176, opened 2019-04-04
Change the AppDebug warning message to indicate that only functional OpenCL API call will trigger the loading of XDP

tianhaoz95:
Some users were confused by the message that AppDebug will only become available after the first OpenCL call, but it doesn't after the OpenCL object construction. 

The behavior is correct, but the warning message is not completely correct. OpenCL API call is referring to functional API calls. The default OpenCL object constructor only does the memory allocation which won't trigger any loading actions.

Added the word "functional" and "variable initialization won't work" into the warning message.


tianhaoz95-xilinx:
Changed the wording

PR #1177, opened 2019-04-04
final coverity issue in emulation

chvamshi-xilinx:
Hi ,

This change contains one final coverity issue in emulation


maxzhen:
ok to test

PR #1178, opened 2019-04-04
Fixing Return value when no memory is available in specific DDR/PLRAM

chvamshi-xilinx:
Hi ,

This change contains a fix which returns -1 when no memory is available in requested DDR/PLRAM

This change also contains a coverity fixes merged from 2018.3

Thanks


maxzhen:
ok to test

PR #1182, opened 2019-04-05
mailbox pseudo device support

xuhz:
Signed-off-by: Brian Xu <brianx@xilinx.com>

Create mailbox pseudo device, and add open/close/read/write/poll fops entries, so that user app by using select/poll syscall can wait on multiple fpga boards for both tx from within kernel and rx from outside (recv())


maxzhen:
ok to test

xuhz:
new commit addresses Max's comments:
1. issue about mailbox_write() return value once the call failed.
2. coding style issue.

PR #1183, opened 2019-04-06
Fix for CR1027601

jvillarre:
Adding the capability to read dataflow activity from the performance monitors inserted into the design.  This will fix the profile summary compute unit utilization table on Zynq platforms.


maxzhen:
ok to test

PR #1187, opened 2019-04-06
clean up AppDebug naming

tianhaoz95:
Remove all the SDx related namings in AppDebug in preparation for the more generic naming scheme in the future.


maxzhen:
ok to test

PR #1188, opened 2019-04-07
Preparation for next changes

sarab96:
XMA: Preparation for next changes and merge from 2018.3


maxzhen:
ok to test

PR #1190, opened 2019-04-08
fixing emulation driver compilation issue for zynq platform

Himanshu-xilinx:
* fixing emulation driver compilation issue for zynq platform


maxzhen:
ok to test

PR #1193, opened 2019-04-08
data_flow support of emulation

chvamshi-xilinx:
Hi 
This change contains data_flow support for emulation
Thanks


maxzhen:
ok to test

PR #1199, opened 2019-04-09
log cu times on per binary basis

hackwa:


maxzhen:
ok to test

hackwa:
I did not test it specifically against that testcase. I created my own test in hw emulation for this.

hackwa:
hey @pgschuey since the banks aren't guaranteed to be contiguous, I didn't implement PLRAM[0:1] ..

pgschuey:
Actually, they are guaranteed to be contiguous, so a min:max format makes sense.

hackwa:
What If the user decides to call migrate mem objects on DDR 1 and 3?

pgschuey:
I have not tried it but was told awhile ago memory is always contiguous. I would guess it either 1. errors out, or 2. allocated DDR[1:3]

hackwa:
@stsoe @sonals  even if the memory assignment is always contiguous, can the migrate statement still take buffers that reside in non contiguous memory? For example: If user design has access to DDR 1..3 can user choose to use only 1 and 3?

PR #1202, opened 2019-04-10
adding an ini option to use legacy ddr model in hw_emu

chvamshi-xilinx:
Hi.

This change contains adding a ini option in the emulation section to use legacy ddr model.

Thanks


maxzhen:
ok to test

PR #1203, opened 2019-04-10
Error out if specify host pointer and p2p flag

chienwei-lan:
We should error out if user tries to create a P2P buffer and assign host pointer to it.

CR: http://jira.xilinx.com/browse/CR-1024102


maxzhen:
ok to test

maxzhen:
retest this please

PR #1205, opened 2019-04-11
Replacing old terminology of CPU emu with Software Emulation

maheshattarde-xilinx:
This message will be presented to user, user is aware of sw_emu as XCL_EMULATION_MODE.
We dont wish to use multiple terminologies, hence replacing word cpu by software


maxzhen:
ok to test

PR #1207, opened 2019-04-11
Json schema change

rradjabi:
Addresses JIRA CR-1024994

Removes duplicate entries for things like compute units, memory. Instead of having memory entries listed as:
```
memory: {
    mem: {
        index: 0
         ....
    },
    mem: {
         index: 1
         ....
    }
```

We have replaced "mem" with "index" and removed the "index" entry. Here is an example output:
```
        "memory": {
            "mem": {
                "0": {
                    "type": "**UNUSED**",
                    "temp": "32",
                    "tag": "bank0",
                    "enabled": "false",
                    "size": "16 GB",
                    "mem_usage": "0 Byte",
                    "bo_count": "0"
                },
                "1": {
                    "type": "MEM_DDR4",
                    "temp": "35",
                    "tag": "bank1",
                    "enabled": "true",
                    "size": "16 GB",
                    "mem_usage": "0 Byte",
                    "bo_count": "0"
                },
                "2": {
                    "type": "**UNUSED**",
                    "temp": "37",
                    "tag": "bank2",
                    "enabled": "false",
                    "size": "16 GB",
                    "mem_usage": "0 Byte",
                    "bo_count": "0"
                },

```

This conforms to a more-standard format for parsing entry lists. Standard JSON parsers do not support duplicate keys. 

Also adding the json schema version, named "version" at root. This change bumps from 1.0.0 to 1.1.0. I should back-port the version into 2018.3 and keep as 1.0.0.

### DIFF
--- a/src/runtime_src/driver/xclng/tools/xbutil/xbutil.h
+++ b/src/runtime_src/driver/xclng/tools/xbutil/xbutil.h
@@ -84,7 +84,7 @@ enum subcommand {
     STATUS_SPM,
     STATUS_LAPC,
     STATUS_SSPM,
-    STATUS_SPC, 
+    STATUS_SPC,
     STREAM,
     STATUS_UNSUPPORTED,
 };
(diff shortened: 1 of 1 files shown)


maxzhen:
ok to test

PR #1213, opened 2019-04-12
clean up error checking that is no longer needed after xdp refactoring

tianhaoz95:
The reading of the IP has been abstracted from this layer, so the error handling is no longer necessary.


maxzhen:
ok to test

PR #1216, opened 2019-04-12
bug fix in software mailbox

rradjabi:
proposed by Brian Xu, this fixes software mailbox transfers that require userspace buffer resizes to complete


maxzhen:
ok to test

PR #1219, opened 2019-04-12
Only map soft kernel regfile once.

larry9523:
We do not need to map soft kenel regfile BO every time we schedule the soft kernel to run.
Also this PR clean up the BO resources after we exit from soft kernel loop.


maxzhen:
ok to test

PR #1225, opened 2019-04-15
Shared library for core utilities

stsoe:
In order to facilitate sharing of driver/common code, this code is now
delivered in shared library xrt_coreutil.

The original files that resided under runtime_src/xrt/util are removed
or modified to use xrt_coreutil instead.  This consolidates the two
versions we had of config_reader and message into the one version
under driver/common.

This PR also changes shared library linking for all of XRT to require fully 
resolved libraries.


src/runtime_src/CMakeLists.txt
```diff
@@ -62,6 +62,7 @@ endif()
 
 # Attach to the user's linker flags
 set (CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-Bsymbolic")
+set (CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-z,defs")
```
sonals: Does -z defs include the semantics of -Bsymbolic? We used the latter so our internal calls to OCL APIs were resolved within our library and not redirected to ICD.

stsoe: I am not aware of  zdef affecting anything but to enforce that libraries we build are fully resolved.

PR #1228, opened 2019-04-15
Fix race in acquire_context 

stsoe:
CR1027850: Multiple threads in host code compete for acquiring CU context


stsoe:
retest this please

PR #1229, opened 2019-04-16
CR-1028200 After download xclbin with ap_ctrl_none kernel, xbutil pro…

houlz0507:
…gram hangs on u200_qdma_201910_1 DSA


houlz0507:
@stsoe I just had a driver fix to protect ERT. This will unblock Sprite. I think you may add the check to ERT userland code later.

stsoe:
Hi Lizhi.  I don't understand the point the change?  I understand the code, but how is this situation occurring and why will the function bail with return code 1?

houlz0507:
There is new type of qdma kernel which does not have kernel axi-lite interface. This kind of kernel starts to run when it is downloaded. And it looks ERT has issues with accessing non-exist kernel interface(CR-1028200). Instead of looking into ERT itself, I think it makes sense to avoid any ERT command in this scenario because CU axi-lite registers are not accessible.

stsoe:
Shouldn't this be controlled from the place where the configure command is issued?   Should the configure command even be sent in this case?  Does the scheduler have any role at all?

PR #1231, opened 2019-04-16
Fix for malformed xbutil formatting

jvillarre:
Adding a statement to strip away extra null characters copied into a string, which adjusts the formatting.


maxzhen:
ok to test

PR #1232, opened 2019-04-16
Reset ECC when subdev drive probe

chienwei-lan:


chienwei-lan:
http://jira.xilinx.com/browse/CR-1026919

maxzhen:
ok to test

PR #1235, opened 2019-04-16
Adding frame rate property for XMA filter plugin, CR-1023409

uday610:
Added frame rate property for XMA filter plugin, CR-1023409


maxzhen:
ok to test

PR #1236, opened 2019-04-16
[MPSoC] Link libxrt_coreutil.so library

mamin506:
After #1225, we should link libxrt_coreutil.so for MPSoC platform.


maxzhen:
ok to test

PR #1237, opened 2019-04-16
Add PIDs of live processes in warning message 

IshitaGhosh:
Add PIDs of live processes in warning message for unavailable Device debug and profiling data when multiple live processes are running on device


maxzhen:
ok to test

src/runtime_src/driver/xclng/xrt/user_gem/shim.cpp
```diff
@@ -1841,18 +1845,62 @@ uint xocl::XOCLShim::xclGetNumLiveProcesses()
...
+
+    /* Dependent on message format built in kdsstat_show. Checking number of "context" and "client pid" in kdsstat.
+     * kdsstat has
+       "context: <number_of_live_processes>
+        client pid:
+                    <pid> 
+                    <pid>
+       "
+     */ 
     if(stringVec.size() >= 4) {
```
stsoe: This is much too complicated parsing which will be close to impossible to get right when/if we change the format of kdsstat.  Please use std::regex for exacting what you need.

maxzhen:
Closing it for now. Please reopen with updated change to address all conflicts and comments.

PR #1239, opened 2019-04-16
Print * if userpf is not ready

chienwei-lan:


chienwei-lan:
http://jira.xilinx.com/browse/CR-1027132

maxzhen:
ok to test

PR #1240, opened 2019-04-16
CR-1028029 stream example running on different DSA hangs

houlz0507:


src/runtime_src/driver/xclng/drm/xocl/subdev/qdma.c
```diff
@@ -359,6 +362,23 @@ static int stream_sysfs_create(struct stream_queue *queue)
...
 	}
 
+	for (i = 0; i < QDMA_QSETS_MAX; i++) {
+		temp_q = queue->qdma->queues[i];
+		if (!temp_q)
+		       continue;
+		if (temp_q->qconf.c2h && temp_q->flowid == queue->flowid) {
+			xocl_info(&pdev->dev,
+				"flowid overlapped with queue %d", i);
+			return 0;
```
maxzhen: do you need to goto failed?

houlz0507: This is only for statistics. I am not very confident to fail the queue creation if there is flowid, route id overlap.

PR #1245, opened 2019-04-17
Mpd passthru refactor

rradjabi:
* Remove software mailbox IOCTLs, and replaced with xclMailbox and xclMailboxMgmt
* Update the passthrough daemon to use xclMailbox APIs, and use read/write APIs that are supported via pseudo-device (https://github.com/Xilinx/XRT/pull/1182)


maxzhen:
ok to test

PR #1246, opened 2019-04-17
Fix ICD loader problem after adding libxrt_coreutil.so dependency

stsoe:
When using ICD loader all dependencies of libxilinxopencl.so must be
located.  Since LD_LIBRARY_PATH is not set, the dependent
libxrt_coreutil.so cannot be found.  To resolve this, make sure to set
run-time search path (RPATH) in libxilinxopencl.so to ORIGIN.


stsoe:
Sonal.  I am adding you as a reviewer of this change for FYI. There is more fallout from adding dependency on .so, but I think this is probably the most important one.  Until this fix, the ICD loader would not have worked.

PR #1251, opened 2019-04-18
Add changes for using gcc7.3 on AL2 with 4.14.94-89.73.amzn2.x86_64

awshkulkar:
This change only includes changes needed to build (compile/link) the XRT on AL2. A newPR will be submitted for creating install-able rpm package.

### DIFF
--- a/src/runtime_src/xocl/CMakeLists.txt
+++ b/src/runtime_src/xocl/CMakeLists.txt
@@ -13,6 +13,8 @@ include_directories(
   ${XRT_XOCL_API_DIR}
 )
 
+add_compile_options("-Wno-nonnull")
+
 file(GLOB XRT_XOCL_API_FILES
   "${XRT_XOCL_API_DIR}/*.h"
   "${XRT_XOCL_API_DIR}/*.cpp"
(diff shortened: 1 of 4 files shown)


maxzhen:
ok to test

awshkulkar:
I will add the warning suppression conditionally for gcc version 7.3 and up. Pleas do not merge this commit yet.

src/runtime_src/xocl/CMakeLists.txt
```diff
@@ -13,6 +13,8 @@ include_directories(
   ${XRT_XOCL_API_DIR}
 )
 
+add_compile_options("-Wno-nonnull")
```
stsoe: What's the reason for disabling this warning?

awshkulkar: The code is in boot devel package, not in XRT.

stsoe: I don't understand what that means?  This is a CMakeList.txt in XRT.

awshkulkar: I take it back. I was looking at a different log. You are right. I see the issue here:
 499   | if (flags & CL_MEM_COPY_HOST_PTR)
  500   | | std::memcpy(m_host_ptr,host_ptr,sz);
There is no null check. I will fix this code and then remove this supression. Will push a new commit.

PR #1255, opened 2019-04-18
axis monitoring without tlast, Use unique ID for sw CU events and reset accelerator monitor using hal

hackwa:
First commit lets us see data transfer for cases when tlast isn't present in AXI Stream
Second commit lets us use unique ID for every CU in sw emuation, allowing us to show overlaps
Third commit lets us reset accelerator monitors from hal at beginning of program


maxzhen:
ok to test

stsoe:
retest this please

PR #1260, opened 2019-04-19
Fixing KDS issue in emulation

chvamshi-xilinx:
Hi 

This change contains a fix for KDS in emulation. KDS was not scheduling free compute units.

Thanks
Vamshi Krishna


stsoe:
ok to test

PR #1261, opened 2019-04-19
Clarifying the error output from xgdb

jvillarre:
When calling "xstatus all" inside of xgdb, we go through all the IP types and try to read them.  In emulation flows, this results in the same message being printed several times.  This change clarifies that the error messages are coming from different calls and are all valid.


maxzhen:
ok to test

PR #1264, opened 2019-04-19
update help message to indicate how to deal with multi-card system in xbutil status

tianhaoz95:
In a recent CR, the tester was confused by `xbutil status --spm` on a P2P system where he though the `xbutil status --spm` was showing incorrect/empty data. `xbutil` was actually behaving correctly, but the documentation was indicating that xbutil should iterate through the devices as it doesn't take a device index argument. However, `xbutil status` does take a device index and behaves exactly like other `xbutil` commands, so I added in the documentation in the help message in this pull request.


maxzhen:
ok to test

PR #1265, opened 2019-04-19
added rwlock to protect sysfs access for mig and dna

houlz0507:


houlz0507:
The patch is not the latest which does not contain DNA stuff. I will file another one.

PR #1273, opened 2019-04-22
thread safety added to schedule_work and work_item_done APIs

sarab96:
Both APIs schedule_work and work_item_done are allowed to be used asynchronously in different threads.
So added thread safety.

@sonals Please check


src/xma/include/lib/xmahw.h
```diff
@@ -47,12 +49,23 @@ typedef struct XmaHwKernel
     uint32_t    ddr_bank;
     //For execbo:
     int32_t     kernel_complete_count;
+    std::unique_ptr<std::condition_variable> kernel_complete_count_cv;
```
stsoe: Why pointer?  Also, I don't see the condition variable used any where?

src/xma/include/lib/xmahw.h
```diff
@@ -47,12 +49,23 @@ typedef struct XmaHwKernel
...
+    std::unique_ptr<std::atomic<bool>> kernel_complete_count_locked;
+
     void*       kernel_cmd_queue;
     void*       kernel_cmd_completion_queue;
     uint32_t    kernel_execbo_handle[MAX_EXECBO_POOL_SIZE];
     char*       kernel_execbo_data[MAX_EXECBO_POOL_SIZE];
     bool        kernel_execbo_inuse[MAX_EXECBO_POOL_SIZE];
     uint32_t    reserved[16];
+
+    XmaHwKernel(): kernel_complete_count_cv(new std::condition_variable),
```
stsoe: Is there any reason why this is heap allocated?

src/xma/include/lib/xmahw.h
```diff
@@ -47,12 +49,23 @@ typedef struct XmaHwKernel
...
+      kernel_complete_count_locked(new std::atomic<bool>) {
```
stsoe: heap allocation?

maxzhen:
ok to test

src/xma/include/lib/xmahw.h
sarab96: I will remove unused condition_variable.
I'm changing c code to c++ in small portions. Later I may change arrays to vector so I want to keep it as pointer.

src/xma/include/lib/xmahw.h
sarab96: I will remove condition_variable

src/xma/include/lib/xmahw.h
sarab96: I think it is okay to have it on heap. Later I may change arrays to vector. So it should be fine keeping in mind overall code architecture. Any suggestions?

stsoe: Yes, please don't use heap allocation unless you need to. Array or not it doesn't matter; at least I can't see why.

src/xma/include/lib/xmahw.h
sarab96: changed

src/xma/include/lib/xmahw.h
sarab96: changed

src/xma/include/lib/xmahw.h
sarab96: I looked at the code. Later on vector will be used to hold this struct. Num of structs are not fixed as it depends on xclbin. There will be one struct for one CU. So number of std::atomic depends on number of CUs in an xclbin.
Though I have changed to make_unique for better protection. I think this is fine now. Please check.

sarab96:
retest this please

sarab96:
Will resubmit after changing HwSession to internal private structure

PR #1274, opened 2019-04-22
CR-1026385 Adding frame rate property for XMA Scaler plugin

cdeshpan-xlnx:
Added  frame rate property to XmaScalerInOutProperties. 

Added debug logs for triaging specific error when XMA initialization fails.


maxzhen:
ok to test

PR #1277, opened 2019-04-22
xbutil list should only list usable devices

chienwei-lan:
Now we list all devices regardless they are usable or not,

and we should only usable devices.

http://jira.xilinx.com/browse/CR-1027132


maxzhen:
ok to test

PR #1278, opened 2019-04-22
Check Ubuntu version upon package install

rradjabi:
Check for glibc version:
16.04 expects libc6 version 2.23
18.04 expects libc6 version 2.27

Build Machine |  Install Machine |    Status
16.04               |  18.04                 |    negative test, failed to install, PASS
18.04               |  18.04                 |    positive test, successful install, PASS
18.04               |  16.04                 |    negative test, untested
 
Addresses CR-1017740


stsoe:
ok to test

src/CMake/cpack.cmake
```diff
@@ -36,6 +36,11 @@ if (${LINUX_FLAVOR} MATCHES "^(Ubuntu|Debian)")
   SET(CPACK_DEBIAN_AWS_PACKAGE_DEPENDS "xrt (>= ${XRT_VERSION_MAJOR}.${XRT_VERSION_MINOR}.${XRT_VERSION_PATCH})")
   # xrt component dependencies
   SET(CPACK_DEBIAN_XRT_PACKAGE_DEPENDS "ocl-icd-opencl-dev (>= 2.2.0), libboost-dev (>=1.58), libboost-filesystem-dev (>=1.58), uuid-dev (>= 2.27.1), dkms (>= 2.2.0), libprotoc-dev (>=2.6.1), protobuf-compiler (>=2.6.1), libncurses5-dev (>=6.0), lsb-release, libxml2-dev (>=2.9.1), libyaml-dev (>= 0.1.6)")
+  if (${LINUX_VERSION} STREQUAL 16.04)
+    SET(CPACK_DEBIAN_XRT_PACKAGE_DEPENDS "${CPACK_DEBIAN_XRT_PACKAGE_DEPENDS}, libc6 (>= 2.23), libc6 (< 2.27)")
+  else ()
```
sonals: Add an explicit check for 18.04 version as well

rradjabi: What should we do in the else condition where version isn't 18.04 or 16.04?

sonals: Maybe we can get the glibc version dynamically?
Can we try something like this with proper credits--
https://gist.github.com/likema/f5c04dad837d2f5068ae7a8860c180e7

keryell: Try on xsjsycl40 and you will know... :-) Otherwise, add 18.10 and 19.04, since they are still valid versions. Probably that picking the highest LTS will just work because the package versions on the target will have higher versions anyway...

rradjabi: I'm not sure how to use that glibc macro and ensure that the build version matches the install version.

maxzhen:
retest this please

src/CMake/cpack.cmake
```diff
@@ -21,6 +21,21 @@ execute_process(
...
+    OUTPUT_VARIABLE GLIBC
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+get_filename_component(GLIBC ${GLIBC} REALPATH)
+get_filename_component(GLIBC_VERSION ${GLIBC} NAME)
+string(REPLACE "libc-" "" GLIBC_VERSION ${GLIBC_VERSION})
+string(REPLACE ".so" "" GLIBC_VERSION ${GLIBC_VERSION})
+if(NOT GLIBC_VERSION MATCHES "^[0-9.]+$")
+    message(FATAL_ERROR "Unknown glibc version: ${GLIBC_VERSION}")
+endif(NOT GLIBC_VERSION MATCHES "^[0-9.]+$")
+
```
sonals: Please keep this in its own file like glibc.cmake, include it in cpack.cmake.

maxzhen:
retest this please

PR #1283, opened 2019-04-23
CR-1023520 CTRL-C with xbutil top causes kernel panic

maxzhen:


maxzhen:
retest this please

maxzhen:
The pipeline failures are all emulation failures, which has nothing to do with this change. Merging it now.

PR #1284, opened 2019-04-23
Fixing emulation issue for ZynqMP platform, Also fixing scheduler issue in emulation

chvamshi-xilinx:
Hi,

This change contains following fixes
- populating xclbin info in emulation for ZynqMP platform
- Scheduling multiple compute unit of same kernel in emulation


maxzhen:
ok to test

PR #1286, opened 2019-04-23
Providing support for scout software emulation 

maheshattarde-xilinx:
Software emulation launches emulated device process from  software emulation driver. This device process is located relative to  XILINX_SCOUT area. In case of XILINX_SDX is set along with XILINX_SCOUT, Emulation process should be launched from XILINX_SCOUT area.


maxzhen:
ok to test

PR #1290, opened 2019-04-23
    Cleaned up the build flow

rozumx:
Work Done
    ---------
    + Replaced curl calls with CMake equivalent functionality
    + Changed the BMC generation schema.  Use '-' instead of ':'.
    + Replaced mkdir bash calls with CMake equivalent functionality
    + Removed curl support

### DIFF
--- a/src/CMake/native.cmake
+++ b/src/CMake/native.cmake
@@ -29,7 +29,7 @@ ENDIF(OPENCL_FOUND)
 find_package(Git)
 
 IF(GIT_FOUND)
-  message("-- Looking for GIT - found at ${GIT_EXECUTABLE}")
+  MESSAGE(STATUS "Looking for GIT - found at ${GIT_EXECUTABLE}")
 ELSE(GIT_FOUND)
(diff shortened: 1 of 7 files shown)


stsoe:
ok to test

src/CMake/native.cmake
```diff
@@ -25,10 +25,11 @@ ELSE(OPENCL_FOUND)
   MESSAGE(FATAL_ERROR "Looking for OPENCL - not found")
 ENDIF(OPENCL_FOUND)
 
+# -- Git --
 find_package(Git)
 
 IF(GIT_FOUND)
-  message("git found: ${GIT_EXECUTABLE}")
+  message("-- Looking for GIT - found at ${GIT_EXECUTABLE}")
```
mamin506: Should this be "message(STATUS "Looking for GIT - found at ${GIT_EXECUTABLE}")" ?

rozumx: Update this pull request with the requested changes.

PR #1295, opened 2019-04-24
local variables shouldn't exceed 5-10

chienwei-lan:
Follow Linux kernel coding style guide:

The local variables should not exceed 5-10


maxzhen:
ok to test

PR #1296, opened 2019-04-24
Fix the testcase issue: disable unsupported DSA and update --sp tag with new syntax

haeseung81:
Fix the testcase issue: 
1) disable unsupported DSA (xilinx_u280-es1_xdma_201910_1 and xilinx_u200_qdma_201910_1)
2) update --sp tag with new syntax


stsoe:
ok to test

PR #1302, opened 2019-04-25
Disable unsupported Shell for 15_buffer_size 

haeseung81:
15_buffer_size allocates large device buffers that HBM in u280-es1 can't handle.
So, disable the u280-es1 for this test.


stsoe:
ok to test

PR #1307, opened 2019-04-25
Enhance clEnqueueCopyBuffer to cover p2p

stsoe:
clEnqueueCopyBuffer now handles buffers local to enqueued device and
cases where either src or dst is imported from another device.  This
eliminates the need for xclEnqueuePeerToPeerCopyBuffer, which will be
removed once all tests have been updated.

HW emulation driver is expected to handle copy buffer exactly as HW
driver.  SW emulation does not support p2p because scheduling is
through software scheduler (sws) which is not enhanced to handle
ERT_START_COPYBO.


stsoe:
Vamshi.  This is the PR for what we talked about Thursday.   The change of interest is xocl::device::copy_buffer, all the other files touched are minor supporting changes.  I will send you two working p2p tests that pass in HW with these changes.  I didn't try HW emulation.

PR #1308, opened 2019-04-25
Changing Zynq Shim masking of virtual apertures

jvillarre:
Since some profile IP addresses were masked and inaccessible, this changes the low-order bit mask for all profiling IP to be able to identify IP with 4K and 8K ranges.

Also, changing the number of elements we are reading from the actual hardware to reflect the amount actually stored on the hardware.


stsoe:
ok to test

PR #1314, opened 2019-04-28
Fix a memory leak in auto-configure

sonals:
Correct xmc message to info from error


keryell:
An interesting example of bad OO design, where the functionalities of `buffer_object` have been actually implemented inside `buffer`, leading to these kinds of problems... :-(

stsoe:
There is no OO design here.  The buffer_object is a wrapper that wraps C objects.  Please elaborate as to what you mean?

keryell:
I have the feeling that the wrapping logic should be moved inside `buffer_object` (which is still C++ here) instead of having some mechanics implemented outside...

PR #1315, opened 2019-04-28
CR-1026071 Multi-process : Add 50 ms delay before checking read bit of database file

cdeshpan-xlnx:
If two processes tries to access database file and first process get the lock on data base file and before 1st process release lock on database file or before first process change permission of database file to 666, 2nd process timed out after 5ms as database file doesn't have read permission after waiting for 5 ms on slower systems.

To resolve above problem we have increased time delay from 5ms to 50ms.


stsoe:
ok to test

PR #1319, opened 2019-04-29
Fixing issue in emulation on ZynqMP Platform

chvamshi-xilinx:
Hi,

This change contains a fix in emulation drivers where loading xclbin is failing to parse xclbin xml file.

Thanks


sonals:
ok to test

PR #1353, opened 2019-05-04
CU init support on embedded

larry9523:
1. Add support to initialize CUs on embedded platform. This is used to initiliaze CUs just once whle it is not running.
2. Eliminate duplicated ert commands definition in zocl driver

### DIFF
--- a/src/runtime_src/driver/include/ert.h
+++ b/src/runtime_src/driver/include/ert.h
@@ -109,6 +109,44 @@ struct ert_start_kernel_cmd {
   uint32_t data[1];          /* count-1 number of words */
 };
 
+/**
+ * struct ert_init_kernel_cmd: ERT initialize kernel command format
+ * this command initializes CUs by writing CU registers. CUs are
+ * represented by cu_mask and extra_cu_masks.
+ *
+ * @state:           [3-0] current state of a command
+ * @extra_cu_masks:  [11-10] extra CU masks in addition to mandatory mask
+ * @count:           [22-12] number of words following header
+ * @opcode:          [27-23] 0, opcode for init_kernel
+ * @type:            [31-27] 0, type of init_kernel
+ *
+ * @cu_mask:         first mandatory CU mask
+ * @data:            count-9 number of words representing interpreted payload
+ *
+ * The packet payload is comprised of reserved id field, 8 reserved fields,
+ * a mandatory CU mask, and extra_cu_masks per header field, followed by a
+ * CU register map of size (count - (9 + extra_cu_masks)) uint32_t words.
+ */
+struct ert_init_kernel_cmd {
+  union {
+    struct {
+      uint32_t state:4;          /* [3-0]   */
+      uint32_t unused:6;         /* [9-4]  */
+      uint32_t extra_cu_masks:2; /* [11-10]  */
+      uint32_t count:11;         /* [22-12] */
+      uint32_t opcode:5;         /* [27-23] */
+      uint32_t type:4;           /* [31-27] */
+    };
+    uint32_t header;
+  };
+
+  uint32_t reserved[8];      /* reserved for future use */
+
+  /* payload */
+  uint32_t cu_mask;          /* mandatory cu mask */
+  uint32_t data[1];          /* count-9 number of words */
+};
+
 #define KDMA_BLOCK_SIZE 64   /* Limited by KDMA CU */
 struct ert_start_copybo_cmd {
   uint32_t state:4;          /* [3-0], must be ERT_CMD_STATE_NEW */
@@ -313,6 +351,7 @@ enum ert_cmd_opcode {
   ERT_SK_CONFIG     = 8,
   ERT_SK_START      = 9,
   ERT_SK_UNCONFIG   = 10,
+  ERT_INIT_CU       = 11,
 };
 
 /**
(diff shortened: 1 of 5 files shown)


maxzhen:
ok to test

src/runtime_src/driver/zynq/drm/zocl/sched_exec.c
```diff
@@ -454,6 +477,105 @@ static irqreturn_t sched_exec_isr(int irq, void *arg)
...
+		 * We don't have uninitialized CU for this 32 mask_id.
+		 * Move to next 32 CUs.
+		 */
+		if (!uninited_mask)
+			continue;
+
+		for (i = 0; i < 32; ++i) {
+			int cu_idx;
+
+			if (!(uninited_mask & (1 << i)))
```
mamin506: Not sure whether "ffs_or_neg_one(u32 mask)" is more efficiency than check each bit in a mask. Pointing this out for more thought. I'm good with current implementation.

src/runtime_src/driver/zynq/drm/zocl/sched_exec.c
```diff
@@ -454,6 +477,105 @@ static irqreturn_t sched_exec_isr(int irq, void *arg)
...
+
+			cu_idx = cu_idx_from_mask(i, mask_idx);
+
+			if (busy_mask & (1 << i)) {
+				DRM_WARN("Can not init CU %d while running.\n",
+				   cu_idx);
+				continue;
+			}
+
+			if (cu_idx >= zdev->exec->num_cus) {
```
mamin506: This is good. I'm just thinking, in the future, we can have sanity check for user command at a earlier stage.

src/runtime_src/driver/zynq/drm/zocl/sched_exec.c
```diff
@@ -454,6 +477,105 @@ static irqreturn_t sched_exec_isr(int irq, void *arg)
...
+static void
+init_cu_by_idx(struct sched_cmd *cmd, int cu_idx)
+{
+	u32 i, size = regmap_size(cmd);
+	u32 *virt_addr = cu_idx_to_addr(cmd->ddev, cu_idx);
+	struct ert_init_kernel_cmd *ik;
+
+	ik = (struct ert_init_kernel_cmd *)cmd->packet;
+
+	/* Initialize CU by writing register map starting at
```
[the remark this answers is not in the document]
larry9523: done

src/runtime_src/driver/zynq/drm/zocl/sched_exec.c
larry9523: This should not matter that much since this cu_init is done just once. The reason I did not choose ffs is because we need an extra step to clear the bit we found in uninited_mask by ffs.

src/runtime_src/driver/zynq/drm/zocl/sched_exec.c
larry9523: Yes I agree. And we also need to check the start CU command. Now, a mistaken cu_mask in ert command can easily panic the system by unmapped CU address.

src/runtime_src/driver/include/ert.h
```diff
@@ -109,6 +109,44 @@ struct ert_start_kernel_cmd {
...
+ * @type:            [31-27] 0, type of init_kernel
+ *
+ * @cu_mask:         first mandatory CU mask
+ * @data:            count-9 number of words representing interpreted payload
+ *
+ * The packet payload is comprised of reserved id field, 8 reserved fields,
+ * a mandatory CU mask, and extra_cu_masks per header field, followed by a
+ * CU register map of size (count - (9 + extra_cu_masks)) uint32_t words.
+ */
+struct ert_init_kernel_cmd {
```
stsoe: Besides the reserved[8] what is the difference between this command and ert_start_kernel_cmd? How is this command used differently?

larry9523: There are cases that CUs are required to be initialized once and only once before it can be run. This init kernel command is designed for that purpose. The differences between this one and ert_start_kernel cmd are
1) This command just initialize CUs but not trigger them to run. For given CUs, this command copy the regmap to the CUs register map, but not the first 16bytes (first 4 WORDs); while ert_start_kernel_cmd will not only copy the regmap, but set AP_START and wait for AP done, etc.
2) For a given CU, if it has been initilized already,  it won't be intialized again; while ert_start_kernel_cmd can be triggered multiple times.
3) This command will initialize all CUs in cu_mask and cu_ext_mask; while ert_start_kerne_cmd will pick only one free CU from the masks and run it.
4) Some of the reserved field will be replaced by CU timeout value. The CU timeout/reset feature will follow after this one when the hardware (CU) support reset feature.

stsoe: ert_start_kernel_cmd is just a packet, it's behavior is from the opcode assigned to the packet.  In hindsight maybe including _start_ in the type name is bad, e.g. there is no reason the existing packet couldn't be used as you describe with the opcode you define.

The only difference that may warrant the new command is the reserved fields, but I am not sure how they are used, e.g. why the are exposed in the packet itself?

larry9523: some of the reserved fields will be replaced by cu_timeout and cu_reset_timeout.

PR #1355, opened 2019-05-04
 KDMA copying requires 64 byte block size

stsoe:
Don't use KDMA if src or dst offset or total copy size is not a
multiple of 64 bytes.

This PR also reverts a mistake made in PR #1307 where exec_buf in C++
land was changed to return a code.  XRT C++ error handling is via
exceptions and exec_buf indeed throws at a lower level thus never
returns error code.

The complication in this PR was primarily due to kds.cpp where cmds
are added to submitted (monitor) list before they are successfully
submitted via exec_buf.  If exec_buf throws, then these unsubmitted
commands must be removed again.  Unfortunately this seemingly reverse
order cannot be avoided (code comment explains why), but is kosher
because exec_buf throwing is really on the error path, e.g. not on a
critical success path.

CR: 1030607

### DIFF
--- a/src/runtime_src/driver/hw_em/generic_pcie_hal2/mbscheduler.cxx
+++ b/src/runtime_src/driver/hw_em/generic_pcie_hal2/mbscheduler.cxx
@@ -70,19 +70,19 @@ namespace xclhwemhal2 {
 
     for(unsigned i=0; i <MAX_SLOTS; ++i)
       submitted_cmds[i] = NULL;
-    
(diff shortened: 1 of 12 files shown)


stsoe:
Hi @chvamshi-xilinx the changes to hwemu are in this PR, I verified that CR 1030607 is now passing.

chvamshi-xilinx:
@stsoe..  I missed the file. Thanks for making changes. Changes are looking good

PR #1357, opened 2019-05-04
update port map for STM rev16.

karenx-xilinx:
STM is upgraded v15->v16. The portmap needs to be changed for STM v16. The platform is already submitted.


sonals:
ok to test

PR #1359, opened 2019-05-05
2019.1

chvamshi-xilinx:
Hi,

This change contains a fix in sw_emu in loading multiple xclbin.
calling resetProgram before loading new xclbin in sw_emu

Thanks


sonals:
ok to test

PR #1362, opened 2019-05-06
Disable CU trimming for MPSoC/Zynq

stsoe:


stsoe:
http://jira.xilinx.com/browse/CR-1030897

PR #1365, opened 2019-05-06
Cherry pick PR-1347 and use std:: list instead of std::queue in trace parser

hackwa:
* better approximation for incomplete cu trace

* add reset state


stsoe:
ok to test

hackwa:
Please do not merge yet. will update this PR to use std::list instead of std::queue

PR #1366, opened 2019-05-06
Add HAL log like feature into XDP (discussion)

tianhaoz95:
**Some discussion is still needed. Please do not merge just yet.**

# Feature Requested

When there is a CR on XDP, for example, `AWS` incompatibility issue, it is hard to locate the bug without inserting checkpoints or stepping through the code. There should be checkpoints that come with the code, but can be turned off with switches.

# Proposed Solution

Give `XDP` a `HAL` log like feature to enable quick locating of the bugs. The user can set a logging file in the `xrt.ini` file to enable the logging of `XDP` activities.

# Discussion

As @hackwa suggested, a global variable under `XDP` namespace will give access to all the `XDP` objects and limit the access outside `XDP`.


maxzhen:
Closing as no discussion has died down and it can't be merged due to unresolved conflicts.

PR #1376, opened 2019-05-07
Updated

uday610:
1) Elaborated and introduced XMA topic
2) Modified Annotated CU names topic. We dont need to use annotated CU names in all the cases, only in specific scenarios


sonals:
ok to merge

PR #1385, opened 2019-05-08
Resolve driver crashes on PPC64LE due to misaligned data structures

sonals:


maxzhen:
This PR fixed CR-1030262 and CR-1030514

PR #1388, opened 2019-05-08
Private Cloud Solution

rradjabi:
* Supports virtualization where mgmtpf is owned by host and userpf by guest VMs
* Uses TCP for passing messages between PFs (Software Mailbox)
* Requires configuration of /opt/xilinx/xrt/etc/msd-host.config to insert host ip, and host port for communication
* Works in a baremetal configuration (as sudo: msd; rmmod xocl; modprobe xocl; mpd) then xbutil validate will pass messages through software mailbox
* Uses a rudimentary identification method 0,1,2,3 that map to device ID on host. This will be updated to increase security.


maxzhen:
ok to test

rradjabi:
@xuhz 

> 1. in mailbox_daemon()
>    singleshot is not correct. It is possible there are IO on multiple FDs
In that case I'll need to call the routine to check and free the buffer after a read from both FDs. That's not a problem.

> 2. msd
>    line 119, the loop is not necessary, since parent will always block to wait for new connection.
I see what you mean. Thanks for catching that!

> 3. msd,
>    line 117, as discussed, the id should be limited as the index of the board. We can generate a long enough random number for each board and keep a mapping.
I will add this in the next PR.

> 4. msd
>    line 125. If id not match, some error handling needs to be taken. for example, send error msg to mpd, close the socket etc.
Same as response to (3), I'll add in the next PR.

> 5. parsing msg is missing.
>    code to be added include, 'dlopen' to load user provided plugin if any, init the hooks provided in the plugin, parse msg (saying if it is a fake xclbin) and call the hook function
Let's add this later, since we need more definition on msg parsing.

PR #1393, opened 2019-05-09
check for active devices in start counters and trace (2019.1)

hackwa:
fix program crash on profiling with p2p


maxzhen:
ok to test

PR #1400, opened 2019-05-09
Qdma 2019.1 update 

karenx-xilinx:
- CR 1029288: swap QDMA bar config and user bar location so they are the same as XDMA
- limit max queue pairs to 256 and move 2K ring size index to 0, thus allow co-existing of other qdma based drivers on the other PF (PF > 2)
- stm v17 update: added programming of h2c/c2h_mode registers based on the hint register.


maxzhen:
ok to test

PR #1401, opened 2019-05-10
Fix unexpected died KDS command monitor

larry9523:
When poll() is interrupted by SIGNAL, we should not treat it as an error. We should just return 0 so that we can restart the poll.

### DIFF
--- a/src/runtime_src/xrt/device/hal2.cpp
+++ b/src/runtime_src/xrt/device/hal2.cpp
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2017 Xilinx, Inc
+ * Copyright (C) 2016-2019 Xilinx, Inc
  *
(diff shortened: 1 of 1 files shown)


stsoe:
ok to test

src/runtime_src/xrt/device/hal2.cpp
```diff
@@ -459,7 +459,11 @@ exec_wait(int timeout_ms) const
 {
   auto retval = m_ops->mExecWait(m_handle,timeout_ms);
   if (retval==-1)
```
stsoe: Larry.  Add curly braces around outer if.   Compiler barfs without it  for standard warnings.

larry9523: fixed.

PR #1405, opened 2019-05-10
Different HP ports can connect to same memory bank

stsoe:
Don't treat all HP ports as separate memory banks.  Disable the CU
trimming from kernels.  Re-work in 2019.2 to determine when ports are
connected to same memory bank.

CR: 1030777
(cherry picked from commit f1b6fdd22336b60b9a8505da2f60609564126b7e)


maxzhen:
retest this please

maxzhen:
retest this please

PR #1406, opened 2019-05-10
The minimum to do power profiling

tianhaoz95:
**Feature Requested**

Do power profiling and show it together with other kernel and host activities.

**Implemented**

* Check if the power profiling feature has been turned on in the `xrt.ini` config file
* Start a thread polling the `sysfs` to get the power consumption reading for each device.
* Store the readings in a `vector`
* At the end of the application, write out the readings to a `csv` file

Implemented the polling and storing of the raw power status. Post processing will be another following PR.


maxzhen:
retest this please

maxzhen:
retest this please

src/runtime_src/xdp/profile/plugin/ocl/ocl_profiler.cpp
```diff
@@ -85,6 +85,13 @@ namespace xdp {
     if ((Plugin->getFlowMode() == xdp::RTUtil::HW_EM))
       xoclp::platform::start_device_trace(platform, XCL_PERF_MON_ACCEL, numComputeUnits);
 
+    if ((Plugin->getFlowMode() == xdp::RTUtil::DEVICE)) {
+      for (auto device : platform->get_device_range()) {
+        auto power_profile = std::make_unique<OclPowerProfile>(device->get_xrt_device(), Plugin, device->get_unique_name());
```
hackwa: @jvillarre @tianhaoz95  can we read power profiling data if a device isn't active or is a subdevice? should we read only the physical devices (root nodes) ?

tianhaoz95: That is a good point. We can read data back if the device is not active (not sure about sub-devices as they are not visible in PCIe platforms), but whether we should be reading from an inactive device is a different discussion. I am wondering is there any value letting the user know the complete power consumption status?

PR #1413, opened 2019-05-13
Extracted the aie pdi data from the xclbin 

venkatp36:
Extracted the aie pdi data from the xclbin  and stored the changes to physical runs dir which is further utilized by the sdx_emulation which lauches QEMU.


maxzhen:
ok to test

PR #1416, opened 2019-05-13
Fix unexpected died KDS command monitor

larry9523:
Cherry-pick this fix from 2019.1


larry9523:
not include the cstype fix. will create a new one

PR #1426, opened 2019-05-13
Fix for Extraneous Top Data transfers entry in Profile Summary with name "aws_0" 

IshitaGhosh:
CR-1031500
Host AXI MM data incorrectly shown in Top Kernel Data Transfer in Profile Summary in AWS Flow.
This was due to use of incorrect value for property of Monitor IPs. xclGetProfilingSlotProperties is now added to AwsXcl to retrieve the correct property value for Monitor IPs like SPM, SAM, SSPM.


maxzhen:
ok to test

PR #1429, opened 2019-05-14
adds mailbox pseudo-device creation support for multiple devices

rradjabi:
* Need to cherry-pick into 2019.1


maxzhen:
ok to test

PR #1433, opened 2019-05-15
Adding catch to XDP intitialization from OpenCL 

jvillarre:
In cases where OpenCL profiling/debug fails during initialization (such as a failure to open a file due to permissions issues), send a warning message and return without registering any callbacks.

### DIFF
--- a/src/runtime_src/xdp/profile/plugin/ocl/xocl_profile_cb.cpp
+++ b/src/runtime_src/xdp/profile/plugin/ocl/xocl_profile_cb.cpp
@@ -654,8 +654,16 @@ extern "C"
 void
 initXDPLib()
 {
-  (void)xdp::RTSingleton::Instance();
-  (void)xdp::OCLProfiler::Instance();
+  try {
+    (void)xdp::RTSingleton::Instance();
+    (void)xdp::OCLProfiler::Instance();
+  } catch (std::runtime_error& e) {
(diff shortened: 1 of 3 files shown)


src/runtime_src/xdp/profile/plugin/ocl/xocl_profile_cb.cpp
```diff
@@ -654,8 +654,16 @@ extern "C"
...
 {
-  (void)xdp::RTSingleton::Instance();
-  (void)xdp::OCLProfiler::Instance();
+  try {
+    (void)xdp::RTSingleton::Instance();
+    (void)xdp::OCLProfiler::Instance();
+  } catch (std::runtime_error& e) {
+    xrt::message::send(xrt::message::severity_level::XRT_WARNING, e.what());
+    // Don't register any of the callbacks.  Something went wrong during
+    //  initialization.
```
keryell: Strange indent.

src/runtime_src/xdp/profile/plugin/ocl/xocl_profile_cb.cpp
```diff
@@ -654,8 +654,16 @@ extern "C"
...
+    return ;
```
keryell: I am unsure the coding style require a space before `;` here

PR #1434, opened 2019-05-15
Add xocl_exec_reconfig to force scheduler reconfiguration

stsoe:


stsoe:
First step to force reconfiguration after re-clock and xclbin download.

PR #1443, opened 2019-05-17
Suppress benign messages if command returns successful (weak-modules)

rradjabi:
RHEL/CentOS 'weak-modules' feature produces benign 'Error' messages from depmod and dkms. Instead of printing these to stdout, we are capturing them in a variable and only displaying if the call returns non-zero. This avoids confusion of seeing 'Error' when it is actually benign. 

Addresses JIRA CR-1031286

### DIFF
--- a/src/CMake/config/postinst-aws.in
+++ b/src/CMake/config/postinst-aws.in
@@ -22,15 +22,9 @@ echo "Invoking xrt-aws common.postinst"
 #that prints benign 'Error' messages. These are not truly errors, so (for RHEL/CentOS)
 #we are capturing them in the 'res' variable, and only displaying them if the
 #command returns non-zero.
-if [ -f /etc/lsb-release ]; then #Ubuntu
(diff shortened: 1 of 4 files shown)


stsoe:
ok to test

rradjabi:
retest this please

maxzhen:
retest this please

keryell:
I am always scared when people try to hide stuff from me, specially when I have to use `sudo` to run *your* code on *my* computer and it might be a Trojan horse... :-)
Would it be safer to test for the existing installed modules and then just use the required commands to remove what has to be done?
If I do not understand correctly the problem, well, this code is not commented either and should not be open-sourced anyway... :-/
At least add a lot of comments explaining the issue and the solution.

rradjabi:
@keryell Thanks for your feedback. I updated the title and description, as well as added comments to the block of code. I hope that makes sense now.

keryell:
Thanks for the clarification in the script. Quite clearer now! :-)

rradjabi:
Struggling to find a command that works on CentOS and Ubuntu. Closing for now.

PR #1445, opened 2019-05-17
Bug Fix: Appending to a DTC section failed to add it back to the xclbin archive

rozumx:
Issue: When appending a DTC section, the section wasn't being added back to the section collection resulting. This resulted in the output xclbin archive not containing the changes.

Resolution: The DTC section is now added to the section collection.


stsoe:
ok to test

PR #1457, opened 2019-05-22
skd: Re-use file path from first cu

rathaval:
This fix re-uses the file path that was created when dealing with the first instance of the soft kernel and re-using it when num_cus is more than one.


rathaval:
@larry9523 @mamin506 

Please take a look at this fix and let me know if this can be merged into master ?
What about 2019.1 ?

maxzhen:
ok to test

larry9523:
> @larry9523 @mamin506
> 
> Please take a look at this fix and let me know if this can be merged into master ?
> What about 2019.1 ?

I think you can cherry-pick this fix to 2019.1

PR #1462, opened 2019-05-22
Build XRT on AL2

awshkulkar:
This has few tweaks to cmake to build XRT RPM on AL2. EPEL must be enables to install the pre-requisites on AL2.


sonals:
ok to test

awshkulkar:
> Are these same changes in other branches also?
What is the plan with merge? Are you planning to have this on master and 2018.3 or some more?

deeppat:
@sonals @stsoe Correct me if I'm wrong, but I believe this should also be added to master/2019.1 so we can have XRT building for AL2 moving forward.

stsoe:
@deeppat  Agreed, that's why I originally asked.   @awshkulkar can you create the corresponding PRs for this?

PR #1469, opened 2019-05-22
pyopencl-example

AShivangi:
A rough layout of pyopencl verify test case. This example currently doesn't work since it is not able to create compute program at L 72.


maxzhen:
Half done, closing for now.

PR #1476, opened 2019-05-23
XMA changes for DDR mapping to support HBM and xclbins with more kernels (2018.3)

hasansayani-xilinx:
1. Changing limits to handle more connections (eg. HBM, more kernels) & putting a place holder for MAX_REGS_PER_IP = 1, to take it into consideration in future
2. Changing a temporary struct variable to move from stack to heap to avoid causing stack smashing


hasansayani-xilinx:
@sarab96 / @sonals , please can we have it reviewed and pulled it?

maxzhen:
ok to test

maxzhen:
ok to test

PR #1491, opened 2019-05-29
Replace/remove obsoleted code related to XML meta data

stsoe:


hackwa:
profiling code looks good @jvillarre can take a look at the debug part.

PR #1493, opened 2019-05-30
skd: check for args being mapped successfully

rathaval:
getKernelArg() leads to a mapBO call which can fail as per the case with POSIX style
mmap() system calls. This commit adds a check for MAP_FAILED ((void *)(-1)).

### DIFF
--- a/src/runtime_src/core/edge/skd/sk_daemon.cpp
+++ b/src/runtime_src/core/edge/skd/sk_daemon.cpp
@@ -149,13 +149,19 @@ static void softKernelLoop(char *name, char *path, uint32_t cu_idx)
   }
 
   syslog(LOG_INFO, "%s_%d start running\n", name, cu_idx);
-    
+
   /* Set Kernel Ops */
   ops.getHostBO = &getHostBO;
   ops.mapBO     = &mapBO;
   ops.freeBO    = &freeBO;
 
   args_from_host = (unsigned *)getKernelArg(boh, cu_idx);
+  if (args_from_host == MAP_FAILED) {
+      syslog(LOG_ERR, "Failed to map soft kernel args for %s_%d", name, cu_idx);
+      freeBO(boh);
+      dlclose(sk_handle);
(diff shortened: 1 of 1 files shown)


rathaval:
@mamin506 @larry9523 

Hi Min and Larry, this is a small fix that checks for MAP_FAILED after the mmap() in getKernelArg().
We had talked about it several weeks ago, but I didn't get a chance to send this until now.

src/runtime_src/core/edge/skd/sk_daemon.cpp
```diff
@@ -149,13 +149,17 @@ static void softKernelLoop(char *name, char *path, uint32_t cu_idx)
...
+
   /* Set Kernel Ops */
   ops.getHostBO = &getHostBO;
   ops.mapBO     = &mapBO;
   ops.freeBO    = &freeBO;
 
   args_from_host = (unsigned *)getKernelArg(boh, cu_idx);
+  if (args_from_host == MAP_FAILED) {
+      syslog(LOG_ERR, "Failed to map soft kernel args for %s_%d", name, cu_idx);
+      return;
```
larry9523: You may want to free register BO and close sk_handle before bail out.

maxzhen:
ok to test

src/runtime_src/core/edge/skd/sk_daemon.cpp
rathaval: @larry9523 
I've updated the pull request with the error path handling.

larry9523: @rathaval I think it is better to keep the original code but just close and free before return. The reason for that is in destroySoftKernel()l, we will do munmap. Since here we are failing because we can not do mmap, it should not do munmap in this situation.

rathaval: Done. New patchset uploaded.

PR #1495, opened 2019-05-30
update mailbox doc for daemon interface

maxzhen:


maxzhen:
@arpitxilinx @sonals , after this PR, mailbox doc should be good enough for now for windows mailbox work.

PR #1504, opened 2019-05-31
Fatal Bug Fix - Fixed issue where the xclbin archive wasn't be formed correct.

rozumx:
This fix reverts a change made previously for the windows port.  This previous submission unfortunately causes the xrtbin image to be malformed.


stsoe:
ok to test

PR #1508, opened 2019-06-01
properly ref count pid

maxzhen:
@sonals, @houlz0507, I think we should use get/pud_pid() to properly ref count the struct pid before we cache the pointer to it.
Also, instead of caching group leader's pid, we probably should just cache the one for the current process.
Please let me know what you think.


maxzhen:
retest this please

maxzhen:
retest this please

maxzhen:
retest this please

maxzhen:
retest this please

PR #1512, opened 2019-06-04
skd: Change process name to match soft kernel name

Rohit-Athavale:
The commit tries to change the process name to match the soft kernel name.
This proves to be useful when trying to debug the performance of specific
soft kernels, unique names aids in tracking in these details.

top shows the process names by their kernel names, ps still shows skd.

### DIFF
--- a/src/runtime_src/core/edge/skd/sk_daemon.cpp
+++ b/src/runtime_src/core/edge/skd/sk_daemon.cpp
@@ -20,6 +20,7 @@
 #include <string.h>
 #include <sys/wait.h>
 #include <sys/mman.h>
+#include <sys/prctl.h>
 
 #include "sk_types.h"
 #include "sk_daemon.h"
@@ -126,7 +127,7 @@ static void softKernelLoop(char *name, char *path, uint32_t cu_idx)
   unsigned *args_from_host;
   unsigned int boh;
   int ret;
- 
+
   devHdl = initXRTHandle(0);
 
   ret = createSoftKernel(&boh, cu_idx);
@@ -273,6 +274,7 @@ static void sigAct(int sig)
   wait((int *)0);
 }
 
+#define PNAME_LEN	(16)
 void configSoftKernel(xclSKCmd *cmd)
 {
   pid_t pid;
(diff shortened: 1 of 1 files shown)


Rohit-Athavale:
@larry9523 @mamin506 

Hi Team,

This PR lets us view the kernel names via top, so we can notice abnormal behaviours. 
prctl() seemed more standard than pthread_setname_np(). Please take a look.

maxzhen:
ok to test

larry9523:
This looks good.  Thanks for making this change. 

A nit, can we move the process name setting prctl() right after we fork this new process (at line 306 in the new code)?

Just curious, I know top will show the new process name, what does ps -ef show? I did a simple test on x86, where ps -ef still showing the old process name unless I modify argv[0].

rathaval:
> 
> 
> This looks good. Thanks for making this change.
> 
> A nit, can we move the process name setting prctl() right after we fork this new process (at line 306 in the new code)?
> 
> Just curious, I know top will show the new process name, what does ps -ef show? I did a simple test on x86, where ps -ef still showing the old process name unless I modify argv[0].

Hi Larry, not sure if I should go change the argv[0] value, I believe its not portable, and I've also read online the the name can be only over-written, as in, if the original name is 4 characters long, then changing it to something longer than that is questionable.

prctl changes the names to "kernel_source0", "kernel_source1", ..., "kernel_sink0" etc in top which match the kernel names provided when downloading the soft kernels from x86. However, ps still shows skd as the process name.

Please comment on whats the best way to proceed ? Proceed with argv[0] change or just stick with prctl after fork() system call ? @larry9523

larry9523:
> > This looks good. Thanks for making this change.
> > A nit, can we move the process name setting prctl() right after we fork this new process (at line 306 in the new code)?
> > Just curious, I know top will show the new process name, what does ps -ef show? I did a simple test on x86, where ps -ef still showing the old process name unless I modify argv[0].
> 
> Hi Larry, not sure if I should go change the argv[0] value, I believe its not portable, and I've also read online the the name can be only over-written, as in, if the original name is 4 characters long, then changing it to something longer than that is questionable.
> 
> prctl changes the names to "kernel_source0", "kernel_source1", ..., "kernel_sink0" etc in top which match the kernel names provided when downloading the soft kernels from x86. However, ps still shows skd as the process name.
> 
> Please comment on whats the best way to proceed ? Proceed with argv[0] change or just stick with prctl after fork() system call ? @larry9523

@Rohit-Athavale , I am fine with the current change (just think move the prctl right after fork is better). At least we have top works better and can give us map between ps name and pid which can be used in other tools like ps.

rathaval:
@larry9523 Updated the PR with the suggested changes. Please take a look.

Unrelated question : Why is the build failing ? I'm able to build on petalinux side with 2018.3. 
Does the build trigger a petalinux compilation of the bitbake recipes ?

larry9523:
> @larry9523 Updated the PR with the suggested changes. Please take a look.
> 
> Unrelated question : Why is the build failing ? I'm able to build on petalinux side with 2018.3.
> Does the build trigger a petalinux compilation of the bitbake recipes ?

Don't worry about the Build failure. It is the pipeline issue and is being worked on.

maxzhen:
retest this please

PR #1514, opened 2019-06-04
adding interrupt sysfs debugging info for zocl driver

xdavidz:
While we were doing the debugging for interrupt bugs, we realized that the info of handling interrupts is the key info to narrow down the issue. It is better to have those code around when we have to debug it for some reason (as of today, we don't support data flow in zocl driver yet, in the future we will need this change). So,  just keep this code for debugging only, product code should not enable this. Hence, by default it is disabled.

In the future, we would enhance the build.sh to make zocl driver built automatically with debug/non-debug flags.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/sched_exec.c
+++ b/src/runtime_src/core/edge/drm/zocl/sched_exec.c
@@ -505,6 +505,10 @@ static irqreturn_t sched_exec_isr(int irq, void *arg)
 	isr = ioread32(virt_addr + 3);
 	iowrite32(isr, virt_addr + 3);
 
+#ifdef ZOCL_DEBUG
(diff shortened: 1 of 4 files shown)


maxzhen:
retest this please

xdavidz:
please do not merge for now, I have updates.

/David

On Jun 4, 2019, at 8:04 PM, Max Zhen <notifications@github.com<mailto:notifications@github.com>> wrote:


EXTERNAL EMAIL

retest this please

xdavidz:
The cmd submit has existing function ++cmd->exec->cu_usage[cmd->cu_idx], even though it is hard for debug when we have multiple CUs but better than nothing. Adding extra debug info seems over engineered for some reason, so that we hold this push until it becomes necessary.

PR #1520, opened 2019-06-04
2018_3 xmc ins

chienwei-lan:


maxzhen:
manually verified by @chienwei-lan

PR #1524, opened 2019-06-05
Windows Bug Fix 

rozumx:
+ Bug Fix: On Windows the xclbin archive was populated with invalidate data.  This bug was the result of some code refactoring that was necessary when we ported from Linux to Windows.  

Note: In a previous release, a gtest unit test was created to detect and report if the xclbin archive should happen to become malformed.  This test passes with these changes.


stsoe:
ok to test

PR #1531, opened 2019-06-06
update embedded platform build script

mamin506:
1. Add dsa_build.sh -- build DSA by dsa.tcl script and create dsa_build/ directory as workspace
2. Add peta_build.sh -- build petalinux project from a DSA file in <dsa_name>/ directory
3. Add pfm_build.sh -- build platform from a pfm.tcl script and create platform/ directory
4. rewrite petalinux.sh to build to use above scripts

This let us easier to build what we need instead of rebuild everything.

### DIFF
--- a/src/runtime_src/tools/scripts/peta_build.sh
+++ b/src/runtime_src/tools/scripts/peta_build.sh
@@ -1,0 +1,164 @@
+#! /bin/bash --
+#  (C) Copyright 2018-2019, Xilinx, Inc.
+
+# This script is to build an embedded system PetaLinux image
+# which contain XRT library and driver.
+#
+
+error()
+{
+	echo "ERROR: $1" 1>&2
+	usage_and_exit 1
+}
+
+usage()
+{
+	echo "Usage: $PROGRAM [options] /path/to/<dsa_name>.dsa"
+	echo "	options:"
+	echo "		--help,-v              print this usage"
+	echo ""
+	echo "This script will create <dsa_name>/ directory for petalinux project"
+	echo "Note: Do NOT move this script to another place. It depends on XRT repo"
+}
+
+usage_and_exit()
+{
+	usage
+	exit $1
+}
+
+SAVED_OPTIONS=$(set +o)
+
+# Don't print all commands
+set +x
+# Error on non-zero exit code, by default:
+set -e
+
+THIS_SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+PATH_TO_DSA=""
+DSA_NAME=""
+RECIPE_DIR=`readlink -f ${THIS_SCRIPT_DIR}/../../../platform/recipes-xrt/`
+
+while [ $# -gt 0 ]; do
+	case $1 in
+		--help | -h )
+			usage_and_exit 0
+			;;
+		--* | -* )
+			error "Unregognized option: $1"
+			;;
+		* )
+			PATH_TO_DSA=$1
+			break
+			;;
+	esac
+	shift
+done
+
+# Sanity Check
+[ ! -e "$PATH_TO_DSA" ] && error "DSA is not exist"
+
+[ -z "$PETALINUX" ] && error "PetaLinux is not installed"
+
+[ ! -d "$RECIPE_DIR" ] && error "Could not found recipe-xrt/ directory"
+# Sanity check done
+
+ORIGINAL_DIR=`pwd`
+
+DSA_FILE=`basename $PATH_TO_DSA`
+DSA_DIR=`dirname $PATH_TO_DSA`
+DSA_NAME=${DSA_FILE%.dsa}
+
+PETALINUX_NAME=$DSA_NAME
+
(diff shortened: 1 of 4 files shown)


xdavidz:
There scripts are very helpful to me. 3 small nits:
1) The dsa is being renamed to XSA or Shell or Platform. We can keep the current version unchanged and fix it once we see the new name from other dependencies.
2) petalinux-config --oldconfig seems obsoleted? should we use --silentconfig?
3) Do we have a README for the over all usage for those scripts? It seems that we should have a small instruction and recommend people start using the script. As long as we keep using those script, we can keep the script up-to-date and save time.

mamin506:
Hi @xdavidz ,
1. Yes, we should change DSA to XSA in 2019.2.
2. Let's use --silentconfig after 2019.2. Since this option is new in Petalinux 2019.1, it will not work on petalinux 2018.x.
3. This is a good point. Let's create a internal confluence page as the next step.

xdavidz:
> Hi @xdavidz ,
> 
> 1. Yes, we should change DSA to XSA in 2019.2.
> 2. Let's use --silentconfig after 2019.2. Since this option is new in Petalinux 2019.1, it will not work on petalinux 2018.x.
> 3. This is a good point. Let's create a internal confluence page as the next step.
Actually, what about a markdown text just beside those scripts? Something like:
https://github.com/Xilinx/XRT/blob/master/README.rst

https://github.com/Xilinx/XRT/tree/master/src/runtime_src/tools/scripts/README.rst or txt in case of being compiled? Could be very simple for scripts inside this directory.

src/runtime_src/tools/scripts/peta_build.sh
```diff
@@ -0,0 +1,164 @@
...
+echo 'IMAGE_INSTALL_append = " zocl"'               >> $PETALINUX_IMAGE_BBAPPEND
+echo 'IMAGE_INSTALL_append = " opencl-headers-dev"' >> $PETALINUX_IMAGE_BBAPPEND
+echo 'IMAGE_INSTALL_append = " opencl-clhpp-dev"'   >> $PETALINUX_IMAGE_BBAPPEND
+
+echo " * Adding XRT Kernel Node to Device Tree"
+#echo "******************* Device Tree hook ********************"
+echo "cat ${RECIPE_DIR}/../../runtime_src/core/edge/fragments/xlnk_dts_fragment_mpsoc.dts >> recipes-bsp/device-tree/files/system-user.dtsi"
+cat ${RECIPE_DIR}/../../runtime_src/core/edge/fragments/xlnk_dts_fragment_mpsoc.dts >> recipes-bsp/device-tree/files/system-user.dtsi
+echo "cat ${ORIGINAL_DIR}/dsa_build/zcu102_fragment.dts >> recipes-bsp/device-tree/files/system-user.dtsi"
+cat ${ORIGINAL_DIR}/dsa_build/zcu102_fragment.dts >> recipes-bsp/device-tree/files/system-user.dtsi
```
chvamshi-xilinx: Hi Min,

peta_build.sh contains following lines which are specific to zcu102.dts addtion. We should give some extra option to specify the file which contains all the dts paths.

example: --dts-file <file which contains all extra dts files>

mamin506: Good eye! I will fix this in next commit.

mamin506:
> > Hi @xdavidz ,
> > 
> > 1. Yes, we should change DSA to XSA in 2019.2.
> > 2. Let's use --silentconfig after 2019.2. Since this option is new in Petalinux 2019.1, it will not work on petalinux 2018.x.
> > 3. This is a good point. Let's create a internal confluence page as the next step.
> 
> Actually, what about a markdown text just beside those scripts? Something like:
> https://github.com/Xilinx/XRT/blob/master/README.rst
> 
> https://github.com/Xilinx/XRT/tree/master/src/runtime_src/tools/scripts/README.rst or txt in case of being compiled? Could be very simple for scripts inside this directory.


Good idea. I will do that in next commit.

src/runtime_src/tools/scripts/petalinux.sh
```diff
@@ -49,152 +53,39 @@ ORIGINAL_DIR=$PWD
...
-echo "petalinux-build"
-petalinux-build
+# dsa_build.sh will create ${PLATFORM_NAME}/ directory for petalinux project
+${XRT_REPO_DIR}/src/runtime_src/tools/scripts/peta_build.sh ${ORIGINAL_DIR}/dsa_build/${PLATFORM_NAME}.dsa
 
-cd $ORIGINAL_DIR
 echo " * Copying PetaLinux boot files (from: $PWD)"
-cp ./${PETALINUX_NAME}/images/linux/image.ub ${ORIGINAL_DIR}/dsa_build/src/a53/xrt/image/image.ub
+cd $ORIGINAL_DIR
+cp ./${PLATFORM_NAME}/images/linux/image.ub 	${ORIGINAL_DIR}/dsa_build/src/a53/xrt/image/image.ub
```
Himanshu-xilinx: is it ok to have src/a53/xrt path fixed? but this will make it specific for zynqmp

PR #1534, opened 2019-06-07
aarch64 warning on alignas

xdavidz:
When I was compiling the XRT on native ARM64 system, this was the only warning. It seems that some aarch64-g++ doesn't support the alignas c++11 api. Either we disable the warning or we don't use the alignas api on ARM64. I think disabling the warning may cause some potential performance issue.


keryell:
I am just scared that we need to care about pre-C++11 compilers here instead of focusing on the present...

PR #1536, opened 2019-06-07
msleep should not be used in write_lock automic context, use ndelay

xdavidz:
On ARM64 embedded system, we saw "BUG: scheduling while atomic: ..." errors during xclbin loading. The issue is that we are using msleep inside write_lock, which is an atomic context. Using ndelay is recommended base on linux doc: https://www.kernel.org/doc/Documentation/timers/timers-howto.txt


sonals:
retest this please

PR #1541, opened 2019-06-11
Refactor and Fix for clEnqueueCopyBufferRect/clEnqueueReadBufferRect

agozillon:
I don't think these OpenCL functions were working as intended so I have applied what I believe are fixes to them. 

In the case of clEnqueueReadBufferRect the function was not invoking setIfZero like it's mirror OpenCL call clEnqueueCopyBufferRect. This was against specification (https://www.khronos.org/registry/OpenCL/sdk/1.1/docs/man/xhtml/clEnqueueReadBufferRect.html sections on row pitch and slice pitch)  and also functionality breaking. It would cause data to be incorrectly mem copied into the host pointer from the device buffer when users passed zero values for certain arguments.

In the case of clEnqueueCopyBufferRect it wasn't possible to copy a full 2D buffer as the computation for the extent_in_bytes was incorrect (I took inspiration from POCL to tweak it: https://github.com/pocl/pocl/blob/master/lib/CL/pocl_util.c#L694 although it has an off by 1 error that I have a pull request in to fix at the moment, in the case of XRT I just remove the subtraction of 1 from the region[0] calculation in extent_in_bytes).

I also refactored a little bit while I was fixing things, hopefully that's OK.

I unfortunately do not have OpenCL tests to back this up, as these problems showed up as a problem in the TriSYCL/sycl runtime (I do have a test case in SYCL if anyone wishes to give it a spin with and without this change to XRT though...) and I do not believe XRT currently tests these functions, although I could be incorrect and just have not found them!

Thanks for considering the changes.

### DIFF
--- a/src/runtime_src/xocl/api/clEnqueueCopyBufferRect.cpp
+++ b/src/runtime_src/xocl/api/clEnqueueCopyBufferRect.cpp
@@ -60,9 +60,9 @@ src_extent_in_bytes(const size_t* region,
                     size_t        src_row_pitch, 
                     size_t        src_slice_pitch)
 {
-  return src_origin_in_bytes(src_origin,src_row_pitch,src_slice_pitch)  
-    + region[2] * src_slice_pitch
-    + region[1] * src_row_pitch
+  return src_origin_in_bytes(src_origin,src_row_pitch,src_slice_pitch)
+    + (region[2] - 1) * src_slice_pitch
+    + (region[1] - 1) * src_row_pitch
     + region[0];
 }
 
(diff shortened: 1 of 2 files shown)


src/runtime_src/xocl/api/clEnqueueCopyBufferRect.cpp
```diff
@@ -255,13 +231,13 @@ clEnqueueCopyBufferRect(cl_command_queue     command_queue ,
                         const cl_event *     event_wait_list ,
                         cl_event *           event_parameter)
 {
+  setIfZero(src_row_pitch,src_slice_pitch,dst_row_pitch,dst_slice_pitch,region);
```
keryell: Definitely a good idea to have this before the `validOrError`!

stsoe:
ok to test

PR #1543, opened 2019-06-11
enhance embedded system build scripts and add README.md

mamin506:
1. Enhance embedded build scripts to automatically build platform for all purposes
2. These scripts supports zynq DSA for 2019.2
3. The peta_build.sh could easily be configured to build a BOOT.BIN for Heterogeneous device
4. Add README.md for src/platform and src/runtime_src/tools/scripts directory


src/platform/README.md
```diff
@@ -0,0 +1,43 @@
...
+====================
+
+# mnt-sd/
+This directory is a recipe of the initial script.
+It would auto mount partition 1 of the SD card to /mnt and source init.sh script in /mnt if it exists.
+To use this recipe, copy it to a yocto or petalinux project. If you still don't know how to use it, please google yocto/petalinux and learn.
+
+# recipes-xrt/
+This directory including all of the necessary recipes to enable XRT on an embedded system.
+Copy the whole directory to your yocto or petalinux project. You could reference src/runtime\_src/tools/scripts/peta\_build.sh to learn how to use it.
```
keryell: All these `\` in this file does not seem the correct way to typeset these paths and other symbols...

keryell: Use backtick instead. Same for the other `.md` files.

mamin506: @keryell I didn't aware what should be the best way to typeset all paths. Use backtick would be a good idea. Changed in #1548.

PR #1547, opened 2019-06-12
PR-13389 SDx_OCL_Runtime: DSA to Shell / XSA Renaming

xdavidz:
Hi folks,

Could you please review this change for the effort of PR-13389 SDx_OCL_Runtime: DSA to Shell / XSA Renaming? Here is the simplified function specification:
http://confluence.xilinx.com/pages/viewpage.action?pageId=113004970

The way I changed the code is:
1) Existing code should not be broken. For example, we will try to find firmware from xsabin, if we cannot find it, we will try to find dsabin.
2) Internal code and script are unchanged.
3) Json interface have been tied to customer's application, so no change for now. Note: I am still asking the root PR owner what is the appropriate way to handle this.

I have compiled, installed, run "hello world". Please let me know if any other unit tests should be done.


src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
```diff
@@ -1372,8 +1412,7 @@ static int icap_download_boot_firmware(struct platform_device *pdev)
 	if (!ICAP_PRIVILEGED(icap))
 		return -EPERM;
 
-	/* Read dsabin from file system. */
-
+	/* Read xsabin first, if failed, try dsabin from file system. */
```
maxzhen: why not just push all these xsabin first and dsabin second logic into find_firmware()?

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -908,14 +909,27 @@ int runShellCmd(const std::string& cmd, std::string& output)
...
     const std::string& xclbin, std::string& output)
 {
-    std::string testCasePath = dsaPath +
+    std::string dirPath;
+    struct stat st;
+
+    output.clear();
+    
+    if (stat(xsaPath.c_str(), &st) != 0) {
+      if (stat(dsaPath.c_str(), &st) != 0) {
```
maxzhen: 4-space indent please

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
xdavidz: thanks, moved code into find_firmware

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
xdavidz: thanks, adjusted it. we still have some cpp code with 2 space indents, I will adjust my .vimrc config.

rozumx:
I will be reviewing these changes shortly....

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -908,14 +909,27 @@ int runShellCmd(const std::string& cmd, std::string& output)
 int xcldev::device::runTestCase(const std::string& exe,
     const std::string& xclbin, std::string& output)
 {
-    std::string testCasePath = dsaPath +
+    std::string dirPath;
+    struct stat st;
+
+    output.clear();
+    
+    if (stat(xsaPath.c_str(), &st) != 0) {
```
rozumx: While going forward in 2019.2, we will be releasing all of our deployment platforms in the /opt/xilinx/xsa directory.  For some customers, they may still have deployment packages from an earlier time.   

_This means that both the /opt/xilinx/dsa and /opt/xilinx/xsa will need to co-exist._  

This code seems to prevent examining the /opt/xilinx/dsa directories once a 2019.2 deployment platform (e.g., xsabin) is installed.  

Was that the intent?

src/runtime_src/core/pcie/driver/aws/kernel/mgmt/mgmt-bit.c
```diff
@@ -418,17 +429,20 @@ long load_boot_firmware(struct awsmgmt_dev *lro)
...
+	//try xsabin suffix first, if failed, then try dsabin suffix
+	create_fw_name(fw_name, sizeof(fw_name), lro, "xsabin");
 	err = request_firmware(&fw, fw_name, &lro->pci_dev->dev);
 
 	if (err) {
 		printk(KERN_WARNING "Unable to find firmware %s\n", fw_name);
-		return err;
+
+		create_fw_name(fw_name, sizeof(fw_name), lro, "dsabin");
+		printk(KERN_INFO "Try to find firmware %s\n", fw_name);
```
rozumx: Could the message be updated to be more "user" friendly indicating that the legacy fw image is being sought after.  For example:
`printk(KERN_WARNING "Searching for legacy firmware image %s\n", fw_name);`

This might be important for legacy platforms will always produce a the warning on line 437 stating that the firmware image could not be found.

xdavidz: Thanks! Added legacy info into the warning message. Please see the most recent commit.

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
xdavidz: Talked with Lizhi, this is for xbutil validate, which will try to load test xclbin from /opt/xilinx/xsa or /opt/xilinx/dsa. Once user installed /opt/xilinx/xsa newer version, the tests in the xsa directory should be valid. Users don't have the ability to chose default/selected directory by using xbutil.

xdavidz: Chatted with Steve, more thoughts needed here because we might have both dsa and xsa installed, but the xbutil will search for /opt/xilinx/xxx/firmware name/test, if the firware name is in dsa rather than xsa, we will fail to find tests. Will fix this.

xdavidz: fixed

xdavidz:
Request re-review for xbutil changes

PR #1562, opened 2019-06-18
Support added to profiling to create a run_summary file

rozumx:
The run_summary file has a file index to the profile and timing
files on disk and contains the system profile metadata.
Work Done
---------
1) When the RTProfile class destructor is called, the run_summary file is produced.
2) Added a RunSummary class that is used to store file and metadata that is later marshed to disk.
   + Extracts the System Diagram metadata from the in memory xclbin archive image
   + Encodes the System Diagram metadata as a hex representation when storing to disk
   + Stores the contents of the run_summary file produced on disk in the JSON format
3) Update RTProfile to create the RunSummary class and to add the files and xclbin image to the RunSummary class.
4) Updated the trace and profile classes with a method that exposes the filename produced on disk.
5) Fixed some spelling mistakes

### DIFF
--- a/src/runtime_src/tools/xclbin/XclBinClass.cxx
+++ b/src/runtime_src/tools/xclbin/XclBinClass.cxx
@@ -120,7 +120,7 @@ XclBin::readXclBinBinaryHeader(std::fstream& _istream) {
   _istream.read((char*)&m_xclBinHeader, sizeof(axlf));
 
   if (_istream.gcount() != expectBufferSize) {
-    std::string errMsg = "ERROR: Input stream is smaller then the expected header size.";
+    std::string errMsg = "ERROR: Input stream is smaller than the expected header size.";
     throw std::runtime_error(errMsg);
   }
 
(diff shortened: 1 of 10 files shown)


rozumx:
Hello Paul,
Thanks for the code review earlier.  I have incorporated your feedback into this pull request and have run it on board which has produced the expected output.

maxzhen:
ok to test

rozumx:
This warning/error is valid...working on addressing it.

maxzhen:
retest this please

pgschuey:
@jvillarre - Are you okay with boost being used here?

rozumx:
Hello Paul,
As requested, I have changed the enumeration names.

rozumx:
Jason with regards to the boost functionality.   
Due to the nature of the changes, I did include only in the run_summary.cpp file inclusion of two boost templates for property tree manipulations.  

**Note**

1. Doing so had zero impact to the cmake build scripts or dependency on any link boost library.
2. I did look into not using the property tree and producing the JSON directly from the class, unfortunately this code became rather difficult to work with.
3. If needed, I can refactor the code to not use the boost libraries, but doing so will most likely result in higher maintenance in the code base, especially when we go to the Windows XRT platform.  The issue here has to with with the file paths values being stored in the JSON syntax.  Boost "automatically" takes care of this format.

PR #1572, opened 2019-06-24
Xbutil for Edge

Himanshu-xilinx:
* Initial xbutil infra for Edge
* Implemented query command


src/runtime_src/core/common/CMakeLists.txt
```diff
@@ -16,6 +16,9 @@ file(GLOB XRT_CORECOMMON_LIB_FILES
 # Files to include in object list
 file(GLOB XRT_CORECOMMON_OBJ_FILES
   "scheduler.*"
+  "dd.*"
```
stsoe: What's the reason dd sensor and utils files were added to this list?  Do they themselves depend on the coreutils library?   

Unfortunately dd* of can't be under core/common  because its linux specific, e.g. won't compile on windows.   It should be rewritten to free form command line parsing or rewritten to use boost::program_options.   If you can do this great, if not, then you have to copy dd* to edge.

sensor and utils should be in the corecommon_lib_files list.

Himanshu-xilinx: 1. No, sensor and utils dont depend on coreutils library. Moved sensor and utils in the corecommon_lib_files list.
2. Actually we are not implementing dd in this phase. So for now we don't need to move dd to common. But point taken, i will implement dd with boost::program_options.

src/runtime_src/core/edge/tools/xbutil/base.h
```diff
@@ -0,0 +1,131 @@
...
+#include <string>
+#include <fstream>
+#include <chrono>
+#include <ctime>
+
+
+
+namespace xcldev {
+
+static std::string driver_version(std::string driver)
```
stsoe: const std::string&

maxzhen:
ok to test

PR #1577, opened 2019-06-25
Changing the permissions of the NIFD character device driver so that only root can open it

jvillarre:


maxzhen:
retest this please

PR #1579, opened 2019-06-25
[Edge] Remove prebuild/, which is redundant in Scout

mamin506:
Siva mentioned that prebuild/ is redundant in Scout. We could remove it and all of the new platform would not need to have that directory.


src/platform/zcu102ng/config.sh
```diff
@@ -20,12 +20,15 @@ TEMPLATE=zynqMP
...
-#  config_peta <petalinux_project>/project-pec/configs/config
+#  config_peta <petalinux_project>/project-spec/configs/config
 #
 #config_peta()
 #{
 #	PETA_CONFIG_FILE=$1
+#   Enable debug tweaks would avoid enter user name and password at boot time
 #	echo "CONFIG_YOCTO_ENABLE_DEBUG_TWEAKS=y" >> $PETA_CONFIG_FILE
+#   Specify TMP directory. This will be useful when the petalinux project is on NFS
+#   echo "CONFIG_TMP_DIR_LOCATION="/scratch/tmp" >> $PETA_CONFIG_FILE
```
maxzhen: missing a '"'?

mamin506: Should be '.  Fix at hand.

PR #1582, opened 2019-06-26
Recognize DRM Buffer object and CU in mmap

Himanshu-xilinx:
* Fixed bug in zocl driver. It was not able to recognize DRM BO and CU if CU address greater than ZOCL_FILE_PAGE_OFFSET
* Added DRM_IOCTL_ZOCL_INFO_CU ioctl


src/runtime_src/core/edge/drm/zocl/zocl_drv.c
```diff
@@ -371,12 +372,12 @@ static int zocl_mmap(struct file *filp, struct vm_area_struct *vma)
 		return -EINVAL;
 	}
 
-	phy_addr = vma->vm_pgoff << PAGE_SHIFT;
-
 	/* Only allow user to map register ranges in apertures list.
 	 * Could not map from the middle of an aperture.
 	 */
-	apt_idx = get_apt_index(zdev, phy_addr);
+	apt_idx = vma->vm_pgoff;
```
mamin506: Check if apt_idx is valid.

if (apt_idx >= zdev->num_apts)
        // Bad apt_idx, map failed.

src/runtime_src/core/edge/drm/zocl/zocl_ioctl.c
```diff
@@ -562,3 +563,21 @@ zocl_read_axlf_ioctl(struct drm_device *dev, void *data, struct drm_file *filp)
...
+	struct drm_zocl_info_cu *args = data;
+	struct drm_zocl_dev *zdev = dev->dev_private;
+	struct sched_exec_core *exec = zdev->exec;
+
+	if (!exec->configured) {
+		DRM_ERROR("Schduler is not configured\n");
+		return -EINVAL;
+	}
+ 
+	if(args->paddr != 0xFFFFFFFF) {
```
larry9523: Can you please explain why do you check args->paddr != 0xFFFFFFFF?

src/runtime_src/core/edge/drm/zocl/zocl_ioctl.c
```diff
@@ -562,3 +563,21 @@ zocl_read_axlf_ioctl(struct drm_device *dev, void *data, struct drm_file *filp)
...
+		args->apt_idx = get_apt_index(zdev, args->paddr);
+	}
```
larry9523: if args->paddr does equal to 0xFFFFFFFF, this function will return 0 and apt_idx is not defined. Is it what expected here?
I think the better way is check the return value of get_apt_index, if it is -EINVAL, just return -EINVAL of your ioctl function. So that the shim.cpp code (ioctl caller) does not need to check if the apt_index is -EINVAL. And if we have other failuer case in kernel, just return error in ioctl handler and ioctl caller does not need to add extra check

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -153,7 +153,18 @@ int ZYNQShim::mapKernelControl(const std::vector<std::pair<uint64_t, size_t>>& o
...
+        int result = ioctl(mKernelFD, DRM_IOCTL_ZOCL_INFO_CU, &info);
+        if(result) {
+            printf("failed to find CU info 0x%lx\n", offset_it->first);
+            return -1;
+        }
+        size_t psize = getpagesize();
+        if(info.apt_idx == -EINVAL) {
+            printf("Failed to find CU in aperture list 0x%lx\n", offset_it->first);
+            return -1;
+        }
```
larry9523: As commented, this seems not necessary. just fail the ioctl if we can not find the CU physical address in kernel

larry9523:
Overall it looks good. I would suggest adding some comments explaining why we need an extra ioctl to get CU index instead of just using CU physical address. Either shim.cpp or zocl_mmap could be a good place

sonals:
Can you share the testcase which is running into this. BO fake offsets start from 4 GB pages (4K) which is 16 TB offset. Do our AXI-Lite address space offsets that high?

larry9523:
> Can you share the testcase which is running into this? BO fake offsets start from 4 GB pages (4K) which is 16 TB offset. So our kernel AXI-Lite address can go as high as 16 TB. Is this not sufficient for some testcase?

I think this is for 32bit Zynq device, where BO fake offset is 0x1_0000 page (4K). So the offset is 0x1000_0000. CU starting from 0xA000_0000 overlap with BO.

Himanshu-xilinx:
> Can you share the testcase which is running into this? BO fake offsets start from 4 GB pages (4K) which is 16 TB offset. So our kernel AXI-Lite address can go as high as 16 TB. Is this not sufficient for some testcase?

On Zynq7000, PL has two address ranges 0x4000_0000 to 0x7FFF_FFFF and 0x8000_0000 to 0xBFFF_FFFF. Both are greater than DRM_FILE_PAGE_OFFSET_START(0x1000_0000). In this case, mmap() is not able to recognize DRM buffer object and CU.

On Zynq UltraScale+, PL has 3 address ranges which are lesser than DRM_FILE_PAGE_OFFSET_START(0x1_0000_0000). There are 4 address ranges greater than DRM_FILE_PAGE_OFFSET_START. Scout compiler would place CUs at the lower ranges. In this case, mmap() is able to recognize buffer object and CU.

src/runtime_src/core/edge/user/shim.cpp
Himanshu-xilinx: removed

src/runtime_src/core/edge/drm/zocl/zocl_ioctl.c
Himanshu-xilinx: code updated

src/runtime_src/core/edge/drm/zocl/zocl_ioctl.c
Himanshu-xilinx: actually not needed. 
removed

src/runtime_src/core/edge/drm/zocl/zocl_drv.c
Himanshu-xilinx: updated code

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -153,7 +153,14 @@ int ZYNQShim::mapKernelControl(const std::vector<std::pair<uint64_t, size_t>>& o
     if ((offset_it->first & (~0xFF)) != (-1UL & ~0xFF)) {
       auto it = mKernelControl.find(offset_it->first);
       if (it == mKernelControl.end()) {
-        ptr = mmap(0, offset_it->second, PROT_READ | PROT_WRITE, MAP_SHARED, mKernelFD, offset_it->first);
+        drm_zocl_info_cu info = {offset_it->first, -1};
+        int result = ioctl(mKernelFD, DRM_IOCTL_ZOCL_INFO_CU, &info);
+        if(result) {
```
mamin506: Should use space after 'if'.
"        if (result) {"

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -153,7 +153,14 @@ int ZYNQShim::mapKernelControl(const std::vector<std::pair<uint64_t, size_t>>& o
...
+            printf("failed to find CU info 0x%lx\n", offset_it->first);
+            return -1;
+        }
+        size_t psize = getpagesize();
+        ptr = mmap(0, offset_it->second, PROT_READ | PROT_WRITE, MAP_SHARED, mKernelFD, info.apt_idx*psize);
         if (ptr == MAP_FAILED){
```
mamin506: "        if (ptr == MAP_FAILED) {". Space before '{'.

Himanshu-xilinx: done

src/runtime_src/core/edge/user/shim.cpp
Himanshu-xilinx: done

src/runtime_src/core/edge/drm/zocl/zocl_ioctl.c
```diff
@@ -562,3 +563,26 @@ zocl_read_axlf_ioctl(struct drm_device *dev, void *data, struct drm_file *filp)
...
+	struct drm_zocl_dev *zdev = dev->dev_private;
+	struct sched_exec_core *exec = zdev->exec;
+
+	if (!exec->configured) {
+		DRM_ERROR("Schduler is not configured\n");
+		return -EINVAL;
+	}
+ 
+	args->apt_idx = get_apt_index(zdev, args->paddr);
+	if(args->apt_idx == -EINVAL) {
```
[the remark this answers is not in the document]
Himanshu-xilinx: done

src/runtime_src/core/edge/user/shim.cpp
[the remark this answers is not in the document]
Himanshu-xilinx: done

maxzhen:
ok to test

PR #1584, opened 2019-06-26
Device Memory Only Buffer

chienwei-lan:
1. Remove xclBOKind from emu, and hw hal2 interface
2. Unifying BO flag


src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_bo.h
```diff
@@ -88,14 +92,40 @@ static inline struct drm_xocl_dev *bo_xocl_dev(const struct drm_xocl_bo *bo)
 	return bo->base.dev->dev_private;
 }
 
-static inline unsigned xocl_bo_ddr_idx(unsigned flags)
+static inline unsigned xocl_bo_ddr_idx(unsigned user_flags)
+{
+        return user_flags & 0xFFFFFF;
```
maxzhen: please define a macro for user flag mask used for retrieving ddr idx and use ~macro to fetch bo type. make sure this macro is also shared with user land code when user flag is created.

src/runtime_src/xrt/device/hal2.h
```diff
@@ -76,7 +76,7 @@ class device : public xrt::hal::device
     void *hostAddr = nullptr;
     size_t size = 0;
     size_t offset = 0;
-    xclBOKind kind;
+    int kind;
```
stsoe: Can we remove kind from this internal struct?  It doesn't look like kind is used any where.

PR #1586, opened 2019-06-26
Minor changes. Changed to xrt.h, added enum type, etc

sarab96:
Minor changes. Changed to xrt.h, added enum type for gcc compatibiliy and removed xma internal header files from install.
@max Please do the needful


maxzhen:
ok to test

PR #1587, opened 2019-06-27
Create XMA singleton on stack rather than on heap.

sarab96:
Create XMA singleton on stack rather than on heap.
@max Please do the needful


maxzhen:
ok to test

sarab96:
retest this please

PR #1616, opened 2019-07-04
Fixed for CR-1034151: added missing func in sw_emu.

vboggara-xilinx:
Added functions xclOpenContext, xclExecWait, xclExecBuf and xclCloseContext  for cpu emulation.


maxzhen:
ok to test

PR #1618, opened 2019-07-06
xclCopyBO() now leverages newly added CMD BO caching framework to improve M2M throughput.

sonals:
Only tested on PCIe platform.

### DIFF
--- a/src/runtime_src/core/common/bo_cache.h
+++ b/src/runtime_src/core/common/bo_cache.h
@@ -1,0 +1,79 @@
+/**
+ * Copyright (C) 2019 Xilinx, Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may
+ * not use this file except in compliance with the License. A copy of the
+ * License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef core_common_bo_cache_h_
+#define core_common_bo_cache_h_
+
+#include <stack>
+#include <utility>
+
+#include "xrt.h"
+#include "ert.h"
+
+namespace xrt_core {
+    typedef std::pair<const unsigned, struct ert_start_copybo_cmd * const> cmd_bo;
+    class bo_cache {
+        xclDeviceHandle device_handle;
+        unsigned cmd_bo_cache_max_size;
+        std::stack<cmd_bo> cmd_bo_cache;
+        std::mutex cmd_bo_cache_lock;
+
+    public:
+        bo_cache(xclDeviceHandle handle, unsigned max_size) : device_handle(handle),
+                                                              cmd_bo_cache_max_size(max_size) {}
+        ~bo_cache() {
+            std::lock_guard<std::mutex> lock(cmd_bo_cache_lock);
+            while (!cmd_bo_cache.empty()) {
+                cmd_bo bo = cmd_bo_cache.top();
+                destroy(bo);
+                cmd_bo_cache.pop();
+            }
+        }
+
+        cmd_bo alloc() {
+            {
+                std::lock_guard<std::mutex> lock(cmd_bo_cache_lock);
+                if (!cmd_bo_cache.empty()) {
+                    cmd_bo bo = cmd_bo_cache.top();
+                    cmd_bo_cache.pop();
+                    return bo;
+                }
+            }
+            unsigned execHandle = xclAllocBO(device_handle, sizeof(ert_start_copybo_cmd),
+                                             0, XCL_BO_FLAGS_EXECBUF);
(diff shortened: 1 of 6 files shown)


xdavidz:
> Only tested on PCIe platform.

Looks good to me. 

If you would like to test on zynqu+, I have test code here
https://github.com/xdavidz/temp/tree/master/built-in-dma.  Please see: https://github.com/xdavidz/temp/blob/master/built-in-dma/copy_bo_main.cpp :: ertCopyBO

src/runtime_src/core/common/bo_cache.h
```diff
@@ -0,0 +1,90 @@
...
+#include <stack>
+#include <utility>
+
+#include "xrt.h"
+#include "ert.h"
+
+namespace xrt_core {
+    // Helper typedef for std::pair. Note the elements are const so that the
+    // pair is immutable. The clients should not change the contents of cmd_bo.
+    typedef std::pair<const unsigned, struct ert_start_copybo_cmd * const> cmd_bo;
```
maxzhen: this seems to be a generic mechanism for cmd bo caching, but the pointer is specific to copybo cmd?

sonals: I indeed want to go there. However ERT cmds currently do not have the concept of base struct. Will work out the change later. Another option is to templatize the implementation with ert cmd type.

maxzhen: struct ert_start_copybo_cmd is a fixed size ert cmd type. but size of other ert cmd type may not be fixed. so, it sounds like you can't use ert cmd type to templatize the implementation...
we may be able to implement a generic bo_cache to alloc and cache a bo of a certain size. it's up to user to cast the mapped bo pointer to the ert cmd they want as long as the bo size is big enough to hold the cmd data.

stsoe: The cache does not need to cache specific ert_cmd types.   In the upper level code and cmd bo is a struct that wraps the device handle and the mapped allocated BO.  Since a command BO is really just memory, a free BO can be used for any command type, which is what the upper level code does.  

Couple of general C++ comments.   
- Never use just unsigned, please specify the type.   This code will not compile on Windows
- typedefs in C++ are a thing of the past, please use 'using' which is more generic.

src/runtime_src/core/common/bo_cache.h
```diff
@@ -0,0 +1,90 @@
...
+
+    // Create a cache of CMD BO objects -- for now only used for M2M -- to reduce
+    // the overhead of BO cycle management.
+    class bo_cache {
+        xclDeviceHandle device_handle;
+        // Maximum number of BOs that can be cached in the pool
+        unsigned cmd_bo_cache_max_size;
+        std::stack<cmd_bo> cmd_bo_cache;
```
stsoe: std::vector will do just fine, no need for stack.   The default container for stack is deque, which is probably not what you want here.

src/runtime_src/core/common/bo_cache.h
```diff
@@ -0,0 +1,90 @@
...
+        std::mutex cmd_bo_cache_lock;
+
+    public:
+        bo_cache(xclDeviceHandle handle, unsigned max_size) : device_handle(handle),
```
stsoe: unsigned int

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -117,12 +117,14 @@ ZYNQShim::ZYNQShim(unsigned index, const char *logfileName, xclVerbosityLevel ve
     mLogStream << "FUNCTION, THREAD ID, ARG..." << std::endl;
     mLogStream << __func__ << ", " << std::this_thread::get_id() << std::endl;
   }
+  mCmdBOCache = new xrt_core::bo_cache(this, 4);
```
stsoe: Better to save help allocated objects in unique ptr.

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -117,12 +117,14 @@ ZYNQShim::ZYNQShim(unsigned index, const char *logfileName, xclVerbosityLevel ve
...
 }
 
 #ifndef __HWEM__
 ZYNQShim::~ZYNQShim()
 {
-  if (profiling != nullptr) delete profiling;
+  delete mCmdBOCache;
```
stsoe: No need for explicit delete if std::unique_ptr is used.

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -371,6 +373,37 @@ int ZYNQShim::xclSyncBO(unsigned int boHandle, xclBOSyncDirection dir, size_t si
...
+{
+    int ret = -EOPNOTSUPP;
+#ifdef __aarch64__
+    xrt_core::cmd_bo bo = mCmdBOCache->alloc();
+    ert_fill_copybo_cmd(bo.second, src_boHandle, dst_boHandle,
+                        src_offset, dst_offset, size);
+
+    ret = xclExecBuf(bo.first);
+    if (ret) {
+        mCmdBOCache->release(bo);
```
stsoe: I don't think we should expose the concept of a command bo cache inside functions that use / allocate command BOs.  The cache is a implementation detail to hide within existing xclAlloc/Free APis.

sonals:
Superseded by https://github.com/Xilinx/XRT/pull/1655

PR #1625, opened 2019-07-09
use dma_addr_t instead of uint64_t so its compatible with both 32bit and 64 bit system

Himanshu-xilinx:
use dma_addr_t instead of uint64_t so its compatible with both 32bit and 64 bit system

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
@@ -44,7 +44,7 @@ void zocl_describe(const struct drm_zocl_bo *obj)
 }
 
 static inline void
-zocl_bo_describe(const struct drm_zocl_bo *bo, uint64_t *size, uint64_t *paddr)
+zocl_bo_describe(const struct drm_zocl_bo *bo, uint64_t *size, dma_addr_t *paddr)
 {
 	if (bo->flags & (ZOCL_BO_FLAGS_CMA | ZOCL_BO_FLAGS_USERPTR)) {
 		*size = bo->cma_base.base.size;
--- a/src/runtime_src/core/edge/include/zynq_ioctl.h
+++ b/src/runtime_src/core/edge/include/zynq_ioctl.h
@@ -27,6 +27,12 @@
 #include <uapi/drm/drm_mode.h>
 #endif /* !__KERNEL__ */
 
+#if defined(CONFIG_ARM64)
+typedef uint64_t addr_t;
+#else
+typedef uint32_t addr_t;
+#endif
+
 //#define XCLBIN_DOWNLOAD
 
 enum {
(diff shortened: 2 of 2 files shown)


src/runtime_src/core/edge/include/zynq_ioctl.h
```diff
@@ -102,7 +108,7 @@ struct drm_zocl_sync_bo {
 struct drm_zocl_info_bo {
 	uint32_t	handle;
 	uint64_t	size;
-	uint64_t	paddr;
+	addr_t	paddr;
 };
```
sonals: 1. Please keep the same indentation
2. Userspace interface could remain 64-bit? For example the size above is still 64 bit and all other ioctls like user_ptr pwrite, etc. expect 64 bit physical address type as well. 32-bit application should still use 64-bit data types for the ioctls.

Himanshu-xilinx: 2.  if userspace interface remains 64-bit, it give build error 
 zocl_bo.c:651:36: error: passing argument 3 of 'zocl_bo_describe' from incompatible pointer type [-Werror=incompatible-pointer-types]
|   zocl_bo_describe(bo, &args->size, &args->paddr);

where args->paddr is uint64_t

mamin506: Change the interface of user space and kernel space is danger. We should only fix this issue in kernel space.

xdavidz: Let's keep the interface unchanged. I am testing a new proposal, will send you update today.

xdavidz:
> use dma_addr_t instead of uint64_t so its compatible with both 32bit and 64 bit system
What about the following changes (compiled on zcu104 and zc702).
https://github.com/xdavidz/XRT1/commit/f3e84f73d70e02f70ee205fb8607abc9be5b7b7e

It seems a little bit safer to not change the interface, thus we always use uint64 for those addresses (note the mm_node->start is u64, casting to u32 is dangerous). Then, we convert those only when we want to use them as dma address and size. Because those address and size info are passed back from the kernel, it should be safe to cast them from (driver)32 to (user)64, then (user)64 to (driver)32.

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -44,7 +44,7 @@ void zocl_describe(const struct drm_zocl_bo *obj)
 }
 
 static inline void
-zocl_bo_describe(const struct drm_zocl_bo *bo, uint64_t *size, uint64_t *paddr)
```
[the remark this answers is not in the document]
chvamshi-xilinx: @xdavidz, Your new proposal looks good and safe. **Please send a pull request with these changes.** 
We have verified your changes for both zc702 and zcu102 platforms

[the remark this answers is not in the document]
maxzhen: So, we should close this PR and wait for David's?

maxzhen:
After integration of PR#1643, we should not need this PR. Closing it now.

PR #1627, opened 2019-07-09
Updated the packaging script to allow packaging future platforms.

rozumx:
Work Done
---------
1) Added a new option to point to other XRT images
2) Updated the packaging flow to recongize the xilinx_u200_xdma_201820_2 and xilinx_u250_xdma_201820_2 platforms and if found to package the scheduler FW image from alternate XRT installations.
3) Removed an unneeded echo command


rozumx:
OK to test

PR #1630, opened 2019-07-10
cpp changes and hal_probe updated for new hwcfg without yaml stuff

sarab96:
cpp changes and hal_probe updated for new hwcfg without yaml stuff

### DIFF
--- a/src/xma/include/lib/xmaapi.h
+++ b/src/xma/include/lib/xmaapi.h
@@ -30,6 +30,7 @@ typedef struct XmaSingleton
 {
     //XmaSystemCfg      systemcfg;
     XmaHwCfg          hwcfg;
+    bool              xma_initialized;
     //XmaConnect        connections[MAX_CONNECTION_ENTRIES];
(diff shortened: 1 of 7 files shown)


stsoe:
Why are the atomic data members allocated on the heap?
I suspect its related to missing deletes of copy ctor and assignment ctor.

sarab96:
yes Soren that is correct. atomic do not allow copy or move.

stsoe:
But this should be changed.   Allocating these on the heap makes no sense.
Please disable the copy and assignment ctor.
Definitely atomics can be moved so make sure move ctor/assignment is defined, but that corresponding copies are deleted.

stsoe:
Ok, so by design std::atomic cannot be moved.  Alas objects with std::atomic data members cannot be inserted into a vector if the vector will resize itself when new objects are emplace constructed.  Given this. I guess it makes sense that the std::atomic data member is on the heap.  Case rested :-)

PR #1634, opened 2019-07-10
remove ert config cmd from python testcases

AShivangi:


tests/python/22_verify/main.py
```diff
@@ -130,9 +105,8 @@ def runKernel(opt):
         print("Now wait until the kernel finish")
 
     print("Wait until the command finish")
-    while ecmd.m_uert.m_cmd_struct.state < ert_cmd_state.ERT_CMD_STATE_COMPLETED:
-        while xclExecWait(opt.handle, 100) == 0: 
-            print(".")
+    while xclExecWait(opt.handle, 100) == 0:
```
maxzhen: the previous code looks right. and the new version looks wrong. why make this change?

AShivangi: ecmd was configure command. That while loop was checking for the state of the configure command

PR #1642, opened 2019-07-11
Bug Fix: Incorrect bitmask was being used

rozumx:
Issue
-----
The bit mask mProfileFlags instead of mFileFlags was being use to determine what profiling output files to record in the run_summary file.


rozumx:
OK to test

PR #1647, opened 2019-07-11
XDP rearchitecture to avoid calling HW SHIM methods for profiling 

IshitaGhosh:
1. Add Profile Monitor IP classes with specialized start, read, stop functionality for accessing registers from the IP HW 
2. Add Device Interface objects to OCLProfiler which reads debug_ip_layout and creates ProfileIP objects


IshitaGhosh:
I am debugging the failures.

src/runtime_src/core/include/xclbin.h
```diff
@@ -307,7 +307,9 @@ extern "C" {
         AXI_MONITOR_FIFO_FULL,
         ACCEL_MONITOR,
         AXI_STREAM_MONITOR,
-	AXI_STREAM_PROTOCOL_CHECKER
+	    AXI_STREAM_PROTOCOL_CHECKER,
```
sonals: Indentation?

src/runtime_src/xdp/profile/device/aim.cpp
```diff
@@ -0,0 +1,246 @@
+/**
+ * Copyright (C) 2016-2018 Xilinx, Inc
```
sonals: License year 2019

src/runtime_src/xdp/profile/device/aim.h
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright (C) 2019-2020, Xilinx Inc - All rights reserved
```
sonals: 2020 is in future so just 2019.

src/runtime_src/xdp/profile/device/am.cpp
```diff
@@ -0,0 +1,275 @@
+/**
+ * Copyright (C) 2016-2018 Xilinx, Inc
```
sonals: For new files it should be just 2019. If older file then change 2018 to 2019.

src/runtime_src/xdp/profile/device/am.cpp
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+}
+
+bool AM::has64bit()
+{
+    return ((properties & XAM_64BIT_PROPERTY_MASK) ? true : false);
+}
+
+bool AM::hasDataflow()
```
sonals: const function

src/runtime_src/xrt/device/device.h
```diff
@@ -786,6 +786,24 @@ class device : public xrt_device
     return m_hal->readTrace(type,vec);
   }
 
+  hal::operations_result<void>
```
stsoe: I am not sure I fancy profiling relying on xrt::device thereby adding more methods.
Can profiling stay at the shim level?  I don't see anything in these methods that wrap automatic objects, so all that's added in hal/device.h, hal.h, hal2.h, hal2.cpp are pass through functions to shim.  

 I can imagine that the device handle may be missing thus making it difficult to call shim, but then expose that, rather than adding a slew of new functions and additional profiling specific code that really don't provide any object abstractions.

src/runtime_src/core/pcie/user_aws/shim.cpp
```diff
@@ -71,7 +71,42 @@ extern char* get_afi_from_axlf(const axlf *);
...
+
+// Profiling
+#define AXI_FIFO_RDFD_AXI_FULL          0x1000
+#define MAX_TRACE_NUMBER_SAMPLES                        16384
+#define XPAR_AXI_PERF_MON_0_TRACE_WORD_WIDTH            64
+
+
 namespace awsbwhal {
+
+#if 0
```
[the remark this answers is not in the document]
IshitaGhosh: Currently, there are two copies of AlignedAllocator : in linux/shim.cpp and user_aws/shim.h. I can move them to the common area.

PR #1648, opened 2019-07-12
 xclRegRead and xclRegWrite support 

maxzhen:
This PR introduces two new XRT APIs for accessing register space of a specific CU. Before access, application should first exclusively reserve the target CU.

### DIFF
--- a/src/runtime_src/core/pcie/linux/shim.cpp
+++ b/src/runtime_src/core/pcie/linux/shim.cpp
@@ -72,6 +72,12 @@ is_multiprocess_mode()
   return val;
 }
 
+static inline void errlog(std::string& errstr)
+{
+    xrt_core::message::send(xrt_core::message::severity_level::XRT_ERROR,
+        "XRT", errstr.c_str());
+}
+
 /*
  * numClocks()
  */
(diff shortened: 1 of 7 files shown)


src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1334,6 +1350,49 @@ uint shim::xclGetNumLiveProcesses()
     return 0;
 }
 
+int shim::xclRegRW(bool rd, uint32_t cu_index, uint32_t offset, uint32_t *datap)
+{
+    std::lock_guard<std::mutex> l(mCuMapLock);
+
+    if (cu_index >= mCuMaps.size()) {
+        std::cerr << "ERROR: xclRegRW: invalid CU index: " <<
```
stsoe: Can we use xrt_core::message::send for all cerr calls?

maxzhen:
> 
> 
> This looks good. Much safer than current xclRead/Write, but in reality just a side way to control CUs outside the official execution model.

This is mainly for providing a way to get access to non-control registers, such as read customized IP status...

src/runtime_src/core/pcie/linux/shim.cpp
maxzhen: > 
> 
> Can we use xrt_core::message::send for all cerr calls?

Will do!

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1334,6 +1356,55 @@ uint shim::xclGetNumLiveProcesses()
...
+{
+    std::lock_guard<std::mutex> l(mCuMapLock);
+
+    if (cu_index >= mCuMaps.size()) {
+        std::string err = "xclRegRW: invalid CU index: ";
+        err += std::to_string(cu_index);
+        errlog(err);
+        return -EINVAL;
+    }
+    if (offset >= mCuMapSize || (offset % sizeof(uint32_t)) != 0) {
```
sonals: offset & 0x3 to determine if it is multiple of 4?

PR #1649, opened 2019-07-12
Enabling interrupts in 28nm Zynq devices

Himanshu-xilinx:
* Enabled interrupts in 28nm Zynq devices 
* Dts and Dynamic post link script provided by Siva Rajesh Jarugula


maxzhen:
ok to test

PR #1651, opened 2019-07-12
Fix a historic legacy -- always run DMA test in multi-channel mode

sonals:


maxzhen:
What is the DMA test performance number this change gives you?

PR #1653, opened 2019-07-12
fix xbutil reset panic

houlz0507:


maxzhen:
Tested on xsjmaxz50 with two boards. The fix worked fine.

PR #1660, opened 2019-07-15
python test added to xbutil validate

AShivangi:
- added python as a dependency
- packaged verify python test in _/opt/xilinx/test_

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
+++ b/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
@@ -908,8 +908,9 @@ int runShellCmd(const std::string& cmd, std::string& output)
 }
 
 int searchXsaAndDsa(std::string xsaPath, std::string 
-    dsaPath,struct stat &st, std::string& path, std::string &output) 
+    dsaPath, std::string& path, std::string &output) 
 {
+    struct stat st;
     if (stat(xsaPath.c_str(), &st) != 0) {
             if (stat(dsaPath.c_str(), &st) != 0) {
                 output += "ERROR: Failed to find test in ";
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -906,33 +907,55 @@ int runShellCmd(const std::string& cmd, std::string& output)
     return ret;
 }
 
+int searchXsaAndDsa(std::string xsaPath, std::string 
+    dsaPath,struct stat &st, std::string& path, std::string &output)
```
maxzhen: why pass in struct stat? it should be small enough to fit in stack, right?

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -906,33 +907,55 @@ int runShellCmd(const std::string& cmd, std::string& output)
...
+            if (stat(dsaPath.c_str(), &st) != 0) {
+                output += "ERROR: Failed to find test in ";
+                output += xsaPath;
+                output += " and ";
+                output += dsaPath;
+                return -ENOENT;
+            }
+            path =  dsaPath;
+            return EXIT_SUCCESS;
+        } else {
```
maxzhen: The indent looks strange, please fix.

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -906,33 +907,55 @@ int runShellCmd(const std::string& cmd, std::string& output)
...
+            path = xsaPath;
+            return EXIT_SUCCESS;
+        }
+}
+
 int xcldev::device::runTestCase(const std::string& exe,
-    const std::string& xclbin, std::string& output)
+    const std::string& xclbin, const bool isPython, std::string& output)
```
maxzhen: there is no need to pass in isPython since you do the search internally anyway. You should know what exe to use based on the search result.

PR #1663, opened 2019-07-15
Removing the usages of unused xclAllBO() argument

haeseung81:
@maxzhen I have removed all the usages of "int unused" argument of xclAllocBO() for all the xrt testcases.
Can you check this out?

If my changes are good to go, then you can go ahead and restore the changes in https://github.com/Xilinx/XRT/pull/1584/


maxzhen:
ok to test

PR #1670, opened 2019-07-16
[EDGE] support xclRegRead and xclRegWrite

maxzhen:
This PR is mostly ported from DC platform and was tested on zcu102ng platform.

There is one difference in behavior b/w edge and DC. Since we don't have real CU reservation support on edge, we do not require user to exclusively reserve a CU before it can access the reg space using xclRegRead() and xclRegWrite().

### DIFF
--- a/src/runtime_src/core/common/bo_cache.h
+++ b/src/runtime_src/core/common/bo_cache.h
@@ -20,6 +20,7 @@
 #include <vector>
 #include <utility>
 #include <mutex>
+#include <sys/mman.h>
 
 #include "xrt.h"
 #include "ert.h"
--- a/src/runtime_src/core/edge/user/shim.h
+++ b/src/runtime_src/core/edge/user/shim.h
@@ -26,11 +26,9 @@
 #include <fstream>
 #include <map>
 #include <vector>
-
-// Forward declaration
-namespace xrt_core {
-    class bo_cache;
-}
+#include <mutex>
+#include <memory>
+#include "core/common/bo_cache.h"
 
 namespace ZYNQ {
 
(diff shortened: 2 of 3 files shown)


src/runtime_src/core/common/bo_cache.h
```diff
@@ -20,6 +20,7 @@
 #include <vector>
 #include <utility>
 #include <mutex>
+#include <sys/mman.h>
```
sonals: sys/mman.h is not portable across windows. Have the clients include this file.

maxzhen:
Merging it for now since it fixes compilation breakage. Let me know if you have other comments. I'll fix them right away.

src/runtime_src/core/edge/user/shim.h
```diff
@@ -26,11 +26,9 @@
...
 #include <map>
 #include <vector>
-
-// Forward declaration
-namespace xrt_core {
-    class bo_cache;
-}
+#include <mutex>
+#include <memory>
+#include "core/common/bo_cache.h"
```
stsoe: I do think we should stick with forward declaration.  It looks like the real problem was xrt::core vs xrt_core?

src/runtime_src/core/common/bo_cache.h
stsoe: I see this needed for Linux since the header files does call munmap.  On windows it doesn't though, because xrt.h on windows includes an munmap function defined in a header file included conditionally.   Leave this #include but guard it by #ifndef _WIN32

src/runtime_src/core/edge/user/shim.h
maxzhen: > 
> 
> I do think we should stick with forward declaration. It looks like the real problem was xrt::core vs xrt_core? and missing

There are multiple issues. xrt::core vs xrt_core is one of them. The other one is caused by forward declaration and unique_ptr. We will hit a static_assert in unique_ptr.h after I include <memory>. The last issue is the missing sys/mman.h for mmap/munmap calls.

src/runtime_src/core/common/bo_cache.h
maxzhen: > 
> 
> I see this needed for Linux since the header files does call munmap. On windows the include isn't needed here, because xrt.h on windows includes an munmap function defined in a header file included conditionally. Leave this #include but guard it by #ifndef _WIN32

Will fix with a new PR.

PR #1671, opened 2019-07-17
[EDGE] change dsa to xsa

mamin506:
Vivado and PetaLinux 2019.2 use XSA (Xilinx Shell Archive) to replace DSA.


xdavidz:
This look ok to me. Can we mention that we must use 2019.2 or later petalinux with those scripts in the README.md?

PR #1679, opened 2019-07-17
Add metadata files for new sprite flow

haeseung81:
Since we are moving to new sprite flows, I put new metadata files for new sprite flow.


maxzhen:
ok to test

PR #1701, opened 2019-07-23
clean up 32bit lint errors of zocl

xdavidz:
seems that's the only lint issues in 32bit zocl code. Verified both on 32 and 64 platforms.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_ert.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_ert.c
@@ -60,8 +60,8 @@ static int zocl_ert_probe(struct platform_device *pdev)
 		return PTR_ERR(map);
 	}
 	ert->hw_ioremap = map;
-	ert_info(pdev, "IP(embedded_scheduler_hw) IO start %llx, end %llx",
-			      res->start, res->end);
+	ert_info(pdev, "IP(embedded_scheduler_hw) IO start %lx, end %lx",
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/edge/drm/zocl/zocl_ert.c
```diff
@@ -61,7 +61,7 @@ static int zocl_ert_probe(struct platform_device *pdev)
 	}
 	ert->hw_ioremap = map;
 	ert_info(pdev, "IP(embedded_scheduler_hw) IO start %llx, end %llx",
-			      res->start, res->end);
+	      (unsigned long long)res->start, (unsigned long long)res->end);
```
larry9523: size_t is long unsigned int type, would it make more sense to use "lx" and cast it to (unsigned long) ?

xdavidz:
please hold this request for me. I found more lint errors for 32bit.

src/runtime_src/core/edge/drm/zocl/zocl_ert.c
xdavidz: agree

xdavidz:
Please re-review the changes. What I intent to do is to make 32 and 64 code clean enough for those addresses. Only make the code work for 64 is simple but it may block 32 badly. Hopefully, there are not too many issues any more. Now the build of 32 and 64 are both clean (no lint errors).

PR #1702, opened 2019-07-23
Added thread safety for session create & destroy. And buffer free cha…

sarab96:
Added thread safety for session create & destroy. And buffer free changes.


abalasa:
retest this please

PR #1711, opened 2019-07-24
Fix test metadata issue for new sprite flow

haeseung81:
Fix test metadata issue for new sprite flow.
All testinfo.yml files for unit_test and xrt tests are updated to fix metadata issue.


maxzhen:
ok to test

PR #1724, opened 2019-07-26
Updated signature info in app/.. header files

sarab96:
Updated signature info in app/.. header files


abalasa:
retest this please

PR #1731, opened 2019-07-29
Support context/queue in KDS

stsoe:
Add support for context / queue feature in KDS.

Add exec_write method to specify context.

Modify exec_write to use CU offset rather than global offset, this is
a breaking change.

Reserve exec_write payload [0..6] corresponding to 0x0..0x14 for CU
control so that addr,value pair starts at 0x18.  0x10 and 0x14 are for
context in/out and are used if and only if kernel enables context feature.


abalasa:
retest this please

maxzhen:
retest this please

PR #1747, opened 2019-07-31
Added OpenSSL development library dependency.

rozumx:
xclbinutil will start signing xclbin archives and needs the OpenSSL development library to do so, hence the dependency.


rozumx:
> I think you also need to add openssl as dependency for xrt package?

I have updated the XRT package dependencies to include the OpenSll Development library.

Note
-----
While it looks to be working as expected (build and installing on an OS), this is the first time I have modified this file and flow.  @maxzhen Please do review and run against the with the changes to validate correct operation as well.

PR #1752, opened 2019-08-01
Feature Added: Signing xclbin archives

rozumx:
Work Done
---------
+ Created C functions to sign and verify the signatures of an xclbin archive using the openssl library
+ Updated xclbinutil to support passing in the private and certificate keys
+ Updated xclbintuil to support verifying a signed xclbin archive
+ Updated cmake scripts to be aware of the openssl library
  - Disabled use of openssl libraries for window builds
+ Replaced unsigned variable 'char axlf.m_cipher[28]' with 'int32_t m_signature_length' and 'char reserved[28]'
+ Updated depricated xclbincat to no longer support setting the axlf.m_cipher value
+ Updated FormattedOutput and XclBinClass classes to be signature aware

### DIFF
--- a/src/runtime_src/core/include/xclbin.h
+++ b/src/runtime_src/core/include/xclbin.h
@@ -197,7 +197,9 @@ extern "C" {
 
     struct axlf {
         char m_magic[8];                            /* Should be "xclbin2\0"  */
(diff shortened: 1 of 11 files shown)


rozumx:
Working on addressing library dependency issues for cento.
Ubuntu compiles successfully

PR #1753, opened 2019-08-01
Updated to read  the install path from the new env

venkatp-xilinx:


venkatp-xilinx:
As per the mail communication with Sudipto & Amit, conclusion is that we can use the Internal name as part of the source code, however not yet the user visible areas like comments, description etc.

venkatp-xilinx:
Can someone take a stand and merge my change.

PR #1755, opened 2019-08-01
xclCopyBO enhancements

xdavidz:
While I am doing documentation of zocl dma, I think we'd better make the code even clearer for the future linux kernel 5.0x. I think we should handle all possible status and most importantly handle error conditions and return appropriate error. 

Also found that the xclCopyBO needs to be enabled for client.

Found a lint error.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/sched_exec.c
+++ b/src/runtime_src/core/edge/drm/zocl/sched_exec.c
@@ -1352,16 +1352,6 @@ scu_unconfig_done(struct sched_cmd *cmd)
 	return true;
 }
 
-static inline int
-dma_done(struct sched_cmd *cmd)
-{
-	if (cmd->dma_handle.dma_flags & ZOCL_DMA_DONE) {
-		cmd->dma_handle.dma_flags &= ~ZOCL_DMA_DONE;
-		return true;
-	}
-	return false;
-}
-
 /**
  * notify_host() - Notify user space that a command is complete.
  */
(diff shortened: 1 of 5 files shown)


src/runtime_src/core/edge/drm/zocl/zocl_dma.h
```diff
@@ -22,9 +22,10 @@
 #include <linux/dmaengine.h>
 #include <linux/dma-mapping.h>
 
-#define ZOCL_DMA_DONE (1 << 0)
+#define ZOCL_DMA_DONE 	(1 << 0)
```
mamin506: Alignment of these two macro.

xdavidz: fixed, thanks!

src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -2219,29 +2219,29 @@ static void
...
 {
 	u32 opc = opcode(cmd);
-	bool cmd_complete = false;
 
 	SCHED_DEBUG("-> penguin_queury() slot_idx=%d\n", cmd->slot_idx);
 	switch (opc) {
 	case ERT_START_COPYBO:
-		if (dma_done(cmd))
-			cmd_complete = true;
+		if (dma_done(cmd)) {
```
larry9523: I suggest a little wrapper on your dam_* functions. Since we have a wrapper to handle dma_handle.dma_flags, it would be neater to treat dma_flags as a private field.
dma_done (just check done bit but not clear it)
dma_error (just check error bit but not clear it)
dma_clear (clear a specific flag ro all flags)

src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -2208,6 +2198,28 @@ fini_scheduler_thread(void)
 	return retval;
 }
 
+static inline int
+dma_done(struct sched_cmd *cmd)
+{
+	if (cmd->dma_handle.dma_flags & ZOCL_DMA_DONE) {
+		cmd->dma_handle.dma_flags &= ~ZOCL_DMA_DONE;
```
larry9523: Any particular reason we have to clear DONE bit here? I think DONE and ERROR flags should have same life cycle and you can clear them altogether in your penguin query call.

PR #1756, opened 2019-08-01
Have to use relative path. WORKDIR didn't compatible with EXTERNAL_SRC

mamin506:
Revert this change #1750. Because ${WORKDIR} didn't compatible with EXTERNALSRC. We rely on EXTERNALSRC to build from local source code.
The XRT recipe in meta-xilinx repo, for PetaLinux team, would use ${WORKDIR}. Since PetaLinux Yocto hate relative path in a recipe.

Note: The recipe in the XRT repo is the latest. Sometime, when we update the recipe in meta-xilinx repo, we need to take care of this difference.


abalasa:
retest this please

PR #1760, opened 2019-08-02
Add use of new AIM, AM, ASM profile ip objects and xclRead/Write based profile data collection in HW Emu flow.

IshitaGhosh:
This support is turned off by default. It can be turned on using "system_dpa=true" configuration in Emulation section of xrt.ini. However, this expects that the xclbin has been created using new Emulation models which use block automation and System DPA. The new emulation model is under test now. So support in XRT has been put under "system_dpa=true" configuration control. The full support, both in tools and XRT, is expected to be default flow in 2019.2 Release.


maxzhen:
retest this please

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -39,6 +39,8 @@ namespace xclhwemhal2 {
   const unsigned HwEmShim::CONTROL_AP_IDLE  = 4;
   const unsigned HwEmShim::CONTROL_AP_CONTINUE  = 0x10;
 
+#define AXI_FIFO_RDFD_AXI_FULL          0x1000
```
pgschuey: Isn't this already defined in xcl_perfmon_parameters.h?

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -2223,6 +2229,101 @@ ssize_t HwEmShim::xclUnmgdPread(unsigned flags, void *buf, size_t count, uint64_
   return xclCopyBufferDevice2Host(buf, offset, count, 0 , 0);
 }
 
+
```
pgschuey: Was there a reason why this entire section was added to shim.cxx instead of perf.cxx? Moving it to perf.cxx would be a better place so all debug/profile code gets isolated and shim.cxx is not overloaded with it.

IshitaGhosh: With the new ProfileIP class based flow in re-architected XDP, the reading of profile/trace data from monitors is handled inside XDP using xclRead/Write and only a couple of flow dependent Shim methods (added in the current section). When this support is fully verified, the older implementation of profiling can be removed entirely. We may not need a separate file for profiling then and perf.cxx may be removed. That is why, I have added these 3 new SHIM methods for profiling in shim.cxx.

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
IshitaGhosh: Yes, but xcl_perfmon_parameters.h is not included in this shim.cxx file. Please check my reply on your next comment for details.

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
pgschuey: Okay, makes sense.

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
pgschuey: Can we not include it? Seems odd to define something twice in two different locations.

IshitaGhosh: Yes, we can include it. I have added this change now. Thanks!

PR #1762, opened 2019-08-03
Python/pyopencl based xbutil validate

AShivangi:
- removed testcase exe dependency
- python based verify and pyopencl based bandwidth
- works for multiple devices

Output:
```
INFO: Found 1 cards

INFO: Validating card[0]: xilinx_u250_xdma_201830_2
INFO: == Starting DMA test:
Host -> PCIe -> FPGA write bandwidth = 9420.86 MB/s
Host <- PCIe <- FPGA read bandwidth = 11940.8 MB/s
INFO: == DMA test PASSED
INFO: == Starting device memory bandwidth test:
..........
Maximum throughput: 52428 MB/s
INFO: == device memory bandwidth test PASSED
INFO: == Starting PCIE peer-to-peer test:
P2P BAR is not enabled. Skipping validation
INFO: == PCIE peer-to-peer test SKIPPED
INFO: == Starting memory-to-memory DMA test:
M2M is not available. Skipping validation
INFO: == memory-to-memory DMA test SKIPPED
INFO: Card[0] validated successfully.

INFO: All cards validated successfully.
```


src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -971,27 +960,15 @@ int xcldev::device::runTestCase(const std::string& exe,
...
-    if (m_idx != 0)
-        idxOption = "-d " + std::to_string(m_idx);
-
-    std::string cmd = "";
-    if (isPython) {
-        cmd = "python " + exePath + " -k " + xclbinPath + " " + idxOption;
-    } else {
-        cmd = exePath + " " + xclbinPath + " " + idxOption;
-    }
+    std::string cmd = "python " + xrtTestCasePath + " -k " + xclbinPath + " -d " + std::to_string(m_idx) + " 2>&1"; //catch stderr
```
maxzhen: 2>&1 is bash specific? does it work with any shell, like csh?

AShivangi: I just tried it on csh shell, both passing and failing scenarios. It works fine.

maxzhen: > 
> 
> I just tried it on csh shell, both passing and failing scenarios. It works fine.

what do you get if you manually run this cmd string in csh?

AShivangi: It doesn't work when I run the command manually... According to "popen" doc, the command is passed to /bin/sh using the -c flag

maxzhen: do we want to explicitly specify bash in the cmd string? /bin/sh can point to any shell...

AShivangi: No we don't

maxzhen: so what if /bin/sh is pointing to csh?

AShivangi: Oooh. Ok I see your point. This won't work in that case. I'll change this

abalasa:
retest this please

PR #1779, opened 2019-08-07
Added dataflow kernel support

sarab96:
Added dataflow kernel support

### DIFF
--- a/src/xma/src/xmaplugin/xmaplugin.cpp
+++ b/src/xma/src/xmaplugin/xmaplugin.cpp
@@ -465,6 +465,14 @@ xma_plg_schedule_work_item(XmaSession s_handle)
     cu_cmd->data[2] = kernel_tmp1->cu_mask3;
     // Copy reg_map into execBO buffer 
     memcpy(&cu_cmd->data[3], src, size);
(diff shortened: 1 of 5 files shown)


src/xma/src/xmaplugin/xmaplugin.cpp
```diff
@@ -465,6 +465,14 @@ xma_plg_schedule_work_item(XmaSession s_handle)
     cu_cmd->data[2] = kernel_tmp1->cu_mask3;
     // Copy reg_map into execBO buffer 
     memcpy(&cu_cmd->data[3], src, size);
+    if (kernel_tmp1->dataflow_kernel) {
```
stsoe: Not all dataflow kernels will support the concept of channels.   Are all video dataflow kernels supporting this?   How will they advertise to the scheduler that the kernel has multiple channels.   In 2019.2 there is no xclbin support to tell XRT that a kernel has multiple channels.   I worked around this using a new key in xrt.ini with that information.

sarab96: Using ini setting for this works for XMA as well. I will call you to discuss more.

PR #1785, opened 2019-08-08
CL_USE_HOST_MEM_PTR with zero copy for edge

Himanshu-xilinx:
* Api to alloc/free mem aligned and continues host buffer for edge.
* Api to alloc/free mem aligned host buffer for pcie.
* hal api xclAllocHostPtr xclFreeHostPtr
* opencl api clAllocHostPtr clFreeHostPtr

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
@@ -379,6 +379,7 @@ zocl_userptr_bo_ioctl(struct drm_device *dev, void *data, struct drm_file *filp)
 	struct drm_zocl_userptr_bo *args = data;
 	struct page **pages;
 	unsigned int sg_count;
+	DRM_INFO("%s:%d args->addr:%x\n",__func__,__LINE__,args->addr);
 
 	if (offset_in_page(args->addr)) {
 		DRM_ERROR("User ptr not PAGE aligned\n");
@@ -405,6 +406,7 @@ zocl_userptr_bo_ioctl(struct drm_device *dev, void *data, struct drm_file *filp)
 		goto out1;
 	}
 
+/*
 	ret = get_user_pages_fast(args->addr, page_count, 1, pages);
 	if (ret != page_count) {
 		DRM_ERROR("Unable to get user pages\n");
@@ -427,8 +429,11 @@ zocl_userptr_bo_ioctl(struct drm_device *dev, void *data, struct drm_file *filp)
 	}
 
 	bo->cma_base.paddr = sg_dma_address((bo->cma_base.sgt)->sgl);
+*/
+	bo->cma_base.paddr = args->addr;
 
 	/* Physical address must be continuous */
+/*
 	if (sg_count != 1) {
 		DRM_ERROR("User buffer is not physical contiguous\n");
 		ret = -EINVAL;
@@ -436,7 +441,7 @@ zocl_userptr_bo_ioctl(struct drm_device *dev, void *data, struct drm_file *filp)
 	}
 
 	bo->cma_base.vaddr = (void *)(uintptr_t)args->addr;
-
+*/
 	ret = drm_gem_handle_create(filp, &bo->cma_base.base, &args->handle);
 	if (ret) {
 		ret = -EINVAL;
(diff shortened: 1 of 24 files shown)


larry9523:
A couple of high level questions
1) I thought the problem we are trying to solve is, user application has a pre allocated buffer (not allocated by XRT interfaces), which might be not physically contiguous and 4K aligned, and they try to use that buffer by CL_USE_HOST_MEM_PTR flag. How would this new API can help with that?
2) Anything new that this API buy us if applications are not using xclAllocBO, xclMapBO and xclGetBOProperties?

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -338,6 +350,42 @@ void *ZYNQShim::xclMapBO(unsigned int boHandle, bool write)
...
+{
+  std::cout << "ZYNQShim::xclFreeHostPtr called" << std::endl;
+  std::lock_guard<std::mutex> l(mBOMapLock);
+  auto it = mBoMap.find(reinterpret_cast<uint64_t>(ptr));
+  if (it == mBoMap.end()) {
+    std::cout  << "xclFreeHostPtr Bad Host PTR" << std::endl;
+    return;
+  }
+  munmap(ptr,it->second.size);  
+  xclFreeBO(it->second.boHandle);
```
stsoe: Wouldn't it be better to have BOData constructor and destructor manage xclAllocBO and xclFreeBO

src/runtime_src/core/include/xrt_helper.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include <vector>
+
+namespace xrt_util {
+
+class helper {
+    static helper* mInstance;
+	std::vector<xclDeviceHandle> mHandleList;
+	helper() {
+		xclDeviceInfo2 deviceInfo;
+		int ndevice = xclProbe();;
```
stsoe: double ;;

src/runtime_src/core/include/xrt_helper.hpp
```diff
@@ -0,0 +1,67 @@
...
+    		if(ndevice  == 0) {
```
stsoe: No curly braces around single statement.

src/runtime_src/core/include/xrt_helper.hpp
```diff
@@ -0,0 +1,67 @@
...
+			throw std::runtime_error("No device found");
+		}
+
+		for(int i=0;i<ndevice; i++) {
+			xclDeviceHandle handle;
+			handle = xclOpen(i, NULL, XCL_INFO);
```
stsoe: nullptr in C++
also why three lines of code?   just write ...push_back(xclOpen(...));

src/runtime_src/core/include/xrt_helper.hpp
```diff
@@ -0,0 +1,67 @@
...
+			mHandleList.push_back(handle);
+		}
+	}	
+public:
+    static helper* getinstance(unsigned deviceIndex=0) {
+		if(mInstance == nullptr) {
+			mInstance = new helper();
```
stsoe: Who will delete this?
Just use a static variable and return address of that.

src/runtime_src/core/include/xrt_helper.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+	~helper() {
+		for (auto it : mHandleList) {
+			xclClose(it);
+		}
+	}
+
+};
+
+helper* helper::mInstance = nullptr;
```
stsoe: No needed.

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -427,16 +428,19 @@ zocl_userptr_bo_ioctl(struct drm_device *dev, void *data, struct drm_file *filp)
 	}
 
 	bo->cma_base.paddr = sg_dma_address((bo->cma_base.sgt)->sgl);
+*/
```
mamin506: Why do we remove this check? What if user use a wrong pointer at the call? We would lost protect.

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -427,16 +428,19 @@ zocl_userptr_bo_ioctl(struct drm_device *dev, void *data, struct drm_file *filp)
...
+/*
 	if (sg_count != 1) {
 		DRM_ERROR("User buffer is not physical contiguous\n");
 		ret = -EINVAL;
 		goto out0;
 	}
 
 	bo->cma_base.vaddr = (void *)(uintptr_t)args->addr;
-
+*/
```
mamin506: see above.

sonals:
As @mamin506 and @larry9523 indicated let's take this offline and discuss how it solves the host ptr needs from specific applications.

maxzhen:
@mamin506 will implement a work around for now.

PR #1789, opened 2019-08-08
QDMA update for SMARTN

karenx-xilinx:


src/runtime_src/core/pcie/driver/linux/xocl/lib/libqdma/qdma_context.c
```diff
@@ -186,6 +186,7 @@ static int make_sw_context(struct qdma_descq *descq, u32 *data, int cnt)
 	data[3] = PCI_DMA_H(descq->desc_bus);
 	data[2] = PCI_DMA_L(descq->desc_bus);
 	data[1] = (1 << S_DESC_CTXT_W1_F_QEN) |
+			(1 << S_DESC_CTXT_W1_F_MRKR_DIS) |
```
houlz0507: Will it cause any issue to add (1 << S_DESC_CTXT_W1_F_MRKR_DIS) unconditonally?

karenx-xilinx: No, the marker is not enabled in SDx QDMA image, I want to disable it explicitly since driver does not support this feature.

PR #1796, opened 2019-08-09
patch from Michael for ACC adapter support in the xclbin

mamin506:
This is from Michael's patch.


mamin506:
This take extra 30 minutes for the pipeline incremental build.

maxzhen:
the change has actually reviewed by @rozumx offline, merging it now.

PR #1805, opened 2019-08-12
add version option to xbutil and xbmgmt and xbutil top -i fix

AShivangi:
CR1:
added version option to xbutil and xbmgmt

CR2:
xbutil top doesn't crash when the interval <1. If the user gives a value <1, the interval is set to 1.


abalasa:
retest this please

PR #1812, opened 2019-08-13
Transfer Rate calculations using AXI Times instead of CU times

hackwa:
I have chosen to not make this change compatible with previous time calculation. So if AXI times are missing, then transfer rate is **not**  calculated using CU times. This was done to avoid confusion as both transfer rates are different by a magnitude and aren't even comparable.


abalasa:
retest this please

hackwa:
retest this please

abalasa:
retest this please

PR #1817, opened 2019-08-14
xclLogMsg changes

sarab96:
xclLogMsg changes

### DIFF
--- a/src/runtime_src/core/pcie/linux/shim.cpp
+++ b/src/runtime_src/core/pcie/linux/shim.cpp
@@ -1515,13 +1515,14 @@ int xclLogMsg(xclDeviceHandle handle, xrtLogMsgLevel level, const char* tag, con
 {
     va_list args;
     va_start(args, format);
-    xocl::shim *drv = xocl::shim::handleCheck(handle);
-    int ret = drv ? drv->xclLogMsg(level, tag, format, args) : -ENODEV;
+    //Do NOT add device handle check
+    //This is logging function independent of device
+    //XMA passes NULL device handle to this API
+    int ret = xocl::shim::xclLogMsg(level, tag, format, args);
     va_end(args);
 
     return ret;
 }
-
 
 size_t xclWrite(xclDeviceHandle handle, xclAddressSpace space, uint64_t offset, const void *hostBuf, size_t size)
 {
--- a/src/runtime_src/core/pcie/user_aws/shim.cpp
+++ b/src/runtime_src/core/pcie/user_aws/shim.cpp
@@ -51,6 +51,7 @@
 #include <sys/file.h>
 
 #include <cstdio>
+#include <cstdarg>
 #include <cstring>
 #include <cassert>
 #include <algorithm>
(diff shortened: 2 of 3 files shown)


src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1515,14 +1515,15 @@ int xclLogMsg(xclDeviceHandle handle, xrtLogMsgLevel level, const char* tag, con
 {
     va_list args;
     va_start(args, format);
-    xocl::shim *drv = xocl::shim::handleCheck(handle);
-    int ret = drv ? drv->xclLogMsg(level, tag, format, args) : -ENODEV;
+    //Do NOT add device handle check
```
maxzhen: maybe we want to introduce a new api and remove handle from arg list and mark this one obsolete?

src/runtime_src/core/pcie/user_aws/shim.cpp
```diff
@@ -930,6 +931,51 @@ namespace awsbwhal {
         return size;
     }
     
+    /*
```
stsoe: This code is common between linux shim, aws shim, edge shim (probably), and windows shim.  The common part should be moved to src/runtime_src/core/common/message,h,cpp.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1515,14 +1515,15 @@ int xclLogMsg(xclDeviceHandle handle, xrtLogMsgLevel level, const char* tag, con
...
+    //This is logging function independent of device
+    //XMA passes NULL device handle to this API
```
AShivangi: I am using xclLog() for internal logging in shim.cpp without the handle. Didn't realize I could pass NULL handle. But maybe we should make a new API as Max suggested? If not then I can remove xclLog and use xclLogMsg with NULL handle

sarab96: Let me check on aws change and the handle arg. Now this API is used in all emulation code as well. so need to check what all will be impacted.

src/runtime_src/core/pcie/user_aws/shim.cpp
sarab96: It is already using xrt_core::message::send common code.

src/runtime_src/core/pcie/linux/shim.cpp
sarab96: How do we ensure that it will make it to common code for aws also? Right now this is missing from aws code. This API is now used in emulation code also. So changing it and removing device handle impacts them also.

maxzhen: after we remove aws version of shim.cpp, we'll use our version of shim.cpp where the api has already been available. i mean add a new one without device handle. i did not mean changing existing one.

src/runtime_src/core/pcie/linux/shim.cpp
maxzhen: If it's only used in our code (including emulation), we can change them all. it should not be hard to do since we are removing arg, not adding one, which might be difficult. however, this is not the reason we can't change it. i believe that this api has been shipped with 2019.1, so it may be used by end users already. i think that if we want to change it, we'll need to mark existing one as obsolete and remove it in later release.

sarab96: ok. I will make that change.

src/runtime_src/core/pcie/linux/shim.cpp
sarab96: ok.  I will make that change.

abalasa:
retest this please

sarab96:
Closing this PR. As per discussion and comments I made the changes and submitted new PR# 1819 for latest changes.

PR #1819, opened 2019-08-14
xclLog API changes

sarab96:
xclLog API changes

### DIFF
--- a/src/runtime_src/core/include/xrt.h
+++ b/src/runtime_src/core/include/xrt.h
@@ -432,6 +432,7 @@ XCL_DRIVER_DLLESPEC unsigned int xclVersion();
 
 
 /**
+ * Deprecated API
  * xclLogMsg() - Send message to log file as per settings in ini file.
  *
  * @handle:        Device handle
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/include/xrt.h
```diff
@@ -442,7 +443,19 @@ XCL_DRIVER_DLLESPEC unsigned int xclVersion();
...
+ * xclLog() - Send message to log file as per settings in ini file.
+ *
+ * @level:         Severity level of the msg
+ * @tag:           Tag supplied by the client, like "OCL", "XMA", etc.
+ * @format:        Format of Msg string to write to log file
+ * @...:           All other arguments as per the format
+ *
+ * Return:         0 on success or appropriate error number
+ */
+XCL_DRIVER_DLLESPEC int xclLog(enum xrtLogMsgLevel level, const char* tag, const char* format, ...);
```
AShivangi: Can you also add this to xrt_binding.py?

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1523,6 +1523,15 @@ int xclLogMsg(xclDeviceHandle handle, xrtLogMsgLevel level, const char* tag, con
     return ret;
 }
 
+int xclLog(xrtLogMsgLevel level, const char* tag, const char* format, ...)
+{
```
stsoe: Please compare the log level against the current verbosity level to avoid the va_ stuff.  Ultimately xocl::shim::xclLogMsg does that, but it probably should be done here too.

src/runtime_src/core/include/xrt.h
sarab96: done

src/runtime_src/core/pcie/linux/shim.cpp
sarab96: done

sonals:
I believe the device handle should not be removed. How would you distinguish between messages for different devices?

sarab96:
Two versions of API are available with this PR. One with device handle and other without device handle. XMA generic messages like "xma_initialize failed" do not require device id as this is common for all devices.

maxzhen:
not sure about the usage of device handle. the device handle is used to redirect the call to shim::xclLogMsg() which does not add any device specific info to the output?

stsoe:
I don't think the XRT logging API belongs with shim implementation.   It should be the same for all implementations of shim, alas factored out and implemented in core/common.   
The API will be used by shim implementations when errors/warnings/info are necessary and the message can include device identification.

sarab96:
Closing. As discussed will change current API

PR #1822, opened 2019-08-15
Use sysfs nodes in xbutil readSensors

AShivangi:


src/runtime_src/core/pcie/tools/xbutil/xbutil.h
```diff
@@ -714,13 +771,13 @@ class device {
...
         }
+        
         getMemTopology( devstat );
-        // stream
 
         // xclbin
-        std::string errmsg, xclbinid;
-        pcidev::get_dev(m_idx)->sysfs_get("", "xclbinuuid", errmsg, xclbinid);
-        if(errmsg.empty()) {
+        {
```
maxzhen: why remove the check for errmsg?

src/runtime_src/core/pcie/tools/xbutil/xbutil.h
```diff
@@ -748,12 +805,6 @@ class device {
         sensor_tree::put("debug_profile.device_info.nifd_name", std::string(info.nifd_name));
         /** End of debug and profile device information */
 
-        // p2p enable
```
maxzhen: why removed this?

src/runtime_src/core/pcie/tools/xbutil/xbutil.h
```diff
@@ -814,50 +867,55 @@ class device {
         }
         ostr << "~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~\n";
         ostr << "Temperature(C)\n";
-        // use get_pretty for Temperature and Electrical since the driver may rail unsupported values high
         ostr << std::setw(16) << "PCB TOP FRONT" << std::setw(16) << "PCB TOP REAR" << std::setw(16) << "PCB BTM FRONT" << std::endl;
-        ostr << std::setw(16) << sensor_tree::get_pretty<unsigned short>( "board.physical.thermal.pcb.top_front" )
-             << std::setw(16) << sensor_tree::get_pretty<unsigned short>( "board.physical.thermal.pcb.top_rear"  )
-             << std::setw(16) << sensor_tree::get_pretty<unsigned short>( "board.physical.thermal.pcb.btm_front" ) << std::endl;
+        ostr << std::setw(16) << sensor_tree::get( "board.physical.thermal.pcb.top_front", -1 )
```
maxzhen: why get_pretty() -> get()?

AShivangi: get_pretty was converting ints to hex, but now we can directly get hex so no need to convert!

src/runtime_src/core/pcie/tools/xbutil/xbutil.h
AShivangi: Moved it up so that it is with the rest of board.info

src/runtime_src/core/pcie/tools/xbutil/xbutil.h
AShivangi: No sysfs get in this function is checking for errmsg. So I thought it would be good to have some consistency.

src/runtime_src/core/pcie/tools/xbutil/xbutil.h
AShivangi: seems like we need get_pretty for when the value returned by sysfs_get is invalid. So if a node is invalid, it returns int_max instead of 0 or NULL. I can either change these back to get_pretty, or change get func to do what get pretty is doing

PR #1824, opened 2019-08-16
Enable zcu102_base platform which has PR flow enabled

sivarajeshjarugula:
this is the replacement platform for zcu102_revmin


maxzhen:
ok to test

PR #1825, opened 2019-08-16
Error when xbutil and xrt versions don't match

AShivangi:
- check INTERNAL_BUILD env variable, if not true or not defined, it gives an error. Only when the variable is set to true [INTERNAL_BUILD = true], it doesn't give an error
- When the internal build is not set:

  - "xbutil help" doesn't give any error
  - all other xbutil commands give the error before running 
  - "xbutil version" runs and then gives the error:
```
       XRT Build Version: 2.3.0
    Build Version Branch: ver-warning
      Build Version Hash: feb684954f8befe5e09ee32464275ff80d25c95b
 Build Version Hash Date: Wed, 14 Aug 2019 21:09:20 -0700
      Build Version Date: Fri, 16 Aug 2019 09:38:42 -0700
                    XOCL: 2.3.0,c9e50a842f11be668e838165ccc2c8325408a4c7
                 XCLMGMT: 2.3.0,c9e50a842f11be668e838165ccc2c8325408a4c7
XRT build version: 2.3.0
Build hash: feb684954f8befe5e09ee32464275ff80d25c95b
Build date: 2019-08-16 09:38:42
Git branch: ver-warning
PID: 8534
UID: 31939
[Fri Aug 16 09:54:16 2019]
HOST: xsjsagarwal
EXE: /proj/xsjhdstaff1/sagarw/XRT/build/Debug/opt/xilinx/xrt/bin/xbutil
[XRT] ERROR: Mixed versions of XRT and xbutil are not supported.
Please install matching versions of XRT and xbutil.
```


src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -110,6 +110,27 @@ void print_pci_info(std::ostream &ostr)
...
 }
 
+int xrt_xbutil_version_cmp() 
+{
+    /*check xbutil tools and xrt versions*/
+    std::string xrt = sensor_tree::get<std::string>( "runtime.build.version", "N/A" ) + "," 
+        + sensor_tree::get<std::string>( "runtime.build.hash", "N/A" );
+    if ( std::strcmp( xrt.c_str(), xcldev::driver_version("xocl").c_str() ) != 0 ) {
+        xocl::shim::xclLog(XRT_ERROR, "XRT", "Mixed versions of XRT and xbutil are not supported. \
+            \nPlease install matching versions of XRT and xbutil. \n");
```
maxzhen: please also mention the environment variable to let developer know how to bypass the check.

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -110,6 +110,27 @@ void print_pci_info(std::ostream &ostr)
...
+        return -1;
+    }
+    return 0;
+}
+
+const char *getEnvVar( std::string const & key )
+{
+    const char * val = getenv( key.c_str() );
+    if(val == NULL)
```
stsoe: nullptr in C++

Here is how I think you should implement this.   Not important, but one liners are better, IMHO.
```
inline const char* value_or_default(const char* cstr, const char* default)
{ return cstr ? cstr : default ; }
inline const char* getenv_or_false(const char* env)
{ return value_or_default(getenv(env),"false"); }
inline const char* getenv_or_false(const std::string& env)
{ return value_or_false(env.c_str()); }
```

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -197,7 +218,15 @@ int main(int argc, char *argv[])
                                        << std:: endl;
         std::cout.width(26); std::cout << std::internal << "XCLMGMT: " << sensor_tree::get<std::string>( "runtime.build.xclmgmt", "N/A" ) 
                                        << std::endl;
-        return 1;
+
```
stsoe: I think requiring a value for INTERNAL_BUILD to be true is confusing.  Either INTERNAL_BUILD is defined or it is not.  If latter, then following previous comment I'd probably implement
```
inline bool getenv_or_null(const char* env)
{ return value_or_default(getenv(env),nullptr); }

```
then call as
`if (!getenv_or_null("INTERNAL_BUILD")) { ... }`

Saves string comparison.

AShivangi: Thanks for the input. I have made the changes, but I am waiting for Sarabjeet's xclLog API changes PR to get merged so that I can remove xclLog from shim class.

AShivangi:
@stsoe made the changes you requested. I haven't removed xclLog() from shim.h yet since it is being called in shim.cpp. https://github.com/Xilinx/XRT/pull/1819 is changing the same code so I do not want to touch that code right now. Here is what the output looks like now:

```
       XRT Build Version: 2.3.0
    Build Version Branch: ver-warning
      Build Version Hash: c289de4b53cecd00aad62e58a62570b5fadb38bd
 Build Version Hash Date: Tue, 20 Aug 2019 10:31:24 -0700
      Build Version Date: Tue, 20 Aug 2019 10:34:39 -0700
                    XOCL: 2.3.0,8542fc344f3e2108971142bf9053183624ddb964
                 XCLMGMT: 2.3.0,8542fc344f3e2108971142bf9053183624ddb964
ERROR: Mixed versions of XRT and xbutil are not supported.
Please install matching versions of XRT and xbutil.
```

I will change this code to use a logging API once we decide which one we want to use.

AShivangi:
@maxzhen output:
```
       XRT Build Version: 2.3.0
    Build Version Branch: ver-warning
      Build Version Hash: 51bd4a66b55232d9d9a808e2fb2e9a92d89c1bfa
 Build Version Hash Date: Tue, 20 Aug 2019 10:36:21 -0700
      Build Version Date: Tue, 20 Aug 2019 15:23:53 -0700
                    XOCL: 2.3.0,c9ac0d5da2747b33f54b8d93140736a1349d9a35
                 XCLMGMT: 2.3.0,c9ac0d5da2747b33f54b8d93140736a1349d9a35

ERROR: Mixed versions of XRT and xbutil are not supported.
Please install matching versions of XRT and xbutil or
define env variable INTERNAL_BUILD to disable this check

```

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -110,6 +110,25 @@ void print_pci_info(std::ostream &ostr)
     }
 }
 
+int xrt_xbutil_version_cmp()
```
stsoe: static, right?  not exported I assume?

PR #1833, opened 2019-08-19
Fix xclbinutil for Ubuntu 19.04

j-stephan:
The compilation changes introduced for Ubuntu 18.04 need to be applied to Ubuntu 19.04 as well, otherwise `xclbinutil` will fail with a segmentation fault. I didn't include Ubuntu 18.10 since that version is no longer supported by Canonical.

### DIFF
--- a/src/CMake/nativeLnx.cmake
+++ b/src/CMake/nativeLnx.cmake
@@ -65,7 +65,7 @@ INCLUDE (FindBoost)
 # link errors when XRT shared objects try to link with it.
 
 # Static linking with Boost is enabled on Ubuntu 18.04 and later versions.
-if ((${LINUX_FLAVOR} STREQUAL Ubuntu) AND (${LINUX_VERSION} STRGREATER_EQUAL 18.04))
+if ((${LINUX_FLAVOR} STREQUAL Ubuntu) AND (${LINUX_VERSION} VERSION_GREATER 17.10))
    set(Boost_USE_STATIC_LIBS  ON)
 endif()
 
--- a/src/CMake/nativeLnx.cmake
+++ b/src/CMake/nativeLnx.cmake
@@ -65,7 +65,7 @@ INCLUDE (FindBoost)
 # link errors when XRT shared objects try to link with it.
 
 # Static linking with Boost is enabled on Ubuntu 18.04 and later versions.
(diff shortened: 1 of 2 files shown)


maxzhen:
ok to test

j-stephan:
Could you post the error message, please?

src/CMake/nativeLnx.cmake
```diff
@@ -64,8 +64,8 @@ INCLUDE (FindBoost)
 # On older systems libboost_system.a is not compiled with -fPIC which leads to
 # link errors when XRT shared objects try to link with it.
 
-# Static linking with Boost is enabled on Ubuntu 18.04.
-if ((${LINUX_FLAVOR} STREQUAL Ubuntu) AND (${LINUX_VERSION} STREQUAL 18.04))
+# Static linking with Boost is enabled on Ubuntu 18.04 and later versions.
+if ((${LINUX_FLAVOR} STREQUAL Ubuntu) AND ((${LINUX_VERSION} STREQUAL 18.04) OR (${LINUX_VERSION} STREQUAL 19.04)))
```
stsoe: GREATER_EQUAL ?

j-stephan: I believe we'd want STRGREATER_EQUAL here. I wasn't aware that there is such an operator in CMake, will fix this now.

rozumx: Initially, I would have asked you why didn't you use GREATER_EQUAL instead of STRGREATER_EQUAL since the later comparison is a lexicographically comparison. 

But looking at the conditional line, it looks like you will always be comparing numeric values, so this shouldn't be an issue.

j-stephan:
Ugh, apparently Ubuntu 16.04's `cmake` doesn't like `STRGREATER_EQUAL` (or `GREATER_EQUAL` for that matter). Weirdly enough, this only happens for the `ppc64le` architecture, not `amd64`. I think I will go with `VERSION_GREATER 17.10` as a workaround.

keryell:
Curious we did not have this problem earlier with 19,04 (and 18.10 & 19.10 too...).
Perhaps it is related to a recent change to `xclbinutil`...

j-stephan:
@rozumx @stsoe Is this PR ok to merge?

PR #1836, opened 2019-08-19
disable zcu102ng platform specific dtsi

mamin506:
This  zcu102ng platform specific dts file has conflict with BSP flow.
Keep this as comment since it could hint someone who wants to use template flow and add his own device tree node.

NOTE: After this change, if you use template flow to build zcu102ng platform, the ethernet and SD card slot would not work.


abalasa:
retest this please

PR #1844, opened 2019-08-20
Added documentation on XSA security. 

sonals:
Updates to doc CMakeLists.txt so kernel-doc is not downloaded if already downloaded before.


src/runtime_src/doc/toc/security.rst
```diff
@@ -0,0 +1,120 @@
...
+operators.
+
+Signing of Xclbins
+==================
+
+xclbin signing process is similar to signing of Linux kernel modules. xclbins can be signed by XRT utility,
+``xclbinutil``. The signing adds a PKCS7 signature at the end of xclbin. The signing certificate is then
+registered with system key-ring. When xclbin is provided to xclmgmt driver it will refuse to load xclbin
+if xclbin signature valiation fails against the certificates in its key ring.
+
```
maxzhen: There is another security level (level 1) at which ".xilinx_fpga_xclbin_keys" key-ring (instead of system key-ring, which is level 2) will be used. Any one with root privilege can add keys to this key-ring at run time. No need to go through UEFI.
If secure boot is used (driver automatically detects it), level 1 is not available and level 2 will be activated automatically and system key-ring will be used. Otherwise, sys admin can promote secure level from 0 (nothing) to 1 and/or to 2 at run time using xbmgmt.

sonals: Added description of levels. Please review.

PR #1845, opened 2019-08-20
Download Partial bitstream for edge platforms by checking whether platform is PR or not

chvamshi-xilinx:
HI ,

This change contains a data driven way to figure out whether to download partial bitstream or not

If platform is a non-PR-platform, Following check will fail. Dont download the partial bitstream
  // if Platform is a PR-platform, Following check passes as enable_pr value is true by default. Download the partial bitstream
  // If platform is a PR-platform, but v++ generated a full bitstream (using some v++ param). User need to add enable_pr=false in xrt.ini (Though this case is not a regular use case, we should provide a way)

also changed the name of the API to match the ini option name

Thanks
Vamshi Krishna


src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -471,8 +471,14 @@ int ZYNQShim::xclLoadAxlf(const axlf *buffer)
...
 	}
 
-	auto runtime_pr_en = xrt_core::config::get_pr_enable();
-	if (runtime_pr_en == true) {
+  // If platform is a non-PR-platform, Following check will fail. Dont download the partial bitstream
+  // if Platform is a PR-platform, Following check passes as enable_pr value is true by default. Download the partial bitstream
+  // If platform is a PR-platform, but v++ generated a full bitstream (using some v++ param). User need to add enable_pr=false in xrt.ini.
+
+  auto is_pr_platform = (buffer->m_header.m_mode == XCLBIN_PR ) ? true : false;
+	auto runtime_pr_en = xrt_core::config::get_enable_pr(); //default value is true
```
mamin506: Indentation

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -471,8 +471,14 @@ int ZYNQShim::xclLoadAxlf(const axlf *buffer)
...
+	
+  if (is_pr_platform && runtime_pr_en) {
```
mamin506: Indentation

chvamshi-xilinx: > 
> 
> Indentation for this function.

Hi Min,
Fixed the indentation of whole function. This is now similar to rest of the functions

PR #1849, opened 2019-08-20
Update build file with additional fields 

abalasa:


abalasa:
@NaAl-Xlnx please review

PR #1850, opened 2019-08-21
Lockdown Xclbin before running test

chienwei-lan:


src/runtime_src/core/pcie/tools/xbutil/xbutil.h
```diff
@@ -148,6 +148,53 @@ static const std::map<MEM_TYPE, std::string> memtype_map = {
 
 static const std::map<std::string, command> commandTable(map_pairs, map_pairs + sizeof(map_pairs) / sizeof(map_pairs[0]));
 
+class xclbin {
+public:
+        xclbin(unsigned index, unsigned cu_idx)
```
stsoe: Please specify type of unsigned

src/runtime_src/core/pcie/tools/xbutil/xbutil.h
```diff
@@ -148,6 +148,53 @@ static const std::map<MEM_TYPE, std::string> memtype_map = {
...
+        try 
+        {
+
+            m_idx = index;
+            m_cuidx = cu_idx;
+            m_handle = nullptr;
+
+            std::string devstr = "device[" + std::to_string(m_idx) + "]";
+            m_handle = xclOpen(m_idx, NULL, XCL_QUIET);
```
stsoe: nullptr in C++ always

src/runtime_src/core/pcie/tools/xbutil/xbutil.h
```diff
@@ -148,6 +148,53 @@ static const std::map<MEM_TYPE, std::string> memtype_map = {
 
 static const std::map<std::string, command> commandTable(map_pairs, map_pairs + sizeof(map_pairs) / sizeof(map_pairs[0]));
 
+class xclbin {
+public:
+        xclbin(unsigned index, unsigned cu_idx)
+        try 
+        {
+
+            m_idx = index;
```
stsoe: These data members should be initialized in constructor initialization list.

src/runtime_src/core/pcie/tools/xbutil/xbutil.h
```diff
@@ -148,6 +148,53 @@ static const std::map<MEM_TYPE, std::string> memtype_map = {
 
 static const std::map<std::string, command> commandTable(map_pairs, map_pairs + sizeof(map_pairs) / sizeof(map_pairs[0]));
 
+class xclbin {
```
stsoe: This class looks like a mix of responsibilities.  It is better to separate the device concept from the xclbin concept.  Have  a device class that opens the device and manages this.  Then have this xclbin class get the uuid and open and manage a context on the xclbin itself.  

The xclbin class constructor should take a device object as argument, it looks like we already have such a class. It doesn't look like the cuidx is needed as argument since we default to virtual CU (-1).

src/runtime_src/core/pcie/tools/xbutil/xbutil.h
```diff
@@ -148,6 +148,53 @@ static const std::map<MEM_TYPE, std::string> memtype_map = {
...
+
+            if (uuid_is_null(uuid))
+                throw std::runtime_error("'uuid' invalid, please re-program xclbin.");
+
+            uuid_copy(m_uuid, uuid);
+
+            if (xclOpenContext(m_handle, m_uuid, m_cuidx, true))
+                throw std::runtime_error("ERROR: Unable to lockdown xclbin.");
+       
+        } catch (std::runtime_error) {
```
stsoe: Never catch by value.  Do "catch (const std::exception& ex) {" 
You should rethrow the exception, otherwise you end up with a bogus object.
There is nothing to close if handle or context wasn't successfully opened.   While you need to close the handle, it doesn't appear there can ever be a context to close.

With proper device object passed to constructor there is no need for try / catch in the xclbin ctor.

PR #1858, opened 2019-08-21
Minor change for log msg API

sarab96:
Minor change for log msg API
Note: adding more device info will be a separate PR


src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1574,13 +1574,26 @@ int xclLoadXclBin(xclDeviceHandle handle, const xclBin *buffer)
...
-    va_end(args);
+    static auto verbosity = xrt_core::config::get_verbosity();
+    if (level <= verbosity) {
+        va_list args;
+        va_start(args, format);
+        int ret = -1;
+        if (handle) {
+            xocl::shim *drv = xocl::shim::handleCheck(handle);
+            ret = drv ? drv->xclLogMsg(level, tag, format, args) : -ENODEV;
+            va_end(args);
```
maxzhen: va_end() and return are common for both cases, move them to outside of if statement?

sarab96: done

PR #1859, opened 2019-08-21
Enable zed platform in new json data format

abalasa:


abalasa:
@mamin506 @NaAl-Xlnx please review

PR #1864, opened 2019-08-22
handle userptr bad alloc exception by copy the user buffer to xrt buffer

mamin506:
1. If XRT is not able to allocate a buffer for the host pointer, copy user buffer to a regular XRT buffer.
2. If 1) happen, OpenCL would warn user and lead them to dmesg for the root cause.
3. This is safe, since we did the same for non aligned host pointer.


maxzhen:
retest this please

PR #1871, opened 2019-08-23
mpd azure plugin

xuhz:
Build & Test:

Built successfully on both ubuntu and centos.
Tested on ubuntu VM on ZT02

Note:
No azure pkg is going to be created for this branch. The plugin is also going into xrt package.
Push to the master will create azure pkg.


xuhz:
I missed changes of 2 files. I am going to close this request and open a new one.

PR #1873, opened 2019-08-23
added check for env variable for sw emu

vboggara-xilinx:
added check for env variable for sw emu


chvamshi-xilinx:
retest this please

PR #1877, opened 2019-08-23
Accept cu name also in addition to cu index

sarab96:
Accept cu name also in addition to cu index


NaAl-Xlnx:
retest this please

PR #1879, opened 2019-08-23
Fixing an issue where iterating through a binary file was incorrect

jvillarre:
Fixing an issue where iterating through a binary file was incorrect due to incorrect pointer arithmetic


NaAl-Xlnx:
retest this please

PR #1882, opened 2019-08-23
Version cmp for xbmgmt and xrt

AShivangi:


src/runtime_src/core/pcie/tools/xbmgmt/cmd_version.cpp
```diff
@@ -42,5 +42,10 @@ int versionHandler(int argc, char *argv[])
     xrt::version::print(std::cout);
     std::cout.width(26); std::cout << std::internal << "XOCL: " << driver_version("xocl") << std:: endl;
     std::cout.width(26); std::cout << std::internal << "XCLMGMT: " << driver_version("xclmgmt") << std::endl;    
+
+    if ( !getenv_or_null("INTERNAL_BUILD") ) {
+        if ( xrt_xbmgmt_version_cmp() != 0 )
+        return 0;
```
maxzhen: no need to return 0 here since you'll return 0 anyway in the end.

PR #1889, opened 2019-08-24
skip 6 data in ACCEL adapter configure

mamin506:
OpenCL will skip the first 6 data. Follow the same rule.


src/runtime_src/core/edge/drm/zocl/zocl_cu.c
```diff
@@ -366,7 +366,7 @@ zocl_acc_configure(void *core, u32 *data, size_t sz, int type)
 	 * Same open issue like HLS adapter
 	 * Skip 6 data,this is how user layer construct the command.
```
keryell: Since you are here, it would worth to make the comment understandable...

PR #1890, opened 2019-08-24
For user host pointer

mamin506:
1. Remove aligned check in sync_to_ubuf() and sync_to_hbuf(). Either buffer is non-aligned or bad alloc by xclAllocUserPtrBO, sync buffer should happen. The ubuf and hbuf check is good enough.

2. If buffer is CL_MEM_WRITE_ONLY, we don't need to do memcpy() on this buffer. In this case, the only sync direction, which make sense, is from device to host. Since this buffer is wrote by kernel and read by host application.

### DIFF
--- a/src/runtime_src/xocl/core/device.cpp
+++ b/src/runtime_src/xocl/core/device.cpp
@@ -133,6 +133,9 @@ static void
 sync_to_ubuf(xocl::memory* buffer, size_t offset, size_t size,
              xrt::device* xdevice, const xrt::device::BufferObjectHandle& boh)
 {
+  if (!buffer->need_extra_sync())
+      return;
+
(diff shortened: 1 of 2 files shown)


mamin506:
> What's the meaning of set_extra_sync? Was the aligned check not enough?
In two situations, OpenCL runtime should do extra sync,
1. User pointer is not 4K aligned.
2. m_xdevice->alloc(sz,xrt::device::memoryDomain::XRT_DEVICE_RAM,memidx,host_ptr) throw bad_alloc exception.

The set_extra_sync() method just set a flag for the buffer object to let us know it needs an extra sync or memory copy at the synchronization point.

stsoe:
I understand the change now and the need for the extra sync flag.

PR #1894, opened 2019-08-26
enable timestamp for exec buf in kds

maxzhen:
Enable KDS to record timestamp of various state when exec buf is being executed. Useful for performance analysis.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
@@ -61,6 +61,7 @@
 #include <linux/list.h>
 #include <linux/eventfd.h>
 #include <linux/kthread.h>
+#include <linux/ktime.h>
 #include <ert.h>
 #include "../xocl_drv.h"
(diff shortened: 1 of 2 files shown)


maxzhen:
> 
> 
> Changes for zocl?

Will be done in a separate pull request, if DC change looks OK to everybody.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
```diff
@@ -614,6 +631,23 @@ cmd_abort(struct xocl_cmd *xcmd)
 	SCHED_DEBUGF("xcmd(%lu) [-> abort]\n", xcmd->uid);
 }
 
+static inline bool
+can_enable_timestamps(struct xocl_cmd *xcmd)
```
stsoe: Prefix with cmd_ for consistency; if this was c++ it would be a member function of a command class and I have tried to use that convention here for all object types.

maxzhen: Got it. Fixed now.

PR #1918, opened 2019-08-28
Zcu102 all h pports 28th aug

subhransu-xilinx:
This merge request enables all HP ports and profiling supports in the zcu102_base platform. Also removed some redundant commands from platform.

Please retain the intermediate commits.


subhransu-xilinx:
Withdrawing this to verify few platform properties before resubmitting.

PR #1935, opened 2019-08-30
IBM PowerPC(PPC64le) redhat uses 4.14 as of 7.5 and 7.6

xdavidz:
Please review the fix for building/installing XOCL on PowerPC redhat 7.6.

Do not merge for now, need to re-verify it after code review.


xdavidz:
verified. note: this is for 2019.1 only, for master branch, we are not able to verify fix due to the xocl installation will fail with python error on IBM's powerPC redhat. We will have to hold off till we have fully functional master branch code for IBM powerPC redhat. T

PR #1936, opened 2019-08-30
HAL level profiling with counters for PCIe and Zynq

IshitaGhosh:


src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1429,6 +1432,14 @@ int shim::xclReadTraceData(void* traceBuf, uint32_t traceBufSz, uint32_t numSamp
     return size;
 }
 
+
+#if 0
```
maxzhen: if this is really needed for future use, please create a proper macro name - #ifdef XXX. Put #if 0 in multiple files for related code is not that clean.

src/runtime_src/xrt/device/halops2.h
```diff
@@ -118,6 +118,9 @@ class operations
                                       bool shared);
   typedef int (* closeContextFuncType)(xclDeviceHandle handle, const uuid_t xclbinId, unsigned ipIndex);
 
+  // xdp hal plugin switch
+//  typedef int (* switchProfilingFuncType)(xclDeviceHandle handle);
```
maxzhen: remove?

src/runtime_src/core/pcie/linux/shim.cpp
IshitaGhosh: Done

src/runtime_src/xrt/device/halops2.h
IshitaGhosh: Done

PR #1940, opened 2019-08-30
Versal V350 polling mode support

xdavidz:
Versal V350 initial support, test report is here:
https://confluence.xilinx.com/display/~davidzha/v350+zocl+test+plan

### DIFF
--- a/src/platform/v350/system.dts
+++ b/src/platform/v350/system.dts
@@ -1,0 +1,568 @@
+/dts-v1/;
+
+/ {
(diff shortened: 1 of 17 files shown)


xdavidz:
Once Larry verified both with xocl code together, we can merge. Thus remove the label.

xdavidz:
> Please fix compile error for 32 bits system.

There are warnings in 32bits system (fixed), but build should pass if ert.h has been updated to the version in versal host PR. I just include the ert.h in this PR to make the build pass. Eventually, we might see a merge conflict if we include ert.h in both host and embedded PR for versal.

PR #1946, opened 2019-08-31
no support for 'xbmgmt flash scan'

maxzhen:


src/runtime_src/core/pcie/tools/xbmgmt/xbmgmt.cpp
```diff
@@ -41,6 +41,8 @@ static const std::map<std::string, struct subCmd> subCmdList = {
     { "--version", {versionHandler, subCmdVersionDesc, subCmdVersionUsage} },
     { "scan", {scanHandler, subCmdScanDesc, subCmdScanUsage} },
     { "flash", {flashHandler, subCmdFlashDesc, subCmdFlashUsage} },
+    // for xbutil flash
+    { "-lash", {flashXbutilFlashHandler, subCmdXbutilFlashDesc, subCmdXbutilFlashUsage} },
```
keryell: `lash`?
If this is really `lash`, please add a comment explaining why, otherwise other people will ask...

maxzhen: > 
> 
> `lash`?
> If this is really `lash`, please add a comment explaining why, otherwise other people will ask...

No, this is not 'lash', it's a hidden option for xbmgmt. Only xbutil will internally call this to handle "xbutil flash" sub-command. This is needed for backward compatibility. The hidden option will be removed after we remove "xbutil flash" eventually.

PR #1953, opened 2019-09-03
Fix for design intent 

subhransu-xilinx:
This fixes design intent for zcu102 and zcu104 base platforms so that the tools don't identify these platforms as data center platforms.


maxzhen:
ok to test

PR #1954, opened 2019-09-03
Updated ERT firmware to support Emulation shared library creation

hemantk-xilinx:
Hi 

This is a pull request to create a shared library for ERT code so that Emulation can use latest ERT firmware instead of keeping a copy of it in the Emulation model.

### DIFF
--- a/src/runtime_src/ert/scheduler/scheduler.cpp
+++ b/src/runtime_src/ert/scheduler/scheduler.cpp
@@ -18,11 +18,7 @@
  * Embedded runtime scheduler
  */
 
-#ifndef ERT_HW_EMU
 #include "core/include/ert.h"
-#else
-#include "ert.h"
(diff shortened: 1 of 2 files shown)


maxzhen:
ok to test

hemantk-xilinx:
Hi Soren,
Can you please review this?

src/runtime_src/ert/scheduler/scheduler.cpp
```diff
@@ -31,8 +27,10 @@
 #else
 #define xil_printf printf
 #define print printf
+#define u32 uint32_t
 #endif
 #include <stdlib.h>
+#include <stdio.h>
```
stsoe: Let's move to ERT_HW_EMU, since firmware itself uses xil_printf

PR #1955, opened 2019-09-03
Live messages fixes

sahilg-xilinx:
1. Disabled old messages if 'system_dpa' is enabled
2. Method to calculate time changed from clock() to chrono
3. Truncating the extra characters in messages


maxzhen:
ok to test

PR #1971, opened 2019-09-05
Zynq compilation issue

chvamshi-xilinx:
Hi,

This change contains a fix for 32bit system compilation issue along with crash when ert-thread started when ERT is not enabled.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/sched_exec.c
+++ b/src/runtime_src/core/edge/drm/zocl/sched_exec.c
@@ -690,8 +690,7 @@ configure(struct sched_cmd *cmd)
 	}
 	write_unlock(&zdev->attr_rwlock);
 
-	if(zdev->ert)
-	{
+	if(zdev->ert) {
 	  /* Enable interrupt from host to PS when new commands are ready */
 	  if (exec->cq_interrupt) {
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -690,21 +690,24 @@ configure(struct sched_cmd *cmd)
...
-		/* At this point we are good. No one is polling CQ */
-		cq_irq = zdev->ert->irq[ERT_CQ_IRQ];
-		ret = request_irq(cq_irq, sched_cq_isr, 0, "zocl_cq", zdev);
-		if (ret) {
-			DRM_WARN("Failed to initial CQ interrupt. "
-				 "Fall back to polling\n");
-			exec->cq_interrupt = 0;
-			exec->cq_thread = kthread_run(cq_check, zdev, name);
-		}
+	if(zdev->ert)
```
xdavidz: Thanks for fixing this, could we do it a little bit neat with other coding style, like
if (zdev->ert) {
    ...
}

chvamshi-xilinx: Thanks for pointing this out. Fixed it.

PR #1973, opened 2019-09-05
Add filtering metadata for new infra 

haeseunglee:
New infra requires 'regression' metadata for filtering.
So, updated all the testinfo.yml files to have the filtering metadata


maxzhen:
ok to test

PR #1974, opened 2019-09-06
account for non-linearity in clock training

hackwa:


src/runtime_src/xdp/profile/device/traceS2MM.cpp
```diff
@@ -195,14 +195,16 @@ void TraceS2MM::parseTraceBuf(void* buf, uint64_t size, xclTraceResultsVector& t
       count = MAX_TRACE_NUMBER_SAMPLES;
     }
     auto pos = static_cast<uint64_t*>(buf);
+    // Each packet is Always writtn in 4 chunks
```
pgschuey: "Always writtn" should be "always written"

hackwa:
This seems to be happening in main thread. Adding sleep to main thread adds some delay. I'll move clock training to a new thread. That'll make it possible for us to do training in background with a small chance of race condition in drift calculations. Race condition will not occur as long as we don't do a lot of sleeping in clock training.

hackwa:
Will reopen with hw changes in place

PR #1983, opened 2019-09-07
remove unharmful ERROR

mamin506:
Remove some unharmful and misleading ERROR messages.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_drv.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_drv.c
@@ -653,19 +653,17 @@ static int zocl_drm_platform_probe(struct platform_device *pdev)
 		zdev->ert = (struct zocl_ert_dev *)platform_get_drvdata(subdev);
 	}
 
+	/* For Non PR platform, there is not need to have FPGA manager
+	 * For PR platform, the FPGA manager is required. No good way to
+	 * determin if it is a PR platform at probe.
+	 */
 	fnode = of_get_child_by_name(of_root,
 	    zdev->zdev_data_info->fpga_driver_name);
 	if (fnode) {
 		zdev->fpga_mgr = of_fpga_mgr_get(fnode);
-		if (IS_ERR(zdev->fpga_mgr)) {
-			DRM_ERROR("FPGA Manager not found %ld\n",
-			    PTR_ERR(zdev->fpga_mgr));
+		if (IS_ERR(zdev->fpga_mgr))
 			zdev->fpga_mgr = NULL;
-		}
 		DRM_INFO("FPGA programming device %s founded.\n",
-		    zdev->zdev_data_info->fpga_driver_name);
-	} else {
-		DRM_ERROR("FPGA programming device %s not found\n",
 		    zdev->zdev_data_info->fpga_driver_name);
 	}
 
--- a/src/runtime_src/core/edge/drm/zocl/zocl_ioctl.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_ioctl.c
@@ -199,10 +199,17 @@ zocl_fpga_mgr_load(struct drm_device *ddev, char *data, int size)
 {
 	struct drm_zocl_dev *zdev = ddev->dev_private;
 	struct device *dev = ddev->dev;
-	/* zocl_pcap_download_ioctl() has checked if fpga_mgr is NULL */
 	struct fpga_manager *fpga_mgr = zdev->fpga_mgr;
 	struct fpga_image_info *info;
 	int err = 0;
+
+	 /* On Non PR platform, it shouldn't never go to this point.
+	  * On PR platform, the fpga_mgr should be alive.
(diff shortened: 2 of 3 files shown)


src/runtime_src/core/edge/drm/zocl/zocl_drv.c
```diff
@@ -653,20 +653,18 @@ static int zocl_drm_platform_probe(struct platform_device *pdev)
...
 			zdev->fpga_mgr = NULL;
-		}
 		DRM_INFO("FPGA programming device %s founded.\n",
 		    zdev->zdev_data_info->fpga_driver_name);
-	} else {
-		DRM_ERROR("FPGA programming device %s not found\n",
-		    zdev->zdev_data_info->fpga_driver_name);
 	}
 
 	if (of_property_read_u32(pdev->dev.of_node, "xlnx,pr-isolation-addr",
```
xdavidz: Is this the way to determine if it is a PR platform? For non-PR platform, if we still have fpga-mgr configured in device-tree, it seems an error too. What if we move this line above the fpga-mgr code, and print errors for PR platform only?

src/runtime_src/core/edge/drm/zocl/zocl_ioctl.c
```diff
@@ -318,6 +319,14 @@ int zocl_pcap_download_ioctl(struct drm_device *dev, void *data,
 	uint64_t primary_fw_len;
 	struct axlf_section_header *primaryHeader;
 
+	 /* On Non PR platform, it shouldn't never go to this point.
+	  * On PR platform, the fpga_mgr should be alive.
+	 */
+	if (!zdev->fpga_mgr) {
```
xdavidz: It is called from 3 places now zocl_pcap_download, zocl_load_pdi and zocl_read_axlf_ioctl.

src/runtime_src/core/edge/drm/zocl/zocl_ioctl.c
```diff
@@ -199,6 +199,7 @@ zocl_fpga_mgr_load(struct drm_device *ddev, char *data, int size)
 {
 	struct drm_zocl_dev *zdev = ddev->dev_private;
 	struct device *dev = ddev->dev;
+	/* zocl_pcap_download_ioctl() has checked if fpga_mgr is NULL */
```
larry9523: Looks like the better place to check if fpga_mgr is NULL is inside this function, because except pcap_download, pdi download also make use of this function

src/runtime_src/core/edge/drm/zocl/zocl_ioctl.c
```diff
@@ -318,6 +319,14 @@ int zocl_pcap_download_ioctl(struct drm_device *dev, void *data,
 	uint64_t primary_fw_len;
 	struct axlf_section_header *primaryHeader;
 
+	 /* On Non PR platform, it shouldn't never go to this point.
+	  * On PR platform, the fpga_mgr should be alive.
+	 */
+	if (!zdev->fpga_mgr) {
+		DRM_ERROR("FPGA manager is not found.\n");
+		return -EFAULT;
```
larry9523: ENXIO would be more proper here

src/runtime_src/core/edge/drm/zocl/zocl_drv.c
mamin506: For non-PR platform, the FPGA manager is not used by us. But it could be used for other purpose, such as reprogram the whole PL. In this case, it is not an error. If we have solid way to know this is a PR platform, we could print error if fgpa-mgr is not found.

src/runtime_src/core/edge/drm/zocl/zocl_ioctl.c
mamin506: I will add this check for all of them. Thanks

src/runtime_src/core/edge/drm/zocl/zocl_ioctl.c
mamin506: Will do.

src/runtime_src/core/edge/drm/zocl/zocl_ioctl.c
xdavidz: LGTM so far

src/runtime_src/core/edge/drm/zocl/zocl_drv.c
xdavidz: ok

mamin506:
> According to me, Checking fpga_mgr in three functions is still fine. We should error out ASAP after entering into function.

Yes. If something is obviously wrong, zocl should error out ASAP. I think the best place is probe().
For these check, let's only check it in the function who used it.

src/runtime_src/core/edge/drm/zocl/zocl_ioctl.c
mamin506: Done

PR #1984, opened 2019-09-07
fix multi-thread application

maxzhen:
Mistakenly forced all threads to use device handle in pci_device. It's now fixed.
Pass in xclbin_id to exec_reset() so that it does not need to call into icap to get it. This may avoid dead lock when icap is calling exec_reset(). @stsoe @larry9523 , please take a look.
Touched cloud daemon and NIFD due to the API change in pci_device, @xuhz @jvillarre , please take a look.

Tested with a process with 200 threads, each doing xclOpen/xclOpenContext/xclExecBuf/xclExecWait. It's working now.


maxzhen:
retest this please

PR #1985, opened 2019-09-07
Partial Reconguration Issue

chvamshi-xilinx:
Hi ,
This change contains resetting the configured flag after programming the new bitstream


src/runtime_src/core/edge/drm/zocl/zocl_ioctl.c
```diff
@@ -292,6 +292,7 @@ static int zocl_pcap_download(struct drm_zocl_dev *zdev,
 
 	if (zdev->pr_isolation_addr && !IS_ERR_OR_NULL(map))
 		iowrite32(0x3, map);
+	zdev->exec->configured = 0;
```
larry9523: We should really be cautious with allowing reconfigure (download another xlcbin) under multithread situation. The safest way to do is clear this flag after we have full context support on edge

larry9523:
As discussed, we can make sure single thread (and not break multiple threads with same xclbin) works. Even with single thread, we need to take care of more places than just clear this configured flag. For example, in zocl_read_axlf_ioctl, it assumes all those zdev->ip, zdev->debug_ip, zdev->connectivity, etc were NULL. We need to free them before give them new address. Otherwise, we will generate memory leak.

src/runtime_src/core/edge/drm/zocl/zocl_ioctl.c
xdavidz: > As discussed, we can make sure single thread (and not break multiple threads with same xclbin) works. Even with single thread, we need to take care of more places than just clear this configured flag. For example, in zocl_read_axlf_ioctl, it assumes all those zdev->ip, zdev->debug_ip, zdev->connectivity, etc were NULL. We need to free them before give them new address. Otherwise, we will generate memory leak.

Yes, enabling multiple configuration will need to consider recycle memory and status(basically reset to initialized status for next new xclbin). As of today, the zocl_read_axlf_ioctl only allocate memory but never free the memory for next xclbin. Also, some other status should be well taken care of.

xdavidz:
> As discussed, we can make sure single thread (and not break multiple threads with same xclbin) works. Even with single thread, we need to take care of more places than just clear this configured flag. For example, in zocl_read_axlf_ioctl, it assumes all those zdev->ip, zdev->debug_ip, zdev->connectivity, etc were NULL. We need to free them before give them new address. Otherwise, we will generate memory leak.

Just had a con-call with Vamshi, here are the AIs:
1) David will create a confluence page for this project, list test cases and designs; Everyone is welcome to contribute.
2) Vamshi will re-think single thread, single xclbin at one time case and make sure no painc/memory/other outstanding bugs after enabling re-configuration.
3) More complex multi-threading and context switch will be done as further phases.

PR #1999, opened 2019-09-09
Default outputtype

subhransu-xilinx:
Adds fixed clocks, defaults type set to xclbin and removes redundant profile changes from metadata


maxzhen:
ok to test

PR #2018, opened 2019-09-10
add polling throttle option to contorl sws

mamin506:
One customer is using polling_throttle option since 2017.2. Just take it back here.
This only impact 'sws' but has no effect on 'kds'.

### DIFF
--- a/src/runtime_src/xrt/scheduler/sws.cpp
+++ b/src/runtime_src/xrt/scheduler/sws.cpp
@@ -769,6 +769,11 @@ class xocl_scheduler
   {
     wait();
     queue_cmds();
+
+    // throttle polling for cu completion
+    if (auto us = xrt::config::get_polling_throttle())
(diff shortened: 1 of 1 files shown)


src/runtime_src/xrt/scheduler/sws.cpp
```diff
@@ -769,6 +769,11 @@ class xocl_scheduler
   {
     wait();
     queue_cmds();
+
+    // throttle polling for cu completion
+    if (auto us = xrt::config::get_polling_throttle())
```
stsoe: This throttles not only polling but also starting of CUs.
Move this code into the wait() function.
Think about changing it to be more intelligent, e.g. only sleep if no new commands have been added or no running commands have completed since last iteration.

mamin506: Thanks for the comment. I will provide another commit in this PR.

PR #2028, opened 2019-09-11
Change JSON writing to use a boost::property_tree::ptree::path_type t…

emenchen:
…o prevent keys from creating arbitrary JSON hierarchy. Also fix a typo in a key.
(Stephen Rozum has reviewed this code change.)


maxzhen:
ok to test

PR #2029, opened 2019-09-11
set flags for import buffer

mamin506:
The flag for import DMABUF buffer was not been set.

drm_gem_prime_fd_to_handle() will call gem_prime_import() to import the DMABUF fd.
This is the best time to set flags for import buffer.

BTW,  drm_gem_prime_import_dev() will call gem_prime_import_sg_table(), which points to drm_gem_cma_prime_import_sg_table(). This function will check if buffer is physical contiguous and create CMA buffer object for the imported buffer.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
@@ -945,3 +945,19 @@ void zocl_clear_mem(struct drm_zocl_dev *zdev)
 
 	mutex_unlock(&zdev->mm_lock);
 }
+
+struct drm_gem_object *zocl_gem_import(struct drm_device *dev,
+				       struct dma_buf *dma_buf)
+{
+	struct drm_gem_object *gem_obj;
+	struct drm_zocl_bo *zocl_bo;
+
+	gem_obj = drm_gem_prime_import_dev(dev, dma_buf, dev->dev);
+	zocl_bo = to_zocl_bo(gem_obj);
+	/* gem_prime_import_sg_table = drm_gem_cma_prime_import_sg_table
+	 * The import buffer is a CMA buffer.
+	 */
+	zocl_bo->flags |= ZOCL_BO_FLAGS_CMA;
+
+	return gem_obj;
+}
--- a/src/runtime_src/core/edge/drm/zocl/zocl_bo.h
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_bo.h
@@ -75,4 +75,7 @@ static inline uint32_t zocl_convert_bo_uflags(uint32_t uflags)
 	return zflags;
 }
 
+struct drm_gem_object *
+zocl_gem_import(struct drm_device *dev, struct dma_buf *dma_buf);
+
 #endif /* _ZOCL_BO_H */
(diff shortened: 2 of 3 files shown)


mamin506:
Fix CR-1039670

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -945,3 +945,19 @@ void zocl_clear_mem(struct drm_zocl_dev *zdev)
...
 	mutex_unlock(&zdev->mm_lock);
 }
+
+struct drm_gem_object *zocl_gem_import(struct drm_device *dev,
+				       struct dma_buf *dma_buf)
+{
+	struct drm_gem_object *gem_obj;
+	struct drm_zocl_bo *zocl_bo;
+
+	gem_obj = drm_gem_prime_import_dev(dev, dma_buf, dev->dev);
```
larry9523: any particular reason we don't want to use drm_gem_prime_import ?
and I think we'd better check the returned gem_obj, in the case of PTR_ERR(obj), we don't want to keep calling to_zocl_bo and set the flags field.

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -945,3 +945,19 @@ void zocl_clear_mem(struct drm_zocl_dev *zdev)
...
+	zocl_bo = to_zocl_bo(gem_obj);
+	/* gem_prime_import_sg_table = drm_gem_cma_prime_import_sg_table
```
larry9523: Could you please explain what these comments are about and why they are put here?

mamin506: The comment is trying to hint that drm_gem_cma_prime_import_sg_table() make sure the importing buffer is CMA buffer.
drm_gem_cma_prime_import_sg_table() actually check if buffer is physical contiguous.

larry9523: > The comment is trying to hint that drm_gem_cma_prime_import_sg_table() make sure the importing buffer is CMA buffer.
> drm_gem_cma_prime_import_sg_table() actually check if buffer is physical contiguous.

so maybe just comment explicitly instead of hint? something like DRM will check if buffer is physically contiguous?

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
mamin506: You are right. We should use drm_gem_prime_import(), since we didn't want to attach dma-buf to another device. 
We should check PTR_ERR.

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -945,3 +945,22 @@ void zocl_clear_mem(struct drm_zocl_dev *zdev)
...
+	struct drm_zocl_bo *zocl_bo;
+
+	gem_obj = drm_gem_prime_import(dev, dma_buf);
+	if (!IS_ERR(gem_obj)) {
+		zocl_bo = to_zocl_bo(gem_obj);
+		/* drm_gem_cma_prime_import_sg_table() is used for hook
+		 * gem_prime_import_sg_table. It will check if import buffer
+		 * is a CMA buffer and create CMA object.
+		 */
+		zocl_bo->flags |= ZOCL_BO_FLAGS_CMA;
```
maxzhen: Does this mean that we can only import CMA BO on edge?

PR #2036, opened 2019-09-12
add docs for cloud vendor support

xuhz:


xuhz:
remove the "azure" and "aws" part in the doc at current stage. Will add it back after 2019.2 official release and support for azure and aws is official announced

PR #2037, opened 2019-09-12
Get compute unit status from sysfs

stsoe:
The backdoor sysfs execution of ctrl commands was unsafe and is
removed in this PR. The sysfs node could add ctrl cmd to kds, but
regular flow could immediately unload xclbin and call exec_stop which
in turn would prevent scheduler from draining the outstanding ctrl
command (see #1705).

Now scheduler stats is updated explicitly through xbutil top and query
through regular command execution.

Fixes CR: 1027416


stsoe:
retest this please

PR #2041, opened 2019-09-13
xbutil mem read/write for edge

Himanshu-xilinx:
* Kernel should have CONFIG_STRICT_DEVMEM disabled (CONFIG_STRICT_DEVMEM=n)
* Does not consider PL DDR


chvamshi-xilinx:
@Sonal, I assume this implementation is same as devmem. Can you please suggest if there is a way to write to specific address.

chvamshi-xilinx:
Closing as this is not going to be supported in 2019.2

PR #2042, opened 2019-09-13
zcu102/104_base: Use same dynamic.tcl for both emu and hw

subhransu-xilinx:
Signed-off-by: Subhransu S. Prusty <subhransu.sekhar.prusty@xilinx.com>


maxzhen:
ok to test

PR #2056, opened 2019-09-14
Fixing xilinx.icd issue for edge

chvamshi-xilinx:
Hi,

This change contains a fix to create a icd file as part of the rootfs. 
Added in xrt_git.bb as icd creation should be part of libxilinxopencl.so recipe


chvamshi-xilinx:
retest this please

PR #2063, opened 2019-09-16
Renamed the SDX ENVs to new tool name ENVs to resolve the CR-1041912

venkatp-xilinx:
In process of Unified Naming PR, we modified all the SDX ENVs to new tool name ENVs in the RDI code base, and hence we have done respective changes even in the XRT code base


chvamshi-xilinx:
retest this please

PR #2066, opened 2019-09-16
added partial support for TRACE_S2MM_FULL (AIE flows only)

pgschuey:


src/runtime_src/core/include/xclbin.h
```diff
@@ -313,6 +313,7 @@ extern "C" {
         AXI_STREAM_MONITOR,
         AXI_STREAM_PROTOCOL_CHECKER,
         TRACE_S2MM,
+        TRACE_S2MM_FULL,
```
rozumx: Being that the enumeration values are inserted into the xclbin archive, shouldn't TRACE_S2MM_FULL be added to the end of the enumeration list.  

Currently how the code is implemented, past AXI_DMA enumeration values contained in a previous xclbin archive are now mapped onto TRACE_S2MM_FULL, which I don't believe is the desired behavior.

pgschuey: moved new enum value to end as requested

pgschuey:
Please retest

NaAl-Xlnx:
retest this please

PR #2071, opened 2019-09-17
removed dependency on device path Env variable for sw emu

vboggara-xilinx:
removed dependency on device path Env variable(XILINX_PCIE_MODEL) for sw emu


NaAl-Xlnx:
retest this please

PR #2077, opened 2019-09-17
Use XRT_INSTALL_INCLUDE_DIR for header files

mamin506:
.


maxzhen:
@stsoe , could you take a look?

PR #2085, opened 2019-09-19
fixing an issue of p2p buffer. Shouldnt call migrate buffer for p2p

saifuddin-xilinx:
There is a check added recently in migrate_buffer to errorout if p2p buffer is getting migrated. This makes sense of host is using this. But EnqueuNdRangeKernel also migrates all the READ_ONLY/READ_WRITE buffers. We should not call migrate from this function aswell


chvamshi-xilinx:
ok to test

chvamshi-xilinx:
retest this please

chvamshi-xilinx:
retest this please

chvamshi-xilinx:
retest this please

stsoe:
retest this please

PR #2086, opened 2019-09-19
zcu102/104_base: Add interface and support for profiling 

subhransu-xilinx:


maxzhen:
ok to test

subhransu-xilinx:
Not required. Merged with different pull requests.

PR #2092, opened 2019-09-19
Display stderr and fix env vars for xbutil validate

AShivangi:
- Fixed stderr pipe, now displays the stderr when the test fails
 - setting environment variables for validating from XILINX_XRT path 
 - tested on u200, u250, u280, u50


src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -918,6 +918,17 @@ void testCaseProgressReporter(bool *quit)
     }
 }
 
+void set_shell_env(std::string var_name, std::string trailing_path, int overwrite)
```
maxzhen: this only works for "path"-like env, so probably call it set_shell_path_env()?

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -938,13 +951,13 @@ int runShellCmd(const std::string& cmd, std::string& output)
         ret = -EINVAL;
     }
 
-    close(stderr_fds[0]);
     dup2(stderr_fds[1], 2);
     std::shared_ptr<FILE> pipe(popen(cmd.c_str(), "r"), pclose);
-    close(stderr_fds[1]);
+    close(stderr_fds[1]);
```
maxzhen: extra space in the end.

keryell: Perhaps it would make sense to use something like a Python-esque https://www.boost.org/doc/libs/1_71_0/doc/html/boost_process/tutorial.html or https://www.boost.org/doc/libs/1_71_0/doc/html/interprocess.html  instead of a low-level C interface from the 70's impossible to debug...

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
stsoe: Never pass std::string by value unless you truly need a copy.

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -918,6 +918,17 @@ void testCaseProgressReporter(bool *quit)
...
+{
+    std::string xrt_path(getenv("XILINX_XRT"));
+    std::string new_path;
+    if (getenv(var_name.c_str()) != NULL)
+        new_path = std::string(getenv(var_name.c_str()));
+    xrt_path += trailing_path + ":";
+    new_path = xrt_path + new_path;
+    setenv(var_name.c_str(), new_path.c_str(), overwrite);
+}
+
```
keryell: What about https://www.boost.org/doc/libs/1_71_0/doc/html/boost_process/tutorial.html#boost_process.tutorial.env ?

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -918,6 +918,17 @@ void testCaseProgressReporter(bool *quit)
     }
 }
 
+void set_shell_path_env(const std::string& var_name, const std::string& trailing_path, int overwrite)
+{
+    std::string xrt_path(getenv("XILINX_XRT"));
```
stsoe: This is not safe.  getenv can return nullptr.

I strongly recommend this simple helper function:
```
const char* value_or_empty(const char* value) { return value ? value : "" ; }
std::string get_env(const char* env) { return value_or_empty(std::getenv(env)); }
```
now you can get env without explicitly checking the return value of std::getenv.
`std::string xrt_path(get_env("XILINX_XRT"));`

Also, xrt_path doesn't change per call, so make it static within function so as to compute just once.

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -928,8 +939,10 @@ int runShellCmd(const std::string& cmd, std::string& output)
 
     // Run test case
     setenv("XILINX_XRT", "/opt/xilinx/xrt", 0);
-    setenv("PYTHONPATH", "/opt/xilinx/xrt/python", 0);
-    setenv("LD_LIBRARY_PATH", "/opt/xilinx/xrt/lib", 1);
+    set_shell_path_env(std::string("PYTHONPATH"), std::string("/python"), 0);
```
stsoe: remote std::string around the string constants, not needed.

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -938,13 +951,13 @@ int runShellCmd(const std::string& cmd, std::string& output)
...
     }
 
-    close(stderr_fds[0]);
     dup2(stderr_fds[1], 2);
     std::shared_ptr<FILE> pipe(popen(cmd.c_str(), "r"), pclose);
     close(stderr_fds[1]);
 
     if (pipe == nullptr) {
         std::cout << "ERROR: Failed to run " << cmd << std::endl;
+        close(stderr_fds[0]);
```
stsoe: prone to error.  fds are resources and should be treated as such, use RAII via a smart ptr so close happens automatically at scope exit.

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -918,6 +918,17 @@ void testCaseProgressReporter(bool *quit)
     }
 }
 
+const char* value_or_empty(const char* value) { return value ? value : "" ; }
```
stsoe: inline

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -918,6 +918,17 @@ void testCaseProgressReporter(bool *quit)
...
+
+void set_shell_path_env(const std::string& var_name, const std::string& trailing_path, int overwrite)
```
stsoe: static?

PR #2096, opened 2019-09-19
wireserver product version doesn't support // in uri

xuhz:
Microsoft wireserver product version doesn't allow "//" in uri, although both the standalone version and RFC allow.

There is no difference to the standalone wireserver with or without the fix. The fix has been verified by Microsoft.


xuhz:
build failed?

The changes are just the content of the string, it is impossible to cause build failure.

maxzhen:
pipeline probably is not set up properly for 2019.2_XDF branch.

PR #2102, opened 2019-09-20
Clarified Reset support functionality through comments and correct return type description of xclUnmgdPread and xclUnmgdPwrite

uday610:
Clarified Reset support functionality through comments


maxzhen:
ok to test

PR #2104, opened 2019-09-20
Fix to check if TRACE_HUB is available in System_dpa

sahilg-xilinx:
If not available, avoid reading live messages as it is leading to random
value and crash in few scenarios


maxzhen:
@chvamshi-xilinx , this pull request is merged without going through pipeline?

chvamshi-xilinx:
@maxzhen , **It was my mistake.** I have reviewed the code and merged as it is blocking the daily builds to be created. Just checked the "check passed" string and merged it.
I will be more careful from the next time.

PR #2123, opened 2019-09-23
Metadata update to exclude versal device for unsupported tests

haeseunglee:
DDR banks won't work with versal. So, update metadata to exclude unsupported tests


maxzhen:
ok to test

maxzhen:
retest this please

PR #2124, opened 2019-09-23
[Edge] kds state timestamp

mamin506:
The same solution like PCIe platform.

The edge zocl driver recorded timestamps of CU start and done in the first 4 data of regmap. Keep this approach for one more release.


mamin506:
> One question (very trivial though), do we plan to remove the set_cmd_ext_timestamp later? When ERT command says don't record timestamps, we may want to honour that.

Yes, set_cmd_ext_timestamp() is obsoleted. Let's remove it in the next release.

PR #2127, opened 2019-09-23
coverity fixes and modify debug message level

chienwei-lan:


maxzhen:
please address conflict.

PR #2131, opened 2019-09-24
Fix dynamic postlink for multiple kernels and clock constraint added to both the kernel clocks

subhransu-xilinx:


chvamshi-xilinx:
ok to test

PR #2133, opened 2019-09-24
sysfs_get API change

AShivangi:
- remove sysfs_get for bool
- use template for getting int and bool
- explicitly specify the default return type


src/runtime_src/core/pcie/linux/scan.cpp
```diff
@@ -353,7 +340,7 @@ pcidev::pci_device::pci_device(const std::string& sysfs) : sysfs_name(sysfs)
     }
 
     // Determine if device is of supported vendor
-    sysfs_get("", "vendor", err, vendor);
+    sysfs_get("", "vendor", err, vendor, -1);
```
maxzhen: not specifying type here?

src/runtime_src/core/pcie/linux/scan.h
```diff
@@ -61,20 +61,22 @@ class pci_device {
...
     void sysfs_get(const std::string& subdev, const std::string& entry,
         std::string& err_msg, std::string& s);
-    void sysfs_get(const std::string& subdev, const std::string& entry,
-        std::string& err_msg, bool& b);
     void sysfs_get(const std::string& subdev, const std::string& entry,
         std::string& err_msg, std::vector<char>& buf);
     template <typename T>
     void sysfs_get(const std::string& subdev, const std::string& entry,
-        std::string& err_msg, T& i) {
+        std::string& err_msg, T& i, const int& default_val) {
```
maxzhen: should we do const T& default_val?

src/runtime_src/core/pcie/linux/scan.h
```diff
@@ -61,20 +61,22 @@ class pci_device {
...
         std::vector<uint64_t> iv;
 
         sysfs_get(subdev, entry, err_msg, iv);
         if (!iv.empty())
             i = static_cast<T>(iv[0]);
         else
-            i = static_cast<T>(-1); // default value
+            i = static_cast<T>(default_val); // default value
+    }
+    void sysfs_get_sensor(const std::string& subdev, const std::string& entry,
```
maxzhen: it seems that caller don't need to pass in err_msg, don't care about error for sensor readings. if we can't get it, we print out N/A
BTW, is -1 a good default value for sensor readings? should we use 0? we should print out N/A when we get default value.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -555,59 +555,59 @@ void shim::xclSysfsGetDeviceInfo(xclDeviceInfo2 *info)
     std::string s;
     std::string errmsg;
 
-    mDev->sysfs_get("", "vendor", errmsg, info->mVendorId);
-    mDev->sysfs_get("", "device", errmsg, info->mDeviceId);
-    mDev->sysfs_get("", "subsystem_device", errmsg, info->mSubsystemId);
+    mDev->sysfs_get("", "vendor", errmsg, info->mVendorId, -1);
```
maxzhen: some of these calls does not provide type?

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -555,59 +555,59 @@ void shim::xclSysfsGetDeviceInfo(xclDeviceInfo2 *info)
...
-    mDev->sysfs_get("xmc", "xmc_vccint_vol", errmsg, info->mVccIntVol);
-    mDev->sysfs_get("xmc", "xmc_fpga_temp", errmsg, info->mOnChipTemp);
-
-    mDev->sysfs_get("", "link_width", errmsg, info->mPCIeLinkWidth);
-    mDev->sysfs_get("", "link_speed", errmsg, info->mPCIeLinkSpeed);
-    mDev->sysfs_get("", "link_speed_max", errmsg, info->mPCIeLinkSpeedMax);
-    mDev->sysfs_get("", "link_width_max", errmsg, info->mPCIeLinkWidthMax);
-    mDev->sysfs_get("", "mig_calibration", errmsg, info->mMigCalib);
+    mDev->sysfs_get("mb_scheduler", "kds_numcdmas", errmsg, info->mNumCDMA, -1);
+    mDev->sysfs_get("xmc", "xmc_12v_pex_vol", errmsg, info->m12VPex, -1);
```
maxzhen: should we use sysfs_get_sensors() for all these sensor reading?

src/runtime_src/core/pcie/tools/cloud-daemon/pciefunc.cpp
```diff
@@ -84,7 +84,7 @@ bool pcieFunc::loadConf()
     std::vector<std::string> config;
     std::string err;
 
-    dev->sysfs_get("", "config_mailbox_channel_switch", err, chanSwitch);
+    dev->sysfs_get("", "config_mailbox_channel_switch", err, chanSwitch, -1);
```
maxzhen: default should be 0 here, which means that all mailbox msgs should go through HW mailbox.

src/runtime_src/core/pcie/tools/xbmgmt/cmd_config.cpp
```diff
@@ -150,7 +150,7 @@ static void showDevConf(std::shared_ptr<pcidev::pci_device>& dev)
     std::string errmsg;
     int lvl = 0;
 
-    dev->sysfs_get("icap", "sec_level", errmsg, lvl);
+    dev->sysfs_get("icap", "sec_level", errmsg, lvl, -1);
```
maxzhen: 0 would be a good default, which means no security requirement.

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -1671,7 +1671,7 @@ int xcldev::device::testP2p()
     if (dev == nullptr)
         return -EINVAL;
 
-    dev->sysfs_get("", "p2p_enable", errmsg, p2p_enabled);
+    dev->sysfs_get<int>("", "p2p_enable", errmsg, p2p_enabled, 0);
```
maxzhen: is this a bool?

src/runtime_src/core/pcie/tools/xbutil/xbutil.h
```diff
@@ -719,26 +719,26 @@ class device {
...
+        pcidev::get_dev(m_idx)->sysfs_get<int>( "rom", "ddr_bank_count_max", errmsg, ddr_count, 0 );
+        pcidev::get_dev(m_idx)->sysfs_get( "icap", "clock_freqs",            errmsg, clock_freqs ); 
+        pcidev::get_dev(m_idx)->sysfs_get( "dma", "channel_stat_raw",        errmsg, dma_threads ); 
+        pcidev::get_dev(m_idx)->sysfs_get<int>( "", "link_speed",            errmsg, pcie_speed, 0 );
+        pcidev::get_dev(m_idx)->sysfs_get<int>( "", "link_width",            errmsg, pcie_width, 0 );
+        pcidev::get_dev(m_idx)->sysfs_get<bool>( "", "mig_calibration",      errmsg, mig_calibration, false );
+        pcidev::get_dev(m_idx)->sysfs_get( "rom", "FPGA",                    errmsg, fpga );
+        pcidev::get_dev(m_idx)->sysfs_get( "icap", "idcode",                 errmsg, idcode );
+        pcidev::get_dev(m_idx)->sysfs_get( "dna", "dna",                     errmsg, dna );
+        pcidev::get_dev(m_idx)->sysfs_get<int>("", "p2p_enable",             errmsg, p2p_enabled, 0 );
```
maxzhen: bool?

src/runtime_src/core/pcie/tools/xbutil/xbutil.h
```diff
@@ -763,69 +763,66 @@ class device {
...
+        unsigned int fan_rpm, xmc_fpga_temp, xmc_fan_temp;
         std::string fan_presence;
         
-        pcidev::get_dev(m_idx)->sysfs_get( "xmc", "xmc_fpga_temp", errmsg, xmc_fpga_temp );
-        pcidev::get_dev(m_idx)->sysfs_get( "xmc", "xmc_fan_temp",  errmsg, xmc_fan_temp );
-        pcidev::get_dev(m_idx)->sysfs_get( "xmc", "fan_presence",  errmsg, fan_presence );
-        pcidev::get_dev(m_idx)->sysfs_get( "xmc", "xmc_fan_rpm",   errmsg, fan_rpm );
+        pcidev::get_dev(m_idx)->sysfs_get_sensor( "xmc", "xmc_fpga_temp", errmsg, xmc_fpga_temp );
+        pcidev::get_dev(m_idx)->sysfs_get_sensor( "xmc", "xmc_fan_temp",  errmsg, xmc_fan_temp );
+        pcidev::get_dev(m_idx)->sysfs_get( "xmc", "fan_presence",                errmsg, fan_presence );
```
maxzhen: extra spaces before errmsg?

src/runtime_src/core/pcie/linux/scan.h
AShivangi: The default value is what the user is passing. We aren't specifying 0 or -1 here. 
Should I remove the err_msg param? Wouldn't it be better if we keep it so that the caller doesn't get confused when using the template or one of the other functions? Instead, I can return "N/A" in the err_msg

src/runtime_src/core/pcie/linux/scan.h
AShivangi: It is better to leave it as int. "T& i" and "T& default_val" don't use the same T, eg, first one is uint_64 and second one is simple int so it gives an error.

maxzhen:
retest this please

maxzhen:
retest this please

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -634,7 +669,7 @@ int shim::xclGetDeviceInfo2(xclDeviceInfo2 *info)
     info->mHALMinorVersion = XCLHAL_MINOR_VER;
     info->mMinTransferSize = DDR_BUFFER_ALIGNMENT;
     info->mDMAThreads = 2;
-    xclSysfsGetDeviceInfo(info);
+    // xclSysfsGetDeviceInfo(info);
```
maxzhen: @AShivangi , why do you comment out this line? this seems to cause the bandwith test to core dump.

PR #2138, opened 2019-09-24
use sysfs nodes to get device info

AShivangi:
move away from xclDeviceInfo2 and xclErrorStatus structs by using sysfs nodes to get information


src/runtime_src/core/pcie/tools/xbutil/xbutil.h
```diff
@@ -231,6 +223,12 @@ class device {
         xclClose(m_handle);
     }
 
+    std::string name() const {
```
maxzhen: Instead of reading it from sysfs every time, should we cache it since it won't change?

src/runtime_src/core/pcie/tools/xbutil/xbutil.h
```diff
@@ -715,14 +710,13 @@ class device {
...
-        std::string name, vendor, device, subsystem, subvendor, xmc_ver,
+        std::string vendor, device, subsystem, subvendor, xmc_ver,
             ser_num, bmc_ver, idcode, fpga, dna, errmsg, max_power;
         int ddr_size = 0, ddr_count = 0, pcie_speed = 0, pcie_width = 0, p2p_enabled = 0;
         std::vector<std::string> clock_freqs;
         std::vector<std::string> dma_threads;
         bool mig_calibration;
         
-        pcidev::get_dev(m_idx)->sysfs_get( "rom", "VBNV",               errmsg, name ); 
         pcidev::get_dev(m_idx)->sysfs_get( "", "vendor",                errmsg, vendor );
```
maxzhen: Considering cache vendor/device/subdev since they don't change?

src/runtime_src/core/pcie/tools/xbutil/xbutil.h
AShivangi: Yes. Sounds good

src/runtime_src/core/pcie/tools/xbutil/xbutil.h
AShivangi: I am not sure how that might be helpful. I don't see vendor/device/subdev being used anywhere else

PR #2142, opened 2019-09-25
scheduler reset after xclbin load for PR

Himanshu-xilinx:
* Reset scheduler state after loading new xclbin.
* This solves single process doing multiple xclbin loading issue.
* This solves the case where Applications run one after other with different xclbins
* This doesn't solve multiple threads/processes loading different xclbins at same time. This can be done using context (TODO).

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/sched_exec.c
+++ b/src/runtime_src/core/edge/drm/zocl/sched_exec.c
@@ -2785,7 +2785,7 @@ static irqreturn_t sched_cq_isr(int irq, void *arg)
 
 inline void init_exec(struct sched_exec_core *exec_core)
 {
-	unsigned int i;
+        unsigned int i;
         exec_core->scheduler = &g_sched0;
         exec_core->num_slots = 16;
         exec_core->num_cus = 0;
--- a/src/runtime_src/core/edge/drm/zocl/sched_exec.c
+++ b/src/runtime_src/core/edge/drm/zocl/sched_exec.c
@@ -2818,7 +2818,6 @@ inline void init_exec(struct sched_exec_core *exec_core)
  * 
  * This function ensures that the device exec_core state is reset to
  * same state as was when scheduler was originally probed for the device.
- * This gets called after each xclbin load. This gets called for the first time also
  *
  * @drm: Device node to initialize
  *
--- a/src/runtime_src/core/edge/drm/zocl/sched_exec.c
+++ b/src/runtime_src/core/edge/drm/zocl/sched_exec.c
@@ -1556,7 +1556,6 @@ delete_cmd_list(void)
  * Clear stale command object associated with execution core.
  * This can occur if the HW for some reason hangs.
  */
-SCHED_UNUSED
 static void
 reset_exec(struct sched_exec_core *exec)
 {
@@ -2784,6 +2783,36 @@ static irqreturn_t sched_cq_isr(int irq, void *arg)
 	return IRQ_HANDLED;
 }
 
+inline void init_exec(struct sched_exec_core *exec_core)
+{
+	unsigned int i;
+        exec_core->scheduler = &g_sched0;
+        exec_core->num_slots = 16;
+        exec_core->num_cus = 0;
+        exec_core->cu_base_addr = 0;
+        exec_core->cu_shift_offset = 0;
+        exec_core->polling_mode = 1;
+        exec_core->cq_interrupt = 0;
+        exec_core->configured = 0;
+        exec_core->cu_isr = 0;
+        exec_core->cu_dma = 0;
+        exec_core->num_slot_masks = 1;
+        exec_core->num_cu_masks = 0;
+        exec_core->ops = &penguin_ops;
+
+        for (i = 0; i < MAX_SLOTS; ++i)
+                exec_core->submitted_cmds[i] = NULL;
+
+        for (i = 0; i < MAX_U32_SLOT_MASKS; ++i)
+                exec_core->slot_status[i] = 0;
+
+        for (i = 0; i < MAX_U32_CU_MASKS; ++i) {
+                exec_core->cu_status[i] = 0;
+                exec_core->cu_init[i] = 0;
+                exec_core->cu_valid[i] = 0; //default value is invalid(0)
+        }
+}
+
 /**
  * zocl_exec_reset() 
  * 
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -2784,6 +2784,53 @@ static irqreturn_t sched_cq_isr(int irq, void *arg)
...
+
+	for (i = 0; i < MAX_U32_SLOT_MASKS; ++i)
+		exec_core->slot_status[i] = 0;
+
+	for (i = 0; i < MAX_U32_CU_MASKS; ++i) {
+		exec_core->cu_status[i] = 0;
+		exec_core->cu_init[i] = 0;
+		exec_core->cu_valid[i] = 0;
+	}
+
```
xdavidz: Hi, thanks this is a simpler way to fix one issue that single user use different xclbin which may bring different configuration. So we will have to re-configure the scheduler to make sure the following xclbin won't be broken. What is missing is that, we also want to make sure the previous runs have been finished, if it is hung for long time, user should take care of the situation before loading the next new xclbin, aka cu drainage. Those will be bring by context switch. I have a code in testing mode: 	https://github.com/xdavidz/XRT1/commit/04155fbb72ead68adefaebc2413e7b200db3e65f

But for this effort, I think if we really want to reset the scheduler as of now, please avoid dup code, the following code is due with just the code right below in sched_init_exec. And, most importantly there are some hard-code should be refactored in the near future (like num_slots = 16), those should be from the device-tree or vise versa, but single way. If we have different device-tree config, the driver will just panic today. My advise would be to use a function to abstract common code both in sched_reset and sched_init or even sched_fini. The above context switch code could be an example, again it is being tested. What do you think?
 
exec_core->scheduler = &g_sched0;
	exec_core->num_slots = 16;		exec_core->num_slots = 16;
	exec_core->num_cus = 0;		exec_core->num_cus = 0;
	exec_core->cu_base_addr = 0;		exec_core->cu_base_addr = 0;
	exec_core->cu_shift_offset = 0;		exec_core->cu_shift_offset = 0;
	exec_core->polling_mode = 1;		exec_core->polling_mode = 1;
	exec_core->cq_interrupt = 0;		exec_core->cq_interrupt = 0;
	exec_core->cu_isr = 0;		exec_core->cu_isr = 0;
	exec_core->cu_dma = 0;		exec_core->cu_dma = 0;
	exec_core->num_slot_masks = 1;		exec_core->num_slot_masks = 1;
	exec_core->num_cu_masks = 0;		exec_core->num_cu_masks = 0;
	exec_core->ops = &penguin_ops;		exec_core->ops = &penguin_ops;

src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -2784,6 +2784,53 @@ static irqreturn_t sched_cq_isr(int irq, void *arg)
 	return IRQ_HANDLED;
 }
 
+/**
+ * zocl_exec_reset() 
+ * 
+ * This function ensures that the device exec_core state is reset to
+ * same state as was when scheduler was originally probed for the device.
+ * This gets called after each xclbin load. This gets called for the first time also
```
mamin506: This function is good! We don't need to mention where it will get called.

src/runtime_src/core/edge/drm/zocl/sched_exec.c
mamin506: I see a lot common code in zocl_exec_reset(struct drm_device *drm) and sched_init_exec(struct drm_device *drm).

Could we create a new function like:
inline void init_exec(struct sched_exec_core *exec_core) {
    exec_core->scheduler = &g_sched0;
    exec_core->num_slots = 16;
    ...
}

larry9523:
A few comments.
1) As mentioned by @mamin506 and @xdavidz, please remove the duplicated code in sched_init_exec
2) Please take care of all the memory we allocated from the previous xclbin downloading to avoid memory leak. (e.g. zdev->ip). 
3) Please double check all sysfs node accessing won't crash during xclbin replacement.
4) You may want to remove SCHED_UNUSED before reset_exec.

Himanshu-xilinx:
> A few comments.
> 
> 1. As mentioned by @mamin506 and @xdavidz, please remove the duplicated code in sched_init_exec
> 2. Please take care of all the memory we allocated from the previous xclbin downloading to avoid memory leak. (e.g. zdev->ip).
> 3. Please double check all sysfs node accessing won't crash during xclbin replacement.
> 4. You may want to remove SCHED_UNUSED before reset_exec.

1. Restructured the code
2. Sections(ip,debug_ip,connectivity,topology) already getting reset. 
     Added zocl_clear_mem to clean mem before init.
3. All sysfs working. Tested
4. Removed

src/runtime_src/core/edge/drm/zocl/sched_exec.c
chvamshi-xilinx: @xdavidz , These changes are good enough for the single process cases. We should reset the scheduler after loading the new xclbin anyway. It is better if we push these changes sothat DSV can validate these cases too.

chvamshi-xilinx: I agree with you that context switch change is the right change to cover all the cases, but as this support may take some more time, i would suggest, we should go with atleast resetting the scheduler changes.

xdavidz: thanks, I am ok with ahead with the fix, please make sure no panic during the test. Code looks ok.

src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -2784,6 +2783,56 @@ static irqreturn_t sched_cq_isr(int irq, void *arg)
 	return IRQ_HANDLED;
 }
 
+inline void init_exec(struct sched_exec_core *exec_core)
+{
+	unsigned int i;
+        exec_core->scheduler = &g_sched0;
```
mamin506: Use 'tab' at the start of the line.

mamin506: No. All these line are using space. Linux driver required 'tab'.

Himanshu-xilinx: fixed

larry9523:
There seems to be race window between zocl_free_sections and the sysfs node reading. It could crash the system on some conditions. For example, when you free zdev->topology, before calling CLEAR, another thread is reading sysfs mem node. zdev->topology is not set to NULL yet but it is actually freed. We may want to put "if (!zdev->topology)" under read_lock protection and "zocl_free_sections()" under write_lock protection. It was supposed to be OK because we can only load xclbin once so that zdev->topology is NULL before we load xclbin.

xdavidz:
I have a question about switching bitstreams (different xclbin kernels) on zcu104. I made 2 simple xclbin, but it seems that I cannot swap them without reboot? Could you please share your xclbins that I can verify if it works?

Himanshu-xilinx:
> I have a question about switching bitstreams (different xclbin kernels) on zcu104. I made 2 simple xclbin, but it seems that I cannot swap them without reboot? Could you please share your xclbins that I can verify if it works?

---> sent mail

PR #2148, opened 2019-09-25
Coverity fixes

AShivangi:


src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1760,7 +1760,7 @@ int xclResetDevice(xclDeviceHandle handle, xclResetKind kind)
 int xclP2pEnable(xclDeviceHandle handle, bool enable, bool force)
 {
     xocl::shim *drv = xocl::shim::handleCheck(handle);
-    return drv ? drv->p2pEnable(enable, force) : -ENODEV;
+    return drv ? drv->p2pEnable(enable, force) : ENODEV;
```
maxzhen: hmm...why?

src/runtime_src/core/pcie/tools/xbmgmt/cmd_part.cpp
```diff
@@ -76,6 +76,8 @@ int program_prp(unsigned index, const std::string& xclbin, bool force)
         if (!errmsg.empty())
         {
             std::cout << errmsg << std::endl;
+            delete [] buffer;
```
maxzhen: can we use smart pointer here for the buffer so that it'll be freed automatically?

src/runtime_src/core/pcie/linux/shim.cpp
AShivangi: strerror() doesn't take -ve value in xclP2p xbutil.cpp. But I think it is better to change "int ret" to unsigned instead of this change.

maxzhen: it's more consistent to return negative value in similar functions here. we should pass -val to strerror().

src/runtime_src/core/pcie/common/memaccess.h
```diff
@@ -164,37 +164,39 @@ namespace xcldev {
      * Caller's responsibility to do sanity checks. No sanity checks done here
      */
     int readBank(std::ofstream& aOutFile, unsigned long long aStartAddr, unsigned long long aSize) {
-      char *buf = 0;
+      auto buf = std::make_shared<char>();
```
stsoe: This char* wrapped by 'buf' will be deleted using 'delete', but posix_memalign probably expects "free".  While this may work on Linux, it is not correct.  Match new  with delete, and malloc with free.

Also, it doesn't look like this should be shared ownership, looks like 'buf' is really strictly owned within the function scope.

I think we could wrap aligned_alloc (or posix_memalign) to return a unique_ptr.
Here is an example that works.  For shared usage, I think the allocation function should be declared in  memalign.h and defined in memalign.cpp (runtime_src/core/common)
```
#include <memory>
#include <cstdlib>

namespace xrt_core {
static auto deleter = [](void* ptr) { free(ptr); };
std::unique_ptr<void, decltype(deleter)>
aligned_alloc(size_t align, size_t size)
{
  return std::unique_ptr<void, decltype(deleter)>(::aligned_alloc(align, size),deleter);
}
} // namespace xrt_core
```
On success aligned_alloc returns a valid ptr, on failure it returns nullptr.  
Your code would simply call
`auto buf = xrt_core::aligned_alloc(get_pagesize(), blockSize);`

src/runtime_src/core/pcie/common/memaccess.h
```diff
@@ -164,37 +164,39 @@ namespace xcldev {
...
       unsigned long long blockSize = 0x20000;
-      if (xrt_core::posix_memalign((void**)&buf, getpagesize(), blockSize))
+      if (xrt_core::posix_memalign((void**)buf.get(), getpagesize(), blockSize))
         return -1;
-      std::memset(buf, 0, blockSize);
+      std::memset(buf.get(), 0, blockSize);
 
       size_t count = aSize;
       uint64_t incr;
+      std::ios_base::fmtflags f(std::cout.flags());
```
stsoe: Nitpicking, you should use an automatic object for this, so that destructor automatically restores the fmt.

src/runtime_src/core/pcie/common/memaccess.h
```diff
@@ -164,37 +164,39 @@ namespace xcldev {
...
           //error
           std::cout << "Error (" << strerror (errno) << ") reading 0x" << std::hex << incr << " bytes from DDR/HBM/PLRAM at offset 0x" << std::hex << phy << std::dec << "\n";
-          free(buf);
+          std::cout.flags(f);
           return -1;
         }
         count -= incr;
         if (incr) {
-          aOutFile.write((const char*)buf, incr);
+          aOutFile.write((const char*)buf.get(), incr);
```
stsoe: Use reinterpret_cast<const char*>, .e.g no c-style cast in C++.

src/runtime_src/core/pcie/common/memaccess.h
stsoe: Actually on second thought, this aligned alloc would have to be in header file but within namespace.  The type of 'deleter' is needed.

stsoe: How about this
```
% cat src/runtime_src/core/common/memalign.h
#include <memory>
#include <cstdlib>

namespace xrt_core {
...
auto aligned_ptr_deleter = [] (void* ptr) { free(ptr); } ;
using aligned_ptr_type = std::unique_ptr<void, decltype(aligned_ptr_deleter)>;
inline aligned_ptr_type
aligned_alloc(size_t align, size_t size)
{
  return aligned_ptr_type(::aligned_alloc(align, size),aligned_ptr_deleter);
}
} // xrt_core
```

src/runtime_src/core/pcie/common/memaccess.h
AShivangi: I have come across this in almost all the files. Right now I am making an RAII class in memaccess.h and in xbmgmt.h (all xbmgmt sub commands are using this now). I have seen this in xbutil as well, but I don't want to make that change in this PR. Is there a better place I can put this class in?

src/runtime_src/core/pcie/common/memaccess.h
```diff
@@ -40,6 +40,17 @@
 #include "xclhal2.h"
 #include "xclbin.h"
 
+class IosBaseFlags {
```
stsoe: lets move this code to runtime_src/core/common/utils.h
Put it under namespace xrt_core { ... }
And calls it `class ios_flags_restore`.

PR #2150, opened 2019-09-25
coverity fixes

chienwei-lan:


src/runtime_src/core/pcie/driver/linux/xocl/subdev/nifd.c
```diff
@@ -575,7 +575,7 @@ static int nifd_probe(struct platform_device *pdev)
     struct xocl_nifd *nifd;
     struct resource *res;
     struct xocl_dev_core *core;
-    struct FeatureRomHeader rom;
+    struct FeatureRomHeader rom = { 0 };
```
maxzhen: should do rom = {{0}}, or centos build may emit warning.

PR #2157, opened 2019-09-26
Add CRI register address for health check

chienwei-lan:


src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
```diff
@@ -446,11 +446,20 @@ static int health_check_cb(void *data)
...
+	void __iomem *shutdown_clk = xocl_iores_get_base(lro, IORES_CLKSHUTDOWN);
+	uint32_t latched;
 
 	if (!health_check)
 		return 0;
 
+	if (shutdown_clk) {
+		latched = XOCL_READ_REG32(shutdown_clk);
+		if (latched & 0x1)
+			mgmt_info(lro, "clk shutdown latched");
```
maxzhen: Should this be an error? Can we print out something more meaningful than "latched"? Something like "Kernel clock is shut down. Board may temp / power may be too high"?

PR #2158, opened 2019-09-26
versal PR with decouple and reset

xdavidz:
I would like to start the code review for versal PR a little earlier, even though we haven't had a chance to test on v350 yet. Mrinal said the design will be the same.  Please let me know if anything can be designed better, we still have chance to push design a little bit better before it's being finalized.

Meantime, I will set it to do not merge for now.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_cu.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_cu.c
@@ -331,7 +331,8 @@ zocl_hls_cu_init(struct zocl_cu *cu, phys_addr_t paddr)
 		return -ENOMEM;
 	}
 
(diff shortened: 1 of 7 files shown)


src/runtime_src/core/edge/drm/zocl/zocl_cu.c
```diff
@@ -342,6 +343,7 @@ zocl_hls_cu_init(struct zocl_cu *cu, phys_addr_t paddr)
 	default:
 		core->max_credits = 1;
 	}
+	core->max_credits = 1;
```
mamin506: Is this debug code? Why set max_credits to 1 here?

xdavidz:
Closed it due to v350 PR is available, no need to have special code for tenzing.

PR #2161, opened 2019-09-27
zcu102/104_base: Common postlink for emulation/hw and cleanup of redundant files

subhransu-xilinx:


chvamshi-xilinx:
ok to test

PR #2163, opened 2019-09-27
use trace clock to calculate slope

hackwa:


src/runtime_src/xdp/profile/device/trace_parser.cpp
```diff
@@ -618,13 +618,19 @@ namespace xdp {
...
     if (!y1 && !x1) {
       y1 = static_cast <double> (hostTimestamp);
       x1 = static_cast <double> (deviceTimestamp);
     } else {
       y2 = static_cast <double> (hostTimestamp);
       x2 = static_cast <double> (deviceTimestamp);
-      mTrainSlope[type] = (y2 - y1) / (x2 - x1);
+      // slope in ns/cycle
+      if (isDeviceFlow) {
+        mTrainSlope[type] = 1000.0/mTraceClockRateMHz;
```
pgschuey: Do we know for certain that mTraceClockRateMHz is correct for a given design? In other words, if it's scaled to less than 300 MHz, do we know that and is reflected in this variable?

hackwa: Yes the clock is correct. I tried a couple designs with clock scaling

jvillarre:
The change requested by pgschuey has been resolved.

PR #2173, opened 2019-09-29
xbutil validate hangs during stress test

maxzhen:
xbutil validate may hang due to that read from stderr of child process is stuck.

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
+++ b/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
@@ -918,9 +918,13 @@ void testCaseProgressReporter(bool *quit)
     }
 }
 
-inline const char* value_or_empty(const char* value) { return value ? value : "" ; }
+inline const char* value_or_empty(const char* value)
+{
+    return value ? value : "";
+}
 
-static void set_shell_path_env(const std::string& var_name, const std::string& trailing_path, int overwrite)
+static void set_shell_path_env(const std::string& var_name,
+    const std::string& trailing_path, int overwrite)
 {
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -950,33 +954,35 @@ int runShellCmd(const std::string& cmd, std::string& output)
...
         }
     }
-
-    //Read stderr
-    if (output.find("PASS") == std::string::npos) {
-        char buffer[256];
-        int count = read(stderr_fds[0], buffer, sizeof(buffer)-1);
-        buffer[count] = 0;
-        std::cout << buffer << std::endl;
+    while (ret == 0 && stderr_child && !feof(stderr_child.get())) {
```
AShivangi: We need to explicitly check if the test passed or not. Even if the test passes, I get something like "pci id for fd x: 10ee:5005, driver (null)" in my stderr. This is what we are trying to avoid with the fds. I believe we need to check that with "output.find("PASS") == std::string::npos"

maxzhen: The idea is that runShellCmd() will just run the cmd and collect and return output from both stdout and stderr back to caller. It is the caller who should decide if test has passed or not based on the content of the output. If caller decides that the test has passed, it will not show the output anyway. In this case, whatever is in stderr (now is part of combined output returned from runShellCmd()) does not matter. I tested xbutil validate on my system. When everything works, I don't see anything unexpected printed out...

AShivangi: Hmm. I see... I think this should be fine then :)

PR #2174, opened 2019-09-30
change zcu104_base platform from DFX to non DFX type

sivarajeshjarugula:
change zcu104_base platform from  DFX (dynamic function exchange or partial reconfiguration ) to non DFX type


chvamshi-xilinx:
ok to test

PR #2176, opened 2019-09-30
xmc fixes

houlz0507:
* the order of stopping ERT and MB does not look right
* only hold reset GPIO for loading ERT and CMC image


houlz0507:
Create another one later.

PR #2179, opened 2019-09-30
Keep separate scheduler queue for running commands

stsoe:
This allows for immediate processing of completed commands rather
than having to iterate entire command list to look for the commands
that are in running state


stsoe:
Reconsider

PR #2182, opened 2019-09-30
pyopencl install change for ubuntu...

AShivangi:
... using pip so that the newest version is installed. Apt installs 2015 version which doesn't behave correctly with CUDA

- tested on ubuntu 16.04, 18.04
- tested on centOS 7.6


keryell:
There is also the `python3-pyopencl` package...
But what is the motivation for having the latest PyOpenCL for *CUDA*, and for example not the latest CUDA version or not the latest C++ version? :-)
I am unsure where PyOpenCL is used when I use XRT on my machine...
Should we ship different versions or sub-packages of XRT?

PR #2183, opened 2019-09-30
install fw in /lib/firmware/xilinx

stsoe:


stsoe:
Lizhi.   Can we change the driver to load scheduler firmware from /lib/firmware/xilinx/sched.bin (sched_u50.bin) for 2019.2?  XRT 2019.2 must be in sync with the ERT fw built with XRT.   If the firmware is not present in /lib/firmware/xilinx/ then default to dsabin content.

PR #2200, opened 2019-10-03
added dummy frame type. more features for zero-copy

sarab96:
added dummy frame type. more features for zero-copy


sarab96:
retest this please

PR #2204, opened 2019-10-03
get_free_cu can always return -1 for invalid cu_mask

xdavidz:
Found the invalid cu_mask, like 0 or only try to mask the invalid cu, the cmd will be set to cu_idx = -1 and will stay in scheduler thread forever.

Aslo, used latest checkpatch, but only fixed errors instead of all warnings, we should get those warnings fixed for code clean with linux upstream.

/proj/rdi/staff/davidzha/XRT_zocl/build/checkpatch/checkpatch.pl --max-line-length=80 --no-tree -f
total: 0 errors, 102 warnings, 2959 lines checked

Verified on 104_base, ert_cmd->state has been set to ERT_CMD_STATE_ERROR.

Kernel log will print:  Cannot find valid cu from cu_mask

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/sched_exec.c
+++ b/src/runtime_src/core/edge/drm/zocl/sched_exec.c
@@ -741,7 +741,7 @@ configure(struct sched_cmd *cmd)
 	}
 	write_unlock(&zdev->attr_rwlock);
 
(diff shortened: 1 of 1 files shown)


xdavidz:
retest this please

PR #2217, opened 2019-10-07
fix zocl when iommu attach fails

larry9523:
This PR fixes two issues
1) When the IOMMU is present on amba bus, if we fail to attach iommu, we should clear flag in zocl. Otherwise, we will still work under IOMMU mode which is obviously wrong.
2) For any reason we fail to initialize iommu, do not fail the whole zocl probe but keep moving on after clearing all the resources allocated during iommu initialization.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_drv.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_drv.c
@@ -497,6 +497,35 @@ static unsigned int zocl_poll(struct file *filp, poll_table *wait)
 	return ret;
 }
 
+static int zocl_iommu_init(struct drm_zocl_dev *zdev,
+		struct platform_device *pdev)
+{
+	struct iommu_domain_geometry *geometry;
+	u64 start, end;
+	int ret;
+
+	zdev->domain = iommu_domain_alloc(&platform_bus_type);
+	if (!zdev->domain)
+		return -ENOMEM;
+
+	ret = iommu_attach_device(zdev->domain, &pdev->dev);
+	if (ret) {
+		DRM_INFO("IOMMU attach device failed. ret(%d)\n", ret);
+		iommu_domain_free(zdev->domain);
+		zdev->domain = NULL;
+		return ret;
+	}
+
+	geometry = &zdev->domain->geometry;
+	start = geometry->aperture_start;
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/edge/drm/zocl/zocl_drv.c
```diff
@@ -671,26 +700,12 @@ static int zocl_drm_platform_probe(struct platform_device *pdev)
...
-		end = geometry->aperture_end;
-
-		DRM_INFO("IOMMU aperture initialized (%#llx-%#llx)\n",
-				start, end);
+		/*
+		 * Note: we ignore the return value of zocl_iommu_init().
+		 * In the case of failing to initialize iommu, zocl
+		 * driver will keep working with iommu disabled.
+		 */
+		(void) zocl_iommu_init(zdev, pdev);
```
sonals: Should we at least warn if the IOMMU attach fails?

src/runtime_src/core/edge/drm/zocl/zocl_drv.c
```diff
@@ -497,6 +497,35 @@ static unsigned int zocl_poll(struct file *filp, poll_table *wait)
...
+static int zocl_iommu_init(struct drm_zocl_dev *zdev,
+		struct platform_device *pdev)
+{
+	struct iommu_domain_geometry *geometry;
+	u64 start, end;
+	int ret;
+
+	zdev->domain = iommu_domain_alloc(&platform_bus_type);
+	if (!zdev->domain)
+		return -ENOMEM;
```
xdavidz: For ENOMEM, why we still continue? I guess we should bail out right away?

src/runtime_src/core/edge/drm/zocl/zocl_drv.c
larry9523: Yes we do log warning message if zocl_iommu_init() if attache fails

larry9523:
retest this please

src/runtime_src/core/edge/drm/zocl/zocl_drv.c
xdavidz: make sense. it is actually no-resource than no memory.

PR #2222, opened 2019-10-07
Using zocl ert to handle dataflow CU on versal

larry9523:
On versal device, if we have dataflow CU, we need to use ert_full mode to send CU control command to zocl ert. This PR force ert_full mode on versal if ert is configured and disable ert_poll regardless dataflow is configured or not.


larry9523:
retest this please

PR #2226, opened 2019-10-08
remove memcpy_toio

houlz0507:


stsoe:
Hi Lizhi.  Changes look good, but what's the underlying reason for this change?

PR #2228, opened 2019-10-09
Dequeue the msg so even if peer is offline the msg can timeout

arpitxilinx:
This will allow xocl driver to be unloaded on VM in Hyper-v, after the reset when UPF is in bad state.


sonals:
ok to test

PR #2231, opened 2019-10-09
Multiple init.sh calling issue

chvamshi-xilinx:
Hi,

This change contains a removal of mnt-sd.sh from init.d as this is not used anywhere and causing problems

Thanks
Vamshi Krishna


mamin506:
Thanks Vamshi for the fix. Looks good.

PR #2239, opened 2019-10-10
Bug Fix: The option "--info" output file wasn't checking for an existing file.

rozumx:
When using the "--info" option for an output file, it wasn't checking for an already existing file.   This pull request adds functionality to check if the file exists and if it does to not write out the file, unless of course the "--force" option is used.


rozumx:
retest this please

PR #2241, opened 2019-10-10
CR-1044962 Report a warning message during xbutil validate to let user know, if AUX power cable is not plugged in

AShivangi:


maxzhen:
the previous change has passed pipeline. the later one is just changing one msg word. merging now.

PR #2242, opened 2019-10-10
Reverting new subdev Info from U2 

chvamshi-xilinx:
Hi,

Reverting new subdev Info from U2 as it is causing a random hang .


chvamshi-xilinx:
> 
> 
> What are 0x6988 and 0x5035 used for.

HI @houlz0507 , These are PCI-DEVICE IDs used for U.2 Platform (Samsung SSD)

mamin506:
This fix is for CR-1045550.

PR #2249, opened 2019-10-11
no lock on mgmt pf for xclbin lockdown on user pf

maxzhen:
Xocl will try to bump up ref count on mgmt pf on first xclbin user lockdown and release ref count on last user unlocking the xclbin. However, there is no clean solution to reset ref count on mgmt side when things go wrong.

If we remove ref count on mgmt pf side and make it stateless, we don't need to solve the mgmt ref count reset issue. The question is: is ref count on mgmt side really needed? 

@sonals , let's have a discussion.


maxzhen:
> 
> 
> Do we need to update MPD/MSD documents on GitHub.IO?

This is not specific about mpd/msd. The Mailbox Inter-domain Communication Protocol doc will be updated since it comes directly from code.

PR #2252, opened 2019-10-11
more clocks are added to zynq7000 platform to align with the SDSoC platforms

sivarajeshjarugula:
more clocks are added to zynq7000 platform to align with the SDSoC platforms


chvamshi-xilinx:
ok to test

PR #2254, opened 2019-10-11
Flow is hanging when kernel is not supporting Interrupt

chvamshi-xilinx:
Hi,

This change contains a fix of hang when one of the kernel is not supporting intterupt.
If any of the kernel is not supporting interrupt, we should fall back to polling mode.

Thanks


src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -723,6 +723,13 @@ configure(struct sched_cmd *cmd)
 		SCHED_DEBUG("++ configuring penguin scheduler mode\n");
 		exec->ops = &penguin_ops;
 		exec->polling_mode = cfg->polling;
+		//Interrupt may not be enabled for some of the kernel, 
+		//Need to use polling mode in that case
+		if(!cfg->cu_isr) {
```
mamin506: cu_isr is not use for this purpose. In NON-ERT mode, we should not care about cu_isr.
What do you mean "kernel is not supporting interrupt"? Is it a RTL kernel?
If you know some of your kernel didn't has interrupt, set "ert_polling = true" in xrt.ini.

chvamshi-xilinx: Hi @mamin506 , 
cuisr in the config command depends on following two values (both ERT/non-ERT mode)
1) ert_cuisr (xrt.ini option. which is by default true) 
2) All kernels supports interrupts 

cuisr comes false only if any of the kernel is not supporting interrupt. Thatswhy we can consider this value in non-ERT mode also. As we can determine this condition of polling vs interrupt data driven, i would suggest we should go with this option. It is not a good idea to ask user to use ert_polling=true.

Please suggest

Please look at get_cuisr function in following file
runtime_src/core/common/xclbin_parser.cpp

src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -723,6 +723,13 @@ configure(struct sched_cmd *cmd)
...
+			DRM_WARN("Interrupt is not enabled for atleast one kernel."
+			   "Fall back to polling mode\n");
+		  	exec->polling_mode = 1;
```
mamin506: There are two space in before exec->poling_mode. Please use tab for indent in the driver code.

src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -723,6 +723,13 @@ configure(struct sched_cmd *cmd)
 		SCHED_DEBUG("++ configuring penguin scheduler mode\n");
 		exec->ops = &penguin_ops;
 		exec->polling_mode = cfg->polling;
+		//Interrupt may not be enabled for some of the kernel, 
+		//Need to use polling mode in that case
+		if(!cfg->cu_isr) {
+			DRM_WARN("Interrupt is not enabled for atleast one kernel."
```
mamin506: *at least

src/runtime_src/core/edge/drm/zocl/sched_exec.c
mamin506: I still don't know which kind of kernel didn't support interrupt. Maybe streaming kernel?
I dislike the name cuisr, since it is a dedicated RTL IP in PCIe platform.
But I got your point about respect the CU properties in the XCLBIN.

maxzhen:
@chvamshi-xilinx , please address review comments with a separate pull request. I'll merge this one to unblock sprite test noww.

src/runtime_src/core/edge/drm/zocl/sched_exec.c
chvamshi-xilinx: This is mainly for the RTL kernels only. So many rtl kernel designs are in sprite. 
Yes, cuisr is bit confusing, but There is no other field in config packet has this information

PR #2265, opened 2019-10-13
Fixing free running kernel issue in Zynq

chvamshi-xilinx:
Hi,

This change contains a fix for free running kernel issue in Zynq. This file is using uint64_t and auto combination which needs to be cleaned up.
I am fixing only free running kernel issue as it is causing many sprite issues.

Thanks


chvamshi-xilinx:
> 
> 
> This code is also used by PCIE, right? Do we see similar issue on PCIE as well for free-running CUs? The comment above get_base_addr() worries me. We should be getting CU index from sysfs node, which is determined by driver. We can't have both user land and driver to determine CU index and hope they will work...

Hi Max,
This issue is specific to 32bit system. This issue is not there for zyncMP as following line returns 64bit max() in zynq platform
addr = std::numeric_limits<uint64_t>::max() & ~0xFF;

This change effects only streaming free running kernels as its base address comes empty.

mamin506:
> This change effects only streaming free running kernels as its base address comes empty.

I remember the address is '0xFFFFFFFF' for streaming free running kernels. What do you mean empty?

maxzhen:
@chvamshi-xilinx , please file another pull request to address @mamin506 's comment. I'm merging it now to unblock sprite test.

chvamshi-xilinx:
> 
> 
> > This change effects only streaming free running kernels as its base address comes empty.
> 
> I remember the address is '0xFFFFFFFF' for streaming free running kernels. What do you mean empty?

In xclbin, This value is empty. xclbin parser converts empty string to integer which gets converted to 0xFFFFFFFF. Following sample from xclbin for free running kernels.
 <instance name="increment_1">
            <addrRemap base="" port=""/>
</instance>
get_base_addr is returning 64Bit in zynq platforms due to following code which is creating a problem while mapping a kernel. 
 addr = std::numeric_limits<uint64_t>::max() & ~0xFF;

Anyway, We will send a seperate PR by using paddr_t

PR #2268, opened 2019-10-13
Fall back to regular copy if DMA copy cannot be performed

chvamshi-xilinx:
Hi,
This change contains a fix to fall back to regaular copy if DMA copy cannot be performed.
We should check the conditions before adding it to execbuf command list.

Thanks


xdavidz:
Overall it's ok, some nits please fix. So the root cause is the execbuf should fail?

I don't think that is by design, the exexbuf will success, instead the wait execbuf should return ERT_CMD_ERROR. Have you checked the ERT_CMD_ERROR is not returned by zocl driver? The code in shim.cpp will catch up the error and switch back to CPU copy.

If there is a special reason that we really need the execbuf to fail, I am ok with this. Thanks for fixing it.

chvamshi-xilinx:
> 
> 
> Overall it's ok, some nits please fix. So the root cause is the execbuf should fail?
> 
> I don't think that is by design, the exexbuf will success, instead the wait execbuf should return ERT_CMD_ERROR. Have you checked the ERT_CMD_ERROR is not returned by zocl driver? The code in shim.cpp will catch up the error and switch back to CPU copy.
> 
> If there is a special reason that we really need the execbuf to fail, I am ok with this. Thanks for fixing it.

Hi David,
PCIe platforms also uses same approach for copy buffer, If xocl cannot do kdma/p2p, then it returns an error. That is why am also using same approach.

xdavidz:
> > Overall it's ok, some nits please fix. So the root cause is the execbuf should fail?
> > I don't think that is by design, the exexbuf will success, instead the wait execbuf should return ERT_CMD_ERROR. Have you checked the ERT_CMD_ERROR is not returned by zocl driver? The code in shim.cpp will catch up the error and switch back to CPU copy.
> > If there is a special reason that we really need the execbuf to fail, I am ok with this. Thanks for fixing it.
> 
> Hi David,
> PCIe platforms also uses same approach for copy buffer, If xocl cannot do kdma/p2p, then it returns an error. That is why am also using same approach. Please check convert_execbuf in mb_scheduler.c

That's why I am curious, because it is by design. And should pass the test unless something changed.
Let me test it a little bit.

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -586,14 +580,14 @@ int zocl_copy_bo_async(struct drm_device *dev,
 		ZOCL_BO_FLAGS_SVM);
 	if ((dst_bo->flags & unsupported_flags) ||
 	    (src_bo->flags & unsupported_flags)) {
-		DRM_ERROR("Failed not supported dst flags 0x%x and "
+	  	DRM_ERROR("Failed: Not supported dst flags 0x%x and "
```
xdavidz: Not space in between tabs for clean checkpatch.

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -547,36 +547,30 @@ int zocl_sync_bo_ioctl(struct drm_device *dev,
...
 	uint64_t		        dst_size, src_size;
 	int 				unsupported_flags = 0;
-	int 				rc = 0;
-
-	if (dma_handle->dma_func == NULL) {
-		DRM_ERROR("Failed: no callback dma_func for async dma");
-		return -EINVAL;
-	}
-
+	bool rc = true;
```
xdavidz: could this be aligned with other declaration? Like in    unsupported_flags?

xdavidz:
Just some nits, overall looks good to me.

xdavidz:
the nits can be fixed later, just approved it

chvamshi-xilinx:
> 
> 
> the nits can be fixed later, just approved it

Feedback is incorporated in this PR itself. Already merged.

PR #2272, opened 2019-10-14
Coverity fixes

AShivangi:
- ios -> ios_base: seems like ios_flags_restore class wasn't really fixing the coverity issue. Not sure if this change will help

- getenv needs to be sanitized before being used. Coverity requires an annotation to be added for custom sanitization functions. This change should fix that issue

- fixing a couple of other coverity issues as well

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbmgmt/xqspips.cpp
+++ b/src/runtime_src/core/pcie/tools/xbmgmt/xqspips.cpp
@@ -231,12 +231,12 @@ XQSPIPS_Flasher::XQSPIPS_Flasher(std::shared_ptr<pcidev::pci_device> dev)
     // By default, it is 'perallel'
     mConnectMode = 0;
     // U30 Rev.A use single
-    if (typeStr.find("single")) {
+    if (typeStr.find("single") != std::string::npos) {
         mConnectMode = 1;
     }
 
     mBusWidth = 4;
-    if (typeStr.find("x2")) {
+    if (typeStr.find("x2") != std::string::npos) {
         // U30 Rev.B use x2
         mBusWidth = 2;
     }
--- a/src/runtime_src/core/common/utils.h
+++ b/src/runtime_src/core/common/utils.h
@@ -35,7 +35,7 @@ namespace xrt_core {
 
     private:
         std::ostream& ios;
-        std::ios::fmtflags f;
+        std::ios_base::fmtflags f;
     };
 } //xrt_core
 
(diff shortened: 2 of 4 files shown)


AShivangi:
retest this please

abalasa:
retest this please

src/runtime_src/core/pcie/tools/xbmgmt/xqspips.cpp
```diff
@@ -231,12 +231,12 @@ XQSPIPS_Flasher::XQSPIPS_Flasher(std::shared_ptr<pcidev::pci_device> dev)
     // By default, it is 'perallel'
     mConnectMode = 0;
     // U30 Rev.A use single
-    if (typeStr.find("single")) {
+    if (typeStr.find("single") != std::string::npos) {
```
maxzhen: shouldn't this needs to be if (typeStr.find("single") != 0)?

src/runtime_src/core/pcie/tools/xbmgmt/xqspips.cpp
```diff
@@ -231,12 +231,12 @@ XQSPIPS_Flasher::XQSPIPS_Flasher(std::shared_ptr<pcidev::pci_device> dev)
...
         mConnectMode = 1;
     }
 
     mBusWidth = 4;
-    if (typeStr.find("x2")) {
+    if (typeStr.find("x2") != std::string::npos) {
```
maxzhen: ditto.

src/runtime_src/core/pcie/tools/xbmgmt/xqspips.cpp
AShivangi: I thought that's how "find" is used. Search for "single" until the end of the string

maxzhen: But std::string::npos is -1, how does it match the original code here? @mamin506 , please confirm.

mamin506: typeStr.find("single") != std::string::npos is correct. If no matched was found, find() returns string::npos. The original code will failed if 'single' is the start of the string.

src/runtime_src/core/pcie/tools/xbmgmt/xqspips.cpp
mamin506: See above

PR #2276, opened 2019-10-14
rename shell to platform in xbmgmt/xbutil 

AShivangi:
visible changes in xbmgmt flash, xbutil query and xbutil validate error path


maxzhen:
can't afford breaking pipeline and (maybe) sprite as this moment. close it for now.

PR #2282, opened 2019-10-16
read msg on tcp socket should block until the full request is statisfied

xuhz:
msd/mpd read 'communication id', "verify status', 'sw mailbox msg length' should block until the full request is satisfied. 
It is possible a write()/send() of even a 4 bytes msg  in one side of a TCP connection results in arriving the other side in multiple packets.


xuhz:
Please retest

maxzhen:
retest this please

PR #2284, opened 2019-10-16
getting interrupt information from cu, not relying on cu_isr

Himanshu-xilinx:


chvamshi-xilinx:
retest this please

mamin506:
Read IP layout in scheduler is not ideal but acceptable.
Let's keep it here and we could clean it up at next release.

xdavidz:
> Read IP layout in scheduler is not ideal but acceptable.
> Let's keep it here and we could clean it up at next release.

If we are using xclbin.h, I would prefer move all those logic into a separate file like zocl_xclbin.c in which we include the xclbin.h only. The scheduler should not include the xclbin.h and else where. We will get this fixed in next release.

src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -662,6 +662,26 @@ init_cus(struct sched_cmd *cmd)
...
+get_cus_support_intr(struct drm_zocl_dev *zdev)
+{
+	struct ip_data *ip;
+        int i;
+        if (!zdev->ip)
+                return false;
+
+        for (i = 0; i < zdev->ip->m_count; ++i) {
+                ip = &zdev->ip->m_ip_data[i];
+                if(!(ip->properties & 0x1)) {
```
mamin506: Confirmed with system compiler, if the CU has a interrupt pin, the bit 0 of properties will be set to 1. This is the correct way to determine if a CU supports interrupt.

mamin506:
> If we are using xclbin.h, I would prefer move all those logic into a separate file like zocl_xclbin.c in which we include the xclbin.h only. The scheduler should not include the xclbin.h and else where. We will get this fixed in next release.

Yes. Let's do code refactoring for the next release. I think we should have scheduler, CU, xclbin, sysfs etc. components in zocl driver.

chvamshi-xilinx:
Thanks @mamin506  and @xdavidz , We have verified multiple cases and enabled and merged this change, 
We will move it to separate file in master.

david2011qq:
> Thanks @mamin506 and @xdavidz , We have verified multiple cases and enabled and merged this change,
> We will move it to separate file in master.

Thanks for the comment, the xclbin related code are not just from here. I have done the code change with context switch for versal. Will add you as a code reviewer. Thanks!

PR #2290, opened 2019-10-17
cma_obj overflows on typecasting to drm_zocl_bo

Himanshu-xilinx:


mamin506:
Thanks for the fix. The buffer object should be able to contain drm_zocl_bo.

Himanshu-xilinx:
Verified (CR-1046915)

mamin506:
> Verified (CR-1046915)

Thanks for pointing out the CR. This is helpful.

PR #2294, opened 2019-10-17
Merge from xrtwindows

stsoe:


src/runtime_src/core/include/xstream.h
```diff
@@ -0,0 +1,59 @@
+/*
```
maxzhen: indentation in this file looks strange.

src/runtime_src/core/include/xrt.h
```diff
@@ -120,7 +143,7 @@ struct xclDeviceInfo2 {
   unsigned short mMgtVtt;
   unsigned short m1v2Bottom;
   unsigned long long mDriverVersion;
-  unsigned mPciSlot;
+  unsigned int mPciSlot;
```
rozumx: Instead of "int" should this value be a long for on linux an int is 64bits and windows it is 32bits.

src/runtime_src/core/include/xrt.h
```diff
@@ -200,11 +223,11 @@ struct xclDeviceUsage {
     size_t h2c[8];
     size_t c2h[8];
     size_t ddrMemUsed[8];
-    unsigned ddrBOAllocated[8];
-    unsigned totalContexts;
+    unsigned int ddrBOAllocated[8];
```
rozumx: Instead of "int" should this value be a long for on linux an int is 64bits and windows it is 32bits.
There seems to be a few of these items, so I will not repeat this comment in this file :).

stsoe: I don't have the liberty to change the size of this struct on Linux.  'unsigned' is a misnomer for 'unsigned int' so I essentially didn't change anything , which is what I wanted.

src/runtime_src/core/include/ert.h
```diff
@@ -569,10 +574,10 @@ ert_fill_copybo_cmd(struct ert_start_copybo_cmd *pkt, uint32_t src_bo,
   pkt->cu_mask[1] = 0;
   pkt->cu_mask[2] = 0;
   pkt->cu_mask[3] = 0;
-  pkt->src_addr_lo = src_offset;
+  pkt->src_addr_lo = (uint32_t)src_offset;
```
rozumx: Should this code be instead: (uint32_t)(src_offset & 0xFFFFFFFF).  Shouldn't be an issue, but it does prevent a potential 1's compliment  issue.

src/runtime_src/core/include/xrt.h
stsoe: Ok. I see what you are saying.   This struct is not well designed by using 'int'.  Should be explicit about size.

PR #2299, opened 2019-10-18
Fix race computing memidx for CU

stsoe:


stsoe:
Combined with PR #2311

PR #2305, opened 2019-10-18
Adjust command queue size

stsoe:
This is a temporary change to get 2RP running again after upping CQ size from 16K to 64K.
XRT relies on cq size from meta data, but ERT uses a constant. After 2RP change they are out-of-sync. Use the minimum of hard-coded constant and meta data advertised size.


stsoe:
retest this please

PR #2311, opened 2019-10-18
Don't release virtual CU context with OCL subdevice

stsoe:
Also, fix race computing memidx for  compute unit


stsoe:
retest this please

PR #2316, opened 2019-10-19
bandwidth test core dump

maxzhen:


AShivangi:
I removed that since that function was getting information from sysfs and putting it in the xcldevinfo2 stuct which we are trying to move away from. But seems some code is still trying to read of the struct, instead of directly from sysfs. I checked and this also fixes the getting dev name problem (for qdma) in bandwidth testcase. Let's merge this for now so that validate isn't broken for different shells while I try to figure out where xcldevinfo2 is still being used. Thanks for debugging!

PR #2319, opened 2019-10-19
Changing the zocl/xrt recipes to use 2019.2 branch instead of master

chvamshi-xilinx:
Hi,

This change contains a fix to zocl/xrt recipes to use xrt 2019.2 branch instead of master.
This is not a final fix as we need to use tags once 2019.2 is released.
Will need to find out a scalable solution for this instead of changing this every release.


mamin506:
> Hi,
> 
> This change contains a fix to zocl/xrt recipes to use xrt 2019.2 branch instead of master.
> This is not a final fix as we need to use tags once 2019.2 is released.
> Will need to find out a scalable solution for this instead of changing this every release.

@chvamshi-xilinx After Petalinux 2019.2 released, we could remove these recipes folder from XRT repo. Since these are built-in.

PR #2323, opened 2019-10-21
mutex lock for live message function

sahilg-xilinx:
Mutex lock for live message function, as it gets accessed by two threads

It is to avoid race around condition happening in some corner scenarios


maxzhen:
ok to test

abalasa:
retest this please

PR #2324, opened 2019-10-21
read p2p_enabled as int

AShivangi:


houlz0507:
@AShivangi I approved change. Please make sure you have verified it.

maxzhen:
"p2p_enabled" is really confusing, it's actually a state, not just enabled or disabled. i'll merge it for now. but we probably should consider a better name so that it won't break again.

PR #2334, opened 2019-10-22
"help" command return 0

iomartin:
Issuing `xbutil help` should return 0, as it is not an error. This also applies to the `awssak` utility.


maxzhen:
ok to test

PR #2337, opened 2019-10-22
CR-1018070 update clock freq table

chienwei-lan:
CR-998101
CR-1018070


abalasa:
retest this please

cellery:
Did you try these new frequencies in HW to verify the configuration? You can use the frequency counter blocks to check the output frequency for each setting.

chienwei-lan:
@cellery  Yes, I've tested all 3 clocks(kernel, data, hbm). 

The frequency increase in increments of 5MHz,  from 10MHz to the clock limit.

PR #2339, opened 2019-10-22
CRs for EoU

AShivangi:
CR-1044842: xbutil mem read should be able to tell what sort of memory is being read
CR-1044030: XRT: enhance dmatest warning message when bitstream not loaded


abalasa:
retest this please

PR #2352, opened 2019-10-25
fix context switch on versal

larry9523:
This is a fix to support swapping xclbin on versal before we have icap subdev available. By caching the mem_topology in xdev, we can clear mem and reintialize memory every time an xclbin is downloaded.

Also, this PR added support to download BITSTREAM_PARTIAL_PDI

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_ioctl.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_ioctl.c
@@ -385,10 +385,15 @@ xocl_read_axlf_helper(struct xocl_drm *drm_p, struct drm_xocl_axlf *axlf_ptr)
 	}
 
 	/* Switching the xclbin, make sure none of the buffers are used. */
-	if (!preserve_mem && !XOCL_DSA_IS_VERSAL(xdev)) {
+	if (!preserve_mem) {
 		err = xocl_cleanup_mem(drm_p);
 		if (err)
 			goto done;
+
+		if (XOCL_DSA_IS_VERSAL(xdev)) {
+			vfree(xdev->mem_topo);
+			xdev->mem_topo = NULL;
+		}
 	}
 
 	if (!XOCL_DSA_IS_VERSAL(xdev)) {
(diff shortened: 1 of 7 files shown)


maxzhen:
Why we cannot have icap subdevice for versal? It's there already and can be used right away.

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_ioctl.c
```diff
@@ -411,16 +416,17 @@ xocl_read_axlf_helper(struct xocl_drm *drm_p, struct drm_xocl_axlf *axlf_ptr)
 			(void) xocl_pci_rbar_refresh(xdev->core.pdev,
 					xdev->p2p_bar_idx);
 		}
+	} else if (!preserve_mem) {
+		xdev->mem_topo = vmalloc(size);
```
xdavidz: 393 we freed mem_topo for versal, so we malloc it at 420, what about non-versal case? should we still alloc mem here?

mamin506:
> Why we cannot have icap subdevice for versal? It's there already and can be used right away.

AFAIK, Versal has no ICAP support. All of the configuration, for PS/PL/AIE should be done by PLM.

larry9523:
> > Why we cannot have icap subdevice for versal? It's there already and can be used right away.
> 
> AFAIK, Versal has no ICAP support. All of the configuration, for PS/PL/AIE should be done by PLM.

We need a better design to either reuse ICAP to cache all xclbin metadata but not download bitstream for versal or decouple xclbin metadata caching from ICAP. But that is post 2019.2.

maxzhen:
> 
> 
> > > Why we cannot have icap subdevice for versal? It's there already and can be used right away.
> > 
> > 
> > AFAIK, Versal has no ICAP support. All of the configuration, for PS/PL/AIE should be done by PLM.
> 
> We need a better design to either reuse ICAP to cache all xclbin metadata but not download bitstream for versal or decouple xclbin metadata caching from ICAP. But that is post 2019.2.

We do not have icap on user pf, but we still have icap subdevice enabled on user pf. this way, we share the meta data management code b/w mgmt pf and user pf. we should have a talk about how this can be done for versal.

larry9523:
> > > > Why we cannot have icap subdevice for versal? It's there already and can be used right away.
> > > 
> > > 
> > > AFAIK, Versal has no ICAP support. All of the configuration, for PS/PL/AIE should be done by PLM.
> > 
> > 
> > We need a better design to either reuse ICAP to cache all xclbin metadata but not download bitstream for versal or decouple xclbin metadata caching from ICAP. But that is post 2019.2.
> 
> We do not have icap on user pf, but we still have icap subdevice enabled on user pf. this way, we share the meta data management code b/w mgmt pf and user pf. we should have a talk about how this can be done for versal.

Agree and that is the plan too. But this might not be this PR. So far, we don't have management driver enabled on versal. (due to the limited MSI-X capability)

PR #2357, opened 2019-10-26
sched_exec should use zocl_xclbin to interact with xclbin.h

xdavidz:
as we discussed before, move xclbin related code outside of scheduler. In the future, we should make the scheduler code only for scheduling.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
@@ -942,3 +942,26 @@ zocl_xclbin_accel_adapter(int kds_mask)
 {
 	return kds_mask == ACCEL_ADAPTER;
 }
+
+/*
+ * returns false if any of the cu doesnt support interrupt
+ */
+bool
+zocl_xclbin_cus_support_intr(struct drm_zocl_dev *zdev)
+{
+	struct ip_data *ip;
+	int i;
+
+	if (!zdev->ip)
+		return false;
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -584,6 +584,82 @@ zocl_xclbin_refcount(struct drm_zocl_dev *zdev)
...
+
+		if (!strstr(md->m_tag, "MIG")) {
+			memp->zm_type = ZOCL_MEM_TYPE_CMA;
+			continue;
+		}
+
+		memp->zm_mm = vzalloc(sizeof(struct drm_mm));
+		memp->zm_type = ZOCL_MEM_TYPE_PLDDR;
+
+		drm_mm_init(memp->zm_mm, memp->zm_base_addr, memp->zm_size);
```
mamin506: The drm_mm_init() is DRM API. I don't think it is necessary to move init mem to xclbin.c

xdavidz: Agree, let's discuss more about this, for now, I just revoke the changes in zocl_bo.c. The issue is that we still have #include xclbin.h elsewhere in zocl_bo.c and zocl_sysfs.c. We might have 2 choices: 1) include it once and everyone uses it; 2) using zocl_xclbin.h as the only interface to interact with xclbin internal data;

PR #2358, opened 2019-10-27
Core based scheduling

stsoe:
Change KDS to managing execution cores (devices) rather than commands.
Each core manages commands but only iterates running commands to check
for completion.

Also use std::bitset in ERT firmware, but avoid throwing std::bitset
member functions (thanks to Stefan).


stsoe:
Problems with MP

PR #2367, opened 2019-10-29
xbutil2: Refactored basic platform query functions to a common API abstraction library

rozumx:


src/runtime_src/core/common/core_device.h
```diff
@@ -0,0 +1,45 @@
...
+ *    xrt_core_static library.
+ */
+namespace xrt_core {
+namespace device {
+  void get_devices(boost::property_tree::ptree &_pt);
+  void get_device_pcie_info(uint64_t _deviceID, boost::property_tree::ptree &_pt);
+  void get_device_rom_info(uint64_t _deviceID, boost::property_tree::ptree & _pt);
+  void get_device_xmc_info(uint64_t _deviceID, boost::property_tree::ptree & _pt);
+  void get_device_platform_info(uint64_t _deviceID, boost::property_tree::ptree & _pt);
+}
```
maxzhen: It's not clear how do we decide what info goes into which API here. E.g., the serial number (implemented below in core_device.cpp)  is obtained from get_device_xmc_info(), but that seems to be part of platform info here? It seems like we go by subdevices (rom/xmc)? I'm not sure if this is ideal since we don't have these subdevices on Windows... Maybe we can have a chat?

rozumx: Hello Max, 
As you are aware, there are a lot of moving parts here and I did an initial "raw" guesstimate on how this information might be gathered.   By all means, if you have an idea on how this data should be organized, lets do chat so that we can organize it appropriately. 

Note
Being that for xbutil2, there is both refactoring and replacement coding happening at the same time, I am expecting that we will be re-examining methods/functions several times throughout the release.  The good news is that the new code is rather modular so making these changes are really simple and easy.  So keep on questioning and asking.

src/runtime_src/core/common/core_device.h
```diff
@@ -0,0 +1,45 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef CORE_DEVICE_H
+#define CORE_DEVICE_H
+
+// Please keep eternal include file dependencies to a minimum
+#include <boost/property_tree/ptree.hpp>
+#include <stdint.h>
```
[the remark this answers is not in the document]
rozumx: Done - Learned something new today.

src/runtime_src/core/pcie/linux/core_system.cpp
```diff
@@ -28,6 +28,8 @@
 
 #include <chrono>
 
+#include "scan.h"
+
 
 static std::string driver_version(std::string driver)
```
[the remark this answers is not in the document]
rozumx: Done - Recommend changes made.

src/runtime_src/core/common/core_device.h
maxzhen: OK, understood. It just seems that we don't want to tie this interface with subdevice which is implementation details. Yes, feel free to merge for now and we can chat along the way.

PR #2368, opened 2019-10-30
CR-1047439 multi_process example fails for zcu102_base_dfx hardware board run

xdavidz:
multi_process example fails for zcu102_base_dfx hardware board run

somehow, the mutex_init is lost during code migration from other development branches.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_drv.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_drv.c
@@ -771,7 +771,7 @@ static int zocl_drm_platform_probe(struct platform_device *pdev)
 
 	ret = drm_dev_register(drm, 0);
 	if (ret)
-		goto err0;
+		goto err_drm;
 
 	/* During attach, we don't request dma channel */
 	zdev->zdev_dma_chan = NULL;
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/edge/drm/zocl/zocl_drv.c
```diff
@@ -788,13 +789,17 @@ static int zocl_drm_platform_probe(struct platform_device *pdev)
 	/* Initial sysfs */
 	rwlock_init(&zdev->attr_rwlock);
 	ret = zocl_init_sysfs(drm->dev);
-	if (ret)
+	if (ret) {
+		mutex_destroy(&zdev->zdev_xclbin_lock);
```
larry9523: I think you need to clean the error label and error out sequence. For example, if zocl_xclbin_init fail, you will still call zocl_xclbin_fini() where might panic the system because zdev->zdev_xclbin is not valid. After clean up, you may be able to put mutex_destroy(xclbin_lock) into one of the error label.

abalasa:
retest this please

src/runtime_src/core/edge/drm/zocl/zocl_drv.c
xdavidz: thanks, will get this fixed and retest failure cases today.

xdavidz: fixed

xdavidz:
retest this please!

PR #2376, opened 2019-10-31
fix context switch on versal

larry9523:
Cherry pick #2352(fix context switch on versal) to 2019.2 branch. This is needed for versal DFX for 2019.2
I've run test on v350 board with two xclbin vadd_rtl_xclbin and ad1.xclbin. These two have differnent memory connectivity. Without the fix, the second run will fail with bad addloc. With the fix, both test cases can be run many times.

Also a regression test passed on U200 board.


xdavidz:
retest this please

larry9523:
retest this please

PR #2381, opened 2019-11-01
libqdma update: allow non-STMC DSAs and reduce completion status requests.

karenx-xilinx:
…T is not supported.

2. group completion status request instead of on every EOT.


maxzhen:
retest this please

PR #2383, opened 2019-11-01
Add xclUnmapBO to xrt.h

stsoe:
Call xclUnmapBO when unmapping previously mapped BO.


stsoe:
Closing because of windows build errors that must first be resolved

PR #2392, opened 2019-11-04
Skip zocl clinet on zocl/ert mode.

larry9523:
Zocl client is not used on zocl/ert mode. We need to skip client operation when we are running ert mode. Host XRT will be responsible for context switch.


larry9523:
retest this please

larry9523:
retest this please

PR #2398, opened 2019-11-05
u50 1rp golden

xdavidz:
xocl code support xbmgmt flash --scan for u50 golden.


xdavidz:
xocl mailbox has some issues. need further debug

PR #2404, opened 2019-11-06
fix versal DFX

larry9523:
The xclbin_buffer is PS memory that is reserved for host application. These memory can NOT be mapped as cacheable because no cache coherence is guaranteed when host DMA from/to these memory.


xdavidz:
Thanks for fixing this. https://patchwork.kernel.org/patch/7956371/

PR #2405, opened 2019-11-06
Show fallthrough intent in ps_ert_query() to suppress GCC warning

sonals:


src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -2431,6 +2431,7 @@ ps_ert_query(struct sched_cmd *cmd)
 	case ERT_EXEC_WRITE:
 		if (!cu_done(cmd))
 			break;
+                __attribute__ ((fallthrough));
 		/* pass through */
```
larry9523: Looks good. I am wondering if simply changing /* pass through */ to /* fall through */ should work too.

PR #2406, opened 2019-11-06
windows: core_system: remove dependency on atls library

rajkumar-xilinx:
Removed dependency on ATLs library in core_system.cpp


src/runtime_src/core/pcie/windows/core_system.cpp
```diff
@@ -17,7 +17,9 @@
 
 #include "common/core_system.h"
 #include "gen/version.h"
-#include <atlstr.h>
+#include <windows.h>
+
+#define BUFFER 128
```
rozumx: Thank you for "defining" the magic variables.  
With you next submission, could you move it closer to where it is used (e.g., inside of the get_os_info() function).  

Something to think about
You might wish to define this value as a "static const unsigned int" inside of the method.   Doing so will allow anyone debugging this code to easily  determine its value.

src/runtime_src/core/pcie/windows/core_system.cpp
```diff
@@ -92,15 +66,22 @@ static std::string getmachinename()
 void xrt_core::system::get_os_info(boost::property_tree::ptree &_pt)
 {
   char tnow[26];
```
rozumx: With the next release, could you please define the "magic 26" value.

src/runtime_src/core/pcie/windows/core_system.cpp
```diff
@@ -92,15 +66,22 @@ static std::string getmachinename()
...
 
   ctime_s(tnow, sizeof tnow, &result);
-  _pt.put("sysname", GetStringFromReg(HKEY_LOCAL_MACHINE,
-				"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", "ProductName").c_str());
-  _pt.put("release", GetStringFromReg(HKEY_LOCAL_MACHINE,
-				"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", "BuildLab").c_str());
-  _pt.put("version", GetStringFromReg(HKEY_LOCAL_MACHINE,
-				"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", "CurrentVersion").c_str());
+
+  RegGetValueA(HKEY_LOCAL_MACHINE, "SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion", "ProductName", RRF_RT_ANY, NULL, (PVOID)&value, &BufferSize);
```
rozumx: In the previous version of this submission, you had encapsulated the calls with a helper function.  You might wish to consider doing the same here and add the following DRCs:
1) Validate there is indeed a return value (e.g., the BufferSize is NOT zero ;)
2) Report a default value or no entry at all if the call failed.

PR #2410, opened 2019-11-07
support U30 reset

mamin506:
Add Processor System(PS) sub device
Add scratchpad register
Add two types of reset (--softkernel and --ert) in xbmgmt

Not support yet: No sysfs nodes like microblaze sub device.


src/runtime_src/core/pcie/driver/linux/include/mgmt-ioctl.h
```diff
@@ -71,6 +71,8 @@ enum XCLMGMT_IOC_TYPES {
 	XCLMGMT_IOC_FREQ_SCALE,
 	XCLMGMT_IOC_OCL_RESET,
 	XCLMGMT_IOC_HOT_RESET,
+	XCLMGMT_IOC_SOFTKERNEL_RESET,
```
chienwei-lan: New enum should be the latest one, to not to break compatibility

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ps.c
```diff
@@ -0,0 +1,202 @@
+/*
+ * A GEM style device manager for PCIe based OpenCL accelerators.
```
maxzhen: overall, yes, but doesn't apply to ps.c

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ps.c
```diff
@@ -0,0 +1,202 @@
...
+ *
+ * Copyright (C) 2016-2018 Xilinx, Inc. All rights reserved.
```
maxzhen: pls fix copyright year.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ps.c
```diff
@@ -0,0 +1,202 @@
...
+	xocl_info(&pdev->dev, "Wait Processor System ready...");
+	ps = platform_get_drvdata(pdev);
+	if (!ps)
+		return;
+
+	/*
+	 * The PS need about 30 ~ 40 seconds for reboot
+	 * 50 seconds are more than enough
+	 */
+	msleep(50000);
```
maxzhen: why not poll the status every, say, 5 seconds, so that you might finish waiting earlier, if the reboot is done for < 50 seconds?

mamin506: Done.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ps.c
mamin506: Fixed

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ps.c
mamin506: I got this from microblaze.c. Anyway, I have updated ps.c

mamin506:
> We should probably consolidate all resets into one sysfs node and have users echo the reset kind to the node? Having a unique ioctl cmd for each reset type looks like an overkill.

Done.

mamin506:
> the ps_reset routine is exposed to userland code via IOCTLs. It'd be safer if you can serialize all reset requests with a mutex, or just refuse a reset request and return -EBUSY when another reset is already going on.
> 
> make sure you test these new xbmgmt reset options on a non-u30 board and handles the error gracefully.

Test on U250 board, --softkernel and --ert is no-op. "--hot" works good.

mamin506:
Fix conflict. Re-test on U30.

PR #2356, opened 2019-10-25
get_header_from_dtb can fail due to[C ioread8 may return -1(ff)

xdavidz:
without the fix:
[root@xsjhemn41 ~]# xbmgmt flash --scan --verbose
Card [0001:45:00.0]
    Card type:          u200
    Flash type:         SPI
    Flashable partition running on FPGA:
        xilinx_u200_xdma_201920_1,[ID=0xffffff9fffffff5e]
            Logic UUID:
            ffffff9fffffff5effffff40ffffffb5ffffffceffffffd1ffffffc8ffffffac
    Flashable partitions installed in system:
        xilinx_u200_gen3x16-ea_blp,[ID=0x272eb49f3c9fb05e]
            Logic UUID:
            272eb49f3c9fb05e27db9f40f01419b5f8c245ce1cba59d11f8b97c8787e96ac

with the fix:
[root@xsjhemn41 ~]# xbmgmt flash --scan --verbose
Card [0001:45:00.0]
    Card type:          u200
    Flash type:         SPI
    Flashable partition running on FPGA:
        xilinx_u200_gen3x16-ea_blp,[ID=0x272eb49f3c9fb05e]
            Logic UUID:
            272eb49f3c9fb05e27db9f40f01419b5f8c245ce1cba59d11f8b97c8787e96ac
    Flashable partitions installed in system:
        xilinx_u200_gen3x16-ea_blp,[ID=0x272eb49f3c9fb05e]
            Logic UUID:
            272eb49f3c9fb05e27db9f40f01419b5f8c245ce1cba59d11f8b97c8787e96ac

PR #2411, opened 2019-11-07
get_header_from_dtb can fail due to ioread8 may return -1(ff) (#2356)

xdavidz:
This is needed for powerpc with ubuntu, otherwise system will panic so hard.


maxzhen:
@houlz0507 , we need your fix for Travis CI build in 2019.2 as well...

PR #2412, opened 2019-11-07
Save xmc board info only if everything is fine

chienwei-lan:


chienwei-lan:
root@alveo:/opt/xilinx/xrt-2.3.1301/driver/xocl/subdev# xbutil validate -q
INFO: Found 1 cards

INFO: Validating card[0]: xilinx_u200_xdma_201920_1
INFO: == AUX power connector check PASSED

PR #2415, opened 2019-11-07
fix versal DFX (#2404)

larry9523:
Cherry-pick this to support Versal DFX in 2019.2. 
Test done:
Build two xclbin with different kernel.cl and switch them. Without the fix, the second download will fail. And with the fix, the two xclbin can be switched many times without issue.

This is a very low risk backport because it is only on the softkernel configuration (download xclbin) data path, which is only used by Versal.


abalasa:
retest this please

larry9523:
retest this please

PR #2416, opened 2019-11-08
CR-1048424 add xmc version

chienwei-lan:


maxzhen:
did you test new xocl with old xclmgmt and new xclmgmt with old xocl?

chienwei-lan:
new xocl with old xclmgmt -> send out mailbox request 10, get 0.

old xocl with new xclmgmt -> get 0.


As good as what we expect

PR #2417, opened 2019-11-08
Soft kernel command should have ERT_CTRL command type.

larry9523:
Softkernel should set ERT_CTRL command type. Otherwise, the command will fail to execute with unknonw command type.


larry9523:
retest this please

PR #2418, opened 2019-11-08
xrt-next.h: Add typedef to profiling related struct declaration 

rathaval:
Add typedef to struct declaration to make visible in the namespace

Several routines in this header refer to ProfileResults, however it was declared as
a simple struct. Without the typedef specifier it will not be resolved when using
an application that eventually needs compile with xclhal2.h

ProfileResults is invoked in the following functions xclCreateProfileResults,
xclGetProfileResults, and xclDestroyProfileResults

Typedef needs to be added to the definitions of KernelTransferData, CuExecData, and
StreamTransferData as they are used without the struct * being used in the
definition of ProfileResults.

Fixes: 9b40ccb HAL level profiling with counters for PCIe and Zynq (#1936)

### DIFF
--- a/src/runtime_src/core/include/experimental/xrt-next.h
+++ b/src/runtime_src/core/include/experimental/xrt-next.h
@@ -133,13 +133,13 @@ struct ProfileResults
   char* deviceName;
 
   uint64_t numAIM;
-  KernelTransferData* kernelTransferData;
+  struct KernelTransferData *kernelTransferData;
 
   uint64_t numAM;
-  CuExecData* cuExecData;
+  struct CuExecData *cuExecData;
 
   uint64_t numASM;
-  StreamTransferData* streamData;
+  struct StreamTransferData  *streamData;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


rathaval:
@stsoe 
Thanks for the review Soren.

I wish to stay away from the typedef v/s struct debate as it is not my place to decide that. Personally I prefer the Linux Kernel's coding guidelines of using structs so as not hide what the underlying type is.

1. I chose typedef as it needed the least number of lines/words/character changes. 

2. The other idea which crossed my mind is that any apps/libraries that rely on the interface being the same may have to recompile when we go to the foo(struct * ) version. Without knowing this dependency I thought it would be simplest to use typedef as I don't need to touch the function definitions at all. 

3.  Any app/library that relied on xrt-next.h would not compile to begin with. With that in mind I'll resend this with the structs being used instead of typedef.

Could you add the author of that commit as a reviewer so as to make sure [2] above is mitigated.
@IshitaGhosh : Could you take a look at this change as well ?

maxzhen:
ok to test

PR #2428, opened 2019-11-11
minor cleanup-1

sarab96:
minor cleanup-1

### DIFF
--- a/src/xma/include/lib/xma_utils.hpp
+++ b/src/xma/include/lib/xma_utils.hpp
@@ -22,10 +22,18 @@
 #include <map>
 
 namespace xma_core {
-   extern std::map<XmaSessionType, const char*> sessionMap;
+   static const std::map<XmaSessionType, const std::string> sessionMap = {
+      { XmaSessionType::XMA_SCALER, "scaler"},
+      { XmaSessionType::XMA_ENCODER, "encoder"},
+      { XmaSessionType::XMA_DECODER, "decoder"},
+      { XmaSessionType::XMA_FILTER, "filter"},
+      { XmaSessionType::XMA_KERNEL, "kernel"},
+      { XmaSessionType::XMA_ADMIN, "admin"},
+      { XmaSessionType::XMA_INVALID, "invalid"}
+   };
 
-   int32_t finalize_ddr_index(XmaHwKernel* kernel_info, int32_t req_ddr_index, int32_t* ddr_index, char* prefix);
-   int32_t create_session_execbo(XmaHwSessionPrivate *priv, int32_t count, char* prefix);
+   int32_t finalize_ddr_index(XmaHwKernel* kernel_info, int32_t req_ddr_index, int32_t& ddr_index, const std::string& prefix);
+   int32_t create_session_execbo(XmaHwSessionPrivate *priv, int32_t count, const std::string& prefix);
 }
 
 namespace xma_core { namespace utils {
--- a/src/xma/include/lib/xma_utils.hpp
+++ b/src/xma/include/lib/xma_utils.hpp
@@ -17,6 +17,17 @@
 #ifndef xma_utils_lib_h_
 #define xma_utils_lib_h_
 
+#include "plg/xmasess.h"
+#include "lib/xmahw_lib.h"
+#include <map>
+
+namespace xma_core {
+   extern std::map<XmaSessionType, const char*> sessionMap;
+
+   int32_t finalize_ddr_index(XmaHwKernel* kernel_info, int32_t req_ddr_index, int32_t* ddr_index, char* prefix);
+   int32_t create_session_execbo(XmaHwSessionPrivate *priv, int32_t count, char* prefix);
+}
+
 namespace xma_core { namespace utils {
 
 int32_t load_libxrt();
--- a/src/xma/src/xmaapi/xma_utils.cpp
+++ b/src/xma/src/xmaapi/xma_utils.cpp
@@ -20,6 +20,7 @@
 #include "app/xmalogger.h"
 #include "app/xmaparam.h"
 #include "lib/xmaapi.h"
+#include "lib/xmalimits_lib.h"
 #include "ert.h"
 #include "core/common/config_reader.h"
 #include "core/pcie/linux/scan.h"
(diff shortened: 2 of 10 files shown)


src/xma/include/lib/xma_utils.hpp
```diff
@@ -17,6 +17,17 @@
...
 
+#include "plg/xmasess.h"
+#include "lib/xmahw_lib.h"
+#include <map>
+
+namespace xma_core {
+   extern std::map<XmaSessionType, const char*> sessionMap;
+
+   int32_t finalize_ddr_index(XmaHwKernel* kernel_info, int32_t req_ddr_index, int32_t* ddr_index, char* prefix);
+   int32_t create_session_execbo(XmaHwSessionPrivate *priv, int32_t count, char* prefix);
```
rozumx: Again, look at what can be done to remove the dependency on pointers.

src/xma/src/xmaapi/xma_utils.cpp
```diff
@@ -28,11 +29,79 @@
...
 #include <iostream>
 #include <sstream>
+#include <bitset>
 
 #define XMAUTILS_MOD "xmautils"
 
 extern XmaSingleton *g_xma_singleton;
 
+namespace xma_core {
+    std::map<XmaSessionType, const char*> sessionMap = {
```
rozumx: Should be declared `static const`.  

in addition, it would be best to change the "char string" value to actual contain a string, for example:

`static const std::map<XmaSessionType, std::string> sessionMap = {`

src/xma/src/xmaapi/xma_utils.cpp
```diff
@@ -28,11 +29,79 @@
...
+        { XmaSessionType::XMA_ENCODER, "encoder"},
+        { XmaSessionType::XMA_DECODER, "decoder"},
+        { XmaSessionType::XMA_FILTER, "filter"},
+        { XmaSessionType::XMA_KERNEL, "kernel"},
+        { XmaSessionType::XMA_ADMIN, "admin"},
+        { XmaSessionType::XMA_INVALID, "invalid"}
+    };
+
+    int32_t finalize_ddr_index(XmaHwKernel* kernel_info, int32_t req_ddr_index, int32_t* ddr_index, char* prefix) {
+        *ddr_index = -1;
```
rozumx: I don't like magic numbers (e.g., numbers without any description).  Yes, I know that is it "obvious" that this is an invalid index, but it would be create to indicate so.  For example:
```
static const int32_t INVALID_INDEX = -1;
*ddr_index = INVALID_INDEX;
```

src/xma/src/xmaapi/xma_utils.cpp
```diff
@@ -28,11 +29,79 @@
...
+        if (kernel_info->soft_kernel) {
+            if (req_ddr_index != 0) {
+                xma_logmsg(XMA_WARNING_LOG, prefix, "XMA session with soft_kernel only allows ddr bank of zero\n");
+            }
+            //Only allow ddr_bank == 0;
+            *ddr_index = 0;
+            xma_logmsg(XMA_DEBUG_LOG, prefix, "XMA session with soft_kernel default ddr_bank: %d\n", *ddr_index);
+            return XMA_SUCCESS;
+        }
+        if (req_ddr_index < 0) {
```
rozumx: Going with the previous them, this line becomes:

`if (req_ddr_index == INVALID_INDEX) {`

src/xma/src/xmaapi/xma_utils.cpp
```diff
@@ -28,11 +29,79 @@
...
+            *ddr_index = kernel_info->default_ddr_bank;
+            xma_logmsg(XMA_DEBUG_LOG, prefix, "XMA session default ddr_bank: %d\n", *ddr_index);
+            return XMA_SUCCESS;
+        }
+        std::bitset<MAX_DDR_MAP> tmp_bset;
+        tmp_bset = kernel_info->ip_ddr_mapping;
```
rozumx: To prevent a copy of the std::bitset, one could use a reference, for example:

`std::bitset<MAX_DDR_MAP> & tmp_bset = kernel_info->ip_ddr_mapping;`

src/xma/src/xmaapi/xma_utils.cpp
```diff
@@ -28,11 +29,79 @@
...
+            xma_logmsg(XMA_DEBUG_LOG, prefix, "Using user supplied default ddr_bank. XMA session default ddr_bank: %d\n", *ddr_index);
+            return XMA_SUCCESS;
+        }
+        xma_logmsg(XMA_ERROR_LOG, prefix,
+            "User supplied default ddr_bank is invalid. Valid ddr_bank mapping for this CU: %s\n", tmp_bset.to_string().c_str());
+        return XMA_ERROR;
+    }
+
+    int32_t create_session_execbo(XmaHwSessionPrivate *priv, int32_t count, char* prefix) {
+        for (int32_t d = 0; d < count; d++) {
```
rozumx: While it might not make any difference regarding the amount of data you are working one, it is always better to perform a `++d` operation instead of `d++`.

src/xma/src/xmaapi/xmadecoder.cpp
```diff
@@ -170,44 +172,12 @@ xma_dec_session_create(XmaDecoderProperties *dec_props)
...
-                free(dec_session);
-                return NULL;
-            }
-        }
+    if (xma_core::finalize_ddr_index(kernel_info, dec_props->ddr_bank_index, 
+        &dec_session->base.hw_session.bank_index, (char*)XMA_DECODER_MOD) != XMA_SUCCESS) {
+        //Release singleton lock
+        g_xma_singleton->locked = false;
+        free(dec_session);
+        return NULL;
```
rozumx: Should be `nullptr `instead of `NULL`.

src/xma/include/lib/xma_utils.hpp
```diff
@@ -17,6 +17,17 @@
...
 #define xma_utils_lib_h_
 
+#include "plg/xmasess.h"
+#include "lib/xmahw_lib.h"
+#include <map>
+
+namespace xma_core {
+   extern std::map<XmaSessionType, const char*> sessionMap;
+
+   int32_t finalize_ddr_index(XmaHwKernel* kernel_info, int32_t req_ddr_index, int32_t* ddr_index, char* prefix);
```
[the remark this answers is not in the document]
sarab96: done. Rest will change later along with change of base structures

src/xma/include/lib/xma_utils.hpp
sarab96: done. Rest will change later along with change of base structures

src/xma/src/xmaapi/xma_utils.cpp
sarab96: done

src/xma/src/xmaapi/xma_utils.cpp
sarab96: done

src/xma/src/xmaapi/xma_utils.cpp
sarab96: keeping same for more flexibility

src/xma/src/xmaapi/xma_utils.cpp
sarab96: keeping same. uint64 to bitset conversion

src/xma/src/xmaapi/xma_utils.cpp
sarab96: keeping same. compilers now handle this properly

src/xma/src/xmaapi/xmadecoder.cpp
sarab96: done

PR #2432, opened 2019-11-12
u50 pf entry size should be ready from header instead of hard code

xdavidz:


src/runtime_src/core/pcie/driver/linux/xocl/xocl_subdev.c
```diff
@@ -1060,7 +1101,7 @@ xocl_subdev_vsec(xdev_handle_t xdev, u32 type,
 		off_high = ioread32(bar_addr + i + 4);
 		off = ((u64)off_high << 16) | (off_low & 0xffff0000) >> 16;
 		if (bar_idx)
-			*bar_idx = (off_low >> 8) & 0xff;
+			*bar_idx = (off_low >> 8) & 0xf0;
```
houlz0507: Should it be (off_low >> 12) & 0xf?

xdavidz: changed.

PR #2433, opened 2019-11-12
doc: xclRegWrite has been implemented in 2019.2

imrickysu:
- Also with some minor updates of "19.2" -> "2019.2"


sonals:
ok to test

superligen:
is this function thread safe？

maxzhen:
No, it does not prevent multiple threads from writing to the CU regs at the same time.

PR #2435, opened 2019-11-12
minor cleanup-2

sarab96:
minor cleanup-2

### DIFF
--- a/src/xma/include/lib/xma_utils.hpp
+++ b/src/xma/include/lib/xma_utils.hpp
@@ -22,8 +22,8 @@
 #include <map>
 
 namespace xma_core {
-   extern const std::map<XmaSessionType, const std::string> sessionMap;
-
+   std::string get_session_name(XmaSessionType eSessionType);
+   
    int32_t finalize_ddr_index(XmaHwKernel* kernel_info, int32_t req_ddr_index, int32_t& ddr_index, const std::string& prefix);
    int32_t create_session_execbo(XmaHwSessionPrivate *priv, int32_t count, const std::string& prefix);
 }
--- a/src/xma/include/lib/xma_utils.hpp
+++ b/src/xma/include/lib/xma_utils.hpp
@@ -22,15 +22,7 @@
 #include <map>
 
 namespace xma_core {
-   static const std::map<XmaSessionType, const std::string> sessionMap = {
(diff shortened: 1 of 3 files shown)


src/xma/src/xmaapi/xma_utils.cpp
```diff
@@ -37,6 +37,16 @@
 extern XmaSingleton *g_xma_singleton;
 
 namespace xma_core {
+    const std::map<XmaSessionType, const std::string> sessionMap = {
```
rozumx: The sessionMap value should be static and access to it should be done via a function.   Code support can be extremely difficult if every the signature for the map should change.

For pretty print functionality like this, the following code might be of interest:

```
std::string get_pretty_name(XmaSessionType eSessionType)
{
  std::map<XmaSessionType, const std::string>::const_iterator it = sessionMap.find(eSessionType);

  if (it == sessionMap.end()) {
    return std::string("Unknown XmaSessionType");
  }

  return it->second;
}
```

src/xma/include/lib/xma_utils.hpp
```diff
@@ -22,15 +22,7 @@
...
-   static const std::map<XmaSessionType, const std::string> sessionMap = {
-      { XmaSessionType::XMA_SCALER, "scaler"},
-      { XmaSessionType::XMA_ENCODER, "encoder"},
-      { XmaSessionType::XMA_DECODER, "decoder"},
-      { XmaSessionType::XMA_FILTER, "filter"},
-      { XmaSessionType::XMA_KERNEL, "kernel"},
-      { XmaSessionType::XMA_ADMIN, "admin"},
-      { XmaSessionType::XMA_INVALID, "invalid"}
-   };
+   extern const std::map<XmaSessionType, const std::string> sessionMap;
```
rozumx: It would be best to not refer to variables defined in other source files.  See below on how you might possible wish to address this.

src/xma/src/xmaapi/xma_utils.cpp
sarab96: excellent suggestion. done

src/xma/include/lib/xma_utils.hpp
sarab96: excellent suggestion. done

sarab96:
retest this please

PR #2436, opened 2019-11-13
reset fix and u50 reset workaround

houlz0507:


houlz0507:
separate this into two CRs

PR #2437, opened 2019-11-13
xbutil2: Complete abstraction library query functionality

rozumx:
Work Done
    ---------
    + Implemented multilay device_core <|-- device_pci <|-- device_linux or device_windows abstraction layers
    + Abstracted all query calls to enumeration values
    + Abstracted all sysdev or IOCTL calls to the lower level layers (e.g. device_linux and device_windows)
    + Added helper methods to collect query data and format it into a property table
    + Added stub body file non linux/window libraries that depends on core_common
    + On windows, the target executable is now xbutil
    + Refactored how sub-commands register themself to the xbutil main().
      - It is now cmake driven
    + Updated CMake to only build program, query, clock, reset, and version sub-commands
    + Sub-command help is now passed along with the other options


rozumx:
please retest

rozumx:
retest this please

src/runtime_src/core/tools/xbutil2/SubCmdFan.cpp
```diff
@@ -27,23 +27,33 @@ namespace po = boost::program_options;
 // System - Include Files
 #include <iostream>
 
+// ======= R E G I S T E R   T H E   S U B C O M M A N D ======================
+#include "SubCmd.h"
+static const unsigned int registerResult = 
+                    register_subcommand("fan",
```
maxzhen: just curious, we are able to set fan speed from command line?

src/runtime_src/core/tools/xbutil2/SubCmdP2P.cpp
```diff
@@ -27,14 +27,22 @@ namespace po = boost::program_options;
 // System - Include Files
 #include <iostream>
 
+// ======= R E G I S T E R   T H E   S U B C O M M A N D ======================
+#include "SubCmd.h"
+static const unsigned int registerResult = 
+                    register_subcommand("p3p",
```
maxzhen: you mean "p2p"?

src/runtime_src/core/tools/xbutil2/SubCmdRun.cpp
```diff
@@ -27,24 +27,34 @@ namespace po = boost::program_options;
 // System - Include Files
 #include <iostream>
 
+// ======= R E G I S T E R   T H E   S U B C O M M A N D ======================
+#include "SubCmd.h"
+static const unsigned int registerResult = 
+                    register_subcommand("run",
```
maxzhen: not sure what "run" is for?

src/runtime_src/core/tools/xbutil2/SubCmdScan.cpp
```diff
@@ -27,22 +27,35 @@ namespace po = boost::program_options;
 // System - Include Files
 #include <iostream>
 
+// ======= R E G I S T E R   T H E   S U B C O M M A N D ======================
+#include "SubCmd.h"
+static const unsigned int registerResult = 
+                    register_subcommand("scan",
```
maxzhen: in xbutil, list is the same as scan and i actually removed "list" from help message and only left scan. maybe we should do the same for xbutil2 as well?

abalasa:
retest this please

src/runtime_src/core/tools/xbutil2/SubCmdP2P.cpp
rozumx: Corrected subcommand name to be: p2p

src/runtime_src/core/tools/xbutil2/SubCmdScan.cpp
rozumx: Great to know!  Will make those changes as well.

src/runtime_src/core/tools/xbutil2/SubCmdFan.cpp
rozumx: The "fan" and "run" subcommands were added prior to examining to see if they are actually implemented.  I will remove them from consideration.

src/runtime_src/core/tools/xbutil2/SubCmdClock.cpp
```diff
@@ -74,7 +85,7 @@ int subCmdClock(const std::vector<std::string> &_options, bool _help)
   }
 
   // Check to see if help was requested or no command was found
-  if (_help == true)  {
+  if (help == true)  {
```
AShivangi: Is the bool_switch working? I tried the same implementation in xbmgmt, but it doesn't seem to work. It works under 2 scenarios:
1. change help=true in the beginning and this condition to check if help=false
2. or change --help to -h

PR #2449, opened 2019-11-14
2018.3

xuhz:
add aws reclock support in 2018.3


xuhz:
the added code is only built on AWS F1 node. How does it get failed in pipeline?
Please check the pipeline and retest it!

maxzhen:
i don't think we have pipeline support for branches other than master and 2019.2

PR #2453, opened 2019-11-16
Compile and link xdp.dll on Windows

IshitaGhosh:
1. Mostly redo-ing compilation and link support for xdp.dll on Windows, like done on XRTWindows branch
2. Use uint64_t for index argument in ProfileIP constructor to match uint64_t ip_index member.  Update constructors for all the classes for profiling IPs : AIM, AM, ASM, TS2MM, TraceFifoLite/Full, TraceFunnel
3. Add call to xclUnmapBO from xrt::hal2::device::unmap using hal2::operations. This helps in skipping direct calls to munmap Linux system call from components using xrt++ library. Those direct calls to munmap system call can now be replaced with xrt::device::unmap. This will resolve and call the applicable SHIM implementation of xclUnmapBO for Linux PCIe/Edge or for Windows
4. Export xrt_core::get_xrt_info for use in XDP. Link xrt_core dynamic library in addition to the xrt_core_static for xbutil2 (for Windows only).


src/runtime_src/xdp/profile/plugin/ocl/ocl_profiler.cpp
```diff
@@ -756,8 +759,7 @@ namespace xdp {
 
     dInt->resetTS2MM();
 
-    auto addr = xrtDevice->map(mDDRBufferForTrace);
-    munmap(addr, mDDRBufferSz);
+    xrtDevice->map(mDDRBufferForTrace);
```
stsoe: Why is map called when the returned ptr is not used?

IshitaGhosh: XRT PR build/test failure seemed to be some setup/infra issue. I wanted to confirm and trigger a rebuild. I have removed it now.

abalasa:
retest this please

src/runtime_src/core/tools/xbutil2/CMakeLists.txt
```diff
@@ -49,6 +49,7 @@ if(WIN32)
     Boost::filesystem 
     Boost::program_options 
     Boost::system 
+    xrt_core
```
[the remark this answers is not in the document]
IshitaGhosh: As discussed, if we want to link xbutil2.exe only to xrt_core.dll, we need to export core_device methods. Currently, only xbutil2 would need that. So for now, we will continue to link to xrt_core_static and xrt_core.

PR #2457, opened 2019-11-18
Add validate of ctrl commands

stsoe:
Coverity issues


src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
```diff
@@ -4294,6 +4299,45 @@ reconfig(struct platform_device *pdev)
 	return 0;
 }
 
+static int
+validate_ctrl(struct platform_device *pdev, struct ert_packet *ecmd)
+{
```
sonals: const struct ert_packet *ecmd?

stsoe: ok

PR #2474, opened 2019-11-21
CR-1049309 1RP Recovery flash doesn't have XMC

xdavidz:
This is the fix for:
CR-1049309 Flashing a BLP while the U50 recovery image is loaded is failing

Basically, the U50 1RP Recovery Golden image doesn't have XMC and also we forget to set the mfg flag in the driver to avoid further probing work failed. When we set the mfg flag in the xclmgmt_probe and exit probe earlier for Golden (mfg), we can flash the card.

To avoid XMC warning in CLI, we use sysfs to mark this as mfg_no_xmc supported.

Verified works:

xbmgmt flash --scan
Card [0001:58:00.0]
    Card type:          u50
    Flash type:         SPI
        xilinx_u50_Recovery BLP
        xilinx_u50_gen3x4-1rp_blp,[ts=0x6bfd439eb11d8d8d]

xbmgmt flash --update --shell xilinx_u50_gen3x4-1rp_blp
         Status: shell needs updating
         Current shell: xilinx_u50_Recovery BLP
         Shell to be flashed: xilinx_u50_gen3x4-1rp_blp
Are you sure you wish to proceed? [y/n]: y

Updating shell on card[0001:58:00.0]
INFO: ***Found 295 ELA Records
Idcode byte[0] ff
Idcode byte[1] 20
Idcode byte[2] bb
Idcode byte[3] 21
Idcode byte[4] 10
Enabled bitstream guard. Bitstream will not be loaded until flashing is finished.
Erasing flash....


xdavidz:
@houlz0507 Redo the code, removed sysfs node no_xmc. However I cannot find a cleaner way to suppress warning message for !mfg with XMC issues of old non-vsec cards. So this fix will only fix the flash issue. They are eager to have flash from golden works. We will need a sophistic/better way to handle the warning from xbmgmt CLI.

PR #2481, opened 2019-11-24
embedded context switch operation needs attr_rwlock, and rephrase con…

xdavidz:
1) ctx_ioctl needs write lock to protect some data, in the future we'd better dump some info into sysfs, like live session etc.
2) rephrase context switch comments


xdavidz:
Could anyone merge this please. thanks!

PR #2488, opened 2019-11-26
enable cmc mutex handshake for ulp new feature resources

xdavidz:
This is the effort for enabling extra handshake for u50 1rp blp, we will need to interact with CMC by extra operations, say CMC Release and CMC Grant.
The CMC Release is needed prior to download BLP/PLP/ULP, it will set and then get response from CMC so that CMC is in status for downloading new dsabin or xsabin.

The CMC Grant is needed after download ULP only. It will tell CMC if there is new features bring up after the xclbin download, CMC should acknowledge by setting ack.

Unit Test:
1) factory reset, flash for regular xdma u50 shell and new 1rp_blp u50 shell;
2) verified xbmgmt program on both xdma u50 (download successfully) and 1rp_blp_u50 (CR http://jira.xilinx.com/browse/CR-1049207).

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-utils.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-utils.c
@@ -547,6 +547,18 @@ int xclmgmt_program_shell(struct xclmgmt_dev *lro)
 
 	xocl_thread_stop(lro);
 
+	/*
+	 * cmc has its own freeze operation, once ert is also enabled, we need
+	 * refactor code to a register design pattern, the freeze operation
+	 * will loop all registerred module and freeze them one by one, for
+	 * example: cmc, ert, etc.
+	 */
+	ret = xocl_cmc_freeze(lro);
+	if (ret) {
+		mgmt_err(lro, "release CMC access failed %d", ret);
+		goto failed;
+	}
+
 	xocl_mb_stop(lro);
 
 	ret = xocl_subdev_destroy_prp(lro);
--- a/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-utils.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-utils.c
@@ -547,18 +547,6 @@ int xclmgmt_program_shell(struct xclmgmt_dev *lro)
 
 	xocl_thread_stop(lro);
 
-	/*
-	 * cmc has its own freeze operation, once ert is also enabled, we need
-	 * refactor code to a register design pattern, the freeze operation
-	 * will loop all registerred module and freeze them one by one, for
-	 * example: cmc, ert, etc.
-	 */
-	ret = xocl_cmc_freeze(lro);
-	if (ret) {
-		mgmt_err(lro, "release CMC access failed %d", ret);
-		goto failed;
-	}
-
 	xocl_mb_stop(lro);
 
 	ret = xocl_subdev_destroy_prp(lro);
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
@@ -2298,6 +2298,10 @@ static int __icap_download_bitstream_axlf(struct platform_device *pdev,
 	ICAP_INFO(icap, "incoming xclbin: %pUb\non device xclbin: %pUb",
 		&xclbin->m_header.uuid, &icap->icap_bitstream_uuid);
 
+	err = xocl_cmc_freeze(xdev);
+	if (err)
+		goto done;
+
 	xocl_subdev_destroy_by_level(xdev, XOCL_SUBDEV_LEVEL_URP);
 	icap_refresh_addrs(pdev);
 
@@ -2356,6 +2360,9 @@ static int __icap_download_bitstream_axlf(struct platform_device *pdev,
 			reg_wr(icap->icap_ucs_control + 8, 1);
 		}
 	}
+
+	if (ICAP_PRIVILEGED(icap))
+		err = xocl_cmc_free(xdev);
 
 done:
 	if (err) {
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
@@ -89,6 +89,7 @@
 #define	XMC_HOST_MSG_OFFSET_REG		0x300
 #define	XMC_HOST_MSG_ERROR_REG		0x304
 #define	XMC_HOST_MSG_HEADER_REG		0x308
+#define	XMC_HOST_NEW_FEATURE_REG1	0xB20
 
 #define	VALID_ID			0x74736574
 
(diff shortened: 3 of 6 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-utils.c
```diff
@@ -547,6 +547,13 @@ int xclmgmt_program_shell(struct xclmgmt_dev *lro)
...
 	xocl_thread_stop(lro);
 
+	/* when cmc release is eligible, release cmc prior to stop it */
+	ret = xocl_cmc_release_access(lro);
+	if (ret) {
+		mgmt_err(lro, "release CMC access failed %d", ret);
+		goto failed;
+	}
+
 	xocl_mb_stop(lro);
```
maxzhen: should we make xocl_cmc_release_access() as part of xocl_mb_stop()?

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
```diff
@@ -2281,6 +2281,10 @@ static int __icap_download_bitstream_axlf(struct platform_device *pdev,
 	ICAP_INFO(icap, "incoming xclbin: %pUb\non device xclbin: %pUb",
 		&xclbin->m_header.uuid, &icap->icap_bitstream_uuid);
 
+	err = xocl_cmc_release_access(xdev);
```
maxzhen: is this needed when we downloaded xclbin? i thought we only need it for plp download?

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
```diff
@@ -2319,6 +2323,9 @@ static int __icap_download_bitstream_axlf(struct platform_device *pdev,
 			xocl_subdev_create(xdev, &subdevs[num_dev].info);
 	}
 
+	if (ICAP_PRIVILEGED(icap))
```
maxzhen: why we don't have if (ICAP_PRIVILEGED(icap)) when we call xocl_cmc_release_access() above?

xdavidz:
> A couple of general questions:
> 
> 1. the new API names sounds strange to me. "release" and "update" are not symmetric, but what they did are exactly symmetric. Do we want to call them "freeze" and "free"?
I am ok with freeze/free, @houlz0507 what do you think?

> 2. What is the grant status when cmc is started?
https://confluence.xilinx.com/display/DCG/CMC+Subsystem+v2.1#CMCSubsystemv2.1-4.3Mutexhandshakefunction
Based on the spec above, so I think the grant status should be 0, then we download BLP/PLP/ULP, and only when ULP has new features detected, we should set grant status to 1.

>    I have a few comments inline as well.

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-utils.c
xdavidz: @houlz0507 think we should somehow get ride of mb_stop for blp? which means we should remove the xocl_mb_stop of those GPIO manipulation?  The current solution is to preserve previous code unchanged in case of any regression. Yes, we can move the new api into the xocl_mb_stop.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
xdavidz: my understanding is that we need this for blp/plp and ulp, please correct me if I am wrong.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
xdavidz: The code above handles mgmt and user both, line 2327 depends on loaded new fdt line (2320 - 2324). So a minimum change is to add if(ICAP_PRIV...).

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
houlz0507: With u50, CMC might be accessing user region. In this case, the accessing need to be stopped before downloading xclbin.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
```diff
@@ -2249,6 +2249,75 @@ static void xmc_clk_scale_config(struct platform_device *pdev)
...
+		    ack, grant);
+		err = -EBUSY;
+		goto fail;
+	}
+
+	xocl_xdev_info(xdev, "%s CMC succeeded.", flags ? "Grant" : "Release");
+fail:
+	return err;
+}
+
```
houlz0507: xocl_iores_read32 and write may return -ENODEV if the resource is not found. Please add code to handle the case that cmc mutex does not exist.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
maxzhen: But, we already have AXI gate freeze and free to protect ULP. I guess that is not protecting CMC from accessing it?

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-utils.c
maxzhen: OK, if you already have a plan, then do whatever you feel like.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
maxzhen: Not sure I'm following...but we don't download xclbin on user side and we don't have CMC on user side. So, what we need to handle from user side?

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-utils.c
xdavidz: Discussed lizhi, I'd leave the name and location as is for this PR to enable the CMC operation. We'd have a better solution to handle all kinds of offline/online or freeze/free operations by a listener/register design module. So that only interested module will register themselves into the interest list, BLP/PLP/ULP download will simply call offline/online adapter, and the adapter will loop the interest list and call every callback.

maxzhen: sounds good

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
xdavidz: It's 'urgly' for now, we would clean this and axi freeze and future ert freeze together in next PR.

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-utils.c
xdavidz: > sounds good

Please wait for final testing done. I will request re-review again. Appreciate that.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
xdavidz: add -ENODEV checking

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-utils.c
xdavidz: Basic test done, xclbin download successful, reg of ep_cmc_mutex_00 is also readable.

xdavidz: > sounds good

I redone the code a little bit, change the name to freeze and free because we eventually will have a unique interface called freeze and free for all kinds of sub IPs, like cmc, ert, etc.

xdavidz:
yield for Lizhi's PR and reset.

xdavidz:
@maxzhen @houlz0507 merged with latest code, also cleanup the code so that xocl_mb_stop will take care of xocl_cmc_freeze. Tested on u50 and u280.

PR #2493, opened 2019-11-26
Windows tools refactor

AShivangi:
- refactored code into tools/common
 - renamed XBUtilMain and XBMgmtMain to XBMain in common and made small changes
 - created method stubs used in different options for xbmgmt flash (these are subject to change depending on their implementation)


src/runtime_src/core/tools/common/XBMain.cpp
```diff
@@ -1,157 +1,149 @@
...
+// 3rd Party Library - Include Files
+#include <boost/program_options.hpp>
+namespace po = boost::program_options;
+#include "boost/format.hpp"
+
+// System - Include Files
+#include <iostream>
+
+
+static void printHelp(po::options_description _optionDescription)
```
stsoe: I think argument should be 'const reference'

src/runtime_src/core/tools/xbmgmt2/CMakeLists.txt
```diff
@@ -15,9 +15,9 @@ include_directories(${CMAKE_BINARY_DIR})
 
 file(GLOB XBMGMT_V2_BASE_FILES
   "xbmgmt.cpp"
-  "XBMgmtMain.cpp"
-  "XBUtilities.cpp"
-  "SubCmd.cpp"
+  "../common/XBMain.cpp"
```
stsoe: consistency please.   how about  #include "tools/common/" like you do in other files?

PR #2494, opened 2019-11-26
Warn about missing sched.bin

stsoe:


maxzhen:
maybe we can save these ert binaries in prerm script and restore them in postinst if they are missing?

stsoe:
I think that is a good idea.  I will create a CR for that.

stsoe:
retest this please

abalasa:
retest this please

abalasa:
retest this please

PR #2495, opened 2019-11-27
Add feature ROM and mgmt function on versal

larry9523:
This PR included the following enahncement and bug fixes on versal
1. Add xclmgmt driver 
2. Add featuro ROM subdriver
3. fix an xdma issue on versal, where interrupt is not supported yet.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/lib/libxdma.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/lib/libxdma.c
@@ -3629,15 +3629,18 @@ void *xdma_device_open(const char *mname, struct pci_dev *pdev, int *user_max,
 	rv = probe_engines(xdev);
 	if (rv)
 		goto err_engines;
-	/* re-determine user_max */
-	xdev->user_max = min(xdev->user_max, pci_msix_vec_count(pdev) -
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/lib/libxdma.c
```diff
@@ -3629,15 +3629,18 @@ void *xdma_device_open(const char *mname, struct pci_dev *pdev, int *user_max,
...
+		xdev->user_max = min(xdev->user_max, pci_msix_vec_count(pdev) -
+				xdev->c2h_channel_max - xdev->h2c_channel_max);
+		if (xdev->user_max < 0) {
+			rv = -EINVAL;
+			pr_err("Invalid number of interrupts. "
+				"pci %d, h2c %d, c2h %d",
+				pci_msix_vec_count(pdev), xdev->h2c_channel_max,
+				xdev->c2h_channel_max);
+			goto err_enable_msix;
+		}
```
houlz0507: Is it possible to remove this condition from libxdma? libxdma is owned and maintained independently.  We need to avoid introducing any DSA concept to libxdma.

PR #2512, opened 2019-12-03
check sc only for u50 2rp flow

houlz0507:


src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -1251,7 +1251,6 @@ int xcldev::device::scVersionTest(void)
     if (!exp_sc_ver.empty() && sc_ver.compare(exp_sc_ver) != 0)
     {
         std::cout << "SC firmware version is not expected. SC firmware running on board: " << sc_ver << ". Expected SC firmware: " << exp_sc_ver << std::endl; 
-	return -EINVAL;
```
maxzhen: Please reformat the output msg following the examples in pcieLinkTest() and auxConnectionTest().
You want to return 1 when sc version does not match so that the xbutil validate can print out proper warning msg in the end. return 0 means success and xbutil validate will not treat it as warning.
These can be fixed as a separate PR.

houlz0507:
retest this please

PR #2515, opened 2019-12-03
bug fix and log msg change

sarab96:
bug fix and log msg change


maxzhen:
merge now since this is an urgent bug fix.

PR #2526, opened 2019-12-04
Update cu stat

mamin506:
For CR-1049515.
xbutil query would send a command to KDS/ERT to check the CU status.
The ERT/PS doesn't support this opcode and does nothing for unknown opcode.
This leads to hang when user call 'xbutil query' on a PCIe hybrid device.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/sched_exec.c
+++ b/src/runtime_src/core/edge/drm/zocl/sched_exec.c
@@ -2351,6 +2351,7 @@ penguin_query(struct sched_cmd *cmd)
 		mark_cmd_complete(cmd, ERT_CMD_STATE_COMPLETED);
 		break;
 	default:
+		mark_cmd_complete(cmd, ERT_CMD_STATE_ERROR);
 		DRM_ERROR("unknown opcode %d", opc);
 	}
 	SCHED_DEBUG("<- %s\n", __func__);
@@ -2552,6 +2553,7 @@ ps_ert_query(struct sched_cmd *cmd)
 		break;
 
 	default:
+		mark_cmd_complete(cmd, ERT_CMD_STATE_ERROR);
 		DRM_ERROR("unknown opcode %d", opc);
 	}
 	SCHED_DEBUG("<- %s()", __func__);
--- a/src/runtime_src/core/edge/drm/zocl/sched_exec.c
+++ b/src/runtime_src/core/edge/drm/zocl/sched_exec.c
@@ -33,6 +33,10 @@
 #define SCHED_UNUSED __attribute__((unused))
 #endif
 
+#ifndef ERT_VERSION
+#define ERT_VERSION 0
+#endif
+
 #define sched_error_on(exec, expr) \
 ({ \
 	unsigned int ret = 0; \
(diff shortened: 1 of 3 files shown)


xdavidz:
LGTM, the unknown op code cause the hung, the reset are for cu_state correct?

mamin506:
> LGTM, the unknown op code cause the hung, the reset are for cu_state correct?

correct.

abalasa:
retest this please

src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -911,6 +915,65 @@ configure(struct sched_cmd *cmd)
...
+	int pkt_idx = 0;
+	int max_idx = (slot_size(cmd->ddev) >> 2) - 1;
+	int i;
+
+	SCHED_DEBUG("-> %s cq_slot_idx %d\n", __func__, cmd->cq_slot_idx);
+	slot_idx = cmd->cq_slot_idx;
+	pkg = cmd->packet;
+
+	/* custat version, update when changing layout of packet */
+	pkg->data[pkt_idx++] = 0x51a10000;
```
larry9523: Just curious, where is the 0x51a10000 from?

src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -911,6 +915,65 @@ configure(struct sched_cmd *cmd)
...
+	/* individual CU status */
+	for (i = 0; i < exec->num_cus && pkt_idx < max_idx; ++i)
+		pkg->data[pkt_idx++] = exec->cu_status[i];
+
+	/* Command slot status
+	 * Hard code QUEUED state. When a NEW command is found,
+	 * ERT/PS will set it to QUEUED. Then no CQ write.
+	 */
+	for (i = 0; i < exec->num_slots && pkt_idx < max_idx; ++i)
+		pkg->data[pkt_idx++] = ERT_CMD_STATE_QUEUED;
```
larry9523: why can't we return the real slot status with idle or queued?

src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -2473,11 +2547,13 @@ ps_ert_query(struct sched_cmd *cmd)
 		__attribute__ ((fallthrough));
 		/* pass through */
 
+	case ERT_CU_STAT:
 	case ERT_CONFIGURE:
 		mark_cmd_complete(cmd, ERT_CMD_STATE_COMPLETED);
 		break;
 
 	default:
+		mark_cmd_complete(cmd, ERT_CMD_STATE_ERROR);
```
larry9523: For an unrecognized command, we probably want take care of it in penguin_submit and ps_ert_submit as well to mark submit error. Looks like then won't be queried here if they are not known command.

src/runtime_src/core/edge/drm/zocl/sched_exec.c
mamin506: https://github.com/Xilinx/XRT/blob/master/src/runtime_src/ert/scheduler/scheduler.cpp#L879

src/runtime_src/core/edge/drm/zocl/sched_exec.c
mamin506: ERT/PS will copy command from CQ to PS memory to avoid AXI Lite access to the PL. If the real slot is CQ, the status is queued. If the "real slot" is the copy in PS memory, then we need to iterate  queued command list(maybe pending list also) and update corresponding CQ slot. I just too lazy to implement it for now. Let's refine this in KDS/ERT 3.0.

src/runtime_src/core/edge/drm/zocl/sched_exec.c
mamin506: I think so. I would like to handle the unrecognized command at cq_check(). But I found I'll need to maintain a supported command list at cq_check() and maintain the consistency with query and submit.
The submit() could return "false" to just skip the unknown opcode. We need more changes to support submit error in current KDS implementation. 
This is the easiest way I can find to handle unknown opcode. Let's refine this at KDS/ERT 3.0.

larry9523: Can we just call mark_cmd_submit_error() in ps_ert_submit and penguin_submit()? But that could be a different PR.

src/runtime_src/core/edge/drm/zocl/sched_exec.c
larry9523: I was thinking to just check the cmd state in each CQ slot just like what we are doing in interate_packets. But I am ok to refine it later.

PR #2529, opened 2019-12-05
changing the max throughput for u2x4 board also. Initially added for

chvamshi-xilinx:
updating bandwidth for u2x4 platform also, earlier changed same for U2x4


AShivangi:
I already made this change

maxzhen:
Closing now since the fix is already merged.

PR #2535, opened 2019-12-05
Windows xbmgmt flash

AShivangi:
- runs in admin shell
 - takes an hour to flash!
 - xmc not tested, the board has a bad sc


maxzhen:
Takes an hour to flash? Hmm.....

stsoe:
Yes. we will have to look at the performance later.

PR #2537, opened 2019-12-06
xrt-next.h: Fix typo in function declaration of xclGetNumLiveProcesses()

rathaval:
In XRT/master commit e04c194 fixes this issue through PR #2294.
This needs to make its way to XRT/2019.2 to allow pure C applications
that using a configure or Cmake.

Typically, the build system compiles a test compile when a pkg-config
dependancy is added, which fails for the FFmpeg being used for the U30.

Fixes: fcadc99 Fix xclLogMsg va_list processing -- cannot call va_start()
	   twice on the same va_list


stsoe:
ok to test

PR #2538, opened 2019-12-06
lift restriction of mismatch SC version

xdavidz:
I finally got it tested on one of the machine. Is the xclbin valid warning message too much?

>SC firmware mismatch, ATTENTION
>SC firmware running on board: 4.0.7. Expected SC firmware from installed Shell: 4.2.0
>Using: xbmgmt flash --update --shell <xsabin file> to sync on board SC firmware
>WARN: == SC firmware version check PASSED with warning

dmesg:
ready_show: installed XSABIN has SC version: (4.2.0) mismatch with loaded SC version: (4.0.7).

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
+++ b/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
@@ -1250,7 +1250,9 @@ int xcldev::device::scVersionTest(void)
     pcidev::get_dev(m_idx)->sysfs_get("xmc", "exp_bmc_ver", errmsg, exp_sc_ver);
     if (!exp_sc_ver.empty() && sc_ver.compare(exp_sc_ver) != 0)
     {
-        std::cout << "SC firmware version is not expected. SC firmware running on board: " << sc_ver << ". Expected SC firmware: " << exp_sc_ver << std::endl; 
+        std::cout << "SC FIRMWARE MISMATCH, ATTENTION" << std::endl;
+        std::cout << "SC firmware running on board: " << sc_ver << ". Expected SC firmware from installed Shell: " << exp_sc_ver << std::endl;
+	std::cout << "Please use \"xbmgmt flash --scan\" to check installed Shell." << std::endl;
 	return 1;
     }
 
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_sysfs.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_sysfs.c
@@ -394,20 +394,16 @@ static ssize_t ready_show(struct device *dev,
 			goto bail;
 		xocl_xmc_get_data(xdev, XCL_BDINFO, board_info);
 		/*
-		 * Note:
-		 * bmc_ver can be cut into 5.0 instead of 5.0.6
-		 * a temporary workaround is to compare first 5.0 part,
-		 * platform should make sure bmc_ver is intact.
-		 *
-		 * with legacy mgmtpf driver, exp_bmc_ver will be NULL.
-		 * And we have to mark ready in this case
+		 * Lift the restriction of mis-matching SC version for
+		 * experienced user to manually update SC firmware than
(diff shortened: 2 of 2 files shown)


src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -1250,7 +1250,9 @@ int xcldev::device::scVersionTest(void)
     pcidev::get_dev(m_idx)->sysfs_get("xmc", "exp_bmc_ver", errmsg, exp_sc_ver);
     if (!exp_sc_ver.empty() && sc_ver.compare(exp_sc_ver) != 0)
     {
-        std::cout << "SC firmware version is not expected. SC firmware running on board: " << sc_ver << ". Expected SC firmware: " << exp_sc_ver << std::endl; 
+        std::cout << "SC firmware mismatch, ATTENTION" << std::endl;
+        std::cout << "SC firmware running on board: " << sc_ver << ". Expected SC firmware from installed Shell: " << exp_sc_ver << std::endl;
+	std::cout << "Using: xbmgmt flash --update --shell <xsabin file> to sync on board SC firmware" << std::endl;
```
houlz0507: Maybe adding --card <bdf>. <bdf> could be the real one.

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -1250,7 +1250,9 @@ int xcldev::device::scVersionTest(void)
     pcidev::get_dev(m_idx)->sysfs_get("xmc", "exp_bmc_ver", errmsg, exp_sc_ver);
     if (!exp_sc_ver.empty() && sc_ver.compare(exp_sc_ver) != 0)
     {
-        std::cout << "SC firmware version is not expected. SC firmware running on board: " << sc_ver << ". Expected SC firmware: " << exp_sc_ver << std::endl; 
+        std::cout << "SC firmware mismatch, ATTENTION" << std::endl;
```
maxzhen: This line should be all upper case.

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
maxzhen: this line should be a real one for user to just copy, paste and run.
will "sudo /opt/xilinx/xrt/bin/xbmgmt flash --update --card xxxx" do?

xdavidz: 1) I will make the change;
2) I found the other way to create this situation, which is just installing an older pkg with older SC FW version, so that even the board has newer FW, we still get this warning. I doubt that if we recommend user to just flash the card with the older pkg?
@maxzhen

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
xdavidz: OK

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
maxzhen: I think that we might want to cache the SC firmware version when driver is attached. Thus, it will not be impacted by what is on the disk.

xdavidz:
@houlz0507 @maxzhen what about now?

SC FIRMWARE MISMATCH, ATTENTION
SC firmware running on board: 4.0.7. Expected SC firmware from installed Shell: 4.2.0
Please use "xbmgmt flash --scan" to check installed Shell.
WARN: == SC firmware version check PASSED with warning

maxzhen:
> 
> 
> @houlz0507 @maxzhen what about now?
> 
> SC FIRMWARE MISMATCH, ATTENTION
> SC firmware running on board: 4.0.7. Expected SC firmware from installed Shell: 4.2.0
> Please use "xbmgmt flash --scan" to check installed Shell.
> WARN: == SC firmware version check PASSED with warning

This looks good.

PR #2539, opened 2019-12-06
Removing ERT configuration code which is not necessary anymore

haeseunglee:
Since we don't have to send a exec buffer to configure the ERT, corresponding code is removed from the HAL API testcases.


stsoe:
ok to test

PR #2541, opened 2019-12-06
support export buffer fd for soft kernel

mamin506:
In before, this is a patch for 2019.1. The VCU team had used this for a long time. Add this to 2019.2 pu1.


maxzhen:
I'm merging this one since it's needed for PU1.

PR #2544, opened 2019-12-06
Windows xbmgmt flash scan

AShivangi:
Sample outputs:

```
C:\Users\sagarw\XRT>xbmgmt.exe --override flash --scan
Card [0]:
        Card type:              u250
        Flash type:             SPI
                xilinx_u250_xdma_201830_2,[ts=0x5d027a10],[SC=3.1]
                None

C:\Users\sagarw\XRT>xbmgmt.exe --override flash --scan -v
Card [0]:
        Card type:              u250
        Flash type:             SPI
                xilinx_u250_xdma_201830_2,[ts=0x5d027a10],[SC=3.1]
                None
        Card name                       AU250A64G
        Card S/N:                       21320493800R
        Config mode:            7
        Fan presence:           A
        Max power level:                225W
        MAC address0:           00:0A:35:05:FE:60
        MAC address1:           00:0A:35:05:FE:61
        MAC address2:           FF:FF:FF:FF:FF:FF
        MAC address3:           FF:FF:FF:FF:FF:FF
```

### DIFF
--- a/src/runtime_src/core/pcie/windows/device_windows.cpp
+++ b/src/runtime_src/core/pcie/windows/device_windows.cpp
@@ -97,14 +97,6 @@ rom(const device_type* device, qr_type qr, const std::type_info&, boost::any& va
   case qr_type::QR_ROM_FPGA_NAME:
     value = std::string(reinterpret_cast<const char*>(hdr.FPGAPartName));
     return;
-  }
-
-  if (device->get_user_handle())
-    throw std::runtime_error("device_windows::rom() unexpected qr("
-                             + std::to_string(qr)
-                             + ") for userpf");
-
-  switch (qr) {
   case qr_type::QR_ROM_UUID:
     value = std::string(reinterpret_cast<const char*>(hdr.uuid),16);
     return;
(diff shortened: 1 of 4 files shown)


maxzhen:
Do you want to change "ts" to "ID" to keep it in sync w/ output on Linux?

src/runtime_src/core/pcie/windows/device_windows.cpp
```diff
@@ -97,14 +97,6 @@ rom(const device_type* device, qr_type qr, const std::type_info&, boost::any& va
   case qr_type::QR_ROM_FPGA_NAME:
     value = std::string(reinterpret_cast<const char*>(hdr.FPGAPartName));
     return;
-  }
```
stsoe: Remove this check for device type will not work.  On the user side  there is only limited ROM info, namely the first 4 query values.  Alas, if you ask for QR_ROM_UUID and EPOCH on the user side, then an exception should be thrown.

AShivangi: Oh. I see.. I was getting an error with this code on mgmt side. I'll look more closely

AShivangi:
Max, I'll do that. Thanks!

PR #2575, opened 2019-12-11
versal ospi support

larry9523:
This PR added Versal OSPI support.
1. Add ospi_versal sub driver in xclmgmt
2. Add zocl_ospi_versal sub driver in zocl
3. The two sub drivers communicate through a BRAM to move Firmware image from host memory to device memory
4. xbmgmt flash is enhanced to flash the FW (PDI) to versal flash device
5. device side ospi driver will talk to user space daemon through sysfs node to flash the FW (PDI) through a user space daemon will will call flashcp to write to the flash.

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbmgmt/xospiversal.cpp
+++ b/src/runtime_src/core/pcie/tools/xbmgmt/xospiversal.cpp
@@ -42,7 +42,7 @@ int XOSPIVER_Flasher::xclUpgradeFirmware(std::istream& binStream)
     total_size = binStream.tellg();
     binStream.seekg(0, binStream.beg);
 
-    std::cout << "INFO: ***BOOT.BIN has " << total_size << " bytes" << std::endl;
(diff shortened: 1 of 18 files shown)


src/runtime_src/core/pcie/tools/xbmgmt/xospiversal.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+int XOSPIVER_Flasher::xclUpgradeFirmware(std::istream& binStream)
+{
+    int total_size = 0;
+
+    binStream.seekg(0, binStream.end);
+    total_size = binStream.tellg();
+    binStream.seekg(0, binStream.beg);
+
+    std::cout << "INFO: ***BOOT.BIN has " << total_size << " bytes" << std::endl;
```
mamin506: Should this be "PDI has *** bytes"?

larry9523: fixed

PR #2583, opened 2019-12-12
Set kernel clock freqs to minimum before partial reconfiguration

chienwei-lan:
http://jira.xilinx.com/browse/CR-980859

Before the HW solution is implemented, the dynamically scalable clock can be set to a very low frequency prior to download bitstream to prevent excessive current draw.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
@@ -1829,15 +1829,6 @@ static int icap_download_bitstream(struct icap *icap, const struct axlf *axlf)
 	icap_freeze_axi_gate(icap);
 
 	err = icap_download_hw(icap, axlf);
-	/*
-	 * Perform frequency scaling since PR download can silenty overwrite
-	 * MMCM settings in static region changing the clock frequencies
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
```diff
@@ -2237,6 +2228,14 @@ static int icap_verify_signature(struct icap *icap,
 	return ret;
 }
 
+static int set_min_freqs(struct icap *icap)
+{
+	unsigned short freqs[ICAP_MAX_NUM_CLOCKS] = {0};
+
+	freqs[KERNEL_CLK] = frequency_table[0].ocl;
```
maxzhen: how about clocks?

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
```diff
@@ -2237,6 +2228,14 @@ static int icap_verify_signature(struct icap *icap,
...
+
+	return set_freqs(icap, freqs, ICAP_MAX_NUM_CLOCKS);
+}
```
maxzhen: need a space line after this.

PR #2587, opened 2019-12-13
Add xbtest from P4 

stsoe:
Preview of xbtest in XRT.

Build as XRT component.
Deliver in separate installable package.


stsoe:
Hey Sonal.  Can you please review the changes to cpack.cmake.  I was surprised to see that all our components (aws, azure, container) include XRT's pre- and post-inst.   It was very odd when I installed xbtest to see it remove and reinstall the XRT kernel drivers.  Same would have been (current is) true for the other existing components.  I changed this, but I was unsure about the original intent.

PR #2596, opened 2019-12-15
support flashing through driver

maxzhen:
This PR contains changes to support shell flashing through driver, instead of mapping xclmgmt's user BAR into xbmgmt to flash.

The PR replicated functionality found in xspi.cpp in flash sub-device driver. However, the implementation is a complete rewrite and:
1. Removed unnecessary interactions with device during flashing.
2. Enabled large page (32K/64K) erasing support besides 4K page used in today's code.
3. Added flash content read back support.

Due to the above enhancement/cleanup, flashing via driver is 2X faster.

The flash sub-device driver presents a standard char device to user land by supporting open/close/read/write/lseek callbacks. Users with root privilege can directly read/write from/to the flash memory on card by talking to the dev node found under /dev/xfpga/flash.xxx without going through xclmgmt or using xbmgmt. Note that when writing to the flash memory, driver will need to break down the IO into smaller erasable pages and even do read-modify-write at smallest page level (4K), if the IO is not page aligned or does not cover the entire page (4K/32K/64K). So, it'll be very inefficient if you write big chunk of data to the flash in small non-flash-page-aligned IOs. It'll work, but slow.

Limitations in this PR that needs addressing in follow-up PRs:
1. Only QSPI is supported in the driver. The QSPI_PS (for U25/30) and OSPI (for Versal) are not supported for now.
2. The existing QSPI code path in xbmgmt is still in use by default even after this change. User needs to set FLASH_VIA_DRIVER environment variable to switch to driver flashing code path explicitly. We'll completely remove userland flashing code path and make permanent switch when this change is tested more widely on all platforms applicable.

The QSPI HW interface is, again, proven to be very difficult to work with and took me quite some time to clean up QSPI code and move it down to driver. But, it worth the time for:
1. now we have an easier to maintain implementation in long term
2. 2X faster flash operation
3. can read back what's flashed on the device (a long desired feature)
4. one big step toward getting rid of supporting mmap user bar from xclmgmt, so it’ll be less frowned upon when we upstream the driver source code

What's more, it also enables flash memory access by xclmgmt as well. On today's platforms we supported, there should still be ~50MB space on flash even after a normal shell is flashed. Potentially, we can save DTB/CMC/ERT/SC images to the flash right after xbmgmt flashes the MCS from the same .dsabin file for xclmgmt to load them during attach. Thus, we'll always have a self-contained board with matching shell and firmware without relying on some data saved on hard drive. The shell package will only be needed when someone wants to update the content on the flash, not when user just want to run applications. More work/thoughts are needed in regarding to this.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
@@ -1,9 +1,10 @@
 /*
- * A GEM style device manager for PCIe based OpenCL accelerators.
+ * Platform driver for flash controllers on Xilinx's Alveo cards.
  *
- * Copyright (C) 2018 Xilinx, Inc. All rights reserved.
+ * Copyright (C) 2019 Xilinx, Inc. All rights reserved.
  *
  * Authors: Lizhi Hou <lizhih@xilinx.com>
+ *          Max Zhen <maxz@xilinx.com>
  *
  * This software is licensed under the terms of the GNU General Public
  * License version 2, as published by the Free Software Foundation, and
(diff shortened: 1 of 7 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
```diff
@@ -18,13 +19,1109 @@
...
+		ret = flash_rx(flash, NULL, len);
+	}
+
+	/* Always need to reset slave select register after each transaction */
+	flash_set_slave(flash, SLAVE_NONE);
+
+	return 0;
+}
+
+static size_t flash_get_fifo_depth(struct xocl_flash *flash)
```
cellery: The FIFO depth should be gotten from the DSA. The depth is configured when the DSA is built. Shouldn't be pushing junk into the flash controller to determine the depth.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
```diff
@@ -18,13 +19,1109 @@
...
+		flash->io_buf[0] = op;
+		flash->io_buf[1] = faddr->addr_hi;
+		flash->io_buf[2] = faddr->addr_mid;
+		flash->io_buf[3] = faddr->addr_lo;
+		*header_len = 4;
+	}
+	return ret;
+}
+
+static bool flash_wait_till_ready(struct xocl_flash *flash)
```
cellery: Grammatically the method name is a bit weird, till is not correct word, should be until.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
```diff
@@ -18,13 +19,1109 @@
...
+	if (ret == 0)
+		ret = flash_buf_rdwr(flash, kbuf, off, thislen, true);
+	return ret;
+}
+
+/*
+ * Write to flash memory page by page from user buf.
+ */
+static ssize_t
+flash_write(struct file *file, const char __user *buf, size_t n, loff_t *off)
```
cellery: Method input names are confusing, can you be more descriptive of what off is referring tot? Input n is in units of what?

src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
```diff
@@ -18,13 +19,1109 @@
...
+	if (!flash_wait_till_ready(flash))
+		ret = -EINVAL;
+
+	while (ret == 0 && cnt < n) {
+		loff_t thisoff = *off + cnt;
+		const char *thisbuf = buf + cnt;
+		size_t thislen = n - cnt;
+
+		/* Try write full page. */
+		ret = flash_page_wr(flash, thisbuf, page, thisoff, &thislen);
```
cellery: Why is there a read,modify,write sequence for flash here? Images stored in flash have to be aligned to 4KB boundaries so you should always be able to do a 4KB erase and not need to store contents of what was in flash at that offset.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
maxzhen: This is how xilinx's upstream spi driver finds out the fifo depth (see xilinx_spi_find_buffer_size() in https://elixir.bootlin.com/linux/latest/source/drivers/spi/spi-xilinx.c). So, I think this should be OK?

src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
maxzhen: this is a generic standard char device interface. user may not do page aligned write at all.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
maxzhen: sure. will change.

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
```diff
@@ -1604,6 +1689,193 @@ int XSPI_Flasher::revertToMFG(void)
...
+        switch (type) {
+        case 0:
+            if (currentAddr == UINT_MAX) {
+                std::cout << "MCS missing page starting address" << std::endl;
+                return -EINVAL;
+            } else if (buf.size() == 0) {
+                // we're the 1st data in this page, updating the current addr
+                assert(pageOffset(currentAddr) == 0);
+                currentAddr |= addr;
+            } else if (pageOffset(currentAddr + buf.size()) != addr) {
```
cellery: Will this fail if the MCS contains multiple images? For example the golden image MCS file contains multiple bitstreams and they are not contiguous and will have a gap between the two bitstreams.

cellery: The main point I'm making here is an MCS does not map directly to a single bitstream. The MCS file can contain multiple images at different offsets.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
maxzhen: this is the write callback function for a standard linux char device driver. you write n bytes starting from off offset of a device.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
cellery: That's an extremely ugly way of getting the depth of the buffer but I guess its fine if that's what they do in our spi drivers.

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
maxzhen: it should work, but i have not tested since i don't have a mcs with more than one chunk to test with. but if there is a gap, you'd expect to see a type 4 line, right? type 0 lines should be contiguous, no?

src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
maxzhen: yes, i agree that it'd be ideal if we have a register to read the fifo length instead of doing this. but that register is not available :-(.

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
cellery: Ah I missed your logic checking in the case 4 statement below. Looks like it would work then. You can test this with the golden image MCS files as those have 2 bitstreams inside that aren't contiguous. You'd have to make sure to unlock the flash first so that you can erase the golden image.

PR #2597, opened 2019-12-16
xbmgmt2 flash --scan update

AShivangi:
- display bdf of the card
- display information about the platform installed in the system

```
Card [0000:04:00.0]:
        Card type:              u250
        Flash type:             SPI
                xilinx_u250_xdma_201830_2,[ID=0x5d14fbe6],[SC=UNKNOWN]
                xilinx_u250_xdma_201830_2,[ID=0x5d027a10],[SC=4.2.0]
```

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/flash/firmware_image.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/flash/firmware_image.cpp
@@ -102,21 +102,26 @@ void getVendorBoardFromDSAName(std::string& dsa, std::string& vendor, std::strin
     board = tokens[1];
 }
 
-void parseDSAFilename(std::string filename, uint64_t& vendor, uint64_t& device, uint64_t& subsystem, uint64_t &ts)
+
+
+void parseDSAFilename(const std::string& filename, uint64_t& vendor, uint64_t& device, uint64_t& subsystem, uint64_t &ts)
 {
 	vendor = 0; device = 0; subsystem = 0; ts = 0;
-    typedef boost::tokenizer<boost::char_separator<char>> tokenizer;
+	using tokenizer = boost::tokenizer< boost::char_separator<char> >;
     boost::char_separator<char> sep("-.");
     tokenizer tokens(filename, sep);
+	int radix = 16;
+
+	// check if we have 5 tokens: vendor, device, subsystem, ts, "dsabin"/"xsabin"
 	if (std::distance(tokens.begin(), tokens.end()) == 5) {
 	    tokenizer::iterator tok_iter = tokens.begin();
-		vendor = std::stoull(std::string(*tok_iter), 0, 16);
+		vendor = std::stoull(std::string(*tok_iter), nullptr, radix);
 		tok_iter++;
-		device = std::stoull(std::string(*tok_iter), 0, 16);
+		device = std::stoull(std::string(*tok_iter), nullptr, radix);
 		tok_iter++;
-		subsystem = std::stoull(std::string(*tok_iter), 0, 16);
+		subsystem = std::stoull(std::string(*tok_iter), nullptr, radix);
 		tok_iter++;
-		ts = std::stoull(std::string(*tok_iter), 0, 16);
+		ts = std::stoull(std::string(*tok_iter), nullptr, radix);
 		tok_iter++;
 	} else
 		ts = NULL_TIMESTAMP;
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/tools/xbmgmt2/flash/firmware_image.cpp
```diff
@@ -398,82 +409,58 @@ std::vector<DSAInfo> firmwareImage::installedDSA;
...
+
+    for (std::string t : suffix) {
+
+        std::regex e("^" FORMATTED_FW_DIR "/([^/]+)/([^/]+)/([^/]+)/.+\\." + t);
+        std::cmatch cm;
+
+        for (boost::filesystem::recursive_directory_iterator iter(formatted_fw_dir, 
+            boost::filesystem::symlink_option::recurse), recursive_end; iter != recursive_end;) {
+            std::string name = iter->path().string();
+            std::regex_match(name.c_str(), cm, e);
```
stsoe: is c_str() really necessary?

src/runtime_src/core/tools/xbmgmt2/flash/firmware_image.cpp
```diff
@@ -398,82 +409,58 @@ std::vector<DSAInfo> firmwareImage::installedDSA;
...
+    if (!installedDSA.empty())
+        return installedDSA;
+
+    // for 2RP
+    p = FORMATTED_FW_DIR;
+	if (!boost::filesystem::is_directory(p))
+        return installedDSA;
+
+    boost::filesystem::path formatted_fw_dir(FORMATTED_FW_DIR);
+    std::vector<std::string> suffix = { XSABIN_FILE_SUFFIX, DSABIN_FILE_SUFFIX};
```
stsoe: It is really necessary to construct a vector of strings here?
Would this maybe work?

> for (auto t : { XSABIN_FILE_SUFFIX, DSABIN_FILE_SUFFIX} ) {
> ...
> }

src/runtime_src/core/tools/xbmgmt2/flash/firmware_image.cpp
```diff
@@ -104,22 +104,22 @@ void getVendorBoardFromDSAName(std::string& dsa, std::string& vendor, std::strin
 
 void parseDSAFilename(std::string filename, uint64_t& vendor, uint64_t& device, uint64_t& subsystem, uint64_t &ts)
```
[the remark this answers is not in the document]
AShivangi: I'll add comments explaining the digits. I talked to Stephen about regex and he thinks boost tokenizer is cleaner than regex

src/runtime_src/core/tools/xbmgmt2/flash/firmware_image.cpp
```diff
@@ -398,82 +409,58 @@ std::vector<DSAInfo> firmwareImage::installedDSA;
...
+
+    // for 2RP
+    p = FORMATTED_FW_DIR;
+	if (!boost::filesystem::is_directory(p))
+        return installedDSA;
+
+    boost::filesystem::path formatted_fw_dir(FORMATTED_FW_DIR);
+    std::vector<std::string> suffix = { XSABIN_FILE_SUFFIX, DSABIN_FILE_SUFFIX};
+
+    for (std::string t : suffix) {
```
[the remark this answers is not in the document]
AShivangi: that gives an error so changing it to const std::string& instead

src/runtime_src/core/tools/xbmgmt2/flash/firmware_image.cpp
AShivangi: yes

src/runtime_src/core/tools/xbmgmt2/flash/firmware_image.cpp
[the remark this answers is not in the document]
stsoe: Yes, you are indeed tokenizing the string.  And I take back the comment of 0,16 those are just conversion args to strtoull.

src/runtime_src/core/tools/xbmgmt2/flash/firmware_image.cpp
stsoe: I don't think so.  I think it is std::smatch sm to stay with std::string.  Please grep code base for std::smatch it's used in different places.

AShivangi: Oh ok. I haven't really explored regex so I don't know much about it. I made the change you asked. I'll have to go through the code again when verifying 2RP so I'll take a closer look then :)

PR #2604, opened 2019-12-18
Patch 2

austino1976:
Fix for CR-1050591

### DIFF
--- a/src/runtime_src/ert/CMakeLists.txt
+++ b/src/runtime_src/ert/CMakeLists.txt
@@ -1,4 +1,4 @@
-if ((EXISTS /tools/batonroot/rodin/devkits) AND (${XRT_NATIVE_BUILD} STREQUAL "yes") AND (${CMAKE_HOST_SYSTEM_PROCESSOR} STREQUAL "x86_64"))
+if ((EXISTS $ENV{XILINX_VITIS}/gnu/microblaze/lin) AND (${XRT_NATIVE_BUILD} STREQUAL "yes") AND (${CMAKE_HOST_SYSTEM_PROCESSOR} STREQUAL "x86_64"))
 
 message("-- Rodin devkits found, preparing ERT build")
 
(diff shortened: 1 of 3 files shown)


rradjabi:
Retest this please

aye20:
> Does aws F1 have ENV(XILINX_VITIS) set?
> I mean, when build on aws, do we make sure the ert is loaded into the emu.so?

We look for the XILINX_VITIS if it exists, the libsched_em.so is added, if not libsched_em.so does not get built.

maxzhen:
ok to test

xuhz:
we have to make sure XILINX_VITIS is set on aws -- that is what this fix is doing, right? 
on aws, we can't access /tools/..., then the libsched_em.so doesn't get built, and then the emu doesn't work. 
what this fix is doing is to make sure libsched_em.so is built. so XILINX_VITIS has to be set on aws

src/runtime_src/ert/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-if ((EXISTS /tools/batonroot/rodin/devkits) AND (${XRT_NATIVE_BUILD} STREQUAL "yes") AND (${CMAKE_HOST_SYSTEM_PROCESSOR} STREQUAL "x86_64"))
+if ((EXISTS $ENV{XILINX_VITIS}/gnu/microblaze/lin) AND (${XRT_NATIVE_BUILD} STREQUAL "yes") AND (${CMAKE_HOST_SYSTEM_PROCESSOR} STREQUAL "x86_64"))
```
stsoe: Where is XILINX_VITIS set?

src/runtime_src/ert/ert.mk
```diff
@@ -1,5 +1,4 @@
-RDI_DEVKITS_HOME := /tools/batonroot/rodin/devkits
-MB_HOME := $(RDI_DEVKITS_HOME)/lnx64/cross/RHEL6/microblaze/lin
+MB_HOME := ${XILINX_VITIS}/gnu/microblaze/lin
```
stsoe: Are we sure this is the same compiler?  This is a pretty scary change unless we are certain the compiler to use.

stsoe: Even if I set XILINX_VITIS and use the MB_HOME, the sched.bin file are not identical to before.

While I agree with the intent of this PR and that using XILINX_VITIS is better that using the internal mb compiler,  making this change in 2019.2 branch is not possible.      

Please redo this PR and make it specific to what you are after, such that it does not affect building of the microblaze firmware used in hardware.

austino1976:
Superseded by PR#2617

PR #2607, opened 2019-12-18
support for multiple devices in xbmgmt2

AShivangi:
- flash --scan should scan all the devices available
- --card option should work for all subcommands

### DIFF
--- a/src/runtime_src/core/common/system.cpp
+++ b/src/runtime_src/core/common/system.cpp
@@ -69,10 +69,10 @@ get_total_devices()
   return instance().get_total_devices();
 }
 
-void
-bdf2index(uint16_t& index, const std::string& bdfStr)
+uint16_t
+bdf2index(const std::string& bdfStr)
 {
-  instance().bdf2index(index, bdfStr);
+  return instance().bdf2index(bdfStr);
 }
 
 } // xrt_core
--- a/src/runtime_src/core/common/system.cpp
+++ b/src/runtime_src/core/common/system.cpp
@@ -72,7 +72,7 @@ get_total_devices()
 void
 bdf2index(uint16_t& index, const std::string& bdfStr)
 {
-  return instance().bdf2index(index, bdfStr);
+  instance().bdf2index(index, bdfStr);
 }
 
 } // xrt_core
--- a/src/runtime_src/core/common/system.cpp
+++ b/src/runtime_src/core/common/system.cpp
@@ -63,4 +63,16 @@ get_mgmtpf_device(device::id_type id)
   return instance().get_mgmtpf_device(id);
 }
 
+std::pair<uint64_t, uint64_t>
+get_total_devices()
+{
+  return instance().get_total_devices();
+}
+
+void
+bdf2index(uint16_t& index, const std::string& bdfStr)
+{
+  return instance().bdf2index(index, bdfStr);
+}
+
 } // xrt_core
--- a/src/runtime_src/core/pcie/common/system_pcie.cpp
+++ b/src/runtime_src/core/pcie/common/system_pcie.cpp
@@ -45,4 +45,40 @@ get_devices(boost::property_tree::ptree& pt) const
   pt.add_child("devices", pt_devices);
 }
 
+void
+system_pcie::
+bdf2index(uint16_t& index, const std::string& bdfStr) const
+{
+  // Extract bdf from bdfStr.
+  int dom = 0, b= 0, d = 0, f = 0;
+  index = std::numeric_limits<uint16_t>::max();
+  char dummy;
+  std::stringstream s(bdfStr);
+  size_t n = std::count(bdfStr.begin(), bdfStr.end(), ':');
+  
+  if (n == 1)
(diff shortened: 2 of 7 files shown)


src/runtime_src/core/common/system.cpp
```diff
@@ -63,4 +63,16 @@ get_mgmtpf_device(device::id_type id)
   return instance().get_mgmtpf_device(id);
 }
 
+std::pair<uint64_t, uint64_t>
+get_total_devices()
+{
+  return instance().get_total_devices();
+}
+
+void
```
stsoe: Can you return the index?

src/runtime_src/core/pcie/common/system_pcie.cpp
```diff
@@ -45,4 +45,40 @@ get_devices(boost::property_tree::ptree& pt) const
...
+    s >> std::hex >> dom >> dummy >> b >> dummy >> d >> dummy >> f;
+  if ((n != 1 && n != 2) || s.fail()) {
+    std::string errMsg = boost::str( boost::format("Can't extract BDF from '%s'") % bdfStr);
+    throw error(errMsg);
+	}
+
+  for (uint16_t i = 0; i < get_total_devices().first; i++) {
+    auto device = get_mgmtpf_device(i);
+    boost::any bus, dev, func;
+    device->query(xrt_core::device::QR_PCIE_BDF_BUS, typeid(b), bus);
```
stsoe: I suggest you short circuit if 'bus' doesn't match 'b', ditto for 'dev' and 'd', and 'func' and 'f'.

src/runtime_src/core/pcie/common/system_pcie.cpp
```diff
@@ -45,4 +45,40 @@ get_devices(boost::property_tree::ptree& pt) const
...
+    device->query(xrt_core::device::QR_PCIE_BDF_DEVICE, typeid(d), dev);
+    device->query(xrt_core::device::QR_PCIE_BDF_FUNCTION, typeid(f), func);
+
+    if (b == boost::any_cast<uint16_t>(bus) && d == boost::any_cast<uint16_t>(dev) && f == boost::any_cast<uint16_t>(func)) {
+      index = i;
+    } else {
+      std::string errMsg = boost::str( boost::format("No mgmt PF found for '%s'") % bdfStr);
+      throw error(errMsg);
```
stsoe: You are throwing out of the for loop for the after just checking first device that doesn't match. Or in other words, you loop handles only one device

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
```diff
@@ -44,52 +45,55 @@ static const unsigned int registerResult =
...
-	std::cout << "\tMAC address0:\t\t" << info.mMacAddr0 << "\n";
-	std::cout << "\tMAC address1:\t\t" << info.mMacAddr1 << "\n";
-	std::cout << "\tMAC address2:\t\t" << info.mMacAddr2 << "\n";
-	std::cout << "\tMAC address3:\t\t" << info.mMacAddr3 << "\n";
+  if (total == 0) {
+    std::string errMsg = "No card found!";
+    throw xrt_core::error(errMsg);
   }
+  
+  for(unsigned i = 0; i < total; i++) {
```
stsoe: unsigned what?

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
```diff
@@ -44,52 +45,55 @@ static const unsigned int registerResult =
...
+	  std::cout << "\tMAC address2:\t\t" << info.mMacAddr2 << "\n";
+	  std::cout << "\tMAC address3:\t\t" << info.mMacAddr3 << "\n";
+    }
+  }
 }
 
 // Update shell on the board.
-static void update_shell(unsigned index, std::string flashType,
+static void 
+update_shell(unsigned index, std::string flashType,
```
stsoe: const std::string& flashType ???

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
```diff
@@ -122,7 +126,8 @@ static void update_shell(unsigned index, std::string flashType,
     std::cout << "Cold reboot machine to load new shell on card" << std::endl;
 }
 
-static void update_SC(unsigned index, const std::string& file)
+static void 
+update_SC(unsigned index, const std::string& file)
```
stsoe: unsigned what?
repeated below.

src/runtime_src/core/common/system.cpp
AShivangi: I thought we wanted all void functions

stsoe: haha :-)  Sorry for the confusion.  No, what we don;'t want is any functions to return an error code, but they can surely return a result and throw on error.

AShivangi: Oh. Ok that makes sense!

PR #2609, opened 2019-12-18
Add infrastructure for build V2.0 of ERT

stsoe:
Enable build by uncommenting v20/sched.bin in
% cat schediler/CMakeLists.txt
...
add_custom_target(scheduler
 DEPENDS legacy/sched.bin u50/sched.bin #v20/sched.bin
)
...


stsoe:
Min.  FYI

PR #2625, opened 2019-12-24
Support updating SC image via driver

maxzhen:
This pull request is for supporting updating SC firmware image via xmc subdevice driver. The new code path can be enabled by setting FLASH_VIA_DRIVER environment variable.


maxzhen:
retest this please

PR #2630, opened 2020-01-03
IOCTL QueryRequest for sensor info

AShivangi:
- updated ioctl calls for xcl_sensor struct
- printing temp info in xbutil2 query
- works on windows and linux
- merge https://github.com/Xilinx/XRT/pull/2629 and update drivers before testing this PR
Partial output for xbutil2 query:
```
...
Accelerator
   Accelerator metadata (e.g., xclbin) unavailable.

Temperature
  PCB top front    : 37 C
  PCB top rear     : 31 C
  PCB bottom front : 37 C
  FPGA             : 37 C
  FAN trig crit    : 36 C
----------------------------------------------------------------
```


src/runtime_src/core/pcie/windows/device_windows.cpp
```diff
@@ -198,6 +198,152 @@ info_mgmt(const device_type* device, qr_type qr, const std::type_info&, boost::a
   }
 }
 
+static void
+sensor_info(const device_type* device, qr_type qr, const std::type_info&, boost::any& value)
```
rozumx: First let me say, that I understand the changes you are making and how you are working on "fitting" them into the code flow.  That said, I do have an issue with the "large" switch statement.  It would be great to place all of the metadata associated with gathering this data into a "collection" and use that collection to obtain the query of choice. 

Does this work need to be done for this PR, I don't think so, but it _**should be done in the next two weeks**_.  That said, could you add a ticket to our scrum board and schedule it to be done in our next sprint.

src/runtime_src/core/pcie/windows/shim.cpp
```diff
@@ -1003,6 +1003,23 @@ struct shim
       throw std::runtime_error("DeviceIoControl IOCTL_XOCL_STAT (get_ip_layout) failed");
   }
 
+  void
+  get_sensor_info(xcl_sensor* value)
+  {
+    DWORD bytes = 0;
+    auto status = DeviceIoControl(m_dev,
```
rozumx: No need to define an auto here, please use the actual return type.

src/runtime_src/core/pcie/windows/shim.cpp
```diff
@@ -1086,6 +1103,15 @@ get_bdf_info(xclDeviceHandle hdl, uint16_t bdf[3])
   shim->get_bdf_info(bdf);
 }
 
+void
+get_sensor_info(xclDeviceHandle hdl, xcl_sensor* value)
+{
+  xrt_core::message::
+    send(xrt_core::message::severity_level::XRT_DEBUG, "XRT", "sensor_info()");
+  auto shim = get_shim_object(hdl);
```
rozumx: Again, no need to define an auto here.

src/runtime_src/core/pcie/windows/device_windows.cpp
AShivangi: Ok. Sounds good. I'll add a ticket for this

src/runtime_src/core/pcie/windows/device_windows.cpp
```diff
@@ -198,6 +198,152 @@ info_mgmt(const device_type* device, qr_type qr, const std::type_info&, boost::a
...
+{
+  auto init_sensor_info = [](const device_type* dev) {
```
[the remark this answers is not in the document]
AShivangi: This can be changed during cleanup. For now, I followed the existing code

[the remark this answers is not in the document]
AShivangi: The return type of lamda function has to be auto, otherwise, it gives a compilation error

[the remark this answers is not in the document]
stsoe: lambda makes sense, no need to pollute with free function.

src/runtime_src/core/pcie/windows/device_windows.cpp
```diff
@@ -198,6 +198,152 @@ info_mgmt(const device_type* device, qr_type qr, const std::type_info&, boost::a
...
+    userpf::get_sensor_info(dev->get_user_handle(), &info);
+    return info;
+  };
+
+  static std::map<const device_type*, xcl_sensor> info_map;
+  static std::mutex mutex;
+  std::lock_guard<std::mutex> lk(mutex);
+  auto it = info_map.find(device);
+  if (it == info_map.end()) {
+    auto ret = info_map.emplace(device,init_sensor_info(device));
```
stsoe: You are caching the sensor info, when does it expire?  Isn't sensor info dynamic data?  It's fine to cache, but maybe expire every second or so?

stsoe: Defer till we actually need sensor  information in a long running process like xbutil top.

PR #2631, opened 2020-01-03
fix softkernel validate

larry9523:
This PR fix the following issue

1) skip cu validate for soft kernel
2) add a new ERT command type ERT_SCU so that ERT_START_SCU can be scheduled deprately.
3) remove two more XOCL_DSA_IS_VERSAL.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018-2019 Xilinx, Inc. All rights reserved.
+ * Copyright (C) 2018-2020 Xilinx, Inc. All rights reserved.
  *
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
```diff
@@ -4091,13 +4123,14 @@ static int convert_execbuf(struct xocl_dev *xdev, struct drm_file *filp,
...
 	case ERT_EXEC_WRITE:
-	case ERT_SK_START:
 		scmd->type = ERT_CU;
 		break;
 	case ERT_SK_CONFIG:
 	case ERT_SK_UNCONFIG:
 		scmd->type = ERT_CTRL;
 		break;
+	case ERT_SK_START:
+		scmd->type = ERT_SCU;
```
larry9523: Just realized that a 'break' is missing. I will add it.

PR #2638, opened 2020-01-06
[Edeg] data-driven interrupt handling(PR-15981)

mamin506:
The legacy interrupt xclbin support would be remove in the future when system linker doesn't support legacy platforms.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/sched_exec.c
+++ b/src/runtime_src/core/edge/drm/zocl/sched_exec.c
@@ -516,7 +516,7 @@ static irqreturn_t sched_exec_isr(int irq, void *arg)
 
 	SCHED_DEBUG("-> %s irq %d", __func__, irq);
 	for (cu_idx = 0; cu_idx < zdev->cu_num; cu_idx++) {
-		if (zdev->irq[cu_idx] == irq)
+		if (zdev->exec->zcu[cu_idx].irq == irq)
 			break;
 	}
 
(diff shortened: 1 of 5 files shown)


src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -815,10 +820,19 @@ configure(struct sched_cmd *cmd)
...
+			if (apt_idx < 0) {
+				DRM_ERROR("CU address %x is not found in XCLBIN\n",
+					  cfg->data[i]);
+				return 1;
+			}
+			if (is_legacy_intr)
+				irq_id = i;
+			else
+				irq_id = zocl_xclbin_intr_id(zdev, apt_idx);
+			exec->zcu[i].irq = zdev->irq[irq_id];
```
larry9523: Initially I was concerned that irq_id might be out of the zdev->irq[MAX_CU_NUM]. But look deep into the code, since IP_INTERRUPT_ID_MASK is 0xFE and IP_INTERRUPT_ID_SHIFT is 1, irq_id will only be between [0, 127] and our MAX_CU_NUM is 128. So we are good :)
Though, I would suggest to add an extra check here to fail gracefully if we get a wrong irq_id in case we bump up MAX CU NUM or something. But given the reason above, I am OK without that check.

xdavidz:
@mamin506 sorry, did not get enough time to go through it, will get it done by today.

src/runtime_src/core/edge/drm/zocl/sched_exec.c
mamin506: Let's refine that in the new version of kds.

PR #2639, opened 2020-01-06
Update XoclUser_INTF.h

igorshr:
Add new IOCTLs for getting SENSOR and ICAP information


igorshr:
I have already cancelled the request. Arpit sent this changes before.

PR #2640, opened 2020-01-07
after PCI reset, mb_scheduler should have clean command buffer

xdavidz:
This is the idea to fix the issue in PU1 compare to 2019.1.

When firewall tripped, the health monitor thread will start issuing PCI reset. After the tripped is resolved for some reason, in 2019.1, we can restart. But in PU1, we cannot restart because of the xdev->outstanding_execs is not zero. Since we have done a hot reset, we should just reclaim all buffers and reset the outstanding_execs back to zero.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
@@ -4495,6 +4495,30 @@ int cu_map_addr(struct platform_device *pdev, u32 cu_idx, void *drm_filp,
 	return 0;
 }
 
+static int
+hot_reset(struct platform_device *pdev)
+{
+	struct exec_core *exec = platform_get_drvdata(pdev);
+	struct xocl_dev	*xdev = xocl_get_xdev(pdev);
+
+	/*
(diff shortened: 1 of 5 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
```diff
@@ -4495,6 +4495,30 @@ int cu_map_addr(struct platform_device *pdev, u32 cu_idx, void *drm_filp,
...
+{
+	struct exec_core *exec = platform_get_drvdata(pdev);
+	struct xocl_dev	*xdev = xocl_get_xdev(pdev);
+
+	/*
+	 * reclaim outstanding cmds, exec_rest_cmds is responsible for
+	 * recycling all cmds without leak.
+	 */
+	exec_reset_cmds(exec);
+	atomic_set(&xdev->outstanding_execs, 0);
```
stsoe: Silently setting outstanding_execs to 0 is maybe not a good idea?    exec_reset_cmds should have decremented this number,   If the count is off (not zero) then there is some other bug,.  Could we at least log a message if this ever happens?

PR #2642, opened 2020-01-07
adding gtest support for xclbinutil for windows

rbramand-xilinx:
Hi,

This PR contains following changes
>updated xrtdeps-win.py script which downloads, builds and links gtest libraries to make xclbin unit tests work for windows
>Changed path of GTest_Root so that find_package gets the path for .lib files

Thanks


chvamshi-xilinx:
ok to test

PR #2643, opened 2020-01-08
CR-1051238 after PCI reset, mb_scheduler should have clean command buffer

xdavidz:
Summary of RCA of xclbin download issue and suggested solutions:

Root Cause Analysis:

cmds running on CU is marked as running_cmds in XRT scheduler, when firewall trip (causes xrt reset), XRT will set those running cmds as ABORTED. The PU1's scheduler will free any commands (status >= COMPLETE, including ABORTED), but forget to update the global value outstanding_execs. Thus, xclbin cannot be swapped due to this outstanding_exec is not ZERO.

Solutions:

1) mini tweak: enforce outstanding_exec to zero after reset. Pro: simple; Con: please see No.2;

2) more sophisticated solution:

     2.1) running cmds on CU cannot be freed unless we can reset the hardware (CU), but we don't reset CU today, so we should not free aborted running cmd before any hardware reset.

     2.2) after reset whole PCI devices, hardware are considered reset, we should clean up any pending cmds now. 

     2.3) after reset, we reclaim aborted cmds, reinitialize pci interrupts, etc.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
@@ -2158,8 +2158,31 @@ exec_isr(int irq, void *arg)
 	return IRQ_HANDLED;
 }
 
-/*
- */
+static void scheduler_intr_register(struct exec_core *exec)
+{
+	struct xocl_dev *xdev = exec_get_xdev(exec);
+	int i;
+
+	for (i = 0; i < exec->intr_num; i++) {
+		userpf_info(xdev, "intr # %d", i);
+		xocl_user_interrupt_reg(xdev, i+exec->intr_base, exec_isr, exec);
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
```diff
@@ -3114,7 +3132,8 @@ exec_running_to_complete(struct exec_core *exec)
 		if (xcmd->state == ERT_CMD_STATE_RUNNING)
 			exec_query_cmd(exec, xcmd);
 
-		if (xcmd->state >= ERT_CMD_STATE_COMPLETED) {
```
stsoe: I am not comfortable with this change.

xdavidz: I would have more conservative fix for this. Close this PR for now.

PR #2661, opened 2020-01-10
Interrupt support from Device to XRT for HW Emulation

ukyatham-xilinx:
Added new socket communication between SIM_QDMA to XRT for passing Device interrupt to XRT.


chvamshi-xilinx:
ok to test

PR #2664, opened 2020-01-10
xbmgmt2 flash --scan 

AShivangi:
- xbmgmt2 flash --scan now works on linux and windows
- shows the correct output when the card is in the golden state
- added "isUser" field in linux query to determine userpf/mgmtpf

Driver work needed:
**"is_mfg"** and **"board_name"** ioctl calls are needed. Currently, these values are hardcoded


src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
```diff
@@ -49,7 +49,7 @@ static void
 scan_devices(bool verbose, bool json)
 {
   json = json;
-  auto total = xrt_core::get_total_devices().first;
+  auto total = xrt_core::get_total_devices(false).first;
```
rozumx: Hmm....
I looks like get_total_devices() returns different information if the user driver is being examined as opposed to the mgmt driver (hence the changes).  By chance do you know why for I would expect the same results.

src/runtime_src/core/tools/xbmgmt2/flash/flasher.cpp
```diff
@@ -318,38 +318,25 @@ std::vector<DSAInfo> Flasher::getInstalledDSA()
...
-    {
+    } else {
         vbnv = xrt_core::query_device<std::string>(m_device, xrt_core::device::QR_ROM_VBNV);
         ts = xrt_core::query_device<uint64_t>(m_device, xrt_core::device::QR_ROM_TIME_SINCE_EPOCH);
-    }
-    else
-    {
-        std::cout << "ERROR: Platform name not found" << std::endl;
+        uuid = xrt_core::query_device<std::string>(m_device, xrt_core::device::QR_ROM_UUID);
+        if (vbnv.empty() || ts == std::numeric_limits<uint64_t>::max())
```
rozumx: 1) Could you break this error condition into two separate errors (one for invalidate timestamp and the other missing vbnv value.

2) Since an error is generated, an exception should be thrown and caught by the calling method.

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
AShivangi: When the device is ready to use, you'll have both xocl and xclmgmt devices. I had to add this so that xbmgmt flash --scan displays the correct information when the golden image is loaded. In that case, the user driver is not loaded but the mgmt driver is.

src/runtime_src/core/tools/xbmgmt2/flash/flasher.cpp
AShivangi: Yes. Will do that

src/runtime_src/core/tools/xbmgmt2/flash/flasher.cpp
```diff
@@ -229,15 +231,15 @@ Flasher::Flasher(unsigned int index) : mFRHeader{}
...
+    {
+       dev->read(MFG_REV_OFFSET, &mGoldenVer, sizeof(mGoldenVer));
+    }
+    // else
+    // {
     //    std::cout << "ERROR: card not supported." << std::endl;
-    //}
-    m_device = dev; // Successfully initialized
+    // }
+    m_device = std::dynamic_pointer_cast<xrt_core::device>(dev); // Successfully initialized
```
stsoe: dynamic_cast.  Not sure I follow what's going on here?

AShivangi: dev is of type pci_device whereas m_device is of type device on linux. Hance the dynamic cast

stsoe: Casting derived to base does not require dynamic_cast.  is it not directly assignable?

PR #2667, opened 2020-01-13
Fix some syntax warnings in xrt_bindings.py for Python>=3.8

giunatale:
This PR fixes some syntax warnings for using 'is' instead of '==' for string and numerical literals. From Python 3.8, doing so will produce a SyntaxWarning, and in general, it is incorrect (see [here](https://bugs.python.org/issue34850)).

The PR is done against master, but the syntax issue should be really fixed across all branches.


AShivangi:
Has this been tested with python 2.7?

giunatale:
Yes, tested on 2.7, and I can confirm it is working.

If there's anything else that need to be done from my end, let me know!

PR #2671, opened 2020-01-14
EoU XRT: Re-architected xbutil2 & xbmgmt2 help subsystems.

rozumx:
Re-architected xbutil2 & xbmgmt2 help subsystem
Work Done
-------------
1. All sub-commands are now contained in their own class.
2. Sub-commands now can be designated as hidden, deprecated, preliminary, etc.
3. New pretty print help created for both sub-command and top level parser.
4. Remove "self registration" of sub-command and replaced it with a more data-driven approached.
5. Cleaned-up CMakelist.txt build scripts
6. Added new xbutil 'examine' sub-command

### DIFF
--- a/src/runtime_src/core/tools/common/SubCmd.h
+++ b/src/runtime_src/core/tools/common/SubCmd.h
@@ -45,7 +45,7 @@ class SubCmd {
 
  // Child class Helper methods
  protected:
-  SubCmd(const std::string _name, const std::string _shortDescription);
+  SubCmd(const std::string & _name, const std::string & _shortDescription);
   void setIsHidden(bool _isHidden) { m_isHidden = _isHidden; };
   void setIsDeprecated(bool _isDeprecated) { m_isDeprecated = _isDeprecated; };
   void setIsPreliminary(bool _isPreliminary) { m_isPreliminary = _isPreliminary; };
--- a/src/runtime_src/core/tools/common/SubCmd.h
+++ b/src/runtime_src/core/tools/common/SubCmd.h
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2019 Xilinx, Inc
+ * Copyright (C) 2019-2020 Xilinx, Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License"). You may
  * not use this file except in compliance with the License. A copy of the
@@ -20,22 +20,56 @@
 // Please keep eternal include file dependencies to a minimum
 #include <vector>
 #include <string>
-#include <map>
+#include <boost/program_options.hpp>
   
-typedef int (*t_subcommand)(const std::vector<std::string> &);
+class SubCmd {
+ public:
+   typedef std::vector<std::string> SubCmdOptions;
+
+ public:
+   virtual void execute(const SubCmdOptions &_options) const = 0;
+
+ public:
+   const std::string &getName() const { return m_subCmdName; };
+   const std::string &getShortDescription() const { return m_shortDescription; };
+   bool isHidden() const { return m_isHidden; };
+   bool isDeprecated() const { return m_isDeprecated; };
+   bool isPreliminary() const { return m_isPreliminary; };
+
+ public:
+   void setExecutableName(const std::string & _name) { m_executableName = _name; };
+   const std::string & getExectuableName() const {return m_executableName; };
+
+ public:
+   virtual ~SubCmd() {};
+
+ // Child class Helper methods
+ protected:
+  SubCmd(const std::string _name, const std::string _shortDescription);
+  void setIsHidden(bool _isHidden) { m_isHidden = _isHidden; };
+  void setIsDeprecated(bool _isDeprecated) { m_isDeprecated = _isDeprecated; };
+  void setIsPreliminary(bool _isPreliminary) { m_isPreliminary = _isPreliminary; };
+  void setLongDescription(const std::string &_longDescription) {m_longDescription = _longDescription; };
+  void setExampleSyntax(const std::string &_exampleSyntax) {m_exampleSyntax = _exampleSyntax; };
+  void printHelp(const boost::program_options::options_description & _optionDescription) const;
+
+ // Methods not supported
+ private:
+  SubCmd() = delete;
+  SubCmd(const SubCmd& obj) = delete;
+  SubCmd& operator=(const SubCmd& obj) = delete;
+
+ // Variables
+ private:
+  std::string m_executableName;
+  std::string m_subCmdName;
+  std::string m_shortDescription;
+  std::string m_longDescription;
+  std::string m_exampleSyntax;
+
+  bool m_isHidden;
+  bool m_isDeprecated;
+  bool m_isPreliminary;
+};
   
-typedef struct {
-  std::string sSubCmd;
-  std::string sDescription;
-  t_subcommand callBackFunction;
-  bool isHidden;
-} SubCmdEntry;
-
-typedef std::map<const std::string, SubCmdEntry> SubCmdTable;
-
-unsigned int register_subcommand(const std::string &_subCommandName, const std::string &_description, t_subcommand _pSubCommand, bool _isHidden = false);
-const SubCmdEntry *getSubCmdEntry(const std::string &_sSubCmdName);
-const SubCmdTable &getSubCmdsTable();
-
-
 #endif
--- a/src/runtime_src/core/tools/common/SubCmd.h
+++ b/src/runtime_src/core/tools/common/SubCmd.h
@@ -53,11 +53,8 @@ class SubCmd {
   void setExampleSyntax(const std::string &_exampleSyntax) {m_exampleSyntax = _exampleSyntax; };
   void printHelp(const boost::program_options::options_description & _optionDescription) const;
 
- // Methods not supported
  private:
   SubCmd() = delete;
-  SubCmd(const SubCmd& obj) = delete;
-  SubCmd& operator=(const SubCmd& obj) = delete;
 
  // Variables
  private:
--- a/src/runtime_src/core/tools/common/XBMain.h
+++ b/src/runtime_src/core/tools/common/XBMain.h
@@ -24,21 +24,12 @@
 // ------------ F O R W A R D - D E C L A R A T I O N S ----------------------
 // Forward declarations - use these instead whenever possible...
 class SubCmd;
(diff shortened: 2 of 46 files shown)


src/runtime_src/core/tools/common/SubCmd.h
```diff
@@ -20,22 +20,56 @@
...
 
-unsigned int register_subcommand(const std::string &_subCommandName, const std::string &_description, t_subcommand _pSubCommand, bool _isHidden = false);
-const SubCmdEntry *getSubCmdEntry(const std::string &_sSubCmdName);
-const SubCmdTable &getSubCmdsTable();
+ public:
+   virtual ~SubCmd() {};
 
+ // Child class Helper methods
+ protected:
+  SubCmd(const std::string _name, const std::string _shortDescription);
```
stsoe: Since you are not moving these value arguments in the ctor, please use const std::string&

src/runtime_src/core/tools/xbutil2/xbutil.cpp
```diff
@@ -14,21 +14,92 @@
...
 
+// Program entry
 int main( int argc, char** argv )
 {
+  // -- Build the supported subcommands
+  SubCmdsCollection subCommands;
+
+  {
+    // Syntax: SubCmdClass( IsHidden, IsDepricated, IsPreliminary)
+    subCommands.emplace_back(new  SubCmdExamine(false, false, false));
```
stsoe: std::make_shared

src/runtime_src/core/tools/xbutil2/xbutil.cpp
```diff
@@ -14,21 +14,92 @@
...
+    subCommands.emplace_back(new  SubCmdProgram(false, false, false));
+    subCommands.emplace_back(new SubCmdValidate(false, false, false));
+  }
+
+  // Add depricated commands
+  #ifdef ENABLE_DEPRECATED_2020_1_SUBCMDS
+  {
+    // Syntax: SubCmdClass( IsHidden, IsDepricated, IsPreliminary)
+    subCommands.emplace_back(new   SubCmdClock(false, true, false));
```
stsoe: std::make_shared

src/runtime_src/core/tools/xbutil2/xbutil.cpp
```diff
@@ -14,21 +14,92 @@
...
+    subCommands.emplace_back(new     SubCmdTop(false, true, false));
+    subCommands.emplace_back(new SubCmdVersion(false, true, false));
+  }
+  #endif
+
+  // -- Determine and set the executable name for each subcommand
+  boost::filesystem::path pathAndFile(argv[0]);
+  const std::string executable = pathAndFile.filename().string();
+
+  for (auto subCommand : subCommands) {
```
stsoe: auto&

src/runtime_src/core/tools/common/XBMain.h
```diff
@@ -19,13 +19,26 @@
...
+class SubCmd;
   
-  typedef enum {
-    RC_SUCCESS = 0,
-    RC_ERROR_IN_COMMAND_LINE = 1,
-    RC_ERROR_UNHANDLED_EXCEPTION = 2
-  } ReturnCodes;
+// --------- E N U M E R A T I O N S   /   T Y P E D E F S--------------------
+typedef enum {
+  RC_SUCCESS = 0,
```
stsoe: Is RC_ too generic?  You are definitely not the first one to use RC_ (just google it).

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -159,4 +162,229 @@ XBUtilities::trace_print_tree(const std::string & _name,
...
+
+      std::string completeOptionName = option->canonical_display_name(po::command_line_style::allow_dash_for_short);
+      buffer << " " << completeOptionName << " arg";
+    }
+
+  return buffer.str();
+}
+
+void 
+XBUtilities::wrap_paragraph( const std::string _unformattedString,
```
stsoe: std::string&

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -159,4 +162,229 @@ XBUtilities::trace_print_tree(const std::string & _name,
...
+    // Write out the line
+    _formattedString.append(lineBeginIter, lineEndIter);
+
+    lineBeginIter = lineEndIter;              
+    linesProcessed++;
+  }
+}   
+
+void 
+XBUtilities::wrap_paragraphs( const std::string _unformattedString,
```
stsoe: std::string&

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -159,4 +162,229 @@ XBUtilities::trace_print_tree(const std::string & _name,
...
+    // Remove leading spaces
+    if ((linesProcessed > 0) && 
+        (*lineBeginIter == ' ')) {
+      lineBeginIter++;
+      continue;
+    }
+
+    // Determine the end-of-the line to be examined
+    std::string::const_iterator lineEndIter = lineBeginIter;
+    int64_t remainingChars = std::distance(lineBeginIter, paragraphEndIter);
```
stsoe: int64_t?

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -159,4 +162,229 @@ XBUtilities::trace_print_tree(const std::string & _name,
...
+      }
+
+      if (firstShortFlagFound == true) 
+        buffer << "]";
+    }
+
+     
+    // Gather up the long simple flags (flags with no short versions)
+    {
+      for (auto option : options) {
```
stsoe: auto&

src/runtime_src/core/tools/common/XBUtilities.h
```diff
@@ -66,7 +69,28 @@ std::string format(const std::string& format, Args ... args) {
...
 
-  void trace_print_tree(const std::string & _name, const boost::property_tree::ptree & _pt);
+  void trace_print_tree(const std::string & _name, 
+                        const boost::property_tree::ptree & _pt);
+
+  // ---------
+  std::string create_usage_string( const std::string &_executableName,
+                                   const std::string &_subCommand,
+                                   const boost::program_options::options_description &_od);
+  void wrap_paragraph( const std::string _unformattedString,
```
stsoe: std::string&

src/runtime_src/core/tools/common/XBUtilities.h
```diff
@@ -66,7 +69,28 @@ std::string format(const std::string& format, Args ... args) {
...
+  // ---------
+  std::string create_usage_string( const std::string &_executableName,
+                                   const std::string &_subCommand,
+                                   const boost::program_options::options_description &_od);
+  void wrap_paragraph( const std::string _unformattedString, 
+                       unsigned int _indentWidth, 
+                       unsigned int _columnWidth, 
+                       bool _indentFirstLine,
+                       std::string &_formattedString);
+  void wrap_paragraphs( const std::string _unformattedString,
```
stsoe: std::string&

src/runtime_src/core/tools/xbutil2/SubCmdExamine.h
```diff
@@ -0,0 +1,37 @@
...
+class SubCmdExamine : public SubCmd {
+ public:
+   virtual void execute(const SubCmdOptions &_options) const;
+
+ public:
+   SubCmdExamine(bool _isHidden, bool _isDepricated, bool _isPreliminary);
+   virtual ~SubCmdExamine() {};
+
+ // Methods not supported
+ private:
```
stsoe: Lets simplify all derived class and remove the explicit delete of these functions which will not be created by the compiler anyway.   You can keep the deleted assignment in the base class, but the ctors don't matter because the is already a non default ctor defined.

The destructor also doesn't have any effect since the default one will do just fine.

src/runtime_src/core/tools/xbutil2/xbutil.cpp
rozumx: done

src/runtime_src/core/tools/xbutil2/xbutil.cpp
rozumx: done

src/runtime_src/core/tools/common/SubCmd.h
rozumx: done

src/runtime_src/core/tools/xbutil2/xbutil.cpp
rozumx: done

src/runtime_src/core/tools/common/XBMain.h
rozumx: Removed the return type entirely.  We should be using exceptions instead and with my next set of submissions, I will update the code accordingly.

src/runtime_src/core/tools/common/XBUtilities.cpp
rozumx: std::distance returns an uint64_t value, and I had assigned it to an unsigned int value originally.  While this worked nicely on Linux it failed on Windows, hence the uint64_t.

Since this is causing some confusion, I instead replaced it with an auto which works both on windows and linux.

src/runtime_src/core/tools/common/XBUtilities.h
rozumx: done

src/runtime_src/core/tools/xbutil2/SubCmdExamine.h
rozumx: I understand where you are coming from and it makes sense to remove this destructor, which I did.  

That said, I don't like it when the the "compiler" creates methods on the fly with sometimes unexpected behavior, hence why I originally added it (a habit to do so).  And yes, it does reading the header file easier as well.

src/runtime_src/core/tools/common/XBUtilities.cpp
rozumx: done

src/runtime_src/core/tools/common/XBUtilities.h
rozumx: done

PR #2672, opened 2020-01-14
Update testcase metadata to add include xilinx_u50_xdma_201920_2 into exclude platform list 

haeseunglee:
Update testcase metadata to add include xilinx_u50_xdma_201920_2 into exclude platform list


maxzhen:
ok to test

PR #2684, opened 2020-01-16
Enhance clGetDeviceInfo to return BDF for a device

stsoe:
This is applicable only for HW flow, emulation defaults to unknown.

Current implementation is not clean, it uses a sysfs path to parse for
BDF.  In future this should go through device classes used by xbutil
and xbmgmt, but more restructuring is needed in order to do this.

Prepare device_linux for accessing more than sysfs data in query
function.


stsoe:
@AShivangi @rozumx Just focus on device_linux changes when reviewing.

PR #2693, opened 2020-01-17
xbmgmt2 flash [factory_reset and scan]

AShivangi:
- separated get_total_devices for user and mgmt pf
- factory_reset works on linux and windows
- flash --scan works on linux and windows
- flash --scan on windows doesn't work when in the golden state. Might is a driver issue, will look into it next week
- a Coverity fix

### DIFF
--- a/src/runtime_src/core/common/device.h
+++ b/src/runtime_src/core/common/device.h
@@ -142,7 +142,8 @@ public:
       QR_FLASH_BAR_OFFSET,
       QR_IS_MFG,
       QR_F_FLASH_TYPE,
-      QR_FLASH_TYPE
+      QR_FLASH_TYPE,
+      QR_BOARD_NAME
   };
 public:
 
--- a/src/runtime_src/core/common/system.cpp
+++ b/src/runtime_src/core/common/system.cpp
@@ -45,12 +45,6 @@ get_devices(boost::property_tree::ptree& pt)
   instance().get_devices(pt);
 }
 
-void
-scan_devices(bool verbose, bool json)
-{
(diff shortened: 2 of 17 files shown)


chvamshi-xilinx:
Hi @AShivangi , Can you please check whether edge platforms are building with this change or not
@Himanshu-xilinx , Please help Shivangi to verify her changes on edge platforms. I suspect, there would be a build failure for edge platforms.

rozumx:
Hello @chvamshi-xilinx, 

These changes are associated with xbmgmt2 which is currently being built for Linux and Windows only.  The edge tools currently have their own modified copy of xbmgmt & xbutil that is contained with the edge directory structure. 

For supporting the edge platforms, an abstract edge library needs to be created and potentially some "pipe-cleaning" to address some compiler differences.

Himanshu-xilinx:
> Hello @chvamshi-xilinx,
> 
> Please disregard my previous comment, I see that @Himanshu-xilinx has started to create the abstract edge library. Being that the XRT-PIPELINE "should" be building are targeted platforms I am "hoping" that any build issues are caught here :-).
> 
> > These changes are associated with xbmgmt2 which is currently being built for Linux and Windows only.  The edge tools currently have their own modified copy of xbmgmt & xbutil that is contained with the edge directory structure.
> > For supporting the edge platforms, an abstract edge library needs to be created and potentially some "pipe-cleaning" to address some compiler differences.

Hi @rozumx  
libxdp.so is common for both edge and pcie platform. It uses some of these API's from "src/runtime_src/core/common/" which lead to runtime errors in edge platform.

PR #2694, opened 2020-01-17
Add power profile in run summary and header for device

hackwa:


src/runtime_src/xdp/profile/plugin/ocl/ocl_power_profile.cpp
```diff
@@ -10,6 +10,7 @@ OclPowerProfile::OclPowerProfile(xrt::device* xrt_device,
     target_device = xrt_device;
     target_xocl_plugin = xocl_plugin;
     target_unique_name = unique_name;
+    output_file_name = "ocl_power_profile_" + target_unique_name + ".csv";
```
pgschuey: Make this:
output_file_name = "power_profile_" + target_unique_name + ".csv";

hackwa: Agreed. OCL shouldn't be there

PR #2700, opened 2020-01-20
Added hot plug support for on-going traffic case. Before removal we a…

saifuddin-xilinx:
This pull request has the following changes:

- Before performing hot removal we are shutdown the userpf.  This will prevent system crash while traffic is going on.
- xbmgmt will poll till shutdown finishes and then try to remove userpf and mgmt pf 
- If we are running xbmgmt on VM environment then it will print a message and gracefully exits.


chvamshi-xilinx:
ok to test

src/runtime_src/core/pcie/tools/xbmgmt/cmd_hotplug.cpp
```diff
@@ -114,6 +134,46 @@ int hotplugHandler(int argc, char *argv[])
...
+{
+    std::string errmsg;
+    int shutdownStatus = -EINVAL;
+    int wait  = 0;
+
+    /* "echo 1 > /sys/bus/pci/<EndPoint>/shutdown" to trigger shutdown of the device */
+    std::string path = dev->get_sysfs_path("", "shutdown");
+    if (!boost::filesystem::exists(path)) 
+    {
+        return -EEXIST;
```
maxzhen: shouldn't you return ENOENT?

saifuddin-xilinx: Thanks for the comments.  I will make the changes and publish them again.

src/runtime_src/core/pcie/tools/xbmgmt/cmd_hotplug.cpp
```diff
@@ -29,9 +29,11 @@
 #define SYSFS_PATH      "/sys/bus/pci/devices"
 #define XILINX_VENDOR   "0x10ee"
 #define XILINX_US       "0x9134"
+#define POLL_TIMEOUT    60      /* Set a pool timeout as 60sec */
 
 static int hotplugRescan(void);
 static int removeDevice(const std::shared_ptr<pcidev::pci_device> dev);
+static int shutdownDevice(const std::shared_ptr<pcidev::pci_device> dev);
```
stsoe: change to 'const pcidev::pci_device* dev' and call as 'shutdownDevice(uDev.get())'
Ditto for removeDevice.

saifuddin-xilinx: I have updated the code. But it is a bit different way. Passed device index as a parameter. The same implementation is followed in other files. Please review it once.

chvamshi-xilinx:
retest this please

src/runtime_src/core/pcie/tools/xbmgmt/cmd_hotplug.cpp
```diff
@@ -29,9 +29,11 @@
 #define SYSFS_PATH      "/sys/bus/pci/devices"
 #define XILINX_VENDOR   "0x10ee"
 #define XILINX_US       "0x9134"
+#define POLL_TIMEOUT    60      /* Set a pool timeout as 60sec */
 
 static int hotplugRescan(void);
-static int removeDevice(const std::shared_ptr<pcidev::pci_device> dev);
+static int removeDevice(unsigned index, bool is_userpf);
```
stsoe: Please specify type.     unsigned what?

PR #2704, opened 2020-01-22
APIs for user manage CU interrupt

mamin506:
1. add prototype of user manage interrupt APIs
2. Implement for edge platforms.
3. Since ERT/MB could not support custom IP, not worth to implement on PCIe device now.
4. Once we have platform with ERT/PS and proper hardware design, could implement APIs at such kind of platform. (Not for tnow)

**NOTE**
For consistent purpose, after discussed with Max and Soren, use IP index on XRT APIs.
The name of below API in xrt-next.h are changed:
xclCuName2Index() -> xclIPName2Index() 

Below APIs description are changed, but the usage is the same (no impact for test):
xclOpenContext()
xclCloseContext()
xclRegRead()
xclRegWrite()


mamin506:
> These custom IPs, they are not really CUs, right? It's just a name, so I really don't feel strongly about this, just want to understand the thinking

The ip_index is a bug. I confirmed it with @maxzhen . The user should use CU index only. The CU index and Ip index (which is index in IP_LAYOUT) might be in different order. Our implementation is actually using CU index ordering.

I don't know if I should introduce another type of accelerator into XRT. Like 'custom_ip'? Then we will needs new lists in the driver to menage them. More important is that we need more user APIs to manage them.

In my thought, the interrupt resources are shared with "cu" and "custom ip". I would say "custom ip" is a set of "cu" with a "AP_CTRL_UNKNOWN" as m_ip_control type. Then we could leverage current load xclbin flow. The KDS creates a CU list and maintain all of it. The only different is that KDS needs to mark those unknown control type CU and do nothing on it. So, treating "custom ip" as CU doesn't harm current KDS flow.

The other benefit is this is flexible for the user. For example, when I test new APIs, I could still use HLS CUs with ap_ctrl_hs or ap_ctrl_chain.

stsoe:
The 'cu' as in cu_index is wrong then.   When we add these custom ips to the mix, they too will be indexed by XRT, but they will not be CUs.   Anyway, it is a minor point, we can call it whatever we want, but personally I'd have preferred ip_index, not to be confused with the index in the IP_LAYOUT.

mamin506:
I don't want to confuse the index in the IP_LAYOUT too.
But in current xrt.h, the comment is " * @ipIndex:       IP/CU index in the IP LAYOUT array".

I though Max's xclCuName2Index() should be the only way that user get a index (for construct start CU command or xclRegRead/write() or interrupt notify). What do you think?

stsoe:
Yeah, that comment is wrong.   And cuname2idx is the only way, correct.
Probably good to be consistent then, even though 'cu' is not the right acronym for covering all the types of IP we index.  And I agree that we should *not* introduce a second list in XRT; there is one list which contains all IPs with assigned indices regardless of type.

mamin506:
> Yeah, that comment is wrong. And cuname2idx is the only way, correct.
> Probably good to be consistent then, even though 'cu' is not the right acronym for covering all the types of IP we index. And I agree that we should _not_ introduce a second list in XRT; there is one list which contains all IPs with assigned indices regardless of type.

:) Totally agree.

maxzhen:
> 
> 
> Yeah, that comment is wrong. And cuname2idx is the only way, correct.
> Probably good to be consistent then, even though 'cu' is not the right acronym for covering all the types of IP we index. And I agree that we should _not_ introduce a second list in XRT; there is one list which contains all IPs with assigned indices regardless of type.

maybe we should change xclCuName2Index to xclIpName2Index?

src/runtime_src/core/edge/drm/zocl/zocl_generic_cu.c
```diff
@@ -0,0 +1,222 @@
...
+	poll_wait(filp, &gcu->waitq, wait);
+
+	/* Safe to read only one object without lock */
+	if (gcu->event > 0)
+		ret = POLLIN;
+
+	return ret;
+}
+
+static ssize_t generic_cu_read(struct file *filp, char __user *buf,
```
maxzhen: the implementation of blocking and waiting looks quite complicated here. i'm wondering if you should just call wait_event_interruptible()? (see mailbox_read() for example).

src/runtime_src/core/edge/drm/zocl/zocl_generic_cu.c
```diff
@@ -0,0 +1,222 @@
...
+	spin_lock_init(&gcu->lock);
+
+	fd = anon_inode_getfd("[generic_cu]", &generic_cu_fops, gcu, O_RDWR);
+	if (fd < 0)
+		kfree(gcu);
+
+	return fd;
+}
+
+int zocl_open_gcu(struct drm_zocl_dev *zdev, struct drm_zocl_ctx *ctx,
```
maxzhen: if you're trying to implement a CU subdevice driver, passing in struct drm_zocl_dev seems to bring in unnecessary coupling b/w zocl and CU subdevice. let's chat.

mamin506:
> We need to decide if we use "cuIndex" or "ipIndex". I don't think you want to use "cu_index" which is c style variable name in *.cxx files...

That is just a name for me. Please let me know which one I should use. I could make changes on all related APIs.

xdavidz:
@mamin506 The changes of context switch look ok for me, but I don't have enough time to go over all code today. I can get it done by this weekend, if you got enough approval for this PR, you may go ahead.

src/runtime_src/core/edge/drm/zocl/zocl_generic_cu.c
mamin506: This is not a subdevice driver yet. But I would like generig_cu to be a subdevice driver in the future. This function is the only place that coupling zocl. We could easily change this later on.

I would add few callback functions in the generic_cu_info, so generic_cu could be the base driver for all other possible IPs and their variations.

src/runtime_src/core/edge/drm/zocl/zocl_generic_cu.c
mamin506: Thanks. Done.

maxzhen:
retest this please

PR #2707, opened 2020-01-22
azure plugin need to maintain static serial number info

xuhz:
azure today gets serial number from sysfs.
For hostreset support through mailbox, when azure gets the mailbox req, the sysfs info is gone already.
So azure needs to maintain static serial number info

### DIFF
--- a/src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.h
+++ b/src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.h
@@ -72,11 +72,27 @@ public:
         cfile.close();    
         return "";
     };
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.h
```diff
@@ -72,11 +72,27 @@ class AzureDev
         cfile.close();    
         return "";
     };
+    static std::vector<std::string> get_serial_number()
+    {
+        std::vector<std::string> ret;
```
arpitxilinx: This ret is uninitialized and could possibly return some garbage value if no device found.

PR #2709, opened 2020-01-22
add timstamp sysfs node

chienwei-lan:
http://jira.xilinx.com/browse/CR-1047682

Add timestamp sysfs node to show when the sensor was updated(in ms).


src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
```diff
@@ -1176,6 +1176,21 @@ static ssize_t cache_expire_secs_store(struct device *dev,
 }
 static DEVICE_ATTR_RW(cache_expire_secs);
 
+static ssize_t timestamp_show(struct device *dev,
```
maxzhen: "timestamp" is too generic. use "sensor_update_timestamp"?

maxzhen:
@chienwei-lan please solve the conflict.

PR #2711, opened 2020-01-23
Cma bo phase 2

chienwei-lan:


src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -743,6 +743,61 @@ int shim::p2pEnable(bool enable, bool force)
     return p2p_enable;
 }
 
+int shim::cmaEnable(bool enable, uint64_t size, uint64_t num, bool force)
+{
+    const std::string input = "1\n";
```
stsoe: Let's move all this declarations and definitions to the scope where they are used.  If you do so, you will find that at least one of these variables is not used at all.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -743,6 +743,61 @@ int shim::p2pEnable(bool enable, bool force)
...
+        else if (size == (2 << 20))
+            hugepage_flag = 0x15;
+
+        if (!hugepage_flag)
+            return -EINVAL;
+
+        for ( i = 0; i < num; ++i ) {
+            addr_local = mmap(0x0, size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_HUGETLB | hugepage_flag << MAP_HUGE_SHIFT, 0, 0);
+            if (addr_local == MAP_FAILED) {
+                std::cout<<"Failed to mmap"<<std::endl;
```
stsoe: This should use our message infrastructure.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -743,6 +743,61 @@ int shim::p2pEnable(bool enable, bool force)
...
+        drm_xocl_alloc_cma_info cma_info;
+        cma_info.page_sz = size;
+        cma_info.nr_page = num;
+        cma_info.reserve = true;
+
+        if (num > 4 || num == 0)
+            return -EINVAL;
+
+        if (size == (1 << 30)) 
+            hugepage_flag = 0x1e;
```
stsoe: what do do these constant values mean?

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -1792,6 +1794,99 @@ int xcldev::xclP2p(int argc, char *argv[])
...
+
+int xcldev::device::setCma(bool enable, uint64_t sz, uint64_t num, bool force)
+{
+    return xclCmaEnable(m_handle, enable, sz, num, force);
+}
+
+int xcldev::xclCma(int argc, char *argv[])
+{
+    int c;
+    unsigned index = 0;
```
stsoe: unsigned int

src/runtime_src/core/pcie/linux/shim.cpp
maxzhen: > 
> 
> that's huge page size in log, 0x1e = 30, 2^30 = 0x40000000 = 1G,
> 
> Once set MAP_HUGETLB, we have to specify bit[26~31] as size

We either need a block of comment to explain 0x1e or a well-named const variable to replace 0x1e here. It's difficult to understand what is happening here with just a const value - 0x1e.

src/runtime_src/core/pcie/driver/linux/include/xocl_ioctl.h
```diff
@@ -462,13 +466,22 @@ struct drm_xocl_reclock_info {
 	unsigned short ocl_target_freq[DRM_XOCL_NUM_SUPPORTED_CLOCKS];
 };
 
+
+struct drm_xocl_alloc_cma_info {
+	uint64_t page_sz;
+	uint64_t nr_page;
+	bool	 reserve;
+	uint64_t user_addr[DRM_XOCL_CMA_CHUNK_MAX];
```
maxzhen: this is not flexible enough for when max number of chunks changes later. can we just do one ioctl for one chunk?

src/runtime_src/core/pcie/driver/linux/xocl/userpf/common.h
```diff
@@ -98,6 +98,9 @@
 #define XOCL_P2P_CHUNK_SHIFT		28
 #define XOCL_P2P_CHUNK_SIZE		(1ULL << XOCL_P2P_CHUNK_SHIFT)
 
+
+
+
```
maxzhen: why we have these empty lines?

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
```diff
@@ -220,10 +224,23 @@ int xocl_gem_fault(struct vm_area_struct *vma, struct vm_fault *vmf)
 		ret = vm_insert_mixed(vma, vmf_address, pfn);
 #else
 		ret = vm_insert_page(vma, vmf_address, xobj->pages[page_offset]);
+#endif
+	} else if (xocl_bo_cma(xobj)) {
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(4, 5, 0) || defined(RHEL_RELEASE_VERSION)
```
maxzhen: add some comments here about why you can't insert struct page, but a pfn?

PR #2713, opened 2020-01-23
added information if incase of XCL_EMULATION_MODE is not set

vboggara-xilinx:
added information if incase of XCL_EMULATION_MODE is not set

### DIFF
--- a/src/runtime_src/xrt/device/hal.cpp
+++ b/src/runtime_src/xrt/device/hal.cpp
@@ -89,7 +89,15 @@ dllpath(const boost::filesystem::path& root, const std::string& libnm)
 static bool
 isEmulationMode()
 {
-  static bool val = (std::getenv("XCL_EMULATION_MODE") != nullptr);
+  static int count = 0;
+  static bool val = (std::getenv("XCL_EMULATION_MODE") != nullptr);  
+  if (!val) {
+    if (!count) {
+      std::string initMsg = "INFO: XCL_EMULATION_MODE is not set, make sure that this environment variable is set to valid value for emulation";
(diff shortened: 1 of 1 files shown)


src/runtime_src/xrt/device/hal.cpp
```diff
@@ -89,7 +89,15 @@ dllpath(const boost::filesystem::path& root, const std::string& libnm)
...
-  static bool val = (std::getenv("XCL_EMULATION_MODE") != nullptr);
+  static int count = 0;
+  static bool val = (std::getenv("XCL_EMULATION_MODE") != nullptr);  
+  if (!val) {
+    if (!count) {
+      std::string initMsg = "INFO: XCL_EMULATION_MODE is not set, make sure that this environment variable is set to valid value for emulation";
+      std::cout << initMsg << std::endl;
+    }
+    count++;
+  }
```
stsoe: This logic does not look right. As far as I can tell the message is printed when running in non emulation mode.  The message makes no sense when running in HW.

vboggara-xilinx: @stsoe , yes soren i agree with you, is there any other way to handle this error?

chvamshi-xilinx: Hi @vboggara-xilinx , There is no way to findout whether you are running on emulation or hw. Main purpose of this Environment variable is to differentiate between both the modes.
This is well documented in all the user guides. We shouldnt give this message while running on board.

chvamshi-xilinx:
Closing this PR as this effects hardware runs too

PR #2722, opened 2020-01-24
Cma bo phase 2

chienwei-lan:


src/runtime_src/core/pcie/driver/linux/include/xocl_ioctl.h
```diff
@@ -462,13 +466,21 @@ struct drm_xocl_reclock_info {
 	unsigned short ocl_target_freq[DRM_XOCL_NUM_SUPPORTED_CLOCKS];
 };
 
+
+struct drm_xocl_alloc_cma_info {
+	uint64_t page_sz;
+	uint64_t user_addr;
+	bool	 reserve;
+};
```
maxzhen: + this data structure is also used to free the previously allocated cma memory. it's confusing to see "alloc" in the name. it'd be cleaner to have one data structure for alloc and one for free.
+ i don't seem to find how the ioctl handler verifies the input memory buffer? driver can't trust any random input from user land.

PR #2724, opened 2020-01-27
Add a formatting library that is terminal aware

rbramand-xilinx:
> This pull request contains functionality of Ansi Escape Sequences for formatting Output on Terminal.


src/runtime_src/core/common/system.cpp
```diff
@@ -69,4 +69,22 @@ bdf2index(const std::string& bdfStr)
...
+setup_terminal()
+{
+  instance().setup_terminal();
+}
+
+void
+restore_terminal()
+{
+  instance().restore_terminal();
+}
```
stsoe: How are these APIs used?  What happens in setup is called without restore?   
I ask out of concern for exception safety.

src/runtime_src/core/tools/common/Escape_Codes.h
```diff
@@ -0,0 +1,81 @@
+/**
```
stsoe: Can we name this file lower case: "escape_codes.h"?

src/runtime_src/core/tools/common/Escape_Codes.h
```diff
@@ -0,0 +1,81 @@
...
+ * under the License.
+ */
+
+#ifndef __ESCAPE_CHAR_H__
+#define __ESCAPE_CHAR_H__
+
+#include <iostream>
+#include <string>
+
+namespace ec {
```
stsoe: spell it out  would be better..  It's nice when namespace (or class) matches the name of the file.

src/runtime_src/core/tools/common/Escape_Codes.h
```diff
@@ -0,0 +1,81 @@
...
+#define EC_CYAN  6
+#define EC_WHITE  7
+#define EC_BRIGHT_BLACK 8
+#define EC_BRIGHT_RED  9
+#define EC_BRIGHT_GREEN  10
+#define EC_BRIGHT_YELLOW  11
+#define EC_BRIGHT_BLUE  12
+#define EC_BRIGHT_MAGENTA  13
+#define EC_BRIGHT_CYAN  14
+#define EC_BRIGHT_WHITE  15
```
stsoe: Replace with constant strings that are scoped to the namespace, and lower case the string names.
But that said, I am not sure where and how these constants are used?
Comments needed.

src/runtime_src/core/tools/common/Escape_Codes.h
```diff
@@ -0,0 +1,81 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef __ESCAPE_CHAR_H__
+#define __ESCAPE_CHAR_H__
+
+#include <iostream>
```
stsoe: why is iostream include?

src/runtime_src/core/tools/common/Escape_Codes.h
```diff
@@ -0,0 +1,81 @@
...
+const std::string show_cursor() { static const std::string s("\033[?25h"); return s; };
+const std::string hide_cursor() { static const std::string s("\033[?25l"); return s; };
+const std::string enable_alternative_screen_buffer() { static const std::string s("\033[?1049h"); return s; };
+const std::string disable_alternative_screen_buffer() { static const std::string s("\033[?1049l"); return s; };
+const std::string fgcolor_reset() { static const std::string s("\033[39m"); return s; };
+const std::string bgcolor_reset() { static const std::string s("\033[49m"); return s; };
+const std::string save_cursor_pos() { static const std::string s("\0337"); return s; };
+const std::string restore_cursor_pos() { static const std::string s("\0338"); return s; };
+const std::string text_reset() { static const std::string s("\033[0m"); return s; };
+const std::string reset_all() { static const std::string s("\033c"); return s; };
```
stsoe: Do not return 'const value_type'   const offers nothing over simply returning "value_type"
Maybe these functions should return const std::string& ?

src/runtime_src/core/common/system.cpp
rbramand-xilinx: Ansi Escape Sequences are supported on Unix like systems by default but Windows does not have that support. From Windows 10 Version 1511 this support is provided but is not enabled by default, Our program should do that. Here setup_terminal function does that and when you call restore_terminal  Ansi Escape chars are not interpreted from that point.

If setup is called without restore terminal still can interpret Ansi Escape Chars if provided in the program.

src/runtime_src/core/tools/common/Escape_Codes.h
rbramand-xilinx: Actually these values are passed as parameters to functions like fgcolour and bgcolour where those functions expects 'int' values. Here I have defined color values that are generally used, but we can pass any value between 0 to 255 to generate different colors, so the parameter should be an integer.

stsoe: Doesn't sound like these should be #defines.  Please change.

src/runtime_src/core/common/system.cpp
stsoe: Please let me know how the API is used.  If the code is 
```
setup();
dosomething();
restore();
```
Then it is not valid code and must be changed.

src/runtime_src/core/tools/common/Escape_Codes.h
rbramand-xilinx: with the latest commit I have changed them to const uint8_t values, please check in files changed

stsoe:
Closed since this particular PR will not be fixed by @rbramand-xilinx.

PR #2731, opened 2020-01-28
xbmgmt2 flash --update 

AShivangi:
- works on windows and linux
- versal flashing code is in here but doesn't work. I'll have to move around some functions to get it working (work for next PR)

### DIFF
--- a/src/runtime_src/core/pcie/linux/scan.cpp
+++ b/src/runtime_src/core/pcie/linux/scan.cpp
@@ -270,7 +270,7 @@ static bool is_admin()
 
 int pcidev::pci_device::open(const std::string& subdev, int flag)
 {
-    if (is_mgmt && !is_admin())
+    if (is_mgmt && !::is_admin())
         throw std::runtime_error("Root privileges required");
     // Open xclmgmt/xocl node
     if (subdev.empty()) {
--- a/src/runtime_src/core/pcie/linux/scan.cpp
+++ b/src/runtime_src/core/pcie/linux/scan.cpp
@@ -263,8 +263,15 @@ static std::string get_devfs_path(bool is_mgmt, uint32_t instance)
     return prefixStr + instStr;
 }
 
+static bool is_admin()
+{     
+    return (getuid() == 0) || (geteuid() == 0);
+}
+
 int pcidev::pci_device::open(const std::string& subdev, int flag)
 {
+    if (is_mgmt && !is_admin())
+        throw std::runtime_error("Root privileges required");
     // Open xclmgmt/xocl node
     if (subdev.empty()) {
         std::string devfs = get_devfs_path(is_mgmt, instance);
--- a/src/runtime_src/core/pcie/windows/system_windows.cpp
+++ b/src/runtime_src/core/pcie/windows/system_windows.cpp
@@ -116,11 +116,7 @@ std::pair<device::id_type, device::id_type>
 system_windows::
 get_total_devices(bool is_user) const
 {
-  unsigned int count = 0;
-  if (is_user)
-    count = xclProbe();
-  else
-    count = mgmtpf::probe();
+  unsigned int count = is_user ? xclProbe() : mgmtpf::probe();
   return std::make_pair(count, count);
 }
 
--- a/src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
@@ -151,6 +151,7 @@ static DSAInfo
 selectShell(uint16_t idx, std::string& dsa, std::string& id)
 {
   uint16_t candidateDSAIndex = std::numeric_limits<uint16_t>::max();
+  boost::format fmtStatus("%|8t|Status: %s"); 
 
   Flasher flasher(idx);
   if(!flasher.isValid())
@@ -162,11 +163,11 @@ selectShell(uint16_t idx, std::string& dsa, std::string& id)
   if (dsa.empty()) {
     std::cout << "Card [" << flasher.sGetDBDF() << "]: " << std::endl;
     if (installedDSA.empty()) {
-      std::cout << "\t Status: no shell is installed" << std::endl;
+      std::cout << fmtStatus % "no shell is installed" << std::endl;
       return DSAInfo("");
     }
     if (installedDSA.size() > 1) {
-      std::cout << "\t Status: multiple shells are installed" << std::endl;
+      std::cout << fmtStatus % "multiple shells are installed" << std::endl;
       return DSAInfo("");
     }
     candidateDSAIndex = 0;
@@ -178,14 +179,14 @@ selectShell(uint16_t idx, std::string& dsa, std::string& id)
       if (!id.empty() && !idsa.matchId(id))
         continue;
       if (candidateDSAIndex != std::numeric_limits<uint16_t>::max()) {
-        std::cout << "\t Status: multiple shells are installed" << std::endl;
+        std::cout << fmtStatus % "multiple shells are installed" << std::endl;
         return DSAInfo("");
       }
       candidateDSAIndex = i;
     }
   }
   if (candidateDSAIndex == std::numeric_limits<uint16_t>::max()) {
-    std::cout << "WARNING: Failed to flash Card["
+    std::cout << "ERROR: Failed to flash Card["
       << flasher.sGetDBDF() << "]: Specified shell is not applicable" << std::endl;
     return DSAInfo("");
   }
@@ -195,18 +196,18 @@ selectShell(uint16_t idx, std::string& dsa, std::string& id)
   bool same_bmc = false;
   DSAInfo currentDSA = flasher.getOnBoardDSA();
   if (!currentDSA.name.empty()) {
-    same_dsa = (candidate.name == currentDSA.name &&
-      candidate.matchId(currentDSA));
-    same_bmc = (currentDSA.bmcVer.empty() ||
-      candidate.bmcVer == currentDSA.bmcVer);
+    same_dsa = ((candidate.name == currentDSA.name) &&
+      (candidate.matchId(currentDSA)));
+    same_bmc = ((currentDSA.bmcVer.empty()) ||
+      (candidate.bmcVer == currentDSA.bmcVer));
   }
   if (same_dsa && same_bmc) {
-    std::cout << "\t Status: shell is up-to-date" << std::endl;
+    std::cout << fmtStatus % "shell is up-to-date" << std::endl;
     return DSAInfo("");
   }
-  std::cout << "\t Status: shell needs updating" << std::endl;
-  std::cout << "\t Current shell: " << currentDSA.name << std::endl;
-  std::cout << "\t Shell to be flashed: " << candidate.name << std::endl;
+  std::cout << fmtStatus % "shell needs updating" << std::endl;
+  std::cout << boost::format("%|8t| %s") % "Current shell: " << currentDSA.name << std::endl;
+  std::cout << boost::format("%|8t| %s") % "Shell to be flashed: " << candidate.name << std::endl;
   return candidate;
 }
 
@@ -243,10 +244,8 @@ updateShellAndSC(uint16_t boardIdx, DSAInfo& candidate, bool& reboot)
   reboot = false;
 
   Flasher flasher(boardIdx);
-  if(!flasher.isValid()) {
-    std::cout << "card not available" << std::endl;
-    return -EINVAL;
-  }
+  if(!flasher.isValid())
+    throw xrt_core::error(boost::str(boost::format("%d is an invalid index") % boardIdx));
 
   bool same_dsa = false;
   bool same_bmc = false;
@@ -263,7 +262,8 @@ updateShellAndSC(uint16_t boardIdx, DSAInfo& candidate, bool& reboot)
   if (!same_bmc) {
     std::cout << "Updating SC firmware on card[" << flasher.sGetDBDF() <<
       "]" << std::endl;
-    auto ret = 0;update_SC(boardIdx, candidate.file.c_str());
+    auto ret = 0;
+    update_SC(boardIdx, candidate.file.c_str());
     if (ret != 0) {
       std::cout << "WARNING: Failed to update SC firmware on card ["
         << flasher.sGetDBDF() << "]" << std::endl;
@@ -301,7 +301,7 @@ auto_flash(uint16_t index, std::string& name,
 
   // Sanity check input dsa and timestamp.
   if (!name.empty()) {
-        bool foundDSA = false;
+    bool foundDSA = false;
     bool multiDSA = false;
     auto installedDSAs = firmwareImage::getIntalledDSAs();
     for (DSAInfo& dsa : installedDSAs) {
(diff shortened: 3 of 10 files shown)


src/runtime_src/core/pcie/linux/scan.cpp
```diff
@@ -263,8 +263,15 @@ static std::string get_devfs_path(bool is_mgmt, uint32_t instance)
...
 }
 
+static bool is_admin()
+{     
+    return (getuid() == 0) || (geteuid() == 0);
+}
+
 int pcidev::pci_device::open(const std::string& subdev, int flag)
 {
+    if (is_mgmt && !is_admin())
```
rozumx: Scope is_admin() to be local (e.g., ::is_admin())

src/runtime_src/core/pcie/windows/system_windows.cpp
```diff
@@ -116,11 +116,7 @@ std::pair<device::id_type, device::id_type>
 system_windows::
 get_total_devices(bool is_user) const
 {
-  unsigned int count = 0;
-  if (is_user)
-    count = xclProbe();
-  else
-    count = mgmtpf::probe();
+  unsigned int count = is_user ? xclProbe() : mgmtpf::probe();
```
rozumx: Thank you :^)

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
```diff
@@ -95,49 +95,119 @@ static void
...
+  if(!flasher.isValid())
+    return DSAInfo("");
+
+  std::vector<DSAInfo> installedDSA = flasher.getInstalledDSA();
+
+  // Find candidate DSA from installed DSA list.
+  if (dsa.empty()) {
+    std::cout << "Card [" << flasher.sGetDBDF() << "]: " << std::endl;
+    if (installedDSA.empty()) {
+      std::cout << "\t Status: no shell is installed" << std::endl;
```
rozumx: Why a tab...for it may or may not be 8 characters.  
For a consistent message you could do the following:
```
boost::format fmtStatus("         Status: %s);
std::cout << fmtStatus % "No shell is installed.";
```
If you don't like the spaces, you could use fixed tabs, for example:
`boost::format fmtStatus("%|8t|Status: %s);`

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
```diff
@@ -95,49 +95,119 @@ static void
...
+      return DSAInfo("");
+    }
+    if (installedDSA.size() > 1) {
+      std::cout << "\t Status: multiple shells are installed" << std::endl;
+      return DSAInfo("");
+    }
+    candidateDSAIndex = 0;
+  } else {
+    for (uint16_t i = 0; i < installedDSA.size(); i++) {
+      DSAInfo& idsa = installedDSA[i];
```
rozumx: const - Since we are not changing it.

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
```diff
@@ -95,49 +95,119 @@ static void
...
+        continue;
+      if (candidateDSAIndex != std::numeric_limits<uint16_t>::max()) {
+        std::cout << "\t Status: multiple shells are installed" << std::endl;
+        return DSAInfo("");
+      }
+      candidateDSAIndex = i;
+    }
+  }
+  if (candidateDSAIndex == std::numeric_limits<uint16_t>::max()) {
+    std::cout << "WARNING: Failed to flash Card["
```
rozumx: Is this a warning or error condition?

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
```diff
@@ -95,49 +95,119 @@ static void
...
+    return DSAInfo("");
+  }
 
-    flasher.upgradeBMCFirmware(bmc.get());
+  DSAInfo& candidate = installedDSA[candidateDSAIndex];
+  bool same_dsa = false;
+  bool same_bmc = false;
+  DSAInfo currentDSA = flasher.getOnBoardDSA();
+  if (!currentDSA.name.empty()) {
+    same_dsa = (candidate.name == currentDSA.name &&
```
rozumx: A lot going on here.
To help the future user reading the code (instead of relying on order of operation), could you add a set of parenthesis around the first and second "equals" operators.  
```
same_dsa = ((candidate.name == currentDSA.name) && 
            (candidate.matchId(currentDSA)));
```

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
```diff
@@ -163,17 +233,159 @@ bool canProceed()
...
+ * Helper method for auto_flash
+ */
+static int 
+updateShellAndSC(uint16_t boardIdx, DSAInfo& candidate, bool& reboot)
+{
+  reboot = false;
+
+  Flasher flasher(boardIdx);
+  if(!flasher.isValid()) {
+    std::cout << "card not available" << std::endl;
```
rozumx: Shouldn't this be an error that is thrown?

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
```diff
@@ -163,17 +233,159 @@ bool canProceed()
...
+  }
+  if (same_dsa && same_bmc)
+    std::cout << "update not needed" << std::endl;
+
+  if (!same_bmc) {
+    std::cout << "Updating SC firmware on card[" << flasher.sGetDBDF() <<
+      "]" << std::endl;
+    auto ret = 0;update_SC(boardIdx, candidate.file.c_str());
+    if (ret != 0) {
+      std::cout << "WARNING: Failed to update SC firmware on card ["
```
rozumx: Hmm..It feels like if we failed one operation we shouldn't be continuing the flow.  Also it should be an error not warning.

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
```diff
@@ -163,17 +233,159 @@ bool canProceed()
...
-  index = index;
-  name = name;
-  id = id;
-  force = force;
+  std::vector<uint16_t> boardsToCheck;
+  std::vector<std::pair<uint16_t, DSAInfo>> boardsToUpdate;
+
+  // Sanity check input dsa and timestamp.
+  if (!name.empty()) {
+        bool foundDSA = false;
```
rozumx: Indention issues

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
```diff
@@ -163,17 +233,159 @@ bool canProceed()
...
+    bool multiDSA = false;
+    auto installedDSAs = firmwareImage::getIntalledDSAs();
+    for (DSAInfo& dsa : installedDSAs) {
+      if (name == dsa.name &&
+        (id.empty() || dsa.matchId(id))) {
+        if (!foundDSA)
```
rozumx: Wouldn't it be better to do the following:
```
multiDSA = multiDSA || foundDSA;
foundDSA = true;

```

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
```diff
@@ -163,17 +233,159 @@ bool canProceed()
...
+      }
+    }
+    if (!foundDSA)
+      throw xrt_core::error("Specified shell not found");
+    if (multiDSA)
+      throw xrt_core::error("Specified shell matched multiple installed shells");
+  }
+
+  // Collect all indexes of boards need checking
+  auto total = xrt_core::get_total_devices(false).first;
```
rozumx: We will be needing to re-evaluating this existing algorithm.  For now, it is O.K., but next week we will need to re-examine the use-cases and refactor this functionality to meet those use-cases.

src/runtime_src/core/tools/xbmgmt2/flash/xospiversal.cpp
```diff
@@ -0,0 +1,53 @@
+/**
+ * Copyright (C) 2019-2020 Xilinx, Inc
+ * Author: Larry Liu
```
rozumx: Never been a fan of cut-n-paste code, let alone, cut-n-paste files.

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
AShivangi: I'm not sure about breaking the flow here. Sometimes SC fails to update but the shell does and the card works fine.

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
AShivangi: Hmm. I don't quite understand this.

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
AShivangi: Ok

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
AShivangi: const gives an error in L178

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
```diff
@@ -95,49 +95,120 @@ static void
...
+
+  flasher.upgradeBMCFirmware(bmc.get());
+}
+
+/* 
+ * Find the correct shell to be flashed on the board
+ * Helper method for auto_flash
+ */
+static DSAInfo 
+selectShell(uint16_t idx, std::string& dsa, std::string& id)
```
stsoe: I don't think this function modifies 'dsa' and 'id'?  const?

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
stsoe:
```
if (!x) 
  x=true;
else 
 y=true
```
same as 
```
y=y || x;
x=true
```

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
```diff
@@ -163,17 +234,158 @@ bool canProceed()
...
+  uint16_t success = 0;
+  bool needreboot = false;
+  if (!boardsToUpdate.empty()) {
+
+    // Prompt user about what boards will be updated and ask for permission.
+    if(!force && !canProceed())
+      return;
+
+    // Perform DSA and BMC updating
+    for (auto p : boardsToUpdate) {
```
stsoe: auto&

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
```diff
@@ -163,17 +234,158 @@ bool canProceed()
...
+      std::cout << std::endl;
+      if (updateShellAndSC(p.first, p.second, reboot) == 0) {
+        //std::cout << "Successfully flashed Card[" << getBDF(p.first) << "]"<< std::endl;
+        success++;
+      }
+      needreboot |= reboot;
+    }
+  }
+
+  std::cout << std::endl;
```
stsoe: implies endl not needed anywhere above.  Don't flush unncessarily

src/runtime_src/core/tools/xbmgmt2/flash/xospiversal.cpp
```diff
@@ -0,0 +1,52 @@
...
+    : m_device(xrt_core::get_mgmtpf_device(device_index))
+{
+}
+
+/**
+ * @brief XQSPIPS_Flasher::~XQSPIPS_Flasher
+ */
+XOSPIVER_Flasher::~XOSPIVER_Flasher()
+{
+}
```
stsoe: Remove

src/runtime_src/core/tools/xbmgmt2/flash/xospiversal.cpp
```diff
@@ -0,0 +1,52 @@
...
+{
+    int total_size = 0;
+
+    binStream.seekg(0, binStream.end);
+    total_size = binStream.tellg();
+    binStream.seekg(0, binStream.beg);
+
+    std::cout << "INFO: ***PDI has " << total_size << " bytes" << std::endl;
+
+    std::unique_ptr<char> buffer(new char[total_size]);
```
stsoe: Nope.  This is wrong.   There will be mismatched delete, allocations with [] and deletion without.
It is much much better to use std::vector<char> for char buffers.

src/runtime_src/core/tools/xbmgmt2/flash/xospiversal.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include <iostream>
+#include "core/common/system.h"
+#include "core/common/device.h"
+
+class XOSPIVER_Flasher
+{
+public:
+    XOSPIVER_Flasher(unsigned int device_index);
+    ~XOSPIVER_Flasher();
```
stsoe: remove since its implementation is  empty

src/runtime_src/core/tools/xbmgmt2/flash/xospiversal.cpp
AShivangi: Ok

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
AShivangi: Thanks. That helped :)

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
AShivangi: yep it should be const

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
```diff
@@ -95,49 +95,119 @@ static void
...
 update_SC(uint16_t index, const std::string& file)
 {
-    Flasher flasher(index);
-    if(!flasher.isValid())
-        return;
+  Flasher flasher(index);
+  if(!flasher.isValid())
+    throw xrt_core::error(boost::str(boost::format("%d is an invalid index") % index));
+
+  std::shared_ptr<firmwareImage> bmc =
```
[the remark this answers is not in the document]
AShivangi: ownership isn't shared

AShivangi:
reset this please

src/runtime_src/core/tools/xbmgmt2/flash/xospiversal.cpp
```diff
@@ -0,0 +1,45 @@
...
+    int total_size = 0;
+
+    binStream.seekg(0, binStream.end);
+    total_size = static_cast<int>(binStream.tellg());
+    binStream.seekg(0, binStream.beg);
+
+    std::cout << "INFO: ***PDI has " << total_size << " bytes" << std::endl;
+
+    std::vector<char> buffer;
+    binStream.read(buffer.data(), total_size);
```
stsoe: Allocate the buffer such that's it size is total_size, otherwise buffer.data() is null.
`std::vector<char> buffer(total_size);`

src/runtime_src/core/tools/xbmgmt2/SubCmdFlash.cpp
rozumx: The error on 178 is the result of DSAInfo::matchId() not being const.  
Please update all of the DSAInfo::matchId() methods to be const.

PR #2739, opened 2020-01-30
CR-1052500

xdavidz:
Don't display SC information or update SC flash when golden/recovery image loaded

test results in the PR:
http://jira.xilinx.com/browse/PR-16518


xdavidz:
Based on last Friday's status, DSV team might have trouble to follow the new procedure (aka flash SC separately after a cool reboot). On Monday, the PR meeting will discuss the issue and then we will resume the work from there.

PR #2742, opened 2020-01-30
Full implementation unmanaged Read/Write in Windows stuff

igorshr:
The Windows XoclUser driver provides unmanaged Read/Write to board's DDR. We need to change interface shim.cpp and XoclUser_INTF.h files to communicate to the driver. The input argument structure copied from the Linux source to make cross platform compatible code.
Tested on XRT build.


src/runtime_src/core/pcie/windows/shim.cpp
```diff
@@ -753,16 +753,20 @@ struct shim
...
       }
 
-      XOCL_PWRITE_BO_UNMGD_ARGS pwriteBO;
+      XOCL_PREAD_PWRITE_UNMGD_ARGS pwriteBO;
       DWORD  code;
       DWORD bytesWritten;
 
-      pwriteBO.Offset = offset;
+      pwriteBO.address_space = DDR_BANK0_ADDRESS;
+      pwriteBO.pad = DDR_PAD0;
```
stsoe: Hard code 0?

src/runtime_src/core/pcie/driver/windows/include/XoclUser_INTF.h
```diff
@@ -420,27 +420,31 @@ typedef struct _XOCL_EXECPOLL_ARGS {
...
+typedef struct _XOCL_PREAD_PWRITE_UNMGD_ARGS {
+    uint32_t address_space; //must be 0. We must to keep it to make the structure compatible with Linux code
+    uint32_t pad;           //Currently is unused. We must to keep it to make the structure compatible with Linux code
+    uint64_t paddr;         //Physical address in the specified address space
+    uint64_t size;          //Length of data to write
+    uint64_t data_ptr;      //User's pointer(virtual address) to write the data to
+}XOCL_PREAD_PWRITE_UNMGD_ARGS, *PXOCL_PREAD_PWRITE_UNMGD_ARGS;
 
+#define DDR_BANK0_ADDRESS   0
+#define DDR_PAD0            0
```
stsoe: These two defines are confusing name wise.  Why DDR, why BANK, and PAD
Would it be better to remove them, and simply hard-code the 0 in shim.cpp?

src/runtime_src/core/pcie/windows/shim.cpp
```diff
@@ -782,20 +786,25 @@ struct shim
...
       DWORD  code;
       DWORD bytesRead;
 
       if (flags) {  // make compatible with Linux code
           return false;
       }
 
-      preadBO.Offset = offset;
+      preadBO.address_space = DDR_BANK0_ADDRESS;
+      preadBO.pad = DDR_PAD0;
```
stsoe: Hard code?

src/runtime_src/core/pcie/windows/shim.cpp
igorshr: Yes, it is. It is unused variable now but I was recommended to keep it.

src/runtime_src/core/pcie/driver/windows/include/XoclUser_INTF.h
igorshr: DDR bank and pad are unused now. I was recommended to keep them. I would like to keep them to make the structure identical to the Linux one.

stsoe: I am saying you should get rid of the #defines in this header file.   #defines with relatively generic names are bad and in this case not needed at all.

igorshr: Removed #define in the last commit.

PR #2743, opened 2020-01-30
GStreamer support-1

sarab96:
GStreamer support-1

### DIFF
--- a/src/xma/include/lib/xmahw_lib.h
+++ b/src/xma/include/lib/xmahw_lib.h
@@ -24,6 +24,7 @@
 #include "lib/xmalimits_lib.h"
 #include "app/xmahw.h"
 #include "app/xmaparam.h"
+#include "app/xmabuffers.h"
 #include "plg/xmasess.h"
 #include "xrt.h"
 #include <atomic>
@@ -34,6 +35,8 @@
 #include <array>
 #include <random>
 #include <chrono>
+#include <list>
+#include <queue>
 
 #define MIN_EXECBO_POOL_SIZE      16
 #define MAX_EXECBO_BUFF_SIZE      4096// 4KB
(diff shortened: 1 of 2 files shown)


src/xma/include/lib/xmahw_lib.h
```diff
@@ -95,6 +98,49 @@ typedef struct XmaHwExecBO
   }
 } XmaHwExecBO;
 
+typedef struct XmaBufferPool
+{
+    std::list<XmaBufferObj>   buffers_busy;
+    std::queue<XmaBufferObj>  buffers_free;
```
stsoe: list and queue are generally bad data structures.  Is there really a need to insert / remove from the middle here, and why are free buffers in a queue as opposed to just a vector?

src/xma/include/lib/xmahw_lib.h
```diff
@@ -95,6 +98,49 @@ typedef struct XmaHwExecBO
...
+    std::atomic<bool> pool_locked;
```
stsoe: Why atomic here.  Shouldn't the pool be using a mutex to ensure atomic operations?

src/xma/include/lib/xmahw_lib.h
sarab96: This is for tracking which buffer from pool is still available and which is already assigned for use by plugin. Can not use vector as assigned buffer can be freed out of order.

src/xma/include/lib/xmahw_lib.h
sarab96: I think in this case mutex can be avoided. atomic is only for internal synchronization. Will add mutex if it can not be avoided.

maxzhen:
@stsoe , are you ok w/ this change?

src/xma/include/lib/xmahw_lib.h
stsoe: I can understand the buffers_busy being a list for random removal, but why is buffers_free is queue?

src/xma/include/lib/xmahw_lib.h
stsoe: An atomic that indicates that a structure is locked cannot be more efficient that using a mutex to ensure mutual exclusion.  I am not understanding.

sarab96:
retest this please

PR #2752, opened 2020-02-01
Xrt reboot

saifuddin-xilinx:
- Flash driver is writing to tx FIFO and reading from RX FIFO, but not checking the emptiness of FIFO before reading. Flash returns Slave error if we read from empty FIFO.
-  Earlier, mgmt is in PF1, As PF1 has less capabilities/error-checks, the system is not hanging, it is just returning all F’s.  Now, mgmt is In PF0, As PF0 has more capabilities/error-checks, the System is hanging and rebooting.
- Added a check in flash driver to read FIFO whenever it is not empty which solved the reboot issue.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
@@ -358,9 +358,14 @@ static inline bool flash_has_err(struct xocl_flash *flash)
 static int flash_rx(struct xocl_flash *flash, u8 *buf, size_t len)
 {
 	size_t cnt;
+    u8 c;
 
 	for (cnt = 0; cnt < len; cnt++) {
-		u8 c = flash_read8(flash);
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
```diff
@@ -467,6 +471,9 @@ static int flash_transaction(struct xocl_flash *flash,
 		ret = flash_rx(flash, NULL, len);
```
maxzhen: we might want to ignore the return code when we just want to drain the fifo here.

saifuddin-xilinx: Yes Max, I have tested the flash with export FLASH_VIA_DRIVER=1. And it is working fine.
So you are asking we should only check the return for  "need_output" case. 
I will make the change.

PR #2760, opened 2020-02-04
All ULP level subdevices should be destroyed before hot reset

houlz0507:


src/runtime_src/core/pcie/driver/linux/xocl/xocl_drv.h
```diff
@@ -1006,7 +1006,7 @@ static inline int xocl_clock_ops_level(xdev_handle_t xdev)
 	return -ENODEV;
 }
 #define CLOCK_CB(xdev, idx, cb)						\
-	(CLOCK_DEV(xdev, idx) && CLOCK_OPS(xdev, idx) && CLOCK_OPS(xdev, idx)->cb)
+	(idx >= 0 && CLOCK_DEV(xdev, idx) && CLOCK_OPS(xdev, idx) && CLOCK_OPS(xdev, idx)->cb)
```
xdavidz: What about other multi level subdevs, like the IORES, should we also add this check?

#define IORES_CB(xdev, idx, cb)		\
	(IORES_DEV(xdev, idx) && IORES_OPS(xdev, idx) &&		\		(IORES_DEV(xdev, idx) && IORES_OPS(xdev, idx) &&		\
	IORES_OPS(xdev, idx)->cb)		IORES_OPS(xdev, idx)->cb)

houlz0507: That is because idx should never be invalidate

PR #2765, opened 2020-02-05
Fix crash in HAL Profiling

IshitaGhosh:
Check whether VPDatabase exists before its possible use during function call logging. This fixes crash in Hal Profiling. 
Details of crash : 
During destruction of xocl::platform, an attempt to log FreeBO call was causing crash as VPDatabase had already been destroyed.
The crash can only be reproduced if pcie/linux/shim.cpp is compiled with ENABLE_HAL_PROFILING defined and "hal_profile" is turned on in xrt.ini.


IshitaGhosh:
Build failure seems to be infrastructure issue. Please retest the PR.

stsoe:
retest this please

PR #2768, opened 2020-02-05
CR-1053003 XRT - xbutil clock command fails due to mgmt firewall trip

xdavidz:
clock ulp handling


houlz0507:
I think we should not introduce pr_gate address to clock subdev. ulp gate should be a subdev eventually. Before that, we may keep the pr gate address only in icap.c. So, we may add the check in icap.c, if clock subdev in ULP level, then toggle ulp gate instead of freezing gate

xdavidz:
retest this please

PR #2771, opened 2020-02-05
Add more statistics for guidance

hackwa:
1. Max parallel enqueues for each kernel
2. Is Command Queue OOO
3. PLRAM Size
4. Buffer arguments and their memory resources
5. If trace buffer got full for a device
6. Bit Widths of each memory type on each device
7. Application Run time
8. Time Duration during which buffer transfers were active
9. Time Duration during which read buffer transfers were active
10. Time Duration during which write buffer transfers were active

### DIFF
--- a/src/runtime_src/xdp/profile/plugin/ocl/ocl_profiler.cpp
+++ b/src/runtime_src/xdp/profile/plugin/ocl/ocl_profiler.cpp
@@ -688,12 +688,12 @@ namespace xdp {
             uint64_t numTraceBytes = 0;
             while (!endLog) {
               auto readBytes = readTraceDataFromDDR(dInt, xdevice, info->mTraceVector);
-              endLog = readBytes != mTraceReadBufChunkSz;
+              endLog = readBytes != mTraceReadBufChunkSize;
               profileMgr->logDeviceTrace(device_name, binary_name, type, info->mTraceVector, endLog);
               numTraceBytes += readBytes;
               info->mTraceVector = {};
             }
-            if (numTraceBytes >= mDDRBufferSz) {
+            if (numTraceBytes >= mDDRBufferSize) {
               Plugin->sendMessage(TS2MM_WARN_MSG_BUF_FULL);
               auto& g_map = Plugin->getDeviceTraceBufferFullMap();
               g_map[device_name] = 1;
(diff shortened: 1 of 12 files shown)


src/runtime_src/xdp/profile/plugin/ocl/xocl_profile_cb.cpp
```diff
@@ -178,6 +228,8 @@ cb_action_ndrange(xocl::event* event,cl_int status,const std::string& cu_name, c
     std::string uniqueName = "KERNEL|" + uniqueDeviceName + "|" + xname + "|" + CuInfo + "|";
     std::string traceString = uniqueName + std::to_string(workGroupSize);
     OCLProfiler::Instance()->getPlugin()->setTraceStringForComputeUnit(kname, traceString);
+    // Log Buffers associated with this kernel
+    log_buffer_guidance(event, kernel);
```
pgschuey: Please make sure this is called only once per kernel.

src/runtime_src/xdp/profile/plugin/ocl/xocl_plugin.cpp
```diff
@@ -275,6 +281,31 @@ namespace xdp {
...
+  void XoclPlugin::getMemBitWidthDevices()
+  {
+    for (auto device : mPlatformHandle->get_device_range()) {
+      if (!device->is_active())
+        continue;
+      // TODO: figure out how to get this from platform
+      auto name = device->get_unique_name();
+      mDeviceMemTypeBitWidthMap[name + "|HBM"] = 256;
+      mDeviceMemTypeBitWidthMap[name + "|DDR"] = 512;
+      mDeviceMemTypeBitWidthMap[name + "|PLRAM"] = 512;
```
pgschuey: Potentially in a future checkin, please add support for SoC platforms.

src/runtime_src/xdp/profile/plugin/ocl/ocl_profiler.cpp
```diff
@@ -689,8 +693,11 @@ namespace xdp {
               numTraceBytes += readBytes;
               info->mTraceVector = {};
             }
-            if (numTraceBytes >= mDDRBufferSz)
+            if (numTraceBytes >= mDDRBufferSz) {
```
pgschuey: Please change variable name from mDDRBufferSz to mDDRBufferSize. Easier to read.

src/runtime_src/xdp/profile/plugin/ocl/xocl_profile_cb.cpp
hackwa: line 158 addresses that already

src/runtime_src/xdp/profile/plugin/ocl/xocl_plugin.cpp
hackwa: @IshitaGhosh  is adding a way to distinguish SOC from PCIE. Will add this after her checkin

src/runtime_src/xdp/profile/plugin/ocl/xocl_profile_cb.cpp
pgschuey: @hackwa Please move the checking code to the beginning of the log_buffer_guidance function and add a comment so it's 1. an early exit; and 2. much clearer that the function only operates once per kernel.

hackwa: done

PR #2775, opened 2020-02-05
Separating out XDP kernel debug into its own plugin

jvillarre:


src/runtime_src/CMakeLists.txt
```diff
@@ -30,6 +30,12 @@ add_library(xdp_hal_api_interface_plugin SHARED
 
 add_dependencies(xdp_hal_api_interface_plugin xdp)
 
+add_library(xdp_debug_plugin SHARED
+  $<TARGET_OBJECTS:xdp_debug_plugin_obj>
+  )
+
+add_dependencies(xdp_debug_plugin xrt_core xilinxopencl)
+
```
stsoe: I have long ment to ask if we could possibly move all xdp related link into src/runtime_src/xdp?

jvillarre: I agree that the linking and installation should be done in the xdp directory CMakeLists.txt file.  I will open a separate pull request for that migration.

PR #2776, opened 2020-02-05
support xsabin platformVBNV for 2rp

houlz0507:


houlz0507:
This intends to resolve platformVBNV related CRs:  CR-1053352 and CR-1053806

PR #2783, opened 2020-02-06
turn on ERT MB in U50

mamin506:
Based on the latest spec, the ERT should read from Base Address GPIO to get peripheral based address.

xbutil validate passed with below config,
scheduler config ert(1), dataflow(0), slots(16), cudma(1), cuisr(1), cdma(0), cus(1)


src/runtime_src/ert/scheduler/scheduler.cpp
```diff
@@ -1097,6 +1096,32 @@ scheduler_loop()
...
+    CU_STATUS_REGISTER_ADDR[0] = ERT_CU_STATUS_REGISTER_ADDR0;
+    CU_STATUS_REGISTER_ADDR[1] = ERT_CU_STATUS_REGISTER_ADDR1;
+    CU_STATUS_REGISTER_ADDR[2] = ERT_CU_STATUS_REGISTER_ADDR2;
+    CU_STATUS_REGISTER_ADDR[3] = ERT_CU_STATUS_REGISTER_ADDR3;
+    CQ_STATUS_REGISTER_ADDR[0] = ERT_CQ_STATUS_REGISTER_ADDR0;
+    CQ_STATUS_REGISTER_ADDR[1] = ERT_CQ_STATUS_REGISTER_ADDR1;
+    CQ_STATUS_REGISTER_ADDR[2] = ERT_CQ_STATUS_REGISTER_ADDR2;
+    CQ_STATUS_REGISTER_ADDR[3] = ERT_CQ_STATUS_REGISTER_ADDR3;
+  #endif
+
```
stsoe: Do this in setup() instead, or since setup() is called also via configuration, maybe instead add a setup_ert_base_addr() function with this code so as to call it only once.   My main objection is polluting scheduler_loop() with this setup specific code.

mamin506:
retest this please

src/runtime_src/core/include/ert.h
```diff
@@ -413,6 +413,18 @@ enum softkernel_type {
...
 
+/*
+ * Base address GPIO per spec
+ * | Offset  | Description
+ * -----------------------
+ * | 0x00    | ERT_MGMT_PF_base_addr (Not sure where this should be use)
+ * | 0x08    | ERT_USER_PF_base_addr. The base address of ERT peripherals
+ */
+#if defined(ERT_BUILD_V20)
+uint32_t ert_base_addr;
```
stsoe: initialize to 0

src/runtime_src/ert/scheduler/scheduler.cpp
```diff
@@ -125,39 +125,38 @@ static const u32 AP_CTRL_ME    = 3;
...
 };
 
-const addr_type CQ_STATUS_REGISTER_ADDR[4] =
+addr_type CQ_STATUS_REGISTER_ADDR[4] =
 {
   ERT_CQ_STATUS_REGISTER_ADDR0
  ,ERT_CQ_STATUS_REGISTER_ADDR1
  ,ERT_CQ_STATUS_REGISTER_ADDR2
  ,ERT_CQ_STATUS_REGISTER_ADDR3
 };
```
stsoe: Maybe remove this up front initialization and just declare / define like this:
```
addr_type STATUS_REGISTER_ADDR[4] = {0,0,0,0};
...
addr_type CQ_STATUS_REGISTER[4] = {0,0,0,0};
```
Then unconditionally initialize the values for all platforms in setup_ert_base_addr() (see  comment below) after conditionally reading the ert_base_addr for ERT_BUILD_V20 also in setup_ert_base_addr().

src/runtime_src/core/include/ert.h
mamin506: Compiler would initial global variable. But I would initial it explicitly here.

src/runtime_src/ert/scheduler/scheduler.cpp
mamin506: done.

src/runtime_src/ert/scheduler/scheduler.cpp
mamin506: done.

houlz0507:
retest this please

PR #2787, opened 2020-02-06
PU2 xbuitl CRs

AShivangi:
CR-1053786: xbutil DUMP - Provide Interface UUIDs

CR-1051971: xbutil dma test should only report memory size for the types of memory on the card _(forgot about the case when both HBM ad DDR are present)_


maxzhen:
could  you paste the changed output here?

PR #2790, opened 2020-02-07
cr-1053561 Ensure SC upgrade can not be performed when application is

houlz0507:


src/runtime_src/core/pcie/linux/scan.cpp
```diff
@@ -577,6 +577,34 @@ int pcidev::pci_device::get_partinfo(std::vector<std::string>& info, void *blob)
     return 0;
 }
 
+std::string pcidev::pci_device::sysfs_get_userpf()
```
maxzhen: you should just use get_dev() and pass false to it to get user pf.

houlz0507: Updated the code

PR #2797, opened 2020-02-07
CR-1053003 XRT - xbutil clock command fails after toggling the gate

xdavidz:
test passed on u50 and u200


xdavidz:
> Please consider to implement ULP gate as a subdev later

You've got it. An ULP will make both icap, iores and clock cleaner.

PR #2807, opened 2020-02-10
Initial skeleton of low overhead profiling before optimizations and O…

jvillarre:
…penCL API calls only


stsoe:
retest this please

PR #2811, opened 2020-02-11
plugin version check updated

sarab96:
plugin version check updated

### DIFF
--- a/src/xma/src/xmaapi/xmaadmin.cpp
+++ b/src/xma/src/xmaapi/xmaadmin.cpp
@@ -132,7 +132,9 @@ xma_admin_session_create(XmaAdminProperties *props)
     int32_t xma_main_ver = -1;
     int32_t xma_sub_ver = -1;
     rc = session->admin_plugin->xma_version(&xma_main_ver, & xma_sub_ver);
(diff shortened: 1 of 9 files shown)


src/xma/src/xmaapi/xmaadmin.cpp
```diff
@@ -132,14 +132,22 @@ xma_admin_session_create(XmaAdminProperties *props)
...
         return nullptr;
     }
+    if ((xma_main_ver == XMA_LIB_MAIN_VER && xma_sub_ver > XMA_LIB_SUB_VER) || xma_main_ver > XMA_LIB_MAIN_VER) {
+        xma_logmsg(XMA_ERROR_LOG, XMA_ADMIN_MOD,
+                   "Initalization of plugin failed. Newer plugin is not allowed with old XMA library\n");
+        //Release singleton lock
+        g_xma_singleton->locked = false;
+        free(session);
+        return nullptr;
+    }
```
maxzhen: Seems like duplicated code in multiple places. Could you create a function for them and call it in these places?

sarab96: done

sonals:
Detailed code review will follow.

PR #2815, opened 2020-02-12
Adjust system and device class hierarchy for xrt_core

stsoe:
Mix-in shim level functions to the device class hierarchy so that all xrt.h functions can be accessed from device class.  

Fix how singletons xrt_core::system class object is initialized so as to allow true system/device base class delivered in coreutils library (as opposed to in lib_xrtcore libraries).  

This in turn makes it possible to share xrt_core::device with code that doesn't directly link with libxrt_core; it also eliminates the need for the halops indirection layer as all shim layer functions can now be made virtual.

Much of this is prep work for moving code to coreutil to facilitate more core level functionality.


stsoe:
@rozumx . FYI mostly.  If you want more reason behind changes, please ping me.  No existing functionality has been changed.

src/runtime_src/core/common/ishim.h
```diff
@@ -0,0 +1,116 @@
...
+  template <typename ...Args>
+  shim(Args&&... args)
+    : DeviceType(std::forward<Args>(args)...)
+  {}
+
+  virtual void
+  open_context(xuid_t xclbin , unsigned int ip_index, bool shared)
+  {
+    if (auto ret = xclOpenContext(DeviceType::get_device_handle(), xclbin, ip_index, shared))
+      throw error(ret, "failed to open ip context");
```
rozumx: Learned something new...the error exception :)

src/runtime_src/core/pcie/linux/CMakeLists.txt
```diff
@@ -15,8 +15,8 @@ file(GLOB XRT_PCIE_LINUX_FILES
   "*.cpp"
   "../common/system_pcie.cpp"
   "../common/device_pcie.cpp"
-  "../../common/system.cpp"
-  "../../common/device.cpp"
+#  "../../common/system.cpp"
```
rozumx: Hmm...Is there a reason why these lines are simply not removed entirely instead of being commented out?

src/runtime_src/core/pcie/windows/system_windows.cpp
```diff
@@ -19,7 +19,7 @@
 // file from importing symbols from libxrt_core we define this source
 // file to instead export with same macro as used in libxrt_core.
 #define XCL_DRIVER_DLL_EXPORT
-
+#define XRT_CORE_PCIE_WINDOWS_SOURCE
```
rozumx: Could you remind me the reason we need to define this value in the source code as opposed to having CMake define it.

src/runtime_src/core/common/ishim.h
```diff
@@ -0,0 +1,116 @@
...
+#include <stdexcept>
+
+namespace xrt_core {
+
+/**
+ * struct ishim - Shim API implemented by core libraries
+ *
+ * All methods throw on error
+ */
+struct ishim
```
rozumx: Hmm...since the majority of methods in the ishim and shim class are helper methods and they don't seem to change the class variables, shouldn't they be "const" or would that make them too restrictive?

stsoe:
retest this please

PR #2826, opened 2020-02-13
xbutil CRs

AShivangi:
- CR-1053689: xbutil validate tries P2P for DDR on xilinx_u280_xdma_201920_3
- CR-1054074: Bandwidth test on U25 expected value
- Don't skip bw testcase for red hat unless it is powerpc


maxzhen:
retest this please

PR #2828, opened 2020-02-13
xocl: Supporting passthrough functionality when userPf in VM and mgmt is in Host

rajkumar-xilinx:
Supporting passthrough functionality when userPf in VM and mgmt is in Host

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
@@ -854,9 +854,9 @@ void xclmgmt_mailbox_srv(void *arg, void *data, size_t len,
 			mgmt_info(lro, "got the p2p bar len = %lld\n", p2p_bar_len);
 			if (!p2p_bar_addr) {
 				pci_write_config_byte(pdev, 0x188, 0x0);
-				pci_write_config_byte(pdev, 0x190, 0x0);
-				pci_write_config_byte(pdev, 0x194, 0x0);
-				pci_write_config_byte(pdev, 0x18C, 0x0);
+				pci_write_config_dword(pdev, 0x190, 0x0);
+				pci_write_config_dword(pdev, 0x194, 0x0);
+				pci_write_config_dword(pdev, 0x18C, 0x0);
 				ret = 0;
 				(void) xocl_peer_response(lro, req->req, msgid, &ret,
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
```diff
@@ -848,34 +848,35 @@ void xclmgmt_mailbox_srv(void *arg, void *data, size_t len,
...
-					(void) xocl_peer_response(lro, req->req, msgid, &ret,
+			p2p_bar_addr = mb_p2p->p2p_bar_addr;
+			p2p_bar_len = mb_p2p->p2p_bar_len;
+			mgmt_info(lro, "got the p2p bar addr = %lld\n", p2p_bar_addr);
+			mgmt_info(lro, "got the p2p bar len = %lld\n", p2p_bar_len);
+			if (!p2p_bar_addr) {
+				pci_write_config_byte(pdev, 0x188, 0x0);
+				pci_write_config_dword(pdev, 0x190, 0x0);
+				pci_write_config_dword(pdev, 0x194, 0x0);
+				pci_write_config_dword(pdev, 0x18C, 0x0);
```
maxzhen: can we add some comments and change numbers like 0x188 to some meaningful macro names here? the current code is not readable and will be broken for sure in the future, if we keep it this way.

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
```diff
@@ -397,6 +407,9 @@ int xocl_hot_reset(struct xocl_dev *xdev, u32 flag)
 		xocl_reset_notify(xdev->core.pdev, false);
 
 		xocl_drvinst_set_offline(xdev->core.drm, false);
+
+		//Reset Virtualization registers
+		(void) xocl_mb_read_p2p_addr(xdev);
```
houlz0507: Why does it have to refresh bridge address table after resetting FPGA?

rajkumar-xilinx: Need to configure these PCIe extended registers again

maxzhen: The question is why you need to configure it again? We do not reset the on-device pcie bridge, so why it lost the configuration?

PR #2834, opened 2020-02-13
[New2] xbutil CRs

AShivangi:
CR-1053689: xbutil validate tries P2P for DDR on xilinx_u280_xdma_201920_3
CR-1054074: Bandwidth test on U25 expected value
Don't skip bw testcase for red hat unless it is powerpc

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
+++ b/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
@@ -1181,7 +1181,7 @@ int xcldev::device::bandwidthKernelTest(void)
     std::string output;
 
     if ((sensor_tree::get<std::string>("system.linux", "N/A").find("Red Hat") != std::string::npos)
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -1692,7 +1694,12 @@ int xcldev::device::testP2p()
     }
 
     for (int32_t i = 0; i < map->m_count && ret == 0; i++) {
-        const std::vector<std::string> supList = { "HBM", "DDR", "bank" };
+        std::vector<std::string> supList = { "HBM", "bank" };
+
+        //p2p is not supported for DDR on u280
+        if(vbnv.find("u280") == std::string::npos)
```
maxzhen: you probably want to search for "_u280_" here?

AShivangi: I'm doing if not u280 then add DDR to the "supList"

PR #2836, opened 2020-02-13
Add DEBUG_IP_LAYOUT information

igorshr:
Already implemented in the Windows XoclUser driver. Changes tested by a special xclbin image which has debug ip layout section.


maxzhen:
ok to test

PR #2837, opened 2020-02-14
Convert remaining query requests

stsoe:


stsoe:
Hi @AShivangi . Can you review xmc and flasher changes (minor)?  Thanks.

PR #2814, opened 2020-02-11
Fixed following CRs:

saifuddin-xilinx:
CR-1054109
    CR-1054175

Simplefied the rescan logic. 
Removed the rescan logic to rescan  from individual RootPort 
Now we are rescanning from the /sys/bus/pci/rescan.

PR #2842, opened 2020-02-14
Hotplug issues 

saifuddin-xilinx:
Creating a new pull request from the previous one i.e. https://github.com/Xilinx/XRT/pull/2814

### DIFF
--- a/src/runtime_src/core/pcie/linux/scan.cpp
+++ b/src/runtime_src/core/pcie/linux/scan.cpp
@@ -31,6 +31,7 @@
 #include <sys/stat.h>
 #include <sys/file.h>
 #include <poll.h>
+#include <boost/filesystem.hpp>
 #include "xclbin.h"
 #include "scan.h"
 #include "core/common/utils.h"
@@ -769,8 +770,7 @@ static bool is_in_use(std::vector<std::shared_ptr<pcidev::pci_device>>& vec)
 
 void pci_device_scanner::pci_device_scanner::rescan_nolock()
 {
-    DIR *dir;
-    struct dirent *entry;
+    std::vector<boost::filesystem::path> vec;
 
     if (is_in_use(user_list) || is_in_use(mgmt_list)) {
         std::cout << "Device list is in use, can't rescan" << std::endl;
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/pcie/linux/scan.cpp
```diff
@@ -605,9 +606,43 @@ std::shared_ptr<pcidev::pci_device> pcidev::pci_device::lookup_peer_dev()
...
+    // Determine if device is a xilinx bridge
+    pdev->sysfs_get<uint16_t>("", "vendor", err, vendor,
+            static_cast<uint16_t>(-1));
+    if (!err.empty()) {
+        std::cout << err << std::endl;
+        return NULL;
+    }
+
+    if (vendor != XILINX_ID)
+        return NULL;
```
maxzhen: Why you don't return parent when bridge is not from xilinx? How does it work for non-U.2 boards?

src/runtime_src/core/pcie/linux/scan.cpp
```diff
@@ -605,9 +606,43 @@ std::shared_ptr<pcidev::pci_device> pcidev::pci_device::lookup_peer_dev()
     return udev;
 }
 
+std::shared_ptr<pcidev::pci_device> pcidev::pci_device::lookup_parent_dev()
```
maxzhen: pci_device is only for mgmt or user functions. i don't think we want it to work for pcie bridge.

src/runtime_src/core/pcie/tools/xbmgmt/cmd_hotplug.cpp
```diff
@@ -226,59 +182,60 @@ static boost::filesystem::path findXilinxRootPort(void)
...
+    return rootportList;
 }
 
 static int hotplugRescan(void)
 {
-    boost::filesystem::ofstream ofile;     
-    boost::filesystem::path sysfs_path = findXilinxRootPort();
+    boost::filesystem::ofstream ofile;
+    /* Find all the RootPorts where Xilinx devices are connected */
+    std::list<std::string> rootportList = findXilinxRootPort(SYSFS_PATH);
```
maxzhen: i'm not sure why not just rescan from the highest level? if you remove the board, do you still see xilinx's pcie bridge?

src/runtime_src/core/pcie/linux/scan.h
```diff
@@ -122,6 +122,7 @@ class pci_device {
     int get_partinfo(std::vector<std::string>& info, void *blob = nullptr);
     int shutdown(bool remove_user = false, bool remove_mgmt = false);
     std::shared_ptr<pcidev::pci_device> lookup_peer_dev();
+    std::shared_ptr<pcidev::pci_device> lookup_parent_dev();
```
maxzhen: no need to introduce this member function.

PR #2845, opened 2020-02-14
zocl changes to handle ap_ctrl_chain

Himanshu-xilinx:
* current implementation of zocl doesnot handle ap_ctrl_chain in which xrt need to write on ap_continue in isr else it will generate infinite interrupts

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/sched_exec.c
+++ b/src/runtime_src/core/edge/drm/zocl/sched_exec.c
@@ -26,6 +26,7 @@
 #include "sched_exec.h"
 #include "zocl_sk.h"
 #include "zocl_xclbin.h"
+#include "xclbin.h"
 
 /* #define SCHED_VERBOSE */
 
@@ -528,6 +529,9 @@ static irqreturn_t sched_exec_isr(int irq, void *arg)
 		return IRQ_NONE;
 	}
 
+	if(zocl_cu_get_control(&zdev->exec->zcu[cu_idx]) == AP_CTRL_CHAIN)
+		zocl_cu_check(&zdev->exec->zcu[cu_idx]);
+	
 	/* This function returns the value of the interrupt status register
 	 * No need to check the interrupt type for now.
 	 */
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -801,6 +806,7 @@ configure(struct sched_cmd *cmd)
 				has_acc_cu = 1;
 		}
 
+		control = cfg->data[i] & 0x7; //bit [2:0]
```
larry9523: It is a bit confusing that we are using two different mask here. ~ZOCL_KDE_MASK and 0x7. Can we unify it? Something like
control = control = cfg->data[i] & 0x7; //bit [2:0]
if (zocl_xclbin_accel_adapter(control))
?

src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -528,6 +529,9 @@ static irqreturn_t sched_exec_isr(int irq, void *arg)
 		return IRQ_NONE;
 	}
 
+	if(zocl_cu_get_control(&zdev->exec->zcu[cu_idx]) == AP_CTRL_CHAIN)
```
larry9523: You may want to add a space between if and (. Can you please run XRT/build/checkpatch to make sure we do not introduce any new warning?

src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -528,6 +529,9 @@ static irqreturn_t sched_exec_isr(int irq, void *arg)
...
+		zocl_cu_check(&zdev->exec->zcu[cu_idx]);
```
larry9523: can we add some comment here to explain why we want o call cu_check in interrupt handler?

src/runtime_src/core/edge/drm/zocl/sched_exec.c
Himanshu-xilinx: actually they are 2 different mask, the control information is in lower 3 bits and zocl_kde_mask is 8 bit.

larry9523: I do believe the ZOCL_KDS_MASK actually mean bit [2:0]. If bits [7:3] (max ctx) are ever used, kds_mask == ACCEL_ADAPTER in zocl_xclbin_accel_adapter would be wrong. 
But I am not going to fight for it but just leave a comment. :)

PR #2849, opened 2020-02-17
Basic value based scope_guard for RAII

stsoe:


src/runtime_src/core/common/scope_guard.h
```diff
@@ -0,0 +1,59 @@
...
+  scope_guard(ValueType v, ExitFunction&& exf)
+    : value(v), exfcn(std::move(exf))
+  {}
+
+  ~scope_guard()
+  {
+    try {
+      exfcn(value);
+    }
+    catch (...) {
```
keryell: I wonder whether it would be a debuggability issue here...
It might end up into a trade-off between an immediate`std::terminate` fast or a latent undefined behaviour because the program goes on with some exceptional case hidden under the carpet...

src/runtime_src/core/common/scope_guard.h
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * class scope_guard - RAII for fundamental types that need to be 
+ * terminated / cleaned-up at scope exit
+ */
+template <typename ValueType, typename ExitFunction>
+class scope_guard
+{
+  static_assert(std::is_fundamental<ValueType>::value,"Invalid ValueType");
+  ValueType value;
+  ExitFunction exfcn;
```
keryell: Since `ExitFunction` can be any callable including a lambda, I am unsure we need a `value` at all

stsoe: We need a value for sure. I realize the lambda and that was my first approach.  However, the guard is there to manage lifetime of the 'value' while providing access to the 'value' for clients who need it.  Before the guard, it was

> int value = open(...);
> DoSomething(value);
> close(value);

with the guard it is

> auto v = open_as_a_guard(...);
> DoSomething(v.get());

C++ is in dire need of a generic scope guard.   I do not necessarily like what I cooked up, but I also do not like the unguarded approach.

src/runtime_src/core/common/scope_guard.h
stsoe: Yes, Indeed a problem.   Unfortunately, the nothrow guarantee can be assumed by the exit function.
Do you have a better suggestion?

PR #2855, opened 2020-02-18
Fix for 15_buffer_size to handle HBM devices

haeseunglee:
xclDeviceInfo2 is not having field for HBM device.
So, temporary fix to handle HBM devices. 
Testcase needs update once we have a new data structure.


maxzhen:
ok to test

PR #2861, opened 2020-02-19
add retry to azure wireserver APIs

xuhz:


xuhz:
retest please

xuhz:
retest this please

src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.h
```diff
@@ -91,6 +91,7 @@ class AzureDev
     // 4 MB buffer to truncate and send
     static const int TRANSFER_SEGMENT_SIZE { 1024 * 4096 };
     static const int REIMAGE_TIMEOUT { 20 }; //in second
+    static const int UPLOAD_RETRY { 15 };
```
stsoe: In general it is a bad idea to use upper cased variable names.  Though unlikely in this case, they are subject to perprocessor replacement

xuhz: These variables are declared as static, and act as Macro defines.

PR #2862, opened 2020-02-19
CR-1054083 Error flashing shell from suggested command

AShivangi:
$xbmgmt flash --update

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbmgmt/firmware_image.cpp
+++ b/src/runtime_src/core/pcie/tools/xbmgmt/firmware_image.cpp
@@ -516,16 +516,16 @@ firmwareImage::firmwareImage(const char *file, imageType type) :
                 return;
             }
             // Load entire BMC section.
-            std::shared_ptr<char> bmcbuf(new char[bmcSection->m_sectionSize]);
+            std::vector<char> bmcbuf(bmcSection->m_sectionSize);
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/pcie/tools/xbmgmt/firmware_image.cpp
```diff
@@ -559,6 +560,22 @@ firmwareImage::firmwareImage(const char *file, imageType type) :
                 mBuf = new char[bufsize];
                 in.seekg(flashSection->m_sectionOffset + flashMeta.m_image_offset);
                 in.read(mBuf, bufsize);
+            } 
+            else if (pdiSection) {
+                // Load entire PDI section.
+                std::shared_ptr<char> pdibuf(new char[pdiSection->m_sectionSize]);
```
stsoe: Either use std::vector<char> or std::unique_ptr<char[]>.   This is not shared ownership, plus there would be a mismatched free since new []  requires delete [].  Line 519 has same mistake.

larry9523:
Just one comment that the PDI section in xclbin really means AIE_PARTIAL_PDI. But given the fact that these piece of code will only be used to parse dsabin and xsabin, I am fine with it for now. Later, we should give a more meaning for name for PDI section for flashing.

salindac:
retest this please

AShivangi:
retest this please

PR #2877, opened 2020-02-20
add flash_type sysfs node for versal

larry9523:
To use "xbmgmt flash", ospi_versal sysfs node needs to be populated.


larry9523:
retest this please

PR #2883, opened 2020-02-21
cpp changes. malloc removed

sarab96:
cpp changes. malloc removed

### DIFF
--- a/src/xma/include/lib/xmaxclbin.h
+++ b/src/xma/include/lib/xmaxclbin.h
@@ -74,7 +74,7 @@ typedef struct XmaXclbinInfo
     uint32_t    reserved[32];
 } XmaXclbinInfo;
 
-std::vector<char> xma_xclbin_file_open(const char *xclbin_name);
+std::vector<char> xma_xclbin_file_open(const std::string& xclbin_name);
 int xma_xclbin_info_get(char *buffer, XmaXclbinInfo *info);
 int xma_xclbin_map2ddr(uint64_t bit_map, int* ddr_bank);
 #endif
--- a/src/xma/include/lib/xmaxclbin.h
+++ b/src/xma/include/lib/xmaxclbin.h
@@ -74,7 +74,7 @@ typedef struct XmaXclbinInfo
     uint32_t    reserved[32];
 } XmaXclbinInfo;
 
-char *xma_xclbin_file_open(const char *xclbin_name);
+std::vector<char> xma_xclbin_file_open(const char *xclbin_name);
 int xma_xclbin_info_get(char *buffer, XmaXclbinInfo *info);
 int xma_xclbin_map2ddr(uint64_t bit_map, int* ddr_bank);
 #endif
--- a/src/xma/src/xmaapi/xmahw_hal.cpp
+++ b/src/xma/src/xmaapi/xmahw_hal.cpp
@@ -193,7 +193,7 @@ bool hal_configure(XmaHwCfg *hwcfg, XmaXclbinParameter *devXclbins, int32_t num_
             return false;
         }
         std::string xclbin = std::string(devXclbins[i].xclbin_name);
-        std::vector<char> xclbin_buffer = xma_xclbin_file_open(xclbin.c_str());
+        std::vector<char> xclbin_buffer = xma_xclbin_file_open(xclbin);
         char *buffer = xclbin_buffer.data();
 
         int32_t rc = xma_xclbin_info_get(buffer, &info);
@@ -201,7 +201,6 @@ bool hal_configure(XmaHwCfg *hwcfg, XmaXclbinParameter *devXclbins, int32_t num_
         {
             xma_logmsg(XMA_ERROR_LOG, XMAAPI_MOD, "Could not get info for xclbin file %s\n",
                        xclbin.c_str());
-            free(buffer);
             return false;
         }
 
@@ -213,7 +212,6 @@ bool hal_configure(XmaHwCfg *hwcfg, XmaXclbinParameter *devXclbins, int32_t num_
         dev_tmp1.handle = xclOpen(dev_index, NULL, XCL_QUIET);
         if (dev_tmp1.handle == NULL){
             xma_logmsg(XMA_ERROR_LOG, XMAAPI_MOD, "Unable to open device  id: %d\n", dev_index);
-            free(buffer);
             return false;
         }
         dev_tmp1.dev_index = dev_index;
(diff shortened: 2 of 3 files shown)


src/xma/include/lib/xmaxclbin.h
```diff
@@ -74,7 +74,7 @@ typedef struct XmaXclbinInfo
     uint32_t    reserved[32];
 } XmaXclbinInfo;
 
-char *xma_xclbin_file_open(const char *xclbin_name);
+std::vector<char> xma_xclbin_file_open(const char *xclbin_name);
 int xma_xclbin_info_get(char *buffer, XmaXclbinInfo *info);
```
stsoe: 'const char* buffer' to make it clear that this function is not changing the buffer content.

src/xma/src/xmaapi/xmahw_hal.cpp
```diff
@@ -193,13 +193,9 @@ bool hal_configure(XmaHwCfg *hwcfg, XmaXclbinParameter *devXclbins, int32_t num_
...
         }
         std::string xclbin = std::string(devXclbins[i].xclbin_name);
-        char *buffer = xma_xclbin_file_open(xclbin.c_str());
-        if (!buffer)
-        {
-            xma_logmsg(XMA_ERROR_LOG, XMAAPI_MOD, "Could not open xclbin file %s\n",
-                       xclbin.c_str());
-            return false;
-        }
+        std::vector<char> xclbin_buffer = xma_xclbin_file_open(xclbin.c_str());
```
stsoe: Since xclbin is already a std::string, I recommend staying with std::string until the very lowest level when/and-only-if a const char* is required.  Alas, you should change xma_xclbin_file_open to take 'const std::string& xclbin_name'.

src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -35,27 +36,32 @@ static int get_xclbin_iplayout(char *buffer, XmaXclbinInfo *xclbin_info);
 static int get_xclbin_mem_topology(char *buffer, XmaXclbinInfo *xclbin_info);
 static int get_xclbin_connectivity(char *buffer, XmaXclbinInfo *xclbin_info);
```
stsoe: const char* buffer

src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -35,27 +36,32 @@ static int get_xclbin_iplayout(char *buffer, XmaXclbinInfo *xclbin_info);
...
 
-char *xma_xclbin_file_open(const char *xclbin_name)
+std::vector<char> xma_xclbin_file_open(const char *xclbin_name)
```
stsoe: const std::string& xclbin_name

src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -35,27 +36,32 @@ static int get_xclbin_iplayout(char *buffer, XmaXclbinInfo *xclbin_info);
...
-    std::ifstream file(xclbin_name, std::ios::binary | std::ios::ate);
-    std::streamsize size = file.tellg();
-    file.seekg(0, std::ios::beg);
+    std::ifstream infile(xclbin_name, std::ios::binary | std::ios::ate);
+    std::streamsize xclbin_size = infile.tellg();
+    infile.seekg(0, std::ios::beg);
 
-    char *buffer = (char*)malloc(size);
-    if (buffer == NULL) {
+    std::vector<char> xclbin_buffer;
```
stsoe: std::vector<char> xclbin_buffer(size), no need for separate reserve

src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -35,27 +36,32 @@ static int get_xclbin_iplayout(char *buffer, XmaXclbinInfo *xclbin_info);
...
+    try {
+        xclbin_buffer.reserve(xclbin_size);
+    } catch (const std::bad_alloc& ex) {
         xma_logmsg(XMA_ERROR_LOG, XMAAPI_MOD, "Could not allocate buffer for file %s\n", xclbin_name);
-        return NULL;
+        xma_logmsg(XMA_ERROR_LOG, XMAAPI_MOD, "Buffer allocation error: %s\n", ex.what());
+        throw;
+    } catch (...) {
+        xma_logmsg(XMA_ERROR_LOG, XMAAPI_MOD, "Could not allocate buffer for xclbin file %s\n", xclbin_name);
+        throw;
```
stsoe: drop the try/catch, just let the exception peculate up the stack; these log messages are not very interesting and only complicates the code

src/xma/include/lib/xmaxclbin.h
sarab96: this change is coming in separate pr for these other functions

src/xma/src/xmaapi/xmahw_hal.cpp
sarab96: done

src/xma/src/xmaapi/xmaxclbin.cpp
sarab96: this change is coming in separate pr for these other functions

src/xma/src/xmaapi/xmaxclbin.cpp
sarab96: done

src/xma/src/xmaapi/xmaxclbin.cpp
sarab96: keeping it to capture alloc exception error in log.

src/xma/src/xmaapi/xmaxclbin.cpp
sarab96: keeping it to capture alloc exception error in log. XMA tries to capture all errors in log. XRT verbosity is used by users to control what they want to see.

src/xma/src/xmaapi/xmaxclbin.cpp
stsoe: std::vector xclbin_buffer(size) will throw std::bad_alloc.  You are not achieving anything with separate call to reserve.

src/xma/src/xmaapi/xmaxclbin.cpp
stsoe: You are rethrowing, so what is the caller doing with the exception?  Also catch (...) is not needed since std::vector throws std::exception only.

src/xma/src/xmaapi/xmaxclbin.cpp
sarab96: below that catching bad_alloc exception with reserve for the log msg

src/xma/src/xmaapi/xmaxclbin.cpp
sarab96: there can be length_error as well so good to have catch(...). It is a fatal error so execution can stop. xma log does need log msg to say that loading of xclbin lead to error

PR #2884, opened 2020-02-21
qdma xocl: support multiple dma requests per aio call

karenx-xilinx:
xocl driver qdma st changes:
- add per libqdma submission control struct, per dma request control struct.
- removed the request cache as the # of request and control struct are dynamic.


karenx-xilinx:
made a mistake in porting

PR #2886, opened 2020-02-21
memory corruption

houlz0507:


houlz0507:
because pipeline is blocked. so merge it directly to unblock pipeline.

PR #2887, opened 2020-02-21
qdma: xocl driver add support of multiple dma requests per aio call

karenx-xilinx:
qdma xocl driver:
- add per libqdma submission control struct and per dma request control struct
- remove the req_cache as the # request is different.


xdavidz:
retest this please

PR #2895, opened 2020-02-24
Update u50 platform exclusion info

haeseunglee:
Update u50 platform exclusion info to reflect updated u50 platform name


maxzhen:
ok to test

maxzhen:
@haeseunglee we're done with PU2 branch. Do you really need these changes?

haeseung81:
@maxzhen Yeah. The updated u50 platform name should be reflected. Otherwise, the sprite keep picking these tests and reports failure.

PR #2903, opened 2020-02-25
Consolidate xrt::time with xrt_core::time

stsoe:


IshitaGhosh:
Changes look good. This change would help eliminate different time "zero" for xrt_core::time_ns and xrt::time_ns and eventually the need for statically linking xrt_core library into xdp.

PR #2907, opened 2020-02-25
add option allowing users to use cmake toolchain

yunqu:
Removed pynqlinux from native build, since pynq is going through the non-native build for now. In the future we probably need better ways to decide which kind of platform users are building upon.

Add extra toolchain option, so users can provide an external toolchain file (toolchain.cmake). This option, if ignored (by default), will not take any effect.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -42,6 +42,7 @@ usage()
     echo "[-nocmake]                 Skip CMake call"
     echo "[-j <n>]                   Compile parallel (default: system cores)"
     echo "[-ccache]                  Build using RDI's compile cache"
+    echo "[-toolchain <file>]        Extra toolchain file to configure CMake"
     echo "[-driver]                  Include building driver code"
     echo "[-checkpatch]              Run checkpatch.pl on driver code"
     echo "[-verbose]                 Turn on verbosity when compiling"
@@ -69,6 +70,7 @@ dbg=1
 edge=0
 nocmake=0
 ertfw=""
+extratoolchain=0
 while [ $# -gt 0 ]; do
     case "$1" in
         -help)
@@ -110,6 +112,12 @@ while [ $# -gt 0 ]; do
             ;;
         -ccache)
             ccache=1
+            shift
+            ;;
+        -toolchain)
+            extratoolchain=1
+            shift
+            toolchain=$1
             shift
             ;;
         -checkpatch)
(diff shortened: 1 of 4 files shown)


build/build.sh
```diff
@@ -180,8 +188,27 @@ if [[ $dbg == 1 ]]; then
...
+		echo "	-DCMAKE_BUILD_TYPE=Debug"
+		echo "	-DCMAKE_EXPORT_COMPILE_COMMANDS=ON"
+		echo "	-DCMAKE_TOOLCHAIN_FILE=$toolchain"
+		echo "	../../src"
+		time $CMAKE -DRDI_CCACHE=$ccache \
+			-DCMAKE_BUILD_TYPE=Debug \
+			-DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
+			-DCMAKE_TOOLCHAIN_FILE=$toolchain \
+			../../src
+	fi
```
stsoe: I believe you can simply always pass -DCMAKE_TOOLCHAIN_FILE=$toolchain without checking if the variable was specified.   If it is empty -DCMAKE_TOOLCHAIN_FILE will simply be ignored.  Please check.

I'd prefer to have this be one line commands as before.

build/build.sh
```diff
@@ -193,8 +220,27 @@ if [[ $opt == 1 ]]; then
...
+		echo "	-DCMAKE_BUILD_TYPE=Release"
+		echo "	-DCMAKE_EXPORT_COMPILE_COMMANDS=ON"
+		echo "	-DCMAKE_TOOLCHAIN_FILE=$toolchain"
+		echo "../../src"
+        time $CMAKE -DRDI_CCACHE=$ccache \
+			-DCMAKE_BUILD_TYPE=Release \
+			-DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
+			-DCMAKE_TOOLCHAIN_FILE=$toolchain \
+			../../src
+	fi
```
stsoe: Same as earlier comment

build/build.sh
yunqu: @stsoe The purpose of using multiple lines is to make the code rendering on github as well as vim editors much better - and if you have a very long single line, then even small changes would be very difficult to track.

stsoe: @yunqu  I hear you, but in this case I thought it made it less readable.  Thanks for fixing, it looks good now (at least to me) :-)

PR #2914, opened 2020-02-26
Updated Mapping of Alveo BMC for both Gen4 and Gen5

rozumx:
Issue: The mapping for both Gen4 and Gen5 FW images were missing in the packaging scripts preventing deployment platforms that use these images from being created.


NaAl-Xlnx:
retest this please

PR #2920, opened 2020-02-26
xclbin parser soft kernel changes

sarab96:
xclbin parser soft kernel changes

### DIFF
--- a/src/runtime_src/core/common/scheduler.cpp
+++ b/src/runtime_src/core/common/scheduler.cpp
@@ -184,8 +184,12 @@ init(xclDeviceHandle handle, const axlf* top)
       scmd->count = sizeof (ert_configure_sk_cmd) / 4 - 1;
       scmd->start_cuidx = start_cuidx;
       scmd->num_cus = sk.ninst;
-      std::strncpy(reinterpret_cast<char*>(scmd->sk_name), sk.symbol_name, 31);
-      reinterpret_cast<char*>(scmd->sk_name)[31] = 0;
+      
+      //std::strncpy(reinterpret_cast<char*>(scmd->sk_name), sk.symbol_name, 31);
+      memset(reinterpret_cast<char*>(scmd->sk_name), 0, 32);
+      sk.symbol_name.copy(reinterpret_cast<char*>(scmd->sk_name), 31);
+      //reinterpret_cast<char*>(scmd->sk_name)[31] = 0;
+
       scmd->sk_addr = skbo->prop.paddr;
       scmd->sk_size = skbo->prop.size;
       std::memcpy(skbo->data, sk.sk_buf, sk.size);
--- a/src/runtime_src/core/common/scheduler.cpp
+++ b/src/runtime_src/core/common/scheduler.cpp
@@ -184,12 +184,7 @@ init(xclDeviceHandle handle, const axlf* top)
       scmd->count = sizeof (ert_configure_sk_cmd) / 4 - 1;
       scmd->start_cuidx = start_cuidx;
       scmd->num_cus = sk.ninst;
-      
-      //std::strncpy(reinterpret_cast<char*>(scmd->sk_name), sk.symbol_name, 31);
-      memset(reinterpret_cast<char*>(scmd->sk_name), 0, 32);
       sk.symbol_name.copy(reinterpret_cast<char*>(scmd->sk_name), 31);
-      //reinterpret_cast<char*>(scmd->sk_name)[31] = 0;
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/common/scheduler.cpp
```diff
@@ -184,8 +184,12 @@ init(xclDeviceHandle handle, const axlf* top)
       scmd->count = sizeof (ert_configure_sk_cmd) / 4 - 1;
       scmd->start_cuidx = start_cuidx;
       scmd->num_cus = sk.ninst;
-      std::strncpy(reinterpret_cast<char*>(scmd->sk_name), sk.symbol_name, 31);
-      reinterpret_cast<char*>(scmd->sk_name)[31] = 0;
+      
+      //std::strncpy(reinterpret_cast<char*>(scmd->sk_name), sk.symbol_name, 31);
+      memset(reinterpret_cast<char*>(scmd->sk_name), 0, 32);
+      sk.symbol_name.copy(reinterpret_cast<char*>(scmd->sk_name), 31);
+      //reinterpret_cast<char*>(scmd->sk_name)[31] = 0;
```
stsoe: Not understanding this?  Why do we need to write to sk_name twice?
Please always remove dead code, e.g. remove the commented out lines, the git history has it.

src/runtime_src/core/common/xclbin_parser.cpp
```diff
@@ -393,13 +393,19 @@ get_softkernels(const axlf* top)
       auto begin = reinterpret_cast<const char*>(top) + pSection->m_sectionOffset;
       auto soft = reinterpret_cast<const soft_kernel*>(begin);
 
-      softkernel_object sko;
+      sks.emplace_back(softkernel_object{});
+      auto sko = sks.back();
```
stsoe: This is a bug.  you will be modifying  an in scope copy of the sks.back().  You probably mean auto&.
Please create sko by value as before, then at end use sks.emplace_back(std::move(sko))

src/runtime_src/core/common/xclbin_parser.cpp
```diff
@@ -393,13 +393,19 @@ get_softkernels(const axlf* top)
...
+      //softkernel_object sko;
       sko.ninst = soft->m_num_instances;
-      sko.symbol_name = const_cast<char*>(begin + soft->mpo_symbol_name);
+      
+      //sko.symbol_name = const_cast<char*>(begin + soft->mpo_symbol_name);
+      sko.symbol_name = std::string(begin + soft->mpo_symbol_name);
+      sko.mpo_name = std::string(begin + soft->mpo_name);
+      sko.mpo_version = std::string(begin + soft->mpo_version);
```
stsoe: This sort of copying worries me, what guarantee is there that mpo_symbol_name, mpo_name, mpo_version are null terminated?   Yes, same problem was there before, but can we fix this?

src/runtime_src/core/common/scheduler.cpp
sarab96: removed. memset 0 of full data structure scmd occurred earlier

src/runtime_src/core/common/xclbin_parser.cpp
sarab96: there is no size limit on these soft kernel strings. What do you suggest?

src/runtime_src/core/common/xclbin_parser.cpp
```diff
@@ -393,13 +393,14 @@ get_softkernels(const axlf* top)
       auto begin = reinterpret_cast<const char*>(top) + pSection->m_sectionOffset;
       auto soft = reinterpret_cast<const soft_kernel*>(begin);
 
-      softkernel_object sko;
+      sks.emplace_back(softkernel_object{});
+      auto& sko = sks.back();
```
stsoe: This is not exception safe.  Please don't modify sks before the sko object has been created *and* populated successfully.
Move the emplace to the end as I suggested before.

src/runtime_src/core/common/xclbin_parser.cpp
stsoe: The code is wrong then (unless I am missing something).   I just looked at xclbin.h, all the _name vars are numbers (uint32_t).  You can't construct a std::string from this using std::string(begin + number) because this string constructor interprets 'begin + number' as the address of a null terminated char* string, which it clearly is not.   The old code was okay, because it just reinterpreted the number as a char*.

Has these code changes been tested?

PR #2921, opened 2020-02-26
XRT-244 Build scripts for Edge platform

Himanshu-xilinx:
1. script to build xrt rpms for edge platform
2. usage:
              ./build_edge.sh  -aarch aarch32
              ./build_edge.sh  -aarch aarch64 
              ./build_edge.sh  -clean
              ./build_edge.sh  -aarch aarch32 -cache <path to sstate-cache>
              ./build_edge.sh  -aarch aarch64 -cache <path to sstate-cache>
3. script also generate rpm.txt contain list of rpm used for sysroot overlay
4. script also generate install_edge.sh and reinstall_edge.sh which is used to install rpms through dnf

### DIFF
--- a/build/build_edge.sh
+++ b/build/build_edge.sh
@@ -1,0 +1,182 @@
+#!/bin/bash
+
+error()
+{
+    echo "ERROR: $1" 1>&2
+    usage_and_exit 1
+}
+
+usage()
+{
+    echo "Usage: $PROGRAM [options] "
+    echo "  options:"
+    echo "          -help                           Print this usage"
+    echo "          -aarch                          Architecture <aarch32/aarch64>"
+    echo "          -clean, clean                   Remove build directories"
+    echo ""
+}
+
+usage_and_exit()
+{
+    usage
+    exit $1
+}
+
+# --- Internal funtions ---
+install_recipes()
+{
+    META_USER_PATH=$1
+
+    SAVED_OPTIONS_LOCAL=$(set +o)
+    set +e
+    mkdir -p ${META_USER_PATH}/recipes-xrt/xrt
+    mkdir -p ${META_USER_PATH}/recipes-xrt/zocl
+    XRT_BB=${META_USER_PATH}/recipes-xrt/xrt/xrt_git.bbappend
+    ZOCL_BB=${META_USER_PATH}/recipes-xrt/zocl/zocl_git.bbappend
+    grep "inherit externalsrc" $XRT_BB
+    if [ $? != 0 ]; then
+        echo "inherit externalsrc" > $XRT_BB
+        echo "EXTERNALSRC = \"$XRT_REPO_DIR/src\"" >> $XRT_BB
+        echo 'EXTERNALSRC_BUILD = "${WORKDIR}/build"' >> $XRT_BB
+        echo 'PACKAGE_CLASSES = "package_rpm"' >> $XRT_BB
+    fi
+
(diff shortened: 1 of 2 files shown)


build/build_edge.sh
```diff
@@ -0,0 +1,182 @@
...
+if [[ $clean == 1 ]]; then
+    echo $PWD
+    echo "/bin/rm -rf $aarch64_dir $aarch32_dir"
+    /bin/rm -rf $aarch64_dir $aarch32_dir
+    exit 0
+fi
+
+# we pick Petalinux BSP
+if [[ -z ${PETALINUX:+x} ]]; then
+    export PETALINUX="/proj/petalinux/2020.1/petalinux-v2020.1_0217_1/tool/petalinux-v2020.1-final"
```
mamin506: As @chvamshi-xilinx 's comment. Never use hard code path.

chvamshi-xilinx:
> 
> 
> Since we moved all edge platforms source code to another repo and removed scripts. Why we need this script in XRT repo not platform repo?
> 
> I would prefer to do this in platform repo, since:
> In the platform repo, it only needs to add one more step after 'petalinux-build' finished.
> Instead, you do two major things here, petalinux-build and got rpm packages.
> 
> Do we have a good reason to do extra petalinux build?

Hi @mamin506 , This is for component level verification of XRT. Platform repo is already building platforms with known good XRT commit id. They are not going to build platform with latest commit-id as they are not sure whether that works or not. Currently, there is no mechanism in XRT to verify the commits for edge platforms.  
There are mainly 3 components now. Petalinux , Hw platform and XRT. Each component should add their own verification mechanism so that no bad check-in goes thru.
We are working on XRT component verification where we are taking known good BSP and known good Petalinux.

build/build_edge.sh
mamin506: Please provide a new commit to fix this.

Himanshu-xilinx: as discussed with vamshi, added petalinux.build to control petalinux tool

build/petalinux.build
```diff
@@ -0,0 +1,3 @@
+export PETALINUX="/proj/petalinux/2020.1/petalinux-v2020.1_0227_1/tool/petalinux-v2020.1-final"
+export PETALINUX_VER=2020.1
```
mamin506: 1. There is still some concerns about if we should have internal path in github.
2. This path points to a specific petalinux. How about if we want to use another build?
3. What is the purpose of this file? Should we let the script user setup the env? In this case, we only check env in build_edge.sh.

build/build_edge.sh
mamin506: In my mind petalinux.build file is a environment configuration for the script. Instead of check it in to XRT repo, we should ask pipeline guys, who setup the environment, to provide that file. Does this make sense to you?

Himanshu-xilinx: i completely agree with you, but Devops is not ready to set the environment.

mamin506: If this is the case, looks like petalinux.build is a workaround for Devops. Please add this to the workaround list and create a SR to Devops. Once they are ready to do correct thing, we will remove petalinux.build.

anupcs-xlnx: Please also add that workaround to our Jira epic tracking these. Please contact me offline if you can't find the epic.

abalasa:
retest this please

PR #2922, opened 2020-02-26
xbmgmt2 options

AShivangi:
Sample outputs:
```
sagarw@xsjsagarwal:/proj/xsjhdstaff1/sagarw/XRT/build$ Debug/opt/xilinx/xrt/bin/xbmgmt2 program --revert-to-golden       
INFO     : Resetting card [ 0000:b3:00.0 ] back to factory mode.
INFO     : Resetting card [ 0000:65:00.0 ] back to factory mode.
Are you sure you wish to proceed? [Y/n]:
INFO     : Shell on [ 0000:b3:00.0 ] is reset successfully.
INFO     : Shell on [ 0000:65:00.0 ] is reset successfully.
Cold reboot machine to load the new image on card(s).


root@xsjsagarwal:/proj/xsjhdstaff1/sagarw/XRT/build# xbmgmt2 status --report --device=0000:b3:00.0
Device BDF : 0000:b3:00.0
  Card type            : u250
  Flash type           : SPI
Flashable partition running on FPGA
  Platform             : xilinx_u250_xdma_201830_1
  SC Version           : UNKNOWN
  Platform ID          : 0x5bf297b0
 
  Platform             : xilinx_u250_xdma_201830_1
  SC Version           : 3.1
  Platform ID          : 0x5bf297b0
WARNING  : 0000:b3:00.0 is not up-to-date.

```


src/runtime_src/core/tools/xbmgmt2/SubCmdStatus.cpp
```diff
@@ -21,11 +21,18 @@
...
+#include "core/common/query_requests.h"
+#include "core/common/utils.h"
+#include "core/common/message.h"
 
 #include "tools/common/XBUtilities.h"
 namespace XBU = XBUtilities;
 
 // 3rd Party Library - Include Files
+#include "boost/format.hpp"
+#include "boost/tokenizer.hpp"
```
stsoe: use angle brackets for system and boost includes

src/runtime_src/core/tools/xbmgmt2/SubCmdStatus.cpp
```diff
@@ -109,5 +134,84 @@ SubCmdStatus::execute(const SubCmdOptions& _options) const
...
+    //get all devices
+    auto total = xrt_core::get_total_devices(false).first;
+    if (total == 0)
+      throw xrt_core::error("No card found!");
+    //better way to do this?
+    for(uint16_t i = 0; i < total; i++) {
+      device_indices.push_back(i);
+    }
+  }
+
```
stsoe: I think I just saw this code earlier .  Can we avoid code duplication?

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -58,11 +57,41 @@ namespace po = boost::program_options;
...
 namespace {
 /*
- * Update shell on the board
+ * Update shell on the board for auto flash
+ */
+static void 
+update_shell(uint16_t index, const std::string& primary, const std::string& secondary)
+{
+  std::unique_ptr<firmwareImage> pri;
+  std::unique_ptr<firmwareImage> sec;
```
stsoe: Better to move declarations with definitions or closer to where they are used.
In this case line 65 can be removed and line 75 can be prefixed with auto.
Line 66 can move below line 75.
This makes it clearer where the variables are used.
Same comments for other parts of code.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -480,7 +507,7 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
 
   // -- Retrieve and parse the subcommand options -----------------------------
   std::string device = "";
```
stsoe: Not specific to this line, but this function is much too long to comprehend.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -480,7 +507,7 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
...
-  std::vector<uint16_t> device_indices = {0};
+  std::vector<uint16_t> device_indices; //instead of saving this, can we save Flasher objects?
```
stsoe: Move closer to usage; I think some 100 lines below.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
AShivangi: I agree. Do you have a way to break down this method? This function is retrieving and parsing the arguments and doing some basic checks before calling the functionality (which is a different function).

PR #2924, opened 2020-02-27
xclbin parser soft kernel changes

sarab96:
xclbin parser soft kernel changes
Tested on local machine

### DIFF
--- a/src/runtime_src/core/common/scheduler.cpp
+++ b/src/runtime_src/core/common/scheduler.cpp
@@ -184,8 +184,12 @@ init(xclDeviceHandle handle, const axlf* top)
       scmd->count = sizeof (ert_configure_sk_cmd) / 4 - 1;
       scmd->start_cuidx = start_cuidx;
       scmd->num_cus = sk.ninst;
-      std::strncpy(reinterpret_cast<char*>(scmd->sk_name), sk.symbol_name, 31);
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/common/scheduler.cpp
```diff
@@ -184,8 +184,7 @@ init(xclDeviceHandle handle, const axlf* top)
       scmd->count = sizeof (ert_configure_sk_cmd) / 4 - 1;
       scmd->start_cuidx = start_cuidx;
       scmd->num_cus = sk.ninst;
-      std::strncpy(reinterpret_cast<char*>(scmd->sk_name), sk.symbol_name, 31);
-      reinterpret_cast<char*>(scmd->sk_name)[31] = 0;
+      sk.symbol_name.copy(reinterpret_cast<char*>(scmd->sk_name), 31);
```
stsoe: is null termination of destination string guaranteed?
I think Coverity will complain here.

sarab96: earlier clear of scmd by "std::memset(scmd, 0, 0x1000);" ensures that this string will be null terminated. copy is only for max of 31 characters. So 32nd char will be always null. I will check coverity output to see what it says.

stsoe: Yep. I missed the memset on line 180.

PR #2928, opened 2020-02-27
per device idx for cdev

chienwei-lan:
Need a per device enumeration index for cdev.

before this change which will break mpd and msd
mailbox.m45824.0
mailbox.m53414.1
mailbox.m61344.2

After and expect 
mailbox.m45824.0
mailbox.m53414.0
mailbox.m61344.0


src/runtime_src/core/pcie/driver/linux/xocl/xocl_subdev.c
```diff
@@ -96,11 +96,14 @@ static struct xocl_subdev *xocl_subdev_reserve(xdev_handle_t xdev_hdl,
 				devid, sdev_info->override_idx);
 			return NULL;
 		}
+		subdev->info.dev_idx = sdev_info->override_idx;
```
houlz0507: This line should not be required.

PR #2942, opened 2020-02-28
Refactoring xrt build information out of get_xrt_info

jvillarre:
This allows XDP to get the information it needs before a singleton is created


jvillarre:
It was not my intention to allow calling directly from the shim, so I centralized the call in one place.

PR #2947, opened 2020-02-29
Get debug_ip_layout in XDP Device abstraction using Windows SHIM API

IshitaGhosh:
Enable processing of debug_ip_layout in XDP on Windows.
However, due to issues in xclRead/xclWrite for Profile Monitors on Windows, device profiling is not yet enabled.


src/runtime_src/xdp/profile/device/device_intf.cpp
```diff
@@ -419,9 +420,21 @@ DeviceIntf::~DeviceIntf()
...
+
     debug_ip_layout *map;
     if (ifs.gcount() > 0) {
       map = (debug_ip_layout*)(buffer);
+#else
+    size_t sz1 = 0, sectionSz = 0;
+    // Get the size of full debug_ip_layout
+    mDevice->getDebugIpLayout(nullptr, sz1, &sectionSz);
+    // Allocate buffer to retrieve debug_ip_layout information from loaded xclbin
+    char* buffer = new char[sectionSz];
```
stsoe:
```
std::vector<char> buffer(sectionSz);
mDevice->getDeviceIPLayout(buffer.data(), sectionSz, &sz1);
auto map = reinterpret_cast<const debug_ip_layout*>(buffer.data());
```

src/runtime_src/xdp/profile/device/device_intf.cpp
```diff
@@ -443,9 +456,14 @@ DeviceIntf::~DeviceIntf()
...
       }
      }
+#ifndef _WIN32
     }
-
     ifs.close();
+#else
+    delete [] buffer;
+    buffer = nullptr;
+    map = nullptr;
```
stsoe: remove

IshitaGhosh:
I agree query request on the new device infrastructure would have been better. As that support is being developed and none of the debug/profile related APIs are in there, I went ahead with the old style. We can update it later.

src/runtime_src/xdp/profile/device/device_intf.cpp
IshitaGhosh: I have added the change. Thanks for the suggestion!

PR #2948, opened 2020-03-01
Declared missing "padding" in the mem_data struct.

rozumx:
In the mem_data struct, the compiler is performing 8 byte aligning of uint64_t types.  To assure "binary" compatibility between various compilers, instead of relying of the compiler to insert the padding bytes, the structure was updated to reflect them. 

Because this structure is used to communicate binary information among a wide range of OSs and platforms, additional testing was was and added to the xclbin.h data structure, which included:
1) Adding a static_assert that validated the size of the mem_data structure.
Note: This static_assert was first added and validated, then the padding was added with no compiler issues.
2) Examine the binary content of the xclbin images produced.


rozumx:
retest this please

PR #2954, opened 2020-03-02
Enabled compile time assertions for all expected xclbin.h structure sizes

rozumx:
Work Done
---------
+ Created a compilier independent assertion macro.  Works for both C++ and Kernel driver compilations.
+ Added expected size assertions for ALL of the structures in xclbin.h

### DIFF
--- a/src/runtime_src/core/include/xclbin.h
+++ b/src/runtime_src/core/include/xclbin.h
@@ -75,6 +75,31 @@
   #endif
 #endif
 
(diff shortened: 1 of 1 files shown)


rozumx:
SIZE_OF_STRUCT is indeed unused for if it was, an error would be produced.   
The purpose behind this macro was to assist me in determining the initial size of the structures.  This value was then later added to the assert.

As for the C-style macros, the choices are limited due to the kernel compiler.  

I left it in for "future" structure size determination.

PR #2957, opened 2020-03-03
xclbinutil: Added AIE_METADATA section & expanded "--quiet" support

rozumx:
Work Done
---------
+ Added support for the AIE_METADATA JSON section
+ Expanded "--quiet" option to suppress 'known' xclbinutil exceptions.
+ Defined the enumeration values for axlf_section_kind.


src/runtime_src/tools/xclbin/SectionAIEMetadata.cxx
```diff
@@ -0,0 +1,86 @@
...
+
+void 
+SectionAIEMetadata::marshalToJSON( char* _pDataSection, 
+                                   unsigned int _sectionSize, 
+                                   boost::property_tree::ptree& _ptree) const
+{
+    XUtil::TRACE("");
+    XUtil::TRACE("Extracting: AIE_METADATA");
+
+    std::unique_ptr<unsigned char> memBuffer(new unsigned char[_sectionSize + 1]);
```
stsoe: mismatched free, either std::vector or std::unique_ptr<unsigned char[]>

rozumx:
> Changes look good. Just curious about the way we parse options, we can't call some API to parse it and have to do it manually looping through all args?

There is a function to do just that in XclBinUtilMain::main_().  

The reason why this "primitive" work is being done here is for two reasons:
1) There is a request to not report "known" exceptions if the --quiet option is used.   
- A "known" exception an XclBinUtilException
- The main code that will be parsing the options is contained within the try/catch block, so we need to look for this value sooner then later, so it is too late to look for this value. 
- I could add the boost programming option support to do the same, but it will cause the code base to become rather large and more difficult to read.

2) There is a "functionality" change with regards to the boost program option parsers.
- This other fix is to address a boost change of functionality with regards to "implicit values".  This issue here is that the documentation stated that an equal assignment value was needed (e.g., --info=console) but the code accepted a blank space (e.g., --info console).  When the boost code was updated to reflect the documentation, use-cases were broken, hence this pre-option parser code.

src/runtime_src/tools/xclbin/SectionAIEMetadata.cxx
rozumx: Thanks for the catch.
Issue resolved and I am working on addressing this issue elsewhere in the code base.
Note: These other changes will not be part of this pull request.

PR #2963, opened 2020-03-04
iops test in xbutil validate

AShivangi:
- added "-h" (hidden) option to run iops test
- running all combinations of  cmd_per_batch = { 8,16,32,64,128,256,1024,2048 } and num_execs = { 8,16,32,64,128,256,1024,2048 } and printing out the highest IOPS
- takes 7-8 seconds to run
Output:
```
sagarw@xsjsagarwal:/proj/xsjhdstaff1/sagarw/XRT/build$ xbutil validate -hq
INFO: Found 1 cards

INFO: Validating card[0]: xilinx_u200_xdma_201830_2
AUX POWER NOT CONNECTED, ATTENTION
Board not stable for heavy acceleration tasks.
WARN: == AUX power connector check PASSED with warning
INFO: == Starting IOPS test:
........
IOPS: 98253
INFO: == IOPS test PASSED
INFO: Card[0] validated with warnings.

INFO: All cards validated successfully but with warnings.
```

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
+++ b/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
@@ -1341,48 +1341,50 @@ int xcldev::device::getXclbinuuid(uuid_t &uuid) {
 /*
  * validate
  */
-int xcldev::device::validate(bool quick)
+int xcldev::device::validate(bool quick, bool hidden)
 {
     bool withWarning = false;
     int retVal = 0;
 
-    // retVal = runOneTest("AUX power connector check",
-    //         std::bind(&xcldev::device::auxConnectionTest, this));
-    // withWarning = withWarning || (retVal == 1);
-    // if (retVal < 0)
-    //     return retVal;
-
-    // // Check pcie training
-    // retVal = runOneTest("PCIE link check",
-    //         std::bind(&xcldev::device::pcieLinkTest, this));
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -1373,10 +1373,19 @@ int xcldev::device::validate(bool quick)
...
+        withWarning = withWarning || (retVal == 1);
+        if (retVal < 0)
+            return retVal;
+    }
+    
     // Skip the rest of test cases for quicker turn around.
     if (quick)
         return withWarning ? 1 : 0;
-
+
```
maxzhen: extra spaces.

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -2019,3 +2032,182 @@ int xcldev::device::testM2m()
...
+            }
+        }
+    }
+    auto end = std::chrono::high_resolution_clock::now();
+    duration = (std::chrono::duration_cast<std::chrono::microseconds>
+    (end - start)).count();
+}
+
+int 
+xcldev::device::iopsTest()
```
stsoe: This function is too long.  For example the triple for-loop starting at line 2177 has to be broken.

PR #2965, opened 2020-03-04
Amend earlier changes to memory bank limit

stsoe:
Amend #2763 and #2844 and make bank limit a symbolic constant.  Fix
hardwired usage of original old limit (63).


stsoe:
retest this please

PR #2971, opened 2020-03-05
XRT-270: Acquire CU context when creating kernel

stsoe:
Support opening kernel in shared or exclusive access mode, such that
underlying compute units are opened accordingly for the lifetime of
the kernel object.


stsoe:
retest this please

abalasa:
retest this please

PR #2972, opened 2020-03-05
xocl: fix coverity bugs

rajkumar-xilinx:
-Fix coverity#202200 Out-of-bounds write
-Fix coverity#208669 Uninitialized pointer field
-Fix coverity#208666 Uninitialized scalar variable
-Fix coverities 208668, 208665 for Uncaught exception cases


src/runtime_src/core/common/config_reader.cpp
```diff
@@ -171,10 +171,18 @@ get_env_value(const char* env)
 bool
 get_bool_value(const char* key, bool default_value)
```
stsoe: Please revert all changes to config_reader.   Do not hide the exceptions, catch them where you want to deal with them. Coverity points to errors in the main program, not in config_reader.cpp.

src/runtime_src/core/common/xclbin_parser.cpp
```diff
@@ -506,7 +506,11 @@ get_kernel_inst_addrs(const axlf* top)
   pt::ptree xml_project;
   std::stringstream xml_stream;
   xml_stream.write(xml_data, xml_size);
-  pt::read_xml(xml_stream, xml_project);
+  try {
+    pt::read_xml(xml_stream, xml_project);
+  } catch(...) {
+    return addrVec;
+  }
```
stsoe: Revert

src/runtime_src/core/pcie/emulation/cpu_em/generic_pcie_hal2/shim.cxx
```diff
@@ -80,6 +80,8 @@ namespace xclcpuemhal2 {
     mCloseAll = false;
     bUnified = _unified;
     bXPR = _xpr;
+    mCore = nullptr;
+    mSWSch = nullptr;
```
stsoe: Why are these and other data members not initialized in the initializer-list ?

src/runtime_src/core/common/config_reader.cpp
chvamshi-xilinx: Thanks @stsoe  for the review. Can you please elaborate on, "coverity points to errors in the main program, not in config_reader.cpp"

rajkumar-xilinx: I see that other APIs like get_string_value()  and get_unit_value() also handles the exceptions in similar way. Hence, I have added catching the exception in this file only.

src/runtime_src/core/pcie/emulation/cpu_em/generic_pcie_hal2/shim.cxx
rajkumar-xilinx: Coverity is showing these variables are being used without initializing them. Hence, added them here.

src/runtime_src/core/common/config_reader.cpp
stsoe: Yes, but that is wrong too.  Only catch exceptions if you are acting on them.  For the Coverity errors you fixed, the problem is all the way in main().  Never write a main() without try/catch

src/runtime_src/core/pcie/emulation/cpu_em/generic_pcie_hal2/shim.cxx
stsoe: You can move them to the constructor member initializer list; that is preferable.

src/runtime_src/core/common/config_reader.cpp
stsoe: Check here.  The error is main.
https://scan9.coverity.com/reports.htm#v37066/p13446/fileInstanceId=62266992&defectInstanceId=7309466&mergedDefectId=208668

rajkumar-xilinx: We need to add below changes in the main() to handle this exception in the root main function. Please let me know if it looks fine?

    try {
      ret = xcldev::xclAwssak(argc, argv);
    } catch (...) {
      std::cout << "Exception caught during xclAwssak()\n";
    }

stsoe: As I mentioned, it would be wrong to try catch these uncaught exceptions in config_reader. Always catch the exceptions where you can take appropriate action;  hiding / ignoring bad user input is not good at all, so it is much better to propagate to main() and take corrective action there.

In general always please catch (const std::exception& ex) and error out with ex.what() so we know what failed and why.  awsak is otherwise almost correct.

However xbmgmt.cpp is not correct.   Lots of code before the later try/catch could throw and is not caught.  Please always write main() c++ programs as something like this:

```
int main(int argc, char* argv[])
{
   try {
     return static_main(argc,argv);  // chose better name 
   }
   catch (const std::exception& ex) {
    xrt_core::send_exception_message(....,..); // or just std::cout
   }
   catch (...) {
     xrt_core::send_exception_message(....); // or just std::cout
  }
  return 1;
}
```

rajkumar-xilinx: Thanks Soren, will make necessary changes for coverity#208670 as you suggested. Remaining coverities fixed in this PR.
I will fix cov#208670 in the next PRs.

src/runtime_src/core/pcie/tools/awssak/main.cpp
```diff
@@ -20,5 +20,13 @@
...
+    int ret = 0;
+
+    try {
+      ret = xcldev::xclAwssak(argc, argv);
+    } catch (const std::exception& ex) {
+      std::cout << ex.what() << std::endl;
+    }
+
+    return ret;
 }
```
stsoe: As written above, the function will return 0 when an exception is thrown.  Please do:
```
try {
  return xcldev::xclAwssak(...);
}
catch (const std::exception& ex) {
   ...;
}
catch (...) {
  .....;
}
return 1;
```

rajkumar-xilinx:
retest this please

rajkumar-xilinx:
retest this please

stsoe:
retest this please

PR #2980, opened 2020-03-06
Config subcmd for xbmgmt2 advanced

AShivangi:
**control flows for xbmgmt2 advanced config**: in an effort to keep changes to the minimum, this PR addresses only the plumbing for config option. 

After talking to Stephen I have removed NIFD related changes and we'll revisit them after 20.1

Closing the old PR


src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
```diff
@@ -19,32 +19,257 @@
...
+    CONFIG_SECURITY = 0,
+    CONFIG_CLK_SCALING,
+    CONFIG_CS_THRESHOLD_POWER_OVERRIDE,
+};
+
+enum memType {
+    UNKNOWN= 0,
+    DDR,
+    HBM, 
+
```
stsoe: I suggest using enum classes and keep the enum enumerator names lower case.
E.g.  
`enum class config_type : uint16_t { config_security, config_clk_scaling, ...};`
The enum classes provide type safety and  the enumerator names are local to the enum.

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
```diff
@@ -19,32 +19,257 @@
...
+namespace {
+
+enum class configType {
+    config_security = 0,
+    config_clk_scaling,
+    config_cs_threshold_power_override,
+};
+
+enum class memType {
+    unknown= 0,
```
rozumx: The individual enumeration values should have a common per-appended value to indicate what enumeration this value is from and to prevent name collisions.  Some possible examples could be:
memType_unknown 
```
MT_unknown
memt_unknown
MemT_unknown
```

src/runtime_src/core/tools/xbmgmt2/OO_Config.h
```diff
@@ -29,6 +29,16 @@ class OO_Config : public OptionOptions {
  private:
   std::string m_device;
   bool m_help;
+  bool daemon;
```
rozumx: Please follow the previous author of this code base and pre-append "m_" to the internal variables.

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
stsoe: The beauty of enum classes is that there is no name collision.   Let's all use enum class so that enum variables are scoped to the class and let's always use lower case enum variable names  that are succinctly named.

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
```diff
@@ -19,32 +19,257 @@
...
+    config_security = 0,
+    config_clk_scaling,
+    config_cs_threshold_power_override,
+};
+
+enum class memType {
+    mem_unknown= 0,
+    mem_ddr,
+    mem_hbm, 
+
```
stsoe: Why the redundant config_ and mem_, please see my comment to Stephen's comment.

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
```diff
@@ -19,32 +19,257 @@
...
+        auto dev = xrt_core::get_mgmtpf_device(idx);
+        update_device_conf(dev, m_clk_scale, configType::clk_scaling);
+      }
+    }
+    
+    //update threshold power override
+    if (!m_power_override.empty()) {
+      for(auto idx : device_indices) {
+        auto dev = xrt_core::get_mgmtpf_device(idx);
+        update_device_conf(dev, m_power_override, configType::cs_threshold_power_override);
```
stsoe: Take it or leave it, but I think the mixed case doesn't look nice with the style otherwise used.  I'd by far prefer enum class config_type over the camel case.

AShivangi: Sorry! missed that

abalasa:
retest this please

PR #2981, opened 2020-03-06
CR-1056508 clock race condition and XRT-280 cmc_freeze/free refactor

xdavidz:
hahaha... here you go

it's going to be wordy...

1) add mutex for health chread
2) clean up cmc_freeze/free
3) adjust the ocl_freqscaling, reduce xclbin download time(we used to do it twice, which is not necessary)
4) refactor callbacks from icap to clock, now it is handled by axigate subdev.

I will test u200, u50 test passed.


xdavidz:
U200 test passed too. here is the script
/proj/rdi/staff/davidzha/share/clock.sh U200
/proj/rdi/staff/davidzha/share/clock.sh U50

maxzhen:
@xdavidz , please resolve the conflict before this pr can be merged.

PR #2985, opened 2020-03-07
U250 2RP support

houlz0507:


src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
```diff
@@ -320,14 +323,6 @@ int xocl_program_shell(struct xocl_dev *xdev, bool force)
 		goto failed;
 	}
 
-	if (xdev->core.fdt_blob) {
-		vfree(xdev->core.fdt_blob);
-		xdev->core.fdt_blob = NULL;
```
maxzhen: where do you do the clean up now?

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.h
```diff
@@ -143,6 +148,9 @@ struct xclmgmt_dev {
 
 	/* save config for pci reset */
 	u32 saved_config[8][16];
+
+	/* programming shell flag */
+	u32 rp_program;
```
maxzhen: don't  you need a lock to protect this when  you touch it in both mailbox request handler and sysfs handler?

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
houlz0507: 1. resolve deadlock to offline mailbox in mailbox handler
2. It is more safe to sync with hot reset request.
We mgmtpf completes PLP programming, it notify userpf. Userpf will call refresh which free the current blob and use the new one.

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.h
houlz0507: The changing should be always in sequence. And the consumer is just the xbmgmt tool to check if programming completes.  the risk is very low.
1) sysfs set rp_program to 1
2) sysfs notify userpf
3) userpf mailbox request programing, mgmtpf receives req and set rp_program to 2
4) mgmtpf notify userpf program completes.
5) userpf mailbox request new metadata, mgmtpf set rp_program to 0

maxzhen: Whenever we design any protocol b/w mgmt and user pf, we have to keep in mind that the user driver is not trusted. It may not do what you expected. Is it possible that the user driver can do something through mailbox to mess up with rp_program and cause mgmt to fail (leaking memory or crash)?

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
maxzhen: ok

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.h
houlz0507: not at all. it can cause xbmgmt  to exit checking loop earlier. It is in userland.  It will not cause any mgmtpf driver issue. rp_program is just a flag for xbmgmt tool. user can re-run PLP program if he wants.

maxzhen: ok

PR #2997, opened 2020-03-10
XRT-287 Support u250 XDMA PLP

houlz0507:


houlz0507:
> default P2P BAR becomes 32M now??

No, this function is actually looking up DMA BAR. I would rule out the user_bar (bar 0) without reading any register in BAR0. Using a 32M check will be simplest change. Otherwise, we need to provide a XDMA API to configure DMA bar.

PR #2999, opened 2020-03-10
adding u55n, u55c and u50lv device ID support

sjmoneill:
Added support for following device IDs to devices.h
u55n - 0x5058. 0x5059, no golden image, using BLP based recovery
u55c - 0x505C. 0x505D, no golden image, using BLP based recovery
u50lv - 0x5060. 0x5061, no golden image, using BLP based recovery
Change is verbose, i.e. largely replicated u50 tables. This could be condensed somewhat as all cards are raptor 2.0 and very similar to u50

Also fixed an issue in xocl_subdev.c. The code to append '_recovery' to the vbnv when VSEC platform info is XOCL_VSEC_PLAT_RECOVERY (0) was crashing the server.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/devices.h
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/devices.h
@@ -2111,6 +2111,60 @@ struct xocl_subdev_map {
 		.board_name = "u50"					\
 	}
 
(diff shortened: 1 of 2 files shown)


maxzhen:
ok to test

PR #3003, opened 2020-03-10
CR-1056370 fix and debug mode added

sarab96:
1) Fix for CR-1056370: timeout fixed for work_item command
2) Added debug modes

### DIFF
--- a/src/runtime_src/core/common/config_reader.h
+++ b/src/runtime_src/core/common/config_reader.h
@@ -264,6 +264,16 @@ get_xclbin_programming()
 }
 
 /**
+ * Enable xma mode2 = only one cu command submission per xma_session at a time
+ */
+inline bool
+get_xma_mode2()
+{
+  static bool value = detail::get_bool_value("Runtime.xma_mode2", false);
+  return value;
+}
+
+/**
  * Enable / Disable kernel driver scheduling when running in hardware.
  * If disabled, xrt will be scheduling either using the software scheduler
  * (sws) or the microblaze scheduler (mbs) if ert is enabled
--- a/src/runtime_src/core/common/config_reader.h
+++ b/src/runtime_src/core/common/config_reader.h
@@ -264,7 +264,7 @@ get_xclbin_programming()
 }
 
 /**
- * Enable xma mode2 = only one cu command submission per xma_session at a time
+ * Enable xma mode2 = only two cu command submission per xma_session at a time
  */
 inline bool
 get_xma_mode2()
--- a/src/xma/include/xma.h
+++ b/src/xma/include/xma.h
@@ -57,6 +57,7 @@ int32_t xma_get_default_ddr_index(int32_t dev_index, int32_t cu_index, char* cu_
 void xma_get_session_cmd_load(void);
 
 void xma_enable_mode1(void);//Hidden mode. To allow bulk submission of cu commands
+void xma_enable_mode2(void);//Hidden mode. To allow only one cu command submission per session at a time
 
 #ifdef __cplusplus
 }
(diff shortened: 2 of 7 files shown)


src/xma/src/xmaapi/xma_utils.cpp
```diff
@@ -506,8 +506,25 @@ int32_t get_default_ddr_index(int32_t dev_index, int32_t cu_index) {
 }
 
 void xma_enable_mode1(void) {
+    if (xrt_core::config::get_xma_mode2()) {
+        xma_logmsg(XMA_ERROR_LOG, XMAUTILS_MOD, "xma_enable_mode1: Mode2 is selected. Use either mode1 or mode2");
```
rozumx: Error message is a little confusing...could you elaborate a little bit more.  Something like:

> xma_enable_mode1: Mismatch between modes - The config currently is using mode2, but the code flow is enabling mode1.

src/xma/src/xmaapi/xma_utils.cpp
```diff
@@ -506,8 +506,25 @@ int32_t get_default_ddr_index(int32_t dev_index, int32_t cu_index) {
...
+        return;
+    }
+    if (g_xma_singleton->num_execbos != 3) {
+        xma_logmsg(XMA_ERROR_LOG, XMAUTILS_MOD, "xma_enable_mode1: Mode already changed");
```
rozumx: Update the error message:
xma_enable_mode1: Mode1 already in effect, no action taken.

src/xma/src/xmaapi/xmaapi.cpp
```diff
@@ -17,11 +17,6 @@
 
 #include <sys/stat.h>
 #include <sys/types.h>
-#include <errno.h>
```
rozumx: Thank you for cleaning up un-needed header files.

src/xma/include/xma.h
```diff
@@ -57,6 +57,7 @@ int32_t xma_get_default_ddr_index(int32_t dev_index, int32_t cu_index, char* cu_
 void xma_get_session_cmd_load(void);
 
 void xma_enable_mode1(void);//Hidden mode. To allow bulk submission of cu commands
```
rozumx: Consider refactoring` xma_enable_mode1()` and `xma_enable_mode2()` to be 1 method, in case there every is a future "mode3".  For example: 
`xma_enable_mode(unsigned int mode) {}`

sarab96: good suggestion. Keeping them separate because mode1 should be rarely used. And separate debug modes seem better for clarity

src/xma/src/xmaapi/xma_utils.cpp
sarab96: done

src/runtime_src/core/common/config_reader.h
```diff
@@ -263,6 +263,16 @@ get_xclbin_programming()
   return get_xclbin_programing();
 }
 
+/**
+ * Enable xma mode2 = only one cu command submission per xma_session at a time
+ */
+inline bool
+get_xma_mode2()
```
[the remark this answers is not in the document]
sarab96: This is for ini setting. So allowed only for a given application run and only one xclbin per application run. More in comment to other question below.

src/xma/src/xmaapi/xma_utils.cpp
sarab96: These are debug modes. Either mode1 or mode2 can be used for an application run. Made change to the code. Safety check if mode is already changed.

abalasa:
retest this please

PR #3009, opened 2020-03-11
XRT-106: Created a bootloader for xbutil2/xbmgmt2

rozumx:
Created a bootloader for both xbutil2/xbmgmt2 for both Windows and Linux
Work Done
-----------
+ Move the target binary output files to xrt/bin/unwrapped
+ Created wrapper scripts that parse the options looking for a "new" option.  If found, then xbutil2/xbmgmt2 will be called on Linux.  On Windows, a simple message is produced indicating that the "new" option is only valid on Linux.
+ Created a loader script that will:
  a. Find the unwrapped executable
  b. Setup the environment
  c. Call the unwrapped executable (minus any reserved options)
+ Added support to the Linux script detect if it is being called from Windows, if so, it will redirect the call to the Windows batch script.
+ Updated various CMake configuration files to support this flow.


src/runtime_src/tools/loader/xbmgmt
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
```
stsoe: I think individual application loaders should live under the application, e.g. tools/xbmgmt2.  I realize the loader is for both legacy xbmgmt and new xbmgmt2 but I still think it belongs with latter.   Ditto for xbutil loader.

sonals: We do need to support use case where end users can run /opt/xilinx/xrt/bin/xbutil without setting up paths, etc. This works today and should work as we move to loaders.

stsoe: @sonals , I don't understand your comment in the context?  My comment merely suggested where the loader lives in our source tree, not where they are installed.  They obviously are installed under /opt/xilinx/xrt/bin.

rozumx: @sonals This wrapper solution support your use-case.  In fact, it is even better, let me explain.

In the past, if the user invoked xbutil directly, there was always a possibility of the wrong XRT shared libraries being used (especially when the user invokes xbutil from the TA and has XILINX_XRT pointing to a local installed version).   If you recall, during 2018.2-XDF we had just this issue which it caused a lot of "random" problems.

With this wrapper solution, it is guaranteed that when the actual binary version of xbutil/xbmgmt is execute, the shared XRT libraries associated with xbutil/xbmgmt will be used (thanks to the loader).

rozumx:
@stsoe The code has been updated to your suggestion.

PR #3012, opened 2020-03-11
CR-1057132: xocl qdma: add setting of stm register base during devive probe

karenx-xilinx:
For CR-1057132


abalasa:
retest this please

karenx-xilinx:
Add an extra parameter of STM register base when probing the qdma device.

Tested with u200 qdma image in the 2020.1_latest.

PR #3026, opened 2020-03-13
xbmgmt2 control flow fixes

AShivangi:
- enforce device specification for xbmgmt2 program but not for xbmgmt2 scan (default is "all")

Eg:
   xbmgmt2 status --report=scan --device=all _[works]_
   xbmgmt2 status --report=scan _[works: reports for all devices]_
   xbmgmt2 program --update --device=all _[works]_
   xbmgmt2 program --update _[throws an error]_

- flash_type is present in many locations, so don't fail when not found. Look check all locations before throwing an error

### DIFF
--- a/src/runtime_src/core/tools/common/XBUtilities.cpp
+++ b/src/runtime_src/core/tools/common/XBUtilities.cpp
@@ -284,7 +284,8 @@ XBUtilities::wrap_paragraphs( const std::string & _unformattedString,
 void
 XBUtilities::parse_device_indices(std::vector<uint16_t> &device_indices, const std::string &device)
 {
-  if(boost::iequals(device, "all")) {
+  //if no device is passed or "all" is specified, parse all devices
+  if(boost::iequals(device, "all") || device.empty()) {
     ::verbose("Sub command : --device");
     //get all devices
     auto total = xrt_core::get_total_devices(false).first;
@@ -294,7 +295,7 @@ XBUtilities::parse_device_indices(std::vector<uint16_t> &device_indices, const s
     for(uint16_t i = 0; i < total; i++) {
       device_indices.push_back(i);
     }
-  } else if (!device.empty()) {
+  } else {
     ::verbose("Sub command : --device");
     using tokenizer = boost::tokenizer< boost::char_separator<char> >;
     boost::char_separator<char> sep(", ");
--- a/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
@@ -491,7 +491,7 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
 //                     Download the accelerator program for card 2
 //                       xbutil program -d 2 -p a.xclbin
 {
-  XBU::verbose("SubCommand: examine");
+  XBU::verbose("SubCommand: program");
 
   XBU::verbose("Option(s):");
   for (auto & aString : _options) {
@@ -589,7 +589,9 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
   }
 
   // get all device IDs to be processed
-  std::vector<uint16_t> device_indices; //instead of saving this, can we save Flasher objects?
+  if(device.empty())
+    throw xrt_core::error("Please specify a device using --device option");
+  std::vector<uint16_t> device_indices;
   XBU::parse_device_indices(device_indices, device);
 
   if (!update.empty()) {
--- a/src/runtime_src/core/tools/xbmgmt2/flash/flasher.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/flash/flasher.cpp
@@ -40,36 +40,38 @@ Flasher::E_FlasherType Flasher::getFlashType(std::string typeStr)
 {
     std::string err;
     E_FlasherType type = E_FlasherType::UNKNOWN;
-    if (typeStr.empty())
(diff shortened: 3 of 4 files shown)


abalasa:
retest this please

src/runtime_src/core/tools/xbmgmt2/flash/flasher.cpp
```diff
@@ -40,36 +40,38 @@ Flasher::E_FlasherType Flasher::getFlashType(std::string typeStr)
...
-    if (typeStr.empty())
-      typeStr = xrt_core::device_query<xrt_core::query::flash_type>(m_device);
 
-    if (typeStr.empty())
-    {
-        getProgrammingTypeFromDeviceName(mFRHeader.VBNVName, type);
-    }
-    else if (typeStr.compare("spi") == 0)
-    {
+    // check various locations for flash_type
```
rozumx: Add some more documentation around this code on why this is needed...for future developers.

src/runtime_src/core/tools/xbmgmt2/flash/flasher.cpp
```diff
@@ -40,36 +40,38 @@ Flasher::E_FlasherType Flasher::getFlashType(std::string typeStr)
...
+        if (typeStr.empty())
+          typeStr = xrt_core::device_query<xrt_core::query::flash_type>(m_device);
+    } catch (...) {}
+    try {
+        if (typeStr.empty())
+            getProgrammingTypeFromDeviceName(mFRHeader.VBNVName, type);
+    } catch (...) {}
+    
+    //map flash_tyoe str to E_FlasherType
+    if (typeStr.compare("spi") == 0) {
```
rozumx: This code is converting a string value to an enumeration value, which could be useful elsewhere in the code.   Please do refactor to code into a helping function.   Something similar to `E_FlasherType getFlasherType(const string & typeStr);`

src/runtime_src/core/tools/xbmgmt2/flash/flasher.cpp
```diff
@@ -36,43 +36,54 @@
...
     }
-    else if (typeStr.compare("ospi_versal") == 0)
-    {
+    else if (typeStr.compare("ospi_versal") == 0) {
         type = E_FlasherType::OSPIVERSAL;
     }
-    else
-    {
-        std::cout << "Unknown flash type: " << typeStr << std::endl;
+    else {
```
rozumx: Being that the enum E_FlasherType has an UNKNOWN value, please return this value instead of throwing an exception.  

It will be the responsibility of the calling code to determine if an exception needs to be thrown.

src/runtime_src/core/tools/xbmgmt2/flash/flasher.cpp
```diff
@@ -36,43 +36,54 @@
...
-    if (typeStr.empty())
-    {
-        getProgrammingTypeFromDeviceName(mFRHeader.VBNVName, type);
-    }
-    else if (typeStr.compare("spi") == 0)
-    {
+/* 
+ * map flash_tyoe str to E_FlasherType
+ */
+void Flasher::typeStr_to_E_FlasherType(const std::string& typeStr, E_FlasherType &type)
```
rozumx: This method should return back the enumeration type, as opposed to via a return variable.   Doing so allows the calling program to act upon the value instead of first creating a variable and then setting it.

src/runtime_src/core/tools/xbmgmt2/flash/flasher.cpp
```diff
@@ -36,43 +36,54 @@
...
+    try {
+        if (typeStr.empty())
+          typeStr = xrt_core::device_query<xrt_core::query::flash_type>(m_device);
+    } catch (...) {}
+    try {
+        if (typeStr.empty())
+            getProgrammingTypeFromDeviceName(mFRHeader.VBNVName, type);
+    } catch (...) {}
+    
+    typeStr_to_E_FlasherType(typeStr, type);
```
rozumx: Check here to determine if the enumeration value is valid, if not then throw an exception.

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -284,7 +284,8 @@ XBUtilities::wrap_paragraphs( const std::string & _unformattedString,
 void
 XBUtilities::parse_device_indices(std::vector<uint16_t> &device_indices, const std::string &device)
 {
-  if(boost::iequals(device, "all")) {
+  //if no device is passed or "all" is specified, parse all devices
+  if(boost::iequals(device, "all") || device.empty()) {
```
rozumx: (minor item)
This functionality does work.  You might wish to consider to break this logic into to separate lines for future EoU reading.  For example:
```
bool parse_all_devices = device.empty() | boost::iequals("device", "all");
if (parse_all_devices) {
```

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -589,7 +589,9 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
   }
 
   // get all device IDs to be processed
-  std::vector<uint16_t> device_indices; //instead of saving this, can we save Flasher objects?
+  if(device.empty())
+    throw xrt_core::error("Please specify a device using --device option");
+  std::vector<uint16_t> device_indices;
```
rozumx: Add a new line, especially since there is no {} for the if statement above.

PR #3030, opened 2020-03-16
XRT-289 Need to flash XSABIN file onto flash

maxzhen:
This change set implement the feature to flash a stripped version of dsabin onto flash together with the MCS bitstream.

In order to test this code path, xbmgmt flash --file option is added to be able to write content of a file onto flash and read it back. Note that this option is for developer use. There is no guarantee that it will not corrupt the bitstream on the flash if it's used to write data.

The data (stripped version of dsabin or content of a plain file) is saved on flash at the very end (of the first chip). There should be some tens of MB of free space in this area based on today's shell size and flash size. This implementation limited the data size to be at most 1MB anyway just to be safe. If the data is corrupted by bitstream, it should be easy to detect since we have parity and magic words in meta data on the flash.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
@@ -676,6 +676,9 @@ static int qspi_probe(struct xocl_flash *flash)
 {
 	int ret;
 
+	/* Probing on first flash only. */
+	flash->qspi_curr_slave = 0;
+
 	flash_set_ctrl(flash, QSPI_CR_INIT_STATE);
 
 	/* Find out fifo depth before any read/write operations. */
@@ -903,7 +906,7 @@ static bool is_valid_offset(struct xocl_flash *flash, loff_t off)
 	/* Assuming all flash are of the same size, we use
 	 * offset into flash 0 to perform boundary check. */
 	faddr.slave = 0;
-	return flash_faddr2offset(&faddr) < flash->flash_size;
+	return flash_faddr2offset(&faddr) <= flash->flash_size;
 }
 
 /*
@@ -920,10 +923,14 @@ flash_read(struct file *file, char __user *buf, size_t n, loff_t *off)
 
 	FLASH_INFO(flash, "reading 0x%lx bytes @0x%llx", n, *off);
 
-	if (n == 0 || !is_valid_offset(flash, *off + n)) {
+	if (!is_valid_offset(flash, *off)) {
 		FLASH_ERR(flash, "Can't read: out of boundary");
 		return -EINVAL;
 	}
+
+	if (n == 0 || *off == flash->flash_size)
+		return 0;
+	n = min(n, flash->flash_size - (size_t)*off);
 
 	page = vmalloc(FLASH_PAGE_SIZE);
 	if (page == NULL)
@@ -1223,10 +1230,22 @@ static ssize_t properties_show(struct device *dev,
 
 static DEVICE_ATTR_RO(properties);
 
+/* Show size of one chip. Double it for total size for dual chip platforms. */
+static ssize_t size_show(struct device *dev,
+	struct device_attribute *attr, char *buf)
+{
+	struct xocl_flash *flash = dev_get_drvdata(dev);
+
+	return sprintf(buf, "%ld\n", flash->flash_size);
+}
+
+static DEVICE_ATTR_RO(size);
+
 static struct attribute *flash_attrs[] = {
 	&dev_attr_bar_off.attr,
 	&dev_attr_flash_type.attr,
 	&dev_attr_properties.attr,
+	&dev_attr_size.attr,
 	NULL,
 };
 
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
@@ -906,7 +906,7 @@ static bool is_valid_offset(struct xocl_flash *flash, loff_t off)
 	/* Assuming all flash are of the same size, we use
 	 * offset into flash 0 to perform boundary check. */
 	faddr.slave = 0;
-	return flash_faddr2offset(&faddr) <= flash->flash_size;
+	return flash_faddr2offset(&faddr) < flash->flash_size;
 }
 
 /*
@@ -923,13 +923,10 @@ flash_read(struct file *file, char __user *buf, size_t n, loff_t *off)
 
 	FLASH_INFO(flash, "reading 0x%lx bytes @0x%llx", n, *off);
 
-	if (!is_valid_offset(flash, *off)) {
+	if (n == 0 || !is_valid_offset(flash, *off)) {
 		FLASH_ERR(flash, "Can't read: out of boundary");
-		return -EINVAL;
-	}
-
-	if (n == 0 || *off == flash->flash_size)
 		return 0;
+	}
 	n = min(n, flash->flash_size - (size_t)*off);
 
 	page = vmalloc(FLASH_PAGE_SIZE);
@@ -1066,10 +1063,11 @@ flash_write(struct file *file, const char __user *buf, size_t n, loff_t *off)
 
 	FLASH_INFO(flash, "writing 0x%lx bytes @0x%llx", n, *off);
 
-	if (n == 0 || !is_valid_offset(flash, *off + n)) {
+	if (n == 0 || !is_valid_offset(flash, *off)) {
 		FLASH_ERR(flash, "Can't write: out of boundary");
 		return -EINVAL;
 	}
+	n = min(n, flash->flash_size - (size_t)*off);
 
 	page = vmalloc(FLASH_HUGE_PAGE_SIZE);
 	if (page == NULL)
--- a/src/runtime_src/core/pcie/tools/xbmgmt/cmd_flash.cpp
+++ b/src/runtime_src/core/pcie/tools/xbmgmt/cmd_flash.cpp
@@ -43,8 +43,8 @@ const char *subCmdFlashUsage =
     "--shell --path file --card bdf [--type flash_type]\n"
     "--sc_firmware --path file --card bdf";
 
-#define fmt_str		"    "
-#define DEV_TIMEOUT	60
+#define fmt_str        "    "
+#define DEV_TIMEOUT    60
 
 static int scanDevices(bool verbose, bool json)
 {
@@ -179,7 +179,7 @@ static int updateSC(unsigned index, const char *file)
     dev->sysfs_put("", "shutdown", errmsg, "0\n");
     if (!errmsg.empty()) {
         std::cout << "ERROR: online userpf failed. Please warm reboot." << std::endl;
-	return ret;
+    return ret;
     }
 
     int wait = 0;
@@ -204,6 +204,7 @@ static int updateShell(unsigned index, std::string flashType,
 {
     std::shared_ptr<firmwareImage> pri;
     std::shared_ptr<firmwareImage> sec;
+    std::shared_ptr<firmwareImage> stripped;
 
     if (!flashType.empty()) {
         std::cout << "CAUTION: Overriding flash mode is not recommended. " <<
@@ -223,14 +224,19 @@ static int updateShell(unsigned index, std::string flashType,
     if (pri->fail())
         return -EINVAL;
 
+    stripped = std::make_shared<firmwareImage>(primary, STRIPPED_FIRMWARE);
+    if (stripped->fail())
+        stripped = nullptr;
+
     if (secondary != nullptr) {
         sec = std::make_shared<firmwareImage>(secondary,
             MCS_FIRMWARE_SECONDARY);
         if (sec->fail())
             sec = nullptr;
     }
-
-    return flasher.upgradeFirmware(flashType, pri.get(), sec.get());
+ 
+    return flasher.upgradeFirmware(flashType, pri.get(), sec.get(),
+        stripped.get());
 }
 
 // Reset shell to factory mode.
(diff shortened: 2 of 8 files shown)


maxzhen:
> 
> 
>     1. Shouldn't we pass this option as --shell --path?
> 
>     2. Should we port this to xbmgmt2?

This new options has nothing to do with shells. It's a generic way to flash a regular file onto the board.
Yes, this should be ported to xbmgmt2. But I'd like to make it work for xbmgmt first.

maxzhen:
> 
> 
> This looks good then. Can you create an XRT ticket to port this to xbmgmt2?

XRT-311 is created to track this.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
```diff
@@ -903,7 +906,7 @@ static bool is_valid_offset(struct xocl_flash *flash, loff_t off)
 	/* Assuming all flash are of the same size, we use
 	 * offset into flash 0 to perform boundary check. */
 	faddr.slave = 0;
-	return flash_faddr2offset(&faddr) < flash->flash_size;
+	return flash_faddr2offset(&faddr) <= flash->flash_size;
```
rozumx: Could you elaborate more on why the "equal" is needed.   I ask because only flash addresses (e.g., 0 to size of flash) makes sense.  Having it less then or equal "feels" wrong. 

For example: 
If the flash size is 0x10 bytes, what does it mean to have an address of 0x10.

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
```diff
@@ -2180,3 +2224,127 @@ int XSPI_Flasher::upgradeFirmware2Drv(std::istream& mcsStream0,
...
+
+int XSPI_Flasher::xclWriteData(std::vector<unsigned char>& data)
+{
+    const int flash_index = 0;
+
+    if (mFlashDev == nullptr)
+        return -EINVAL;
+
+    if (data.size() > xrt_max_data_len) {
+        std::cout << "meta data is too big (" << data.size() << "B) to flash: "
```
rozumx: Glad to see this error check.  Could the message to improved to report the xrt_max_data_len as well?

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
```diff
@@ -2180,3 +2224,127 @@ int XSPI_Flasher::upgradeFirmware2Drv(std::istream& mcsStream0,
     // Disable bitstream guard.
     return removeBitstreamGuard(mDev.get(), mFlashDev, bsGuardAddr);
 }
+
+const std::string magic = "XRTDATA";
+struct flash_data_header {
+    char magic[7] = "";
+    char ver = 0;
+    uint32_t data_len = 0;
+    uint32_t parity = 0;
```
rozumx: As we have seen in the past with the axlf structures, here can be "future" needs to store additional metadata.   My I suggest adding 64 bytes of "reserved" future space.  For example:
```
struct flash_data_header {
  char magic[7];
  char verion = 0;
  uint32_t data_len_bytes;
  uint32_t parity;
  uint8_t reserved[16];
};
```

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
```diff
@@ -2180,3 +2224,127 @@ int XSPI_Flasher::upgradeFirmware2Drv(std::istream& mcsStream0,
...
+    unsigned int addr = size - sizeof(flash_data_header) - data.size();
+    int ret = writeToFlash(mFlashDev, flash_index, addr,
+        data.data(), data.size());
+    if (ret) {
+        std::cout << "failed to write meta data to flash: " << ret << std::endl;
+        return ret;
+    }
+
+    // Write meta data onto flash
+    flash_data_header header;
```
rozumx: Initialize the structure (including padded bytes) all to zeros.  For example:
`flash_data_header header = {0};`

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
```diff
@@ -2180,3 +2224,127 @@ int XSPI_Flasher::upgradeFirmware2Drv(std::istream& mcsStream0,
     // Disable bitstream guard.
     return removeBitstreamGuard(mDev.get(), mFlashDev, bsGuardAddr);
 }
+
+const std::string magic = "XRTDATA";
+struct flash_data_header {
+    char magic[7] = "";
```
rozumx: Is this value meant to be a "null" terminated string?  or are you simply attempting to initialize the structure to known values.  

If it is to be a "null" terminated string, then add room to include the null character.

If you are attempting to initialize the structure to a known value, then use following syntax when creating the structure (as shown later below)
`flash_data_header header = {0};`

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
```diff
@@ -2180,3 +2224,127 @@ int XSPI_Flasher::upgradeFirmware2Drv(std::istream& mcsStream0,
...
+            << std::endl;
+        return -EINVAL;
+    }
+
+    size_t size = getFlashSize(mDev.get());
+    if (size == 0)
+        return -EINVAL;
+
+    // Write file data onto flash
+    unsigned int addr = size - sizeof(flash_data_header) - data.size();
```
rozumx: Interesting...So does this mean that this XRT DATA will ALWAYS be at the "top" of the FLASH, regardless of its size?

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
```diff
@@ -2180,3 +2224,127 @@ int XSPI_Flasher::upgradeFirmware2Drv(std::istream& mcsStream0,
...
+    flash_data_header header;
+    std::memcpy(header.magic, magic.c_str(), magic.size());
+    header.ver = 0;
+    header.data_len = data.size();
+    header.parity = getParity32(data);
+    addr += data.size();
+    ret = writeToFlash(mFlashDev, flash_index, addr,
+        reinterpret_cast<unsigned char *>(&header), sizeof(header));
+    if (ret) {
+        std::cout << "failed to write meata data header to flash: " << ret
```
rozumx: Spelling error

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
```diff
@@ -2180,3 +2224,127 @@ int XSPI_Flasher::upgradeFirmware2Drv(std::istream& mcsStream0,
...
+            << std::endl;
+        return ret;
+    }
+
+    std::cout << "Persisted " << data.size() << " bytes of meta data to flash "
```
rozumx: Being that this code is C++, you might wish to consider using the boost format, which will make your life easier instead of formatting though the iostreams.  For example, the error message becomes:
`std::cout << boost::format("Persisted %d bytes of meta data to flash %d @0x%x\n") % data.size() % flash_index %addr;`

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
```diff
@@ -2180,3 +2224,127 @@ int XSPI_Flasher::upgradeFirmware2Drv(std::istream& mcsStream0,
...
+        << flash_index << " @0x" << std::hex << addr << std::dec << std::endl;
+    return 0;
+}
+
+int XSPI_Flasher::xclReadData(std::vector<unsigned char>& data)
+{
+    const int flash_index = 0;
+
+    if (mFlashDev == nullptr)
```
rozumx: If we are going to do checks, we should at least inform the user why we are leaving this method.

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
```diff
@@ -2180,3 +2224,127 @@ int XSPI_Flasher::upgradeFirmware2Drv(std::istream& mcsStream0,
...
+        return -EINVAL;
+
+    size_t size = getFlashSize(mDev.get());
+    if (size == 0)
+        return -EINVAL;
```
rozumx: If we are going to do checks, we should at least inform the user why we are leaving this method.

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
```diff
@@ -2180,3 +2224,127 @@ int XSPI_Flasher::upgradeFirmware2Drv(std::istream& mcsStream0,
...
+
+    // Read meta data from flash
+    flash_data_header header;
+    unsigned int addr = size - sizeof(header);
+    std::cout << "reading " << sizeof(header) << " bytes of header from flash "
```
rozumx: Please consider using boost formats

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
```diff
@@ -2180,3 +2224,127 @@ int XSPI_Flasher::upgradeFirmware2Drv(std::istream& mcsStream0,
...
+    addr -= header.data_len;
+    std::cout << "reading " << header.data_len << " bytes of data from flash "
+        << flash_index << " @0x" << std::hex << addr << std::dec << std::endl;
+    ret = readFromFlash(mFlashDev, flash_index, addr, data.data(), data.size());
+    if (ret) {
+        std::cout << "failed to read data from flash: " << ret << std::endl;
+        return ret;
+    }
+    if (header.parity ^ getParity32(data)) {
+        std::cout << "data is corrupted" << std::endl;
```
rozumx: More description should be added here.  For example: "ERROR:  XRT Flash Data is corrupted. Expected: 0xFFEF, Actual: 0xABDF

maxzhen:
Just marked it as do not merge until I have time to get back to this work.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
maxzhen: you're right. this does not make sense. i'll revert it back.

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
maxzhen: will do.

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
maxzhen: will do.

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
maxzhen: will do.

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
maxzhen: will use the = { 0 }  syntax.

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
maxzhen: no, it's in the end of the flash, right before header. header data is in the very end, which is a well known location.

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
maxzhen: will fix.

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
maxzhen: will do.

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
maxzhen: will do.

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
maxzhen: will do.

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
maxzhen: will do.

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
maxzhen: will do.

maxzhen:
@rozumx , i have pushed another commit, please take a look.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/flash.c
```diff
@@ -920,10 +923,11 @@ flash_read(struct file *file, char __user *buf, size_t n, loff_t *off)
 
 	FLASH_INFO(flash, "reading 0x%lx bytes @0x%llx", n, *off);
 
-	if (n == 0 || !is_valid_offset(flash, *off + n)) {
+	if (n == 0 || !is_valid_offset(flash, *off)) {
 		FLASH_ERR(flash, "Can't read: out of boundary");
-		return -EINVAL;
+		return 0;
```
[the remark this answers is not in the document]
maxzhen: For read behavior, I was referring to: http://man7.org/linux/man-pages/man2/read.2.html. It specifically talked about this case:
"If the file offset is at or past the end of file, no
       bytes are read, and read() returns zero"
However, for write, the behavior is not clearly defined (E.g., http://man7.org/linux/man-pages/man2/write.2.html).
Maybe I should map it to the below error when the write is beyond the boundary:
"       ENOSPC The device containing the file referred to by fd has no room
              for the data."
I'll change the write return code to -ENOPSC.

src/runtime_src/core/pcie/tools/xbmgmt/cmd_flash.cpp
```diff
@@ -245,12 +251,12 @@ static int resetShell(unsigned index)
...
         return -ECANCELED;
 
-    return flasher.upgradeFirmware("", nullptr, nullptr);
+    return flasher.upgradeFirmware("", nullptr, nullptr, nullptr);
 }
 
 /* We do not take the risk to flash any bmc marked as UNKNOWN */
 static void isSameShellOrSC(DSAInfo& candidate, DSAInfo& current,
-	bool *same_dsa, bool *same_bmc)
+    bool *same_dsa, bool *same_bmc)
```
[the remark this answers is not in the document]
maxzhen: this is a good suggestion and we can fix it later. In this pull request, i was merely changing the indentation, not the logic of the code.

PR #3031, opened 2020-03-16
Move loader script to unwrapped dir

stsoe:
This works around clash with Vivado loader script, which unfortunately
are not called with absolute path but instead picked from PATH.


stsoe:
By passing pipeline for quick fix for Vivado HEAD breakage.

PR #3045, opened 2020-03-17
XRT-317 add retry when calling wireserver APIs in azure mpd plugin

xuhz:
in azure production environment, retry when calling wireserver APIs are required since  there is no guarantee networking is available.
And when xclbin is downloaded, azure wireserver takes extra steps to verify the xclbin in one of the API -- that step usually takes 12-18s, so the default 20s mailbox rx timeout is not enough. change the default to 60s.


xuhz:
retest this please

PR #3050, opened 2020-03-18
Delete old metadata file and update new metadata file for 2020.1

haeseunglee:
Delete old metadata file and update new metadata file for 2020.1 sprite run


stsoe:
ok to test

PR #3052, opened 2020-03-18
CR-1057699 Check for Errors while waiting for completions

pranjalv-xilinx:
XRT-296 Hit xdma panic when debugging u250 2RP
CR-1057699 Check for Errors while waiting for completions

Look for error being returned by wait_event_interruptible_timeout() in addition to timeout. Not checking for error was causing requests to be freed before being removed for list causing panic.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/lib/libxdma.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/lib/libxdma.c
@@ -1437,7 +1437,9 @@ static int xdma_process_requests(struct xdma_engine *engine,
 		timeout = wait_event_interruptible_timeout(req->arbtr_wait,
 				req->sw_desc_cnt == req->desc_completed,
 				msecs_to_jiffies(10000)); /* 10sec timeout */
-		if (timeout == 0) {
+
+		/* Check for timeouts or errors */
+		if (timeout == 0 || timeout < 0) {
 			pr_err("Request completion timeout\n");
 			engine_reg_dump(engine);
 			rv = -EIO;
--- a/src/runtime_src/core/pcie/driver/linux/xocl/lib/libxdma.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/lib/libxdma.c
@@ -1439,11 +1439,16 @@ static int xdma_process_requests(struct xdma_engine *engine,
 				msecs_to_jiffies(10000)); /* 10sec timeout */
 
 		/* Check for timeouts or errors */
-		if (timeout == 0 || timeout < 0) {
+		if (timeout == 0) {
 			pr_err("Request completion timeout\n");
 			engine_reg_dump(engine);
 			rv = -EIO;
+		} else if (timeout < 0) {
+			pr_err("Request aborted with error = %d\n", timeout);
+			engine_reg_dump(engine);
+			rv = -EIO;
 		}
+	
 	}
 	if (req->cb && req->cb->io_done) {
 		req->expiry = jiffies + msecs_to_jiffies(10000);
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/lib/libxdma.c
```diff
@@ -1437,7 +1437,9 @@ static int xdma_process_requests(struct xdma_engine *engine,
 		timeout = wait_event_interruptible_timeout(req->arbtr_wait,
 				req->sw_desc_cnt == req->desc_completed,
 				msecs_to_jiffies(10000)); /* 10sec timeout */
-		if (timeout == 0) {
+
+		/* Check for timeouts or errors */
+		if (timeout == 0 || timeout < 0) {
```
houlz0507: Could we print different dmesg for ctrl-c case?

pranjalv-xilinx: Yes I can make that change. can we say "Request aborted with error code = "

houlz0507: Sounds great to me.

src/runtime_src/core/pcie/driver/linux/xocl/lib/libxdma.c
```diff
@@ -1437,11 +1437,18 @@ static int xdma_process_requests(struct xdma_engine *engine,
...
+
+		/* Check for timeouts or errors */
 		if (timeout == 0) {
 			pr_err("Request completion timeout\n");
 			engine_reg_dump(engine);
 			rv = -EIO;
+		} else if (timeout < 0) {
+			pr_err("Request aborted with error = %d\n", timeout);
+			engine_reg_dump(engine);
+			rv = -EIO;
```
houlz0507: I still have concern about this. What if the hardware completes the request after user canceled it (timeout < 0)? Will that cause any problem?
Maybe we should change wait_event_interruptible_timeout to wait_event_timeout which does not allow user cancel the request.

pranjalv-xilinx: if hardware completes after user cancels it  the request will still be completed with ERROR. 

 I do not see anything bad happening  in terms of corruption or panic. If it is acceptable to wait for 10 seconds after user cancels then we can make it non-interruptible.

houlz0507: I think in this case the current request returns error. Then the hardware completes. Do you mean this hardware completion will be ignored gracefully and will not impact next request?

pranjalv-xilinx: I see. yes it can impact next request because completion for current request will "look" like completion for next request which is not true.  Although in synchronous transfers it would fine but in asynchronous transfers it can impact next request. 

You suggest I should make it  non-interruptible ?

karenx-xilinx:
I think either a copy of the request somehow have to be kept at the driver until the h/w completion comes back or do not allow cancel.

src/runtime_src/core/pcie/driver/linux/xocl/lib/libxdma.c
houlz0507: I suggest to just make it non-interruptible.

karenx-xilinx: The simplest is to not allow cancel as Lizhi suggested.
Otherwise, we need to keep the # descriptor used for the cancelled request (and also  still maintain the same request order) in case the request eventually get completed.

pranjalv-xilinx:
Made waiting for completion uninterruptible.

PR #3054, opened 2020-03-19
 CR-1057259,CR-1057180  segfault if user doesn't call clRelease*

hackwa:
Moving getXrtIniSettings to plugin constructor makes it safe to get these values


abalasa:
retest this please

PR #3067, opened 2020-03-21
Changed CL_DEVICE_MAX_MEM_ALLOC_SIZE from 512Mb to 4GB.

uday610:
A small fix for hardcoded CL_DEVICE_MAX_MEM_ALLOC_SIZE that was hardcoded to 512 since inception.

### DIFF
--- a/src/runtime_src/xocl/api/clGetDeviceInfo.cpp
+++ b/src/runtime_src/xocl/api/clGetDeviceInfo.cpp
@@ -129,7 +129,7 @@ clGetDeviceInfo(cl_device_id   device,
   case CL_DEVICE_MAX_MEM_ALLOC_SIZE:
     buffer.as<cl_ulong>() =
 #ifdef __x86_64__
-      4*1024*1024*1024; // 4GB
+      4ULL *1024*1024*1024; // 4GB
 #else
       128*1024*1024; //128 MB
 #endif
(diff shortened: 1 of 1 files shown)


src/runtime_src/xocl/api/clGetDeviceInfo.cpp
```diff
@@ -129,7 +129,7 @@ clGetDeviceInfo(cl_device_id   device,
   case CL_DEVICE_MAX_MEM_ALLOC_SIZE:
     buffer.as<cl_ulong>() =
 #ifdef __x86_64__
-      512*1024*1024; //512 MB
+      4*1024*1024*1024; // 4GB
```
keryell: Is it a real limitation somewhere or it can be changed to whatever we want?
I am surprised by these raw constants in the middle of some code instead of some global symbolic constants. Curious...

stsoe: The origin of these constants can be traced back to original development coming from CTO org.  In this case I actually have no problem with the raw constants which are 100% self explanatory.   I do however have a problem with the value being hard code here as opposed to data driven based on device.

stsoe:
ok to test

src/runtime_src/xocl/api/clGetDeviceInfo.cpp
keryell: I plead not guilty on this origin. I am too young. :-)

My question was more: why 4 GiB? Actually most of this file is the same with some hard-coded values...

maxzhen:
retest this please

PR #3069, opened 2020-03-22
misc fixes for flashing shell and SC firmware through driver

maxzhen:
Issues fixed:
1. not able to flash SC firmware through when board is running golden image
2. failed to wait long enough for SC firmware erase to finish
3. can't flash shell after hot reset
Also did some code cleanup.

Tested with xbmgmt flash command to flash shell and SC firmware when board is in golden or non-golden image.


abalasa:
retest this please

PR #3073, opened 2020-03-23
No need to lock device in clGetDeviceInfo

stsoe:
Related to #3053.  Device info is cache when the platform is created
and all devices are probed.


abalasa:
retest this please

PR #3081, opened 2020-03-25
unify kds for for xocl driver

mamin506:
Beta version.
1. Enhance common driver code directory "core/common/drv"
2. Modular KDS
3. CU controller subdev
4. Create FIFO PLram CU subdev
5. Basic configure for configure and start ERT command

### DIFF
--- a/src/CMake/dkms.cmake
+++ b/src/CMake/dkms.cmake
@@ -61,6 +61,7 @@ SET (XRT_DKMS_DRIVER_SRCS
   xocl/userpf/xocl_ioctl.c
   xocl/userpf/xocl_sysfs.c
   xocl/userpf/xocl_drv.c
+  xocl/userpf/xocl_kds.c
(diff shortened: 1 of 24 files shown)


mamin506:
Addition changes:
1. Add ECHO macro in cu_plram.c. 'Echo' means response without touching hardware. It would be useful when we want to know the pure software overhead. This would be configurable by a configure command in the future.
2. Optimized pending queue and running queue in cu_ctrl.c.
3. Optimized cu_plram_check().

IOPS test with XRT application: 
- The finial non echo test could reach 500K IOPS. (wo/ 3, the IOPS is 400K)
- The xclExecBuf() could accept 850K commands/sec. (with old KDS, it is 300K)

Submit speed 850K > 500K, so, maybe it is still possible to improve the finial IOPS.

mamin506:
@sonals  Please review the License for each file.

PR #3083, opened 2020-03-25
XRT-341 memory resource definition conflict

chienwei-lan:
Create a dedicated XOCL_DEVINFO_XMC_U2 for U2 platform


chienwei-lan:
U200 also support XMC scaling feature

PR #3090, opened 2020-03-26
Fix xbutil validate error with HAL Profiling turned on

IshitaGhosh:
When HAL profiling is enabled (turn on compile time macro and add xrt_profile=true in xrt.ini), "xbutil validate" hits "undefined symbol" error for HAL APIs like xclAllocBO/xclRead etc. while loading xdp_hal_plugin library when Python tests are launched from "xbutil validate". This happens because, in Python layer when "xrt_core" library is loaded it was using "lazy loading" approach. All the symbols were not resolved unless used. This caused undefined symbols error while dynamically loading xdp_hal_plugin library which expects some HAL APIs to be available from xrt_core.
To fix this, RTLD_GLOBAL mode is set during loading of xrt_core.

This PR contains another change to update paths used in "load_xdp_pluging_library" to be compatible for Windows also. This change is not related to the xbutil validate issue.


src/runtime_src/core/pcie/linux/plugin/xdp/hal_profile.cpp
```diff
@@ -386,36 +402,33 @@ void load_xdp_plugin_library(HalPluginConfig* )
...
-    dlerror();
-    cb = cb_func_type(reinterpret_cast<cb_load_func_type>(dlsym(handle, cb_func_name.c_str())));
-    if(dlerror() != NULL) { // check if dlsym was successful
+    xrt_core::dlerror();
+    cb = cb_func_type(reinterpret_cast<cb_load_func_type>(xrt_core::dlsym(handle, cb_func_name.c_str())));
+#ifndef _WIN32
+    if(xrt_core::dlerror() != NULL) { // check if dlsym was successful
       cb = nullptr;
     }
+#endif
```
stsoe: shouldn't you just check 'if (!cb)' both windows and linux?

src/runtime_src/core/pcie/linux/plugin/xdp/hal_profile.cpp
```diff
@@ -1,14 +1,16 @@
 #include "plugin/xdp/hal_profile.h"
 #include "core/common/config_reader.h"
 #include "core/common/message.h"
+#include "core/common/dlfcn.h"
+#include<unistd.h>
```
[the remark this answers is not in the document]
IshitaGhosh: Yes, we don't need it here. I had some testcode which needed this. I forgot to remove the unistd.h inclusion. Thanks for pointing it out.

src/runtime_src/core/pcie/linux/plugin/xdp/hal_profile.cpp
```diff
@@ -27,18 +33,27 @@ inline bool isDLL(const bfs::path& path) {
...
 
-static int directoryOrError(const bfs::path& path) {
+boost::filesystem::path
+dllpath(const boost::filesystem::path& root, const std::string& libnm)
+{
+#ifdef _WIN32
+  return root / "bin" / (libnm + ".dll");
+#else
+  return root / "lib" / ("lib" + libnm + ".so");
+#endif
```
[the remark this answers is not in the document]
IshitaGhosh: dllExt gives the extension only. Path for the library is different starting from $XILINX_XRT. For Linux it is in $XILINX_XRT/lib. For Windows, it is in $XILINX_XRT/bin.
As anyway, we have to construct the path I used hardcoded string for the extension also.

src/runtime_src/core/pcie/linux/plugin/xdp/hal_profile.cpp
IshitaGhosh: I saw that for Windows, xrt_core::dlerror returns "". So, on Windows (xrt_core::dlerror() !=NULL) will be true resulting in cb = nullptr, even if the xrt_core::dlsym succeeded.
Shall I add change in xrt_core::dlerror ?

stsoe: No, my point was that you should check the return value of xrt_core:dlsym; it it is nullptr, then it failed, same on both linux and and windows.  If the dlsym returns nullptr, then check the error by calling dlerror.

IshitaGhosh: okay. I have updated.

IshitaGhosh:
retest this please

PR #3094, opened 2020-03-26
03_loopback testcase updated with kernel API

uday610:
03_loopback testcase updated with kernel API

### DIFF
--- a/tests/xrt/03_loopback/main.cpp
+++ b/tests/xrt/03_loopback/main.cpp
@@ -23,22 +23,18 @@
 #include <cstring>
 #include <sys/mman.h>
 
-// driver includes
-#include "ert.h"
+#include "experimental/xrt_kernel.h"
+
 
 // host_src includes
 #include "xclhal2.h"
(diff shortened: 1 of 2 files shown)


tests/xrt/03_loopback/main.cpp
```diff
@@ -152,99 +148,90 @@ int main(int argc, char** argv)
...
+       if (std::memcmp(bo2, bo1, DATA_SIZE)) {
             std::cout << "FAILED TEST\n";
             std::cout << "Value read back does not match value written\n";
             return 1;
-        }
+       }
 
         //Clean up stuff
         munmap(bo1, DATA_SIZE);
         munmap(bo2, DATA_SIZE);
```
stsoe: If you replace 'munmap' with 'xclUnmapBO' then this test will almost certainly be working on windows also.

uday610: Done, Thanks

stsoe:
ok to test

PR #3103, opened 2020-03-27
XRT-348 XRT AIE GMIO API

larry9523:
Implement XRT GRAPH API Sync BO between GMIO<==>AIE

### DIFF
--- a/src/runtime_src/core/edge/user/aie/aie.h
+++ b/src/runtime_src/core/edge/user/aie/aie.h
@@ -20,22 +20,54 @@
 #define xrt_core_edge_user_aie_h
 
 #include <vector>
+#include <queue>
 
+#include "core/edge/common/aie_parser.h"
 extern "C" {
 #include <xaiengine.h>
 }
 
 namespace zynqaie {
 
+#define BD_HIGH_ADDR_MASK               0xFF00000000L
+#define BD_LOW_ADDR_MASK                0xFFFFFFFF
+
+#define GET_BD_HIGH_ADDR(addr)          (((addr) & BD_HIGH_ADDR_MASK) >> 32)
+#define GET_BD_LOW_ADDR(addr)           ((addr) & BD_LOW_ADDR_MASK)
+
+struct BD {
+    uint16_t bd_num;
+    uint16_t addr_high;
+    uint32_t addr_low;
+};
+
+struct DMAChannel {
+    std::queue<BD> idle_bds;
+    std::queue<BD> pend_bds;
+};
+
+struct ShimDMA {
+    XAieDma_Shim handle;
+    DMAChannel dma_chan[XAIEDMA_SHIM_MAX_NUM_CHANNELS];
+    bool configured;
+};
+
+
 class Aie {
 public:
+    using gmio_type = xrt_core::edge::aie::gmio_type;
+
     ~Aie();
     Aie();
 
     std::vector<XAieGbl_Tile> tileArray;  // Tile Array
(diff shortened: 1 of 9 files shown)


src/runtime_src/core/edge/user/aie/aie.h
```diff
@@ -20,22 +20,54 @@
...
 #include <xaiengine.h>
 }
 
 namespace zynqaie {
 
+#define BD_HIGH_ADDR_MASK               0xFF00000000L
+#define BD_LOW_ADDR_MASK                0xFFFFFFFF
+
+#define GET_BD_HIGH_ADDR(addr)          (((addr) & BD_HIGH_ADDR_MASK) >> 32)
+#define GET_BD_LOW_ADDR(addr)           ((addr) & BD_LOW_ADDR_MASK)
```
stsoe: I am not positive how you imagine these being used.   I am wary of # defines which are controlled by preprocessor and not scoped to namespace.  It is better if these are inline functions, something like:
```
inline uint64_t
get_bd_high_addr(uint64_t addr)
{
    constexpr uint64_t hi_mask =  0xFF00000000L
    return (addr & hi_mask) >> 32;
}
```
Now if these functions are used only in graph.cpp, then maybe just declare and define them there?

src/runtime_src/core/edge/user/aie/aie.h
```diff
@@ -20,22 +20,54 @@
...
+
+struct BD {
+    uint16_t bd_num;
+    uint16_t addr_high;
+    uint32_t addr_low;
+};
+
+struct DMAChannel {
+    std::queue<BD> idle_bds;
+    std::queue<BD> pend_bds;
```
stsoe: queues are last resort.  I think I see that pend_bds must be a FIFO, so this is probably right.  But idle_bds, does the order matter?  If not make it a std::vector and use push(), back() and pop().

src/runtime_src/core/edge/user/aie/graph.cpp
```diff
@@ -421,6 +504,13 @@ xrtGraphUpdateRTP(xrtGraphHandle ghdl, const char* port, const char* buffer, siz
   graph->update_rtp(port, buffer, size);
 }
 
+void
+xrtSyncBOAIE(xrtGraphHandle ghdl, unsigned bo, const char *dmaID, enum xclBOSyncDirection dir, size_t size, size_t offset)
```
stsoe: This is just my pet peeve, I have never liked bare 'unsigned'.   Please use 'unsigned int' to be explicit.

src/runtime_src/core/edge/user/aie/graph.cpp
```diff
@@ -598,3 +688,20 @@ xrtGraphUpdateRTP(xrtGraphHandle ghdl, const char* port, const char* buffer, siz
     return -1;
   }
 }
+
+int
+xrtSyncBOAIE(xrtGraphHandle ghdl, unsigned bo, const char *dmaID, enum xclBOSyncDirection dir, size_t size, size_t offset)
```
stsoe: unsigned int :-)

src/runtime_src/core/edge/user/aie/graph.cpp
```diff
@@ -310,6 +312,87 @@ update_rtp(const char* port, const char* buffer, size_t size)
     }
 }
 
+void
+graph_type::
+sync_bo(unsigned bo, const char *dmaID, enum xclBOSyncDirection dir, size_t size, size_t offset)
```
stsoe: I don't see 'dir' used any where?

PR #3104, opened 2020-03-28
Add ImportBO and ExportBO in Windows

igorshr:
The fix implemented for CR-1040791. The Windows XRT doesn't support these functions and returns "NON SUPPORTED" status now.


arpitxilinx:
ok to test

maxzhen:
ok to test

PR #3115, opened 2020-03-30
XRT-371 Provide --primary & --secondary switches in xbmgmt flash --shell

chvamshi-xilinx:
Currently, xbmgmt takes one argument --path for the file which contains PRIMARY_MCS. There is no way to provide a different file which has SECONDARY_MCS for xbmgmt.

This PR is providing --primary/–secondary switches in xbmgmt to support this.

--path/--primary switches are identical now. Made two switches identical for the backward compatibilty

Added mail regarding this change in following XRT Task
http://jira.xilinx.com/browse/XRT-371

Testing:
validated on U.2 Shell by giving --primary and --secondary options.
Checked error conditions in all following cases
- Giving both --path and --primary options
- Giving no --path & --primary options
- Giving both primary & secondary options
- Giving only --path option

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbmgmt/cmd_flash.cpp
+++ b/src/runtime_src/core/pcie/tools/xbmgmt/cmd_flash.cpp
@@ -40,7 +40,7 @@ const char *subCmdFlashUsage =
     "--update [--shell name [--id id]] [--card bdf] [--force]\n"
     "--factory_reset [--card bdf] [--force]\n\n"
     "Experts only:\n"
-    "--shell --path file --card bdf [--type flash_type]\n"
+    "--shell --path file --primary primary_file [--secondary secondary_file] --card bdf [--type flash_type]\n"
     "--sc_firmware --path file --card bdf";
 
 #define fmt_str		"    "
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/tools/xbmgmt/cmd_flash.cpp
```diff
@@ -40,7 +40,7 @@ const char *subCmdFlashUsage =
     "--update [--shell name [--id id]] [--card bdf] [--force]\n"
     "--factory_reset [--card bdf] [--force]\n\n"
     "Experts only:\n"
-    "--shell --path file --card bdf [--type flash_type]\n"
+    "--shell --path file --primary primary_file [--secondary secondary_file] --card bdf [--type flash_type]\n"
```
maxzhen: We can stop showing --path option in the help msg for this release and remove it completely in the future release.

src/runtime_src/core/pcie/tools/xbmgmt/cmd_flash.cpp
```diff
@@ -690,17 +694,44 @@ static int shell(int argc, char *argv[])
             file = std::string(optarg);
             break;
```
maxzhen: don't you want to remove these lines so make 1 and 2 the same? thus you have one less case to worry about later when you validate the arguments...

src/runtime_src/core/pcie/tools/xbmgmt/cmd_flash.cpp
chvamshi-xilinx: Makes sense @maxzhen . I will remove --path option from help

src/runtime_src/core/pcie/tools/xbmgmt/cmd_flash.cpp
chvamshi-xilinx: True, but we cannot find out the issues if both --path/--primary is given, but thats a corner case which we can ignore at this point of time as we have removed that --path from help also.. I will make changes.

chvamshi-xilinx:
@maxzhen , Made both the changes. 
Following error message is not updated. We can update once we remove --path option from tool.
"--primary/--path switch is not provided."

PR #3120, opened 2020-03-31
Support for reporting Debug IP status on Windows in xbutil2

IshitaGhosh:
This PR adds support for :
1. Listing the Debug IPs present in xclbin loaded on device and also reporting the counter values using "xbutil examine -r debug-ip-status" on Windows.
2. Generate Json format output for debug-ip-status
3. Enable filtering for certain Debug IP types using "-e[--element] <aim> <am> <asm> <lapc> <spc>" option

However, please note that, on Linux this new "examine -r debug-ip-status" (xbutil2)  cannot give any Debug IP Data. The already existing "xbutil status <--ipname>" should be used on Linux. This issue can be fixed using a CR.


src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -2041,5 +2041,6 @@ xclCloseIPInterruptNotify(xclDeviceHandle handle, int fd)
 void
 xclGetDebugIpLayout(xclDeviceHandle hdl, char* buffer, size_t size, size_t* size_ret)
 {
+  *size_ret = 0;
```
rozumx, line 2044: This code assumes that the `size_ret` pointer will always be valid.  To prevent future issues, could it first be checked for a null pointer.  For example:
```
if (size_ret != nullptr)
  *size_ret = 0;
```

IshitaGhosh: Added the change.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  if(0 == amResults.NumSlots) {
+    return;
+  }
+
+  _output << "\nAccelerator Monitor Counters (hex values are cycle count)\n";
+
+  auto col1 = std::max(cuNameMaxStrLen[ACCEL_MONITOR], strlen("Compute Unit")) + 4;
+
+  auto iosguard = xrt_core::utils::ios_restore(_output);
+  _output << std::left
```
rozumx, line 690: Instead of using the iostream formatting (which can get rather obfuscated), could you use boost::format.  For example:
```
boost::format header(std::to_string("%")+std::to_string(col1)+"s %8s  %16s  %16s  %16s  %16s  %16s  %16s  %16s);
_output << header % "Compute Unit"  % "Ends" % "Starts" % "Max Parallel Itr" % "Execution" % "Memory Stall" % "Pipe Stall" % "Stream Stall" % "Min Exec" % "Max Exec";

```
And the contents can be formatted similarly.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  getStreamName(dbgIpInfo->m_type, dbgIpName, masterName, slaveName);
+  cuNames[AXI_STREAM_MONITOR].emplace_back(masterName);
+  portNames[AXI_STREAM_MONITOR].emplace_back(slaveName);
+
+  size_t size = 0; // The amount of data read from the hardware
+
+  // Fill up the portions of the return struct that are known by the runtime
+
+  // Fill up the return structure with the values read from the hardware
+  uint64_t asm_offsets[] = {
```
rozumx, line 744: Please make this  a` static const` value.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  if(0 == asmResults.NumSlots) {
+    return;
+  }
+  _output << "\nAXI Stream Monitor Counters\n";
+  auto col1 = std::max(cuNameMaxStrLen[AXI_STREAM_MONITOR], strlen("Stream Master")) + 4;
+  auto col2 = std::max(portNameMaxStrLen[AXI_STREAM_MONITOR], strlen("Stream Slave"));
+
+  std::streamsize ss = _output.precision();
+  auto iosguard = xrt_core::utils::ios_restore(_output);
+  _output << std::left
```
rozumx, line 805: Please format this using boost::format.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  // Get Cu and Port Name
+  std::string cuName;
+  std::string portName;
+  getCuNamePortName(dbgIpInfo->m_type, dbgIpName, cuName, portName);
+  cuNames[LAPC].emplace_back(cuName);
+  portNames[LAPC].emplace_back(portName);
+
+  size_t size = 0;
+
+  uint64_t statusRegisters[] = {
```
rozumx, line 854: This should be a `static const` value.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1305 @@
...
+#ifndef _WIN32
+#pragma GCC diagnostic pop
+#endif
+  }
+
+  lapcResults.OverallStatus[index]      = currData[XLAPC_OVERALL_STATUS];
+  std::copy(currData+XLAPC_CUMULATIVE_STATUS_0, currData+XLAPC_SNAPSHOT_STATUS_0, lapcResults.CumulativeStatus[index]);
+  std::copy(currData+XLAPC_SNAPSHOT_STATUS_0, currData+XLAPC_STATUS_PER_SLOT, lapcResults.SnapshotStatus[index]);
+
+  (void)size;
```
rozumx, line 881: No sure what this does.  Is this needed?

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  bool violations_found = false;
+  bool invalid_codes = false;
+  _output << "\nLight Weight AXI Protocol Checkers codes \n";
+  auto col1 = std::max(cuNameMaxStrLen[LAPC], strlen("CU Name")) + 4;
+  auto col2 = std::max(portNameMaxStrLen[LAPC], strlen("AXI Portname"));
+
+  for (size_t i = 0; i < lapcResults.NumSlots; ++i) {
+    if (!xclAXICheckerCodes::isValidAXICheckerCodes(lapcResults.OverallStatus[i],
+                lapcResults.SnapshotStatus[i], lapcResults.CumulativeStatus[i])) {
+      _output << "CU Name: " << cuNames[LAPC][i] << " AXI Port: " << portNames[LAPC][i] << "\n";
```
rozumx, line 904: Please use boost::format.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  }
+  if (!violations_found && !invalid_codes)
+    _output << "No AXI violations found \n";
+
+  if (violations_found && /*aVerbose &&*/ !invalid_codes) {
+    std::ofstream saveFormat;
+    saveFormat.copyfmt(_output);
+
+    _output << "\n";
+    _output << std::left
```
rozumx, line 932: Yep, boost format :-)

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1305 @@
...
+    // Restore formatting
+    _output.copyfmt(saveFormat);
+  }
+}
+
+void 
+DebugIpStatusCollector::readSPChecker(debug_ip_data* dbgIpInfo)
+{
+  // index in results is debugIpNum
+  uint64_t index = debugIpNum[AXI_STREAM_PROTOCOL_CHECKER];
```
rozumx, line 970: Being that this value never changes.  It should be `const`.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1305 @@
...
+            << "  " << "0x" << std::setw(14) << amResults.CuMaxExecCycles[i]
+            << std::dec << std::endl;
+  }
+}
+
+void 
+DebugIpStatusCollector::readASMCounter(debug_ip_data* dbgIpInfo)
+{
+  // index in results is debugIpNum
+  uint64_t index = debugIpNum[AXI_STREAM_MONITOR];
```
rozumx, line 723: Being that this value never changes.  It should be `const`.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  return;
+
+}
+
+
+void 
+DebugIpStatusCollector::readAMCounter(debug_ip_data* dbgIpInfo)
+{
+  // index in results is debugIpNum
+  uint64_t index = debugIpNum[ACCEL_MONITOR];
```
rozumx, line 532: Being that this value never changes.  It should be `const`.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  if (violations_found && /*aVerbose &&*/ !invalid_codes) {
+    auto col1 = std::max(cuNameMaxStrLen[AXI_STREAM_PROTOCOL_CHECKER], strlen("CU Name")) + 4;
+    auto col2 = std::max(portNameMaxStrLen[AXI_STREAM_PROTOCOL_CHECKER], strlen("AXI Portname"));
+
+    std::ofstream saveFormat;
+    saveFormat.copyfmt(_output);
+
+    _output << "\n";
+    _output << std::left
+        << std::setw(col1) << "CU Name"
```
rozumx, line 1067: Boost format is your friend :)

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1305 @@
...
+
+  boost::property_tree::ptree aim_pt;
+
+  for(size_t i = 0; i < aimResults.NumSlots; ++i) {
+    boost::property_tree::ptree entry;
+    std::string entryStr = "AIM " + std::to_string(i);
+
+    entry.put("Region or CU", cuNames[AXI_MM_MONITOR][i]);
+    entry.put("Type or Port", portNames[AXI_MM_MONITOR][i]);
+    entry.put("Write kBytes", static_cast<double>(aimResults.WriteBytes[i])/1000.0);
```
rozumx, line 1142: Thank you for adding the units.  It helps the clients parsing this data greatly.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1305 @@
...
+    entry.put("Write Trans",  aimResults.WriteTranx[i]);
+    entry.put("Read kBytes",  static_cast<double>(aimResults.ReadBytes[i])/1000.0);
+    entry.put("Read Tranx",   aimResults.ReadTranx[i]);
+    entry.put("Outstanding Cnt", aimResults.OutStandCnts[i]);
+    entry.put("Last Wr Addr", aimResults.LastWriteAddr[i]);
+    entry.put("Last Wr Data", aimResults.LastWriteData[i]);
+    entry.put("Last Rd Addr", aimResults.LastReadAddr[i]);
+    entry.put("Last Rd Data", aimResults.LastReadData[i]);
+
+    aim_pt.add_child(entryStr.c_str(), entry);
```
rozumx, line 1152: Q: Can the entryString contain either a period '.' or forward slash '/'?
Not asking to change this value.  Asking for we need to determine a hierarchy search character that isn't being used by a key.

IshitaGhosh: No, the entryStr here willnot have any period '.' or forward slash '/'.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
IshitaGhosh: "size" variable here stores the total number of bytes read from hardware returned by xclRead. But we are not using it actually. However, Coverity can complain about unused return value or unused variable, thats why I have added it.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
IshitaGhosh: Thanks!

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1305 @@
...
+  // Get the size of full debug_ip_layout
+  xclGetDebugIpLayout(handle, nullptr, sz1, &sectionSz);
+  if(sectionSz == 0) {
+    std::cout << "INFO: Failed to find any Debug IPs on the platform. "
+              << "Ensure that a valid bitstream with debug IPs (AIM, LAPC) is successfully downloaded. \n"
+              << std::endl;
+    return;
+  }
+  // Allocate buffer to retrieve debug_ip_layout information from loaded xclbin
+  map = reinterpret_cast<debug_ip_layout*>(new char[sectionSz]);
```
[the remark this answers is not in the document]
IshitaGhosh: Done

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
IshitaGhosh: I added use of boost::format while printing the LAPC violations. At this current line, there is no formatting used. So, I am not adding any change here.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
IshitaGhosh: boost::format is really nice and easy to use and read. Thanks for your suggestion. I have added change at all the required places.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1221 @@
...
+    , aimResults{0}
+    , asmResults{0}
+    , amResults{0}
+    , lapcResults{0}
+    , spcResults{0}
+{
+  // By default, enable status collection for all Debug IP types
+  for(uint64_t i = 0; i < maxDebugIpType ; ++i) {
+    debugIpOpt[i] = true;
+  }
```
stsoe, line 145: why not initialize in the constructor initializer list?

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1221 @@
...
+    if(0 == strcmp(element, "aim")) {
+      debugIpOpt[AXI_MM_MONITOR] = true;
+    } else if(0 == strcmp(element, "am")) {
+      debugIpOpt[ACCEL_MONITOR] = true;
+    } else if(0 == strcmp(element, "asm")) {
+      debugIpOpt[AXI_STREAM_MONITOR] = true;
+    } else if(0 == strcmp(element, "lapc")) {
+      debugIpOpt[LAPC] = true;
+    } else if(0 == strcmp(element, "spc")) {
+      debugIpOpt[AXI_STREAM_PROTOCOL_CHECKER] = true;
```
stsoe, line 239: why not use std::string::operator== ?

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1221 @@
...
+#ifndef _WIN32
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
+  size += xclRead(handle, XCL_ADDR_SPACE_DEVICE_PERFMON,
+                    dbgIpInfo->m_base_address + XAIM_SAMPLE_OFFSET,
+                    &sampleInterval, sizeof(uint32_t));
+#ifndef _WIN32
+#pragma GCC diagnostic pop
+#endif
```
stsoe, line 414: This pattern makes the code look ugly.  Please wrap xclRead in an inline function in unnamed namespace at beginning of this file, then call from all places below.  Only the inline function need to have the pragmas and code will be cleaner and easier to read.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1221 @@
...
+  // Fill up string with 128 characters (padded with null characters)
+  dbgIpName.assign(dbgIpInfo->m_name, sizeof(dbgIpInfo->m_name));
+  // Strip away any extraneous null characters
+  dbgIpName.assign(dbgIpName.c_str());
+
+  // Get Master and Slave Port Name
+  std::string masterName;
+  std::string slaveName;
+  getStreamName(dbgIpInfo->m_type, dbgIpName, masterName, slaveName);
+  cuNames[AXI_STREAM_MONITOR].emplace_back(masterName);
```
stsoe, line 693: I think you mean to 'move' masterName and slaveName into the vectors, so do it this way
```
{
  std::string masterName;
  std::string slaveName;
  getStreamName(dbgIpInfo->m_type, dbgIpName, masterName, slaveName);
  cuNames[AXI_STREAM_MONITOR].emplace_back(std::move(masterName));
  portNames[AXI_STREAM_MONITOR].emplace_back(std::move(slaveName));
}
```
The scope ensures that masterName and slaveName are not used later, and the std::move creates a rvalue from the lvalue strings so that the string content is moved during emplace rather than copied.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1221 @@
...
+  dbgIpName.assign(dbgIpInfo->m_name, sizeof(dbgIpInfo->m_name));
+  // Strip away any extraneous null characters
+  dbgIpName.assign(dbgIpName.c_str());
+
+  // Get Cu and Port Name
+  std::string cuName;
+  std::string portName;
+  getCuNamePortName(dbgIpInfo->m_type, dbgIpName, cuName, portName);
+  cuNames[LAPC].emplace_back(cuName);
+  portNames[LAPC].emplace_back(portName);
```
stsoe, line 793: same comment as earlier

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
stsoe: if Coverity complains, I think the right place for void is with the calls to the (wrapped) xclRead, e.g.
`(void) xclReadWrap(...);`
No need for size variable.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1221 @@
...
+  // Strip away any extraneous null characters
+  dbgIpName.assign(dbgIpName.c_str());
+
+  // Get Cu and Port Name
+  std::string cuName;
+  std::string portName;
+  getCuNamePortName(dbgIpInfo->m_type, dbgIpName, cuName, portName);
+  cuNames[AXI_STREAM_PROTOCOL_CHECKER].emplace_back(cuName);
+  portNames[AXI_STREAM_PROTOCOL_CHECKER].emplace_back(portName);
+
```
stsoe, line 913: as ealier

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -0,0 +1,1221 @@
...
+#endif
+  size += xclRead(handle, XCL_ADDR_SPACE_DEVICE_CHECKER,
+              dbgIpInfo->m_base_address + XSPC_PC_ASSERTED_OFFSET,
+              &pc_asserted, sizeof(uint32_t));
+  size += xclRead(handle, XCL_ADDR_SPACE_DEVICE_CHECKER,
+              dbgIpInfo->m_base_address + XSPC_CURRENT_PC_OFFSET,
+              &current_pc, sizeof(uint32_t));
+  size += xclRead(handle, XCL_ADDR_SPACE_DEVICE_CHECKER,
+              dbgIpInfo->m_base_address + XSPC_SNAPSHOT_PC_OFFSET,
+              &snapshot_pc, sizeof(uint32_t));
```
stsoe, line 933: just (void) the function call itself and get rid of size

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
IshitaGhosh: If constructor initializer list is used, then only the first element gets initialized to "true" and rest of the elements remain "false". I had tried this earlier and because of the issue had to set all the elements to "true" in the current manner.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
IshitaGhosh: One of the operand here is always a const char* (i.e. "aim", "am" etc.). To use std::string::operator==, a string object would need to be created for those "const char*" s. So, it does not give much of an advantage.
Do you see any issue with the current implementation? If there is an issue, then I could definitely use std::string::operator== here.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
stsoe: ok. you are right indeed.   then how about 
`std::fill(debugIpOpt, debugIpOpt + maxDebugIpType, true);`

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
stsoe: I am wary of out-smarting the standard library.  It's better to use std::string::operator== and let it do the most efficient comparison.  Incidentally the string comparison operator is overloaded to take second argument as a const char*, so under the hood it may do what you are doing, but we don't know for sure and we shouldn't.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
IshitaGhosh: I have added the wrapper. The code now looks so much better. Thanks for the suggestion.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
IshitaGhosh: Done

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
IshitaGhosh: That works. I have added the change.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
IshitaGhosh: okay. Done.

PR #3123, opened 2020-03-31
edge: Fix drm_mode.h include

matwey:
Original commit

    01178203434a ("[Edge] Remove drm.h and drm_mode.h (#2614)")

says that drm_mode.h has to come from libdrm, so change the path
accordingly to fix the build.


maxzhen:
ok to test

maxzhen:
retest this please

PR #3125, opened 2020-03-31
XRT-375 Multiple completion interrupts within descriptor chain cause failures

pranjalv-xilinx:
Two changes:

1. Increase default size of descriptor chain to 32
2. Disable setting COMPLETION bit in multiple descriptors within linked chain of descriptors. Only the last descriptor will generate interrupt for completion.

There is issue with descriptor completion count register in XDMA when in a single chain of linked descriptors multiple descriptors have completion bit set. This causes increased interrupts but when driver goes and reads the completion count register the engine has moved further and completed even more descriptors. There are instances of stale interrupts during next engine run and looks like even instances of stale count value being returned in register.

The above has been documented well  "XDMA-1071 MSI-X Interrupt firing prematurely".


pranjalv-xilinx:
retest this please

PR #3126, opened 2020-03-31
CR-1045886: Add XCL_STREAM flags for consistency with CL_MEM flags

stsoe:
XCL_STREAM flags are kernel centric to match CL_MEM flags.


uday610:
@stsoe , this will break all existing test cases? Right ?

stsoe:
@uday610 I left the old flags and marked them deprecated.  If they are used, the compiler will issue a warning during compilation.

PR #3128, opened 2020-03-31
xrt-379 need to recreate firewall subdevice after programming PLP

houlz0507:


src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
```diff
@@ -2931,7 +2939,13 @@ static int xmc_offline(struct platform_device *pdev)
...
 static int xmc_online(struct platform_device *pdev)
 {
-	return xmc_access(pdev, XOCL_XMC_FREE);
+	int ret;
+
+	ret = xmc_access(pdev, XOCL_XMC_FREE);
+	if (ret && ret != -ENODEV)
+		return ret;
+
+	return 0;
```
xdavidz, line 2948: could be changed to: return (ret && ret != -ENODEV) ? ret : 0;

PR #3135, opened 2020-04-02
xbutil2/xbmgmt2 : Removed classic command signatures / help

rozumx:
With the introduction of xbutil and xbmgmt loader, there isn't a need to implemented the "classic" xbutil/xbmgmt commands in the new version of these applications.  This is because the loader is smart enough to select which version to use.

Work Done
---------
+ Removed "most" the subcommands that will not be supported in xbmgmt2/xbutil2 in 2020.1
+ A few remaining subcommands (e.g., flash and query) will remain until their functionality has been validated in the examine subcommand.


rozumx:
Retest this please

rozumx:
First build failed with regards to missing boost include files on windows.   Retesting to see if this issue still exists.

PR #3138, opened 2020-04-02
add temperature sampling

pgschuey:


hackwa:
looks good. GUI can pick and choose what to display

PR #3145, opened 2020-04-03
CR1060218: Fix issue with reporting kernel enqueues incorrectly in LOP

jvillarre:
Now tracing the correct event and only marking the start when CL_RUNNING occurs

### DIFF
--- a/src/runtime_src/xocl/api/clEnqueueNDRangeKernel.cpp
+++ b/src/runtime_src/xocl/api/clEnqueueNDRangeKernel.cpp
@@ -421,6 +421,10 @@ clEnqueueNDRangeKernel(cl_command_queue command_queue,
   xocl::profile::set_event_action(umEvent.get(),xocl::profile::action_ndrange_migrate,mEvent,kernel);
   xocl::appdebug::set_event_action(umEvent.get(),xocl::appdebug::action_ndrange_migrate,mEvent,kernel);
 
+#ifndef _WIN32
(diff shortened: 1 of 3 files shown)


src/runtime_src/xocl/api/clEnqueueNDRangeKernel.cpp
```diff
@@ -420,9 +420,6 @@ clEnqueueNDRangeKernel(cl_command_queue command_queue,
   xocl::enqueue::set_event_action(umEvent.get(),xocl::enqueue::action_ndrange_migrate,mEvent,kernel);
   xocl::profile::set_event_action(umEvent.get(),xocl::profile::action_ndrange_migrate,mEvent,kernel);
   xocl::appdebug::set_event_action(umEvent.get(),xocl::appdebug::action_ndrange_migrate,mEvent,kernel);
```
hackwa, line 422: this event also needs a callback to record ndrange migrate right before kernel starts

hackwa: In case the buffer is already resident, this event does nothing. lop callback can check that condition and not log it. however, if the buffer isn't resident, this actually does a buffer transfer.

jvillarre: Added callback to catch buffer writes due to ndrange migration in latest commit

PR #3154, opened 2020-04-04
XRT-396 XRT needs not support partition metadata partial overlay

houlz0507:


src/runtime_src/core/pcie/driver/linux/xocl/xocl_subdev.c
```diff
@@ -1426,6 +1455,7 @@ int xocl_subdev_create_prp(xdev_handle_t xdev)
 {
 	int		ret;
 
+	xocl_subdev_destroy_dup(xdev);
```
maxzhen, line 1458: is this for destroying mailbox subdev in blp and keeping mailbox subdev in plp? but the mailbox in plp has not been created yet, right? isn't it created by the next line of code?

houlz0507: PLP Meta-data has been parsed  here. So this line checks if blp sub devices are re-defined in PLP and destroy it. Mailbox is the only case for now. the next line will create all PLP sub devices include PLP mailbox

PR #3170, opened 2020-04-07
flashing shell and update SC firmware via driver by default

maxzhen:
By default, shell and sc firmware update will go through driver. Setting FLASH_VIA_USER environment variable could switch to old code path which will go through mapped PCIE BAR on mgmt PF.


maxzhen:
retest this please

AShivangi:
Does this mean we need to set FLASH_VIA_USER variable on windows everytime we flash?

maxzhen:
> 
> 
> Does this mean we need to set FLASH_VIA_USER variable on windows everytime we flash?

it should fall through, i think, and use existing code path. but i don't know how to test it.

PR #3176, opened 2020-04-08
fix embedded xrt build

larry9523:
Fix embedded XRT build


stsoe:
Hi @matwey .  Apparently embedded petalinux build failed because of the MODULE change.   Here is the error 

ERROR: xrt-202010.2.6.0-r0 do_package_qa: QA Issue: -dev package contains non-symlink .so: xrt-dev path '/work/aarch64-xilinx-linux/xrt/202010.2.6.0-r0/packages-split/xrt-dev/usr/lib/libxdp_lop_plugin.so'
-dev package contains non-symlink .so: xrt-dev path '/work/aarch64-xilinx-linux/xrt/202010.2.6.0-r0/packages-split/xrt-dev/usr/lib/libxdp_hal_plugin.so'
-dev package contains non-symlink .so: xrt-dev path '/work/aarch64-xilinx-linux/xrt/202010.2.6.0-r0/packages-split/xrt-dev/usr/lib/libxdp_debug_plugin.so'
-dev package contains non-symlink .so: xrt-dev path '/work/aarch64-xilinx-linux/xrt/202010.2.6.0-r0/packages-split/xrt-dev/usr/lib/libxdp_appdebug_plugin.so'
-dev package contains non-symlink .so: xrt-dev path '/work/aarch64-xilinx-linux/xrt/202010.2.6.0-r0/packages-split/xrt-dev/usr/lib/libxdp_hal_api_interface_plugin.so' [dev-elf]

matwey:
> Hi @matwey . Apparently embedded petalinux build failed because of the MODULE change. Here is the error
> 
> ERROR: xrt-202010.2.6.0-r0 do_package_qa: QA Issue: -dev package contains non-symlink .so: xrt-dev path '/work/aarch64-xilinx-linux/xrt/202010.2.6.0-r0/packages-split/xrt-dev/usr/lib/libxdp_lop_plugin.so'
> -dev package contains non-symlink .so: xrt-dev path '/work/aarch64-xilinx-linux/xrt/202010.2.6.0-r0/packages-split/xrt-dev/usr/lib/libxdp_hal_plugin.so'
> -dev package contains non-symlink .so: xrt-dev path '/work/aarch64-xilinx-linux/xrt/202010.2.6.0-r0/packages-split/xrt-dev/usr/lib/libxdp_debug_plugin.so'
> -dev package contains non-symlink .so: xrt-dev path '/work/aarch64-xilinx-linux/xrt/202010.2.6.0-r0/packages-split/xrt-dev/usr/lib/libxdp_appdebug_plugin.so'
> -dev package contains non-symlink .so: xrt-dev path '/work/aarch64-xilinx-linux/xrt/202010.2.6.0-r0/packages-split/xrt-dev/usr/lib/libxdp_hal_api_interface_plugin.so' [dev-elf]

Sorry for that. But what is the technical reason to have symlink-ed .so for plugins? I understand when this rule is applied for shared libraries, but here nobody is going to link anything with this plugins.

stsoe:
No worries.   I am not sure, I don't understand the details of the petalinux build.

matwey:
> No worries. I am not sure, I don't understand the details of the petalinux build.

I missed `-dev` suffix in the package name. Do you test that you can use plugins without `xrt-dev` package installed? I suppose this `.so` plugins should go to the main package instead of `xrt-dev` because they are part of runtime.

larry9523:
> > Hi @matwey . Apparently embedded petalinux build failed because of the MODULE change. Here is the error
> > ERROR: xrt-202010.2.6.0-r0 do_package_qa: QA Issue: -dev package contains non-symlink .so: xrt-dev path '/work/aarch64-xilinx-linux/xrt/202010.2.6.0-r0/packages-split/xrt-dev/usr/lib/libxdp_lop_plugin.so'
> > -dev package contains non-symlink .so: xrt-dev path '/work/aarch64-xilinx-linux/xrt/202010.2.6.0-r0/packages-split/xrt-dev/usr/lib/libxdp_hal_plugin.so'
> > -dev package contains non-symlink .so: xrt-dev path '/work/aarch64-xilinx-linux/xrt/202010.2.6.0-r0/packages-split/xrt-dev/usr/lib/libxdp_debug_plugin.so'
> > -dev package contains non-symlink .so: xrt-dev path '/work/aarch64-xilinx-linux/xrt/202010.2.6.0-r0/packages-split/xrt-dev/usr/lib/libxdp_appdebug_plugin.so'
> > -dev package contains non-symlink .so: xrt-dev path '/work/aarch64-xilinx-linux/xrt/202010.2.6.0-r0/packages-split/xrt-dev/usr/lib/libxdp_hal_api_interface_plugin.so' [dev-elf]
> 
> Sorry for that. But what is the technical reason to have symlink-ed .so for plugins? I understand when this rule is applied for shared libraries, but here nobody is going to link anything with this plugins.

Maybe this is related to the QA error?
https://wiki.yoctoproject.org/wiki/Technical_FAQ

Why are .so files in the -dev package instead of the main package for a recipe?
In standard Unix library packaging, non-versioned .so symlinks (e.g. /usr/lib/libgd.so) are intended for development purposes only. At runtime, binaries should be linked to the major-versioned .so file/symlink e.g. /usr/lib/libgd.so.3. This (theoretically) allows multiple major versions of the same library as well as binaries that depend upon each of them to coexist on the same system. If the library is versioned but you have a binary that links to the unversioned .so file, it has almost certainly been linked incorrectly.

Non-symlink .so files on the other hand are sometimes produced and are entirely legal - however these will be picked up in the -dev package in OpenEmbedded simply by virtue of their name, which is almost always not what you want. In this case you can do one of two things:

Fix the build of the library so it gets versioned. This may not always be appropriate, especially not for things like plugins.
Set FILES_${PN}-dev within the recipe so that it does not include ${FILES_SOLIBSDEV}. If the software the recipe is building also produces symlink .so files you'll need to set FILES_${PN}-dev such that those do still get packaged in the -dev package though, or you'll get a package QA warning.

matwey:
@larry9523 this guys are not dynamically linked shared libraries. They are loaded using `dlopen` in code by their filepath, so they do not need to have soname.

larry9523:
> @larry9523 this guys are not dynamically linked shared libraries. They are loaded using `dlopen` in the code by their filepath, so they do not need to have soname.

@matwey Good point. I agree with you that MODULE is more proper for those guys. Let's get the build back and find out a proper way to elimiate QA warning with MODULE. 

https://wiki.yoctoproject.org/wiki/TipsAndTricks/Packaging_Prebuilt_Libraries
Some Background
Libraries in Linux systems are generally versioned so that it is possible to have multiple versions of the same library installed, which eases upgrades and support for older software. For example, suppose that in a versioned library an actual library is called "libfoo.so.1.2", a symbolic link named "libfoo.so.1" points to "libfoo.so.1.2", and a symbolic link named "libfoo.so" points to "libfoo.so.1.2". Given these conditions, when you link a binary against a library, you typically provide the unversioned file name (i.e. -lfoo to the linker). However, the linker follows the symbolic link and actually links against the versioned filename. The unversioned symbolic link is only used at development time. Consequently, the library is packaged along with the headers in the development package ${PN}-dev along with the actual library and versioned symbolic links in ${PN}. Because versioned libraries are far more common than unversioned libraries, the default packaging rules assume versioned libraries.

Yocto Packaging Overview
It follows that packaging an unversioned library requires a bit of work in the recipe. By default, "libfoo.so" gets packaged into ${PN}-dev, which triggers a QA warning that a non-symlink library is in a -dev package, and binaries in the same recipe link to the library in ${PN}-dev, which triggers more QA warnings. To solve this problem, you need to package the unversioned library into ${PN} where it belongs. Following are the abridged default FILES variables in bitbake.conf:

Himanshu-xilinx:
Hi All,

Raised pull request https://github.com/Xilinx/XRT/pull/3184  to fix this.
yocto while creating dev rpm expects only softlinks in dev rpm. 
we disable this qa check for dev rpm by adding below code in xrt_git.bbappend file 

INSANE_SKIP_${PN}-dev += "dev-elf"

stsoe:
This pull request will be merged as a temporary work-around until we have moved the plugin libraries out of the lib directory.

mamin506:
> yocto while creating dev rpm expects only softlinks in dev rpm.
> we disable this qa check for dev rpm by adding below code in xrt_git.bbappend file
> 
> INSANE_SKIP_${PN}-dev += "dev-elf"

Hi @Himanshu-xilinx, I would not said disable Yocto's QA check is a good idea.
But it is acceptable workaround before the pipeline could run petalinux build.

Please enable QA check once pipeline is ready.
Thanks for looking into the issue and provide quick fix.

PR #3181, opened 2020-04-08
More test clean up and migration to PL APIs

sonals:
*Update xrt/02_simple/kernel.cl to use workgroup size of 1024 for iteration which removes dependency on host application passing group ids to kernel
*Update 02_simple host code for both python and xrt test case
*Update 22_verify xrt test case host code to use new PL APIs
*Capture errno for XRT allocation APIs

### DIFF
--- a/src/runtime_src/core/pcie/linux/shim.cpp
+++ b/src/runtime_src/core/pcie/linux/shim.cpp
@@ -122,8 +122,8 @@ namespace xocl {
  * handles the actual interaction with the kernel on the i/o data path.
  *
  * Configuration:
- * - qAioEn:		per queue aio context enabled or not 
- * - qAioCtx:		per queue aio context, valid only if qAioEn = true 
+ * - qAioEn:		per queue aio context enabled or not
+ * - qAioCtx:		per queue aio context, valid only if qAioEn = true
  * - set_option():	optional configurations for aio context and batching
  *
  * i/o data path:
@@ -156,7 +156,7 @@ public:
     const int queue_get_handle(void) { return (int)qhndl; }
 
     // optional configurations
-    int set_option(int type, uint32_t val) 
+    int set_option(int type, uint32_t val)
     {
         switch(type) {
         case STREAM_OPT_AIO_MAX_EVENT:
@@ -523,8 +523,13 @@ size_t shim::xclRead(xclAddressSpace space, uint64_t offset, void *hostBuf, size
 unsigned int shim::xclAllocBO(size_t size, int unused, unsigned flags)
 {
     drm_xocl_create_bo info = {size, mNullBO, flags};
+    unsigned int bo = mNullBO;
     int result = mDev->ioctl(mUserHandle, DRM_IOCTL_XOCL_CREATE_BO, &info);
-    return result ? mNullBO : info.handle;
+    if (result)
+        errno = result;
+    else
+        bo = info.handle;
+    return bo;
 }
 
 /*
@@ -534,8 +539,13 @@ unsigned int shim::xclAllocUserPtrBO(void *userptr, size_t size, unsigned flags)
 {
     drm_xocl_userptr_bo user =
         {reinterpret_cast<uint64_t>(userptr), size, mNullBO, flags};
+    unsigned int bo = mNullBO;
     int result = mDev->ioctl(mUserHandle, DRM_IOCTL_XOCL_USERPTR_BO, &user);
-    return result ? mNullBO : user.handle;
+    if (result)
+        errno = result;
+    else
+        bo = user.handle;
+    return bo;
 }
 
 /*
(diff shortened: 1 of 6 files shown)


src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -523,8 +523,13 @@ size_t shim::xclRead(xclAddressSpace space, uint64_t offset, void *hostBuf, size
...
 {
     drm_xocl_create_bo info = {size, mNullBO, flags};
+    unsigned int bo = mNullBO;
     int result = mDev->ioctl(mUserHandle, DRM_IOCTL_XOCL_CREATE_BO, &info);
-    return result ? mNullBO : info.handle;
+    if (result)
+        errno = result;
+    else
+        bo = info.handle;
+    return bo;
```
stsoe, line 532: This does not look right.   ioctl should return -1 on error and errno should be set.   You are overwriting it with -1.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -534,8 +539,13 @@ unsigned int shim::xclAllocUserPtrBO(void *userptr, size_t size, unsigned flags)
 {
     drm_xocl_userptr_bo user =
         {reinterpret_cast<uint64_t>(userptr), size, mNullBO, flags};
+    unsigned int bo = mNullBO;
     int result = mDev->ioctl(mUserHandle, DRM_IOCTL_XOCL_USERPTR_BO, &user);
-    return result ? mNullBO : user.handle;
+    if (result)
+        errno = result;
+    else
+        bo = user.handle;
```
stsoe, line 547: ditto

tests/xrt/02_simple/main.cpp
```diff
@@ -20,33 +20,22 @@
...
-#if defined(DSA64)
-#include "xsimple_hw_64.h"     
-#else
-#include "xsimple_hw.h"  
-#endif
-
-static const int count = 1024;
-int foo;
+// This value is shared with worgroup size in kernel.cl
+static const int COUNT = 1024;
```
stsoe, line 35: It's much better to use lower case.    You run a serious risk of preprocessor playing tricks with the code.   I think we need to mandate lower case for everything but #defines which all should be proceeded with // anyway.

tests/xrt/02_simple/main.cpp
```diff
@@ -20,33 +20,22 @@
...
 
-// driver includes
+// XRT includes
+#include "xrt.h"
 #include "ert.h"
-
-// host_src includes
-#include "xclhal2.h"
+#include "experimental/xrt_kernel.h"
+#include "experimental/xrt_xclbin.h"
```
haeseunglee, line 28: At this point, XRT tests with the experimental APIs won't work on the sprite because libxrt++ is not official library (for now) and sprite does not include libxrt++ for host compilation. So, either we need to disable this test on the sprite or find a way to add libxrt++ for this test.

tests/xrt/22_verify/main.cpp
```diff
@@ -22,21 +22,17 @@
...
 #include <time.h>
 
-// driver includes
-#include "ert.h"
-
 // host_src includes
-#include "xclhal2.h"
+#include "xrt.h"
+#include "experimental/xrt_kernel.h"
+#include "experimental/xrt_xclbin.h"
```
haeseunglee, line 28: Same to my previous comment for the experimental APIs.

src/runtime_src/core/pcie/linux/shim.cpp
sonals: The alloc bo ioctls return the error code from the driver. So far we have been throwing away that information which does not help the client since it only sees a null handle when the API fails.

tests/xrt/02_simple/main.cpp
sonals: These APIs are defined in xrt_coreutil.so which is available so the tests should work. The library is currently included in common.mk makefile

haeseunglee: The sprite is not using common.mk for compilation. It is using metadata to generate the makefile for the sprite run. It's because the existing makefile finds the common.mk from upper directory, which can't be done in sprite. So, it looks like I'll have to do necessary changes for the sprite.

src/runtime_src/core/pcie/linux/shim.cpp
stsoe: Ok, that is wrong then; it should set errno according to the ioctl man page.

stsoe:
@haeseunglee  These new APIs are not experimental for long.   Test cases that use these APIs (xrt_kernel.h) must link explicitly with xrt_coreutil.so.   Sprite will have to do this for specific tests we write against these APIs.  These APIs will forever be in xrt_coreutil.so because their implementation is shared between all platforms and OSes.    

For tests that don't use symbols from xrt_coreutils, nothing changes, e.g. they should *not* link we this library. 

Doesn't the yml files that create makefiles support adding arbitrary libraries to link with?

haeseunglee:
@stsoe It is. I can modify the metadata to add libxrt++. So, I'll have to make that change to support these experimental APIs.

sonals:
retest these please

sonals:
retest this please

sonals:
retest this please

sonals:
retest this please

PR #3183, opened 2020-04-09
XRT-408 zocl driver shows page request fails while unload

mamin506:
There is no cmd->client in ERT mode.
Only when this is ERT mode and there are garbage commands in the queue, it would lead to NULL pointer error or page request failed.


mamin506:
All of 'cmd->client' was protect by is_ert(), except this one. I would like to solve this issue in the unify KDS.

maxzhen:
Please use single space b/w task number and synopsis, not ":". Use same standard msg in both PR title and commit msg, please.

PR #3187, opened 2020-04-09
Changing the build of the XDP plugins to MODULE

jvillarre:
XDP plugins (xdp_debug_plugin, xdp_appdebug_plugin, xdp_hal_api_interface_plugin, xdp_lop_plugin, and xdp_hal_plugin) that are loaded at runtime and linked via dlopen and dlsyms are changed to be MODULE in the CMakeList files instead of shared libraries on Linux for edge and pcie.

Also, moving the install location to lib/xrt/module on Linux and changing the loading of the plugins in XRT code.

The elimination of duplicate code and consolidation of a single way to load the plugins will be addressed in a future pull request.


mamin506:
This would revert #3176. 
Hi @matwey, Thanks for the contribution!

PR #3188, opened 2020-04-09
XRT-409 Support ERT_EXIT command

mamin506:
Add ERT_EXIT command support in zocl.

In polling mode, it would stop cq_check thread.
In CQ interrupt mode, disable CQ interrupt

As ERT MB, it would not be able to handle new commands after this.
The only way to let ERT come back is hot reset.

Tested on U30.

After this, we are able to revert #3175


abalasa:
retest this please

PR #3189, opened 2020-04-10
AIE error and event handling

larry9523:
This PR implement the XRT AIE error/event handler.
1. Error handler is registered during AIE initialization.
2. Event handler is registerer during graph initializtion because event is reported on tiles based.
3. The current error/event handler will log message using existing XRT logging framework. 
4. Errors are classified according to its severity.
5. Events are logged as NOTICE

### DIFF
--- a/src/runtime_src/core/edge/user/aie/aie.cpp
+++ b/src/runtime_src/core/edge/user/aie/aie.cpp
@@ -17,6 +17,7 @@
  */
 
 #include "aie.h"
+#include "core/common/error.h"
 
 #include <iostream>
 #include <cerrno>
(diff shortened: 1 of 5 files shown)


src/runtime_src/core/edge/user/aie/aie.cpp
```diff
@@ -89,25 +90,129 @@ Aie::Aie()
...
-    /* TODO Register Event, Error XRT handler here ? */
-    xaie_register_event_handler(event, module, xrtAieEventCallBack);
-    xaie_register_error_notification(error, module, xrtAieErrorCallBack, arg);
-#endif
+    /* Disable AIE interrupts */
+    u32 reg = XAieGbl_NPIRead32(XAIE_NPI_ISR);
+    XAieGbl_NPIWrite32(XAIE_NPI_ISR, reg);
+
+    if (XAieTile_EventsHandlingInitialize(&aieInst) != XAIE_SUCCESS)
+        throw xrt_core::error("Failed to initialize AIE Events Handling.");
```
stsoe, line 98: Is there a particular error code that should end up getting propagated out of the C-API itself?  If so, then add that error code to the exception.

larry9523: The AIE driver just returns XAIE_FAILURE (which is 1). And reading its code, it is not clear to me what error code should be put here.

PR #3190, opened 2020-04-10
CR-1061236 : added LD_LIBRARY_PATH for sw emu

vboggara-xilinx:
added $XILINX_VIVADO/lib/lnx64.o/Default/ to LD_LIBRARY_PATH for sw emulation.


chvamshi-xilinx:
Good to create a loader for this process instead of adding LD_LIBRARY_PATH here...
Merging this change as it is critical..

PR #3193, opened 2020-04-13
CR-1050583 CR-1057548 : Merge overlapping memory banks into groups

stsoe:
Merge overlaping banks into groups, overlap is currently defined as
same base address.

The index of the group (grpidx) becomes the mem_topology mem_data[]
index (memidx) of exactly one of the memory banks in the group. This
ensures that grpidx can be used directly to index mem_topology
entries, which in turn simplifies upstream code that work with mem
indices and are blissfully unaware of the concept of group indices.

Only functions that return indicies into mem_topology need to be
changed to now return the corresponding grpidx.  The grpidx is a valid
memidx of a memory bank in the group.

### DIFF
--- a/src/runtime_src/xocl/xclbin/xclbin.cpp
+++ b/src/runtime_src/xocl/xclbin/xclbin.cpp
@@ -733,7 +733,7 @@ public:
         auto memidx = (used != upper) ? (*used).memidx : (*itr).memidx;
 
         // process the range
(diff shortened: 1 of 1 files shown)


stsoe:
retest this please

abalasa:
retest this please

stsoe:
retest this please

PR #3195, opened 2020-04-13
move qdma error intr from 0 to 8

houlz0507:


houlz0507:
This is required because u250 2rp uses vector 0 for mailbox.

PR #3197, opened 2020-04-13
XRT-430 xbutil top to print CMA_BANK

chienwei-lan:
To print out CMA_BANK related information in xbutil top

1. Change metrics function to property_tree based

2. Add sensor_tree clear function

3. Parse string with nice and clean way: sscanf

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
+++ b/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
@@ -778,6 +778,8 @@ static void topPrintUsage(const xcldev::device *dev, xclDeviceUsage& devstat)
 {
     std::vector<std::string> lines;
 
+    dev->readSensors();
+
     dev->m_mem_usage_bar(devstat, lines);
 
     dev->sysfs_stringize_power(lines);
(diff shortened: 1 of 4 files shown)


chienwei-lan:
retest this please

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -788,6 +790,7 @@ static void topPrintUsage(const xcldev::device *dev, xclDeviceUsage& devstat)
 
     dev->m_cu_usage_stringize_dynamics(lines);
 
+    dev->clearSensorTree();
```
AShivangi, line 793: Why do we need to clear the sensor tree? Doesn't it gets overwritten?

chienwei-lan: it didn't, instead it will have duplicate fields if you add the same entries.

and sensor_tree free when xbutil terminated. So we have to find a way to clear it.

AShivangi: I see.

PR #3200, opened 2020-04-14
XRT-431 New KDS driver has NULL pointer when there are 2 CUs in a xclbin

mamin506:
The CU size should be 64KB.
When it is failed to create a CU, the CU controller need to check if CU threads are allocated.


mamin506:
retest this please

PR #3201, opened 2020-04-14
XRT-285 incorrect feature rom header for u.2

AShivangi:
Output:
```
xbmgmt.exe status --report=platform
ERROR: Failed to detect XMC, bad magic number: 0
ERROR: Failed to detect XMC, bad magic number: 0
Device BDF : 0000:0e:00.0
  Flash type           : spi
Flashable partition running on FPGA
  Platform             : xilinx_samsung_u2x4_201920_2
  SC Version           : query request (21) not supported for mgmtpf on windows
  Platform ID          : 0x5e4fce30

  Platform             : xilinx_samsung_u2x4_201920_2
  SC Version           :
  Platform ID          : 0x0x5e4fce30

WARNING  : SC image on the device is not up-to-date.
```

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
@@ -355,7 +355,7 @@ auto_flash(xrt_core::device_collection& deviceCollection, bool force)
   // Collect all indexes of boards need updating
   std::vector<std::pair<unsigned int , DSAInfo>> boardsToUpdate;
   for (const auto & device : deviceCollection) {
-    DSAInfo dsa(_pt.get<std::string>(std::to_string(device->get_device_id()) + ".platform.installed_shell.file"));
+    DSAInfo dsa(_pt.get<std::string>(std::to_string(device->get_device_id()) + ".platform.installed_shell.0.file"));
     //if the shell is not up-to-date and dsa has a flash image, queue the board for update
     if (!_pt.get<bool>(std::to_string(device->get_device_id()) + ".platform.shell_upto_date") ||
           !_pt.get<bool>(std::to_string(device->get_device_id()) + ".platform.sc_upto_date")) {
--- a/src/runtime_src/core/pcie/driver/windows/include/XoclMgmt_INTF.h
+++ b/src/runtime_src/core/pcie/driver/windows/include/XoclMgmt_INTF.h
@@ -97,9 +97,9 @@ typedef union _DRIVER_VERSION
 {
     struct
     {
-        /* [Minor Version Number] Indicates the minor version is �0�. */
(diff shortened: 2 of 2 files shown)


AShivangi:
retest this please

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -355,7 +355,7 @@ auto_flash(xrt_core::device_collection& deviceCollection, bool force)
   // Collect all indexes of boards need updating
   std::vector<std::pair<unsigned int , DSAInfo>> boardsToUpdate;
   for (const auto & device : deviceCollection) {
-    DSAInfo dsa(_pt.get<std::string>(std::to_string(device->get_device_id()) + ".platform.installed_shell.file"));
+    DSAInfo dsa(_pt.get<std::string>(std::to_string(device->get_device_id()) + ".platform.installed_shell.0.file"));
```
rozumx, line 358: This doesn't look right to have a "0" in the path.  Would you be so kind as to share the property_tree structure that requires this syntax.

AShivangi: This is to support the case when multiple shells are installed on the system: https://github.com/Xilinx/XRT/commit/8aee70e82d7afbe1bdd59e6f85d848d50d9f7f99. We won't reach this code if multiple shells are installed on the system (we don't do auto update in that case). Whenever we reach this point in the code, we only have one shell "0" which is why i think it is fine to have 0 in the path. After I made the change, the ptree looks something like this:
```
{
    "schema_version": {
        "schema": "JSON-2020.1"
    },
    "devices": [
        {
            "platform": {
                "Description": "Platform Information",
                "bdf": "0000:b3:00.0",
                "flash_type": "spi",
                "shell_on_fpga": {
                    "vbnv": "xilinx_u200_xdma_201830_2",
                    "sc_version": "4.2.0",
                    "id": "0x5d1211e8"
                },
                "number_of_installed_shells": "2",
                "shell_upto_date": "false",
                "sc_upto_date": "false",
                "installed_shell": {
                    "0": {
                        "vbnv": "xilinx_u200_xdma_201830_1",
                        "sc_version": "3.1",
                        "id": "0x5bece8e1",
                        "file": "\/lib\/firmware\/xilinx\/10ee-5000-000e-000000005bece8e1.dsabin"
                    },
                    "1": {
                        "vbnv": "xilinx_u200_qdma_201920_1",
                        "sc_version": "4.3.7",
                        "id": "0x5dccb0ca",
                        "file": "\/lib\/firmware\/xilinx\/10ee-5010-000e-000000005dccb0ca.xsabin"
                    }
                }
            }
        }
    ]
}
```

AShivangi:
retest this please

PR #3203, opened 2020-04-14
Removing redundant code in loading XDP plugin modules from the OpenCL layer

jvillarre:
Refactored the loading of XDP plugin modules loaded from the OpenCL layer to remove duplicate code.


jvillarre:
retest this please

PR #3213, opened 2020-04-15
Resolve the mismatch of the cu name between the host code and the metadata

haeseunglee:
Resolve the mismatch of the cu name between the host code and the metadata
Both the host code and metadata will use hello:hello_1


maxzhen:
ok to test

PR #3220, opened 2020-04-16
CR-1056667 support batching of asynch i/o on stream queues

karenx-xilinx:
During stream performance testing, we found that syscall io_submit() is a bottleneck, this patch adds the support to allow accumulating then submitting multiple stream i/o in a single io_submit() call.

To enable asynchronous i/o batching, the host app needs to configure the batching threshold via clSetStreamOpt(). There are two thresholds, they can be used individually or combined:

1. timer - accumulating the request until the timer pops.
2. total buffer count: total # of buffers in all of the i/o requests

When either of the above condition is met, all of the accumulated i/o submitted to the kernel.

### DIFF
--- a/src/runtime_src/core/pcie/linux/shim.cpp
+++ b/src/runtime_src/core/pcie/linux/shim.cpp
@@ -149,6 +149,7 @@ private:
     bool qAioBatchEn;		/* per queue aio req. batching is enabled */
     bool qExit;			/* queue is being stopped/released */
     bool h2c; 			/* queue is for H2C direction */
+    uint64_t qhndl;		/* queue handle */
     unsigned int aio_max_evts;	/* for io_setup() max. event concurrently */
 
     /* aio batching threshold to flush the queued i/o */
@@ -163,7 +164,6 @@ private:
     int cbErrCnt;		/* submission error */
     int cbErrCode;		/* submission error code */
 
-    uint64_t qhndl;		/* queue handle */
     aio_context_t qAioCtx;	/* per queue aio context */
 
     std::thread qWorker;	/* aio batching thread */
@@ -316,7 +316,7 @@ private:
             std::unique_lock<std::mutex> lck(reqLock);
 	    /* set a time out to prevent holding the i/o forever */
             if (!timer_ms)
-		cv.wait_for(lck, std::chrono::milliseconds(1000));
+		cv.wait(lck);
             else
 		cv.wait_for(lck, std::chrono::milliseconds(timer_ms));
 
@@ -361,12 +361,11 @@ private:
 
 public:
     queue_cb(struct xocl_qdma_ioc_create_queue *qinfo)
-        : qAioEn{false}, qAioBatchEn{false}, qExit{false}, aio_max_evts{0},
-          timer_ms{0}, byteThresh{0}, pktThresh{0}, byteCnt{0}, bufCnt{0},
-          cbSubmitCnt{0}, cbPollCnt{0}, cbErrCnt{0}
+        : qAioEn{false}, qAioBatchEn{false}, qExit{false},
+          h2c{qinfo->write ? true : false}, qhndl{qinfo->handle},
+          aio_max_evts{0}, timer_ms{0}, byteThresh{0}, pktThresh{0},
+          byteCnt{0}, bufCnt{0}, cbSubmitCnt{0}, cbPollCnt{0}, cbErrCnt{0}
     {
-       qhndl = qinfo->handle;
-       h2c = qinfo->write ? true : false;
        memset(&qAioCtx, 0, sizeof(qAioCtx));
     }
 
@@ -506,11 +505,10 @@ public:
             for (unsigned int i = 0; i < wr->buf_num; i++) {
                 struct iovec iov[2];
 		struct iocb cb;
-		struct iocb *cbs[1];
+		struct iocb *cbs[1] = {&cb};
 
                 prepare_io(&cb, iov, &header, wr->bufs[i].va, wr->bufs[i].len,
 			 (uint64_t)wr->priv_data);
-                cbs[0] = &cb;
                 int rv = io_submit(*aio_ctx, 1, cbs);
                 if (rv <= 0)
                     break;
@@ -521,7 +519,7 @@ public:
         } else {
             for (unsigned int i = 0; i < wr->buf_num; i++) {
                 struct iovec iov[2];
-                prepare_io(NULL, iov, &header, wr->bufs[i].va, wr->bufs[i].len, 0);
+                prepare_io(nullptr, iov, &header, wr->bufs[i].va, wr->bufs[i].len, 0);
                 if (h2c)
                     rc = writev((int)qhndl, iov, 2);
                 else
--- a/src/runtime_src/core/pcie/linux/shim.cpp
+++ b/src/runtime_src/core/pcie/linux/shim.cpp
@@ -145,12 +145,10 @@ private:
        struct xocl_qdma_req_header header;	/* used for io_submit */
     };
 
-    uint64_t qhndl;		/* queue handle */
-    bool h2c; 			/* queue is for H2C direction */
     bool qAioEn;		/* per queue aio is enabled */
     bool qAioBatchEn;		/* per queue aio req. batching is enabled */
     bool qExit;			/* queue is being stopped/released */
-    aio_context_t qAioCtx;	/* per queue aio context */
+    bool h2c; 			/* queue is for H2C direction */
     unsigned int aio_max_evts;	/* for io_setup() max. event concurrently */
 
     /* aio batching threshold to flush the queued i/o */
@@ -164,6 +162,9 @@ private:
     unsigned int cbPollCnt;	/* total # of polled/completed i/o */
     int cbErrCnt;		/* submission error */
     int cbErrCode;		/* submission error code */
+
+    uint64_t qhndl;		/* queue handle */
+    aio_context_t qAioCtx;	/* per queue aio context */
 
     std::thread qWorker;	/* aio batching thread */
     std::mutex reqLock;		/* lock to protect i/o related info. */
@@ -184,9 +185,36 @@ private:
             byteCnt -= qio.len;
             reqList.pop_front();
         }
-        
+
         return i;
     } 
+
+    int queue_up_request(xclQueueRequest *wr)
+    {
+        std::lock_guard<std::mutex> lk(reqLock);
+
+        /* if there was io submission error, wait till all requests are
+         * drained or list full*/
+        if (cbErrCnt || (reqList.size() == aio_max_evts))
+            return -EAGAIN;
+
+        /* queue up this async i/o request */
+        unsigned int bytes = 0;
+        for (unsigned int i = 0; i < wr->buf_num; i++) {
+             bytes += wr->bufs[i].len; 
+             reqList.push_back({wr->flag, (uint64_t)wr->priv_data,
+				wr->bufs[i].va, wr->bufs[i].len});
+        }
+        bufCnt += wr->buf_num;
+        byteCnt += bytes;
+
+	/* wake up the batch worker thread if threshold is reached */
+        if ((pktThresh && bufCnt >= pktThresh) ||
+            (byteThresh && byteCnt >= byteThresh))
+            cv.notify_one();
+
+        return wr->buf_num;
+    }
 
     int check_io_submission_error(int nr_comps, struct xclReqCompletion *comps)
     {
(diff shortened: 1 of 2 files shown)


maxzhen:
Please use single space b/w CR number and synopsis, not ":" in pr title and commit msg

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -131,28 +132,220 @@ namespace xocl {
...
+
+        do {
+            /* submit all queued requests */
+            unsigned int cb_max = bufCnt;
+            reqLock.unlock();
+
+            unsigned int cb_cnt = 0;
+            struct iocb **cbpp = (struct iocb **)malloc(cb_max * sizeof(struct iocb *));
+
+            reqLock.lock();
```
stsoe, line 231: explicit lock and unlock of mutexes are not exception safe. Please use std::lock_guard

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -131,28 +132,220 @@ namespace xocl {
...
+            return 0;
+        }
+
+        do {
+            /* submit all queued requests */
+            unsigned int cb_max = bufCnt;
+            reqLock.unlock();
+
+            unsigned int cb_cnt = 0;
+            struct iocb **cbpp = (struct iocb **)malloc(cb_max * sizeof(struct iocb *));
```
stsoe, line 229: malloc followed by a later free looks unsafe.  Can this be scoped to be exception safe?
Could it maybe be a std::vector?
std::vector<iocb*> cbpp(cb_max);
which is automatically scoped and much preferable over raw new and delete.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -131,28 +132,220 @@ namespace xocl {
...
+            reqLock.lock();
+            auto end = reqList.end();
+            for (auto it = reqList.begin(); it != end && cb_cnt < cb_max;
+                ++it, cb_cnt++) {
+                auto qio = (*it);
+
+                qio.header.flags = qio.flags;
+                prepare_io(&qio.cb, qio.iov, &qio.header, qio.buf_va, qio.len,
+				qio.priv_data);
+                *(cbpp+cb_cnt) = &qio.cb;
```
stsoe, line 240: too cryptic; I read it though :-)  much simpler with vector

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -131,28 +132,220 @@ namespace xocl {
...
+            }
+
+            int submitted = io_submit(qAioCtx, cb_cnt, cbpp);
+            free(cbpp);
```
stsoe, line 244: with vector this would change to
`io_submit(,,cbpp.data());`
and subsequent free would go away

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -131,28 +132,220 @@ namespace xocl {
...
+            if (submitted < 0) {
+                if (submitted != -EAGAIN) {
+		    /* something went wrong, flush all of the queued request
+                     * with this error */
+                    cbErrCnt = reqList.size();
+                    cbErrCode= submitted;
+		}
+                reqLock.unlock();
```
stsoe, line 252: not good

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -131,28 +132,220 @@ namespace xocl {
...
+                return submitted;
+
+            } else if (submitted > 0) {
+                release_request(submitted);
+                cbSubmitCnt += submitted;
+            }
+        } while(reqList.size());
+        reqLock.unlock();
```
stsoe, line 260: not good.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -131,28 +132,220 @@ namespace xocl {
...
+                    cbErrCode= submitted;
+		}
+                reqLock.unlock();
+                return submitted;
+
+            } else if (submitted > 0) {
+                release_request(submitted);
+                cbSubmitCnt += submitted;
+            }
+        } while(reqList.size());
```
stsoe, line 259: I feel the whole do ... while loop is too long and complicated, can it be refactored?
You know the best comment I ever heard was from Sean Parent: If you write a loop, you must be inventing an algorithm, so maybe you should file a pattern.   It worries me when I see long complicated nested loops.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -131,28 +132,220 @@ namespace xocl {
...
+                reqLock.unlock();
+                continue;
+            }
+            reqLock.unlock();
+
+            if (qExit) break;
+
+            queue_flush_aio_request();
+          
+        } while(!qExit);
```
stsoe, line 292: why this test?  there is already a break on qExit, so I infer !qExit must always be true if tested.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -131,28 +132,231 @@ namespace xocl {
...
+    }
+
+    /* worker thread for aio batching */
+    void queue_aio_worker(void)
+    {
+        byteCnt = bufCnt = 0;
+        cbSubmitCnt = cbPollCnt = cbErrCnt = 0;
+
+        std::mutex mtx;
+        std::unique_lock<std::mutex> lck(mtx);
```
stsoe, line 288: Are these really needed.  Should the loop just acquire reqLock and wait on that lck?  I doubt the 'if .. else ...' with cv.wait has any significant time that allows reqLock to be locked by other thread before it is immediately locked at line 296 in the loop.  In short, I would have expected:
```
do {
  std::unique_lock<std::mutex> lck(reqLock);
  if ...
  else ..
  ...
}
```
with no additional locking of reqLock.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -131,28 +132,231 @@ namespace xocl {
...
+	    /* set a time out to prevent holding the i/o forever */
+            if (!timer_ms)
+		cv.wait_for(lck, std::chrono::milliseconds(1000));
+            else
+		cv.wait_for(lck, std::chrono::milliseconds(timer_ms));
+
+            std::lock_guard<std::mutex> lk(reqLock);
+            if (reqList.empty()) {
+                continue;
+            }
```
stsoe, line 299: convention is no curly base for single statement.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -131,28 +132,231 @@ namespace xocl {
...
+        std::mutex mtx;
+        std::unique_lock<std::mutex> lck(mtx);
+        do {
+	    /* set a time out to prevent holding the i/o forever */
+            if (!timer_ms)
+		cv.wait_for(lck, std::chrono::milliseconds(1000));
+            else
+		cv.wait_for(lck, std::chrono::milliseconds(timer_ms));
+
+            std::lock_guard<std::mutex> lk(reqLock);
```
stsoe, line 296: Not needed if you use reqLock for the cv lock also.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -131,28 +132,231 @@ namespace xocl {
...
+    }
+
+    void queue_aio_batch_disable_check(void)
+    {
+        if (qAioBatchEn && !byteThresh && !pktThresh && !timer_ms) {
+            std::lock_guard<std::mutex> lk(reqLock);
+            /* stop the worker thread */
+            qExit = true;
+            cv.notify_one();
+            if (qWorker.joinable())
```
stsoe, line 313: end the 'if' statement block here, so that reqLock is released and the worker thread can finish (that is if you do as I suggested regarding mutexes in the worker thread).

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -131,28 +132,231 @@ namespace xocl {
...
     queue_cb(struct xocl_qdma_ioc_create_queue *qinfo)
     {
        qhndl = qinfo->handle;
        h2c = qinfo->write ? true : false;
-       qAioEn = false;
+       qAioEn = qAioBatchEn = qExit = false;
+       byteThresh = pktThresh = timer_ms = 0;
 
        memset(&qAioCtx, 0, sizeof(qAioCtx));
+       aio_max_evts = 0;
```
stsoe, line 339: Constructors tend to use member initializer list, initializing all members in the order they are declared in the class.  Leave the body for complex initialization, like the memset.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -131,28 +132,231 @@ namespace xocl {
...
+    {
+        if (!qAioEn)
+           return;
+
+        if (qAioBatchEn) {
+            qExit = true;
+            cv.notify_one();
+            if (qWorker.joinable())
+                qWorker.join();
+        }
```
stsoe, line 352: ?  Isn't this the same as queue_aio_batch_disable_check()

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -131,28 +132,231 @@ namespace xocl {
...
+	
+        io_destroy(qAioCtx);
     }
+
     const bool queue_aio_ctx_enabled(void) { return qAioEn; }
     const bool queue_is_h2c(void) { return h2c; }
+    const bool queue_aio_batch_enabled(void) {return qAioBatchEn; }
     const int queue_get_handle(void) { return (int)qhndl; }
```
stsoe, line 360: declare these as 'const' member functions.
c-style cast are not really c++'ish.  static_cast<int>  ?

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -211,53 +438,96 @@ class queue_cb {
...
+             rc = check_io_submission_error(max_compl - num_evt, comps + num_evt);
+             num_evt += rc;
+        }
+
+        *actual = num_evt;
         return 0;
     }
 
     // submit the read/write i/o to the queue
     ssize_t queue_submit_io(xclQueueRequest *wr, aio_context_t *mAioCtx)
```
stsoe, line 452: This function is too long.  I can't quickly comprehend it.  Maybe refactor by creating synchronous and asynchronous submit_io functions ?

src/runtime_src/core/pcie/linux/shim.cpp
karenx-xilinx: I don't want to hold the lock while waiting for the timer to pop. A user could submit an i/o during this time and it can be added to the reqList.

src/runtime_src/core/pcie/linux/shim.cpp
karenx-xilinx: I moved queuing up of the wr to a separate function.
In this function, it first checks if batching is enabled or not.  If it does and this i/o is async, it just queue up the new i/o. If this i/o is synchronous then first flush all of the queued async i/o to the kernel before submitting this sync i/o.
If no batching, just go ahead and submit to the kernel.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -211,53 +462,75 @@ class queue_cb {
...
+        /* synchronous i/o or no batching configured on the queue, submit the
+         * i/o right away */
+        struct xocl_qdma_req_header header;
+        header.flags = wr->flag;
+        if (aio) {
+            aio_context_t *aio_ctx = qAioEn ? &qAioCtx : mAioCtx;
+            for (unsigned int i = 0; i < wr->buf_num; i++) {
+                struct iovec iov[2];
+		struct iocb cb;
+		struct iocb *cbs[1];
```
stsoe, line 509: Do it this way:
```
struct iocb* cbs[1] = {nullptr};
prepare_io(cbs, ...);
```
no need for line 513.

karenx-xilinx: struct iocb* cbs[1] = {nullptr};
prepare_io(cbs, ...);
would not compile, if I changes to prepare_io(*cbs, ....) then it compile, but failed the later io_submit() call

so I did 
struct iocb cb;
struct iocb *cbs[1] = {&cb};

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -131,47 +132,294 @@ namespace xocl {
...
+    {
+        byteCnt = bufCnt = 0;
+
+        do {
+            std::unique_lock<std::mutex> lck(reqLock);
+	    /* set a time out to prevent holding the i/o forever */
+            if (!timer_ms)
+		cv.wait(lck);
+            else
+		cv.wait_for(lck, std::chrono::milliseconds(timer_ms));
```
stsoe, line 321: The timer is irrelevant.  Here is how I think it should be:
```
std::unique_lock<std::mutex> lck(reqLock);
cv.wait(lck, []{return !reqList.empty()});
queue_flush_aio_request();
```

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -131,47 +132,294 @@ namespace xocl {
...
+
+            if (reqList.empty())
+                continue;
```
stsoe, line 324: scratch

src/runtime_src/core/pcie/linux/shim.cpp
karenx-xilinx: Sure, let me remove the setting of the timer then.

PR #3223, opened 2020-04-17
XRT-423 Get 2020.1 lenet work with XRT Graph API

larry9523:
This PR addressed the following issues:
1) Fix a few bugs to make lenet work with native XRT Graph API
   a) AIE driver does not support unregister all errors and register/unregister all events
   b) tile.row should be added by one to include shim tile
2) Implement getting GMIO metadata from Xclbin
3) Address some comments from after last PR (https://github.com/Xilinx/XRT/pull/3103)

### DIFF
--- a/src/runtime_src/core/edge/common/aie_parser.cpp
+++ b/src/runtime_src/core/edge/common/aie_parser.cpp
@@ -29,6 +29,7 @@ namespace {
 namespace pt = boost::property_tree;
 using tile_type = xrt_core::edge::aie::tile_type;
 using rtp_type = xrt_core::edge::aie::rtp_type;
+using gmio_type = xrt_core::edge::aie::gmio_type;
 
 inline void
 throw_if_error(bool err, const char* msg)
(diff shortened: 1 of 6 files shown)


src/runtime_src/core/edge/common/aie_parser.cpp
```diff
@@ -122,6 +123,24 @@ get_rtp(const pt::ptree& aie_meta)
...
+  std::vector<gmio_type> gmios;
+
+  for (auto& gmio_node : aie_meta.get_child("aie_metadata.GMIOs")) {
+    gmio_type gmio;
+
+    gmio.id = gmio_node.second.get<std::string>("id");
+    gmio.name = gmio_node.second.get<std::string>("name");
+    gmio.type = gmio_node.second.get<uint16_t>("type");
+    gmio.shim_col = gmio_node.second.get<uint16_t>("shim_column");
+    gmio.burst_len = gmio_node.second.get<uint16_t>("burst_length_in_16byte");
```
sonals, line 138: Did we miss capturing gmio object into the gmios vector?
`gmios.push_back(gmio)`

larry9523: You are right. I just added. Apparently I am not able to test it without a valid metadata file. Good catch!

larry9523:
retest this please

sonals:
retest this please

PR #3230, opened 2020-04-18
[XRT-318] flashing takes too long

AShivangi:
- chrono thread sleep resolution on windows is 1 ms which is higher than the required sleep. Replacing the thread sleep with a busy loop. 
- xbmgmt2 flash times are comparable with xbmgmt flash times
- u.2 flashing on windows takes < 5 min

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/flash/xspi.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/flash/xspi.cpp
@@ -297,7 +297,7 @@ static void stripe_data(uint8_t *intrlv_buf, uint8_t *buf0, uint8_t *buf1, uint3
 static void delay(std::chrono::microseconds us) 
 {
 	std::chrono::high_resolution_clock::time_point currTime;
-	std::chrono::high_resolution_clock::time_point startTime = std::chrono::high_resolution_clock::now();
+	const std::chrono::high_resolution_clock::time_point startTime = std::chrono::high_resolution_clock::now();
 	do {
 		currTime = std::chrono::high_resolution_clock::now();
 	} while (std::chrono::duration<double>(currTime - startTime) < us);
@@ -764,8 +764,9 @@ int XSPI_Flasher::writeReg(unsigned int RegOffset, unsigned int value)
 
 
 bool XSPI_Flasher::waitTxEmpty() {
-    long long _delay = 0;
-    while (_delay < 30000000000) {
+    std::chrono::high_resolution_clock::time_point currTime;
+	const std::chrono::high_resolution_clock::time_point startTime = std::chrono::high_resolution_clock::now();
+	do {
         uint32_t StatusReg = XSpi_GetStatusReg();
         if(StatusReg & XSP_SR_TX_EMPTY_MASK )
             return true;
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/tools/xbmgmt2/flash/xspi.cpp
```diff
@@ -24,7 +24,6 @@
 #include <string>
 #include <fstream>
 #include <cassert>
-#include <thread>
```
rozumx, line 27: Thank you for removing unused header files.

src/runtime_src/core/tools/xbmgmt2/flash/xspi.cpp
```diff
@@ -291,6 +290,19 @@ static void stripe_data(uint8_t *intrlv_buf, uint8_t *buf0, uint8_t *buf1, uint3
...
 }
 
+/*
+ * Chronos sleep resolution on Windows is 1 ms which is exponentially bigger than
+ * the required sleep. This is a busy loop to mimick the accurate amount of sleep.
+ */
+static void delay(std::chrono::microseconds us) 
+{
+	std::chrono::high_resolution_clock::time_point currTime;
+	std::chrono::high_resolution_clock::time_point startTime = std::chrono::high_resolution_clock::now();
```
rozumx, line 300: Should be a const.

src/runtime_src/core/tools/xbmgmt2/flash/xspi.cpp
```diff
@@ -752,25 +764,25 @@ int XSPI_Flasher::writeReg(unsigned int RegOffset, unsigned int value)
 
 
 bool XSPI_Flasher::waitTxEmpty() {
-    long long delay = 0;
-    while (delay < 30000000000) {
+    long long _delay = 0;
```
rozumx, line 767: As you know, I don't like magic numbers. This loop makes a lot of assumptions, but if one does the calculations (e.g. 30000000000*5us) the total delay here is 3 seconds.  To better help the user it would be better to rewrite the code to reflect that desire.

Similar to the delay() code that was written earlier, create a loop around this code that will expire after 3 seconds and produce the error message when it does.

src/runtime_src/core/tools/xbmgmt2/flash/xspi.cpp
```diff
@@ -1491,7 +1503,7 @@ bool XSPI_Flasher::prepareXSpi(uint8_t slave_sel)
 #endif
 
     //TODO: Do we need this short delay still?
```
rozumx, line 1505: With the works others are doing, we no longer need this comment.

src/runtime_src/core/tools/xbmgmt2/flash/xspi.cpp
```diff
@@ -784,8 +796,8 @@ bool XSPI_Flasher::isFlashReady() {
         }
         //TODO: Try resetting. Uncomment next line?
         //XSpi_WriteReg(XSP_SRR_OFFSET, XSP_SRR_RESET_MASK);
-        std::this_thread::sleep_for(std::chrono::nanoseconds(5000));
-        delay += 5000;
+        delay(std::chrono::microseconds(5));
```
rozumx, line 799: I like what you did here....it is clear to the user that we are delaying 5 microseconds.

src/runtime_src/core/tools/xbmgmt2/flash/xspi.cpp
AShivangi: This is a great catch! Thanks for explaining it :)

PR #3233, opened 2020-04-19
clean up coverity warnings & remove debug printf

karenx-xilinx:
added missing initializer and removed the lock for setting qExit as setting of it is only from clSetStreamOpt() or clReleaseStream(), assume no concurrency here.
Also removed the debug printf accidentally checked in.

### DIFF
--- a/src/runtime_src/core/pcie/linux/shim.cpp
+++ b/src/runtime_src/core/pcie/linux/shim.cpp
@@ -324,7 +324,6 @@ private:
     void stop_aio_worker(void)
     {
         if (qAioBatchEn) {
-            std::lock_guard<std::mutex> lk(reqLock);
             /* stop the worker thread */
             qExit = true;
             cv.notify_one();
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -324,7 +324,6 @@ class queue_cb {
     void stop_aio_worker(void)
     {
         if (qAioBatchEn) {
-            std::lock_guard<std::mutex> lk(reqLock);
```
stsoe, line 327: really?   I wouldn't have left this one; I think synchronization and a memory point is missing if you remove this.  In my experience, the thread routine will never see qExit changed.

karenx-xilinx: I see. I guess the value is cached. Let me fix it.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -344,7 +344,8 @@ class queue_cb {
         /* to enable aio batching, the stream needs to have its own context
          * and any of the 
          */
-        if (qAioEn && (byteThresh || pktThresh)) {
+        if (qAioEn && (byteThresh || pktThresh) && !qAioBatchEn) {
+            std::lock_guard<std::mutex> lk(reqLock);
```
stsoe, line 348: yeah, oh well, this is a work-around for a false positive Coverity defect, since this function is called from main thread only before worker thread shares qExit.

stsoe:
retest this please

PR #3236, opened 2020-04-19
CR-1062504 CR-1049689 multiple thread hang waiting for commands to finish

stsoe:
This is specific to AP_CTRL_CHAIN and ERT assisted polling for CU
completion.  KDS can be ahead of ERT in that it doesn't have to wait
for an interrupt form ERT before determining that a CU has completed.
When KDS gets ahead, there could be a pending interrupt for one of CUs
that KDS itself determined as completed.  Without acknowledging the
interrupt, ERT cannot send future interrupts and KDS can in certain
cases go into a waiting state when no further commands arrive and none
appear to have completed.

The fix is to force one extra iteration of the scheduler after
transitioning new commands from submitted to running.  This way the
interrupt register gets cleared by the code that checks if running
commands are done.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
@@ -1369,6 +1369,28 @@ ert_cfg(struct xocl_ert *xert, unsigned int cq_size, unsigned int num_slots, boo
 	xert->ctrl_busy = false;
 }
 
+/**
+ * Clear the ERT command queue status register
+ *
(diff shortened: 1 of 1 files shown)


stsoe:
Sorry, the PR has unrelated minor refactoring.  The change is really only in two lines in 3173-3174

maxzhen:
So we are assuming that the spurious interrupt from ERT will come before the next forced iteration starts (so that it can be ack'ed in that iteration)? Can it come after?

stsoe:
Yes it can and that would be ideal because in that case a next iteration will be enforced by the interrupt calling the same function I just inserted.

PR #3237, opened 2020-04-19
CR-1036487 slot number,size should be consistent

xdavidz:
Please take a look at the comment. I am not sure if that code is valid, I think the slot_size of cq should be always 4k.


src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -776,6 +776,7 @@ configure(struct sched_cmd *cmd)
 	}
 
 	SCHED_DEBUG("Configuring scheduler\n");
+	/* ??? so slot_size can be not 4k, but cq_size is always 64k? */
```
mamin506, line 779: In current implementation.

PR #3239, opened 2020-04-20
CR-1034331 xbutil dmatest crashes if not enough system memory available

sonals:
* Use xclAllocUserPtrBO() to allocate buffers for dmatest which is more robust than xclAllocBO()
* DMARunner now throws exceptions which are supposed to be caught by the callers

### DIFF
--- a/src/runtime_src/core/pcie/common/dmatest.h
+++ b/src/runtime_src/core/pcie/common/dmatest.h
@@ -52,14 +52,15 @@ namespace xcldev {
         // DMARunner now uses xclAllocUserPtrBO() to allocate buffers. This reduces memory pressure on
         // Linux kernel which other wise tries very hard inside xocl to allocate and pin pages when
         // xlcAllocBO() is used may oops.
-        std::vector<std::pair<xclBufferHandle, xrt_core::aligned_ptr_type>> mBOList;
+        using buffer_and_deleter = std::pair<xclBufferHandle, xrt_core::aligned_ptr_type>;
+        std::vector<buffer_and_deleter> mBOList;
         xclDeviceHandle mHandle;
         size_t mSize;
         unsigned mFlags;
-        char pattern;
+        char mPattern;
 
-        int runSyncWorker(std::vector<std::pair<xclBufferHandle, xrt_core::aligned_ptr_type>>::const_iterator b,
-                          std::vector<std::pair<xclBufferHandle, xrt_core::aligned_ptr_type>>::const_iterator e,
+        int runSyncWorker(std::vector<buffer_and_deleter>::const_iterator b,
+                          std::vector<buffer_and_deleter>::const_iterator e,
                           xclBOSyncDirection dir) const {
             int result = 0;
             while (b < e) {
@@ -99,8 +100,8 @@ namespace xcldev {
         }
 
         int validate() const {
-            std::vector<char> bufCmp(mSize, pattern);
-            for (const std::pair<xclBufferHandle, xrt_core::aligned_ptr_type> &bo : mBOList) {
+            std::vector<char> bufCmp(mSize, mPattern);
+            for (const buffer_and_deleter &bo : mBOList) {
                 if (!std::memcmp(bo.second.get(), bufCmp.data(), mSize))
                     continue;
                 throw xrt_core::error(-EIO, "DMA test data integrity check failed.");
@@ -110,7 +111,7 @@ namespace xcldev {
 
     public:
         DMARunner(xclDeviceHandle handle , size_t size, unsigned flags=0) : mHandle(handle), mSize(size),
-                                                                            mFlags(flags), pattern('x') {
+                                                                            mFlags(flags), mPattern('x') {
             long long count = 0x100000000/size;
 
             if (count == 0)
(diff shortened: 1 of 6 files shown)


src/runtime_src/core/pcie/common/dmatest.h
```diff
@@ -80,74 +89,73 @@ namespace xcldev {
...
             int result = 0;
-            for_each(threads.begin(), threads.end(), [&](std::future<int> &v) {result += v.get();});
+            std::for_each(threads.begin(), threads.end(), [&](std::future<int> &v) {result += v.get();});
+            return result;
+        }
+
+        void clear() {
+            //Clear out the host shadow buffer
+            std::for_each(mBOList.begin(), mBOList.end(), [&](std::pair<xclBufferHandle, xrt_core::aligned_ptr_type> &v) {
+                    std::memset(v.second.get(), 'x', mSize);
```
stsoe, line 99: Strictly formatting.  If the lambda doesn't fit on the same line as the std::algorithm, then I really like it like this:
```
std::for_each(mBOList.begin(), mBOList.end(), 
              [&](std::pair<xclBufferHandle, xrt_core::aligned_ptr_type> &v) {
                    std::memset(v.second.get(), 'x', mSize);
              });
```
That way it looks and feels a lot more like a regular function.

src/runtime_src/core/pcie/common/dmatest.h
```diff
@@ -80,74 +89,73 @@ namespace xcldev {
...
-                    break;
-		}
-
-                if (std::memcmp(buf, bufCmp.get(), mSize)) {
-                    std::cout << "DMA Test data integrity check failed\n";
-                    break;
-                }
-            }
-            return static_cast<int>(result);
+            std::for_each(mBOList.begin(), mBOList.end(), [&](std::pair<xclBufferHandle, xrt_core::aligned_ptr_type> &bo) {
```
stsoe, line 140: you know that lambda in C++14 takes auto as argument, e.g.
`std::for_each(mBOList.begin(), mBOList.end(), [&](auto& bo) { xclFreeBO(mHandle, bo.first); });`
This may fit on one line.

src/runtime_src/core/pcie/common/dmatest.h
```diff
@@ -80,74 +89,73 @@ namespace xcldev {
...
+        void clear() {
+            //Clear out the host shadow buffer
+            std::for_each(mBOList.begin(), mBOList.end(), [&](std::pair<xclBufferHandle, xrt_core::aligned_ptr_type> &v) {
+                    std::memset(v.second.get(), 'x', mSize);
+                });
+        }
+
+        int validate() const {
+            std::unique_ptr<char[]> bufCmp(new char[mSize]);
+            std::memset(bufCmp.get(), pattern, mSize);
```
stsoe, line 105: std::vector<char> is probably preferable for this type of RAII:
`std::vector<char> bufCmp(pattern, pattern+mSize);`

src/runtime_src/core/pcie/common/dmatest.h
```diff
@@ -44,17 +47,23 @@ namespace xcldev {
     };
 
     class DMARunner {
-        std::vector<xclBufferHandle> mBOList;
+        // Ideally I would use a C++ BO object which hides this detail completely inside
+        // but that feature is coming in a future release. For now use a poor man implementation.
+        // DMARunner now uses xclAllocUserPtrBO() to allocate buffers. This reduces memory pressure on
+        // Linux kernel which other wise tries very hard inside xocl to allocate and pin pages when
+        // xlcAllocBO() is used may oops.
+        std::vector<std::pair<xclBufferHandle, xrt_core::aligned_ptr_type>> mBOList;
```
rozumx, line 55: You might wish to consider defining a "using-directive" for the "pair" in the vector since this syntax is used in multiple locations below.  Something like:
```
using std::pair<xclBufferHandle, xrt_core::aligned_ptr_type>  buffer_and_deleter;
std::vector<buffer_and_deleter> mBOList;
```

src/runtime_src/core/pcie/common/dmatest.h
```diff
@@ -44,17 +47,23 @@ namespace xcldev {
...
         xclDeviceHandle mHandle;
         size_t mSize;
         unsigned mFlags;
+        char pattern;
```
rozumx, line 59: Being that this is a member variable, it would be best to follow the naming convention of the other member variables and add rename "pattern" to "mPattern".

src/runtime_src/core/pcie/common/dmatest.h
rozumx, line 99: Any particular reason for using the character 'x' to initialize memory as opposed to the "pattern" member variable.

src/runtime_src/core/pcie/common/dmatest.h
sonals: Updated

src/runtime_src/core/pcie/common/dmatest.h
sonals: Updated

src/runtime_src/core/pcie/common/dmatest.h
sonals: Done

src/runtime_src/core/pcie/common/dmatest.h
sonals: Updated

src/runtime_src/core/pcie/common/dmatest.h
sonals: Updated

PR #3240, opened 2020-04-20
CR-1062426 cma temp out of range

chienwei-lan:
Don't add cma_bank to sensor_tree if not enable


chienwei-lan:
retest this please

PR #3241, opened 2020-04-20
CR-1061606 "xbutil flash scan" support for flat shell. 

saifuddin-xilinx:
This pull request has the following fixes:
1. Distinguish the flat shell in firmware_image.cpp
2. Updated pkg_hw_platform accordingly.

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbmgmt/firmware_image.cpp
+++ b/src/runtime_src/core/pcie/tools/xbmgmt/firmware_image.cpp
@@ -264,7 +264,7 @@ DSAInfo::DSAInfo(const std::string& filename, uint64_t ts, const std::string& id
         }
         // For 2RP platform, only UUIDs are provided
         //timestamp = ap->m_header.m_featureRomTimeStamp;
-        hasFlashImage = (xclbin::get_axlf_section(ap, MCS) != nullptr) ||
+        hasFlashImage = (xclbin::get_axlf_section(ap, MCS) != nullptr) || (ap->m_header.m_mode == XCLBIN_FLAT) ||
             (xclbin::get_axlf_section(ap, ASK_FLASH) != nullptr) || (xclbin::get_axlf_section(ap, PDI) != nullptr);
 
         // Find out BMC version
--- a/src/runtime_src/tools/scripts/pkg_hw_platform.sh
+++ b/src/runtime_src/tools/scripts/pkg_hw_platform.sh
@@ -181,6 +181,7 @@ xsabinOutputFile=""
 SatelliteControllerFamily=""
 CardMgmtControllerFamily=""
 SchedulerFamily=""
+SysMode="hw_pr"
 
 
 XBMGMT=/opt/xilinx/xrt/bin/xbmgmt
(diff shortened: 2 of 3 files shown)


saifuddin-xilinx:
Currently, there is no flag to distinguish between the flat shell and a normal shell. Hence, some of the xrt functionality like "xbutil flash scan" failing.  Now this is working fine.

src/runtime_src/core/pcie/tools/xbmgmt/firmware_image.cpp
```diff
@@ -264,7 +264,7 @@ DSAInfo::DSAInfo(const std::string& filename, uint64_t ts, const std::string& id
         }
         // For 2RP platform, only UUIDs are provided
         //timestamp = ap->m_header.m_featureRomTimeStamp;
-        hasFlashImage = (xclbin::get_axlf_section(ap, MCS) != nullptr) ||
+        hasFlashImage = (xclbin::get_axlf_section(ap, MCS) != nullptr) || (ap->m_header.m_mode == XCLBIN_FLAT) ||
```
AShivangi, line 267: Can we please make this change in xbmgmt2/flash/firmware_image.cpp as well?

saifuddin-xilinx: Added the fix in xmgmt2. Please review.

chvamshi-xilinx:
@rozumx , Can you please review this change (src/runtime_src/tools/scripts/pkg_hw_platform.sh)

src/runtime_src/tools/scripts/pkg_hw_platform.sh
```diff
@@ -503,12 +504,19 @@ initXsaBinEnvAndVars()
           if [ "${1}" != "NOT_DEFINED" ]; then
             SchedulerFamily="${1}"
           fi
+          
+          # Sys Mod
+          set -- `cat "${metaDataJSONFile}" | python -c "import sys, json; print json.load(sys.stdin)['xsabin'].get('Sys Mode','NOT_DEFINED')"`
```
rozumx, line 509: Why is this additional information being stored in the ext_metadata file?

This file is meant to be use only to help package additional data.  Not to drive xclbinutil option values.

chvamshi-xilinx:
> 
> 
> Please help me understand why there was a need to:
> 
>     1. Change the platform packager
>     
>     2. Change how the the platform archive is created
> 
>     3. Examine if the given xsabin (e.g., xclbin) needed to be looked at.
> 
> 
> Looking at this PR suggests a "custom" one of solution for just the U.2 platform.

Thanks @rozumx for reviewing.

1) This is not specific to U.2. This flow works for all flat shells (non-DFX).  U.2 is the only flat shell which we have today.

2) We need to know whether platfom is DFX/non-DFX to show the "flashable partiton installed on system". XSABIN should contain this information as non-DFX shells doesnt contain **MCS** in it

3) This information is extracted from ext_metadata.json as it is already used to find out the CMC , SC and ERT information to package.

Let us know if there is any other way to find out from XSABIN.

rozumx:
Vamshi,
Could you please schedule a meeting for us to discuss this solution.  Anytime you and your team is available works for me.

Thanks, Stephen

chvamshi-xilinx:
> 
> 
> Vamshi,
> Could you please schedule a meeting for us to discuss this solution. Anytime you and your team are available works for me.
> 
> Thanks, Stephen

Thanks @rozumx  for the quick call and explaining the concerns. I agree with your points. Let me explain this to platform team as well.

saifuddin-xilinx:
After discussing, it is clear that these changes may not need now. For Flat shell it is expected to see the difference between flashed image and installed image. Hence closing this pull request.

PR #3244, opened 2020-04-20
xclbinutil : Renamed xclbinutil parent directory and added boost program-options dependency.

rozumx:
Work Done
---------
+ Rename the parent directory of xclbinutil from 'xclbin' to 'xclbinutil'
+ Added to the XRT installation the dependent package boost-program-options


stsoe:
@rozumx enlighten me on << ...ONE_GREATER  what is that,  strictly less that the following version?  Why is this necessary?

rozumx:
I was hoping you could have enlighten me as well.  

**_Further Dissusion_**
For this change, I decide not too look into why the previous author went to great lengths to support this functionality.  Mainly because understanding, possibility removing, and testing this functionality would take a considerable amount of time to complete.  Comparing the priority of doing so to the current work needed for 2020.1, I instead choice to simply follow the previous pattern. 

That said, we should look into why this syntax is there, hence I have created a task (XRT-454) to investigate why post 2020.1.

rozumx:
retest this please

PR #3246, opened 2020-04-20
CR-1062551 XRT:bad_alloc error

stsoe:
MEM_STREAMING entries should be ignored when grouping banks because
their route_id may just happen to match that of a DDR/HBM base address.
The streaming entries should probably be removed all together rather
than attempting to group them, but minimize changes now by preserving
old behavior.


stsoe:
retest this please

xilinxgitops:
please retest this

chvamshi-xilinx:
retest this please

PR #3247, opened 2020-04-21
CR-1062783 Emulation processes are not closed

stsoe:
Pull request #3225 for OpenCL, removed static close of devices from
platform destructor.  This was done because the platform no longer
itself opens devices found in the system.  In OpenCL only cl_context
opens devices and the context destructor closes the devices.
Unfortunately this relies on the host code properly disposing of
OpenCL objects holding on a cl_context objects, and we have lots and
lots of host code that does not do this.

For HW runs this is not a problem because the OS itself takes care of
cleaning up remaining resources, but in emulation xclClose() must be
called in order for the underlying xsim process to exit.

The pull request adds back the missign xclClose() for the HW emulation
flow only. It might be better if HW emulation could manage the xsim
processes without relying xclClose().

### DIFF
--- a/src/runtime_src/xrt/device/hal2.cpp
+++ b/src/runtime_src/xrt/device/hal2.cpp
@@ -34,11 +34,10 @@
 namespace {
 
 static bool
-is_hw_emulation()
+is_emulation()
 {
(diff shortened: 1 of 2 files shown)


stsoe:
retest this please

PR #3251, opened 2020-04-21
CR-1062973 xbutil top prints out the CU usage

chienwei-lan:
xbutil top prints  CU usage

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbutil/xbutil.h
+++ b/src/runtime_src/core/pcie/tools/xbutil/xbutil.h
@@ -122,10 +122,10 @@ enum cmacommand {
     CMA_SIZE,
 };
 
-enum cu_stat {
-    CU_USAGE = 0x0,
-    CU_ADDR,
-    CU_STATUS,
+enum class cu_stat : unsigned short { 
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/tools/xbutil/xbutil.h
```diff
@@ -122,6 +122,12 @@ enum cmacommand {
     CMA_SIZE,
 };
 
+enum cu_stat {
+    CU_USAGE = 0x0,
+    CU_ADDR,
+    CU_STATUS,
+};
```
stsoe, line 129: When we introduce new enums in C++ code, it would be good to make them enum classes with lower case enumeration variables, e.g.
```
enum class cu_stat : unsigned short { 
  usage = 0, 
  addr, 
  stat
 };
```
Enum classes scopes the enumeration variables and there is zero risk of perprocessor interfering provided you use lower case enum variables and no one has #defined a lower case symbolic name.  It also means that you can use short  symbolic to the point names of the enum variables, since these can only be used when scoped with the class name, for example:
`if (kind == cu_stat::usage) ...`

PR #3255, opened 2020-04-22
XRT-464 enhance XRT with verbose logs

xdavidz:
print logs for every 1M data transferred, it makes our live easier for debugging.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/ospi_versal.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/ospi_versal.c
@@ -64,8 +64,6 @@
 #include "../xocl_drv.h"
 #include "xrt_drv.h"
 
-// #define OSPI_VERBOSE
-
 #define	OSPI_VERSAL_DEV_NAME "ospi_versal" SUBDEV_SUFFIX
 
 /* Timer interval of checking OSPI done */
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/ospi_versal.c
```diff
@@ -64,6 +64,8 @@
 #include "../xocl_drv.h"
 #include "xrt_drv.h"
 
+// #define OSPI_VERBOSE
```
larry9523, line 67: I think it is better not to use "c99" comments style in kernel code

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ospi_versal.c
```diff
@@ -84,6 +86,13 @@ struct ospi_versal {
 #define	OV_INFO(ov, fmt, arg...)    \
 	xocl_info(&ov->ov_pdev->dev, fmt "\n", ##arg)
 
+#ifdef OSPI_VERBOSE
```
larry9523, line 89: any concern we need to use #ifdef here? Though without this, we can turn the debug information without recompiling the code.

larry9523:
Also, you may want to update the year of copyright

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ospi_versal.c
xdavidz: Agree, so just remove the ifdef for deb logs.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ospi_versal.c
xdavidz: removed

xdavidz:
> Also, you may want to update the year of copyright

Fixed the copyright year. Also removed unnecessary ifdef.

PR #3264, opened 2020-04-23
XRT-461 Chiron bringup

houlz0507:


src/runtime_src/core/pcie/driver/linux/xocl/subdev/feature_rom.c
```diff
@@ -392,7 +392,11 @@ static int load_firmware_from_flash(struct platform_device *pdev,
 	xocl_info(&pdev->dev, "try loading fw from flash");
 
 	(void) xocl_flash_get_size(xdev, &flash_size);
-	BUG_ON(flash_size == 0);
+	if (flash_size == 0) {
+		xocl_err(&pdev->dev,
```
maxzhen, line 396: if the return code is -ENODEV, then we don't need to print out this error, right?

houlz0507: I think if platform does not have flash, it might be good to know. Maybe use xocl_info for -ENODEV case?

maxzhen: > 
> 
> I think if platform does not have flash, it might be good to know. Maybe use xocl_info for -ENODEV case?

Agree!

houlz0507: done the change.

xuhz: I just worked on a crash caused by this BUG_ON():
u25 blp pkg is not installed and flash type is qspi.
...

PR #3268, opened 2020-04-23
Trace Enhancements and Race Condition Fixes

hackwa:
1. Use double locking to make device trace lower priority offload. This means host events can still get logged while device data is being dumped.
2. Fix Race condition where big trace buffer allocation lead to delay in trace offload start.
3. Fix FIFO trace vector indexing.
4. Use references in trace writer to eliminate string copy.


abalasa:
retest this please

PR #3270, opened 2020-04-23
With HAL profiling enabled, xbutil query hits undefined symbol error while loading libxdp_hal_plugin.so

IshitaGhosh:
When xbutil query is used with HAL profiling enabled, it hits a runtime error for undefined symbol xclUnmgdPread while loading libxdp_hal_plugin.so. xclUnmgdPread is defined in xrt_core_static library and xbutil executable links with it. But xbutil executable does not export symbols as dynamic symbol by default. So, when libxdp_hal_plugin.so is loaded dynamically within xbutil, symbols like xclUnmgdPread expected from xrt_core_static remains unresolved for libxdp_hal_plugin.so. To fix this, xbutil should be enabled to export symbols for loadable modules like libxdp_hal_plugin.so. This is done by setting ENABLE_EXPORTS property to true for xbutil TARGET in CMake.

Also, during construction HAL profiler (HALPlugin) checks how many devices are available using xclOpen repeatedly till it returns nullptr. xclOpen issues "Cannot find index" when device index exceeds number of available devices. Now, when xbutil query is called with HAL profiling enabled, "Cannot find index <available_device_index+1>" message was coming up always at the end. To prevent this, I have added a verbosity check and am treating this message as XRT_INFO.


src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -2149,8 +2149,10 @@ xclDeviceHandle
 xclOpen(unsigned int deviceIndex, const char*, xclVerbosityLevel)
 {
     if(pcidev::get_dev_total() <= deviceIndex) {
+      if(XRT_INFO <= xrt_core::config::get_verbosity()) {
         printf("Cannot find index %u \n", deviceIndex);
-        return nullptr;
+      }
+      return nullptr;
```
stsoe, line 2155: Can you  use xrt_logmsg?

src/runtime_src/core/pcie/tools/xbutil/CMakeLists.txt
```diff
@@ -24,5 +24,7 @@ target_link_libraries(xbutil
   ${CURSES_LIBRARIES}
   )
 
+set_property(TARGET xbutil PROPERTY ENABLE_EXPORTS 1)
+
```
stsoe, line 28: xbutil should not export symbols.  I understand the problem, but we need to find a way to not load xdp when xbutil is running.   The proposed fix is just a work-around, isn't it?

IshitaGhosh: This export is only for the loadable modules within the executable. It helps the libraries which are dynamically loaded from within the executable. Do you see a potential issue with this?

Also, xbutil query is not loading full "xdp" library. It is loading "xdp_hal_plugin" library as it has been specifically enabled in the xrt.ini with xrt_profile=true under Debug configuration. In addition to this, currently loading of xdp_hal_plugin is actually under a compilation macro "ENABLE_HAL_PROFILING" to be defined in pcie/linux/shim.cpp. The plugin library is not loaded otherwise. As the user has specifically enabled this, we don't have a way to block it except some hardcoded executable name checks.

stsoe: Yes, I know the export thing.  I am not comfortable with this change.  dynamically loaded modules don't necessarily want to resolve symbols against the executable.  Imagine 
```
a.so -> { b.so } // a.so linked with b.so
x.exe  // exporting arbitrary symbols one of happens to be same as one in b.so
```
If x.exe dlopens a.so, will a.so resolve symbols against b.so or x.exe? 
As an example, consider b.so to be a boost library and a.so using a  different version  of boost compared to x.exe.  This could lead to segmentation fault.

I understand what you are doing, and it is safe in our context, but it is just not a good solution, and I really would like to avoid xbutil exporting symbols.

As we have talked about in the past, I don't think it makes sense to hal profile xbutil.  But assume it does and that it can't be prevented why can't xdp_hal resolve  the missing symbols by simply statically with core_static as xrt_core?

IshitaGhosh: Thanks for explaining in details.
Linking with xrt_core_static does not help. However, dynamically linking xrt_core with xdp_hal_plugin helps here.
In XDP, we had worked towards removing direct dependency and linking with xrt_core for "xdp_core", "xdp", "xdp_hal_plugin" libraries. We were only linking with xrt_coreutil.
But dynamically linking with xrt_core seems to be the only acceptable solution here.
Another possible solution would be to add the required APIs to "xrt_core::system" class. But it  may not be justifiable for this corner case.

src/runtime_src/core/pcie/linux/shim.cpp
IshitaGhosh: xrt_logmsg is not available in the stub function xclOpen. 
xrt_logmsg is a private member method of xrt_core::xocl::shim class. As this a negative scenario, we may not want to create an object of xocl::shim here.

stsoe: Ok. can you use xrt_core::message::send then?

IshitaGhosh: Yes, added.

PR #3278, opened 2020-04-25
add kds_echo param for test purpose

mamin506:
As per Max's request, add kds_echo parameter to make KDS doesn't touch hardware. With this, it is easier to measure the software overhead in scheduling.

This is only for test purpose. It is not a product level support.


mamin506:
> I don't think the extra level of scoping is cool; looks really bad. How about overriding iowrite and read to do the check but leave main flow alone. The override of ioread32 should have a second argument for the short circuited returned value.

Will do.

PR #3283, opened 2020-04-27
deprecated more xcl APIs

maxzhen:
Obsoleting the following XRT public APIs from xrt.h. These instead should be private hooks for our tools xbutil/xbmgmt as we do not want any application call these:
a.	xclUpgradeFirmware()
b.	xclUpgradeFirmware2()
c.	xclUpgradeFirmwareXSpi()
d.	xclBootFPGA()
e.	xclRemoveAndScanFPGA()
f.      xclResetDevice()

### DIFF
--- a/src/runtime_src/core/include/xrt.h
+++ b/src/runtime_src/core/include/xrt.h
@@ -300,7 +300,7 @@ xclClose(xclDeviceHandle handle);
  * NOTE: Only implemeted Reset kind through user pf is XCL_USER_RESET
  *
  * This API is deprecated and will be removed in future release.
- * New clients should run xbutil reset instead
+ * New clients should use xbutil utility to reset device instead
  */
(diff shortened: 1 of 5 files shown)


src/runtime_src/core/include/xrt.h
```diff
@@ -299,6 +299,7 @@ xclClose(xclDeviceHandle handle);
  * running kernel(s) to finish.
  * NOTE: Only implemeted Reset kind through user pf is XCL_USER_RESET
```
sonals, line 300: This API is deprecated. Please use xbutil utility to reset device.

maxzhen:
> 
> 
> For each deprecated API we should put a one line note saying how to achieve the functionality using XRT tools?

Added notes for these APIs. Please take a look.

PR #3290, opened 2020-04-28
CR-1063635 Updating the Variable name used my emulation for enabling P2P feature

hemantk-xilinx:
Fixing the env variable used to control P2P implementation in HW emulation


chvamshi-xilinx:
ok to test

PR #3297, opened 2020-04-29
CR-1041924 Adding timeout support for clPollStream in emulation

hemantk-xilinx:
Adding timeout support for clPollStream API in emulation which will be enabled through a ini switch
[Emulation]
timeout_scale=na/ms/sec/min
na -> timeout disabled (default)
ms- > timeout scale is milliseconds
sec -> timeout scale is second
min -> timeout scale is minutes

Scaling is needed as HW based timeout will not work for emulation which is usually millisec


chvamshi-xilinx:
ok to test

PR #3304, opened 2020-04-30
XRT-504 Refine CU thread

mamin506:
Use 4 lists, pending queue, run queue, submitted queue and completed queue
One process function per queue
avoid call kthread_should_stop() frequently
More comments to explain the full name of pq, rq, sq and cq, which make shorter line possible :)

I didn't see obvious change of "OPS" on test machine (kds_mode=1 and kds_echo=1). This is better code and same performance.

### DIFF
--- a/src/runtime_src/core/common/drv/xrt_cu.c
+++ b/src/runtime_src/core/common/drv/xrt_cu.c
@@ -74,8 +74,9 @@ int xrt_cu_thread(void *data)
 {
 	struct xrt_cu *xcu = (struct xrt_cu *)data;
 	unsigned long flags;
+	int ret = 0;
 
-	while (1) {
+	while (!xcu->stop) {
 		// Check num_pq here
 		spin_lock_irqsave(&xcu->pq_lock, flags);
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/common/drv/xrt_cu.c
```diff
@@ -9,107 +9,147 @@
...
+	/* Add command to pending queue
+	 * wakeup CU thread if it is the first command
+	 */
 	spin_lock_irqsave(&xcu->pq_lock, flags);
 	list_add_tail(&xcmd->list, &xcu->pq);
-	if (xcu->num_pq == 0)
-		up(&xcu->sem);
 	++xcu->num_pq;
+	if (xcu->num_pq == 1)
+		up(&xcu->sem);
```
maxzhen, line 152: you could cache the result of "xcu->num_pq == 1" and do the up after releasing the lock.

mamin506: done

PR #3314, opened 2020-05-01
CR-1063652 System crash on multi-threaded host accessing multiple eng…

maxzhen:
…ines in same device (part 1)

This change fixes the crash described in CR-1063652.

KDS does not release exec buf BOs after application is killed. In the very next hot reset the system will crash when KDS is freeing left over exec buf BOs due to that it's torn down after DRM is gone.

Note that those left over BOs are not for CU cmds that are stuck on device. They are BOs for CU cmds that have not submitted to device yet. CU cmds that are on device is aborted and freed properly when application is killed.

The fix is simply freeing all stale CU cmds when KDS is being reset, no matter what stage they are on.

Tested with steps described in the CR and the crash is not seen. After hot reset, board can be recovered properly. Tested with full xbutil validate and it worked fine.


xilinxgitops:
retest this please

chvamshi-xilinx:
retest this please

PR #3319, opened 2020-05-01
Change_Windows_ExportBO_input_Param

igorshr:
CR-1040791. ExportBO was missed as an Export DLL function. Needed to correct an input parameter. XRT build is passed. Tested by Sid's 30vadd_c application.

### DIFF
--- a/src/runtime_src/core/pcie/windows/shim.cpp
+++ b/src/runtime_src/core/pcie/windows/shim.cpp
@@ -1414,11 +1414,11 @@ xclExecWait(xclDeviceHandle handle, int timeoutMilliSec)
   return shim->exec_wait(timeoutMilliSec);
 }
 
-int xclExportBO(xclDeviceHandle handle, unsigned int boHandle)
+int xclExportBO(xclDeviceHandle handle, xclBufferHandle boHandle)
 {
   xrt_core::message::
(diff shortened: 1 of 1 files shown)


sonals:
ok to test

src/runtime_src/core/pcie/windows/shim.cpp
```diff
@@ -1414,11 +1414,11 @@ xclExecWait(xclDeviceHandle handle, int timeoutMilliSec)
   return shim->exec_wait(timeoutMilliSec);
 }
 
-int xclExportBO(xclDeviceHandle handle, unsigned int boHandle)
+int xclExportBO(xclDeviceHandle handle, xclBufferHandle boHandle)
 {
   xrt_core::message::
     send(xrt_core::message::severity_level::XRT_DEBUG, "XRT", "xclExportBO() NOT IMPLEMENTED");
-  return ERROR_INVALID_FUNCTION;
+  return EDOM;
```
arpitxilinx, line 1421: Minor comment - 
Can we return instead of EDOM to ENOSYS or ENOTSUP, which would be more appropriate.

igorshr:
Changed the return code to ENOSYS

igorshr:
Could we merge this fix?

PR #3328, opened 2020-05-04
CR-1028065 fixed missing return code of clPollStream & clSetStreamOpt()

karenx-xilinx:
Make sure the return code is set in errcode_ret.
Also checking for negative min_num_completion in clPollStream()


stsoe:
retest this please

PR #3331, opened 2020-05-04
added missing check for execbuf waitlist

sarab96:
added missing check for execbuf waitlist


src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -76,6 +76,9 @@
 
 #define SHIM_QDMA_AIO_EVT_MAX   1024 * 64
 
+//#include "core/pcie/driver/linux/xocl/userpf/common.h"
+//MAX_DEPs copied from above include file.
+#define MAX_DEPS        8
```
maxzhen, line 81: can we put MAX_DEPS in a header file that is common for both driver and user space code?

PR #3335, opened 2020-05-04
Add builds.sh switch to disable unit test

stsoe:


maxzhen:
is ctest causing any trouble so that we would like to disable it?

stsoe:
Yes, it is preventing my local build from completing; I have changes that break the units tests.  Until I look closer at these,  I need a way to disable unit tests temporarily without having to edit build.sh.

PR #3337, opened 2020-05-04
Updated xclIPName2Index() API to return index or error as return valu…

sonals:
…e in line with XRT APIs


chvamshi-xilinx:
We may need to post this change in xrt_general as this API is used by several people.

PR #3339, opened 2020-05-05
CR-1064295 Stop Buffer deallocation emulation calls for versal

hemantk-xilinx:
Versal emulation does not support P2P buffer allocation/deallocation


chvamshi-xilinx:
ok to test

PR #3343, opened 2020-05-05
XRT-520 huge page dma_map_sg setup

chienwei-lan:


src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
```diff
@@ -860,40 +868,13 @@ int xocl_init_mem(struct xocl_drm *drm_p)
 	return err;
```
maxzhen, line 868: before return, you might want to set cma_mem->sgt, cma_mem->pages and cma_mem_vaddr to NULL.

xilinxgitops:
retest this please

PR #3345, opened 2020-05-06
NSPR-1387 : Xrt hbm support 

saifuddin-xilinx:
Hi, 
Creating this PR for review this HBM changes. 
Please review the changes follows:
   1. Userspace changes for XRT layer
   2. User_PF Driver changes for HBM support

New changes support the following functionality:
1. User can specify a range of HBM banks in the v++ connectivity section, Example: 
    -sp CU_0.in1:HBM[0:3] . The user  needs to specify bank-id information in his host application 
    as well using Extension Pointer.
2. If the user doesn't specify anything through the Extension pointer, Then XRT should automatically 
    detect the HBM banks using the connectivity section present in rtd file.

### DIFF
--- a/src/runtime_src/core/include/xrt.h
+++ b/src/runtime_src/core/include/xrt.h
@@ -717,9 +717,25 @@ xclExportBO(xclDeviceHandle handle, xclBufferHandle boHandle);
  * Import a BO exported by another device.     *
  * This operation is backed by Linux DMA-BUF framework
  */
+
 XCL_DRIVER_DLLESPEC
 xclBufferHandle
 xclImportBO(xclDeviceHandle handle, int fd, unsigned int flags);
+
+/**
+ * xclGetBOGroup() - Obtain Memory Group id for the argument
+ *
+ * @handle:        Device handle
+ * @argidx:        Argument index
+ * @cuidx:         CU index
+ * Return:         -1 on failure
+ *                 +ve group index
+ *
+ * This is the prefered method for obtaining Group id for the given argument.
+ */
+XCL_DRIVER_DLLESPEC
+int
+xclGetBOGroup(xclDeviceHandle handle, unsigned int cuidx, unsigned int argidx);
 
 /**
  * xclGetBOProperties() - Obtain xclBOProperties struct for a BO
--- a/src/runtime_src/core/include/xrt.h
+++ b/src/runtime_src/core/include/xrt.h
@@ -242,6 +242,18 @@ struct xclDeviceUsage {
     uint64_t memSize[XCL_DEVICE_USAGE_COUNT];
 };
 
+struct xcl_mem_group {
+    uint32_t cu_id;
+    uint32_t arg_id;
+    uint32_t grp_id;
+};
+
+#define XCL_MEM_GROUP_MAX 128
+struct xcl_mem_conn {
+    uint32_t m_count;
+    struct xcl_mem_group *m_conn[XCL_MEM_GROUP_MAX];
+};
+
 struct xclBOProperties {
     uint32_t handle;
     uint32_t flags;
--- a/src/runtime_src/core/include/xrt_mem.h
+++ b/src/runtime_src/core/include/xrt_mem.h
@@ -38,6 +38,10 @@ extern "C" {
 #define	XCL_BO_FLAGS_P2P		(1 << 30)
 #define	XCL_BO_FLAGS_EXECBUF		(1 << 31)
 
+/* This Flag is used to identify between BANK or GROUP specified for memory allocation */
+#define XCL_FLAGS_BANK_ONLY     (1 << 16)
+#define XRT_FLAGS_BANKIDX_MASK  (0xFFFFUL)
+
 /**
  * This is the legacy usage of XCL DDR Flags.
  *
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
@@ -617,8 +617,13 @@ int xocl_mm_insert_node_range(struct xocl_drm *drm_p, unsigned user_flags,
        
         start_addr_range = mem_info->start_addr;
         end_addr_range = mem_info->end_addr;
+#if defined(XOCL_DRM_FREE_MALLOC)
         err = drm_mm_insert_node_in_range(drm_p->mm, node, size, PAGE_SIZE, 0, 
-                start_addr_range, end_addr_range, 0);
+			start_addr_range, end_addr_range, 0);
+#else
+	err = drm_mm_insert_node_in_range(drm_p->mm, node, size, PAGE_SIZE,
+			start_addr_range, end_addr_range);
+#endif
         
         *ddr_id = mem_id;
     }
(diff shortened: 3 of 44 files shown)


chvamshi-xilinx:
This PR is anyway for 2020.1_PU1. Lets have a meeting to go thru all changes.

saifuddin-xilinx:
> Just want to clarify how I think grp ids should map to mem_topology sections:
> 
> A grp id should match a mem_topology entry. For example,
> 
> * Assume we have hbm[0] through hbm[6] used and in correspondingly indexed mem_topology entries.
> * Assume that [0:2] and [5:6] are grouped.
> * I’d expect the driver to advertise 4 groups with ids [0,3,4,5].   Grp ids are not necessarily consecutive.
> 
> With this scheme user space can use the group ids as if they were mem topology indices when it is correlating CU connectivity and allocating buffers.
> 
> Say CU0 is connected to mem_topology entry 5 or 6, any connection for that CU will have grpid 5 and any buffer created for that CU will be allocated by passing ‘5’ to xclAllocBO.
> 
> If CU1 connected to 0,1, or 2, any connection of that CU will have grp id 0 and buffers created for this CU will be allocated by passing ‘0’ to xclAllocBO.
> 
> All OCL code remains unchanged, from OCL’s point of view CU0 is connected to mem_topology index 5 even though under the hood it is connected to 5 or 6. CU1 is connected to mem_topology index 0, even though under the hood it is connected to 0,1, or 2. The changes in OCL for these are none at all (except for the memidx -> grpid mapping), see for example:
> 
> * #3193
> * #3246

If I understand your comment correctly, I believe there can be a issue while the BANKs are overlapped between multiple groups. 
Let’s take an example, 
If we have three groups, i.e. [0:2], [0:5], [5:6]
•	What would be the id’s of the groups 
•	In this case, how can we map the groups with the indexed mem_topology entries

stsoe:
> If I understand your comment correctly, I believe there can be a issue while the BANKs are overlapped between multiple groups.
> Let’s take an example,
> If we have three groups, i.e. [0:2], [0:5], [5:6]
> • What would be the id’s of the groups
> • In this case, how can we map the groups with the indexed mem_topology entries
Hi @saifuddin-xilinx.   In your example  the mem_topology section in the xclbin has 7 entries, right?   But what is the base address and size of mem_topology entry with index 0?  Does the size span [0:2] or [0:5] ?   It can’t be both, it must be one or the other.   Similarly what is the base address of mem_topology.m_data[5] ?

I don’t see how overlapping groups make sense?

src/runtime_src/core/include/xrt_mem.h
```diff
@@ -38,6 +38,10 @@ extern "C" {
 #define	XCL_BO_FLAGS_P2P		(1 << 30)
 #define	XCL_BO_FLAGS_EXECBUF		(1 << 31)
 
+/* This Flag is used to identify between BANK or GROUP specified for memory allocation */
+#define XCL_FLAGS_BANK_ONLY     (1 << 16)
+#define XRT_FLAGS_BANKIDX_MASK  (0xFFFFUL)
```
maxzhen, line 43: i'm not sure if we ever need to identify b/w bank or group?

src/runtime_src/core/include/xrt.h
```diff
@@ -717,10 +729,26 @@ xclExportBO(xclDeviceHandle handle, xclBufferHandle boHandle);
...
+ * @argidx:        Argument index
+ * @cuidx:         CU index
+ * Return:         -1 on failure
+ *                 +ve group index
+ *
+ * This is the prefered method for obtaining Group id for the given argument.
+ */
+XCL_DRIVER_DLLESPEC
+int
+xclGetBOGroup(xclDeviceHandle handle, unsigned int cuidx, unsigned int argidx);
```
maxzhen, line 750: why would a normal application wants to call this API? shouldn't it just be an internal one?

src/runtime_src/core/pcie/driver/linux/xocl/xocl_drm.h
```diff
@@ -52,12 +54,96 @@ struct xocl_cma_bank {
...
+
+struct xocl_mem_bank_info {
+    uint32_t bank_id;
+    uint64_t start_addr;
+    uint64_t end_addr;
+    uint32_t grp_id;
+    size_t   size;
+};
+
+struct xocl_mem_bank {
```
maxzhen, line 107: defined above already?

src/runtime_src/core/pcie/driver/linux/xocl/xocl_drm.h
```diff
@@ -52,12 +54,96 @@ struct xocl_cma_bank {
...
+    uint32_t b_count;
+    struct xocl_mem_bank_info *m_bank[DRM_XOCL_MEM_GROUP_MAX];
+};
+
+struct xocl_mem_range {
```
maxzhen, line 112: defined above already?

src/runtime_src/core/pcie/driver/linux/xocl/xocl_drm.h
```diff
@@ -52,12 +54,96 @@ struct xocl_cma_bank {
...
+    uint32_t l_bank_idx;
+    uint64_t l_start_addr;
+    uint32_t h_bank_idx;
+    uint64_t h_end_addr;
+};
+
+struct xocl_mem_group {
```
maxzhen, line 119: defined above already?

src/runtime_src/core/pcie/driver/linux/xocl/xocl_drm.h
```diff
@@ -52,12 +54,96 @@ struct xocl_cma_bank {
...
+    uint32_t g_count;
+    struct xocl_mem_range *m_range[DRM_XOCL_MEM_GROUP_MAX];
+};
+
+struct xocl_mem_conn_map {
```
maxzhen, line 124: defined above already?

src/runtime_src/core/pcie/driver/linux/xocl/xocl_drm.h
```diff
@@ -52,12 +54,96 @@ struct xocl_cma_bank {
...
+    uint32_t cu_id;
+    uint32_t arg_id;
+    uint32_t grp_id;
+};
+
+struct xocl_mem_conn {
```
maxzhen, line 130: defined above already?

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
```diff
@@ -512,13 +514,63 @@ void xocl_mm_get_usage_stat(struct xocl_drm *drm_p, u32 ddr,
...
 
-void xocl_mm_update_usage_stat(struct xocl_drm *drm_p, u32 ddr,
-	u64 size, int count)
+int get_start_bank_id(struct xocl_drm *drm_p, u64 addr)
 {
-	BUG_ON(!drm_p->mm_usage_stat[ddr]);
+    size_t bank_size = 0;
+	uint64_t start_addr = 0;
+    struct xocl_mem_bank *mem_bank;
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 7, 0)
```
maxzhen, line 522: we never intend to support < 3.7.0 kernel anyway.

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
```diff
@@ -512,13 +514,63 @@ void xocl_mm_get_usage_stat(struct xocl_drm *drm_p, u32 ddr,
...
+    struct xocl_mm_wrapper *wrapper;
+#else
+    int i;
+#endif
+    
+    mem_bank = drm_p->m_connect->mem_bank;
+    /* As we support only equal size bank */
```
maxzhen, line 529: this assumption looks strange. why?

saifuddin-xilinx:
Reopen an another pull request for the new design.
NSPR-1387 : Xrt hbm support #3456

PR #3355, opened 2020-05-07
xbmgmt2 bug fixes

AShivangi:
- When the device is not specified, the following is printed

```
$ xbmgmt --new program --update
 Involking next generation of the xbmgmt application

ERROR: Device not specified.

List of available devices:
[0000:b3:00.0] : xilinx_u200_xdma_201830_2
```
- Fixed misleading "successfully flashed" message when the card isn't flashed

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
@@ -66,7 +66,7 @@ namespace {
  * Update shell on the board for auto flash
  */
 static void 
-update_shell(unsigned int  index, const std::string& primary, const std::string& secondary)
+update_shell(unsigned int index, const std::string& primary, const std::string& secondary)
 {
   Flasher flasher(index);
   if(!flasher.isValid())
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -76,17 +76,20 @@ update_shell(unsigned int  index, const std::string& primary, const std::string&
...
   if (!secondary.empty()) {
     sec = std::make_unique<firmwareImage>(secondary.c_str(),
       MCS_FIRMWARE_SECONDARY);
     if (sec->fail())
       sec = nullptr;
   }
-
-  flasher.upgradeFirmware("", pri.get(), sec.get());
+  
+  if (flasher.upgradeFirmware("", pri.get(), sec.get()) != 0)
```
rozumx, line 90: Minor correction. 
For calls like this it might be best if we simply do the following:
` if (!flasher.upgradeFirmware("", pri.get(), sec.get()))`
Doing so removes the "magic" number and implies we are only interested in failures.  

Unfortunately, the underlying code uses "error codes" and not exceptions so information is lost.  All we know that an error occurred.   For the next release, we should look into update the underlying code to instead:
1. Throw exceptions.
2. Have the exception caught here, reported, and a new exception thrown.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -110,16 +113,19 @@ update_shell(unsigned int index, const std::string& flashType,
...
-  std::unique_ptr<firmwareImage> sec;
   if (pri->fail())
     throw xrt_core::error(boost::str(boost::format("Failed to read %s") % primary));
+
+  std::unique_ptr<firmwareImage> sec;
   if (!secondary.empty()) {
     sec = std::make_unique<firmwareImage>(secondary.c_str(), MCS_FIRMWARE_SECONDARY);
     if (sec->fail())
-      sec = nullptr;
+      throw xrt_core::error(boost::str(boost::format("Failed to read %s") % secondary));
```
rozumx, line 123: Not required
For readability it sometimes is best to split functionality between two lines, for example:
```
   if (sec->fail()) {
      const std::string errMsg = boost::str(boost::format("Failed to read %s") % secondary);
      throw xrt_core::error(errMsg);
  }
```

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
AShivangi: I agree!

PR #3360, opened 2020-05-08
XRT-531 Zocl panic when outstanding command is not completed.

larry9523:
Skip print free-running CU because there is no zcu_core structure allocated.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_cu.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_cu.c
@@ -100,6 +100,9 @@ zocl_cu_status_print(struct zocl_cu *cu)
 {
 	struct zcu_core *cu_core = cu->core;
 
+	if (!cu_core)
+		return;
+
 	DRM_INFO("addr 0x%llx, status 0x%x",
 	    (u64)cu_core->paddr, ioread32(cu_core->vaddr));
 }


akashm-xilinx:
retest this please

larry9523:
> I think the panic is due to the cu is null not the cu->core is null?
> Otherwise, how if (!NULL) can panic the system?

I am confused.
1) the panic is due to cu->core is null but not cu is null. We allocate exec->zcu for every cu, valid or invalid. that is why exec->zcu[0,..ncu-1] are all non pointer. But we only allocate cu_core structure whenever the cu is valid.
2) Not sure I understand "how if (!NULL) can panic the system". It is the fix to avoid the panic, isn't it? The real problem is cu_core is null, so trying to pring cu_core->paddr, panic the system

xdavidz:
> > I think the panic is due to the cu is null not the cu->core is null?
> > Otherwise, how if (!NULL) can panic the system?
> 
> I am confused.
> 
> 1. the panic is due to cu->core is null but not cu is null. We allocate exec->zcu for every cu, valid or invalid. that is why exec->zcu[0,..ncu-1] are all non pointer. But we only allocate cu_core structure whenever the cu is valid.
> 2. Not sure I understand "how if (!NULL) can panic the system". It is the fix to avoid the panic, isn't it? The real problem is cu_core is null, so trying to pring cu_core->paddr, panic the system

ou, I see, ok

PR #3362, opened 2020-05-08
CR-1064049 Added an exception instead of just cout message if any of the System operations fails

venkatp-xilinx:
Added an exception instead of just cout message if any of the System operations fails.

Reviewer: @vboggara-xilinx 
CR: 1064049

### DIFF
--- a/src/runtime_src/core/pcie/emulation/common_em/system_utils.cxx
+++ b/src/runtime_src/core/pcie/emulation/common_em/system_utils.cxx
@@ -13,62 +13,74 @@ namespace systemUtil {
   
   void makeSystemCall (std::string &operand1, systemOperation operation, std::string operand2 )
   {
-    switch (operation) {
-
-      case CREATE :
-        {
-          if (boost::filesystem::exists(operand1) == false)
+    std::string operationStr;
+    try {
+      switch (operation) {
+      
+        case CREATE :
           {
-            boost::filesystem::create_directories(operand1);
+            operationStr = "CREATE";
+            if (boost::filesystem::exists(operand1) == false)
+            {
+              boost::filesystem::create_directories(operand1);
+            }
+            break;
           }
-          break;
-        }
-      case REMOVE :
-        {
-          if (boost::filesystem::exists(operand1) )
+        case REMOVE :
           {
-            boost::filesystem::remove_all(operand1);
+            operationStr = "REMOVE";
+            if (boost::filesystem::exists(operand1) )
+            {
+              boost::filesystem::remove_all(operand1);
+            }
+            break;
           }
-          break;
-        }
-      case COPY   :
-        {
-          std::stringstream copyCommand;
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/pcie/emulation/common_em/system_utils.cxx
```diff
@@ -3,10 +3,19 @@
...
+        throw status; 
+      } else {
+        return;
+      }
+    } 
+    catch (int status) { 
+      std::cout << "ERROR: [EMU 60-600] " << command.c_str() << " Exception Caught - Failed with the error code " << status << ". PLEASE CHECK YOUR PERMISSIONS " << std::endl;
+    } catch (...) {
+      std::cout << "ERROR: [EMU 60-600] " << command.c_str() << " Exception Caught - Failed with the error code " << status << ". PLEASE CHECK YOUR PERMISSIONS " << std::endl;
+    }
```
stsoe, line 18: I don''t understand this code at all?   Throwing the status and catching it is a no-op and second catch is never applicable.

src/runtime_src/core/pcie/emulation/common_em/system_utils.cxx
```diff
@@ -3,10 +3,19 @@
...
   }
   
   void makeSystemCall (std::string &operand1, systemOperation operation, std::string operand2 )
```
stsoe, line 21: I don't see operand1 modified by this function?   And why is operand2 passed by value?

src/runtime_src/core/pcie/emulation/common_em/system_utils.cxx
venkatp-xilinx: removed the second catch which never hits.

src/runtime_src/core/pcie/emulation/common_em/system_utils.cxx
venkatp-xilinx: corrected the operand to pass by reference

src/runtime_src/core/pcie/emulation/common_em/system_utils.cxx
```diff
@@ -2,11 +2,18 @@
...
+    try { 
+      if (status) { 
+        throw status; 
+      } else {
+        return;
+      }
+    } 
+    catch (int status) { 
+      std::cout << "ERROR: [EMU 60-600] " << command.c_str() << " Exception Caught - Failed with the error code " << status << ". PLEASE CHECK YOUR PERMISSIONS " << std::endl;
+    }
```
stsoe, line 16: Hi @venkatp-xilinx Still makes no sense to me can you explain the reason for this try/catch?  It's a no-op, what's the point throwing and catching immediately?

venkatp-xilinx: Thanks for the suggestion. My requirement is to immediate exit from the flow when any of the system operation fails. Now I have corrected the code accordingly

PR #3363, opened 2020-05-08
Remove deprected APIs from python bindings all together

sonals:


tests/pyopencl/23_bandwidth.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/python3
```
AShivangi, line 1: should we remove this? We should let xbutil decide which pyhton to run

sonals: shebang should not apply to xbutil? This is only relevant if someone attempts to directly run the script like `./22_bandwidth.py`

PR #3364, opened 2020-05-08
Deprecate legacy APIs

sarab96:
deprecate legacy APIs


sarab96:
changes done as per comments above

PR #3370, opened 2020-05-08
XRT-535 Unify slave bridge Host memory management

chienwei-lan:


src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -213,7 +213,7 @@ int main(int argc, char *argv[])
         return xcldev::xclReset(argc, argv);
     } else if( std::strcmp( argv[1], "p2p" ) == 0 ) {
         return xcldev::xclP2p(argc, argv);
-    } else if( std::strcmp( argv[1], "cma" ) == 0 ) {
+    } else if( std::strcmp( argv[1], "host" ) == 0 ) {
```
maxzhen, line 216: "host" is too generic, let's call it "host_mem" which matches your error msg.

PR #3371, opened 2020-05-09
XRT-536 After reboot programming PLP immediately could fail

houlz0507:


maxzhen:
lro->rp_program should be naturally aligned, which means that reading from it or writing to it is atomic already. i'm not sure what can be fixed by explicitly making it atomic?

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
```diff
@@ -708,7 +708,11 @@ int xocl_refresh_subdevs(struct xocl_dev *xdev)
 		offset += resp->size;
 	} while (resp->rtncode == XOCL_MSG_SUBDEV_RTN_PARTIAL);
 
-	if (resp->rtncode == XOCL_MSG_SUBDEV_RTN_UNCHANGED &&
+	if (resp->rtncode == XOCL_MSG_SUBDEV_RTN_PENDINGPLP) {
+		(void) xocl_program_shell(xdev, true);
+		ret = -EAGAIN;
+		goto failed;
+	} else if (resp->rtncode == XOCL_MSG_SUBDEV_RTN_UNCHANGED &&
```
maxzhen, line 715: I'd suggest you also explicitly check for XOCL_MSG_SUBDEV_RTN_COMPLETE return code so that next time xclmgmt return something new, old xocl can still gracefully error out.

houlz0507: added the check

houlz0507:
retest this please

houlz0507:
Installed this pr locally and reboot. I did not see any issue. I kick off a re-run and see if it reproduces.

PR #3372, opened 2020-05-10
Build on Ubuntu 20.04

stsoe:
These are quick fixes to build XRT on Ubuntu 20.04. Python is a bit of a mess and needs to be cleaned up in xrtdeps.sh, cpackLin.cmake, and postinst.in.


stsoe:
These are quick fixes to build XRT in 20.04.  Python is a bit of a mess and needs to be cleaned up in xrtdeps.sh, cpackLin.cmake, and postinst.in.

stsoe:
retest this please

PR #3373, opened 2020-05-11
CR-1032457 Display warning if OS or Kernel isn't officially supported

AShivangi:
removed 5.3.0


maxzhen:
Marked as do-not-merge until we hear back from PRT.

PR #3380, opened 2020-05-11
XmaEncoder Properties: Updates need for Custom Rate Control integration with Look Ahead

rathaval:
This pull-request contains updates to both xma and xma_legacy XmaEncoder properties to absorb lookahead depth and custom-rate-control flags.

- The Lookahead IP collects statistics about the next N frames (N could be between 1 and 30).
- Custom Rate Control utilizes this statistics, sends Frame SAD &Frame Activity to the U30 soft-kernels
- For low latency use-cases the Look Ahead Depth could be 1 frame
- For regular latency use-case the Look-Ahead Depth could be 16, 18 or 20 frames.
-  Custom RC = 1 (Enabled), XMA Plugins need to send this enable/disable flag to U30 soft-kernels

Such details come from the FFmpeg command line arguments, and need to be captured and made a part of the Encoder Properties for the plugins to consume and take appropriate action.

### DIFF
--- a/src/xma/include/app/xmaencoder.h
+++ b/src/xma/include/app/xmaencoder.h
@@ -118,6 +118,8 @@ typedef struct XmaEncoderProperties
     /** aq_mode  */
     int32_t         aq_mode;
     int32_t         minQP;
+    /** Rate Control Mode for Custom Rate Control **/
+    int32_t 	    rc_mode;
     /** force property values to be accepted by encoder plugin */
     int32_t         force_param;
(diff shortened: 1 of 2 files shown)


rathaval:
@sarab96 
Sarab please take a look at this PR and let me know what you think of this for XRT/master ?

sarab96:
This will break compatibility of legacy compiled applications and will require recompile. For this release legacy changes which break compatibility are not recommended.

PR #3384, opened 2020-05-12
CR-1065045 xilinx_vck5000-es1 or xilinx_vck5000_es1 will confuse xbmg…

xdavidz:
this fixes the issue, but not sure if we should accept this solution?


xdavidz:
put onto shelf, platform team can rollback their changes for 2020.1 temporarily

PR #3388, opened 2020-05-12
XRT-393 Basic Infrastucture to support xbutil2 on Edge Platform

rbramand-xilinx:
> This pull request contains changes to support xbutil2 on edge platform.
> As edge platforms does not have mgmtpf added check so that xbmgmt and xbmgmt2 code doesnot build for edge.


src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -30,9 +31,159 @@ namespace {
...
+  {
+    auto edev = get_edgedev(device);
+    switch (key)
+    {
+    case key_type::edge_vendor:
+      return deviceInfo.mVendorId;
+    case key_type::rom_vbnv:
+      return std::string(deviceInfo.mName);
+    case key_type::rom_ddr_bank_size:
+      return (deviceInfo.mDDRSize >> 30);
```
rozumx, line 58: Magic numbers.  Could you define what is happening here.

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -30,9 +31,159 @@ namespace {
...
+    get(const xrt_core::device* device,key_type key)
+  {
+    auto edev = get_edgedev(device);
+    switch (key)
+    {
+    case key_type::edge_vendor:
+      return deviceInfo.mVendorId;
+    case key_type::rom_vbnv:
+      return std::string(deviceInfo.mName);
+    case key_type::rom_ddr_bank_size:
```
rozumx, line 57: For this key, it might be best to rename it to represent what the "size" represents.  Is it Kbytes, bytes, MBytes, etc.

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -30,9 +31,159 @@ namespace {
...
+      return (deviceInfo.mDDRSize >> 30);
+    case key_type::rom_ddr_bank_count_max:
+      return static_cast<uint64_t>(deviceInfo.mDDRBankCount);
+    case key_type::clock_freqs:
```
rozumx, line 61: Same here, is the frequency Hhz, MHz, etc.

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -30,9 +31,159 @@ namespace {
...
+      {
+        std::vector<std::string> clk_freqs;
+        for(int i = 0; i < 3; i++)
```
rozumx, line 64: Magic numbers.  What does '3' represent.  You might wish to consider determining the size directly from the structure.  For example: sizeof(xclDeviceInfo2.mOCLFrequency)

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -30,9 +31,159 @@ namespace {
...
+          clk_freqs.push_back(std::to_string(deviceInfo.mOCLFrequency[i]));
+        return clk_freqs;
+      }
+    default:
+      return std::string("NA");
```
rozumx, line 69: Shouldn't an exception be thrown here.

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -30,9 +31,159 @@ namespace {
...
+};
+
+template <typename QueryRequestType>
+struct sysfs_getter : QueryRequestType
+{
+  const char* entry;
+
+  sysfs_getter(const char* e)
+    : entry(e)
+  {}
```
rozumx, line 130:
```
For empty function bodies, it is best practice to add "/* empty */' to the body.  For example
 sysfs_getter(const char* e)
    : entry(e)
  { /* empty */ }
```

src/runtime_src/core/tools/xbutil2/XBReport.cpp
```diff
@@ -31,6 +31,15 @@ using namespace XBReport;
 
 
 // ------ F U N C T I O N S ---------------------------------------------------
+void
+report_driver_version(boost::property_tree::ptree _pt, std::string pt_var, std::string msg_var)
+{
+  if (_pt.get<std::string>(pt_var, "NA") != std::string("NA")) {
```
rozumx, line 37: The previous author of the code used "N/A" and here "NA" is being used.   Did you really intend to change the default return value to something else.  If so, what is the reasoning behind it.

src/runtime_src/core/edge/user/device_linux.cpp
rbramand-xilinx: Here I am using the key which is already defined in query_requests.h and as the to_string method in that key structure uses  "value << 30" we can infer that it is expecting a Gbytes value which  it is trying to convert into bytes by using left shift.
I dint change the name as it is already in use by pcie platform.

src/runtime_src/core/edge/user/device_linux.cpp
rbramand-xilinx: As stated the value expected is in GBytes and I am reading the value in bytes, converted it to expected value.

src/runtime_src/core/edge/user/device_linux.cpp
rbramand-xilinx: The value is in MHz, again the key was already defined so was using it, to avoid confusion we can rename the key which suits for both edge and pcie platforms.

src/runtime_src/core/edge/user/device_linux.cpp
rozumx: O.K. Then we need to update the PCI platforms as well.

src/runtime_src/core/edge/user/device_linux.cpp
rozumx: Please remove the magic number and comment this code.
For example:
```
case key_type::rom_ddr_bank_size:
{ 
  static const uint32_t BYTES_TO_GBYTES = 30;
  return (deviceInfo.mDDRSize >> BYTES_TO_GBYTES);
}
```

src/runtime_src/core/edge/user/device_linux.cpp
rozumx: Please update the names.

PR #3390, opened 2020-05-12
Fix CR-1054554: Change the threshold value based on the platform

haeseunglee:
Fix for CR-1054554: Change the threshold value based on the platform


maxzhen:
ok to test

PR #3400, opened 2020-05-13
CR-1064997, CR-1063824

AShivangi:
- CR-1063824: Remove 'Experts Only' commands from xbutil help message
- CR-1064997: promote xbmgmt partition and config sub-commands for external use
- CR-1065226: Cannot upgrade XRT to 2.6-576 on Centos 7.5


maxzhen:
retest this please

PR #3401, opened 2020-05-13
HAL profiling fixes

IshitaGhosh:
1. Performance improvement : Use of xclGetDebugIPlayoutPath to determine unique device ID causes slow down. To avoid calling xclGetDebugIPlayoutPath everytime in HALPlugin::getDeviceId, use and check a map of xclDeviceHandle and its unique device ID. Call xclGetDebugIPlayoutPath only if device handle is not found in the map.
Also, disable writing and logging records for hal_summary.csv which was actually intended for developer debugging.
2. Change coarse type data transfer trace default configuration to fine
3. Enable creation of "hal.run_summary" for Vitis Analyzer which contains name of files created with HAL profiling.


maxzhen:
retest this please

PR #3402, opened 2020-05-13
CR-1065044 terminate called after throwing an instance of 'std::invalid_argument'...

AShivangi:
...what(): stoi

Catch the error cleanly. Now if the SC is not able to update, it shows:
```
[root@xsjhemn41 xbmgmt]# ./xbmgmt flash --update
...
Updating SC firmware on card[0000:18:00.0]
Stopping user function...
...................................
ERROR: Failed to update SC firmware, err=-1
ERROR: Please refer to dmesg for more details
WARNING: Failed to update SC firmware on card [0000:18:00.0]
...
```

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbmgmt/xmc.cpp
+++ b/src/runtime_src/core/pcie/tools/xbmgmt/xmc.cpp
@@ -600,7 +600,11 @@ static void tiTxtStreamToBin(std::istream& tiTxtStream,
         switch (line[0]) {
         case '@':
             // Address line
-            currentAddr = std::stoi(line.substr(1), NULL , 16);
+            try {
+                currentAddr = std::stoi(line.substr(1), NULL, 16);
+            } catch (...){
+                return;
+            }
             break;
         case 'q':
         case 'Q':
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/tools/xbmgmt/xmc.cpp
```diff
@@ -600,7 +600,11 @@ static void tiTxtStreamToBin(std::istream& tiTxtStream,
         switch (line[0]) {
         case '@':
             // Address line
-            currentAddr = std::stoi(line.substr(1), NULL , 16);
+            try {
+                currentAddr = std::stoi(line.substr(1), NULL, 16);
+            } catch (...){
+                return;
```
rozumx, line 606: This is great that we are catch the exception and the original issue is addressed.

The next step is to let the user "know" what just happened.  For example, and error message similar to:
`ERROR: The format of the current address '<the given string>' isn't valid.  No action taken.`
Or something like that.

src/runtime_src/core/pcie/tools/xbmgmt/xmc.cpp
```diff
@@ -617,8 +621,15 @@ static void tiTxtStreamToBin(std::istream& tiTxtStream,
...
             std::stringstream ss(line);
             std::string token;
-            while (std::getline(ss, token, ' '))
-                buf.push_back(std::stoi(token, NULL, 16));
+            unsigned char int_token;
+            while (std::getline(ss, token, ' ')) {
+                try {
+                    int_token = std::stoi(token, NULL, 16);
+                } catch (...) {
+                    return;
```
rozumx, line 629: Print out some "helpful" error message.

PR #3403, opened 2020-05-13
new header fields

sarab96:
new header fields


sarab96:
@maxzhen This is used in one XMA2 user space API. It is ok to change in this 2020.1 release. Backward compatibility for XMA2 will be required from next release.

maxzhen:
retest this please

PR #3404, opened 2020-05-14
XRT-486 data-driven msix interrupt for versal

xdavidz:
Please take your time and carefully review this for me. We want to make mb_scheduler data-driven by the name of the resources. 

The special case for vck5000 is that:
1) the resource of IRQ0 interrupt base is also the resource of versal mailbox.
2) versal doen'st have mb ERT (u30 doesn't have mb ERT too).

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/devices.h
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/devices.h
@@ -227,6 +227,7 @@ enum subdev_id {
 	XOCL_SUBDEV_DMA,
 	XOCL_SUBDEV_IORES,
 	XOCL_SUBDEV_FLASH,
+	XOCL_SUBDEV_MAILBOX_VERSAL,
 	XOCL_SUBDEV_MB_SCHEDULER,
 	XOCL_SUBDEV_XVC_PUB,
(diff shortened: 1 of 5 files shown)


xdavidz:
We would prefer to hold off the merge for 2020.1, it should be a fix for 2020.1_web for versal. Please go ahead and review this carefully for me.  Thank you so much!

maxzhen:
> 
> 
> Defer the data-driver changes for versal due to the mb_scheduler is designed to handle its own resources, fetching resources from other subdevs is error-prone.
> 
> Two way to resolve this:
> 
>     1. mb_scheduler to expose APIs for other subdevs to use; or
> 
>     2. convert other subdevs into mb_scheduler's library.
> 
> 
> Any of these 2 efforts require careful thought. Thus, defer the PR.

Per our discussion over phone, for now, let's keep other subdev as is and move data driven code into mb_scheduler. We can clean up the code in new KDS architecture where ERT/CU subdevice has a cleaner interface with KDS.

xdavidz:
New KDS is almost ready. Willing to see if we can utilize the new KDS which will make versal_mailbox as a separate IP(subdev). That's a clear design for the furture.

PR #3405, opened 2020-05-14
CR-1065259 Fixing the way emulation identifies the Versal DC platform

hemantk-xilinx:
Improved the check to identify a versal platform


chvamshi-xilinx:
ok to test

PR #3415, opened 2020-05-17
CID 210272 (#1 of 1): Dereference after null check (FORWARD_NULL)

xdavidz:
ov should not be null now.


src/runtime_src/core/pcie/driver/linux/xocl/subdev/ospi_versal.c
```diff
@@ -287,8 +287,8 @@ static int ospi_versal_remove(struct platform_device *pdev)
 	platform_set_drvdata(pdev, NULL);
 	xocl_drvinst_free(hdl);
 
-done:
 	OV_INFO(ov, "return: %d", ret);
+done:
```
larry9523, line 291: It is OK, but I don't see much value to use a goto in this case. Returning -EINVAL at line 279 will be just fine

PR #3418, opened 2020-05-18
Refactoring the loading of XDP modules in XRT

jvillarre:
Modifying the loading of the hal_api_interface and hal_profiling XDP plugin modules to use the same common XDPLoader class as the XDP plugin modules loaded by the OpenCL level.  This required moving the XDPLoader class functionality into the xrt_coreutil library, which necessitated adding the "dl" library during linking on Linux.


jvillarre:
I changed the names of the files, class, and include guards to better reflect that this is a generic module loader instead of XDP specific.  Also, added comments to describe intended usage and what the constructor parameters should be used for.

jvillarre:
retest this please

src/runtime_src/core/common/module_loader.h
```diff
@@ -0,0 +1,85 @@
...
+   * 
+   * This class is responsible for loading a plugin module from the
+   * appropriate directory under the XILINX_XRT installation.  The loading
+   * happens at object construction time, so the XRT side implementation should
+   * contain a function that instantiates a single static instance of this 
+   * class to handle the loading of a module once in a thread safe manner.
+   *
+   */
+
+  class module_loader
```
stsoe, line 42: Why is this a class?

src/runtime_src/core/common/module_loader.h
```diff
@@ -0,0 +1,85 @@
...
+  {
+  private:
+    // Useful functions
+    const char* emptyOrValue(const char* cstr) ;
+    boost::filesystem::path& dllExt() ;
+    bool isDLL(const boost::filesystem::path& path) ;
+    boost::filesystem::path modulePath(const boost::filesystem::path& root,
+				       const std::string& libname);
+    boost::filesystem::path moduleDir(const boost::filesystem::path& root) ;
```
stsoe, line 51: These looks like they should be static functions in unnamed namespace in module_loader.cpp

src/runtime_src/core/common/module_loader.h
```diff
@@ -0,0 +1,85 @@
...
+ * under the License.
+ */
+
+#ifndef xrtcore_util_module_loader_h_
+#define xrtcore_util_module_loader_h_
+
+// This file contains a loader utility class for plugin modules
+//  that are loaded from either OpenCL or XRT level applications.
+
+#include <boost/filesystem/path.hpp>
```
stsoe, line 23: This include should be in the .cpp file.

src/runtime_src/core/common/module_loader.h
jvillarre: The utility functions were changed into static functions in an unnamed namespace in the .cpp file.

src/runtime_src/core/common/module_loader.h
jvillarre: This include was moved into the cpp file along with the utility functions

src/runtime_src/core/common/module_loader.h
jvillarre: The loading of the plugin module should occur only once and is triggered by the first call into XRT source code from the user's side.  All of the entry points specific to a plugin will call a single function to load that plugin module.  In order to handle multithreaded access from the user and guarantee that the loading only occurs once, the loading functions consist of a single static declaration of an instance of this class.  This instance is only constructed once and is thread safe per C++-11, and simplifies the code by not having to create global locks for each plugin and check flags on every callback to see if loading should occur.

stsoe: ok. makes sense

src/runtime_src/core/common/module_loader.h
stsoe: Thanks

jvillarre:
retest this please

PR #3420, opened 2020-05-18
XRT-290 Enable clock subdev for versal

xdavidz:
This is the initial support for displaying clock for xbutil. So that user can check the frequency of clock now.

We are waiting for platform team to provide clock configuration guide, that's sightly different than existing clock config. For now, xbutil clock -f will just fail. No hardware hung or panic, but it won't work. This will take longer time to fix.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
@@ -842,9 +842,9 @@ void xclmgmt_mailbox_srv(void *arg, void *data, size_t len,
 			break;
 		}
 
-		ret = xocl_has_icap(lro) ?
-		    xocl_icap_ocl_update_clock_freq_topology(lro, clk) :
-		    xocl_clock_update_freq(lro, clk->ocl_target_freq,
+		ret = xocl_icap_ocl_update_clock_freq_topology(lro, clk);
+		if (ret == -ENODEV)
+		    ret = xocl_clock_update_freq(lro, clk->ocl_target_freq,
 		        ARRAY_SIZE(clk->ocl_target_freq), 1);
 
 		(void) xocl_peer_response(lro, req->req, msgid, &ret,
--- a/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
@@ -544,7 +544,7 @@ struct xocl_pci_funcs xclmgmt_pci_ops = {
 	.reset = xclmgmt_reset,
 };
 
-static void xclmgmt_icap_get_data_impl(struct xclmgmt_dev *lro, void *buf)
+static int xclmgmt_icap_get_data_impl(struct xclmgmt_dev *lro, void *buf)
 {
 	struct xcl_pr_region *hwicap = NULL;
 	int err = 0;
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/xocl_fdt.c
```diff
@@ -196,7 +196,7 @@ static void devinfo_cb_plp_gate(void *dev_hdl, void *subdevs, int num)
 {
 	struct xocl_subdev *subdev = subdevs;
 
-	subdev->info.level = XOCL_SUBDEV_LEVEL_BLD;
```
xdavidz, line 199: Based on my test, it should be PRP, please let me know if you think it should be BLD.

src/runtime_src/core/pcie/driver/linux/xocl/xocl_fdt.c
```diff
@@ -196,7 +196,7 @@ static void devinfo_cb_plp_gate(void *dev_hdl, void *subdevs, int num)
...
+	subdev->info.level = XOCL_SUBDEV_LEVEL_PRP;
```
houlz0507, line 199: it should be BLP. plp_gate is implemented in BLP and it isolates BLP.

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
```diff
@@ -822,7 +841,12 @@ void xclmgmt_mailbox_srv(void *arg, void *data, size_t len,
 			mgmt_err(lro, "peer request dropped, wrong size\n");
 			break;
 		}
-		ret = xocl_icap_ocl_update_clock_freq_topology(lro, clk);
+
+		ret = xocl_has_icap(lro) ?
```
houlz0507, line 845: It is better to check the return value (-ENODEV) instead of defining xocl_has_icap().
Why do we still keep xocl_icap_ocl_update_clock_freq_topology in icap?  Should it be better to call xocl_clock_update _freq in both cases?

src/runtime_src/core/pcie/driver/linux/xocl/xocl_fdt.c
xdavidz: Thanks, adjust it back. Actually the clock.c should have use BLD instead of PRP.

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
xdavidz: Adjust the code to check -ENODEV.
The xocl_icap_ocl_update_clock_freq_topology is still need here due to it checks xclbin min/max freq. After we can split the xclbin and icap, we can move this checking to better place.

xdavidz:
retest this please

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
```diff
@@ -570,6 +570,25 @@ static void xclmgmt_icap_get_data(struct xclmgmt_dev *lro, void *buf)
...
+	hwicap->freq_1 = xocl_clock_get_data(lro, CLOCK_FREQ_1);
+	hwicap->freq_2 = xocl_clock_get_data(lro, CLOCK_FREQ_2);
+	hwicap->freq_cntr_0 = xocl_clock_get_data(lro, FREQ_COUNTER_0);
+	hwicap->freq_cntr_1 = xocl_clock_get_data(lro, FREQ_COUNTER_1);
+	hwicap->freq_cntr_2 = xocl_clock_get_data(lro, FREQ_COUNTER_2);
+}
+
+static void xclmgmt_icap_get_data(struct xclmgmt_dev *lro, void *buf)
+{
+	return xocl_has_icap(lro) ? xclmgmt_icap_get_data_impl(lro, buf) :
```
houlz0507, line 588: Please use -ENODEV and remove xocl_has_icap().

xdavidz: The current API will return 0 instead of -ENODEV, should I change it?
XOCL_GET_XCLBIN_ID

PR #3422, opened 2020-05-19
Put host_mem error message in order

chienwei-lan:
1. Driver return -EBUSY if user try to disable/re-enable Host_mem feature with different size

2. Driver Return -ENODEV if user try to enable HOST_MEM w/o address translator

3. Clean all the bitstream related metadata(mem topo, uuid, etc) if xocl_read_axlf_ioctl fails


xilinxgitops:
retest this please

PR #3428, opened 2020-05-19
Hooks for Cardano

stsoe:


src/runtime_src/core/common/exec/kernel.cpp
```diff
@@ -697,6 +743,28 @@ class run_update_type
...
+
+  void
+  update_arg_at_index(size_t index, const void* value, size_t bytes)
+  {
+    auto& arg = kernel->args.at(index);
+    if (arg.index() == argument::no_index)
+      throw std::runtime_error("Bad argument index '" + std::to_string(index) + "'");
+    bytes = std::max(sizeof(uint32_t), bytes);
+    auto uval = reinterpret_cast<const uint32_t*>(value);
+    update_arg_value(arg, {uval, uval + bytes / sizeof(uint32_t)});
```
keryell, line 766: All these rounding codes could benefit from a function.

stsoe: Elaborate please.  I don't see any rounding ?

PR #3435, opened 2020-05-20
CR-1065849 opmode ERT_EXEC_WRITE does not set kernel arguments correctly

mamin506:
Support EXEC_WRITE in ert mode

PR #3436, opened 2020-05-20
CR-1065849 opmode ERT_EXEC_WRITE does not set kernel arguments correctly

mamin506:
This for 2020.1 branch. #3435


maxzhen:
this is not a regression. we'll fix it in 2020.1_web for versal. closing now.

PR #3441, opened 2020-05-21
Enable validate_cus for __arm__

stsoe:
The __arm__ check is no longer needed after mem_data entries mapping to to same
base address are merged and treated as one per #3193.


abalasa:
retest this please

PR #3444, opened 2020-05-21
CR-1065952: NV12 get num of frame planes

sarab96:
NV12 get num of frame planes
CR-1065952


maxzhen:
This fix is approved by Sumit. Merging now.

PR #3447, opened 2020-05-22
XRT-459 Add context support in KDS

mamin506:
1. Support xclOpenContext() and xclCloseContext at unify KDS.
- The first client open context would lock the bitstream. Others who try to change bitstream will fail.
- Other clients could open context with the same bitstream (bump up the ref count).
- Only when the ref count return to 0, the bitstream is unlocked.
2. Unify CU controller code for xocl and zocl. The logic is almost the same.

After this, zocl scheduler would fully support multi processes. (In before, exclusive reserved CU was not support).

### DIFF
--- a/src/runtime_src/core/common/drv/kds_core.c
+++ b/src/runtime_src/core/common/drv/kds_core.c
@@ -278,9 +278,9 @@ u32 kds_live_clients(struct kds_sched *kds, pid_t **plist)
 			i++;
 		}
 	}
-	mutex_unlock(&kds->lock);
 
 out:
+	mutex_unlock(&kds->lock);
 	return count;
 }
 
--- a/src/runtime_src/core/common/drv/kds_core.c
+++ b/src/runtime_src/core/common/drv/kds_core.c
@@ -260,7 +260,7 @@ u32 kds_live_clients(struct kds_sched *kds, pid_t **plist)
 	/* Find out number of active client */
 	list_for_each(ptr, &kds->clients) {
 		client = list_entry(ptr, struct kds_client, link);
-		if (CLIENT_NUM_CU(client) > 0)
+		if (client->num_ctx > 0)
 			count++;
 	}
 	if (count == 0 || plist == NULL)
@@ -273,7 +273,7 @@ u32 kds_live_clients(struct kds_sched *kds, pid_t **plist)
 
 	list_for_each(ptr, &kds->clients) {
 		client = list_entry(ptr, struct kds_client, link);
-		if (CLIENT_NUM_CU(client) > 0) {
+		if (client->num_ctx > 0) {
 			pl[i] = pid_nr(client->pid);
 			i++;
 		}
@@ -299,6 +299,7 @@ void notify_execbuf(struct kds_command *xcmd, int status)
 	wake_up_interruptible(&client->waitq);
 }
 
+/* User space execbuf command related functions below */
 void cfg_ecmd2xcmd(struct ert_configure_cmd *ecmd,
 		   struct kds_command *xcmd)
 {
@@ -339,3 +340,10 @@ void start_krnl_ecmd2xcmd(struct ert_start_kernel_cmd *ecmd,
 	memcpy(xcmd->info, &ecmd->data[4], xcmd->isize);
 }
 
+inline int cu_mask_to_cu_idx(struct kds_command *xcmd)
+{
+	/* TODO: balance the CU usage if multiple bits are set */
+
+	/* assume there is alwasy one CU */
+	return 0;
+}
--- a/src/runtime_src/core/common/drv/kds_core.c
+++ b/src/runtime_src/core/common/drv/kds_core.c
@@ -7,11 +7,25 @@
  * Authors: min.ma@xilinx.com
  */
 
-#include <linux/device.h>
 #include <linux/slab.h>
 #include <linux/sched.h>
 #include <linux/vmalloc.h>
+#include <linux/device.h>
 #include "kds_core.h"
+
+int kds_init_sched(struct kds_sched *kds)
+{
+	INIT_LIST_HEAD(&kds->clients);
+	mutex_init(&kds->lock);
+	kds->num_client = 0;
+
+	return 0;
+}
+
+void kds_fini_sched(struct kds_sched *kds)
+{
+	mutex_destroy(&kds->lock);
+}
 
 struct kds_command *kds_alloc_command(struct kds_client *client, u32 size)
 {
@@ -65,7 +79,7 @@ void kds_free_command(struct kds_command *xcmd)
 int kds_submit_cu(struct kds_command *xcmd)
 {
 	struct kds_client *client = xcmd->client;
-	struct kds_controller *ctrl = client->ctrl[KDS_CU];
+	struct kds_ctrl *ctrl = client->ctrl[KDS_CU];
 
 	if (ctrl) {
 		/* NOTE: If still has errors, the controller
@@ -108,10 +122,29 @@ int kds_add_command(struct kds_command *xcmd)
 	return err;
 }
 
-int kds_init_client(struct kds_client *client)
+int kds_init_client(struct kds_sched *kds, struct kds_client *client)
 {
+	struct kds_ctx_info info;
+	struct kds_ctrl *ctrl;
+
+	client->pid = get_pid(task_pid(current));
+	client->ctrl = kds->ctrl;
+	mutex_init(&client->lock);
+
+	/* Initial controller context private data */
+	ctrl = client->ctrl[KDS_CU];
+	info.flags = CU_CTX_OP_INIT;
+	if (ctrl)
+		ctrl->control_ctx(ctrl, client, &info);
+
 	init_waitqueue_head(&client->waitq);
 	atomic_set(&client->event, 0);
+
+	mutex_lock(&kds->lock);
+	list_add_tail(&client->link, &kds->clients);
+	kds->num_client++;
+	mutex_unlock(&kds->lock);
+
 #if PRE_ALLOC
 	client->max_xcmd = 0x8000;
 	client->xcmd_idx = 0;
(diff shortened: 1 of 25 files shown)


mamin506:
This is the last big change for new KDS. The framework and concepts are clear now.
@larry9523, @xdavidz, @chvamshi-xilinx Please help to review this.

xdavidz:
@mamin506 this is 1000+ lines of code. Do you have any confluence page or doc talk about the over all design. Why it is clear than previous version?

mamin506:
> @mamin506 this is 1000+ lines of code. Do you have any confluence page or doc talk about the over all design. Why it is clear than previous version?

Oh.. I didn't realize that this change has 1000+ lines. I have a confluence page for unify KDS. Let me add how context work to that page and shared it with you.

xdavidz:
> > @mamin506 this is 1000+ lines of code. Do you have any confluence page or doc talk about the over all design. Why it is clear than previous version?
> 
> Oh.. I didn't realize that this change has 1000+ lines. I have a confluence page for unify KDS. Let me add how context work to that page and shared it with you.

I am looking for the new way to handle ERT. Basically, we are having mailbox_versal subdev, which is actually an ERT. The best way to do this is that the mailbox_versal subdev should register itself into the new KDS, then the KDS will also pass in a callback (used to be named as notify). Therefore, the subdev will handle all requests by itself (KDS does not interact with the ERT), after the CU is done, the ERT subdev(in this case the mailbox_versal) will call the given callback (still named notify?).

A simple illustation:

Client-> scheduler ->"submit"-> ERT1 
Client<- scheduler <-"notify"<- ERT1 

Client-> scheduler ->"submit"-> ERT2 
Client<- scheduler <-"notify"<- ERT2

As you can see the scheduler should have no knowledge about inside of ERT.

As of today, if this won't happen soon in 2020.1_web, we will have to move all ERT code into mb_scheduler.c. Otherwise, there are too many hacks in the code.

xdavidz:
I basically have one single comment, why we separate xocl and zocl code, like zocl_kds.c and xocl_kds.c. What I think that we could put some code in common area (even compiled under different env).

mamin506:
> > > @mamin506 this is 1000+ lines of code. Do you have any confluence page or doc talk about the over all design. Why it is clear than previous version?
> > 
> > 
> > Oh.. I didn't realize that this change has 1000+ lines. I have a confluence page for unify KDS. Let me add how context work to that page and shared it with you.
> 
> Seems this is only for the context switch, but ERT handling is not there yet?
> 
> I am looking for the new way to handle ERT. Basically, we are having mailbox_versal subdev, which is actually an ERT. The best way to do this is that the mailbox_versal subdev should register itself into the new scheduler, then the scheduler will also pass in a callback (used to be named as notify). Therefore, the subdev will handle all requests by itself (scheduler does not interact with the ERT), after the CU is done, the ERT subdev(in this case the mailbox_versal) will call the given callback (still named notify?).
> 
> A simple illustation:
> 
> Client-> scheduler ->"submit"-> ERT1
> Client<- scheduler <-"notify"<- ERT1
> 
> Client-> scheduler ->"submit"-> ERT2
> Client<- scheduler <-"notify"<- ERT2
> 
> As you can see the scheduler should have no knowledge about inside of ERT.
> 
> As of today, if this won't happen soon in 2020.1_web, we will have to move all ERT code into mb_scheduler.c. Otherwise, there are too many hacks in the code.

Yes. This is for context switch. ERT is the next milestone. Let's do that step by step. Otherwise the change would be more then 2000+ lines.

This is code is disabled in 2020.1. You could use module param kds_mode=1 to enable it and try it out.

mamin506:
> I basically have one single comment, why we separate xocl and zocl code, like zocl_kds.c and xocl_kds.c. What I think that we could put some code in common area (even compiled under different env).

Because xocl use xdev and zocl use zdev. They are different. Right now, this two files are pretty simple and they do have shared code in common/drv/kds_core.c.

src/runtime_src/core/common/drv/kds_core.c
```diff
@@ -108,10 +122,29 @@ int kds_add_command(struct kds_command *xcmd)
...
+
+	/* Initial controller context private data */
+	ctrl = client->ctrl[KDS_CU];
+	info.flags = CU_CTX_OP_INIT;
+	if (ctrl)
+		ctrl->control_ctx(ctrl, client, &info);
+
 	init_waitqueue_head(&client->waitq);
 	atomic_set(&client->event, 0);
+
```
larry9523, line 142: seems we don't need atomic_set but just (client->event = 0) should be good, right?

src/runtime_src/core/common/drv/kds_core.c
```diff
@@ -130,13 +163,125 @@ int kds_init_client(struct kds_client *client)
...
+	mutex_lock(&kds->lock);
+	/* Find out number of active client */
+	list_for_each(ptr, &kds->clients) {
+		client = list_entry(ptr, struct kds_client, link);
+		if (client->num_ctx > 0)
+			count++;
+	}
+	if (count == 0 || plist == NULL)
+		goto out;
```
larry9523, line 267: do we need to release lock before bailing out?

src/runtime_src/core/common/drv/kds_core.c
```diff
@@ -130,13 +163,125 @@ int kds_init_client(struct kds_client *client)
...
+
+	/* Collect list of PIDs of active client */
+	pl = (pid_t *)vmalloc(sizeof(pid_t) * count);
+	if (pl == NULL)
+		goto out;
```
larry9523, line 272: same as 267

src/runtime_src/core/common/drv/kds_cu_ctrl.c
```diff
@@ -0,0 +1,294 @@
...
+	int i;
+
+	/* Do not use this search in critical path */
+	for (i = 0; i < kcuc->num_cus; ++i) {
+		if (kcuc->xcus[i]->info.addr == addr)
+			break;
+	}
+
+	return i;
+}
```
larry9523, line 24: It is a little bit confusing here. What if we could not find the cu (idx?) by given address? how about
static int get_cu_idx_by_addr (struct kds_cu_ctrl *kcuc, u32 addr, int *cu_idx)
?

src/runtime_src/core/common/drv/kds_cu_ctrl.c
```diff
@@ -0,0 +1,294 @@
...
+	case CU_CTX_OP_ADD:
+		return add_ctx(kcuc, client, info);
+	case CU_CTX_OP_DEL:
+		return del_ctx(kcuc, client, info);
+	}
+
+	return 0;
+}
+
+int add_cu(struct kds_cu_ctrl *kcuc, struct xrt_cu *xcu)
```
larry9523, line 229: I would suggest a clearer interface for the functions like this (remove_cu as well).
negative value means unsuccessful
0 means successful
if we need to return the index where the cu is added (or removed), we can add an extra pointer argument.

src/runtime_src/core/edge/drm/zocl/zocl_kds.c
```diff
@@ -24,6 +25,144 @@ module_param(kds_echo, int, (S_IRUGO|S_IWUSR));
...
+	ret = kds_add_context(client, &info);
+
+out:
+	if (!client->num_ctx) {
+		vfree(client->xclbin_id);
+		client->xclbin_id = NULL;
+		(void) zocl_unlock_bitstream(zdev, id);
+	}
+	vfree(id);
+	mutex_unlock(&client->lock);
```
larry9523, line 86: Nit, you may want to put mutex_unlock before vfree() and make a sequence pair 
alloc
  mutex_lock
  mutex_unlock
free

src/runtime_src/core/edge/drm/zocl/zocl_kds.c
```diff
@@ -24,6 +25,144 @@ module_param(kds_echo, int, (S_IRUGO|S_IWUSR));
...
+	uuid_t *id;
+	int ret;
+
+	id = vmalloc(sizeof(uuid_t));
+	if (!id)
+		return -ENOMEM;
+
+	ret = copy_from_user(id, uuid_ptr, sizeof(uuid_t));
+	if (ret)
+		goto out;
```
larry9523, line 58: if copy_from_user fails, do we have issue here to unlock a mutex which is not locked before?

src/runtime_src/core/edge/drm/zocl/zocl_kds.c
```diff
@@ -24,6 +25,144 @@ module_param(kds_echo, int, (S_IRUGO|S_IWUSR));
...
+
+	if (!client->num_ctx) {
+		vfree(client->xclbin_id);
+		client->xclbin_id = NULL;
+		(void) zocl_unlock_bitstream(zdev, id);
+	}
+
+out:
+	vfree(id);
+	mutex_unlock(&client->lock);
```
larry9523, line 141: same nit as above

src/runtime_src/core/pcie/driver/linux/xocl/subdev/cu_ctrl.c
```diff
@@ -18,93 +18,62 @@
...
+	/* This is call echo type 1.
+	 * The echo type 2 could be enabled by module param 'kds_echo=1'
 	 */
 	//xcmd->cb.notify_host(xcmd, KDS_COMPLETED);
 	//xcmd->cb.free(xcmd);
 	//return;
 
-	inst_idx = xcuc->xcus[cu_idx]->info.inst_idx;
-	(void) xocl_cu_submit_xcmd(xdev, inst_idx, xcmd);
+	/* In here we needs to know the CU subdevice instance id,
```
larry9523, line 67: need :)

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_kds.c
```diff
@@ -20,11 +20,119 @@ module_param(kds_echo, int, (S_IRUGO|S_IWUSR));
...
+	switch(args->op) {
+	case XOCL_CTX_OP_ALLOC_CTX:
+		ret = xocl_add_context(xdev, client, args);
+		break;
+	case XOCL_CTX_OP_FREE_CTX:
+		ret = xocl_del_context(xdev, client, args);
+		break;
+	default:
+		ret = -EINVAL;
+	}
```
larry9523, line 133: It is legal. But it is just my suggestion we add a break for every case including default. :) The default case could be not the last case, though rare.

src/runtime_src/core/common/drv/kds_core.c
mamin506: Yes. Will fix.

src/runtime_src/core/common/drv/kds_core.c
mamin506: fixed

src/runtime_src/core/common/drv/kds_core.c
mamin506: fixed

src/runtime_src/core/common/drv/kds_cu_ctrl.c
mamin506: If CU is not found, it would return num_cus, which is out of range.
If this function would return any other error type, we could change this. Let me add a command for this function.

BTW, at the end of the day, once all of the modules become stable, we should commend all functions.

src/runtime_src/core/common/drv/kds_cu_ctrl.c
mamin506: Fixed

src/runtime_src/core/edge/drm/zocl/zocl_kds.c
mamin506: fixed

src/runtime_src/core/edge/drm/zocl/zocl_kds.c
mamin506: fixed

src/runtime_src/core/edge/drm/zocl/zocl_kds.c
mamin506: fixed

src/runtime_src/core/pcie/driver/linux/xocl/subdev/cu_ctrl.c
mamin506: fixed

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_kds.c
mamin506: fixed

mamin506:
> Overall looks good. Just in a few rare cases, there could be some leak. And most of the review comments are just suggestions.

Thanks for the comments. I believe there should be some hiding bugs. We will find they in the future.
The exception handling(ex. exit without close context, ctrl-c) is not completed. I will take care of leak issue at that time.

src/runtime_src/core/common/drv/kds_core.c
mamin506: We can't directly assign 0 to atomic_t.

mamin506: As discussed offline, we observed polling cause performance issue. Let's review this part later.

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_ioctl.c
```diff
@@ -349,6 +329,39 @@ xocl_read_axlf_helper(struct xocl_drm *drm_p, struct drm_xocl_axlf *axlf_ptr)
 	if (xclbin_downloaded(xdev, &bin_obj.m_header.uuid))
 		goto done;
 
+	/*
```
maxzhen, line 332: i don't see anything wrong to separate xclbin download and xclbin lock/unlock operation. but i agree that the coupling here is not ideal.

mamin506: I agree. Separate download and lock is okay. I will change the comment.
If anyone try to load another xclbin before KDS unlock prev xclbin, ICAP would failed here. If it does need to check KDS status in here, we could improve this in the future.

mamin506:
> the part of code that handles cu cmds that get stuck in hw seems missing. we're adding the code in follow up pull request, right?

That part is about the exception handling. I would create another PR for that.
It would needs to add more operations on CU controller and CU subdevice.

src/runtime_src/core/common/drv/kds_cu_ctrl.c
```diff
@@ -0,0 +1,294 @@
...
+	if (test_and_set_bit(cu_idx, cu_priv->cu_bitmap)) {
+		kds_err(client, "CU(%d) has been added", cu_idx);
+		return -EINVAL;
+	}
+
+	/* Remove OPs from flag */
+	info->flags &= ~CU_CTX_OP_MASK;
+	shared = (info->flags != CU_CTX_EXCLUSIVE);
+
+	mutex_lock(&kcuc->lock);
```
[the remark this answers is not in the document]
mamin506: cu_priv->cu_bitmap is per client. The xocl_add_context() or zocl_add_context() would hold client->lock. From the same client, thread A and B could not go to this function at the same time.
If thread A and B are from different clients, if only one could take kcuc->lock and update kcuc->cu_refs.

The add_ctx() and del_ctx() are called by control_ctx(). I could check if clinet->lock is locked at control_ctx().

I would comment these in add_ctx() and del_ctx().

src/runtime_src/core/common/drv/kds_cu_ctrl.c
```diff
@@ -0,0 +1,294 @@
...
+
+	cu_priv = client->ctrl_priv[KDS_CU];
+	if (test_and_set_bit(cu_idx, cu_priv->cu_bitmap)) {
+		kds_err(client, "CU(%d) has been added", cu_idx);
+		return -EINVAL;
+	}
+
+	/* Remove OPs from flag */
+	info->flags &= ~CU_CTX_OP_MASK;
+	shared = (info->flags != CU_CTX_EXCLUSIVE);
```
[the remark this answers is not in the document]
mamin506: fixed

src/runtime_src/core/common/drv/kds_cu_ctrl.c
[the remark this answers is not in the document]
maxzhen: > 
> 
> cu_priv->cu_bitmap is per client. The xocl_add_context() or zocl_add_context() would hold client->lock. From the same client, thread A and B could not go to this function at the same time.
> If thread A and B are from different clients, if only one could take kcuc->lock and update kcuc->cu_refs.
> 
> The add_ctx() and del_ctx() are called by control_ctx(). I could check if clinet->lock is locked at control_ctx().
> 
> I would comment these in add_ctx() and del_ctx().

OK, I see. That is fine, then.

src/runtime_src/core/common/drv/kds_cu_ctrl.c
```diff
@@ -52,9 +58,10 @@ add_ctx(struct kds_cu_ctrl *kcuc, struct kds_client *client,
 	}
 
 	/* Remove OPs from flag */
-	info->flags &= ~CU_CTX_OP_MASK;
-	shared = (info->flags != CU_CTX_EXCLUSIVE);
+	prop = info->flags & CU_CTX_OP_MASK;
+	shared = (prop != CU_CTX_EXCLUSIVE);
```
maxzhen, line 62: I'm confused, prop contains only CU_CTX_OP_xxx now, but CU_CTX_EXCLUSIVE is not part of it, why do you compare them?

mamin506: Oh.. It should be CU_CTX_PROP_MASK.

PR #3448, opened 2020-05-22
new ini switch added

uday610:
New ini switch documentation

### DIFF
--- a/src/runtime_src/doc/toc/xrt_ini.rst
+++ b/src/runtime_src/doc/toc/xrt_ini.rst
@@ -37,41 +37,52 @@ The following table lists all supported groups, keys, valid key values, and shor
 Runtime Group
 =============
 
-+-----------------+------------------------------+-------------------------------------------+
-|  Key            |  Valid Values                |             Description                   |
-+=================+==============================+===========================================+
-| api_checks      |  [true|false]                |Enable or disable OpenCL API checks:       |
-|                 |                              |                                           |
-|                 |                              |     - true: enable                        |
-|                 |                              |     - false: disable                      |
-|                 |                              |                                           |
-|                 |                              |Default: true                              |
-+-----------------+------------------------------+-------------------------------------------+
-| runtime_log     |[null|console|syslog|filename]|Specify where the runtime logs are printed:|
-|                 |                              |                                           |
-|                 |                              |     - null: Do not print any logs.        |
-|                 |                              |     - console: Print logs to stdout       |
-|                 |                              |     - syslog: Print logs to Linux syslog  |
-|                 |                              |     - filename: Print logs to the         |
-|                 |                              |       specified file.                     |
-|                 |                              |       Example, runtime_log=my_run.log     |
-|                 |                              |                                           |
-|                 |                              |Default: console                           |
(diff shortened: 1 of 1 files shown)


maxzhen:
ok to test

src/runtime_src/doc/toc/xrt_ini.rst
```diff
@@ -37,41 +37,52 @@ The following table lists all supported groups, keys, valid key values, and shor
...
+| polling_throttle    | [N]                          |Specify the time interval in microseconds  |
+|                     |                              |that the runtime library polls the device. |
+|                     |                              |                                           |
+|                     |                              |Default: 0                                 |
++---------------------+------------------------------+-------------------------------------------+
+| verbosity           | [0|1|2|3]                    |verbosity level of log messages. Higher    |
+|                     |                              |number implies more verbosity              |
+|                     |                              |                                           |
+|                     |                              |Default: 0                                 |
++---------------------+------------------------------+-------------------------------------------+
```
stsoe, line 74: @uday610 Let's fix this one.   verbosity has many more values and default is 4.

src/runtime_src/doc/toc/xrt_ini.rst
```diff
@@ -37,41 +37,52 @@ The following table lists all supported groups, keys, valid key values, and shor
...
+|                     |                              |Default: console                           |
++---------------------+------------------------------+-------------------------------------------+
+| cpu_affinity        | [{N,N,...}]                  |Pin all runtime threads to specified CPUs. |
+|                     |                              |                                           |
+|                     |                              |Example: cpu_affinity = {4,5,6}            |
++---------------------+------------------------------+-------------------------------------------+
+| polling_throttle    | [N]                          |Specify the time interval in microseconds  |
+|                     |                              |that the runtime library polls the device. |
+|                     |                              |                                           |
+|                     |                              |Default: 0                                 |
```
stsoe, line 68: This is only applicable for userspace software scheduler which is for internal use only.  Lets remove from documentation.

src/runtime_src/doc/toc/xrt_ini.rst
uday610: @stsoe  , done. But I think the way the internal code is written and this switch name is confusing. 
If I see https://github.com/Xilinx/XRT/blob/master/src/runtime_src/core/include/xrt.h#L219 , xclVerbosityLevel is still 4, but this switch does not correspond to that, though from the name it seems like. 
This switch corresponds to xrtLogMsgLevel (https://github.com/Xilinx/XRT/blob/master/src/runtime_src/core/include/xrt.h#L203 ). So the name of the switch better fits "loglevel" or something like this.

src/runtime_src/doc/toc/xrt_ini.rst
uday610: done. Thanks

src/runtime_src/doc/toc/xrt_ini.rst
stsoe: yes that's confusing.  i will try look at that and clean it up.  thanks!

PR #3450, opened 2020-05-22
XRT-472 Porting XRT code to RHEL 8.1

rbramand-xilinx:
> Updated xrtdeps.sh to support rhel 8.1

### DIFF
--- a/src/runtime_src/tools/scripts/xrtdeps.sh
+++ b/src/runtime_src/tools/scripts/xrtdeps.sh
@@ -89,7 +89,6 @@ rh_package_list()
      unzip \
      zlib-static \
      libcurl-devel \
-     openssl-devel \
     )
 
     # Centos8
@@ -99,8 +98,16 @@ rh_package_list()
          systemd-devel \
          python3 \
          python3-pip \
-         systemd-devel \
         )
+
+	if [ $FLAVOR == "rhel" ]; then
+  
+            RH_LIST+=(\
+             kernel-devel-$(uname -r) \
+             kernel-headers-$(uname -r) \
+            )
+  
+        fi
 
     else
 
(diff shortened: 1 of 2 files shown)


src/runtime_src/tools/scripts/xrtdeps.sh
```diff
@@ -276,15 +289,18 @@ prep_rhel()
     echo "Enabling EPEL repository..."
     rpm -q --quiet epel-release
     if [ $? != 0 ]; then
-	yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+	if [ $MAJOR -lt "8" ]; then
+	    yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+	else
+	    yum install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
+	fi
```
stsoe, line 296: Let's move this in prep_rhel7 and prep_rhel8.  Also move cmake3 install after prep_rhel8 and rhel7 ?

src/runtime_src/tools/scripts/xrtdeps.sh
```diff
@@ -99,9 +98,17 @@ rh_package_list()
          systemd-devel \
          python3 \
          python3-pip \
-         systemd-devel \
```
stsoe, line 102: Did you try building XRT from a clean centos8?  I am pretty sure this list is needed as it was.

chvamshi-xilinx: @stsoe Looks like these are duplicates.

stsoe: Duh :-)

src/runtime_src/tools/scripts/xrtdeps.sh
rbramand-xilinx: @stsoe made changes as suugested, please review

PR #3459, opened 2020-05-27
CR-1054524 XRT - Remove the five second delay workaround for CMC/SC b…

xdavidz:
I am still testing it. Will update test report when testing is done.

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbmgmt/xmc.cpp
+++ b/src/runtime_src/core/pcie/tools/xbmgmt/xmc.cpp
@@ -90,6 +90,9 @@ XMC_Flasher::XMC_Flasher(std::shared_ptr<pcidev::pci_device> dev)
             std::cout << "Failed to open XMC device on card" << std::endl;
     }
 
+    initCMCStatusMap();
+    initSCStatusMap();
+
 nosup:
     return;
 }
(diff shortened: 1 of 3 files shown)


xdavidz:
@AShivangi @houlz0507 Got clear definition for checking stable SC via CMC. Could you please help me review if this change is appropriate?

src/runtime_src/core/pcie/tools/xbmgmt/xmc.cpp
```diff
@@ -500,15 +503,48 @@ int XMC_Flasher::writeReg(unsigned RegOffset, unsigned value) {
...
+{
+    auto entry = map.find(status);
+
+    return entry != map.end() ?
+        "0x" + std::to_string(entry->first) + "(" + entry->second + ")":
+	"0x" + std::to_string(entry->first);
+}
+
+/* XMC means CMC */
+void XMC_Flasher::initCMCStatusMap()
```
AShivangi, line 516: Maybe make this an enum?

xdavidz: We need to print the string like "NOT READY" or "READY". You mean we convert the 0,1,2,4 to enum?

AShivangi: I see. This looks good.

keryell:
There should not be a mix of spaces and tabs in the code.

xdavidz:
> There should not be a mix of spaces and tabs in the code.

C in kernel uses 8 character as one single tab, CPP in userland uses 4 spaces.
Please let me know which line(s) is(are) broken, I will fix them.

PR #3466, opened 2020-05-27
CR-1065562 ASTeR TC 18.01 - xbutil dmatest returns Read bandwidth val…

pranjalv-xilinx:
…ues out of expected range.

on u50 HBM bank size is 256MB causing dmatest to only create 1 BO. DMA bandwidth test results will not be accurate because we will be measuring transfer for 1 BO over 2 xdma hannels.  16MB seems to be fair IO size for measure BW on different platforms.


pranjalv-xilinx:
retest this please

PR #3475, opened 2020-05-30
Move deprecated APIs to sub-folder

stsoe:
And remove inline documentation.  These APIs are deprecated no
need to document other than replacement.


stsoe:
retest this please

PR #3476, opened 2020-06-01
Load AIE only PDI

larry9523:


src/runtime_src/core/edge/user/aie/graph.cpp
```diff
@@ -538,6 +538,14 @@ xrtSyncBOAIE(xrtGraphHandle ghdl, unsigned int bo, const char *dmaID, enum xclBO
   graph->sync_bo(bo, dmaID, dir, size, offset);
 }
 
+void
+xrtResetAieArray(xclDeviceHandle handle)
+{
+  auto device = xrt_core::get_userpf_device(handle);
+  XAieLib_NpiAieArrayReset(XAIE_RESETENABLE);
+  XAieLib_NpiAieArrayReset(XAIE_RESETDISABLE);
```
stsoe, line 546: Doesn't look the device is used here?  Can you remove line 544?

larry9523: My intention here is to do a handle check. The caller should open the device before operate on it. Do you think it is necessary? Or do we have a better way to do it?

stsoe: Ok. intention is fine, makes sense.  Maybe comment to that effect and maybe don't save the return value.

larry9523: Done. Thanks!

larry9523:
This PR depends on https://github.com/Xilinx/XRT/pull/3479
I will remove the "do not merge" label after the above PR is merged

PR #3479, opened 2020-06-01
AIE Support: Updated the xclbin header to support an action mask

rozumx:
Updated the xclbin header to support an action mask value.  This value represents a collection of bits used to indicate an action that the drivers should perform.  In this case, LOAD_AIE pid image.
Work Done
---------
+ Reduced the size of the m_mode value from an unit32_t to uintt16_t.
+ Added a new variable "m_actionMask" after m_mode
+ Verified via a hex editor that the m_mode value placement in the xclbin didn't regress
+ Added a new SYS parameter to xclbinutil (e.g., action_mask) that supports setting the LOAD_AIE bit
+ Update the --info report to report any "set" bit values in the action_mask variable.

### DIFF
--- a/src/runtime_src/core/include/xclbin.h
+++ b/src/runtime_src/core/include/xclbin.h
@@ -1,5 +1,5 @@
 /**
- *  Copyright (C) 2015-2018, Xilinx Inc
+ *  Copyright (C) 2015-2020, Xilinx Inc
  *
  *  This file is dual licensed.  It may be redistributed and/or modified
  *  under the terms of the Apache 2.0 License OR version 2 of the GNU
@@ -192,6 +192,10 @@ extern "C" {
         IP_MEM_HBM
     };
 
+    enum ACTION_MASK {
+      AM_LOAD_AIE = 0x1                     /* Indicates to the driver to load the AIE PID section */
+    };
+
     struct axlf_section_header {
         uint32_t m_sectionKind;             /* Section type */
         char m_sectionName[16];             /* Examples: "stage2", "clear1", "clear2", "ocl1", "ocl2, "ublaze", "sched" */
@@ -207,7 +211,8 @@ extern "C" {
         uint16_t m_versionPatch;            /* Patch Version */
         uint8_t m_versionMajor;             /* Major Version - Version: 2.1.0*/
         uint8_t m_versionMinor;             /* Minor Version */
-        uint32_t m_mode;                    /* XCLBIN_MODE */
+        uint16_t m_mode;                    /* XCLBIN_MODE */
+        uint16_t m_actionMask;              /* Bit Mask */
 	union {
 	    struct {
 		uint64_t m_platformId;      /* 64 bit platform ID: vendor-device-subvendor-subdev */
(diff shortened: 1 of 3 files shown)


abalasa:
retest this please

rozumx:
**_Test Failure_**
18:25:19 SC FIRMWARE MISMATCH, ATTENTION
18:25:19 SC firmware running on board: 4.0.0. Expected SC firmware from installed Shell: 4.0.7
18:25:19 Please use "xbmgmt flash --scan" to check installed Shell.
18:25:19 WARN: == SC firmware version check PASSED with warning
18:25:19 INFO: == Starting verify kernel test: 
18:25:19 
18:25:19 ERROR: One or more tests failed to run.
18:25:19 ERROR: XRT test FAILED.
18:25:19 INFO: /usr/bin/ssh -t -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -q xcosda83 "cat /opt/xilinx/dmesg_log_name" 

Being investigated by the XRT_pipeline team

rozumx:
retest this please

maxzhen:
retest this please

src/runtime_src/core/include/xclbin.h
```diff
@@ -207,7 +211,8 @@ extern "C" {
         uint16_t m_versionPatch;            /* Patch Version */
         uint8_t m_versionMajor;             /* Major Version - Version: 2.1.0*/
         uint8_t m_versionMinor;             /* Minor Version */
-        uint32_t m_mode;                    /* XCLBIN_MODE */
+        uint16_t m_mode;                    /* XCLBIN_MODE */
+        uint16_t m_actionMask;              /* Bit Mask */
```
sonals, line 215: Confirming that this should not break any xclbin on little endian architecture and all our host CPUs are little endian so we should be good.

src/runtime_src/tools/xclbinutil/FormattedOutput.cxx
```diff
@@ -331,6 +331,18 @@ reportXclbinInfo( std::ostream & _ostream,
...
+    // Are there any masks set
+    if (_xclBinHeader.m_header.m_actionMask) {
+      _ostream << XUtil::format("   %-23s ", "Action Mask(s):");
+      if (_xclBinHeader.m_header.m_actionMask & AM_LOAD_AIE) {
+        _ostream << "LOAD_AIE ";
+      }
+      _ostream << std::endl;
+    }
+  }
+
```
sonals, line 345: What does this do for older xclbins? Do we always get zero for higher bits so we are good?

src/runtime_src/core/include/xclbin.h
rozumx: Yes, this piece of code concerned me greatly as well.   So I:
1) Created an xclbin archive image before and after the changes and verify correctness in a binary editor.
2) Additional testing. 
i) Modified the code and set the uint32_t m_mode to 0x1122, validated that the bytes in the xclbin header were: 0x22 0x11, 0x00, 0x00.
ii) Read in modified xclbin and validated that theuint16_t m_mode value was 0x1122.
3) Had multiple code reviews for this piece of code.

src/runtime_src/tools/xclbinutil/FormattedOutput.cxx
rozumx: Yes, previous version of xclbin headers have the m_actionMask initialized to zeros.

maxzhen:
@sonals and @stsoe has approved it offline. merging it now.

PR #3480, opened 2020-06-01
should goto skip1 in kds_mode

mamin506:
.


houlz0507:
This is fixing the pipeline crash. Merge it right not.

PR #3486, opened 2020-06-02
CR-1066723 xbutil validate does not catch that the SC is in a bad state

chienwei-lan:


maxzhen:
have we tested it with a u.2 board which does not have a SC?

chienwei-lan:
> have we tested it with a u.2 board which does not have a SC?
It passes, won't break U.2-like board which doesn't have SC

-bash-4.2$ ./xbutil validate
INFO: Found 1 cards

INFO: Validating card[0]: xilinx_samsung_u2x4_201920_2
INFO: == Starting Kernel version check:
Developer's build. Skipping validation
INFO: == Kernel version check SKIPPED
AUX power connector not available. Skipping validation
INFO: == AUX power connector check SKIPPED

PR #3495, opened 2020-06-05
CR-1067055 host applications fails with XRT_202010.2.6.550 and later

xuhz:


mamin506:
I'm okay to check if ert is on here.

But if this condition leads to unexpected result, I guess cfg->slot_size is 0. If this is true, this is more like a test case issue.

Please comment the root cause here or in the CR.

xuhz:
In case of no ert resource(eg. u25 EA), cfg->cq_size is 0. ert config will fail, and hence open context, xbutil validate... fail

mamin506:
> In case of no ert resource(eg. u25 EA), cfg->cq_size is 0. ert config will fail, and hence open context, xbutil validate... fail

Correct my previous comment. The configure command is not constructed by user.
Not sure why cq_size is 0 in this case. But obviously, this change is simpler and safer than change scheduler.cpp.

xuhz:
cq_size is calculated from hardware resource during _probe(). There is no such kind of resource when ert is not available so it is 0

PR #3499, opened 2020-06-08
Fix missed #include <stdexcept>

matwey:
<stdexcept> is required for std::runtime_error

Brand new gcc 10.0 fails to compile the code pointing out that `std::runtime_error` is unknown name.


stsoe:
Thank you!

PR #3505, opened 2020-06-08
NSPR-7258 Streamline PDI download from host to VCK5000 via mgmt only PF

larry9523:
-------------To XRT developers--------------
This PR enable loading PDI, which is packed into xclbin, via mgmt PF on VCK5000.
1) Remove the current mechanism of transferring XCLBIN through userpf by XDMA
2) Remove loadXclbinToPS and combine it into xclLoadXclbin
3) Remove supporting the "pdi_load" key in xrt.ini
4) Remove the softkernel xclbin type
5) Remove "static_xclbin" operation on MPSOC in zocl driver
6) XCLBIN will be transferred from host memory to device memory through a 32K BRAM on mgmt function. This BRAM will be shared by transferring OSPI image and XCLBIN.

-------------To VCK5000 platform team--------------
Once merged, VCK5000 platform team will need to respin a platform including device side XRT.

-------------To VCK5000 users--------------
This PR will be an incompatible change for the old platform (including golden image). You will need to update to the new platform once install host XRT with this PR.

If new host XRT operates on old platform (including golden image):
1) You will NOT be able to load your PDI. xclLoadXclbin() will return fail with -ENOTSUPP
2) You will need to use "xbmgmt flash" to update your platform

After host XRT and device platform are all up-to-date,
1) You will NOT be able to use ”pdi_load=false" to prevend XRT from load PDI. Instead, make sure your XCLBIN_MODE in your XCLBIN header is set to XCLBIN_FLAT to prevent XRT loading PDI
2) PDI load time could be a little bit longer because we use PCIe PIO to replace XDMA when transfering XCLBIN to device memory. But this will only be observed on the first time you load XCLBIN. Once loaded, same XCLBIN will not be transferred any more.
3) You will NOT be able to load XCLBIN (first time) while you are updating your OSPI image

Please report bug if you see any issue or regression.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/ospi_versal.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/ospi_versal.c
@@ -16,11 +16,11 @@
  */
 
 /*
- * This subdriver is used to transfer Firmware Image from host memory
- * to device memory through a dedicated BRAM. This BRAM is mapped to
- * mgmt function and both drivers running on host and device can access
- * this BRAM. The first 4 Bytes of the BRAM is the packet header and
- * others are data payload.
+ * This subdriver is used to transfer data, e.g. firmware image, XCLBIN
+ * from host memory to device memory through a dedicated BRAM. This BRAM
+ * is mapped to mgmt function and both drivers running on host and device
+ * can access this BRAM. The first 4 Bytes of the BRAM is the packet
+ * header and others are data payload.
  *
  *                      ------------------
  *                     |    pkt_status    |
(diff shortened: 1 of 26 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/ospi_versal.c
```diff
@@ -16,11 +16,11 @@
  */
 
 /*
- * This subdriver is used to transfer Firmware Image from host memory
- * to device memory through a dedicated BRAM. This BRAM is mapped to
- * mgmt function and both drivers running on host and device can access
- * this BRAM. The first 4 Bytes of the BRAM is the packet header and
- * others are data payload.
+ * This subdriver is used to transfer data, e.g. firmware image, XCLBIN
```
mamin506, line 19: Should we change the name of this file to xfer_versal.c?

larry9523: I am glad you ask. And yes, we'd better make it consistency between the file name, module name, variable pre-fix, etc, including the device side zocl driver too. While I am making the change from ospi-->xfer, when we decide to call the BRAM from OSPI cache to XFER cache, the name is being discussed again. xfer wil not be the final name. I will follow up with a name change after we finalized the ssv2 name for this BRAM.

larry9523:
> additional question:
> Does versal still need icap.user for memory topology and clock min/max etc?

So far, yes. This PR does not address any of above.

larry9523:
> Have you tested this CR? http://jira.xilinx.com/browse/CR-1047555
> 
> I think the new design will remove this concern because we are using exact same flow as other Alevo broads.

No, I did not. I don't have a proper xcblin. I will let DSV to verify it.

PR #3506, opened 2020-06-09
CR-1064750 XRT WINDOWS : Timestamp mismatch not reported

igorshr:
CR-1064750. Because there is mismatch between the kernel return errors (NTSTATUS) and Win32 error codes the driver returns errors in the output parameters.


igorshr:
@stsoe. I added a comment. They are values returned from the driver. Unlike Linux, Windows has separate sets of the return values. One set is called NTSTATUS and used only in kernel, including driver. NTSTATUS values defined in <ntstatus.h>. This header file could not be compiled in a Win32 application. For Win32 application Windows uses another set of return codes. 
https://docs.microsoft.com/en-us/windows/win32/debug/system-error-codes
Both sets are not compatible. Because of this I added definitions of the error codes returned from XRT Windows device drivers.

PR #3509, opened 2020-06-09
CR-1067004 xbutil clock accuracy is not good enough for VCK5000

xdavidz:
We adjusted the clock wizard method, using a formula to avoid inaccuracy.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/clock.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/clock.c
@@ -509,7 +509,7 @@ static inline unsigned int floor_acap_o(int freq)
  */
 static inline unsigned int floor_acap_m(int freq)
 {
-	return (floor_acap_o(freq) * freq * 10000/333333);
+	return (floor_acap_o(freq) * freq * 1000/33333);
 }
 
 /*
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/clock.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/clock.c
@@ -310,163 +310,6 @@ const static struct xclmgmt_ocl_clockwiz {
 	{/*1462.500*/   650.000,        0x02710E01,     0x0000FA02}
 };
 
-/*
- * Due to float caculation is needed, the following table is generated
- * by this python script.
- *    #!/usr/bin/python
- *
- *    import re
- *    # import numpy as np
- *
- *    maxFreq = 650.0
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/clock.c
```diff
@@ -654,6 +497,21 @@ static inline int clock_wiz_busy(struct clock *clock, int idx, int cycle,
...
 }
 
+static inline unsigned int floor_acap_o(int freq)
+{
+	return (4320 / freq);
+}
+
+/*
+ * Kernel compiler even has disabled SSE(floating caculation) for preprocessor,
+ * we need a simple match to count floor without losing too much accuracy.
```
larry9523, line 507: do you mean math?

xdavidz: > do you mean math?

yes, fixed

src/runtime_src/core/pcie/driver/linux/xocl/subdev/clock.c
```diff
@@ -654,6 +497,21 @@ static inline int clock_wiz_busy(struct clock *clock, int idx, int cycle,
 	return 0;
 }
 
+static inline unsigned int floor_acap_o(int freq)
+{
+	return (4320 / freq);
+}
```
larry9523, line 503: Since we are removing the comment about the python script, does it make sense to comment or (better, if possible) predefine some macro about 4320, 1000, 333333?

xdavidz: For 4320:
There is a full document: https://confluence.xilinx.com/display/XIP/VCK5000+kernel+clocking. I don't know how reliable to put confluence into the code, possible shouldn't do that. 

I think 1000 with 33.333 is good enough. I tested 10000 with 333333, rolled it back.

larry9523: > For 4320:
> There is a full document: https://confluence.xilinx.com/display/XIP/VCK5000+kernel+clocking. I don't know how reliable to put confluence into the code, possible shouldn't do that.
> 
> I think 1000 with 33.333 is good enough. I tested 10000 with 333333, rolled it back.

definitely not the conf page. I am just suggesting a macro to explain what is 4320, something like 
#define VERSAL_CLK_MAX_VCO 4320
#define VERSAL_CLK_INPUT_FEQ 33.333

PR #3511, opened 2020-06-09
XRT-595 Add KDS IOPS test

mamin506:
Add IOPS test case.
This test include two implementations, the only different is the kernel related APIs.
One is using xclExecBuf() and xclExecWait(), which I see better result.
One is using xrtRunStart() and xclRunWait().


mamin506:
Thanks for the pointer. I would update the new XRT API test accordingly

PR #3523, opened 2020-06-10
Migrate to python3

AShivangi:


stsoe:
retest this please

stsoe:
I think we do.  We need to run ctest, if any of these tests are in python?  Also, you can run XRT without installing installing it, say emulation for example.

PR #3528, opened 2020-06-11
Additional information added to the System operation functions to determine the root cause of the issues for hw_emu

venkatp-xilinx:
Additional information added to the System operation functions to determine the root cause of the issues


venkatp-xilinx:
Can someone help me in merging this change. We have an issue which is appearing only in the regressions. Not able to reproduce it locally. This fix will enable us to get more messaging on the system operations and would be helpful in identifying the issue. This is just a messaging change, Please do merge this change.

maxzhen:
@chvamshi-xilinx , could you take a brief look and merge?

PR #3540, opened 2020-06-12
linux-driver: Fix mmap and build error

AlooGS:
1)Should use the linux/uaccess.h, it include all arches.

2)Make XOCL_FILE_PAGE_OFFSET be more applicable

static int xocl_mmap(struct file *filp, struct vm_area_struct *vma)
{
        /*
         * If the offset is > than 4G, then let GEM handle that and do what
         * it thinks is best, we will only handle offsets less than 4G.
         */
        if (likely(vma->vm_pgoff >= XOCL_FILE_PAGE_OFFSET))
                return xocl_bo_mmap(filp, vma);

Different architectrues have different page size, and can be choosed,
PPC64 support 4/16/64/256KB, MIPS support 4/8/16/32/64KB to choose.

So set XOCL_FILE_PAGE_OFFSET = (SZ_4G / PAGE_SIZE) is more applicable.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/qdma.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/qdma.c
@@ -28,7 +28,6 @@
 #include "../lib/libqdma/libqdma_export.h"
 #include "qdma_ioctl.h"
 
-#define XOCL_FILE_PAGE_OFFSET   0x100000
 #ifndef VM_RESERVED
 #define VM_RESERVED (VM_DONTEXPAND | VM_DONTDUMP)
 #endif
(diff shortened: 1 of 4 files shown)


maxzhen:
ok to test

src/runtime_src/core/pcie/driver/linux/xocl/subdev/qdma.c
```diff
@@ -28,7 +28,11 @@
 #include "../lib/libqdma/libqdma_export.h"
 #include "qdma_ioctl.h"
 
-#define XOCL_FILE_PAGE_OFFSET   0x100000
+#if (LINUX_VERSION_CODE < KERNEL_VERSION(4, 17, 0))
```
maxzhen, line 31: it might be better to use #ifndef here. centos may backport code from new kernel to their 3.10 kernel without bumping up the kernel version. so, the check against kernel version might not be accurate.

AlooGS: OK, it is a good idea, i have get and fix it.

maxzhen:
@AlooGS , Could you add "sign-off-by" as I mentioned? I'll then be happy to merge it.

AlooGS:
> @AlooGS , Could you add "sign-off-by" as I mentioned? I'll then be happy to merge it.

Ok.

PR #3567, opened 2020-06-16
NSPR-1387 : Xrt hbm support

saifuddin-xilinx:
Hi,
Creating this PR for review this HBM changes.
Please review the changes follows:

Userspace changes for XRT layer
User_PF Driver changes for HBM support
New changes support the following functionality:

User can specify a range of HBM banks in the v++ connectivity section, Example:
-sp CU_0.in1:HBM[0:3] . The user needs to specify bank-id information in his host application
as well using Extension Pointer.
If the user doesn't specify anything through the Extension pointer, Then XRT should automatically
detect the HBM banks using the connectivity section present in rtd file.

### DIFF
--- a/src/runtime_src/core/common/device.cpp
+++ b/src/runtime_src/core/common/device.cpp
@@ -24,6 +24,7 @@
 #include "core/include/xrt.h"
 #include "core/include/xclbin.h"
 #include "core/include/ert.h"
+#include "core/common/drv/include/mem_group.h"
 #include <boost/format.hpp>
 #include <string>
 #include <iostream>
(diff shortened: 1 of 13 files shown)


src/runtime_src/core/common/device.h
```diff
@@ -235,6 +236,13 @@ class device : public ishim
   std::pair<size_t, size_t>
   get_ert_slots() const;
 
+  /**
+   * populate_mem_group_info() - Store the memory group information
+   *
+   */
+  void
+  populate_mem_group_info(const char *infoBuff);
```
stsoe, line 244: No camel case

src/runtime_src/core/common/device.h
```diff
@@ -24,6 +24,7 @@
 #include "scope_guard.h"
 #include "uuid.h"
 #include "core/include/xrt.h"
+#include "core/common/drv/include/mem_group.h"
```
stsoe, line 27: Does mem_group.h need to be included in header file (device.h)?

src/runtime_src/core/common/device.cpp
```diff
@@ -78,6 +78,53 @@ register_axlf(const axlf* top)
...
   // Build modified CONNECTIVITY and MEM_TOPOLOGY section based on memory group ids
   // Base groups off data from driver
+  auto m_itr = m_axlf_sections.find(MEM_TOPOLOGY);
+  int g_count = (int)m_grp_info.size();
+  if (g_count && (m_itr != m_axlf_sections.end())) {
+    auto m_mem = reinterpret_cast<::mem_topology*>((*m_itr).second.data());
+    
+    /* Store the original memory topology before modify */
+    size_t size = (*m_itr).second.size();
+    char *buf = new char[size];
```
stsoe, line 88: No raw news.  use std::vector<char> buf(size);

src/runtime_src/core/common/device.cpp
```diff
@@ -78,6 +78,53 @@ register_axlf(const axlf* top)
 
   // Build modified CONNECTIVITY and MEM_TOPOLOGY section based on memory group ids
   // Base groups off data from driver
+  auto m_itr = m_axlf_sections.find(MEM_TOPOLOGY);
+  int g_count = (int)m_grp_info.size();
+  if (g_count && (m_itr != m_axlf_sections.end())) {
+    auto m_mem = reinterpret_cast<::mem_topology*>((*m_itr).second.data());
```
stsoe, line 84: Can you use get_axlf_section<const mem_topology*>() ?

src/runtime_src/core/common/device.cpp
```diff
@@ -78,6 +78,53 @@ register_axlf(const axlf* top)
...
+	    mem.m_used = 0;
+    }
+    // Update the new memory topology count 
+    m_mem->m_count = g_count;
+    delete []buf;
+  }
+
+  auto c_itr = m_axlf_sections.find(CONNECTIVITY);
+  if (c_itr != m_axlf_sections.end()) {
+    auto m_con = reinterpret_cast<::connectivity*>((*c_itr).second.data());
```
stsoe, line 118: Maybe use the get_axlf_section templated membor function?

src/runtime_src/core/common/device.h
```diff
@@ -307,6 +315,12 @@ class device : public ishim
   // cache xclbin meta data loaded by this process
   uuid m_xclbin_uuid;
   std::map<axlf_section_kind, std::vector<char>> m_axlf_sections;
+
+  // Store memory mapping information <<cuIdx, argIdx>, grpIdx>
+  std::map<std::pair<uint32_t, uint32_t>, uint32_t> m_grp_map;
+
+  // Store memory group information <grpIdx, <startAddr, size>>
+  std::map<uint32_t, std::pair<uint64_t, uint64_t>> m_grp_info;
```
stsoe, line 323: Are these needed as data members?    Seems like temporary implementation data used during construction of modified sections.  I think these should be create and returned by proper function called by user of the data.

src/runtime_src/core/common/device.h
saifuddin-xilinx: Ok. I will move it to device.cpp file.

src/runtime_src/core/common/device.cpp
saifuddin-xilinx: The return type of get_axlf_section if std::pair<**const char***, size_t>. But in our case, we need to modify the mem_topology, which is not allowed.

src/runtime_src/core/common/device.cpp
saifuddin-xilinx: The return type of get_axlf_section if std::pair<const char*, size_t>. But in our case, we need to modify the connectivity, which is not allowed because of const.

src/runtime_src/core/common/device.h
saifuddin-xilinx: You are correct. This is a temporary data. 
Currently, from shim.cpp layer updating those data member and from device.cpp we are accessing and updating the mem_topology..
If we want to avoid the temporary data member then we need to access the following path:
device.cpp -> ishim.h -> shim.cpp 
Please suggest if we need to do that.

src/runtime_src/core/common/device.cpp
stsoe: I see.  This makes sense.  I still think you should use the existing method and maybe const_cast the data, or alternatively provide a non-const private method that does just that.

src/runtime_src/core/common/device.cpp
stsoe: yep, see other response.

src/runtime_src/core/common/device.cpp
saifuddin-xilinx: I will update it.

src/runtime_src/core/common/device.h
stsoe: Ok. got it.  I suggest you change register_axlf that takes both the axlf and the pGrpInfo (default nullptr), then merge / refactor xrt_core::device::populate_mem_group_info and register_axlf appropriately.  In xclLoadXclBin you call the modified register_axlf with the extra pGrpInfo argument.

saifuddin-xilinx: ok. This make sense.. I will modify this..

PR #3572, opened 2020-06-17
CR-1067817 add new condition check for runtime clock scaling feature config

rajkumar-xilinx:
The new bit 30 (in 0xB20 reg) set to '1' indicates that the currently
loaded SC firmware does not support clock throttling.
XRT can set bit 29 (Feature present) if the ULP metadata indicates that
hardware support for throttling is loaded, but should not set bit 28
(Feature enabled) if the new bit 30 is set to '1'.
A dmesg report should be generated to indicate that runtime clock
throttling is not supported if the user tries to enable clock throttling
when bit 30 is set to '1'.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
@@ -1460,45 +1460,54 @@ static bool scaling_condition_check(struct xocl_xmc *xmc, struct device *dev)
 	bool runtime_cs_enabled = false;
 	bool sc_no_cs = false;
 
+	if (!XMC_PRIVILEGED(xmc)) {
+		xocl_dbg(dev, "Runtime clock scaling is not supported in non privileged mode\n");
+		return false;
+	}
+
 	if (!xmc->sc_presence) {
 		void *xdev_hdl = xocl_get_xdev(xmc->pdev);
-		if (xocl_clk_scale_on(xdev_hdl)) {
+		if (xocl_clk_scale_on(xdev_hdl))
 			cs_on_ptfm = true;
-			reg = READ_RUNTIME_CS(xmc, XMC_CLOCK_CONTROL_REG);
-			if (reg & XMC_CLOCK_SCALING_EN)
-				runtime_cs_enabled = true;
-		}
 	} else {
 		//Feature present bit may configured each time an xclbin is downloaded,
 		//or following a reset of the CMC Subsystem. So, check for latest
 		//status every time.
 		reg = READ_REG32(xmc, XMC_HOST_NEW_FEATURE_REG1);
-		if (reg & XMC_HOST_NEW_FEATURE_REG1_SC_NO_CS) {
+		if (reg & XMC_HOST_NEW_FEATURE_REG1_SC_NO_CS)
 			sc_no_cs = true;
-		} else {
-			if (reg & XMC_HOST_NEW_FEATURE_REG1_FEATURE_PRESENT)
-				cs_on_ptfm = true;
-			if (reg & XMC_HOST_NEW_FEATURE_REG1_FEATURE_ENABLE)
-				runtime_cs_enabled = true;
-		}
+		if (reg & XMC_HOST_NEW_FEATURE_REG1_FEATURE_PRESENT)
+			cs_on_ptfm = true;
 	}
 
-	if (!runtime_cs_enabled) {
-		if (!XMC_PRIVILEGED(xmc)) {
-			xocl_dbg(dev, "Runtime clock scaling is not supported in non privileged mode\n");
-		} else if (sc_no_cs) {
-			xocl_dbg(dev, "Loaded SC firmware does not support Runtime clock scalling\n");
-			return false;
-		} else if (cs_on_ptfm) {
-			xocl_dbg(dev, "Runtime clock scaling is not enabled\n");
-			return true;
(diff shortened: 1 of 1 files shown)


rajkumar-xilinx:
> Clock throttling is enabled/disabled based on indicators on HW (sc_no_cs, cs_on_ptfm and runtime_cs_enabled). We fetch these indicators every time we are asked to disable or enable clock throttling via sysfs node. Do you think we still need to fetch and cache them in xmc_probe()? Can we just make them local variables in scaling_condition_check()?

Hi @maxzhen ,
Currently, we are supporting two flows in xmc driver, one for Alveo U50 and second is for Samsung U2.
For U2, we are reading hw indicators only once and cache them during xmc_probe.

But, for U50, we have to read them every time as this feature present/enable bit might configured each time xclbin is download or following a reset of the CMC Subsystem. So, check for latest status every time. This change was added as part of http://jira.xilinx.com/browse/CR-1051402 for Alveo U50.

maxzhen:
> 
> 
> > Clock throttling is enabled/disabled based on indicators on HW (sc_no_cs, cs_on_ptfm and runtime_cs_enabled). We fetch these indicators every time we are asked to disable or enable clock throttling via sysfs node. Do you think we still need to fetch and cache them in xmc_probe()? Can we just make them local variables in scaling_condition_check()?
> 
> Hi @maxzhen ,
> Currently, we are supporting two flows in xmc driver, one for Alveo U50 and second is for Samsung U2.
> For U2, we are reading hw indicators only once and cache them during xmc_probe.
> 
> But, for U50, we have to read them every time as this feature present/enable bit might configured each time xclbin is download or following a reset of the CMC Subsystem. So, check for latest status every time. This change was added as part of http://jira.xilinx.com/browse/CR-1051402 for Alveo U50.

Why do you need two flows at all? Can checking-every-time flow cover U.2 case?

rajkumar-xilinx:
Hi @maxzhen ,
> Why do you need two flows at all? 
We need two flows as U50 is using new CMC register 0xB20 to indicate feature present/enable and U2 uses feature rom to know if clock scaling is supported on the current platform.

> Can checking-every-time flow cover U.2 case?
Yes, it can be done. But, everytime we need to invoke feature rom sub device to check if the clock scaling is supported or not, leads to more function callbacks from xmc at runtime. So, we just cached the same in xmc_probe as cs_on_ptfm (one time operation).

maxzhen:
> 
> 
> Hi @maxzhen ,
> 
> > Why do you need two flows at all?
> > We need two flows as U50 is using new CMC register 0xB20 to indicate feature present/enable and U2 uses feature rom to know if clock scaling is supported on the current platform.
> 
> > Can checking-every-time flow cover U.2 case?
> > Yes, it can be done. But, everytime we need to invoke feature rom sub device to check if the clock scaling is supported or not, leads to more function callbacks from xmc at runtime. So, we just cached the same in xmc_probe as cs_on_ptfm (one time operation).

This design does not make sense. We are not doing it in critical path. more function calls is completely fine and you can't avoid it when you are on U50. There is no point optimizing the code here and we suffer from maintaining two flows.

rajkumar-xilinx:
> sorry that i might not be clear here.... i'm thinking that we should remove the checking for clock throttling from xmc_probe() completely. you're going to do the check when you are requested to disable / enable it later anyway. let me know if this is not going to work for all cases.

It will work. We can have these checks in probe just to display runtime clock scaling feature supported or not on the platform/board during drivers loading itself.
If supported --> "Runtime clock scaling is supported." will be displayed,
else no such print will come on dmesg.
Please let me know if it is not required to display such messages?

maxzhen:
> 
> 
> > sorry that i might not be clear here.... i'm thinking that we should remove the checking for clock throttling from xmc_probe() completely. you're going to do the check when you are requested to disable / enable it later anyway. let me know if this is not going to work for all cases.
> 
> It will work. We can have these checks in probe just to display runtime clock scaling feature supported or not on the platform/board during drivers loading itself.
> If supported --> "Runtime clock scaling is supported." will be displayed,
> else no such print will come on dmesg.
> Please let me know if it is not required to display such messages?

I'm OK with just displaying it. But you should not duplicate the code in two places. You can call the check condition function in probe.

rajkumar-xilinx:
Hi @maxzhen ,
I have taken care of your suggestions, please review it again, thank you.

rajkumar-xilinx:
Hi @chvamshi-xilinx ,
Please review the change and let me know.

rajkumar-xilinx:
> I am ok with the changes. Could you please double confirm if the current change is backward compatible with older CMC version?

It is compatible since the added new field position 30th bit (in 0xB20 reg)'s default value is 0 hence sc fw supports clock scaling feature and new conditions which are added in xmc driver will be negative. So, functionality will not be changed for older cmc fw. I validated this change using shell image where this bit is not used (means 30th bit is 0). Did I answer your query?

xdavidz:
> > I am ok with the changes. Could you please double confirm if the current change is backward compatible with older CMC version?
> 
> HI @xdavidz ,
> It is compatible since the added new field position 30th bit (in 0xB20 reg)'s default value is 0 hence sc fw supports clock scaling feature and new conditions which are added in xmc driver will be negative. So, functionality will not be changed for older cmc fw. I validated this change using shell image where this bit is not used (means 30th bit is 0). Did I answer your query?

Based on your explanation, it looks good to me.  Thanks for taking care of backward compatibility.

PR #3573, opened 2020-06-17
CR-1046399 XRT to support QDMA-MM on Versal

xdavidz:
We have been trying to test qdma for versal. There are indeed several other bugs in shell or qdma driver or metadata somewhere else, but the above code is just to enable the new device id for qdma pcie device.

I think we can enable it in XRT, so that other team can start their testing without asking us to send them customized patch.

If there are more code to be changed in XRT, we can always add necessary code later on.


src/runtime_src/core/pcie/driver/linux/xocl/devices.h
```diff
@@ -2997,8 +2998,9 @@ struct xocl_subdev_map {
 	{ XOCL_PCI_DEVID(0x10EE, 0x5019, PCI_ANY_ID, USER_QDMA) },	\
 	{ XOCL_PCI_DEVID(0x10EE, 0x501D, PCI_ANY_ID, USER_QDMA) },	\
 	{ XOCL_PCI_DEVID(0x10EE, 0x5031, PCI_ANY_ID, USER_SMARTN) },	\
-	{ XOCL_PCI_DEVID(0x10EE, 0x5029, PCI_ANY_ID, USER_XDMA_VERSAL) },	\
-	{ XOCL_PCI_DEVID(0x10EE, 0x5045, PCI_ANY_ID, USER_XDMA_VERSAL) }
+	{ XOCL_PCI_DEVID(0x10EE, 0x5029, PCI_ANY_ID, USER_XDMA_VERSAL) },\
+	{ XOCL_PCI_DEVID(0x10EE, 0x5045, PCI_ANY_ID, USER_XDMA_VERSAL) },\
+	{ XOCL_PCI_DEVID(0x10EE, 0x5049, PCI_ANY_ID, USER_XDMA_VERSAL) }
```
houlz0507, line 3003: Why do not directly use VERSAL_USER_RAPTOR2 here? Is this new added pci-id raptor2 only?

xdavidz: Yes, changed the code.

src/runtime_src/core/pcie/driver/linux/xocl/devices.h
```diff
@@ -3055,6 +3057,14 @@ struct xocl_subdev_map {
...
 		.priv_data = &XOCL_BOARD_VERSAL_USER_RAPTOR2,		\
 		.type = XOCL_DSAMAP_RAPTOR2 },				\
+	{ 0x10EE, 0x5048, PCI_ANY_ID,					\
+		.vbnv = "xilinx_vck5000",				\
+		.priv_data = &XOCL_BOARD_VERSAL_MGMT_RAPTOR2,		\
+		.type = XOCL_DSAMAP_RAPTOR2 },				\
+	{ 0x10EE, 0x5049, PCI_ANY_ID,					\
+		.vbnv = "xilinx_vck5000",				\
+		.priv_data = &XOCL_BOARD_VERSAL_USER_RAPTOR2,		\
+		.type = XOCL_DSAMAP_RAPTOR2 },				\
```
houlz0507, line 3067: Can these two be removed? Just curious

xdavidz: Seems working too. The .vbnv and .type should be override by the definition.

PR #3587, opened 2020-06-19
fixed srsr index

houlz0507:


chienwei-lan:
ddr_idx won't be accumulated if all banks are used?

houlz0507:
> ddr_idx won't be accumulated if all banks are used?

ok, added back ddr_idx++ after srsr call

PR #3595, opened 2020-06-22
CR-1054646 Need robust xbutil reset in VCK5000

xdavidz:
This is implemented based on the spec here:
https://confluence.xilinx.com/display/XIP/Versal+PCIe-Reset+Sequence

We have to keep ep_pmc_intr_00 for this release but it will be removed for next release and handled in CIPs. Simply removing ep_pmc_intr_00 from the metadata will just work. This design should be backward and forward compatible with different version of shells. Thanks for raptor2.0 framework.

Add @mamin506, please take a look and share your thoughts about U30. We do not have PMC, but the follow should be the same?


src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-utils.c
```diff
@@ -524,6 +524,10 @@ static void xclmgmt_reset_pci(struct xclmgmt_dev *lro)
 	xocl_pci_restore_config_all(lro);
 
 	xclmgmt_config_pci(lro);
+
+	if (xocl_pmc_enable_reset(lro) == -ENODEV) {
+		mgmt_info(lro, "no support for pmc pcie reset.\n");
```
houlz0507, line 529: Most of the platforms do not have pmc. Could you move the print message inside pcm sub device?

xdavidz: thanks, removed those logs.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/pmc.c
```diff
@@ -0,0 +1,237 @@
...
+	xocl_warn(&(pmc)->pmc_pdev->dev, fmt "\n", ##arg)
+#define	PMC_INFO(pmc, fmt, arg...)	\
+	xocl_info(&(pmc)->pmc_pdev->dev, fmt "\n", ##arg)
+#define	PMC_DBG(pmc, fmt, arg...)	\
+	xocl_dbg(&(pmc)->pmc_pdev->dev, fmt "\n", ##arg)
+
+#define	PMC_ERR1_STATUS_MASK	(1 << 24)
+#define	PMC_ERR_OUT1_EN_MASK	(1 << 24)
+#define	PMC_POR1_EN_MASK	(1 << 24)
+#define	PMC_ERR_OUT1_MASK	0x20
```
larry9523, line 30: A nit to the Macro name. It would make the code more readable if we name PMC_ERR_OUT1_MASK as PMC_REG_ERR_OUT1_MASK ? (just to differentiate from PMC_ERR1_STATUS_MASK)

src/runtime_src/core/pcie/driver/linux/xocl/subdev/pmc.c
```diff
@@ -0,0 +1,237 @@
...
+			val &= ~PMC_ERR1_STATUS_MASK;
+			XOCL_WRITE_REG32(val, pmc_intr); 
+		}
+
+		XOCL_WRITE_REG32(PMC_ERR_OUT1_EN_MASK, pmc_intr + PMC_ERR_OUT1_EN);
+		val = XOCL_READ_REG32(pmc_intr + PMC_ERR_OUT1_MASK);
+		if (val & PMC_ERR_OUT1_EN_MASK) {
+			PMC_ERR(pmc, "mask 0x%x for PMC_ERR_OUT1_MASK 0x%x "
+			    "should be 0.\n", PMC_ERR_OUT1_EN_MASK, val);
+			rc = -EINVAL;
```
larry9523, line 79: Nit: this seems to be a hardware misbehavior than a invalid argument? Would -EIO be more proper

src/runtime_src/core/pcie/driver/linux/xocl/subdev/pmc.c
```diff
@@ -0,0 +1,237 @@
...
+			goto done;
+		}
+
+		XOCL_WRITE_REG32(PMC_POR1_EN_MASK, pmc_intr + PMC_POR1_EN);
+		val = XOCL_READ_REG32(pmc_intr + PMC_POR1_MASK);
+		if (val & PMC_POR1_EN_MASK) {
+			PMC_ERR(pmc, "mask 0x%x for PMC_POR1_MASK 0x%x "
+			    "should be 0.\n", PMC_POR1_EN_MASK, val);
+			rc = -EINVAL;
```
larry9523, line 88: Same nit

src/runtime_src/core/pcie/driver/linux/xocl/subdev/pmc.c
xdavidz: fixed.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/pmc.c
xdavidz: fixed

src/runtime_src/core/pcie/driver/linux/xocl/subdev/pmc.c
xdavidz: fixed

xdavidz:
> Generally looks good.
> 
> A question, when we return from xclmgmt_reset_pci(), do we know if device side reset is done? Do we know if PS linux is up and running?

Thanks, fixed those nits. 

I have tested xbutil reset then a immediate xbutil query. Worked well. The A72 boots in 10~30 seconds.

I think the only concern is the ERT. We might need full test cycle to see if there are windows that we should not allow comments squeeze in.

PR #3597, opened 2020-06-22
xrt::kernel support up to 128 CUs

stsoe:


stsoe:
Larry, please check the updates to run_update where the init command is used and has to offset for multiple CU masks

PR #3600, opened 2020-06-23
CR-1067079 reduce Over Temp limit to 110c from 125c on SmartSSD U2 board

rajkumar-xilinx:
Reducing over temperature threshold value from 125c to 110c for SmartSSD U2 board.
12 bit ADC code which has to be programmed in over temp upper reg (53h), needs to be calculated using temperature sensor transfer functions defined at equation 4-2 from UG580 document.

The default OT threshold is only overridden to a custom OT upper alarm threshold value when the OT upper alarm threshold register (53h) ends with 0011b (DI[3:0]).

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/devices.h
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/devices.h
@@ -53,7 +53,11 @@ enum {
 	XOCL_DSAFLAG_VERSAL			= (1 << 11),
 	XOCL_DSAFLAG_MPSOC			= (1 << 12),
 	XOCL_DSAFLAG_NOSC			= (1 << 13),
-	XOCL_DSAFLAG_OT_OVERRIDE		= (1 << 14),
+};
+
+/* sysmon flags */
+enum {
+	XOCL_SYSMON_OT_OVERRIDE		= (1 << 0),
 };
 
 #define	FLASH_TYPE_SPI	"spi"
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/devices.h
```diff
@@ -53,6 +53,7 @@ enum {
 	XOCL_DSAFLAG_VERSAL			= (1 << 11),
 	XOCL_DSAFLAG_MPSOC			= (1 << 12),
 	XOCL_DSAFLAG_NOSC			= (1 << 13),
+	XOCL_DSAFLAG_OT_OVERRIDE		= (1 << 14),
```
rajkumar-xilinx, line 56: We can still use "XOCL_DSAFLAG_NOSC" here since this change is only applicable for SmartSSD U.2 board for now.
But, if other platforms do require this kind of change in future, so added new flag "XOCL_DSAFLAG_OT_OVERRIDE" for this. Please let me know if XOCL_DSAFLAG_NOSC flag only to be used in the sysmon driver for this change also. I can make necessary change.

houlz0507: We should avoid using global flag if possible. Could you remove this flag? Instead, you may define a private data for sysmon subdevice to achieve this. Please search "xocl_flash_privdata" in devices.h as an example.

rajkumar-xilinx: Hi @houlz0507 , I have taken care of your suggestions in the change, can you please review the change again and let me know. Thank you.

PR #3602, opened 2020-06-23
should reset ulp axigate during xclbin download for vck5000

xdavidz:
Mark requested this changes. Logically, we should reset axigate for xclbin download. 

Test Result:
```
run this script over 100+ times /proj/rdi/staff/davidzha/share/clock_stress.sh
Test(104): sleep 1 s
verify PASSED.

```

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/ospi_versal.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/ospi_versal.c
@@ -421,12 +421,11 @@ static int xclbin_load_axlf(struct platform_device *pdev, const void *buf)
 	}
 	xocl_subdev_destroy_by_level(xdev, XOCL_SUBDEV_LEVEL_URP);
 
-	ret = xocl_axigate_freeze(xdev, XOCL_SUBDEV_LEVEL_PRP);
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/ospi_versal.c
```diff
@@ -421,15 +421,20 @@ static int xclbin_load_axlf(struct platform_device *pdev, const void *buf)
 	}
 	xocl_subdev_destroy_by_level(xdev, XOCL_SUBDEV_LEVEL_URP);
 
+	ret = xocl_axigate_freeze(xdev, XOCL_SUBDEV_LEVEL_PRP);
+	if (ret)
+		goto done;
```
larry9523, line 426: will this break the legacy platform support? I don't think we have axigate subdriver there.

xdavidz: yes, do just ignore the return value. The axigate freeze and free will only return 0 or -ENODEV;.

larry9523:
nit: line 411 and 415 have extra blank space at the end of line.

PR #3611, opened 2020-06-24
xbutil --new reset

AShivangi:
- support xbutil hot reset (same functionality as xbmgmt hot reset)
- xbmgmt reset default type is now "hot"

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/SubCmdReset.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/SubCmdReset.cpp
@@ -34,34 +34,27 @@ namespace po = boost::program_options;
 // ----- C L A S S   M E T H O D S -------------------------------------------
 
 static void
-pretty_print_action_list(xrt_core::device_collection& deviceCollection, const std::string& reset)
+pretty_print_action_list(xrt_core::device_collection& deviceCollection, XBU::reset_type reset)
 {
-  if(reset.compare("hot") == 0) {
+  if(reset == XBU::reset_type::hot)
     std::cout << "Performing 'hot' reset on " << std::endl;
-  } 
-  else if(reset.compare("kernel") == 0) {
+  else if(reset == XBU::reset_type::kernel)
     std::cout << "Performing DFX region reset on " << std::endl;
-  } 
-  else if(reset.compare("ert") == 0) {
+  else if(reset == XBU::reset_type::ert)
     std::cout << "Performing PS ERT reset on" << std::endl;
-  } 
-  else if(reset.compare("ecc") == 0) {
+  else if(reset == XBU::reset_type::ecc)
     std::cout << "Resetting all ECC counters on " << std::endl;
-  } 
-  else if(reset.compare("soft-kernel") == 0) {
+  else if(reset == XBU::reset_type::soft_kernel)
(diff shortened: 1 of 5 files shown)


src/runtime_src/core/tools/xbmgmt2/SubCmdReset.cpp
```diff
@@ -135,7 +135,7 @@ SubCmdReset::execute(const SubCmdOptions& _options) const
   XBU::verbose("SubCommand: reset");
   // -- Retrieve and parse the subcommand options -----------------------------
   std::vector<std::string> devices;
-  std::string type = "";
+  std::string type = "hot";
```
rozumx, line 138: Let's rename the variable "type" to "resetType"

src/runtime_src/core/tools/xbutil2/SubCmdReset.cpp
```diff
@@ -48,20 +76,20 @@ SubCmdReset::execute(const SubCmdOptions& _options) const
 {
   XBU::verbose("SubCommand: reset");
   // -- Retrieve and parse the subcommand options -----------------------------
-  std::string device = "all";
-  std::string reset = "all";
+  std::vector<std::string> devices = {"all"};
+  std::string reset = "hot";
```
rozumx, line 80: Please rename to "resetType"

src/runtime_src/core/tools/xbutil2/SubCmdReset.cpp
```diff
@@ -28,6 +29,33 @@ namespace po = boost::program_options;
 #include <iostream>
 
 // ----- C L A S S   M E T H O D S -------------------------------------------
+static void
+pretty_print_action_list(xrt_core::device_collection& deviceCollection, const std::string& reset)
```
rozumx, line 33: This method does "more" then what the name implies. 

It does a DRC check to determine if the option is valid AND does a pretty print.  Business rule logic should not be included in this method OR the method should be renamed.

src/runtime_src/core/tools/xbutil2/SubCmdReset.cpp
```diff
@@ -92,9 +120,23 @@ SubCmdReset::execute(const SubCmdOptions& _options) const
...
 
-  XBU::error("COMMAND BODY NOT IMPLEMENTED.");
-  // TODO: Put working code here
+  // Ask user for permission
+  if(!XBU::can_proceed())
+    return;
+
+  //perform reset actions
+  for (const auto & dev : deviceCollection) {
+    reset_device(dev, reset);
```
rozumx, line 138: The code should not be passing reset string names.  It should instead be using enumeration types. 

Please create:

1. An enumeration type to handle all of the reset values.
2. Create two methods to translate "strings to the enumeration type" and "enumeration type to strings".  
3. Update xbmgmt's code base to use the enumeration types as well.  

Note: Most likely this enumeration type and helper methods will be located in core/tools/common.

AShivangi: That's a good idea. I created enum class in XBUtilities and created a string to enum function which uses a map to do the conversion. I don't need the enum to string method.

PR #3614, opened 2020-06-24
XRT-604 xmc: skip loading xmc image from XRT driver when it is loading through FPGA bitfile

rajkumar-xilinx:
New proposed shell(NSPR-7555 & NSPR-7553) will embeds CMC.elf in the bitfile and
it will separates out the ERT & CMC reset, so XRT can avoid
resetting the CMC but continue to load/reset ERT as it is doing today.
Added required changes in XMC driver to support this change.

This change is backward compatible since the register field which
used here to determine if xmc is in bitfile will have 0 value in old shells.
And, the new shell which embeds cmc.elf in bitfile will have 1 value.


src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
```diff
@@ -2861,6 +2876,7 @@ static int load_mgmt_image(struct platform_device *pdev, const char *image,
 
 	if (binary)
 		vfree(binary);
+	xocl_info(&xmc->pdev->dev, "%s: load mgmt image len %d", __func__, len);
```
rajkumar-xilinx, line 2879: I will remove this line.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
```diff
@@ -2710,6 +2723,8 @@ static int load_xmc(struct xocl_xmc *xmc)
 				xmc->mgmt_binary_length);
 			COPY_MGMT(xmc, xmc->mgmt_binary, xmc->mgmt_binary_length);
 		}
+	} else {
+		xocl_info(&xmc->pdev->dev, "Skip copying XMC image");
```
rajkumar-xilinx, line 2727: I will add more info "Skipping loading xmc image since xmc is already part of bitfile"

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
```diff
@@ -2574,40 +2582,42 @@ static int stop_xmc_nolock(struct platform_device *pdev)
...
+			 * has also finished */
+			if (retry >= MAX_XMC_RETRY) {
+				xocl_err(&xmc->pdev->dev, "Failed to stop XMC, Error Reg 0x%x",
+						 READ_REG32(xmc, XMC_ERROR_REG));
+				xmc->state = XMC_STATE_ERROR;
+				return -ETIMEDOUT;
+			}
+			xocl_info(&xmc->pdev->dev, "XMC Stopped, retry %d",	retry);
+		} else {
+			xocl_info(&xmc->pdev->dev, "Skip XMC stop");
```
rajkumar-xilinx, line 2609: Will add more info "Skipping stopping of xmc since xmc is part of bitfile".

chvamshi-xilinx:
> 
> 
> we don't load cmc image as long as the bit is set even when the cmc image from xsabin is newer?

yes @maxzhen . XRT shouldn't load CMC which can break the IIC communication between NVME and FPGA.

This is needed for their Continuous I2C communication between NVME and FPGA.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
```diff
@@ -2574,40 +2582,42 @@ static int stop_xmc_nolock(struct platform_device *pdev)
...
-			WRITE_REG32(xmc, CTL_MASK_STOP, XMC_CONTROL_REG);
-			WRITE_REG32(xmc, 1, XMC_STOP_CONFIRM_REG);
-		}
-
-		retry = 0;
-		while (retry++ < MAX_XMC_RETRY &&
-			!(READ_REG32(xmc, XMC_STATUS_REG) &
-			STATUS_MASK_STOPPED))
-			msleep(RETRY_INTERVAL);
+		u32 misc = READ_VERSION_CTRL_REG32(xmc, VERSION_CTRL_MISC_REG);
```
houlz0507, line 2585: This check applies to all platforms, correct? Will this register be zero for other platforms? (u200, u250, etc)

rajkumar-xilinx: Not guaranteed actually, thanks for pointing this case. I will add one more check as this change is needed only for SmartSSD U.2

houlz0507: > Not guaranteed actually, thanks for pointing this case. I will add one more check as this change is needed only for SmartSSD U.2
Maybe you may define "__RES_VERSION_CTRL" only for U2 in devices.h?

rajkumar-xilinx:
Hi @chienwei-lan , Can you please review this change?

PR #3622, opened 2020-06-25
XRT-619 ep_fpga_configuration_00 for icap and versal xfer

xdavidz:
Sorry, lots of code and it is urgent (the last fix for releasing versal in 2020.1_web).

Basically, both ICAP and XFER of versal IP are sharing same ep name but they have different compatible value.

I ended up with changing the res_names[] array to struct xocl_subdev_res res_array[] {.res_name, .platform_name}.
Then we can use the platform_name to match the name in ip_node(retrieved from fdt blob device tree).

Please review that for me carefully.


xdavidz:
retest this please

xdavidz:
retest this please

abalasa:
retest this please

PR #3624, opened 2020-06-25
memory leak fix

AShivangi:


src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -445,11 +445,11 @@ XBUtilities::get_axlf_section(const std::string& filename, axlf_section_kind kin
...
     throw std::runtime_error("Section not found");
 
-  auto buf = new char[section->m_sectionSize];
+  std::unique_ptr<char> buf(new char[section->m_sectionSize]);
   in.seekg(section->m_sectionOffset);
-  in.read(buf, section->m_sectionSize);
+  in.read(buf.get(), section->m_sectionSize);
 
-  return std::make_pair(buf, section->m_sectionSize);
+  return std::make_pair(buf.get(), section->m_sectionSize);
```
stsoe, line 452: No that wont work.  The unique_ptr goes out of scope at lone 453 and the returned buffer has been deleted.  You will have to pass ownership to the caller via std::unique_ptr in the returned pair.

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -413,7 +413,7 @@ XBUtilities::report_available_devices()
   std::cout << std::endl;
 }
 
-std::pair<const char*, size_t>
+std::pair<std::vector<char>, size_t>
```
stsoe, line 416: just return std::vector.  size of vector is same.

AShivangi: Oh yes, good point ;)

stsoe:
retest this please

PR #3625, opened 2020-06-25
two small fixes to get this compiling under gcc 10

boxerab:
My version of gcc:

```
gcc (Ubuntu 10.1.0-2ubuntu1~18.04) 10.1.0
```

### DIFF
--- a/src/runtime_src/core/include/xrt_mem.h
+++ b/src/runtime_src/core/include/xrt_mem.h
@@ -32,12 +32,12 @@ extern "C" {
 
 #define XRT_BO_FLAGS_MEMIDX_MASK	(0xFFFFFFUL)
 #define	XCL_BO_FLAGS_NONE		(0)
-#define	XCL_BO_FLAGS_CACHEABLE		(1 << 24)
-#define	XCL_BO_FLAGS_SVM		(1 << 27)
-#define	XCL_BO_FLAGS_DEV_ONLY		(1 << 28)
-#define	XCL_BO_FLAGS_HOST_ONLY		(1 << 29)
-#define	XCL_BO_FLAGS_P2P		(1 << 30)
-#define	XCL_BO_FLAGS_EXECBUF		(1 << 31)
+#define	XCL_BO_FLAGS_CACHEABLE		(1U << 24)
+#define	XCL_BO_FLAGS_SVM		(1U << 27)
+#define	XCL_BO_FLAGS_DEV_ONLY		(1U << 28)
+#define	XCL_BO_FLAGS_HOST_ONLY		(1U << 29)
+#define	XCL_BO_FLAGS_P2P		(1U << 30)
+#define	XCL_BO_FLAGS_EXECBUF		(1U << 31)
 
 /**
  * This is the legacy usage of XCL DDR Flags.
--- a/src/runtime_src/xocl/api/plugin/xdp/debug.cpp
+++ b/src/runtime_src/xocl/api/plugin/xdp/debug.cpp
@@ -18,6 +18,7 @@
 #include "core/common/dlfcn.h"
 #include "plugin/xdp/debug.h"
 #include "core/common/module_loader.h"
+#include <stdexcept>
 
 namespace xocl {
 namespace debug {


src/runtime_src/core/include/xrt_mem.h
```diff
@@ -37,7 +37,7 @@ extern "C" {
 #define	XCL_BO_FLAGS_DEV_ONLY		(1 << 28)
 #define	XCL_BO_FLAGS_HOST_ONLY		(1 << 29)
 #define	XCL_BO_FLAGS_P2P		(1 << 30)
-#define	XCL_BO_FLAGS_EXECBUF		(1 << 31)
+#define	XCL_BO_FLAGS_EXECBUF		(unsigned int)(1 << 31)
```
stsoe, line 40: This does not seem right place for cast,  where is the compile failure?

boxerab: Thanks. Since we are using C++14, I've now changed these to `const uint32_t`

src/runtime_src/core/include/xrt_mem.h
```diff
@@ -30,14 +30,14 @@ extern "C" {
...
 
-#define XRT_BO_FLAGS_MEMIDX_MASK	(0xFFFFFFUL)
-#define	XCL_BO_FLAGS_NONE		(0)
-#define	XCL_BO_FLAGS_CACHEABLE		(1 << 24)
-#define	XCL_BO_FLAGS_SVM		(1 << 27)
-#define	XCL_BO_FLAGS_DEV_ONLY		(1 << 28)
-#define	XCL_BO_FLAGS_HOST_ONLY		(1 << 29)
-#define	XCL_BO_FLAGS_P2P		(1 << 30)
-#define	XCL_BO_FLAGS_EXECBUF		(1 << 31)
+const uint32_t 	XRT_BO_FLAGS_MEMIDX_MASK   = 0xFFFFFF;
```
stsoe, line 33: Please include <stdint.h> for C and <cstdint> for __cplusplus;  the header file should be standalone compilable.

stsoe:
ok to test

boxerab:
Thank you.

boxerab:
> Hi @boxerab . Looks like we need some conditional include to handle kernel use of this header file. Something like:
> 
> ```
> #if defined(__KERNEL__)
> # include <linux/types.h>
> #else
> # include <stdint.h>
> #endif
> ```

Thanks, I have added this in.

stsoe:
Failure in kernel driver compilation. const uint32_t is not a compile time integer constant in C.

src/runtime_src/core/common/api/xrt_bo.cpp
```diff
@@ -319,12 +319,12 @@ alloc(xclDeviceHandle dhdl, size_t sz, xrtBufferFlags flags, xrtMemoryGroup grp)
...
-  case XCL_BO_FLAGS_DEV_ONLY:
-  case XCL_BO_FLAGS_HOST_ONLY:
-  case XCL_BO_FLAGS_P2P:
-  case XCL_BO_FLAGS_EXECBUF:
+  case (xrtBufferFlags)XCL_BO_FLAGS_CACHEABLE:
+  case (xrtBufferFlags)XCL_BO_FLAGS_SVM:
+  case (xrtBufferFlags)XCL_BO_FLAGS_DEV_ONLY:
+  case (xrtBufferFlags)XCL_BO_FLAGS_HOST_ONLY:
+  case (xrtBufferFlags)XCL_BO_FLAGS_P2P:
+  case (xrtBufferFlags)XCL_BO_FLAGS_EXECBUF:
```
stsoe, line 327: Isn't the compile failure is because xrtBufferFlags is of wrong type?  I don't think it should be uint64_t as it is in xrt_bo.h, but it should be unsigned int to match shim layer flags usage.   Let's revert this particular change and fix the typedef.  

Which OS are you using where you have access to gcc10?

abalasa:
retest this please

stsoe:
Alright. I installed gcc-10 and tried this.  Your first change with the casts is essentially needed, but let's do it this way:

```
#define XRT_BO_FLAGS_MEMIDX_MASK        (0xFFFFFFUL)
#define XCL_BO_FLAGS_NONE               (0u)
#define XCL_BO_FLAGS_CACHEABLE          (1u << 24)
#define XCL_BO_FLAGS_SVM                (1u << 27)
#define XCL_BO_FLAGS_DEV_ONLY           (1u << 28)
#define XCL_BO_FLAGS_HOST_ONLY          (1u << 29)
#define XCL_BO_FLAGS_P2P                (1u << 30)
#define XCL_BO_FLAGS_EXECBUF            (1u << 31)
```

with no other changes to this file.

src/runtime_src/core/common/api/xrt_bo.cpp
boxerab: I am using Ubuntu 20.x. I believe the error happens because `XCL_BO_FLAGS_EXECBUF` is treated as an `int`, 
but `1<<31` would be treated as -1, then cast to uint64_t would equal `0xFFFFFFFFFFFFFFF`.

stsoe:
Thank you for the changes.

boxerab:
> Thank you for the changes.

My pleasure.

PR #3639, opened 2020-06-29
XRT-608 fixing an issue if we run build_edge from different dir

chvamshi-xilinx:
fixing an issue if we run build_edge from different dir


chvamshi-xilinx:
retest this please

PR #3640, opened 2020-06-29
added device registration code in edge sw emu code base

vboggara-xilinx:
added device registration code in edge sw emu code base during load xclbin.


chvamshi-xilinx:
retest this please

PR #3647, opened 2020-06-30
User space cannot access profiling subdevices while running inside containers. So, use xclRead/Write flow for profiling.

IshitaGhosh:
Add new Debug configuration in xrt.ini : container=true/false. When container=true, profiling will use xclRead/Write based flow instead of open+mmap on subdevices. This is added to work around issue on cloud container based Nimbix where user space cannot access the subdevices in /dev/xfpga.


IshitaGhosh:
retest this please

abalasa:
retest this please

PR #3660, opened 2020-07-02
CR-1067868 Usage of the XRT API (xrtGraphUpdateRTP) for tx chain

larry9523:
Validate Graph RTP interface with a real AIE vadd/fadd design.
1) Added Read RTP interfaces
2) Fixed bugs in Update RTP and GraphWatiDone.

### DIFF
--- a/src/runtime_src/core/edge/user/aie/graph.cpp
+++ b/src/runtime_src/core/edge/user/aie/graph.cpp
@@ -54,6 +54,7 @@ graph_type(std::shared_ptr<xrt_core::device> dev, const uuid_t, const std::strin
     // this is not the right place for creating Aie instance. Should
     // we move this to loadXclbin when detect Aie Array?
     auto drv = ZYNQ::shim::handleCheck(device->get_device_handle());
+
     aieArray = drv->getAieArray();
     if (!aieArray) {
       aieArray = new Aie(device);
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/edge/user/aie/graph.cpp
```diff
@@ -261,34 +266,37 @@ update_rtp(const char* port, const char* buffer, size_t size)
...
 
-    if (rtp->require_lock)
+    /* Don't acquire selector lock for async RTP while graph is not running */
+    bool need_lock = rtp->require_lock && (
+        rtp->is_async && state == graph_state::running ||
+        !rtp->is_async
+	);
+
+    if (need_lock)
         XAieTile_LockAcquire(selector_tile, rtp->selector_lock_id, 0, 0xFFFFFFFF);
```
stsoe, line 296: LockAcquire and LockRelease is not a good pattern.  What happens in an exception is thrown in between?
Don't worry about it now, we can fix this later.

src/runtime_src/core/edge/user/aie/graph.cpp
```diff
@@ -306,33 +314,110 @@ update_rtp(const char* port, const char* buffer, size_t size)
...
+        lock_id = rtp->ping_lock_id;
+        start_addr = rtp->ping_addr;
+    }
+
+    XAieGbl_Tile *update_tile = &(aieArray->tileArray.at(update_pos));
+    if (need_lock) {
+        XAieTile_LockAcquire(update_tile, lock_id, 1, 0xFFFFFFFF);
+	/* sync RTP release lock for write, async RTP relase lock for read */
+        XAieTile_LockRelease(selector_tile, rtp->selector_lock_id, (rtp->is_async ? 1 : 0), 0XFFFFFFFF);
+    }
```
stsoe, line 399: This looks suspiciously identical to update_rtp?  Can some code be shared?

src/runtime_src/core/edge/user/aie/graph.cpp
larry9523: Yeah. I agree with you. Hopefully a later AIE driver could hide the lock acquire/release from the interface (better in kernel space).

src/runtime_src/core/edge/user/aie/graph.cpp
larry9523: I've been trying to provide one interface with a flag update/read RTP. But it seems not easy.
1) Because of the direction, we have to use different AIE driver interface DmReadWord/DmWriteWord
2) The selector lock acquire/release sequence and parameter for update and read are different. 
    a) update needs to pass in 0 while read needs to pass in 1 
    b) for read, selector lock needs to be released before accessing the tile memory
    c) for update, needs to release selector lock even we don't acquiring it when graph is not running. 
3) Parameter for update_tile lock are different

PR #3661, opened 2020-07-03
Enabling support for Rhel/Centos  7.7/7.8 and 8.1

rbramand-xilinx:


maxzhen:
have you tested xbutil validate? don't you need to update centos_rh_kernel_versions list?

rbramand-xilinx:
> have you tested xbutil validate? don't you need to update centos_rh_kernel_versions list?

Actually in this pull request  I added changes to build RPM's and I am in process of creating vm's to test xbutil validate and will update  that also.
Also we need to cherrypick  this commit  572fc1381b8af19f5cac3b28029145a5796343ec  to 2020.1 as  I faced pyopencl issue while installing the rpms.

PR #3667, opened 2020-07-06
Remove conformance code from OCL

stsoe:


keryell:
Just curious about the consequences...
There was no chance to have this useful again?
Was there a performance impact or normal use case?

stsoe:
Hi @keryell .  It was dead code, the flow has not been used for a couple of years.  If we need to resurrect, it should be done differently.

PR #3672, opened 2020-07-06
XDP HAL profiling refactoring

jvillarre:
1. Removing the reading of device profiling IP from the HAL plugin.  Now device offload and HAL API profiling are independently controlled via xrt.ini switches.

2. Hooking the loading of the device offload plugin (specific to HAL level) into the current HAL profiling loading infrastructure.

3. Making slight changes to events and filters so LOP, HAL, and device profiling can all be enabled at the same time.

Note:  These changes are still guarded in the shim by a compile time guard and should have no impact on general builds.


src/runtime_src/xdp/profile/plugin/device_offload/device_offload_plugin.cpp
```diff
@@ -93,9 +113,19 @@ namespace xdp {
 			     trace_buffer_size,            // trbuf_size,
 			     continuous_trace) ;           // start_thread
```
hackwa, line 114: should pass false here if calling read_trace_init. this code needs to be looked into in details.

PR #3677, opened 2020-07-07
Improving the fix to check for the platform using legacy_ert

hemantk-xilinx:
The same check is applicable for all minor version of the platforms so omitting them from comparison as the new minor release can be done later


chvamshi-xilinx:
ok to test

PR #3692, opened 2020-07-09
CR-1064951 Exclusive CU reporting error

rbramand-xilinx:
> xclopencontext with exclusive access is throwing error of Unreserved CU error
> Updated zocl_xclbin.c file to fix this

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
@@ -799,7 +799,7 @@ zocl_xclbin_ctx(struct drm_zocl_dev *zdev, struct drm_zocl_ctx *ctx,
       }
 			if (!ret) {
 				DRM_ERROR("can not remove unreserved cu");
-        ret = -EINVAL;
+        			ret = -EINVAL;
 				goto out;
 			}
 		}
--- a/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
@@ -792,14 +792,14 @@ zocl_xclbin_ctx(struct drm_zocl_dev *zdev, struct drm_zocl_ctx *ctx,
 
 		if (cu_idx != ZOCL_CTX_VIRT_CU_INDEX) {
 			/* Try clear exclusive CU */
-			ret = test_and_clear_bit(cu_idx, client->excus) ?
-                                  0 : -EINVAL;
-			if (ret)
+			ret = test_and_clear_bit(cu_idx, client->excus);
+			if (!ret) {
 				/* Maybe it is shared CU */
-				ret = test_and_clear_bit(cu_idx, client->shcus) ?
-                                  0 : -EINVAL;
-			if (ret) {
+				ret = test_and_clear_bit(cu_idx, client->shcus);
+      }
+			if (!ret) {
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx:
Petalinux Issue :-(

chvamshi-xilinx:
retest this please

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -792,8 +792,9 @@ zocl_xclbin_ctx(struct drm_zocl_dev *zdev, struct drm_zocl_ctx *ctx,
 
 		if (cu_idx != ZOCL_CTX_VIRT_CU_INDEX) {
 			/* Try clear exclusive CU */
-			ret = test_and_clear_bit(cu_idx, client->excus);
-			if (!ret)
+			ret = test_and_clear_bit(cu_idx, client->excus) ?
+                                  0 : -EINVAL;
+			if (ret)
```
larry9523, line 797: It is a little bit weird of using "-EINVAL" here. The "-EINVAL" does not really play a role. It could be any non-zero value. What you really want to to do is to reverse the return value of test_and_clear_bit so that we don't hit the second if statement. I would suggest a different approach, like
if (cu_idx != ZOCL_CTX_VIRT_CU_INDEX) {
        /* Try clear exclusive CU */
        ret = test_and_clear_bit(cu_idx, client->excus);
        if (!ret) {
                /* Maybe it is shared CU */
                ret = test_and_clear_bit(cu_idx, client->shcus) ?
                                0 : -EINVAL;
                if (ret) {
                        DRM_ERROR("can not remove unreserved cu");
                        goto out;
                }
        }
}
or, I prefer the following fix which could scale if we have other CU types.
if (cu_idx != ZOCL_CTX_VIRT_CU_INDEX) {
        /* Try clear exclusive CU */
        ret = test_and_clear_bit(cu_idx, client->excus);
        if (!ret) {
                /* Maybe it is shared CU */
                ret = test_and_clear_bit(cu_idx, client->shcus);
        }
        if (!ret) {
                DRM_ERROR("can not remove unreserved cu");
                ret = -EINVAL;
                goto out;
        }
}

Please let me know what do you think. Thanks!

rbramand-xilinx: In the 1st method it will again hit DRM_ERROR case if it is exclusive cu. Second method is perfect, I will update and push the change. Thanks larry.

rbramand-xilinx: changed the code as suggested, please review

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -793,12 +793,13 @@ zocl_xclbin_ctx(struct drm_zocl_dev *zdev, struct drm_zocl_ctx *ctx,
...
 			/* Try clear exclusive CU */
 			ret = test_and_clear_bit(cu_idx, client->excus);
-			if (!ret)
+			if (!ret) {
 				/* Maybe it is shared CU */
-				ret = test_and_clear_bit(cu_idx, client->shcus) ?
-                                  0 : -EINVAL;
-			if (ret) {
+				ret = test_and_clear_bit(cu_idx, client->shcus);
+      }
```
xdavidz, line 799: ...

PR #3696, opened 2020-07-10
CR-1069847 mgmt probe fail with static xfer on 2019 shell

xdavidz:
Tested on vck5000 2020.1 shell.

RCA:

xfer subdev can fail which causes OS panic. 

Solution:

2020.1_web shell has xfer resource in VSEC, reading from VSEC.

for 2019.2 shell, there is no VSEC, xfer will be loaded from static config.

for 2020.1_web shell, xfer will be loaded from VSEC config

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/xocl_subdev.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/xocl_subdev.c
@@ -1292,6 +1292,16 @@ bool xocl_subdev_is_vsec(xdev_handle_t xdev)
 	return pci_find_ext_capability(pdev, PCI_EXT_CAP_ID_VNDR) != 0;
 }
 
+static inline int xocl_subdev_create_vsec_impl(xdev_handle_t xdev,
+	struct xocl_subdev_info *info, u64 offset, int bar)
+{
+	info->res[0].start = offset;
+	info->res[0].end = offset + 0xfff;
+	info->bar_idx[0] = bar;
+
+	return xocl_subdev_create(xdev, info);
+}
+
 int xocl_subdev_create_vsec_devs(xdev_handle_t xdev)
 {
 	u64 offset;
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/xocl_subdev.c
```diff
@@ -1301,36 +1311,61 @@ int xocl_subdev_create_vsec_devs(xdev_handle_t xdev)
...
-		struct xocl_subdev_info subdev_info = XOCL_DEVINFO_FLASH_VSEC;
+		struct xocl_subdev_info subdev_info =
+		    XOCL_DEVINFO_FLASH_VSEC;
+
+		switch (vtype) {
+		case XOCL_VSEC_FLASH_TYPE_QSPI:
+			memcpy(((struct xocl_flash_privdata *)
+			    (subdev_info.priv_data))->flash_type,
+			    FLASH_TYPE_QSPIPS, strlen(FLASH_TYPE_QSPIPS));
+			/* default is FLASH_TYPE_SPI, thus pass through. */
```
larry9523, line 1322: should we add __attribute__ ((fallthrough)); to suppress GCC warning?
https://github.com/Xilinx/XRT/pull/2405/files

src/runtime_src/core/pcie/driver/linux/xocl/xocl_subdev.c
```diff
@@ -1301,36 +1311,61 @@ int xocl_subdev_create_vsec_devs(xdev_handle_t xdev)
...
+			    FLASH_TYPE_OSPI_VERSAL, strlen(FLASH_TYPE_OSPI_VERSAL));
+
+			ret = xocl_subdev_create_vsec_impl(xdev, &subdev_info,
+				offset, bar);
+
+			if (ret)
+				return ret;
+			break;
+		default:
+			xocl_xdev_info(xdev, "Unsupport flash type 0x%x", vtype);
```
larry9523, line 1355: it will be better to add a break even in default (default does not have to be the last case.)

xdavidz:
Found an issue on u50, debugging.

xdavidz:
> Found an issue on u50, debugging.

It is ok, I flashed a wrong shell (u50_base_4), I should use the shell in the blp.

PR #3708, opened 2020-07-14
XRT-644 enabling Graph API support in hw_emu

chvamshi-xilinx:
This change enables XRT Graph API support for hw_emu.


abalasa:
retest this please

PR #3713, opened 2020-07-15
XDP: Initial implementation of power profiling as a HAL level plugin.

jvillarre:
This loading of this plugin module is still protected by the compile time guard, so by default there should be no change to any current applications.  This check-in does not remove the current OpenCL level power profiling.  When HAL level profiling is enabled, the OpenCL level power profiling should be removed in the same check-in.

### DIFF
--- a/src/runtime_src/core/common/config_reader.h
+++ b/src/runtime_src/core/common/config_reader.h
@@ -160,6 +160,13 @@ get_power_profile()
   return value;
 }
 
+inline unsigned int
+get_power_profile_interval_ms()
+{
+  static unsigned int value = detail::get_uint_value("Debug.power_profile_interval_ms", 20) ;
+  return value ;
+}
+
 inline std::string
 get_stall_trace()
 {
--- a/src/runtime_src/core/common/config_reader.h
+++ b/src/runtime_src/core/common/config_reader.h
@@ -156,7 +156,7 @@ get_data_transfer_trace()
 inline bool
 get_power_profile()
 {
-  static bool value = get_profile() && detail::get_bool_value("Debug.power_profile",false);
+  static bool value = detail::get_bool_value("Debug.power_profile",false);
   return value;
 }
 
--- a/src/runtime_src/xdp/profile/plugin/power/power_plugin.cpp
+++ b/src/runtime_src/xdp/profile/plugin/power/power_plugin.cpp
@@ -1,0 +1,141 @@
+/**
+ * Copyright (C) 2020 Xilinx, Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may
+ * not use this file except in compliance with the License. A copy of the
+ * License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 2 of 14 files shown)


src/runtime_src/xdp/profile/plugin/power/power_plugin.cpp
```diff
@@ -0,0 +1,141 @@
...
+	  std::getline(fs, data) ;
+	  int dp = data.empty() ? 0 : std::stoi(data) ;
+	  values.push_back(dp) ;
+	  fs.close() ;
+	}
+	(db->getDynamicInfo()).addPowerSample(index, timestamp, values) ;
+	++index ;	
+      }
+
+      std::this_thread::sleep_for(std::chrono::milliseconds(20)) ;
```
pgschuey, line 137: Please make the sample period user controllable.

src/runtime_src/xdp/profile/plugin/power/power_plugin.h
```diff
@@ -0,0 +1,51 @@
...
+  private:
+    // These are the files that will be opened and read
+    static const char* powerFiles[] ;
+
+  private:
+    std::vector<std::vector<std::string>> filePaths ;
+
+    // Power profiling requires its own thread
+    bool keepPolling ;
+    std::thread pollingThread ;
```
pgschuey, line 40: For now, this is fine. We might want to consider using a common thread for all power/counter sampling so as to minimize the number of threads we kick off just for profiling.

src/runtime_src/core/common/config_reader.h
```diff
@@ -156,7 +156,7 @@ get_data_transfer_trace()
 inline bool
 get_power_profile()
 {
-  static bool value = get_profile() && detail::get_bool_value("Debug.power_profile",false);
+  static bool value = detail::get_bool_value("Debug.power_profile",false);
```
pgschuey, line 159: If we're not going to use get_profile() as a top-level gating mechanism, then we should only use it for host profiling, change the name to get_host_profile(), and only expose individual types of profiling separately. Perhaps not something for this checkin, but a down the road issue to address.

hackwa: @pgschuey this previously also was used for host profiling as you needed other switches to get device data_transfers etc. It might be a bad idea to break backwards compatibility with last 5 releases.

src/runtime_src/xdp/profile/plugin/power/power_plugin.cpp
```diff
@@ -0,0 +1,141 @@
...
+    uint64_t index = 0 ;
+    void* handle = xclOpen(index, "/dev/null", XCL_INFO) ;
+    while (handle != nullptr)
+    {
+      // For each device, keep track of the paths to the sysfs files
+      std::vector<std::string> paths ;
+      for (auto f : powerFiles)
+      {
+	char sysfsPath[512] ;
+	xclGetSysfsPath(handle, "xmc", f, sysfsPath, 512) ;
```
hackwa, line 70: need a check here in case xclGetSysfsPath returns a null path. In that case we need to skip the device

jvillarre: In order to make sure all devices have the same number of entries (one per file defined in powerFiles) and all the data is aligned, I want to push the null path or invalid paths into our vector.  Later, when we poll the data, if the file cannot be opened we record the value as 0.

PR #3715, opened 2020-07-15
CR-1066337 and pip update

AShivangi:
- CR-1066337: Even though SC flashing fails on U50LV, xrt reports success
- pip3 needs to be updated to install numpy


src/runtime_src/core/pcie/tools/xbmgmt/cmd_flash.cpp
```diff
@@ -307,7 +308,7 @@ static int updateShellAndSC(unsigned boardIdx, DSAInfo& candidate, bool& reboot)
     if (!same_dsa) {
         std::cout << "Updating shell on card[" << flasher.sGetDBDF() <<
             "]" << std::endl;
-        int ret = updateShell(boardIdx, "", candidate.file.c_str(),
+        ret = updateShell(boardIdx, "", candidate.file.c_str(),
```
maxzhen, line 311: I'm not sure about the intention of the code change, but ret will be overwritten here, so if updateSC failed, you can still return 0 if updateShell succeeded, right? is this something we want?

PR #3717, opened 2020-07-16
XRT-647 Add interrupt controller sub-device

mamin506:
Basic INTC sub-device, the idea is to create a sub-device to manage the ERT/CU interrupts. This could decoupling interrupt handling logic from KDS and be able to support CU to host interrupt.

I'd like to highlight devices.h change. I'm trying to split XOCL_RES_SCHEDULRE into two resources. One is XOCL_RES_INTC, which is for this new INTC sub-device. The other is for command queue (no include in this PR). Later on, the ERT sub-device would need the command queue resource.
@larry9523 , I would need your help to check if this change work on Versal (no runtime error).

The ERT/CU sub-device could register an interrupt handler through xocl_intc_request(xdev, id, handler, arg).
    The id is the interrupt ID, from 0 to 127.
    The handler is the same the standard interrupt handler function.
    The arg is the argument that will pass to the handler.

Supported:
    Legacy platforms (Tested on U.2 and U200)
Unsupported:
    SSv2 platform (waiting on U50 board)
    SSv3 platform (waiting on U50 board and new interrupt mapping support)
       - The INTC sub-device would support CU to host interrupts

The other commit is to update plram CU, which is a minor change. Please let me know if I should create an other PR.

### DIFF
--- a/src/runtime_src/core/common/drv/cu_plram.c
+++ b/src/runtime_src/core/common/drv/cu_plram.c
@@ -11,14 +11,14 @@
 
 extern int kds_echo;
 
-static int cu_plram_get_credit(void *core)
+static int cu_plram_alloc_credit(void *core)
 {
 	struct xrt_cu_plram *cu_plram = core;
 
 	return (cu_plram->credits) ? cu_plram->credits-- : 0;
 }
 
-static void cu_plram_put_credit(void *core, u32 count)
+static void cu_plram_free_credit(void *core, u32 count)
 {
 	struct xrt_cu_plram *cu_plram = core;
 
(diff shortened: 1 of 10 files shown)


src/runtime_src/core/common/drv/cu_plram.c
```diff
@@ -68,14 +75,15 @@ static void cu_plram_check(void *core, struct xcu_status *status)
...
-#endif
+out:
 	status->num_done = done_reg;
 	status->num_ready = done_reg;
 }
 
 static struct xcu_funcs xrt_cu_plram_funcs = {
 	.get_credit	= cu_plram_get_credit,
 	.put_credit	= cu_plram_put_credit,
+	.is_zero_credit = cu_plram_is_zero_credit,
```
maxzhen, line 86: maybe you can name these API as: alloc_credit, free_credit, peek_credit (is_zero_credit seems to be too restricted).

mamin506: done

PR #3739, opened 2020-07-22
Temporarily increase xclbin download timeout value on vck5000

larry9523:
Currently on VCK5000, it takes about 55 seconds to download a 43MB xclbin, which including
1. transfer xclbin from user pf->mgmtpf
2. transfer xclbin from host DDR -> device DDR through a 32K BRAM
3. PLM load PDI (in xclbin) 

So the current timeout value is not long enough here. We temporarily bump this value up on Versal. We will revisit this when we figure out the expected time consumption for loading xclbin on VCK5000.


larry9523:
> I am ok with this. However, if the download speed is stable on versal, could we come up with a formula instead of just double the timeout time?

Some investigation need to be done. It seems the bandwidth of the transfer is a bit low. I will come back with a more solid solution.

larry9523:
retest this please

larry9523:
retest this please

larry9523:
retest this please

PR #3747, opened 2020-07-25
Support Cardano and AIE Sim

larry9523:
This PR enhances 
1) XRT AIE/Graph API to work better with Cardano graph API
2) XRT AIE/Graph sources and  API works on AIE SIM


larry9523:
retest this please

PR #3700, opened 2020-07-12
NSPR-1387 : Xrt hbm support

saifuddin-xilinx:
Hi,
Creating this PR for review this HBM changes.
Please review the changes follows:

Userspace changes for XRT layer
User_PF Driver changes for HBM support
New changes support the following functionality:

User can specify a range of HBM banks in the v++ connectivity section, Example:
-sp CU_0.in1:HBM[0:3] . The user needs to specify bank-id information in his host application
as well using Extension Pointer.
If the user doesn't specify anything through the Extension pointer, Then XRT should automatically
detect the HBM banks using the connectivity section present in rtd file.

Further details : https://confluence.xilinx.com/pages/viewpage.action?pageId=205359608

PR #3752, opened 2020-07-27
NSRP-1387 Use GROUP_TOPOLOGY and GROUP_CONNECTIVITY

stsoe:
Use new sections for checking connectivity and acquiring group bank
index.  For legacy xclbins, or xclbins without grouping, the group
sections are same as MEM_TOPOLOGY and CONNECTIVITY.

This PR requires #3700


stsoe:
@chvamshi-xilinx I had to make an additional change to the emulation shims where they access mem topology sections directly from xclbin.   Basically, GROUP_TOPOLOGY should be used if available, otherwise just revert to MEM_TOPOLOGY.   Please review.  Thanks!

PR #3761, opened 2020-07-28
CR-1059337 add subdev icap_controller

rajkumar-xilinx:
Adding icap controller subdev to program & enable flash address
in icap controller ip registers to eliminate reboot after flashing image on card.
This feature will only be enabled on Samsung U2 flat shells.
Added new subdev called version control to read the shell stats and other
information which are needed for enabling features.

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
+++ b/src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
@@ -2125,6 +2125,7 @@ static int programXSpiDrv(std::FILE *mFlashDev, std::istream& mcsStream,
     unsigned int curAddr = UINT_MAX;
     unsigned int nextAddr = 0;
     unsigned int storeAddr = 0;
+    bool store = true;
     int ret;
 
     while (nextAddr != UINT_MAX) {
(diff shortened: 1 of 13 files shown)


rajkumar-xilinx:
> Why is version control moved to feature rom? Does it mean other platforms will use this IP?
> Instead of adding more stuff to feature rom, I would suggest to have a new subdev for version control.
> feature rom is obsoleted with SSv2 SSv3 platform. I think we should avoid adding new stuff to it if possible.

Hi @houlz0507 ,  I have added new subdev version control as per your suggestions. Please review again. Thank you

src/runtime_src/core/pcie/tools/xbmgmt/xspi.cpp
```diff
@@ -2132,6 +2133,8 @@ static int programXSpiDrv(std::FILE *mFlashDev, std::istream& mcsStream,
         if (ret)
             return ret;
         assert(nextAddr == UINT_MAX || pageOffset(nextAddr) == 0);
+        if (!storeAddr)
+            storeAddr = curAddr;
```
chvamshi-xilinx, line 2137: If MCS initial address is "0". Your expectation is to store "0" in storeAddr right?
This logic is not going to store "0". It always stores first non-zero address. You may need to use some boolean that initial address is already stored.

PR #3763, opened 2020-07-28
Install pybind11 as an XRT build time dependency

stsoe:


stsoe:
@NaAl-Xlnx @abalasa FYI, the xrtdeps.sh script is updated to pip3 install pybind11 which is needed in order to build the XRT python bindings.  Can you please make sure the docker images are updated accordingly, or maybe that is automatic?  Thank you!

PR #3764, opened 2020-07-28
add rw_shared switch to allow xclRegRead/xclRegWrite access shared CU

mamin506:
.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
@@ -1768,6 +1768,8 @@ struct exec_core {
 	bool		           configured;
 	bool		           stopped;
 	bool		           flush;
+	/* WORKAROUND: allow xclRegWrite/xclRegRead access shared CU */
+	bool			   rw_shared;
 
 	struct list_head           pending_cu_queue[MAX_CUS];
 	struct list_head           pending_ctrl_queue;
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
```diff
@@ -4613,7 +4618,8 @@ int cu_map_addr(struct platform_device *pdev, u32 cu_idx, void *drm_filp,
 		mutex_unlock(&xdev->dev_lock);
 		return -EINVAL;
 	}
-	if (ip_excl_holder(exec, cu_idx) == 0) {
+	/* WORKAROUND: If rw_shared is true, allow map shared CU */
+	if (!exec->rw_shared && ip_excl_holder(exec, cu_idx) == 0) {
```
maxzhen, line 4622: do we want to check if user is root?

stsoe:
Hi @mamin506 @maxzhen  It would be helpful with pull request comments that explain why this work-around is needed.  What's implemented here is easy to understand, but the why is missing.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
mamin506: We could easily add that once we have to.

mamin506:
Hi @stsoe , this special requirement is from a specific customer. And it would be a non document option.

PR #3768, opened 2020-07-29
NSPR-6216 XRT support on RHEL/CentOs 8.2

rbramand-xilinx:
> Made changes to build userspace and kernel code.
> Added kernel version of Rhel/Centos 8.2 in supported kernel version list.


rbramand-xilinx:
Branch name is making pipeline build to fail. So creating PR with new branch name

PR #3773, opened 2020-07-29
CR-1059337 add subdev icap_controller

rajkumar-xilinx:
Adding icap controller subdev to program & enable flash address
in icap controller ip registers to eliminate reboot after flashing image on card.
This feature will only be enabled on Samsung U2 flat shells.
Added new subdev called version control to read the shell stats and other
information which are needed for enabling features.


rajkumar-xilinx:
This PR is integrated from master branch (https://github.com/Xilinx/XRT/pull/3761)

PR #3782, opened 2020-07-30
CR-1070987 xrt 2.7.76 xbmgmt flash attempts to download SC even thoug…

xdavidz:
tested on U25:
```
Card [0000:66:00.0]
    Card type:          u25
    Flash type:         QSPI_PS
        xilinx_u25_gen3x8_xdma_base_1,[ID=0xc27f347aa668d2f9],[SC=7.8.0]
        xilinx_u25_gen3x8_xdma_base_1,[ID=0xc27f347aa668d2f9]
```
> Fixed version:
```
/proj/xsjhdstaff2/davidzha/XRT_zocl/build/Debug/runtime_src/core/pcie/tools/xbmgmt/xbmgmt flash --update --shell xilinx_u25_gen3x8_xdma_base_1 --card 0000:66:00.0
         Status: shell is up-to-date

Card(s) up-to-date and do not need to be flashed.
```

> without the fix:
```
/proj/xsjhdstaff2/davidzha/XRT_zocl/build/Debug/runtime_src/core/pcie/tools/xbmgmt/xbmgmt flash --update --shell xilinx_u25_gen3x8_xdma_base_1 --card 0000:66:00.0
         Status: SC needs updating
         Current SC: 7.8.0
         SC to be flashed:
Are you sure you wish to proceed? [y/n]: n
Action canceled.

```

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbmgmt/cmd_flash.cpp
+++ b/src/runtime_src/core/pcie/tools/xbmgmt/cmd_flash.cpp
@@ -251,7 +251,7 @@ static int resetShell(unsigned index, bool force)
 }
 
 /*
- * bmcVer (shown as [SC=version]) can be 3 status:
+ * bmcVer (shown as [SC=version]) can be the following status:
  *   1) regular SC version;
  *        example: [SC=4.1.7]
  *   2) INACTIVE;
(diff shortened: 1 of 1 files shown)


xdavidz:
> I don't think we should make this change. XRT should only ignore SC update when SC is either INACTIVE or FIXED. In other cases, XRT should always try to update SC and if the image can't be found, XRT should error out. This fix will hide the issue that the xsabin might not be generated correctly.


@xuhz please check with CMC team if they could provide status that can be marked as (FIXED).

We might not warn for all cases, as you pointed out here, if SC is valid but xclbin doesn't have SC, we should print some warning messages.

maxzhen:
After offline discussion with @xdavidz and @xuhz , we agreed that this PR should be closed and no change should be made in XRT.

PR #3784, opened 2020-07-30
Retrieve clock frequency to be used for trace from xclbin, adjust datatype used for actual device time stamps

IshitaGhosh:


IshitaGhosh:
retest this please

PR #3793, opened 2020-08-02
Add import / export xrt::bo

stsoe:
An buffer object can be exported from one process and imported by
another.

This change exposes xclExportBO and xclImportBO to the XRT Native API
level. It uses an opaque implementation define xclBufferExportHandle
type, which on Linux is a dmabuf file descriptor and is TBD on
Windows.

Also add size() method to xrt::bo.


stsoe:
retest this please after ping xcosda112

PR #3794, opened 2020-08-02
Fix GMIO interfaces

larry9523:
This PR made the following changes
1) Change data type of xrtGraphRun from uint32_t to int32_t to match Cardano context (0 for default and -1 for run forever)
2) Fix xrtSyncBOAIE interfaces
    a) Use device handle other than graph handle
    b) Use xrtBufferHandle other than integer BO handle
    c) Use GMIO name other than GMIO id
3) Move xrtSyncBO implementation from graph class to AIE class which is a device level class
4) Added non block GMIO and GMIO wait (xrtSycnBONB and xrtGMIOWait) for internal uses. Will revisit the non block syncbo later with AIO framework.
5) Small fixes for AIESIM

### DIFF
--- a/src/runtime_src/core/edge/user/aie/aie.cpp
+++ b/src/runtime_src/core/edge/user/aie/aie.cpp
@@ -41,7 +41,7 @@ get_bd_low_addr(uint64_t addr)
   return (addr & low_mask);
 }
 
-Aie::Aie(std::shared_ptr<xrt_core::device> device)
+Aie::Aie(const std::shared_ptr<xrt_core::device>& device)
 {
     /* TODO where are these number from */
     numRows = 8;
@@ -258,7 +258,7 @@ sync_bo_nb(uint64_t paddr, const char *gmioName, enum xclBOSyncDirection dir, si
 
 void
 Aie::
-wait_gmio(const char *gmioName)
+wait_gmio(const std::string& gmioName)
 {
   auto gmio = std::find_if(gmios.begin(), gmios.end(),
             [gmioName](gmio_type it) { return it.name.compare(gmioName) == 0; });
@@ -273,7 +273,7 @@ wait_gmio(const char *gmioName)
 
 void
 Aie::
-submit_sync_bo(uint64_t paddr, std::vector<gmio_type>::iterator gmio, enum xclBOSyncDirection dir, size_t size)
+submit_sync_bo(uint64_t paddr, std::vector<gmio_type>::iterator& gmio, enum xclBOSyncDirection dir, size_t size)
 {
   switch (dir) {
   case XCL_BO_SYNC_BO_GMIO_TO_AIE:
@@ -330,7 +330,7 @@ submit_sync_bo(uint64_t paddr, std::vector<gmio_type>::iterator gmio, enum xclBO
 
 void
 Aie::
-wait_sync_bo(ShimDMA *dmap, uint32_t chan, uint32_t timeout)
+wait_sync_bo(ShimDMA* const dmap, uint32_t chan, uint32_t timeout)
 {
   while ((XAieDma_ShimWaitDone(&(dmap->handle), chan, timeout) != XAIEGBL_NOC_DMASTA_STA_IDLE));
 
--- a/src/runtime_src/core/edge/user/aie/aie.cpp
+++ b/src/runtime_src/core/edge/user/aie/aie.cpp
@@ -18,12 +18,28 @@
 
 #include "aie.h"
 #include "core/common/error.h"
+#ifndef __AIESIM__
 #include "core/common/message.h"
+#endif
 
 #include <iostream>
 #include <cerrno>
 
 namespace zynqaie {
+
+static inline uint64_t
+get_bd_high_addr(uint64_t addr)
+{
+  constexpr uint64_t hi_mask = 0xFFFF00000000L;
+  return ((addr & hi_mask) >> 32);
+}
+
+static inline uint64_t
+get_bd_low_addr(uint64_t addr)
+{
+  constexpr uint32_t low_mask = 0xFFFFFFFFL;
+  return (addr & low_mask);
+}
 
 Aie::Aie(std::shared_ptr<xrt_core::device> device)
 {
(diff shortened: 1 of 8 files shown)


src/runtime_src/core/edge/user/aie/aie.cpp
```diff
@@ -18,13 +18,29 @@
...
+}
+
+static inline uint64_t
+get_bd_low_addr(uint64_t addr)
+{
+  constexpr uint32_t low_mask = 0xFFFFFFFFL;
+  return (addr & low_mask);
+}
+
 Aie::Aie(std::shared_ptr<xrt_core::device> device)
```
stsoe, line 44: Prefer `const std::shared_ptr<xrt_core::device>& device`

src/runtime_src/core/edge/user/aie/aie.h
```diff
@@ -77,8 +87,16 @@ class Aie {
     XAieGbl_Config *aieConfigPtr; // AIE configuration pointer
     XAieGbl aieInst;              // AIE global instance
     XAieGbl_HwCfg aieConfig;      // AIE configuration pointer
+
+    void
+    submit_sync_bo(uint64_t paddr, std::vector<gmio_type>::iterator gmio, enum xclBOSyncDirection dir, size_t size);
```
stsoe, line 92: I am almost positive you don't mean to pass the vector by value.

src/runtime_src/core/edge/user/aie/aie.h
```diff
@@ -66,6 +67,15 @@ class Aie {
...
     XAieGbl *getAieInst();
 
+    void
+    sync_bo(uint64_t paddr, const char *dmaID, enum xclBOSyncDirection dir, size_t size);
+
+    void
+    sync_bo_nb(uint64_t paddr, const char *gmioName, enum xclBOSyncDirection dir, size_t size);
+
+    void
+    wait_gmio(const char *gmioName);
```
stsoe, line 77: Maybe use std::string in place of const char* for all C++ functions.

src/runtime_src/core/edge/user/aie/aie.h
```diff
@@ -77,8 +87,16 @@ class Aie {
     XAieGbl_Config *aieConfigPtr; // AIE configuration pointer
     XAieGbl aieInst;              // AIE global instance
     XAieGbl_HwCfg aieConfig;      // AIE configuration pointer
+
+    void
+    submit_sync_bo(uint64_t paddr, std::vector<gmio_type>::iterator gmio, enum xclBOSyncDirection dir, size_t size);
+
+    /* Wait for all the BD trasters for a given channel */
+    void
+    wait_sync_bo(ShimDMA *dmap, uint32_t chan, uint32_t timeout);
```
stsoe, line 96: Does this function change dmap?   If not it should be const ShimDMA*

src/runtime_src/core/edge/user/aie/graph.cpp
```diff
@@ -36,25 +36,16 @@ extern "C"
 }
 
 #ifdef __AIESIM__
-static zynqaie::Aie s_aie(xrt_core::get_userpf_device(0));
+static zynqaie::Aie* sp_aie = nullptr;
+zynqaie::Aie* getAieArray() {
+  if (sp_aie == nullptr)
+    sp_aie = new zynqaie::Aie(xrt_core::get_userpf_device(0));
```
stsoe, line 42: Where is this deleted?  If this like a singleton you may want to do something like this:

```
auto
getAieArray()
{
   static zynqaie::Aie s_aie(xrt_core::get_user_pf_device(0));
   return &s_aie;
}
```

src/runtime_src/core/edge/user/aie/graph.cpp
```diff
@@ -1001,10 +952,10 @@ xrtGraphReadRTP(xrtGraphHandle ghdl, const char *port, char *buffer, size_t size
 }
 
 int
-xrtSyncBOAIE(xrtGraphHandle ghdl, unsigned int bo, const char *dmaID, enum xclBOSyncDirection dir, size_t size, size_t offset)
+xrtSyncBOAIE(xclDeviceHandle handle, xrtBufferHandle bohdl, const char *gmioName, enum xclBOSyncDirection dir, size_t size, size_t offset)
 {
   try {
-    api::xrtSyncBOAIE(ghdl, bo, dmaID, dir, size, offset);
+    api::xrtSyncBOAIE(handle, bohdl, gmioName, dir, size, offset);
```
stsoe, line 958: This is the C-API implementation receiving const char*, but I think api::xrtSyncBOAIE or maybe what it calls should use std::string.

src/runtime_src/core/edge/user/aie/aie.h
stsoe: Sorry misses that you are passing an iterator.

src/runtime_src/core/edge/user/aie/aie.h
larry9523: The content of dmap pointing to will change. I am using 
wait_sync_bo(ShimDMA* const dmap, uint32_t chan, uint32_t timeout);

src/runtime_src/core/edge/user/aie/aie.h

src/runtime_src/core/edge/user/aie/aie.cpp

src/runtime_src/core/edge/user/aie/graph.cpp
larry9523: done. Thanks for the suggestion

src/runtime_src/core/edge/user/aie/graph.cpp
larry9523: done

src/runtime_src/core/edge/user/aie/aie.h
```diff
@@ -77,8 +87,16 @@ class Aie {
     XAieGbl_Config *aieConfigPtr; // AIE configuration pointer
     XAieGbl aieInst;              // AIE global instance
     XAieGbl_HwCfg aieConfig;      // AIE configuration pointer
+
+    void
+    submit_sync_bo(uint64_t paddr, std::vector<gmio_type>::iterator& gmio, enum xclBOSyncDirection dir, size_t size);
+
+    /* Wait for all the BD trasters for a given channel */
```
keryell, line 94: I am so happy to see a comment! \o/
What is "trasters"? Some secret piece of hardware?
Please feel free to add a lot of comments.
I love open-source code and even more READABLE open-source code... :-)

larry9523: Thanks for catching this.

PR #3797, opened 2020-08-03
remove numpy dependency

AShivangi:
Use ctypes arrays instead of numpy. This adds 10 seconds to the bandwidth test.


tests/python/23_bandwidth/23_bandwidth.py
```diff
@@ -61,8 +60,8 @@ def getInputOutputBuffer(devhdl, krnlhdl, argno, isInput):
         lst = [i%256 for i in range(globalbuffersize)]
     else:
         lst = [0 for i in range(globalbuffersize)]
-    hostbuf = np.array(lst).astype(np.uint8)
-    assert hostbuf.size == globalbuffersize
+    hostbuf = (ctypes.c_int * len(lst))(*lst)
```
rozumx, line 63: If I read this code correctly, the original code created an array 1 byte values.   The replacement code creates an array of 8 byte values (assuming integers are 64-bits in length).  

_**Could you validated that the hostbuf arrays are identical before and after these changes.**_

PR #3799, opened 2020-08-04
qspi flash factory reset support

xuhz:


src/runtime_src/core/pcie/tools/xbmgmt/xqspips.cpp
```diff
@@ -35,6 +35,7 @@
 
 #define SAVE_FILE                   0
 #define FLASH_BASE                  0x040000
+#define GOLDEN_BASE                 0x06000000
```
mamin506, line 38: The golden start from 96MB. But does this the case on U30?

src/runtime_src/core/pcie/tools/xbmgmt/xqspips.cpp
```diff
@@ -375,6 +376,37 @@ bool XQSPIPS_Flasher::waitTxEmpty()
...
+        std::cout << "[ERROR]: Could not get Flash ID" << std::endl;
+        exit(-EOPNOTSUPP);
+    }
+
+    /* Use 4 bytes address mode */
+    enterOrExitFourBytesMode(ENTER_4B);
+
+    // Sectoer size is defined by SECTOR_SIZE
+    std::cout << "Factory resetting " << std::flush;
+    eraseSector(0, GOLDEN_BASE - 3 * SECTOR_SIZE);
```
mamin506, line 404: The eraseSector() would erase extra sectors. It is okay when there is only one image. We would better roundup the byteCount to nearest SECTOR_SIZE in eraseSector().

src/runtime_src/core/pcie/tools/xbmgmt/xqspips.cpp
```diff
@@ -390,6 +422,10 @@ int XQSPIPS_Flasher::xclUpgradeFirmware(std::istream& binStream)
     binStream.seekg(0, binStream.beg);
 
     std::cout << "INFO: ***BOOT.BIN has " << total_size << " bytes" << std::endl;
+    if (total_size >= GOLDEN_BASE) {
+        std::cout << "[ERROR]: BOOT.BIN is too large. Flashing will overwrite golden." << std::endl;
+        exit(-EINVAL);
+    }
```
mamin506, line 428: Please clearly say that the BOOT.BIN should not bigger than 96MB.

BTW, I already let U30 platform developer know this 96MB change. They think it is okay. The typical size of BOOT.BIN is 96MB.

src/runtime_src/core/pcie/tools/xbmgmt/xqspips.cpp
xuhz: yes, eraseSector so far erases more than required. This will have issues when we have bin file ~96M size and we also try to hold gloden at 96M.
will fix eraseSector() to erase accurate number of sectors.

src/runtime_src/core/pcie/tools/xbmgmt/xqspips.cpp
xuhz: changed.

src/runtime_src/core/pcie/tools/xbmgmt/xqspips.cpp
xuhz: u30 will follow the same to u25

PR #3800, opened 2020-08-04
Prevent infinite loop when the emulator dies

Ralender:
Currently when the software emulator crashes, XRT sometimes goes into an infinite loop because it is still waiting on the simulator to write data to a socket.

this patch fixes this by detecting the the emulator died via a SIGCHILD with a CLD_KILLED or CLD_DUMPED code

the noop changes is my editor correcting a CRLF line ending to LF.

### DIFF
--- a/src/runtime_src/core/edge/sw_em/generic_pcie_hal2/shim.cxx
+++ b/src/runtime_src/core/edge/sw_em/generic_pcie_hal2/shim.cxx
@@ -167,43 +167,36 @@ namespace xclcpuemhal2 {
  
   static void sigHandler(int sn, siginfo_t *si, void *sc)
   {
-    switch(sn) {
-      case SIGSEGV:
-        {
-          saveDeviceProcessOutputs();
-          kill(0,SIGSEGV);
-          exit(1);
-          break;
-        }
-      case SIGFPE :
-        {
-          saveDeviceProcessOutputs();
-          kill(0,SIGTERM);
-          exit(1);
-          break;
-        }
-      case SIGABRT:
-        {
-          saveDeviceProcessOutputs();
-          kill(0,SIGABRT);
-          exit(1);
-          break;
-        }
+    switch (sn) {
+    case SIGSEGV:
+      saveDeviceProcessOutputs();
+      kill(0, SIGSEGV);
+      exit(1);
+      break;
+    case SIGFPE:
+      saveDeviceProcessOutputs();
+      kill(0, SIGTERM);
+      exit(1);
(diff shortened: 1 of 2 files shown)


Ralender:
formated the sigHandler function
added comment.
fixed typo

chvamshi-xilinx:
ok to test

src/runtime_src/core/edge/sw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -164,46 +164,43 @@ namespace xclcpuemhal2 {
...
+      saveDeviceProcessOutputs();
+      std::cerr << "Software emulation of compute unit(s) exited unexpectedly"
+                << std::endl;
+      kill(0, SIGTERM);
+      exit(1);
+      break;
+    }
+    default: {
+      break;
+    }
```
keryell, line 203: I see. There was a problem in the original code with a random mix of spaces and tabs...
Curious use of all these useless `{}` almost everywhere. Specially strange with the `default:` case.
@stsoe is this the local coding style?

stsoe: No.  The coding style is early exit and no braces for single statement.

Ralender:
Where are we on this ?

src/runtime_src/core/edge/sw_em/generic_pcie_hal2/shim.cxx
stsoe: @Ralender .  Maybe remove all the unnecessary curly braces?

I don't quite understand the exit(1), are you sure the host application should be killed?  Isn't it enough to kill the emulation process?

keryell: All this seems to have been here before but I agree that it does not prevent us to fix the ugliness.
Perhaps you could start by preventing any non commented code in your team... :-)

Ralender: Removed the Braces.

About the exit(1) optimally we would want to properly return an error to he host code when the emulator dies.
However in specific situation i experienced where the emulator died proper error propagation is quite hard because it happens deep in nested macros that are used i various contexts.
I still think that killing the Host is better then staying stuck in an infinite loop.

Ralender:
I added the signoff

dayeh-xilinx:
retest this please

PR #3806, opened 2020-08-04
Drop numpy usage from XRT and tests

sonals:


chvamshi-xilinx:
It will be better if we cherry pick this change to 2020.1_PU1 also. @sonals , any thoughts.

chvamshi-xilinx:
Looks like @maxzhen already filed a PR for 2020.1_PU1. Thanks.

PR #3808, opened 2020-08-05
NSPR-7298 HW-EMU changes in shim.cxx for uniq name to Device to xrt unix socket

ukyatham-xilinx:
… xrt socket


chvamshi-xilinx:
ok to test

PR #3809, opened 2020-08-05
NSPR-7327 NSPR-7327 XRT package split into dev and deploy

stsoe:
Removed package dependencies on development packages.

Use RPATH with CMake to allow libraries to find XRT dependencies
relative to location.  This in turn enables CPack to discover package
dependencies at package time, greatly simplifying explicit package
dependencies.

By default only one combined XRT package is created for both run-time
and development. To enable separate packages adjust src/CMakeLists.txt
to specify XRT_DEV_COMPONENT.


stsoe:
@jvillarre Just FYI because I changed one xdp makefile.  No functional changes to how xdp is installed at this point.  This PR is just set up for splitting XRT package to development and deployment, plus clean up of package dependencies.

PR #3811, opened 2020-08-05
Flip the default value of Runtime.xrt_bo to true

stsoe:
Set to false if using old style xclBufferHandles with kernel C-APIs.
This is necessary for the C-style variadic args to infer the type.  By
default it is assume that host code use XRT Native APIs for BOs as
well as kernel, in which case the default value is required.


stsoe:
retest this please  previous build is not available

PR #3817, opened 2020-08-06
CR-1071496 Addition of PCIe IDs for U26Z

juliank-xilinx:
Addition of PCIe IDs for U26Z.
Smoke-tested by rebuilding and running through xbutil validate on known-good U250 card.


maxzhen:
ok to test

PR #3819, opened 2020-08-07
qdma4: fix error interrupt enable mask for MM only and ST updates

karenx-xilinx:
- Check if ST enabled when setting up error mask
- allow bigger c2h freelist buffer allocation.
- misc. update ported from ip eqdma driver


houlz0507:
retest this please

PR #3836, opened 2020-08-10
Coverity

maxzhen:


maxzhen:
retest this please due to Jenkins node went down during previous pipeline test.

PR #3846, opened 2020-08-11
Cherry-pick :Remove fpga-mgr dependency for inserting xclmgmt.ko (#3844)

rbramand-xilinx:
> xclmgmt.ko inserting depends on symbols exported by fpga-mgr.c file
> CentOs and RHEL does not support this by default, so removing dependency in them


chvamshi-xilinx:
FPGA Mnager integration is support limited to Ubuntu as this module gets added only in Ubuntu by default.

chvamshi-xilinx:
Merging this change as this is suggested by @sonals himself. I would like to have this change in today's daily_latest.

PR #3848, opened 2020-08-11
xbmgmt2 cleanup

AShivangi:
- cleaned up property trees: removed puts using "x.y.z"
- moved from flash_address to flash_bar_offset because that's how it is originally defined on linux
- added scan and flash support for/from golden image (tested on golden -> 1RP)
- got rid of variables starting with "_"


src/runtime_src/core/tools/xbmgmt2/ReportPlatform.cpp
```diff
@@ -60,78 +60,97 @@ same_sc(const std::string& sc, const DSAInfo& installed)
...
 }
 
-void 
-ReportPlatform::getPropertyTree20202( const xrt_core::device * _pDevice,
-                                      boost::property_tree::ptree &_pt) const
+/* 
+ * standard format for text output
+ */
+static std::string 
+basic_format(const std::string& label, const std::string& value)
```
rozumx, line 67: Is this method really needed? If not, could you please refactor it out.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -233,7 +233,6 @@ initialize_query_table()
   emplace_sysfs_get<query::xclbin_uuid>                 ("", "xclbinuuid");
   emplace_sysfs_get<query::mem_topology_raw>            ("icap", "mem_topology");
   emplace_sysfs_get<query::ip_layout_raw>               ("icap", "ip_layout");
-  emplace_sysfs_get<query::clock_freq_topology_raw>     ("icap", "clock_freq_topology");
```
stsoe, line 236: @AShivangi Why was this removed?

AShivangi: I must have removed it when I was resolving merge conflicts. You can add it back.

stsoe:
ok. thanks!

PR #3861, opened 2020-08-13
New IOCTLs for Linux In-kernel Interface

akum-xilinx:
These new IOCTLs can be only be invoked from the other Linux modules within kernel . The access is via the DRM subsystem IOCTL interface. The access via DRM subsystem already exists for existing IOCTLs. These new IOCTLs provide meaningful usability of the XRT interfaces from within the linux kernel.

DRM_IOCTL_XOCL_KINFO_BO - To get kernel virtual address of ExecBuf and P2P BOs + other INFO BO details
DRM_IOCTL_XOCL_MAP_KERN_MEM - To map Linux kernel buffers and SGLs to Device BOs
DRM_IOCTL_XOCL_EXECBUF_CB - To get a callback once an execbuf been completed
DRM_IOCTL_XOCL_SYNC_BO_CB - To get a callback once a DMA has been completed


maxzhen:
ok to test

houlz0507:
@karenx-xilinx Could you help to review the libxdma change?

src/runtime_src/core/pcie/driver/linux/xocl/lib/libxdma.c
```diff
@@ -941,13 +941,15 @@ static int process_completions(struct xdma_engine *engine,
 
 		if (req->sw_desc_cnt == req->desc_completed) {
 			list_del(&req->entry);
+			spin_unlock(&engine->req_list_lock);
 			if (req->cb && req->cb->io_done) {
 				struct xdma_io_cb *cb = req->cb;
 
-				cb->io_done((unsigned long)cb->private, 0);
+				cb->io_done((unsigned long)cb->private, req->done);
```
karenx-xilinx, line 948: the 2nd parameter in io_done() is error code, please keep it. If you want additional information on the done bytes, please add a another parameter for it.

akum-xilinx: Hi Karen, I am mostly fine with adding another parameter to io_done function. Before that can we also consider 2 other options:
1. Use the count field in xdma_io_cb to load the done bytes count. It does not seem to be used now. 
2. If the count field is to denote requested len, have a new field "done_bytes" for completed transfers.
3. Rename count to "done_bytes" to ensure that this field will contain completed transfers.

karenx-xilinx: Yes, let's do #2. Thanks.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xdma.c
```diff
@@ -92,6 +92,110 @@ static ssize_t xdma_migrate_bo(struct platform_device *pdev,
...
+	struct xdma_async_context *async_ctx = (struct xdma_async_context *)data;
+	//pr_info("%s: async_ctx %llx, ", __func__, (u64)async_ctx);
+
+	if (!err) {
+		async_ctx->xdma->channel_usage
+			[async_ctx->dir][async_ctx->channel] += 
+						async_ctx->iocb->done_bytes;
+	}
+	
+	async_ctx->callback_fn(async_ctx->callback_data, err);
```
pranjalv-xilinx, line 115: It will be good to have a check for callback func like if (async_ctx->callback_fn). Sometimes clients for this async method might not want completions notifications for each transfers.

src/runtime_src/core/pcie/driver/linux/include/xocl_ioctl.h
```diff
@@ -157,6 +157,15 @@ enum drm_xocl_ops {
 	DRM_XOCL_ALLOC_CMA,
 	/* Free allocated CMA chunk through userpf*/
 	DRM_XOCL_FREE_CMA,
+
+	/* The foloowing IOCTLs can only be called from linux kernel space
```
larry9523, line 161: following?

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_bo.c
```diff
@@ -1329,3 +1338,311 @@ int xocl_usage_stat_ioctl(struct drm_device *dev, void *data,
...
+        int i;
+        int ret=0;
+        unsigned int page_count;
+        struct scatterlist *sg;
+        const struct drm_xocl_map_kern_mem *args = data;
+
+        struct drm_xocl_bo *xobj;
+        struct drm_gem_object *gem_obj;
+
+	/* This ioctl should only be called to map to kernel memory */
```
larry9523, line 1535: the indent seems wrong

src/runtime_src/core/pcie/driver/linux/xocl/xocl_drv.h
```diff
@@ -595,6 +598,9 @@ struct xocl_dma_funcs {
 #define	xocl_migrate_bo(xdev, sgt, to_dev, paddr, chan, len)	\
 	(DMA_CB(xdev, migrate_bo) ? DMA_OPS(xdev)->migrate_bo(DMA_DEV(xdev), \
 	sgt, to_dev, paddr, chan, len) : 0)
+#define	xocl_async_migrate_bo(xdev, sgt, to_dev, paddr, chan, len, cb_fn, ctx_ptr)	\
+	(DMA_CB(xdev, migrate_bo) ? DMA_OPS(xdev)->async_migrate_bo(DMA_DEV(xdev), \
```
pranjalv-xilinx, line 602: DMA_CB() macro should be checking for async_migrate_bo and not migrate_bo

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_kds.c
```diff
@@ -315,6 +331,26 @@ static int xocl_command_ioctl(struct xocl_dev *xdev, void *data,
 	xcmd->cb.notify_host = notify_execbuf;
 	xcmd->gem_obj = obj;
 
+	if (in_kernel) {
+		struct drm_xocl_execbuf_cb *args_cb =
+					(struct drm_xocl_execbuf_cb *)data;
```
mamin506, line 336: This is not a must resolve thing.
To make this work, the drm_xocl_execbuf_cb must be an extend of drm_xocl_execbuf. For the short term and for internal quick fix, this is okay to me.
But, ideally, we should not use two structs.
Maybe we can accept this code but we definitely need a little bit refine in the future.

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_kds.c
```diff
@@ -315,6 +331,26 @@ static int xocl_command_ioctl(struct xocl_dev *xdev, void *data,
...
+
+		if (args_cb->cb_func) {
+			xcmd->inkern_cb = kzalloc(sizeof(struct in_kernel_cb),
+								GFP_KERNEL);
+			if (!xcmd->inkern_cb) {
+				kfree(xcmd);
```
mamin506, line 342: xcmd->cb.free(xcmd);

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
```diff
@@ -407,7 +417,7 @@ static struct drm_driver mm_drm_driver = {
 	.gem_vm_ops			= &xocl_vm_ops,
 
 	.ioctls				= xocl_ioctls,
-	.num_ioctls			= ARRAY_SIZE(xocl_ioctls),
+	.num_ioctls			= (ARRAY_SIZE(xocl_ioctls)-NUM_KERNEL_IOCTLS),
```
mamin506, line 420: What would happen if user call DRM_XOCL_EXECBUF_CB just in accident? Hope linux kernel is strong enough to handle this without crash. Have we tried it?

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
```diff
@@ -368,6 +368,16 @@ static const struct drm_ioctl_desc xocl_ioctls[] = {
...
+#define NUM_KERNEL_IOCTLS 4
+	DRM_IOCTL_DEF_DRV(XOCL_KINFO_BO, xocl_kinfo_bo_ioctl,
+			  DRM_AUTH|DRM_UNLOCKED|DRM_RENDER_ALLOW),
+	DRM_IOCTL_DEF_DRV(XOCL_MAP_KERN_MEM, xocl_map_kern_mem_ioctl,
+			  DRM_AUTH|DRM_UNLOCKED|DRM_RENDER_ALLOW),
+	DRM_IOCTL_DEF_DRV(XOCL_EXECBUF_CB, xocl_execbuf_callback_ioctl,
+			  DRM_AUTH|DRM_UNLOCKED|DRM_RENDER_ALLOW),
+	DRM_IOCTL_DEF_DRV(XOCL_SYNC_BO_CB, xocl_sync_bo_callback_ioctl,
+			  DRM_AUTH|DRM_UNLOCKED|DRM_RENDER_ALLOW),
+
```
larry9523, line 380: It is probably a good idea to add some comments, that these kernel ioctls must be put at the end of the array. Otherwise, there might be leak to user space for these kernels which could be a security issue.

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
akum-xilinx: The DRM subsystem in linux kernel will prevent calling of these IOCTLS. Following is a small snippet of code from drm_ioctl function (linux/v4.15/source/drivers/gpu/drm/drm_ioctl.c)
_if (is_driver_ioctl) {
		/* driver ioctl */
		if (nr - DRM_COMMAND_BASE >= dev->driver->num_ioctls)
			goto err_i1;
		ioctl = &dev->driver->ioctls[nr - DRM_COMMAND_BASE]; 
}_
These IOCTLS if called through drm_ioctl function will fail(err_i1 above). Userspace IOCTLs are invoked through drm_ioctl. These new IOCTLs can be only called using drm_ioctl_kernel, which is not exposed to userspace.  Kernel space modules need to call drm_ioctl_kernel. XSS does. Since @larry9523 also asked for this explanation, I am tagging him here.

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
akum-xilinx: I took care of it.

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_kds.c
akum-xilinx: I had thought of it. But the problem I see is the callback function and data pointer may look out of place in user-space struct (drm_xocl_execbuf).  Can we think over the options and make a subsequent check-in?

karenx-xilinx (approved):

PR #3868, opened 2020-08-14
xbutil2 advanced --p2p 

stsoe:
Support p2p validate/enable/disable

### DIFF
--- a/src/runtime_src/core/common/error.h
+++ b/src/runtime_src/core/common/error.h
@@ -58,6 +58,14 @@ public:
   {}
 };
 
+class generic_error : public std::system_error
+{
+public:
+  generic_error(int ec, const std::string& what = "")
+    : std::system_error(ec, std::generic_category(), what)
+  {}
+};
+
 XRT_CORE_COMMON_EXPORT
 void
 send_exception_message(const char* msg, const char* tag="XRT");
--- a/src/runtime_src/core/pcie/windows/shim.cpp
+++ b/src/runtime_src/core/pcie/windows/shim.cpp
@@ -1556,7 +1556,7 @@ xclUnmgdPwrite(xclDeviceHandle handle, unsigned int flags, const void *buf, size
   xrt_core::message::
     send(xrt_core::message::severity_level::XRT_DEBUG, "XRT", "xclUnmgdPwrite()");
   auto shim = get_shim_object(handle);
-  return shim->unmgd_pwrite(flags, buf, count, offset);
+  return shim->unmgd_pwrite(flags, buf, count, offset) ? 0 : 1;
 }
 
 ssize_t
@@ -1565,7 +1565,7 @@ xclUnmgdPread(xclDeviceHandle handle, unsigned int flags, void *buf, size_t coun
   xrt_core::message::
     send(xrt_core::message::severity_level::XRT_DEBUG, "XRT", "xclUnmgdPread()");
   auto shim = get_shim_object(handle);
-  return shim->unmgd_pread(flags, buf, count, offset);
+  return shim->unmgd_pread(flags, buf, count, offset) ? 0 : 1;
 }
 
 size_t xclWriteBO(xclDeviceHandle handle, xclBufferHandle boHandle, const void *src, size_t size, size_t seek)
@@ -1697,8 +1697,15 @@ xclReadTraceData(xclDeviceHandle handle, void* traceBuf, uint32_t traceBufSz,
   return size;
 }
 
-int xclGetSubdevPath(xclDeviceHandle handle,  const char* subdev,
-                        uint32_t idx, char* path, size_t size)
+int
+xclGetSubdevPath(xclDeviceHandle handle,  const char* subdev,
+                 uint32_t idx, char* path, size_t size)
 {
   return 0;
+}
+
+int
+xclP2pEnable(xclDeviceHandle handle, bool enable, bool force)
+{
+  return 1; // -ENOSYS;
 }
--- a/src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
+++ b/src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
@@ -37,6 +37,10 @@ namespace po = boost::program_options;
 #include <iostream>
 #include <cctype>
 
+#ifdef _WIN32
+# pragma warning(disable : 4245)
+#endif
+
 namespace {
 
 enum class action_type { enable, disable, validate };
@@ -45,7 +49,7 @@ action_type
 string2action(std::string str)
 {
   std::transform(str.begin(), str.end(), str.begin(),
-                 [](auto c) { return std::tolower(c); });
+                 [](auto c) { return static_cast<char>(std::tolower(c)); });
 
   if (str == "enable")
     return action_type::enable;
@@ -68,7 +72,7 @@ p2p_config(xrt_core::device* device)
     throw xrt_core::system_error(ENOTSUP, "p2p is not supported");
   }
 }
-  
+
 static void
 p2p_enabled_or_error(xrt_core::device* device)
 {
@@ -146,11 +150,10 @@ chunk(xrt_core::device* device, char* boptr, uint64_t dev_addr, uint64_t size)
   try {
     device->unmgd_pwrite(buf, size, dev_addr);
   }
-  catch (const std::exception& ex) {
+  catch (const std::exception&) {
     std::ostringstream ostr;
-    ostr << "Error (" << strerror (errno) << ") writing 0x"
-         << std::hex << size << " bytes to 0x" << std::hex << dev_addr
-         << std::dec;
+    ostr << "Error writing 0x" << std::hex << size
+         << " bytes to 0x" << dev_addr << std::dec;
     throw xrt_core::system_error(EIO, ostr.str());
   }
   cmp_with_stride(boptr, boptr+size, patternA);
@@ -159,11 +162,10 @@ chunk(xrt_core::device* device, char* boptr, uint64_t dev_addr, uint64_t size)
   try {
     device->unmgd_pread(buf, size, dev_addr);
   }
-  catch (const std::exception& ex) {
+  catch (const std::exception&) {
     std::ostringstream ostr;
-    ostr << "Error (" << strerror (errno) << ") reading 0x"
-         << std::hex << size << " bytes from 0x" << std::hex << dev_addr
-         << std::dec;
+    ostr << "Error reading 0x" << std::hex << size
+         << " bytes from 0x" << dev_addr << std::dec;
     throw xrt_core::system_error(EIO, ostr.str());
   }
   cmp_with_stride(buf, buf+size, patternB);
@@ -305,4 +307,3 @@ OO_P2P::execute(const SubCmdOptions& _options) const
   for (auto& device : XBU::collect_devices(m_device, true))
     p2p(device.get(), string2action(m_action), false);
 }
-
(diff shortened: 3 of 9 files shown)


src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
```diff
@@ -17,12 +17,244 @@
...
+p2p_enabled_or_error(xrt_core::device* device)
+{
+  auto cfg = p2p_config(device);
+
+  long long bar = -1;
+  long long rbar = -1;
+  long long remap = -1;
+  long long exp_bar = -1;
+
+  for (auto& str : cfg) {
```
rozumx, line 86: Looking at the code it is evident that is is iterating over a collection (most likely a vector) of strings, where the string format is <key>:<value>.  

Could you add a comment with example "sample" string that is being parsed to assist the next developer in understanding what is being done.

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
```diff
@@ -17,12 +17,244 @@
...
+  if (bar == exp_bar)
+    return;
+
+  throw xrt_core::system_error(ENOTSUP,"p2p is not supported");
+}
+
+namespace p2ptest {
+
+static void
+fill_with_stride(char* begin, const char* end, char pattern)
```
rozumx, line 115: Could you change the variable "pattern" to something like "fill_byte" or something like that.  I ask for pattern is misleading and the name suggests it could be more then just a fill_byte.

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
```diff
@@ -17,12 +17,244 @@
...
+
+static void
+cmp_with_stride(const char* begin, const char* end, char pattern)
+{
+  auto stride = xrt_core::getpagesize();
+  if (std::distance(begin, end) % stride)
+    throw xrt_core::system_error(EINVAL, "Range not an increment of stride: " + std::to_string(stride));
+  for (auto itr = begin; itr != end; itr += stride) {
+    if ((*itr) != pattern) {
+      std::ostringstream ostr;
```
rozumx, line 132: An alternate way to produce this error message could be:
const string errMsg = boost::format("Error in p2p comparison, expected '0x%x' got '0x%x'" % pattern % (*itr)).str();

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
```diff
@@ -17,12 +17,244 @@
...
+  auto mem = xrt_core::aligned_alloc(xrt_core::getpagesize(), size);
+  auto buf = static_cast<char*>(mem.get());
+
+  fill_with_stride(buf, buf+size, patternA);
+  try {
+    device->unmgd_pwrite(buf, size, dev_addr);
+  }
+  catch (const std::exception&) {
+    std::ostringstream ostr;
+    ostr << "Error writing 0x" << std::hex << size
```
rozumx, line 155: Similar comment as above, consider using boost::format to format the error message.

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
```diff
@@ -17,12 +17,244 @@
...
+         << " bytes from 0x" << dev_addr << std::dec;
+    throw xrt_core::system_error(EIO, ostr.str());
+  }
+  cmp_with_stride(buf, buf+size, patternB);
+}
+
+static void
+bank(xrt_core::device* device, int memidx, uint64_t addr, uint64_t size)
+{
+  const size_t chunk_size = 16 * 1024 * 1024;
```
rozumx, line 177: Can the magic number "16" be assigned to a `static const size_t `variable, especially since it is used more then once.

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
```diff
@@ -17,12 +17,244 @@
...
+  auto mt_raw = xrt_core::device_query<xrt_core::query::mem_topology_raw>(device);
+  auto mt = reinterpret_cast<const mem_topology*>(mt_raw.data());
+  if (!mt)
+    throw xrt_core::system_error(EINVAL, "mem_topology is invalid, cannot validate p2p");
+
+  // support memory types
+  // p2p is not supported for DDR on u280
+  auto vbnv = xrt_core::device_query<xrt_core::query::rom_vbnv>(device);
+  std::vector<std::string> supported = { "HBM", "bank" };
+  if (vbnv.find("_U280_") == std::string::npos)
```
rozumx, line 222: So sad....there needs to be a better way, but unfortunately that doesn't look to be the case.

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
```diff
@@ -17,12 +17,244 @@
...
+    supported.push_back("DDR");
+
+  for (int32_t i = 0; i < mt->m_count; ++i) {
+    if (!mt->m_mem_data[i].m_used)
+      continue;
+
+    auto nm = std::string(reinterpret_cast<const char *>(mt->m_mem_data[i].m_tag));
```
rozumx, line 229: Note: The "tag" strings do not necessarily match that of the m_type.  By chance, can the memory types be used instead of the tag strings.

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
AShivangi: I have this function as check_p2p_config in XBUtilities

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
```diff
@@ -17,12 +17,244 @@
...
+    throw xrt_core::system_error(EIO,"Please WARM reboot to enable p2p");
+  if (remap > 0 && remap != bar)
+    throw xrt_core::system_error(EIO,"p2p remapper is not set correctly");
+  if (bar == exp_bar)
+    return;
+
+  throw xrt_core::system_error(ENOTSUP,"p2p is not supported");
+}
+
+namespace p2ptest {
```
AShivangi, line 112: I added p2p test in SubCmdValidate

stsoe: Oh, I didn't know about these., thanks for pointing that out.  Unfortunately they are not exception safe, I am adamant about exceptions everywhere and no return codes.    We will leave as is for now, but clearly we will have to revisit this.  I am  not clear on how to run the p2p tests you added?  They are not part of validate, are they?

src/runtime_src/core/common/error.h
```diff
@@ -58,6 +58,14 @@ class system_error : public std::system_error
   {}
 };
 
+class generic_error : public std::system_error
+{
+public:
+  generic_error(int ec, const std::string& what = "")
```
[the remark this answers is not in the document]
stsoe: The alternative is multiple constructors same as system_error.  I took a shortcut, I believe empty message makes sense sometimes if std::error_codes are used.

src/runtime_src/core/pcie/windows/shim.cpp
```diff
@@ -1556,7 +1556,7 @@ xclUnmgdPwrite(xclDeviceHandle handle, unsigned int flags, const void *buf, size
   xrt_core::message::
     send(xrt_core::message::severity_level::XRT_DEBUG, "XRT", "xclUnmgdPwrite()");
   auto shim = get_shim_object(handle);
-  return shim->unmgd_pwrite(flags, buf, count, offset);
+  return shim->unmgd_pwrite(flags, buf, count, offset) ? 0 : 1;
```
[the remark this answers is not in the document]
stsoe: This function actually does not return number of bytes to write, it is supposed to return 0 on success and appropriate error code on error.  I should have replaced 1 with an error code, but point is I just fixed a bug.

src/runtime_src/core/pcie/windows/shim.cpp
```diff
@@ -1565,7 +1565,7 @@ xclUnmgdPread(xclDeviceHandle handle, unsigned int flags, void *buf, size_t coun
   xrt_core::message::
     send(xrt_core::message::severity_level::XRT_DEBUG, "XRT", "xclUnmgdPread()");
   auto shim = get_shim_object(handle);
-  return shim->unmgd_pread(flags, buf, count, offset);
+  return shim->unmgd_pread(flags, buf, count, offset) ? 0 : 1;
```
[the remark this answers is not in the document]
stsoe: Same answer :-)

src/runtime_src/core/pcie/windows/shim.cpp
```diff
@@ -1697,8 +1697,15 @@ xclReadTraceData(xclDeviceHandle handle, void* traceBuf, uint32_t traceBufSz,
   return size;
 }
 
-int xclGetSubdevPath(xclDeviceHandle handle,  const char* subdev,
-                        uint32_t idx, char* path, size_t size)
+int
+xclGetSubdevPath(xclDeviceHandle handle,  const char* subdev,
+                 uint32_t idx, char* path, size_t size)
 {
   return 0;
```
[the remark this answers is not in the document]
stsoe: Yes, this should be cleaned up.  It's part of our project to get a handle on error handling and error codes

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
stsoe: Yes, I didn't notice this function in XBUtilities.  However, I want exceptions, so we will have to defer consolidating this.

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
stsoe: Could point.  I was also not liking pattern.

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
stsoe: True.  Likely the same under the hood.  I have not used boost::format a lot, but point taken.

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
stsoe: yeah

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
stsoe: I think a comment will do explaining what the function does.  I pretty much just copied it and cleaned it up a litte

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
stsoe: Maybe.  Again, I just copied the code more or less,  Didn't exactly like the strings either.  This should be cleaned up for sure.

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
AShivangi: Yes, I agree that they should be exception-safe. I don't have a way of running each test manually, though that is a functionality I eventually want to add . p2ptest is a part of validate. I moved the code over from the old xbutil, but since your code is more efficient, we should replace my code with yours. 

If we are going to revisit this later, do you want to add a comment indicating that so that we don't forget later?

stsoe: I don't believe p2p test is part of validate, per the spec I am working off it is part of "advanced --p2p validate" ?

stsoe: I created XRT-694 to track code consolidation

AShivangi: Thanks for creating a task. What spec are you referring to? In legacy xbutil we have p2p as a standalone test as well as as a part of validate

PR #3873, opened 2020-08-15
enhance load xclbin flow for CU index and more

mamin506:
In the new KDS, the CU index is determined in driver.

### DIFF
--- a/src/runtime_src/core/common/xclbin_parser.cpp
+++ b/src/runtime_src/core/common/xclbin_parser.cpp
@@ -581,11 +581,48 @@ get_kernel_arguments(const char* xml_data, size_t xml_size, const std::string& k
   return args;
 }
 
+std::vector<std::string>
+get_kernel_names(const char *xml_data, size_t xml_size)
+{
+  std::vector<std::string> names;
+
+  pt::ptree xml_project;
+  std::stringstream xml_stream;
+  xml_stream.write(xml_data,xml_size);
+  pt::read_xml(xml_stream,xml_project);
+
+  for (auto& xml_kernel : xml_project.get_child("project.platform.device.core")) {
+    if (xml_kernel.first != "kernel")
+      continue;
+
+    names.push_back(xml_kernel.second.get<std::string>("<xmlattr>.name"));
+  }
+
+  return names;
+}
+
 std::vector<kernel_argument>
 get_kernel_arguments(const axlf* top, const std::string& kname)
 {
   auto xml = get_xml_section(top);
   return get_kernel_arguments(xml.first, xml.second, kname);
+}
+
+std::vector<kernel_object>
+get_kernels(const axlf* top)
+{
+  auto xml = get_xml_section(top);
+  std::vector<kernel_object> kernels;
+
+  auto knames = get_kernel_names(xml.first, xml.second);
+  for (auto kname : knames) {
+    kernels.emplace_back(kernel_object{
+       kname
+      ,get_kernel_arguments(xml.first, xml.second, kname)
+    });
+  }
+
+  return kernels;
 }
 
 bool
--- a/src/runtime_src/core/common/xclbin_parser.cpp
+++ b/src/runtime_src/core/common/xclbin_parser.cpp
@@ -615,7 +615,7 @@ get_kernels(const axlf* top)
   std::vector<kernel_object> kernels;
 
   auto knames = get_kernel_names(xml.first, xml.second);
-  for (auto kname : knames) {
+  for (auto& kname : knames) {
     kernels.emplace_back(kernel_object{
        kname
       ,get_kernel_arguments(xml.first, xml.second, kname)
--- a/src/runtime_src/core/edge/user/shim.cpp
+++ b/src/runtime_src/core/edge/user/shim.cpp
@@ -515,19 +515,20 @@ xclLoadAxlf(const axlf *buffer)
 
   auto kernels = xrt_core::xclbin::get_kernels(buffer);
   /* Calculate size of kernels */
-  for (auto kernel : kernels) {
+  for (auto& kernel : kernels) {
       axlf_obj.za_ksize += sizeof(kernel_info) + sizeof(argument_info) * kernel.args.size();
   }
 
   /* Check PCIe's shim.cpp for details of kernels binary */
-  axlf_obj.za_kernels = (char *)malloc(axlf_obj.za_ksize);
-  for (auto kernel : kernels) {
+  std::vector<char> krnl_binary(axlf_obj.za_ksize);
+  axlf_obj.za_kernels = krnl_binary.data();
+  for (auto& kernel : kernels) {
       auto krnl = reinterpret_cast<kernel_info *>(axlf_obj.za_kernels + off);
       strcpy(krnl->name, kernel.name.c_str());
       krnl->anums = kernel.args.size();
 
       int ai = 0;
-      for (auto arg : kernel.args) {
+      for (auto& arg : kernel.args) {
           strcpy(krnl->args[ai].name, arg.name.c_str());
           krnl->args[ai].offset = arg.offset;
           krnl->args[ai].size   = arg.size;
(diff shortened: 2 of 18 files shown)


src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1347,12 +1347,72 @@ int shim::xclLoadXclBin(const xclBin *buffer)
...
+     * |   anums               |
+     * |   argument[0]         |
+     * |   argument[1]         |
+     * |   argument[...]       |
+     * |-----------------------|
+     * | Kernel[...]           |
+     * |   ...                 |
+     * +-----------------------+
+     */
+    axlf_obj.kernels = (char *)malloc(axlf_obj.ksize);
```
stsoe, line 1390: Please no malloc/free/new/delete.  Same comment as earlier

src/runtime_src/core/common/xclbin_parser.cpp
```diff
@@ -581,13 +581,50 @@ get_kernel_arguments(const char* xml_data, size_t xml_size, const std::string& k
...
 }
 
+std::vector<kernel_object>
+get_kernels(const axlf* top)
+{
+  auto xml = get_xml_section(top);
+  std::vector<kernel_object> kernels;
+
+  auto knames = get_kernel_names(xml.first, xml.second);
+  for (auto kname : knames) {
```
stsoe, line 618: auto&

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -508,7 +509,38 @@ xclLoadAxlf(const axlf *buffer)
   drm_zocl_axlf axlf_obj = {
     .za_xclbin_ptr = const_cast<axlf *>(buffer),
     .za_flags = flags,
+    .za_ksize = 0,
+    .za_kernels = NULL,
   };
+
+  auto kernels = xrt_core::xclbin::get_kernels(buffer);
+  /* Calculate size of kernels */
+  for (auto kernel : kernels) {
```
stsoe, line 518: auto&

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -508,7 +509,38 @@ xclLoadAxlf(const axlf *buffer)
...
+      axlf_obj.za_ksize += sizeof(kernel_info) + sizeof(argument_info) * kernel.args.size();
+  }
+
+  /* Check PCIe's shim.cpp for details of kernels binary */
+  axlf_obj.za_kernels = (char *)malloc(axlf_obj.za_ksize);
```
stsoe, line 523: Please no malloc/free/new/delete, this is not exception safe
Use std::vector<char*> data(size) to manage the buffer and extract the buffer where necessary using vec.data().

But I am not seeing a free of this buffer.  Wouldn't it be better if the driver allocated the memory and did the copying?   A dangling malloc is not acceptable.  You already have all the data allocated on the user side, so the copying done here could as well be done in the driver.  You can define a struct like kernel_info that just contains the user pointer addresses already allocated. e.g. kernel.name.c_str() which is valid for the duration of the xclLoad call so no additional strcpy is necessary.

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -508,7 +509,38 @@ xclLoadAxlf(const axlf *buffer)
...
+  for (auto kernel : kernels) {
```
stsoe, line 524: auto&

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -508,7 +509,38 @@ xclLoadAxlf(const axlf *buffer)
...
+      auto krnl = reinterpret_cast<kernel_info *>(axlf_obj.za_kernels + off);
+      strcpy(krnl->name, kernel.name.c_str());
+      krnl->anums = kernel.args.size();
+
+      int ai = 0;
+      for (auto arg : kernel.args) {
```
stsoe, line 530: auto&

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1347,12 +1347,72 @@ int shim::xclLoadXclBin(const xclBin *buffer)
 int shim::xclLoadAxlf(const axlf *buffer)
 {
     xrt_logmsg(XRT_INFO, "%s, buffer: %s", __func__, buffer);
+    drm_xocl_axlf axlf_obj = {const_cast<axlf *>(buffer), 0};
+    int off = 0;
+
+    auto kernels = xrt_core::xclbin::get_kernels(buffer);
+    /* Calculate size of kernels */
+    for (auto kernel : kernels) {
```
stsoe, line 1355: auto&

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1347,12 +1347,72 @@ int shim::xclLoadXclBin(const xclBin *buffer)
...
+     * |   argument[0]         |
+     * |   argument[1]         |
+     * |   argument[...]       |
+     * |-----------------------|
+     * | Kernel[...]           |
+     * |   ...                 |
+     * +-----------------------+
+     */
+    axlf_obj.kernels = (char *)malloc(axlf_obj.ksize);
+    for (auto kernel : kernels) {
```
stsoe, line 1391: auto&

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1347,12 +1347,72 @@ int shim::xclLoadXclBin(const xclBin *buffer)
...
+        auto krnl = reinterpret_cast<kernel_info *>(axlf_obj.kernels + off);
+        strcpy(krnl->name, kernel.name.c_str());
+        krnl->anums = kernel.args.size();
+
+        int ai = 0;
+        for (auto arg : kernel.args) {
```
stsoe, line 1397: auto&

maxzhen:
> 
> 
> Cool. Lots of changes :-). I added some requests for changes.
> 
> A general question: User land does not use the ordering determined by the driver, in other words, xclbin_parser::get_cus() is still computing the order? We want to change this, right? Not needed now of course.

Hi @stsoe , yes, we want to change this so that xclbin_parser::get_cus() can read all CU information from driver including the index. @mamin506 , if the order you have in driver is different than in the get_cus(), we will have to change it now. I think that the cu order in user space and driver should be in sync.

mamin506:
> Hi @stsoe , yes, we want to change this so that xclbin_parser::get_cus() can read all CU information from driver including the index. @mamin506 , if the order you have in driver is different than in the get_cus(), we will have to change it now. I think that the cu order in user space and driver should be in sync.

The order is the same. But current get_cus() still required by mb_scheduler. We would definitely change this function and get CU information from driver.

src/runtime_src/core/pcie/linux/shim.cpp
mamin506: fixed

src/runtime_src/core/pcie/linux/shim.cpp
mamin506: fixed

src/runtime_src/core/pcie/linux/shim.cpp
mamin506: fixed

src/runtime_src/core/edge/user/shim.cpp
mamin506: fixed

src/runtime_src/core/edge/user/shim.cpp
mamin506: fixed

src/runtime_src/core/edge/user/shim.cpp
mamin506: yes, you are right. Driver would allocate a buffer and copy this binary.
Use c++ container is much more better. :)

src/runtime_src/core/edge/user/shim.cpp
mamin506: fixed

src/runtime_src/core/common/xclbin_parser.cpp
mamin506: fixed

src/runtime_src/core/pcie/linux/shim.cpp
mamin506: fixed

PR #3876, opened 2020-08-17
Expose xrt::bo::address

stsoe:
Per request address is needed for non HLS style kernels.


stsoe:
@larry9523   FYI.  xrtBOAddress is moved to public API and no longer internal, shouldn't affect Jerry, but just in case.

PR #3877, opened 2020-08-17
CR-1073017 system crash after canceling application and issuing hot reset

mamin506:
If ctrl-c when downloading xclbin, thread_run() would return error.


maxzhen:
mark as do-not-merge, waiting for sumit's approval.

PR #3880, opened 2020-08-18
xbmgmt advanced --config

stsoe:


src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
```diff
@@ -81,38 +176,53 @@ show_device_conf(std::shared_ptr<xrt_core::device>& dev)
 static void 
 update_daemon_config(const std::string& host)
 {
-  std::cout << "update config with " << host;
+  XBU::sudo_or_throw("Updating daemon configuration requires sudo");
```
AShivangi, line 179: xbmgmt can't run any commands without sudo. Is this needed?

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
```diff
@@ -146,7 +257,8 @@ OO_Config::OO_Config( const std::string &_longName, bool _isHidden)
     ("security", boost::program_options::value<decltype(m_security)>(&m_security), "<add description>")
     ("runtime_clk_scale", boost::program_options::value<decltype(m_clk_scale)>(&m_clk_scale), "<add description>")
     ("cs_threshold_power_override", boost::program_options::value<decltype(m_power_override)>(&m_power_override), "<add description>")
-    ("show", boost::program_options::bool_switch(&m_show), "<add description>")
+    ("cs_reset", boost::program_options::value<decltype(m_cs_reset)>(&m_cs_reset), "<add description>")
+    ("showx", boost::program_options::bool_switch(&m_show), "<add description>")
```
AShivangi, line 261: showx typo

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -386,13 +386,24 @@ XBUtilities::collect_devices( const std::set<std::string> &_deviceBDFs,
 }
 
 xrt_core::device_collection
-XBUtilities::collect_devices( const std::string& _devices, // comma separated no space
-                              bool _inUserDomain )
+XBUtilities::
+collect_devices(const std::vector<std::string>& _devices, bool _inUserDomain)
```
AShivangi, line 390: Should this be exposed? This looks like a helper function to the following func

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
```diff
@@ -32,46 +32,141 @@ namespace po = boost::program_options;
...
+  {
+    ostr << "host=" << host << std::endl;
+  }
+};
+
+static std::string
+get_hostname()
+{
+  std::string hn;
+#ifdef __GNUC__
```
AShivangi, line 74: We should get the hostname from `xrt_core::get_os_info(...)` or maybe we should consider adding device_queries?

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
stsoe: sudo is not enforced by xbmgmt2, don't know if it should be? But certain operations that change the device are not possible unless run as sudo, rather than mysteriously erroring out sudo_or_throw will cause graceful exit with proper message.

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
stsoe: Actually not a typo, but a work-around until command parse can handle -show without it being captured by --show-hidden.

src/runtime_src/core/tools/common/XBUtilities.cpp
stsoe: Not sure what you mean by exposed?  Some commands build a device list as a comma separated string, others build it as a vector of string.  I don't know why there is this inconsistency, but I am hiding it in XBU so commands look the same regardless of device list format.

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
stsoe: Yes, I wanted to consolidate this some place else, but xrt_core::get_os_info take a property tree, to me that is not the right API,

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
AShivangi: Hmm. I think we do enforce sudo:
```
/opt/xilinx/xrt/bin/xbmgmt --new status
XRT build version: 2.8.0
Build hash: 9aa1c2cd35d8ad401c673a9ea6d5630ede456f99
Build date: 2020-08-14 10:31:27
Git branch: master
PID: 219041
UID: 31939
[Tue Aug 18 19:53:29 2020 GMT]
HOST: xsjsagarwal
[xbmgmt] ERROR: Root privileges required
```
All xbmgmt commands should require sudo. If there is a scenario where we are able to run xbmgmt2 without sudo, we need to fix that immediately. Can you point me to a case where you are able to do so? I can take a look.

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
AShivangi: Aah, interesting :)

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
maxzhen: not all xbmgmt subcommands needs sudo. xbmgmt scan is one of them.

src/runtime_src/core/tools/common/XBUtilities.cpp
AShivangi: I didn't realize that there was an inconsistency. My guess is it's happening because of how device gets initialized and captured by the boost program_options: vector captured using multitoken vs capturing a simple comma-separated string. 

We should work on removing: `collect_devices( const std::string& _devices, bool _inUserDomain )` and having only `collect_devices(const std::vector<std::string>& _devices, bool _inUserDomain)`. I can look into this once I have a chance. Created a ticket: XRT-699

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
AShivangi: I agree. I am moving away from xrt_core::get_xx_info(). I did that for electricals, thermals, etc. I'm directly using device queries for them in reports. Do you think we should create device_queries for system info as well?

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
AShivangi: Oh, I see. 
I checked and we don't have sudo enforced globally in xbmgmt2. So this function is good, sorry for the confusion :)

PR #3883, opened 2020-08-18
update zocl version

mamin506:
Per @imrickysu 's request, update zocl driver version for future releases.


mamin506:
Agree, ideally it should be the same like xocl driver. Let's wait Ricky's reply.

mamin506:
Confirm with Ricky. "Take the way you think it's convenient". I would close this one and create a XRT task to refine xbutil on edge.

PR #3888, opened 2020-08-19
Polpulating device's PCI slot address into existing field in xclDeviceInfo2

akum-xilinx:
The application needs FPGA device's PCI  slot address to be able to find its peer NVME device in case of SmartSSD.  The field already existed in the xclDeviceInfo2, which is returned by xclGetDeviceInfo2, but was not filled previously. Added this change to fill the PCI slot address in the following format: (domain<<16) + (bus<<8) + (dev<<3) + function.


chvamshi-xilinx:
ok to test

chvamshi-xilinx (approved):

maxzhen (approved):
xclDeviceInfo2 is pretty much dead now. maybe sysfs node is a better choice. but i have no objection to this change since it does not extending the existing data structure.

PR #3899, opened 2020-08-20
Added support for HBM backward compartibility of existing application

saifuddin-xilinx:
Added support for HBM backward compartibility of existing application

### DIFF
--- a/src/runtime_src/xocl/core/device.cpp
+++ b/src/runtime_src/xocl/core/device.cpp
@@ -713,7 +713,7 @@ get_cu_memidx() const
         mask &= cu->get_memidx_intersect();
 
       // select first common memory bank index if any
-      for (size_t idx=0; idx<mask.size(); ++idx) {
+      for (size_t idx=mask.size() - 1; idx>=0; --idx) {
         if (mask.test(idx)) {
           m_cu_memidx = idx;
           break;
--- a/src/runtime_src/xocl/core/device.cpp
+++ b/src/runtime_src/xocl/core/device.cpp
@@ -712,8 +712,8 @@ get_cu_memidx() const
       for (auto& cu : get_cu_range())
         mask &= cu->get_memidx_intersect();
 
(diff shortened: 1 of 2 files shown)


chvamshi-xilinx:
@saifuddin-xilinx , As discussed offline, Please simplify the code and dont remove the DRC.

chvamshi-xilinx (approved):
Looks good.  Can you please add a comment why DRC checks are performed before adding origincal connections.  It will be helpful

src/runtime_src/xocl/core/device.cpp
```diff
@@ -713,7 +713,7 @@ get_cu_memidx() const
         mask &= cu->get_memidx_intersect();
 
       // select first common memory bank index if any
-      for (size_t idx=0; idx<mask.size(); ++idx) {
+      for (size_t idx=mask.size() - 1; idx>=0; --idx) {
```
stsoe, line 716: Please elaborate on this change.  The comment no longer reflects the loop.  Why was it changed?

stsoe: The OCL code should be using GROUP_CONNECTIVITY and the connectivity in this section should always be valid groups.

@saifuddin-xilinx explains that APPs requested backwards compatibility to old mem topology indices, but we have to revisit this request.  Grouping is a new feature, existing host code and xclbins will continue to work without any changes.  

However changing the xclbin to use grouping of HBMs and changing CU connectivity implies that host code must change also.   One should not expect existing host code to work with xclbins that have been converted to use groups.

We should resolve this and revert the change in xocl/core/device.cpp.

chvamshi-xilinx:
Having offline discussions with @stsoe on this page.
https://confluence.xilinx.com/display/XSW/Specifying+Bank+index+in+Application

saifuddin-xilinx:
This is not required for now. We will see if there is any real use case..

PR #3917, opened 2020-08-25
checking unupgradable sc status need more work

xuhz:


xuhz:
this PR was firstly to fix the the flooding of 'xocl_err: SC is not ready, state=5' that eventually makes the system very very slow.
it also fixes this CR http://jira.xilinx.com/browse/CR-1073732 -- bdinfo not cached

PR #3919, opened 2020-08-25
Dynamically load XRT core library from coreutil

stsoe:
No longer require host applicaiton to link with xrt_core.  Instead
load proper core library at run-time.  Application can continue to
link with core library if necessary for some reason.

Remove explicit linking with xrt_core library from XRT native tests.


stsoe:
@jvillarre I had to re-purpose the module loader to share code with shim loader, please check my changes in module_loader.* to ensure I haven't broken anything when loading modules.  Thanks,

IshitaGhosh:
module loader change looks good.

PR #3922, opened 2020-08-25
CR-1073568 ASTeR - Cannot program the Shell/PLP after installing packages as the process stalls or reboots the server with a kernal panic

chienwei-lan:


xdavidz:
this caused insmod: ERROR: could not insert module /proj/rdi/staff/davidzha/XRT_xocl/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/xclmgmt.ko: Cannot allocate memory
but I am curious why pipeline can pass.

PR #3925, opened 2020-08-26
CR-1073645 Fix Versal build fail

rbramand-xilinx:
> setAieArray function was not defined, which is causing error
> Replaced the function call with equivalent functionality


rbramand-xilinx:
This issue is already addressed in other pull request, so closing this one

PR #3926, opened 2020-08-26
CR-1071805 ICAP Interruption Mitigation - XRT Support

xdavidz:
tested on u200, and u50.

Asked Nigel and Shane if this procedure is exactly what they want.
Nigel is ok with this. Shane confirms the current procedure is ok too.


xilinxgitops:
retest this please

xdavidz:
retest this please

PR #3860, opened 2020-08-13
xbutil2 report memory

Himanshu-xilinx:
* xbutil report memory
```
$ xbutil2 examine -r memory 
Memory Information

  ECC Error Status
    Tag     Errors      CE Count  UE Count  CE FFA              UE FFA              
    bank0   UE           0         1         0x0                   0xfffcffff                                     

  Memory Topology
         Tag         Type        Temp(C)  Size    
    [ 0] bank0       MEM_DDR4    33       16 GB   
    [ 1] bank1       MEM_DDR4    35       16 GB   
    [ 2] bank2       MEM_DDR4    46       16 GB   
    [ 3] bank3       MEM_DDR4    42       16 GB   
    [ 4] PLRAM[0]    MEM_DRAM    N/A      128 KB  
    [ 5] PLRAM[1]    MEM_DRAM    N/A      128 KB  
    [ 6] PLRAM[2]    MEM_DRAM    N/A      128 KB  

  Memory Status
         Tag         Type        Size    Mem Usage       BO count
    [ 0] bank0       MEM_DDR4    16 GB   0 Byte          0       
    [ 1] bank1       MEM_DDR4    16 GB   0 Byte          0       
    [ 2] bank2       MEM_DDR4    16 GB   0 Byte          0       
    [ 3] bank3       MEM_DDR4    16 GB   0 Byte          0       

  DMA Transfer Metrics
    Chan[ 0].h2c:  2048 MB
    Chan[ 0].c2h:  2048 MB
    Chan[ 1].h2c:  2 GB
    Chan[ 1].c2h:  2 GB
	
  Streams
         Tag         Status   Total (B/#)     Pending (B/#)   
    [ 0] dc_0        N/A      N/A             N/A             
    [ 1] dc_1        N/A      N/A             N/A    

	
$ xbutil2 examine -r memory -f JSON-2020.2
{
    "schema_version":
    {
        "schema": "JSON-2020.2",
        "creation_date": "Wed Aug 19 13:48:01 2020 GMT"
    },
    "devices":
    [
        {
            "mem_topology":
            {
                "description": "Memory Information",
                "board":
                {
                    "direct_memory_accesses":
                    {
                        "type": "pcie xdma",
                        "metrics":
                        [
                            {
                                "channel_id": "0",
                                "host_to_card_bytes": "0x800609ff",
                                "card_to_host_bytes": "0x800c09ff"
                            },
                            {
                                "channel_id": "1",
                                "host_to_card_bytes": "0x80000000",
[...]

PR #3932, opened 2020-08-26
Fix up query request with dynamic subdev and entry

stsoe:
Tweaks to #3860.  Avoid fixing query argument to std::pair, make the
dynamic nature of subdev and entry explicit through a modifier accessor
function.

Subject to change, but little cleaner and more explicit than before.


Himanshu-xilinx (approved):
changes looks good

PR #3934, opened 2020-08-26
Add check for NODMA

stsoe:
Cache the check in core device to avoid repeated sysfs access.

Enable nodma check in xrtBOAlloc.

Expose nodma check via clGetDeviceInfo (CL_DEVICE_NODMA) per cl_ext_xilinx.h.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -39,12 +39,14 @@
 #endif
 
 namespace {
+
 static bool
-is_nodma()
+is_nodma(xclDeviceHandle xhdl)
 {
-  // TODO
-  return false;
+  auto core_device = xrt_core::get_userpf_device(xhdl);
+  return core_device->is_nodma();
 }
+
 }
 
 ////////////////////////////////////////////////////////////////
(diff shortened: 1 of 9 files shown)


stsoe:
@maxzhen @houlz0507 Just FYI that I added and enabled nodma xrtBOAlloc flow and exposed nodma via clGetDeviceInfo.

src/runtime_src/core/common/api/xrt_bo.cpp
```diff
@@ -514,7 +516,7 @@ alloc(xclDeviceHandle dhdl, size_t sz, xrtBufferFlags flags, xrtMemoryGroup grp)
   switch (type) {
   case 0:
 #ifndef XRT_EDGE
-    if (is_nodma())
+    if (is_nodma(dhdl))
       return alloc_nodma(dhdl, sz, flags, grp);
```
maxzhen, line 520: will alloc_nodma return error if user tries to alloc normal BO?

stsoe: @maxzhen What is a normal BO?  Do you mean userptr bo?

maxzhen: OK, it's good if driver returns failure.

PR #3938, opened 2020-08-27
CR-1072101 ASTeR TC 7.06/7.07 - xbutil dump is returning a non-zero return code when xbtest is running due to server crash

mamin506:
.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
@@ -1945,7 +1945,6 @@ exec_cfg_cmd(struct exec_core *exec, struct xocl_cmd *xcmd)
 		return 1;
 	}
 
-	mutex_lock(&exec->exec_lock);
 	if (major > 2) {
 		DRM_INFO("Unknown ERT major version, fallback to KDS mode\n");
 		ert_full = 0;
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
```diff
@@ -4699,6 +4702,7 @@ kds_custat_show(struct device *dev, struct device_attribute *attr, char *buf)
 	unsigned int idx = 0;
 	ssize_t sz = 0;
 
+	mutex_lock(&exec->exec_lock);
 	// No need to lock exec, cu stats are computed and cached.
 	// Even if xclbin is swapped, the data reflects the xclbin on
 	// which is was computed above.
```
stsoe, line 4708: Based on these changes the comment is not correct.   Locking here stalls scheduler every time custat is accessed, that doesn't seem right.  

Is it not possible to synchronize num_cus used in this function with the number of cus created by exec_cfg?   Essentially, keep a variable that reflects exec->num_cus but is set onlyafter the CUs have been created.  Since the CUs once created are never deleted (can only increase) they will be valid for access of the data needed by this function, though the data shown may not be valid if the xclbin is being swapped at the same time, but that doesn't matter.

mamin506: Thanks Soren for the comment. I use cfg->num_cus in the loop and update exec->num_cus after CUs have been created. Base on the fact that CUs are never deleted until mb_scheduler_remove()

PR #3939, opened 2020-08-27
xbutil2 reset aie array

Himanshu-xilinx:
```
$ xbutil2 reset -h    
                                                                                                                                                                  
DESCRIPTION: Resets the given device.

USAGE: xbutil reset [--help] [-d arg] [-r arg]

OPTIONS:
  -d, --device       - The Bus:Device.Function (e.g., 0000:d8:00.0) device of interest.  
                       A value of 'all' (default) indicates that every found device 
                       should be examined.
  -r, --type         - The type of reset to perform. Types resets available:
                         hot          - Hot reset (default)
                         aie          - Reset Aie array

  --help             - Help to use this sub-command


$ xbutil2 reset -r aie
Performing 'aie' reset on 
  -[0000:00:00.0]

Successfully reset Device[0000:00:00.0]

### DIFF
--- a/src/runtime_src/core/edge/user/device_linux.cpp
+++ b/src/runtime_src/core/edge/user/device_linux.cpp
@@ -49,7 +49,8 @@ struct bdf
   static result_type
   get(const xrt_core::device* device, key_type)
   {
-    return std::make_tuple(0,0,0);
+    //return std::make_tuple(0,0,0);
+    return {0,0,0};
   }
 
 };
@@ -62,8 +63,7 @@ struct board_name
   get(const xrt_core::device* device, key_type)
   {
     result_type deviceName("edge");
-    std::ifstream VBNV;
-    VBNV.open("/etc/xocl.txt");
+    std::ifstream VBNV("/etc/xocl.txt");
     if (VBNV.is_open()) {
       VBNV >> deviceName;
     }
@@ -295,16 +295,12 @@ reset(const char* subdev, const char* key, const char* value) const
   switch(val) {
   case 1:
     throw error(-ENODEV, "Hot reset not supported");
-    break;
   case 2:
     throw error(-ENODEV, "OCL dynamic region reset not supported");
-    break;
   case 3:
     throw error(-ENODEV, "ERT reset not supported");
-    break;
   case 4:
     throw error(-ENODEV, "Soft Kernel reset not supported");
-    break;
   case 5:
     if (auto ret = xclResetAIEArray(get_device_handle()))
       throw error(ret, "fail to reset aie array");
--- a/src/runtime_src/core/edge/user/device_linux.cpp
+++ b/src/runtime_src/core/edge/user/device_linux.cpp
@@ -42,6 +42,47 @@ get_edgedev(const xrt_core::device* device)
   return zynq_device::get_dev();
 }
 
+struct bdf 
+{
+  using result_type = query::pcie_bdf::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type)
+  {
+    return std::make_tuple(0,0,0);
+  }
+
+};
+
+struct board_name 
+{
+  using result_type = query::board_name::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type)
+  {
+    result_type deviceName("edge");
+    std::ifstream VBNV;
+    VBNV.open("/etc/xocl.txt");
+    if (VBNV.is_open()) {
+      VBNV >> deviceName;
+    }
+    VBNV.close();
+    return deviceName;
+  }
+};
+
+struct is_ready 
+{
+  using result_type = query::is_ready::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type)
+  {
+    return true;
+  }
+};
+
 struct devInfo
 {
   static boost::any
@@ -166,6 +207,14 @@ emplace_func0_get()
   query_tbl.emplace(k, std::make_unique<function0_get<QueryRequestType, Getter>>());
 }
 
+template <typename QueryRequestType, typename Getter>
+static void
+emplace_func0_request()
+{
+  auto k = QueryRequestType::key;
+  query_tbl.emplace(k, std::make_unique<function0_get<QueryRequestType, Getter>>());
+}
+
 static void
 initialize_query_table()
 {
@@ -185,6 +234,9 @@ initialize_query_table()
   emplace_sysfs_get<query::memstat>                   ("memstat");
   emplace_sysfs_get<query::memstat_raw>               ("memstat_raw");
   emplace_sysfs_get<query::error>                     ("error");
+  emplace_func0_request<query::pcie_bdf,                bdf>();
+  emplace_func0_request<query::board_name,              board_name>();
+  emplace_func0_request<query::is_ready,                is_ready>();
 }
 
 struct X { X() { initialize_query_table(); } };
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: src/runtime_src/core/edge/user/device_linux.cpp [hunk 1]
```diff
@@ -42,6 +42,47 @@ get_edgedev(const xrt_core::device* device)
...
+struct board_name 
+{
+  using result_type = query::board_name::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type)
+  {
+    result_type deviceName("edge");
+    std::ifstream VBNV;
+    VBNV.open("/etc/xocl.txt");
```
keryell, line 66: What is this file?
Strange extension in a Unix world...
Should it be under a `/etc/xilinx` or `/etc/default` or whatever?

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -42,6 +42,47 @@ get_edgedev(const xrt_core::device* device)
...
+
+struct board_name 
+{
+  using result_type = query::board_name::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type)
+  {
+    result_type deviceName("edge");
+    std::ifstream VBNV;
```
keryell, line 65: Merge with the next line

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -42,6 +42,47 @@ get_edgedev(const xrt_core::device* device)
...
 }
 
+struct bdf 
+{
+  using result_type = query::pcie_bdf::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type)
+  {
+    return std::make_tuple(0,0,0);
```
keryell, line 52: Perhaps
```suggestion
    return {0,0,0};
```
works?

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -42,6 +42,47 @@ get_edgedev(const xrt_core::device* device)
...
+  using result_type = query::board_name::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type)
+  {
+    result_type deviceName("edge");
+    std::ifstream VBNV;
+    VBNV.open("/etc/xocl.txt");
+    if (VBNV.is_open()) {
+      VBNV >> deviceName;
```
keryell, line 68: What if there are several devices?

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -42,6 +42,47 @@ get_edgedev(const xrt_core::device* device)
...
+  }
+
+};
+
+struct board_name 
+{
+  using result_type = query::board_name::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type)
```
keryell, line 62: Why pointer and not reference?

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -42,6 +42,47 @@ get_edgedev(const xrt_core::device* device)
...
+    return deviceName;
+  }
+};
+
+struct is_ready 
+{
+  using result_type = query::is_ready::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type)
```
keryell, line 80: Why pointer and not reference?

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -233,4 +285,33 @@ write(uint64_t offset, const void* buf, uint64_t len) const
...
+void 
+device_linux::
+reset(const char* subdev, const char* key, const char* value) const 
+{
+  int val = atoi(value);
+
+  switch(val) {
+  case 1:
+    throw error(-ENODEV, "Hot reset not supported");
+    break;
```
keryell, line 298: Give me a `break` here ;-)
I guess you can remove all the breaks...

src/runtime_src/core/edge/user/device_linux.cpp
stsoe: @keryell  This is called from a shared_ptr.get().  It is a normal and good idiom to extract pointer from the smart pointer when calling a function where arguments stay in scope for the duration of the call.   It is extremely awkward in my opinion to deference the pointer to convert it to a reference.

src/runtime_src/core/edge/user/device_linux.cpp
stsoe: haha :-)  For all but 5, yeah.

keryell: Do not brake so hard! Yes, case 5 is a good br(e)aking point :-)

src/runtime_src/core/edge/user/device_linux.cpp
keryell: I see. I am just browsing the PR without the big picture... :-/

stsoe: Understood :-)

src/runtime_src/core/edge/user/device_linux.cpp
Himanshu-xilinx: done

src/runtime_src/core/edge/user/device_linux.cpp
Himanshu-xilinx: it works :)

src/runtime_src/core/edge/user/device_linux.cpp
Himanshu-xilinx: its for embedded device.... will be only one

src/runtime_src/core/edge/user/device_linux.cpp
Himanshu-xilinx: reduced the 'g' force :)

larry9523:
As discussed, we need to comment out calling xrtResetAIEArray for now.

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -233,4 +284,29 @@ write(uint64_t offset, const void* buf, uint64_t len) const
...
 
+
+void 
+device_linux::
+reset(const char* subdev, const char* key, const char* value) const 
+{
+  int val = atoi(value);
+
+  switch(val) {
+  case 1:
```
rozumx, line 295: The current code flow here has issues.  It goes from a "string" to an "enum" to a "string" to a "numeric" value.   

I do understand that you were not the original author here.  It would be most helpful if you did refactor the code so that to use enumeration values,  I do realize that this will impact several areas of the code base when you do.

Himanshu-xilinx: hi Stephen, 
i refactored the code.
Please review.

keryell (commented):
There is almost no comment in the new code introduced... :-(

rozumx (approved):
Thank you for refactoring the code. It is much cleaner and easier to understand.

larry9523 (approved):

src/runtime_src/core/edge/user/device_linux.cpp
stsoe: No, it doesn't work with older compilers, e.g. gcc version 5.4.0.  I am reverting.

keryell: People who do not care about modern current software could use old versions of XRT and old FPGA anyway...
At the end there is a technical debt we all pay just because a lot of irresponsible ignorant managers all over the place... :-(

PR #3940, opened 2020-08-27
Support NODMA BOs in OCL

stsoe:


src/runtime_src/xrt/device/hal2.cpp
```diff
@@ -484,13 +525,21 @@ read(const BufferObjectHandle& boh, void* dst, size_t sz, size_t offset, bool as
...
 {
   xclBOSyncDirection dir = XCL_BO_SYNC_BO_TO_DEVICE;
   if(dir1 == direction::DEVICE2HOST)
     dir = XCL_BO_SYNC_BO_FROM_DEVICE;
 
   BufferObject* bo = getBufferObject(boh);
+  if (bo->nodma) {
+    // sync is M2M copy between host and device handle
+    if (dir == XCL_BO_SYNC_BO_TO_DEVICE)
+      return event(typed_event<int>(m_ops->mCopyBO(m_handle, bo->handle, bo->nodma_host_handle, sz, offset, offset)));
```
keryell, line 539: Perhaps this works:
```suggestion
      return { typed_event<int>(m_ops->mCopyBO(m_handle, bo->handle, bo->nodma_host_handle, sz, offset, offset)) };
```

stsoe: Yeah, sure.  This file is old and will be rewritten to use native APIs, so  will not bother with any style changes.

PR #3944, opened 2020-08-27
CR-1074051 HBM ECC status is not being updated for xbutil unless run as root user

chienwei-lan:
& CR-1074058 ASTeR TC1.01 - ERROR: could not insert 'xclmgmt': Cannot allocate memory

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/ulite.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/ulite.c
@@ -22,10 +22,8 @@
 #include "../xocl_drv.h"
 #include "mgmt-ioctl.h"
 
-#define ULITE_NAME		"ttyUL"
-#define ULITE_MAJOR		204
-#define ULITE_MINOR		187
-#define ULITE_NR_UARTS		64
+#define ULITE_NAME		"ttyXRTUL"
+#define ULITE_NR_UARTS		8
 
 /* ---------------------------------------------------------------------
  * Register definitions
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/ulite.c
```diff
@@ -438,8 +436,6 @@ static struct uart_driver xcl_ulite_driver = {
 	.owner		= THIS_MODULE,
 	.driver_name	= XOCL_DEVNAME(XOCL_UARTLITE),
 	.dev_name	= ULITE_NAME,
-	.major		= ULITE_MAJOR,
-	.minor		= ULITE_MINOR,
 	.nr		= ULITE_NR_UARTS,
```
houlz0507, line 439: This means that if you do not specify Major, then linux uart will allocate one, correct?

chienwei-lan: exactly,

__register_chrdev_region(unsigned int major, unsigned int baseminor,
			   int minorct, const char *name)
{
....

	if (major == 0) {
		ret = find_dynamic_major();

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_sysfs.c
```diff
@@ -412,16 +412,16 @@ static ssize_t ulp_uuids_show(struct device *dev,
...
-		struct device_attribute *da, const char *buf, size_t count)
+static ssize_t mig_cache_update_show(struct device *dev,
+	struct device_attribute *attr, char *buf)
 {
 	struct xocl_dev *xdev = dev_get_drvdata(dev);
 
 	xocl_update_mig_cache(xdev);
 
-	return count;
+	return 0;
```
houlz0507, line 422: This looks odd to me. It is reading but always return nothing. I understand it is triggering cache updating for Docker case.  maybe we can update the cache in ecc_status read?

maxzhen:
I'll merge it now since it contains a critical fix. @chienwei-lan , we might need a per-device subdev managing cache for all ECC status apart from mig subdev.

PR #3954, opened 2020-08-28
Fix build on Ubuntu 20.04 where cmake fails to configure Boost correc…

sonals:
…tly when it attempts to use boost-cmake which is bundled with Boost

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -16,6 +16,13 @@ if [[ $OSDIST == "centos" ]] || [[ $OSDIST == "amzn" ]] || [[ $OSDIST == "rhel" 
     fi
 fi
 
+CMAKE_EXTRA_DEF=""
+
+if [[ $OSDIST == "ubuntu" ]]; then
+    OS_VERSION_ID=`grep '^VERSION_ID=' /etc/os-release | awk -F= '{print $2}' | tr -d '"'`
+    CMAKE_EXTRA_DEF="-DBoost_NO_BOOST_CMAKE=ON"
+fi
+
 if [[ $CPU == "aarch64" ]] && [[ $OSDIST == "ubuntu" ]]; then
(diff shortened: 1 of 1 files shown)


build/build.sh
```diff
@@ -16,6 +16,13 @@ if [[ $OSDIST == "centos" ]] || [[ $OSDIST == "amzn" ]] || [[ $OSDIST == "rhel"
     fi
 fi
 
+CMAKE_EXTRA_DEF=""
+
+if [[ $OSDIST == "ubuntu" ]]; then
+    OS_VERSION_ID=`grep '^VERSION_ID=' /etc/os-release | awk -F= '{print $2}' | tr -d '"'`
```
chvamshi-xilinx, line 22: OS_VERSION_ID is extracted but not used?

keryell: Otherwise for this we just need to `source /etc/os-release`

keryell:
> Sure this is okay, but where did we run into this issue building on Ubuntu 20.04? I am positive I have built XRT on Ubuntu 20.04.

Ubuntu 20.04 ships with either Boost 1.67 or 1.71.
I guess the problem here addresses the case for Boost 1.71? https://gitlab.kitware.com/cmake/cmake/-/issues/18865

keryell:
I have just tried compiling on 2 machines with Ubuntu 20.04, one with vanilla 1.71 and another with manually installed 1.74 and XRT from master branch compiles fine. Perhaps it is an issue with manually installed Boost 1.70?

sonals:
This was an 18.04 to 20.4 upgrade issue. Ensure that boost packages (version 1.65)  from Ubuntu 18.04 are uninstalled and only packages (1.71) from Ubuntu 20.04 packages are present after upgrade.

PR #3955, opened 2020-08-29
Initial add of pybind11 wrappers for the native C++ API

schelleg:
Initial add of pybind11 support for the C++ Native API.  This package can be built with
either cmake or pip install - see instructions in the README.  

A subset of the full API is included in the pull-request.  Currently the device, uuid, 
kernel, and bo classes are supported in order to run the simple designs in the test
folders.

Signed-off-by: Graham Schelle <graham.schelle@xilinx.com>

### DIFF
--- a/src/python/pybind11/CMakeLists.txt
+++ b/src/python/pybind11/CMakeLists.txt
@@ -17,11 +17,6 @@ if (DEFINED ENV{XCL_EMULATION_MODE})
   message("=============== XRT_CORE_LIB=${XRT_CORE_LIBRARY}")
 endif()
 
-find_library(xrt_core_LIBRARY
-  NAMES ${XRT_CORE_LIBRARY}
-  HINTS "${XILINX_XRT}/lib")
-message("xrt_core_LIBRARY=${xrt_core_LIBRARY}")
-
 find_library(xrt_coreutil_LIBRARY
   NAMES xrt_coreutil
   HINTS "${XILINX_XRT}/lib")
@@ -41,5 +36,5 @@ endif (NOT WIN32)
 include_directories(${XILINX_XRT}/include)
 
 pybind11_add_module(pyxrt src/pyxrt.cpp)
-target_link_libraries(pyxrt PRIVATE ${xrt_core_LIBRARY} ${xrt_coreutil_LIBRARY})
+target_link_libraries(pyxrt PRIVATE ${xrt_coreutil_LIBRARY})
 target_link_libraries(pyxrt PRIVATE ${uuid_LIBRARY} pthread)
--- a/src/python/pybind11/CMakeLists.txt
+++ b/src/python/pybind11/CMakeLists.txt
@@ -1,0 +1,45 @@
+project(pyxrt)
+CMAKE_MINIMUM_REQUIRED(VERSION 3.0.0)
+
+find_package(pybind11 REQUIRED)
+
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_VERBOSE_MAKEFILE ON)
+set(XILINX_XRT $ENV{XILINX_XRT})
+set(XRT_CORE_LIBRARY xrt_core)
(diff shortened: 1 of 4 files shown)


maxzhen:
ok to test

src/python/pybind11/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+
+find_library(xrt_core_LIBRARY
+  NAMES ${XRT_CORE_LIBRARY}
+  HINTS "${XILINX_XRT}/lib")
+message("xrt_core_LIBRARY=${xrt_core_LIBRARY}")
+
+find_library(xrt_coreutil_LIBRARY
+  NAMES xrt_coreutil
+  HINTS "${XILINX_XRT}/lib")
+message("xrt_coreutil_LIBRARY=${xrt_coreutil_LIBRARY}")
```
stsoe, line 28: Linking with XRT core library  is no longer needed.   The coreutil library loads the proper shim library at run-time.

src/python/pybind11/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+if (NOT WIN32)
+find_library(uuid_LIBRARY
+  NAMES uuid)
+message("uuid_LIBRARY=${uuid_LIBRARY}")
+endif (NOT WIN32)
+
+include_directories(${XILINX_XRT}/include)
+
+pybind11_add_module(pyxrt src/pyxrt.cpp)
+target_link_libraries(pyxrt PRIVATE ${xrt_core_LIBRARY} ${xrt_coreutil_LIBRARY})
```
stsoe, line 44: Remove linking of xrt_core unless of course you are using APIs from xrt.h which I don't think we should be doing.

src/python/pybind11/src/pyxrt.cpp
```diff
@@ -0,0 +1,176 @@
...
+#include <string>
+#include <cstring>
+#include <tuple>
+#include <typeinfo>
+
+// XRT includes
+#include "experimental/xrt_device.h"
+#include "experimental/xrt_kernel.h"
+#include "experimental/xrt_bo.h"
+#include "xrt.h"
```
stsoe, line 29: Is this include needed here? Using symbols out of xrt.h may imply the need to link with xrt_core.  This in turn complicates using python code with emulation.   If we rely on coreutil.so to load the shim (core) library then emulation automatically works.

src/python/pybind11/src/pyxrt.cpp
```diff
@@ -0,0 +1,176 @@
...
+    .value("XCL_BO_SYNC_BO_TO_DEVICE", xclBOSyncDirection::XCL_BO_SYNC_BO_TO_DEVICE)
+    .value("XCL_BO_SYNC_BO_FROM_DEVICE", xclBOSyncDirection::XCL_BO_SYNC_BO_FROM_DEVICE);
+
+
+/*
+ *
+ * XRT:: Functions
+ *
+ */
+m.def("xclProbe", &xclProbe);
```
stsoe, line 59: Don't wrap this function.   It's only purpose is to provide the number of boards hooked up to the system.  You don't need it, since opening of the device will (gracefully) fail if you provide a device index that doesn't exists.   I suspect xclProbe is the only reason you link with xrt_core.  If you check all the low level tests (tests/xrt) you will see I recently  removed the call to xclProbe from all of these.

src/python/pybind11/CMakeLists.txt
schelleg: removed library dependency in 2ccb6f8

src/python/pybind11/src/pyxrt.cpp
schelleg: removed xrt.h in 2ccb6f8

src/python/pybind11/src/pyxrt.cpp
schelleg: removed xclprobe from cpp and README ... calling device(0) passing simple test

PR #3959, opened 2020-08-31
NSPR-8172 monitor kernel clock shutdown status on U2 card

rajkumar-xilinx:
Add support in xrt to monitor the clock shutdown status in U2 card

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
@@ -453,7 +453,6 @@ static void check_pcie_link_toggle(struct xclmgmt_dev *lro, int clear)
 	u32 sts;
 	int err;
 
-
 	err = xocl_iores_read32(lro, XOCL_SUBDEV_LEVEL_BLD,
 			IORES_PCIE_MON, 0x8, &sts);
 	if (err)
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
```diff
@@ -493,6 +493,8 @@ static int health_check_cb(void *data)
 
 	(void) xocl_clock_status(lro, &latched);
 
+	xocl_xmc_clock_status(lro, &latched);
+
```
chvamshi-xilinx, line 497: Can you please add a comment explaining why do we need to add a separate logic for U.2. example: U.2 doesn't use UCS etc

chvamshi-xilinx (approved):
Logic looks good to me.. Please add some comments explaining the function.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
```diff
@@ -3877,6 +3880,26 @@ static int xmc_access(struct platform_device *pdev, enum xocl_xmc_flags flags)
 		raptor_cmc_access(pdev, flags);
 }
 
+static void clock_status_check(struct platform_device *pdev, bool *latched)
```
chvamshi-xilinx, line 3883: Same here. Please add some comments about this function. As CMC for U.2 is different, its good to add description about U.2 CMC behavior.

dayeh-xilinx:
retest this please

PR #3964, opened 2020-09-01
qdma4: enable libqdma4 debugfs, stmc chagnes

karenx-xilinx:
- enable libqdma4 debugfs
- at queue closing add polling of stmc context to make sure qdma<->stm reset is done
- temporarily change default interrupt mode to be direct interrupt instead of auto, as there is some bugs in interrupt ring being debugged.


karenx-xilinx:
duplicate

PR #3966, opened 2020-09-01
reintroduce xoclv2 to xbmgmt

maxzhen:
This is a new attempt to support xoclv2 driver in xbmgmt.
The difference b/w new and legacy driver lays in the sysfs node and devfs node access, which is now hidden inside pci_device_v2 class.
@AShivangi , I can't make it work for xbmgmt2 due to that it is still doing flash, sc update, etc via mapping of user bar in user space. The new driver does not support mmap of user bar at all...

### DIFF
--- a/src/runtime_src/core/pcie/linux/device_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/device_linux.cpp
@@ -350,6 +350,7 @@ initialize_query_table()
   emplace_sysfs_get<query::mig_ecc_ue_ffa>              ("mig", "ecc_ue_ffa");
   emplace_sysfs_get<query::flash_bar_offset>            ("flash", "bar_off");
   emplace_sysfs_get<query::is_mfg>                      ("", "mfg");
+  emplace_sysfs_get<query::mfg_ver>                     ("", "mfg_ver");
   emplace_sysfs_get<query::is_ready>                    ("", "ready");
   emplace_sysfs_get<query::f_flash_type>                ("flash", "flash_type");
   emplace_sysfs_get<query::flash_type>                  ("", "flash_type");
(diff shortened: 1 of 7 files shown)


src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -350,6 +350,7 @@ initialize_query_table()
   emplace_sysfs_get<query::mig_ecc_ue_ffa>              ("mig", "ecc_ue_ffa");
   emplace_sysfs_get<query::flash_bar_offset>            ("flash", "bar_off");
   emplace_sysfs_get<query::is_mfg>                      ("", "mfg");
+  emplace_sysfs_get<query::mfg_ver>                     ("", "mfg_ver");
```
AShivangi, line 353: Is there a difference between mfg and mfg_ver?

AShivangi (approved):
This looks good, but now I'm concerned about xbmgmt2 and xbmgmt going out of sync. We can take that offline.

src/runtime_src/core/pcie/linux/device_linux.cpp
maxzhen: mfg is indicating if a board is running golden image. mfg_ver is the version of the golden image.

maxzhen:
> 
> 
> This looks good, but now I'm concerned about xbmgmt2 and xbmgmt going out of sync. We can take that offline.

The main change in this PR is in scan.cpp/.h which is common b/w xbmgmt and xbmgmt2 so it does not add more to the mess.
However, we do need to stop accessing device directly from user space in xbmgmt2 and bring it on par with xbmgmt.

PR #3968, opened 2020-09-01
XRT-666 Support CU to host interrupt on SSv3 platform

mamin506:
Basic CU to host interrupt support. "xbutil validate -q" could pass.

NOT include in this commit:
1. Sysfs node to let user space switch interrupt/polling mode
2. User management interrupt

The not include items are on the working list. This change make it possible to achieve them.


mamin506:
> I think we agreed that we should not use void* as api interfaces because of that any pointer can be casted to void* automatically. We could change this later.

Let me change it in this Pull request. That is simple. Max also has some suggestions. I would push a new commit for all those minor changes.

houlz0507 (approved):

PR #3978, opened 2020-09-02
Add missing import ctypes.util

keryell:
This fixes this issue::

  sudo /opt/xilinx/xrt/bin/xbutil validate

  INFO: Found 1 cards

  INFO: Validating card[0]: xilinx_u200_xdma_201830_2
  INFO: == Starting Kernel version check:
  INFO: == Kernel version check PASSED
  INFO: == AUX power connector check PASSED
  INFO: == Starting DMA test:
  Host -> PCIe -> FPGA write bandwidth = 9117.600575 MB/s
  Host <- PCIe <- FPGA read bandwidth = 11773.464290 MB/s
  INFO: == DMA test PASSED
  INFO: == Starting device memory bandwidth test:
  Traceback (most recent call last):
    File "/opt/xilinx/xrt/test/23_bandwidth.py", line 37, in <module>
      libc_name = ctypes.util.find_library("c")
  AttributeError: module 'ctypes' has no attribute 'util'

  ERROR: == device memory bandwidth test FAILED
  INFO: Card[0] failed to validate.

  ERROR: Some cards failed to validate.

I am wondering how it was able to work before...

### DIFF
--- a/tests/python/23_bandwidth/23_bandwidth.py
+++ b/tests/python/23_bandwidth/23_bandwidth.py
@@ -17,6 +17,7 @@
  under the License.
 """
 
+import ctypes.util
 import sys
 import time
 import math
--- a/tests/python/23_bandwidth/host_mem_23_bandwidth.py
+++ b/tests/python/23_bandwidth/host_mem_23_bandwidth.py
@@ -17,6 +17,7 @@
  under the License.
 """
 
+import ctypes.util
 import numpy as np
 import sys
 import time
--- a/tests/python/23_bandwidth/versal_23_bandwidth.py
+++ b/tests/python/23_bandwidth/versal_23_bandwidth.py
@@ -17,6 +17,7 @@
  under the License.
 """
 
+import ctypes.util
 import numpy as np
 import sys
 import time


maxzhen:
ok to test

keryell:
This is with Ubuntu 20.04.
You can try this on `xsjsycl41`.

AShivangi (commented):
We are importing ctypes in utils_binding.py, so I'm not sure why this isn't working...

keryell:
> We are importing ctypes in utils_binding.py, so I'm not sure why this isn't working...

This sounds like a bad programming style anyway to rely on some implementation details of a package to have some side effect on the global `import` namespace.
If someone change implementation details or Python security policy about transitive import improves in a newer version, it might break a consumer...

keryell:
I looked again at the current code and there is nowhere any `import ctypes.util` so I do not understand why it can work on some platforms. Probably the implementation of `ctypes` changed in recent Python? Ubuntu 20.04 comes with `Python 3.8.2`.
Otherwise, "We are importing ctypes in utils_binding.py," explains why there is no need for `import ctypes`, which is another story.

PR #3982, opened 2020-09-03
CR-1069984 using cl_mem_use_host_ptr with xcl_mem_ext_host_only works but gives incorrect report

chienwei-lan:


maxzhen (approved):

src/runtime_src/xocl/api/detail/memory.cpp
```diff
@@ -114,6 +114,9 @@ validHostPtrOrError(cl_mem_flags flags, const void* host_ptr)
     }
     if (bool(ubuf) && bool(ext_flags & XCL_MEM_EXT_P2P_BUFFER))
        throw error(CL_INVALID_HOST_PTR,"host_ptr with P2P buffer flags specified");
+
+    if (bool(ubuf) && bool(ext_flags & XCL_MEM_EXT_HOST_ONLY))
```
keryell, line 118: Strange code...
Is this:
```c++
     if (ubuf && (ext_flags & XCL_MEM_EXT_HOST_ONLY))
```

keryell (commented):
I am confused about the title "works but gives incorrect report".
What does it mean?

PR #3987, opened 2020-09-03
Support for AIE trace offload 

IshitaGhosh:
Add new xdp_aie_trace plugin which is loaded with aie_trace=true configuration in xrt.ini
This plugin offloads the trace data for AIE, for both PLIO and GMIO flows.


pgschuey (approved):
Changes look good to me. Besides AIE trace support, they include necessary changes to remove unused unified reports as well as cleanup and improvements of variable names, etc.

jvillarre (approved):
Changes and additions look good.  I would like two additional pull requests after this one.  One pull request to remove code commented out using #if 0, and one pull request to change the AIETraceWriter to inherit from VPWriter directly since it does not follow the same format as other VPTraceWriter classes and doesn't need the blank virtual functions.

PR #3989, opened 2020-09-03
Only register AIE when AIE_METADATA is present

larry9523:
Check for AIE metadata before registering AIE array


Himanshu-xilinx (approved):
looks good.

chvamshi-xilinx (approved):

PR #3995, opened 2020-09-03
skd: Install signal Handlers and attempt to capture stacktrace

rathaval:
In this PR, I am attempting to add some signal handling and stacktrace logging.

What we have observed while performing longevity tests or other such U30 tests, we would see hangs/failures. On further inspection, we observed that a few of the encode or decode soft-kernels were unresponsive to the XMA2 (host/x86) plugins.
There was no indication beyond some drm logs indicating that a drm device was closed by zocl.

In order to better understand failures of soft-kernels, here I present signal handlers and some backtrace walking to log a few symbols. In case of synchronous signals (like SIGSEGV, SIGABRT, SIGTERM), the relevant stacktrace gets logged to syslog.

There were a few of ways of doing this 

1. Use boost's stacktrace() functionality, however, with stdout being closed for daemonized processes. The other option was to use this and perform file I/O. Boost overloads the "<<" stream operator. Thus it can be logged to a file via that.

2. Using the backtrace() functionality from glibc, here a malloc is called internally by the backtrace_symbols() routine.

3. Using the backtrace_symbols_fd() from glibc, here an fd (file descriptor) is fed to the call, and I/O is performed to the file.

Options 1 & 3 involve performing file I/O within a signal handler and Option 2 involves performing a malloc within the signal handler. Neither is ideal, however, I am open to addressing a better solution. 

I went ahead with Option#2 since it gave us the simplest way to check for logs. We already log the encoder/decoder soft-kernel status to syslog, thus logging backtraces to syslog just meant we had one log to grok/grep/parse.
 
### DIFF
--- a/src/runtime_src/core/edge/skd/sk_daemon.cpp
+++ b/src/runtime_src/core/edge/skd/sk_daemon.cpp
@@ -17,6 +17,7 @@
  */
 
 #include <dlfcn.h>
+#include <execinfo.h>
 #include <string.h>
 #include <sys/wait.h>
 #include <sys/mman.h>
@@ -274,9 +275,35 @@ static int createSoftKernelFile(uint64_t paddr, size_t size, uint32_t cuidx)
   return 0;
 }
 
-/* Define a signal handler to avoid zombie process */
-static void sigAct(int sig)
+#define STACKTRACE_DEPTH	(25)
+static void stacktrace_logger(const int sig)
 {
+  const int stack_depth = STACKTRACE_DEPTH;
+  syslog(LOG_ERR, "%s - got %d\n", __func__, sig);
+  if (sig == SIGCHLD)
+    return;
+  void *array[stack_depth];
+  int nSize = backtrace(array, stack_depth);
+  char **symbols = backtrace_symbols(array, nSize);
+  if (symbols) {
+    for (int i = 0; i < nSize; i++)
+      syslog(LOG_ERR, "%s\n", symbols[i]);
+    free(symbols);
+  }
+}
+
+/* Define a signal handler for the child to handle signals */
+static void sigLog(const int sig)
+{
+  syslog(LOG_ERR, "%s - got %d\n", __func__, sig);
+  stacktrace_logger(sig);
+  exit(EXIT_FAILURE);
+}
+
+/* Define a signal handler for the parent to avoid zombie process */
+static void sigAct(const int sig)
+{
+  syslog(LOG_ERR, "%s - got %d\n", __func__, sig);
   wait((int *)0);
 }
 
@@ -285,11 +312,6 @@ void configSoftKernel(xclSKCmd *cmd)
 {
   pid_t pid;
   uint32_t i;
-
-  struct sigaction act;
-  act.sa_handler = sigAct;
-  sigemptyset(&act.sa_mask);
-  act.sa_flags = 0;
 
   if (createSoftKernelFile(cmd->xclbin_paddr, cmd->xclbin_size,
           cmd->start_cuidx) != 0)
@@ -304,6 +326,26 @@ void configSoftKernel(xclSKCmd *cmd)
     if (pid == 0) {
       char path[XRT_MAX_PATH_LENGTH];
       char proc_name[PNAME_LEN] = {};
+      /* Install Signal Handler for the Child Processes/Soft-Kernels */
+      struct sigaction act;
+      act.sa_handler = sigLog;
+      sigemptyset(&act.sa_mask);
+      act.sa_flags = 0;
+      sigaction(SIGHUP, &act, 0);
+      sigaction(SIGINT, &act, 0);
+      sigaction(SIGQUIT , &act, 0);
+      sigaction(SIGILL, &act, 0);
+      sigaction(SIGTRAP, &act, 0);
+      sigaction(SIGABRT, &act, 0);
+      sigaction(SIGBUS, &act, 0);
+      sigaction(SIGFPE, &act, 0);
+      sigaction(SIGKILL, &act, 0);
+      sigaction(SIGUSR1, &act, 0);
+      sigaction(SIGSEGV, &act, 0);
+      sigaction(SIGUSR2, &act, 0);
+      sigaction(SIGPIPE, &act, 0);
+      sigaction(SIGALRM, &act, 0);
+      sigaction(SIGTERM, &act, 0);
 
       (void)snprintf(proc_name, PNAME_LEN, "%s%d", cmd->krnl_name, i);
       if (prctl(PR_SET_NAME, (char *)proc_name) != 0) {
@@ -318,8 +360,13 @@ void configSoftKernel(xclSKCmd *cmd)
       exit(EXIT_SUCCESS);
     }
 
-    if (pid > 0)
+    if (pid > 0) {
+      struct sigaction act;
+      act.sa_handler = sigAct;
+      sigemptyset(&act.sa_mask);
+      act.sa_flags = 0;
       sigaction(SIGCHLD, &act, 0);
+    }
 
     if (pid < 0)
       syslog(LOG_ERR, "Unable to create soft kernel process( %d)\n", i);
--- a/src/runtime_src/core/edge/skd/CMakeLists.txt
+++ b/src/runtime_src/core/edge/skd/CMakeLists.txt
@@ -20,4 +20,6 @@ target_link_libraries(skd
   dl
   )
 
+set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -rdynamic")
+
 install (TARGETS skd RUNTIME DESTINATION ${XRT_INSTALL_DIR}/bin)


rathaval:
> 
> 
> In this PR, I am attempting to add some signal handling and stacktrace logging.
> 
> What we have observed while performing longevity tests or other such U30 tests, we would see hangs/failures. On further inspection, we observed that a few of the encode or decode soft-kernels were unresponsive to the XMA2 (host/x86) plugins.
> There was no indication beyond some drm logs indicating that a drm device was closed by zocl.
> 
> In order to better understand failures of soft-kernels, here I present signal handlers and some backtrace walking to log a few symbols. In case of synchronous signals (like SIGSEGV, SIGABRT, SIGTERM), the relevant stacktrace gets logged to syslog.
> 
> There were a few of ways of doing this
> 
>     1. Use boost's stacktrace() functionality, however, with stdout being closed for daemonized processes. The other option was to use this and perform file I/O. Boost overloads the "<<" stream operator. Thus it can be logged to a file via that.
> 
>     2. Using the backtrace() functionality from glibc, here a malloc is called internally by the backtrace_symbols() routine.
> 
>     3. Using the backtrace_symbols_fd() from glibc, here an fd (file descriptor) is fed to the call, and I/O is performed to the file.
> 
> 
> Options 1 & 3 involve performing file I/O within a signal handler and Option 2 involves performing a malloc within the signal handler. Neither is ideal, however, I am open to addressing a better solution.
> 
> I went ahead with Option#2 since it gave us the simplest way to check for logs. We already log the encoder/decoder soft-kernel status to syslog, thus logging backtraces to syslog just meant we had one log to grok/grep/parse.
> 
> Signed-off-by: Rohit Athavale [rathaval@xilinx.com](mailto:rathaval@xilinx.com)

@mamin506 @larry9523 : Could you take a look and  review this PR ? It's a few lines of diff

src/runtime_src/core/edge/skd/sk_daemon.cpp
```diff
@@ -274,9 +275,24 @@ static int createSoftKernelFile(uint64_t paddr, size_t size, uint32_t cuidx)
...
 
+static void stacktrace_logger(const int sig)
+{
+  syslog(LOG_ERR, "%s - got %d\n", __func__, sig);
+  if (sig == SIGCHLD)
+    return;
+  void * array[25];
+  int nSize = backtrace(array, 25);
+  char ** symbols = backtrace_symbols(array, nSize);
+  for (int i = 0; i < nSize; i++)
```
larry9523, line 286: You may want to check if symbols is valid in case of backtrace_symbols is running out of time to avoid crash in crash handler. :)

src/runtime_src/core/edge/skd/sk_daemon.cpp
```diff
@@ -290,6 +306,22 @@ void configSoftKernel(xclSKCmd *cmd)
...
+  sigaction(SIGBUS, &act, 0);
+  sigaction(SIGFPE, &act, 0);
+  sigaction(SIGKILL, &act, 0);
+  sigaction(SIGUSR1, &act, 0);
+  sigaction(SIGSEGV, &act, 0);
+  sigaction(SIGUSR2, &act, 0);
+  sigaction(SIGPIPE, &act, 0);
+  sigaction(SIGALRM, &act, 0);
+  sigaction(SIGTERM, &act, 0);
+
```
larry9523, line 324: If I read the code correctly, are these signal handlers registered in the parent process? I think we really want to capture those signal in the softkernel process itself, right? When softkernel process crash, will it send, say SIGSEGV, to the parent process?

larry9523 (changes requested):
This is very good. Thanks!. I was coding the similar functions while you submitted this PR.

Just a couple of questions and suggestions. Please see them in line.

src/runtime_src/core/edge/skd/sk_daemon.cpp
sonals: Reminds of XSIM simulator back trace printer I had implemented many years ago.
Please add additional checks on backtrace() and backtrace_symbols() as suggested.

src/runtime_src/core/edge/skd/sk_daemon.cpp
```diff
@@ -274,9 +275,24 @@ static int createSoftKernelFile(uint64_t paddr, size_t size, uint32_t cuidx)
...
 }
 
+static void stacktrace_logger(const int sig)
+{
+  syslog(LOG_ERR, "%s - got %d\n", __func__, sig);
+  if (sig == SIGCHLD)
+    return;
+  void * array[25];
+  int nSize = backtrace(array, 25);
+  char ** symbols = backtrace_symbols(array, nSize);
```
sonals, line 285: `const char **symbols`

src/runtime_src/core/edge/skd/sk_daemon.cpp
```diff
@@ -274,9 +275,24 @@ static int createSoftKernelFile(uint64_t paddr, size_t size, uint32_t cuidx)
   return 0;
 }
 
+static void stacktrace_logger(const int sig)
+{
+  syslog(LOG_ERR, "%s - got %d\n", __func__, sig);
+  if (sig == SIGCHLD)
+    return;
+  void * array[25];
```
sonals, line 283: Align pointer with variable name:
`void *array[25]`

sonals (changes requested):
Minor formatting changes and sanity tests for glibc backtrace APIs.

src/runtime_src/core/edge/skd/sk_daemon.cpp
rathaval: The installation of the handlers happens in the parent, and should/will carry forward to the child. In some sense this is better, since the original skd(parent) and the rest of the soft-kernels (children) will be logging backtrace incase of signals.

src/runtime_src/core/edge/skd/sk_daemon.cpp
rathaval: error: invalid conversion from 'char**' to 'const char**' [-fpermissive]
   const char **symbols = backtrace_symbols(array, nSize);
                          ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~

I did feel that this would help, not entirely sure why the petalinux compilation complains about this. Getting back char ** got it to compile again. 

In this case -fpermissive is not helping.

src/runtime_src/core/edge/skd/sk_daemon.cpp
rathaval: Fixed.

src/runtime_src/core/edge/skd/sk_daemon.cpp
rathaval: By that I mean, after the fork() the same handling will be installed on parent and child processes.

rathaval:
> 
> 
> This is very good. Thanks!. I was coding the similar functions while you submitted this PR.
> 
> Just a couple of questions and suggestions. Please see them in line.

Thanks for your review Larry :)

rathaval:
> 
> 
> Minor formatting changes and sanity tests for glibc backtrace APIs.

Thanks for your review Sonal :)

src/runtime_src/core/edge/skd/sk_daemon.cpp
larry9523: Good to know it is inherited by the child process.

Then I have a question about wait call in the child process. It does not make sense to me to call wait in the child process. I think we should call exit explicitly there.

larry9523 (commented):
Please see the in line comment about wait() in child process

src/runtime_src/core/edge/skd/sk_daemon.cpp
larry9523: My suggestion is to register different sig handler in parent and child process separately, where you don't call wait in child's handler.

rathaval:
> 
> 
> Please see the in line comment about wait() in child process

Hi Larry, I've installed 

1. Two seperate handlers -- one for the parent and a stack logging handler for the child

2.  After the child logs the stacktrace, it performs an exit() call.

larry9523 (approved):

rathaval:
Device Side Log when simulating a Segfault 

root@U30_xdma_1_1:~# kill -s SEGV 2803
Dec  3 13:37:24 U30_xdma_1_1 user.err VCU encoder kernel[2803]: sigLog - got 11
Dec  3 13:37:24 U30_xdma_1_1 user.err VCU encoder kernel[2803]: stacktrace_logger - got 11
Dec  3 13:37:24 U30_xdma_1_1 user.err VCU encoder kernel[2803]: skd(+0x1c28) [0x5593d2ec28]
Dec  3 13:37:24 U30_xdma_1_1 user.err VCU encoder kernel[2803]: linux-vdso.so.1(__kernel_rt_sigreturn+0) [0x7f8ce9c66c]
Dec  3 13:37:24 U30_xdma_1_1 user.err VCU encoder kernel[2803]: /lib/libc.so.6(ioctl+0xc) [0x7f8cd1ad1c]
Dec  3 13:37:24 U30_xdma_1_1 user.err VCU encoder kernel[2803]: /usr/lib/libxrt_core.so.2(_ZN4ZYNQ4shim11xclSKReportEj13xrt_scu_state+0x40) [0x7f8ce41d70]
Dec  3 13:37:24 U30_xdma_1_1 user.err VCU encoder kernel[2803]: skd(_Z16configSoftKernelP8xclSKCmd+0x4d4) [0x5593d2f1ec]
Dec  3 13:37:24 U30_xdma_1_1 user.err VCU encoder kernel[2803]: skd(main+0x110) [0x5593d2e9b0]
Dec  3 13:37:24 U30_xdma_1_1 user.err VCU encoder kernel[2803]: /lib/libc.so.6(__libc_start_main+0xe4) [0x7f8cc77ce4]
Dec  3 13:37:24 U30_xdma_1_1 user.err VCU encoder kernel[2803]: skd(+0x1a3c) [0x5593d2ea3c]
Dec  3 13:37:24 U30_xdma_1_1 local0.err skd[1733]: sigAct - got 17
Dec  3 13:37:24 U30_xdma_1_1 user.info kernel: [  827.632879] [drm] Pid 2803 closed device

larry9523 (approved):

sonals (approved):

PR #4005, opened 2020-09-04
Turn on HAL profiling hooks in PCIe Linux Shim by default

IshitaGhosh:
Loading of xdp_hal_plugin and actual HAL profiling feature is however still under runtime configuration xrt_profile=true in xrt.ini


jvillarre (approved):
Changes look good to enable the HAL level profiling hooks.  This enables the capability to turn on HAL API tracing, device trace for HAL applications, and the replacement power profiling plugin.

dayeh-xilinx:
retest this please

PR #4008, opened 2020-09-07
xocl: cu: minor fixes in cu driver

rajkumar-xilinx:


mamin506 (approved):
Good! Thanks for this fix.

PR #4011, opened 2020-09-07
Internal API for constructing xrt device from xclDeviceHandle

stsoe:
Added internal API to allow OpenCL to use native APIs that require xrt::device or
xrtDeviceHandles. The APIs are internal for Vitis use, but will likely
disappear as OCL implementation moves to native XRT APIs.


larry9523 (approved):

larry9523 (approved):
I've tested on lab8 OpenCL host code. It worked!

PR #3863, opened 2020-08-13
Forking processes for xbutil2

AShivangi:
**This completes the functionality implementation of xbutil2 validate**

- Created a standalone Process.h to abstract the implementation of "runPythonScript" from SubCmdValidate
- Added new nodes "Xclbin" and "Testcase" in the log for debugging purposes
- Cleanup (removed unwanted warning suppressions), restructured code and documentation 

Tests:
- 1RP  on ubuntu 16.04
- 2RP on ubuntu 16.04
- ubuntu18.04 (boost::process) 
- Not tested on windows since it requires the shell directory to be defined but the code compiles 

Sample output:
```
$ xbutil --new validate -d 0000:b3:00.1
-----------------------------------------------------
 Invoking next generation of the xbutil application
-----------------------------------------------------
Starting validation for 1 devices

Validate device[0000:b3:00.1]
Platform            : xilinx_u200_xdma_201830_3
SC Version          : 4.3.9
Platform ID         : 0x5ee83d83

1/10 Test #1         : Kernel version
    Description     : Check if kernel version is supported by XRT
    [PASSED]
-------------------------------------------------------------------------------
2/10 Test #2         : Aux connection
    Description     : Check if auxiliary power is connected
    Warning         : Aux power is not connected
    Warning         : Device is not stable for heavy acceleration tasks
    [PASSED WITH WARNINGS]
-------------------------------------------------------------------------------
3/10 Test #3         : PCIE link
    Description     : Check if PCIE link is active
    [PASSED]
-------------------------------------------------------------------------------
4/10 Test #4         : SC version
    Description     : Check if SC firmware is up-to-date
    [PASSED]
-------------------------------------------------------------------------------
5/10 Test #5         : Verify kernel
    Description     : Run 'Hello World' kernel test
    Xclbin          : /opt/xilinx/dsa/xilinx_u200_xdma_201830_3/test/verify.xclbin
    Testcase        : /opt/xilinx/xrt/test/22_verify.py
    [PASSED]
-------------------------------------------------------------------------------
6/10 Test #6         : DMA
    Description     : Run dma test
    Details         : Host -> PCIe -> FPGA write bandwidth = 9064.513845 MB/s
    Details         : Host <- PCIe <- FPGA read bandwidth = 11737.398630 MB/s
    [PASSED]
-------------------------------------------------------------------------------
7/10 Test #7         : Bandwidth kernel
    Description     : Run 'bandwidth kernel' and check the throughput
[...]

PR #4012, opened 2020-09-07
Fix building with local static Boost build

stsoe:
Linking was broken after #3863.  Make sure first find_package of Boost
locates and configures based on XRT_BOOST_INSTALL if this variable is
set.


stsoe:
Hi @AShivangi, please check that I didn't break the what you added.  Thanks!

PR #4013, opened 2020-09-08
Added support for No-dma platforms in emulation

vboggara-xilinx:
Added support for No-DMA platforms in sw and hw emulation.
1. Added copy bo support from host_only to read_only buffers and vice-versa in xclCopyBo()
2. Added device_query nodes for m2m and nodma in both sw and hw emulation.


chvamshi-xilinx (approved):
Looks good

PR #4017, opened 2020-09-09
xbutil dmatest fix

houlz0507:


maxzhen (approved):

maxzhen (changes requested):
I believe this is a better fix:
diff --git a/src/runtime_src/core/pcie/linux/scan.cpp b/src/runtime_src/core/pcie/linux/scan.cpp
index e3ff9867..c7262b11 100644
--- a/src/runtime_src/core/pcie/linux/scan.cpp
+++ b/src/runtime_src/core/pcie/linux/scan.cpp
@@ -300,6 +300,7 @@ get(const std::string& name,
   if (!err.empty())
     return;
 
+  buf.clear();
   buf.insert(std::end(buf),std::istreambuf_iterator<char>(fs),
              std::istreambuf_iterator<char>());
 }
I tested on my system and it can fix this issue. Could you give it a try?

maxzhen (approved):

PR #4018, opened 2020-09-09
CR-1074070 general protection fault (scheduler_loop)  race condition mb_scheduler

xdavidz:
This is a draft version of adding a simple lock to protect the potential race condition. But I am not able to reproduce this issue easily. Still trying to crack the issue to the ground.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
@@ -3435,6 +3435,7 @@ pending_cmds_reset(void)
  * @cores: list of execution cores (devices)
  * @intc: boolean flag set when there is a pending interrupt for command completion
  * @poll: number of running commands in polling mode
+ * @scheduler_lock: lock to protect threads access cores list
  */
 struct xocl_scheduler {
 	struct task_struct	  *scheduler_thread;
@@ -3449,6 +3450,7 @@ struct xocl_scheduler {
 
 	unsigned int		   intc; /* pending intr shared with isr, word aligned atomic */
 	unsigned int		   poll; /* number of cmds to poll */
+	struct mutex		   scheduler_lock;
 };
 
 static struct xocl_scheduler scheduler0;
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
```diff
@@ -3539,10 +3546,17 @@ scheduler_service_cores(struct xocl_scheduler *xs)
 	struct list_head *pos, *next;
 
 	SCHED_DEBUGF("-> %s\n", __func__);
+
+	mutex_lock(&xs->scheduler_lock);
 	list_for_each_safe(pos, next, &xs->cores) {
 		struct exec_core *exec = list_entry(pos, struct exec_core, core_list);
+
+		mutex_unlock(&xs->scheduler_lock);
```
maxzhen, line 3554: how can you make sure the exec is not going to be destroyed after you drop the lock here?

maxzhen (changes requested):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
xdavidz: We cannot, as of today, we don't prevent this situation too. Given the exec will be thrown into the queues, running queue, pending_queue, etc. Seems we don't have a race otherwise we should see exec being destroyed very often?

I think the reset procedure has covered most of the cases to prevent this situation. I am not sure if just add lock around line 3555 would be safe enough.

maxzhen: I thought that this is the exact race you're trying to fix, no? If you drop lock here, all your effort goes in vain, right? Can we keep the lock here?

maxzhen (approved):

houlz0507 (approved):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
xdavidz: After an offline chat, we agreed that keep lock to protect this whole operation. Two additional tests to be done:
1) wait for pipeline to finish
2) I have a specific test procedure which can trigger the scheduler_loop panic sooner(within 10~20 reset cycles), will verify the panic is reproducible without the fix, and cannot be seen with the fix.

mamin506 (approved):

stsoe (approved):

xuhz (commented):
I believe the following should also be protected by the lock
mb_scheduler_remove()
{
...
exec_reset_cmds(exec);
...
}

xuhz (approved):

xdavidz:
retest this please

dayeh-xilinx:
retest this please

PR #4019, opened 2020-09-09
Purge LD_LIBRARY_PATH from Vitis and Vivado paths for XRT tools

Ralender:
Maybe this doesn't need to be gated by Distro and Version.

### DIFF
--- a/src/runtime_src/tools/scripts/loader
+++ b/src/runtime_src/tools/scripts/loader
@@ -48,6 +48,11 @@ fi
 
 . "${XRT_SETUP_SCRIPT}" > /dev/null
 
+# XRT uses system librairies whereas Vitis and Vivado ship there own.
+# when XRT tools are invoked from Vitis/Vivado the LD_LIBRARY_PATH will be set to use libraries shipped with Vitis/vivado
+# so we reset the LD_LIBRARY_PATH to make XRT tool select the correct libraries.
+LD_LIBRARY_PATH=$XILINX_XRT/lib:$(echo $LD_LIBRARY_PATH | tr ':' '\n' | grep -E -v "/(Vitis|Vivado).*/(lnx|lin)" | tr '\n' ':')
+
 # -- Execute the wrapped program
 "${XRT_PROG_UNWRAPPED}" "${XRT_LOADER_ARGS[@]}"
 


keryell:
Good question.
@stsoe @sonals Is there a case where XRT tools have to run using libraries shipped by Vitis/Vivado?

stsoe:
No, never.  In fact mixing in LD_LIBRARY_PATH pointing to Vitis is bound to cause errors when using XRT.

src/runtime_src/tools/scripts/setup.sh
```diff
@@ -12,6 +12,12 @@ if [[ $OSDIST == "ubuntu" ]]; then
         echo "ERROR: Ubuntu release version must be 16.04 or later"
         return 1
     fi
+    if (( $OSREL > 1804 )); then
+        # XRT uses system librairies whereas Vitis and Vivado ship there own.
+        # when XRT tools are invoked from Vitis/Vivado the LD_LIBRARY_PATH will be set to use libraries shipped with Vitis/vivado
+        # so we purge LD_LIBRARY_PATH of all path within Vitis/vivado to make XRT tool select the correct libraries.
+        LD_LIBRARY_PATH=$(echo $LD_LIBRARY_PATH | awk -v RS=: -v ORS=: '/Vitis/ {next} {print}' | awk -v RS=: -v ORS=: '/Vivado/ {next} {print}' )
+    fi
```
stsoe, line 20: Which part of XRT uses libraries from Vitis?
I don't think this should be added to XRT's setup script.

stsoe (changes requested):

src/runtime_src/tools/scripts/setup.sh
keryell: Obviously it is a DevOps bug in Vitis like almost any other tool that Xilinx ships but it looks like nobody cares... :-(
I do not really see what you mean by a "user error" if the problem is among Xilinx developers...
But Vitis is close-source while XRT is open-source, so at the end it is cleaner/simpler to put some antidote in XRT. :-)
In the past I have used file-system layering on top of `/proj/xbuilds` to patch some nightly builds or I directly patched some installed tools in `/opt/xilinx` or @Ralender has copied some part of `/proj/xbuilds` and added some spaghetti `ln -s` icing on top of it (this does not requires `root` access :-) ), so at the end I suggested @Ralender to propose this low-hanging antidote in XRT.
Perhaps it is simpler to fork XRT if this is not accepted here than trying to unsuccessfully change Xilinx DevOps methods... :-/

stsoe: I mean that we simply can't go about clearing elements from LD_LIBRARY_PATH arbitrarily thinking that we own every path containing the string "Vitis" or "Vivado".   I  object to removing elements from caller (user) defined paths in a  setup script.   

I also don't quite understand what this change will accomplish in terms of the stated problem (comment). If XRT is used from Vitis/Vivado how does XRT's setup script end up being called as part of this usage?  The setup scripts for Vitis or Vivado do not change LD_LIBRARY_PATH.   Invoking Vivado/Vitis is through a loader which  sets LD_LIBRARY_PATH for the duration of process, presumable this process won't source XRT's setup.sh (will the process even use XRT?).

keryell: Actually we do not really use the setup script directly, because it is not really necessary (or in my case because I cannot, I do not use `bash` but `zsh` and this setup script is full of `bash`-ism).
The problem is that all the XRT tools are actually wrapped by a `/opt/xilinx/xrt/bin/unwrapped/loader` `bash` script which ends with
```bash
# -- Find the setup script and configure environment
XRT_SETUP_SCRIPT="`dirname \"$0\"`/../../setup.sh"

if [ ! -f "$XRT_SETUP_SCRIPT" ]; then
  echo "ERROR: Could not find XRT setup script."
  echo "ERROR: ${XRT_SETUP_SCRIPT} does not exist"
  exit 1
fi

. "${XRT_SETUP_SCRIPT}" > /dev/null

# -- Execute the wrapped program
"${XRT_PROG_UNWRAPPED}" "${XRT_LOADER_ARGS[@]}"
```
Thus at the ends, every time an XRT command is launched, the setup script is sourced.
So, why not adding an `LD_LIBRARY_PATH` sanitizer there? It fits nicely all the design patterns in use at the first place... ;-)
For some reasons `v++` with its atrocious` LD_LIBRARY_PATH` uses some XRT commands and it breaks.

stsoe: This is a good idea indeed.   Let's sanitize the loader script, it makes perfect sense since its purpose is to establish the env needed for the tools.  Great!

keryell: @Ralender The regex on the paths in `LD_LIBRARY_PATH` should be more precise to limit the risk to remove useful library path.

Plausible use case at Xilinx: someone install Homebrew to have some modern environment and decide to do it inside a `Vitis` directory because at the end this is to experiment with... Vitis. Then a lot of shared libraries will be in a directory with `Vitis` in the name with some stuff useful for XRT, such as Boost... ;-)

keryell: @stsoe so, if I understand correctly, what you propose is to do the clean up in `unwrapped/loader` directly instead of in `setup.sh`?
It is probably cleaner in the sense that then a user can also source fist the Vitis `bash` setup, then source the XRT `bash` setup without breaking the Vitis configuration.

stsoe (commented):
Yes, I thought that was what you suggested yesterday.  I liked it.  I think the loader can simply reset LD_LIBRARY_PATH to $XILINX_XRT/lib, there is no reason for the loader to use anything from the callers environment, right?

src/runtime_src/tools/scripts/setup.sh
stsoe: Yes, I thought that was what you suggested yesterday. I liked it. I think the loader can simply reset LD_LIBRARY_PATH to $XILINX_XRT/lib, there is no reason for the loader to use anything from the callers environment, right?

keryell: Except if you are on a system where you need a lot of `LD_LIBRARY_PATH` set-up to run...: -(
For example you work in a company where all the default software is so obsolete that you need to install a recent compiler, libraries such as Boost, etc. without breaking the existing system software. In that case XRT would probably need to use this library path.
Of course this is purely imaginative, there is no company that would run obsolete software on purpose. This is just a conspiracy theory where the IT and DevOps are actually operated by your competitors to keep you behind, right?... :-)

stsoe: haha. sure yes, I did think about the case where XRT was built against non system libraries and you are right this will break the tools if we just reset the path.  I am not sure what we should do then.   Filtering out vivado/vitis is better in loader than in setup.sh, but IMO still evil.

src/runtime_src/tools/scripts/loader
```diff
@@ -48,6 +48,11 @@ fi
 
 . "${XRT_SETUP_SCRIPT}" > /dev/null
 
+# XRT uses system librairies whereas Vitis and Vivado ship there own.
+# when XRT tools are invoked from Vitis/Vivado the LD_LIBRARY_PATH will be set to use libraries shipped with Vitis/vivado
+# so we reset the LD_LIBRARY_PATH to make XRT tool select the correct libraries.
+LD_LIBRARY_PATH=$XILINX_XRT/lib
+
```
keryell, line 55: Making progress.
But I would prefer you make more precise your previous LD_LIBRARY_PATH patching solution and you move it here.

src/runtime_src/tools/scripts/setup.sh
Ralender: > Except if you are on a system where you need a lot of `LD_LIBRARY_PATH` set-up to run...: -(
> For example you work in a company where all the default software is so obsolete that you need to install a recent compiler, libraries such as Boost, etc. without breaking the existing system software. In that case XRT would probably need to use this library path.

a solution maybe to use the LIBRARY_PATH variable if present which used by Vitis/Vivado for that purpose.
so in @keryell very theoretical example one would have to set both LD_LIBRARY_PATH and LIBRARY_PATH.
LIBRARY_PATH already has to be set correctly to use Vitis anyway.

src/runtime_src/tools/scripts/loader
keryell: I think you still need to remove offensive Vitis content from `$LIBRARY_PATH` as you did first. Just need a filtering targeting better Vivado & Vitis.

keryell (approved):
This looks more robust that way.
Thanks.

stsoe (approved):

PR #4020, opened 2020-09-09
Fixes failing tests in sycl

Ralender:
buffers provided to clEnqueueCopyBufferRect may not yet be associated with a device.

the default flag isn't 0xffffffff but XOCL_MEM_BANK_MSK which is 0xffffff.


keryell:
> buffers provided to clEnqueueCopyBufferRect may not yet be associated with a device.
> 
> the default flag isn't 0xffffffff but XOCL_MEM_BANK_MSK which is 0xffffff.

I guess this should be part of the commit message. The PR will disappear, not the commit message...

stsoe (approved):

chvamshi-xilinx (approved):

PR #4027, opened 2020-09-10
CR-1074939 DMESG reports MIG calibration took 0s but the time delta for the step is longer than this

chienwei-lan:
We should just not print out the calibration time at all.


dayeh-xilinx:
retest this please

xdavidz (approved):

PR #4028, opened 2020-09-10
CR-1066313 Default value of XILINX_XRT=/usr for edge platforms

rbramand-xilinx:
> Setting XILINX_XRT=/usr if env variable is not set
> previously runtime error was thrown but changed this behavior on edge platforms to set "/usr" as default value

### DIFF
--- a/src/runtime_src/xrt/device/hal.cpp
+++ b/src/runtime_src/xrt/device/hal.cpp
@@ -175,11 +175,11 @@ loadDevices()
   // xrt
   bfs::path xrt(emptyOrValue(getenv("XILINX_XRT")));
 
-  if (xrt.empty()){
-    #if defined (__aarch64__) || defined (__arm__)
-        xrt /= "/usr";
-    #endif
+#if defined (__aarch64__) || defined (__arm__)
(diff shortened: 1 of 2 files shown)


src/runtime_src/xrt/device/hal.cpp
```diff
@@ -175,6 +175,12 @@ loadDevices()
   // xrt
   bfs::path xrt(emptyOrValue(getenv("XILINX_XRT")));
 
+  if (xrt.empty()){
+    #if defined (__aarch64__) || defined (__arm__)
```
chvamshi-xilinx, line 179: better to move this condition before xrt.empty() check

src/runtime_src/xrt/device/hal.cpp
```diff
@@ -175,6 +175,12 @@ loadDevices()
...
+        xrt /= "/usr";
```
chvamshi-xilinx, line 180: xrt = "/usr"

chvamshi-xilinx (approved):

stsoe (approved):

PR #4031, opened 2020-09-11
CR-1071356 Integrate verify, bandwidth and slave bridge git examples in xbutil validate

AShivangi:
- Cases handled:
```
            |     Old XRT       |    New XRT
Old shell   |   Runs python     |  Runs python
New shell   |  Throws an error  |  Runs cpp
```
- Versal will always run py test case since we don't have the cpp equivalent
- Todo: add to xbutil2

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
+++ b/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
@@ -1197,8 +1197,7 @@ int xcldev::device::runTestCase(const std::string& py,
     // if platform.json exists in the new shell pakcage then use the new testcase
     // else fallback to old python test cases
     
-    // NEW FLOW: runs if platform.json is available and the platform is not versal
-    // Note: we don't have cpp based versal bw testcase
+    // NEW FLOW: runs if platform.json is available
     auto json_exists = [xclbinPath]() { return boost::filesystem::exists(xclbinPath + "platform.json") ? true : false; };
     
     if(json_exists()) {    
--- a/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
+++ b/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
@@ -1186,37 +1186,70 @@ int xcldev::device::runTestCase(const std::string& py,
     std::string devInfoPath = name + "/test/";
     std::string xsaXclbinPath = getXsaPath(vendor) + devInfoPath;
     std::string dsaXclbinPath = dsaPath + devInfoPath;
-    std::string xrtTestCasePath = xrtPath + "test/" + py;
+    std::string xrtTestCasePath = xrtPath + "test/";
 
     output.clear();
 
     std::string xclbinPath;
     searchXsaAndDsa(m_idx, xsaXclbinPath, dsaXclbinPath, xclbinPath, output);
-    xclbinPath += xclbin;
-
-    if (stat(xrtTestCasePath.c_str(), &st) != 0 || stat(xclbinPath.c_str(), &st) != 0) {
-        //if bandwidth xclbin isn't present, skip the test
-        if(xclbin.compare("bandwidth.xclbin") == 0) {
-            output += "Bandwidth xclbin not available. Skipping validation.";
-            return -EOPNOTSUPP;
+    std::string cmd;
+    
+    // if platform.json exists in the new shell pakcage then use the new testcase
+    // else fallback to old python test cases
+    
+    // NEW FLOW: runs if platform.json is available and the platform is not versal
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -1186,37 +1186,70 @@ int xcldev::device::runTestCase(const std::string& py,
...
+    // if platform.json exists in the new shell pakcage then use the new testcase
+    // else fallback to old python test cases
+    
+    // NEW FLOW: runs if platform.json is available and the platform is not versal
+    // Note: we don't have cpp based versal bw testcase
+    auto json_exists = [xclbinPath]() 
+        {
+            return boost::filesystem::exists(xclbinPath + "platform.json") ? true : false;
+        };
+    if(json_exists() && xclbin.compare("versal_23_bandwidth.py") != 0) {
```
maxzhen, line 1206: We can have platform.json for versal platform?

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -1186,37 +1186,70 @@ int xcldev::device::runTestCase(const std::string& py,
...
+        
+        //map old testcase names to new testcase names
+        static const std::map<std::string, std::string> test_map = {
+            { "22_verify.py",             "validate.exe"    },
+            { "23_bandwidth.py",          "kernel_bw.exe"   },
+            { "host_mem_23_bandwidth.py", "slavebridge.exe" }
+        };
+        
+        xrtTestCasePath += test_map.find(py)->second;
+        // in case the user is trying to run a new platform with an older XRT
```
maxzhen, line 1216: Could you clarify "older XRT“？

maxzhen (changes requested):

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
AShivangi: I don't think Heera's team has bw testcase for versal yet. We'll have to rely on python testcase until they add the versal bw

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
AShivangi: By older XRT I mean any XRT before this PR, but I'll update the comment reflecting this

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
maxzhen: do you mean that they can have verify kernel for versal, but no bandwidth for versal?

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
maxzhen: > 
> 
> By older XRT I mean any XRT before this PR, but I'll update the comment reflecting this

you pr will make "older XRT" become "new XRT", so how do you handle "older XRT" case in your pr?

AShivangi: So if we have a new shell (with platform.json) with an older XRT (without the .exe), then we'll throw an error saying: 
`ERROR: Failed to find verify.exe. Please update XRT.`

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
AShivangi: Oh, I guess I'm mistaken. I don't think that verify kernel for versal works either. We should check and see if they are providing platform.json for versal. If not then  `xclbin.compare("versal_23_bandwidth.py") != 0` is redundant

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -1186,37 +1186,67 @@ int xcldev::device::runTestCase(const std::string& py,
...
-        output += ", Shell package not installed properly.";
-        return -ENOENT;
-    }
+    std::string cmd;
+    
+    // if platform.json exists in the new shell pakcage then use the new testcase
+    // else fallback to old python test cases
+    
+    // NEW FLOW: runs if platform.json is available and the platform is not versal
+    // Note: we don't have cpp based versal bw testcase
```
maxzhen, line 1201: I don't think you want to call out versal. there is no difference than other platforms here.

maxzhen (changes requested):

maxzhen (approved):

PR #4039, opened 2020-09-12
ERT 3.0 Firmware compilation support for emulation

hemantk-xilinx:
Adding support for compiling ERT 3.0 firmware for Emulation


maxzhen:
ok to test

chienwei-lan (approved):

src/runtime_src/ert/scheduler/CMakeLists.txt
```diff
@@ -153,3 +153,29 @@ install(TARGETS sched_em
   ARCHIVE DESTINATION ${XRT_INSTALL_LIB_DIR} COMPONENT ${XRT_DEV_COMPONENT}
   LIBRARY DESTINATION ${XRT_INSTALL_LIB_DIR} COMPONENT ${XRT_DEV_COMPONENT} ${XRT_NAMELINK_ONLY}
 )
+
+
+file(GLOB SCH_SRC_FILES
+  "${CMAKE_CURRENT_SOURCE_DIR}/scheduler_v30.cpp"
+  )
+
+add_definitions(-DERT_HW_EMU -DERT_BUILD_V30  -DXCLHAL_MAJOR_VER=1 -DXCLHAL_MINOR_VER=0)
```
stsoe, line 162: I am wary of global add_definition, can we be sure it only affects following targets?  And if it does then it would be sufficient to just add -DERT_BUILD_V30 since preceeding add_defintions has the other ones.

Can you maybe use 
https://cmake.org/cmake/help/latest/command/target_compile_definitions.html

stsoe (commented):
Just a comment, please consider.  Thanks!

src/runtime_src/ert/scheduler/CMakeLists.txt
hemantk-xilinx: Fixed Please check

PR #4041, opened 2020-09-13
Enable builds on Ubuntu 20.04

sonals:


src/runtime_src/tools/scripts/xrtdeps.sh
```diff
@@ -50,6 +50,23 @@ while [ $# -gt 0 ]; do
     esac
 done
 
+dump_host_environment()
+{
+    echo "Build environment"
+    echo "---------------------------------------------------------------------------------"
+    if [ -x /usr/bin/systemd-detect-virt ]; then
+	echo -n "Virtualization: "
```
keryell, line 58: Avoid mixing spaces and tabulations

PR #4047, opened 2020-09-14
Fixes in Device Event Creator

IshitaGhosh:
1. Fix for crash for certain Stream Transfer test
2. Fix for missing Stream Transfers when Stream Starve or Stall events re present
3. Fix for missing AXI MM transfers as reported in http://jira.xilinx.com/browse/CR-1075956


jvillarre (approved):
Changes look good.  The trace IDs for some stream events can be repeated for different types of events, so this added identifier is necessary.

dayeh-xilinx:
retest this please

PR #4049, opened 2020-09-15
Fix aie reset/reload

larry9523:
This PR added the following support
1) xrtResetAIEArray
2) Load AIE only XCLBIN after reset
3) Prevent accessing AIE resources after reset if an AIE only XCLBIN (AIE PDI) is not loaded.

### DIFF
--- a/src/runtime_src/core/edge/user/aie/aie.cpp
+++ b/src/runtime_src/core/edge/user/aie/aie.cpp
@@ -284,4 +284,23 @@ clear_bd(BD& bd)
 #endif
 }
 
+void
+Aie::
+reset(const std::shared_ptr<xrt_core::device>& device)
+{
+#ifndef __AIESIM__
+    int rval = close(fd);
+
+    auto drv = ZYNQ::shim::handleCheck(device->get_device_handle());
+
+    /* TODO get partition id and uid from XCLBIN or PDI */
+    uint32_t partition_id = 1;
+
+    drm_zocl_aie_reset reset = { partition_id };
+    int ret = drv->resetAIEArray(reset);
+    if (ret)
+        throw xrt_core::error(ret, "Fail to reset AIE Array");
+#endif
+}
+
 }
--- a/src/runtime_src/core/edge/user/aie/aie.cpp
+++ b/src/runtime_src/core/edge/user/aie/aie.cpp
@@ -59,7 +59,7 @@ Aie::Aie(const std::shared_ptr<xrt_core::device>& device)
     AieRC rc;
     if ((rc = XAie_CfgInitialize(&DevInst, &ConfigPtr)) != XAIE_OK)
         throw xrt_core::error(-EINVAL, "Failed to initialize AIE configuration: " + std::to_string(rc));
-    devInst = DevInst;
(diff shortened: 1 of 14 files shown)


mamin506 (approved):

src/runtime_src/core/edge/user/aie/aie.cpp
```diff
@@ -284,4 +284,23 @@ clear_bd(BD& bd)
 #endif
 }
 
+void
+Aie::
+reset(const std::shared_ptr<xrt_core::device>& device)
```
stsoe, line 289: My preference is to use "const xrt_core_device*" since the shared_ptr is not used at all and the device is in scope for the duration of the call.

src/runtime_src/core/edge/user/aie/aie.h
```diff
@@ -93,6 +93,9 @@ class Aie {
     void
     wait_gmio(const std::string& gmioName);
 
+    void
+    reset(const std::shared_ptr<xrt_core::device>& device);
```
stsoe, line 97: See above comment

src/runtime_src/core/edge/user/aie/graph.cpp
```diff
@@ -745,9 +745,17 @@ xclGMIOWait(xclDeviceHandle handle, const char *gmioName)
...
-  /* Do a handle check */
-  xrt_core::get_userpf_device(handle);
+#ifndef __AIESIM__
+  auto device = xrt_core::get_userpf_device(handle);
+  auto drv = ZYNQ::shim::handleCheck(device->get_device_handle());
 
+  if (!drv->isAieRegistered())
+    throw xrt_core::error(-EINVAL, "No AIE presented");
+  auto aieArray = drv->getAieArray();
+  aieArray->reset(device);
```
stsoe, line 755: reset(device.get())

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -1449,6 +1447,17 @@ getPartitionFd(drm_zocl_aie_fd &aiefd)
...
+
+int
+shim::
+resetAIEArray(drm_zocl_aie_reset &reset)
+{
+  int ret = ioctl(mKernelFD, DRM_IOCTL_ZOCL_AIE_RESET, &reset);
+  if (ret)
+    return -errno;
+
+  return 0;
```
stsoe, line 1459:
```
return ioctl(...) ? -errno : 0;
```

stsoe (changes requested):

src/runtime_src/core/edge/user/aie/aie.cpp

src/runtime_src/core/edge/user/aie/aie.h

src/runtime_src/core/edge/user/aie/graph.cpp
larry9523: done

src/runtime_src/core/edge/user/shim.cpp
larry9523: done

stsoe (approved):

PR #4057, opened 2020-09-15
Make sure xclProbe() is called when in emulation mode

stsoe:
With new XRT native APIs, xclProbe() is folded into the implementation
and no longer called from host application.  Emulation sets up data
structures as part of xclProbe() so make sure it is called when the
core library is loaded.

See also CR-1076331


stsoe:
@uday610 FYI.  This PR should eliminate the need to call xclProbe() explicitly in emulation (I believe you mentioned you had to do this?)

PR #4061, opened 2020-09-16
 CR-1074448 ASTeR Basic sanity - xbutil dump JSON format broken due to error message printed in output u250 and u50

xdavidz:
@AShivangi  this is need your attention about xbutil future enhancement. I left comments in CR: http://jira.xilinx.com/browse/CR-1074448

Basically, we might clean up existing xbutil or design a better way to handle some cases where we want to return errors instead of silently print errors into output, or we want to skip reading one sysfs node but continue to read next sysfs node.

Thanks,
David


AShivangi (approved):
Looks good. Thanks for making the changes. 
We are being more careful about this in xbutil2. We are throwing exceptions whenever necessary and redirecting errors to cerr.

maxzhen (approved):

PR #4066, opened 2020-09-16
Use xrt APIs in place of xcl APIs for buffer handling

IshitaGhosh:
The xrt APIs for buffer allocation, sync, free like xrtBOAlloc, xrtBOMap, xrtBOSync, xrtBOFree handle platforms with or without DMA. So, replace use of xcl APIs like xclAllocBO, xclSyncBO, xclFreeBO with the new xrt APIs


jvillarre (approved):
Changes look good.

PR #4068, opened 2020-09-16
CR-1074971 ASTeR 1.05 - y/n prompt when plp program executed while fa…

xdavidz:
…ctory image is in use

Just avoid to prompt the y/n.

Maybe, just return fail directly is better?


maxzhen:
> 
> 
> …ctory image is in use
> 
> Just avoid to prompt the y/n.
> 
> Maybe, just return fail directly is better?

I agree that we probably want to fail fast and give out clear error msg to indicate that we can't support download *LP on card running golden image.

dayeh-xilinx:
retest this please

maxzhen (approved):

houlz0507 (approved):

PR #4074, opened 2020-09-17
Added zocl driver support for LPDDR

saifuddin-xilinx:
We have added the LPDDR support for Zocl driver.
 -- Added a new macro for LPDDR 
-- drm mem initialization for LPDDR case

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
@@ -159,11 +159,12 @@ zocl_create_bo(struct drm_device *dev, uint64_t unaligned_size, u32 user_flags)
 
 		bo = to_zocl_bo(&cma_obj->base);
 	} else {
-		/* We are allocating from a separate BANK, i.e. PL-DDR */
+		/* We are allocating from a separate BANK, i.e. PL-DDR or LPDDR */
 		unsigned int bank = GET_MEM_BANK(user_flags);
 
 		if (bank >= zdev->num_mem || !zdev->mem[bank].zm_used ||
-		    zdev->mem[bank].zm_type != ZOCL_MEM_TYPE_PLDDR)
+		    (zdev->mem[bank].zm_type != ZOCL_MEM_TYPE_PLDDR &&
+		    zdev->mem[bank].zm_type != ZOCL_MEM_TYPE_LPDDR)) 
 			return ERR_PTR(-EINVAL);
 
 		bo = kzalloc(sizeof(struct drm_zocl_bo), GFP_KERNEL);
@@ -202,8 +203,7 @@ zocl_create_bo(struct drm_device *dev, uint64_t unaligned_size, u32 user_flags)
 		 * not establish the kernel mapping. We just can not
 		 * access BO directly from kernel.
 		 */
-		if (zdev->mem[bank].zm_type != ZOCL_MEM_TYPE_LPDDR)
-			bo->vmapping = memremap(bo->mm_node->start, size, MEMREMAP_WC);
+		bo->vmapping = memremap(bo->mm_node->start, size, MEMREMAP_WC);
 
 		err = drm_gem_create_mmap_offset(&bo->gem_base);
 		if (err) {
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -202,7 +202,8 @@ zocl_create_bo(struct drm_device *dev, uint64_t unaligned_size, u32 user_flags)
 		 * not establish the kernel mapping. We just can not
 		 * access BO directly from kernel.
 		 */
-		bo->vmapping = memremap(bo->mm_node->start, size, MEMREMAP_WC);
+		if (zdev->mem[bank].zm_type != ZOCL_MEM_TYPE_LPDDR)
```
larry9523, line 205: Any specific reason we can not map a virtual address on LPDDR? Without this, xclBOWrite/xclBORead won't work

larry9523 (changes requested):
Please help me understand why we can not set up a virtual mapping of LPDDR.
Other than that, if we can treat LPDDR and PLDDR equally, I don't think it is necessary  to separate ZOCL_MEM_TYPE_PLDDR and ZOCL_MEM_TYPE_LPDDR. We can find one better name cover these two memory type.
Otherwise, please search all code which is using PLDDR and make enhancement according, e.g. zocl_update_mem_stat, zocl_create_bo, etc.

dayeh-xilinx:
retest this please

sonals (commented):
Are we sure that LPDDR cannot be used as CMA backend by Linux with proper command line like the following:
`cma=nn[MG]@[start[MG][-end[MG]]]`

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
chvamshi-xilinx: > Please help me understand why we can not set up a virtual mapping of LPDDR.
> Other than that, if we can treat LPDDR and PLDDR equally, I don't think it is necessary to separate ZOCL_MEM_TYPE_PLDDR and ZOCL_MEM_TYPE_LPDDR. We can find one better name cover these two memory type.
> Otherwise, please search all code which is using PLDDR and make enhancement according, e.g. zocl_update_mem_stat, zocl_create_bo, etc.

Hi @larry9523 , We should have separate tags for LPDDR and PLDDR as they  are two different types of memories. For example, we need to have cache enabled for LPDDR. we need a clear differentiation.
@saifuddin-xilinx , Please search the PLDDR in the code and make necessary changes

chvamshi-xilinx:
> Are we sure that LPDDR cannot be used as CMA backend by Linux with proper command line like the following:
> `cma=nn[MG]@[start[MG][-end[MG]]]`

Hi @sonals , even if we have a mechanism to extend CMA, This is not going to solve our use case
XRT should have a mechanism to allocate memory in DDR/LPDDR based on user input.  Even if we enable extra CMA, we dont have an option to allocate memory on that CMA
Thatswhy, we are doing based on  SPTAGS.

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
larry9523: @chvamshi-xilinx separate tag is OK. I am saying we can unify the behavior on both tags. The only difference I see is we are not doing memremap on LPDDR, which I don't quite understand why.

chvamshi-xilinx: Yes @larry9523 , I Agree. @saifuddin-xilinx , We need to debug why memremap is not working. We also need to enable Cache for LPDDR.

saifuddin-xilinx: After changing in dtsi file for LPDDR region as "no-map" the memremap issue resolved. Here I have updated zocl code for LPDDR case. And will send a separate PR for plarform changes.

chvamshi-xilinx (approved):

larry9523 (approved):

sonals (approved):

PR #4077, opened 2020-09-20
xbutil2 updates to report machine model, number of CPU cores as part …

sonals:
…of examine host

### DIFF
--- a/src/runtime_src/core/edge/user/system_linux.cpp
+++ b/src/runtime_src/core/edge/user/system_linux.cpp
@@ -20,6 +20,7 @@
 #include "core/common/time.h"
 
 #include <boost/property_tree/ini_parser.hpp>
+#include <boost/format.hpp>
 
 #include <fstream>
 #include <memory>
--- a/src/runtime_src/core/edge/user/system_linux.cpp
+++ b/src/runtime_src/core/edge/user/system_linux.cpp
@@ -31,13 +31,13 @@
 #include <unistd.h>
 
 #if defined(__aarch64__) || defined(__arm__) || defined(__mips__)
-  #define MACHINE_NODE_PATH "/proc/device-tree/model";
+  #define MACHINE_NODE_PATH "/proc/device-tree/model"
 #elif defined(__PPC64__)
-  #define MACHINE_NODE_PATH "/proc/device-tree/model-name";
+  #define MACHINE_NODE_PATH "/proc/device-tree/model-name"
   // /proc/device-tree/system-id may be 000000
   // /proc/device-tree/model may be 00000
 #elif defined (__x86_64__)
-  #define MACHINE_NODE_PATH "/sys/devices/virtual/dmi/id/product_name";
+  #define MACHINE_NODE_PATH "/sys/devices/virtual/dmi/id/product_name"
 #else
 #error "Unsupported platform"
   #define MACHINE_NODE_PATH "";
@@ -104,9 +104,8 @@ glibc_info()
 static std::string
 machine_info()
 {
-  static const char node[] = MACHINE_NODE_PATH;
   std::string model("unknown");
-  std::ifstream stream(node);
+  std::ifstream stream(MACHINE_NODE_PATH);
   if (stream.good()) {
     std::getline(stream, model);
     stream.close();
@@ -196,7 +195,7 @@ get_mgmtpf_device(device::id_type id) const
 
 void
 system_linux::
-program_plp(std::shared_ptr<device> dev, std::vector<char> buffer) const
+program_plp(std::shared_ptr<device> dev, const std::vector<char> &buffer) const
 {
   throw std::runtime_error("plp program is not supported");
 }
--- a/src/runtime_src/core/edge/user/system_linux.cpp
+++ b/src/runtime_src/core/edge/user/system_linux.cpp
@@ -29,6 +29,19 @@
 #include <gnu/libc-version.h>
 #include <unistd.h>
 
+#if defined(__aarch64__) || defined(__arm__) || defined(__mips__)
+  #define MACHINE_NODE_PATH "/proc/device-tree/model";
+#elif defined(__PPC64__)
+  #define MACHINE_NODE_PATH "/proc/device-tree/model-name";
+  // /proc/device-tree/system-id may be 000000
+  // /proc/device-tree/model may be 00000
+#elif defined (__x86_64__)
+  #define MACHINE_NODE_PATH "/sys/devices/virtual/dmi/id/product_name";
+#else
+#error "Unsupported platform"
+  #define MACHINE_NODE_PATH "";
+#endif
+
 namespace {
 
 // Singleton registers with base class xrt_core::system
@@ -90,18 +103,7 @@ glibc_info()
 static std::string
 machine_info()
 {
-#if defined(__aarch64__) || defined(__arm__) || defined(__mips__)
-  const char node[] = "/proc/device-tree/model";
-#elif defined(__PPC64__)
-  const char node[] = "/proc/device-tree/model-name";
-  // /proc/device-tree/system-id may be 000000
-  // /proc/device-tree/model may be 00000
-#elif defined (__x86_64__)
-  const char node[] = "/sys/devices/virtual/dmi/id/product_name";
-#else
-#error "Unsupported platform"
-  const char node[] = "";
-#endif
+  static const char node[] = MACHINE_NODE_PATH;
   std::string model("unknown");
   std::ifstream stream(node);
   if (stream.good()) {
@@ -146,7 +148,7 @@ get_os_info(boost::property_tree::ptree &pt)
   }
   pt.put("model", machine_info());
   pt.put("cores", std::thread::hardware_concurrency());
-  pt.put("memory", sysconf(_SC_PHYS_PAGES) * sysconf(_SC_PAGE_SIZE) / 0x100000);
+  pt.put("memory_bytes", (boost::format("0x%lx") % (sysconf(_SC_PHYS_PAGES) * sysconf(_SC_PAGE_SIZE))).str());
   pt.put("now", xrt_core::timestamp());
 }
 
--- a/src/runtime_src/core/edge/user/system_linux.cpp
+++ b/src/runtime_src/core/edge/user/system_linux.cpp
@@ -23,6 +23,7 @@
 
 #include <fstream>
 #include <memory>
+#include <thread>
 
 #include <sys/utsname.h>
 #include <gnu/libc-version.h>
@@ -65,7 +66,7 @@ driver_version(const std::string& driver)
 
 namespace xrt_core {
 
-void 
+void
 system_linux::
 get_xrt_info(boost::property_tree::ptree &pt)
 {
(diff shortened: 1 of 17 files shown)


AShivangi (approved):

sonals:
retest this please

src/runtime_src/core/edge/user/system_linux.cpp
```diff
@@ -118,17 +144,19 @@ get_os_info(boost::property_tree::ptree &pt)
     }
     ifs.close();
   }
-
+  pt.put("model", machine_info());
+  pt.put("cores", std::thread::hardware_concurrency());
+  pt.put("memory", sysconf(_SC_PHYS_PAGES) * sysconf(_SC_PAGE_SIZE) / 0x100000);
```
rozumx, line 149: After reviewing the code base, I have come full circle to this entry.  Could you please do the following:

1. Change the "memory" entry name to "memory_bytes".
2. Remove the 0x100000 magic values.   It is best to keep the internal values in bytes, like we do with the other values.
3.  Add this value as a hex string value.  For example
`pt.put("memory_bytes", (boost::format("0x%x") %  sysconf(_SC_PHYS_PAGES) * sysconf(_SC_PAGE_SIZE).str().c_str())`

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -133,6 +157,10 @@ get_os_info(boost::property_tree::ptree &pt)
       }
       ifs.close();
   }
+
+  pt.put("model", machine_info());
+  pt.put("cores", std::thread::hardware_concurrency());
+  pt.put("memory", sysconf(_SC_PHYS_PAGES) * sysconf(_SC_PAGE_SIZE) / 0x100000);
```
rozumx, line 163: Same comment as comment regarding changing the JSON property name and how the value is formatted.

src/runtime_src/core/pcie/windows/system_windows.cpp
```diff
@@ -130,6 +131,17 @@ get_os_info(boost::property_tree::ptree &pt)
...
   pt.put("distribution", value);
+
+  BufferSize = sizeof value;
+  RegGetValueA(HKEY_LOCAL_MACHINE, "SYSTEM\\CurrentControlSet\\Control\\SystemInformation", "SystemProductName", RRF_RT_ANY, NULL, (PVOID)&value, &BufferSize);
+  pt.put("model", value);
+
+  MEMORYSTATUSEX mem;
+  mem.dwLength = sizeof(mem);
+  GlobalMemoryStatusEx(&mem);
+  pt.put("memory", mem.ullTotalPhys / 0x100000);
```
rozumx, line 142: Same comment as comment regarding changing the JSON property name and how the value is formatted.

src/runtime_src/core/tools/common/ReportHost.cpp
```diff
@@ -70,16 +70,18 @@ ReportHost::writeReport(const xrt_core::device * _pDevice,
     _output << boost::format("  %-20s : %s\n") % "Release" % _pt.get<std::string>("host.os.release");
     _output << boost::format("  %-20s : %s\n") % "Version" % _pt.get<std::string>("host.os.version");
     _output << boost::format("  %-20s : %s\n") % "Machine" % _pt.get<std::string>("host.os.machine");
+    _output << boost::format("  %-20s : %s\n") % "CPU Cores" % _pt.get<std::string>("host.os.cores");
+    _output << boost::format("  %-20s : %s MB\n") % "Memory" % _pt.get<std::string>("host.os.memory");
```
rozumx, line 74: O.K. - I now see that we are dealing with MBytes.

rozumx: Being that the original data is coming in as a hex string, and it is desired to report it as MBs, the following code could be of assistance:
```
static const uint64_t BYTES_TO_MEGABYTES = 0x1000000;
_output << boost::format("  %-20s : 0x%x MB\n") % "Memory" % strtoll(_pt.get<std::string>("host.os.memory").c_str(),nullptr,16)/BYTES_TO_MEGABYTES);
```

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -85,6 +86,29 @@ glibc_info()
   return _pt;
 }
 
+static std::string machine_info()
+{
+#if defined(__aarch64__) || defined(__arm__) || defined(__mips__)
```
rozumx, line 91: I have never been a fan of compiler #if def of variable declarations (they always are difficult to support).  By chance, could you refactor it to be something that would not require doing this.  For example:


```
#if defined(__aarch64__) || defined(__arm__) || defined(__mips__)
  #define MACHINE_NODE_PATH = "/proc/device-tree/model";
#elif defined(__PPC64__)
  #define MACHINE_NODE_PATH = "/proc/device-tree/model-name";
  // /proc/device-tree/system-id may be 000000
  // /proc/device-tree/model may be 00000
#elif defined (__x86_64__)
  #define MACHINE_NODE_PATH = "/sys/devices/virtual/dmi/id/product_name";
#else
  #error "Unsupported platform"
#endif

static std::string machine_info()
{
  static const char node[] = MACHINE_NODE_PATH;
  std::string model("unknown");
  std::ifstream stream(node);
  if (stream.good()) {
    std::getline(stream, model);
    stream.close();
  }
  return model;
}
```

Doing so keeps the keep readable easy to support.

rozumx (changes requested):
In addition to the JSON name changes below, could you also add to the comments a sample output of the JSON report.  

Thanks

Himanshu-xilinx (approved):
looks good

src/runtime_src/core/edge/user/system_linux.cpp
sonals: Changes made to store memory as as hex string.

src/runtime_src/core/pcie/windows/system_windows.cpp
sonals: Done

src/runtime_src/core/tools/common/ReportHost.cpp
sonals: Thanks for the suggestion -- done!

src/runtime_src/core/pcie/linux/system_linux.cpp
sonals: Done!

src/runtime_src/core/pcie/windows/system_windows.cpp
```diff
@@ -174,7 +186,7 @@ get_mgmtpf_device(device::id_type id) const
 
 void
 system_windows::
-program_plp(std::shared_ptr<device> dev, std::vector<char> buffer) const 
+program_plp(std::shared_ptr<device> dev, std::vector<char> buffer) const
```
stsoe, line 191: Not part of this PR, but still has to be fixed.  Do not pass by value, especially `buffer ` looks very inefficient, but also dev should be pass by const reference.

src/runtime_src/core/pcie/windows/system_windows.cpp
```diff
@@ -29,6 +29,7 @@
 #include <chrono>
 #include <thread>
 #include <ctime>
+#include <thread>
```
stsoe, line 32: Why this extra include?  already included line 30.

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -85,6 +99,18 @@ glibc_info()
   return _pt;
 }
 
+static std::string machine_info()
+{
+  static const char node[] = MACHINE_NODE_PATH;
```
stsoe, line 105: What's the point of node here?  Why not use MACHINE_NODE_PATH directly line 106?

src/runtime_src/core/edge/user/system_linux.cpp
```diff
@@ -165,7 +195,7 @@ get_mgmtpf_device(device::id_type id) const
 
 void
 system_linux::
-program_plp(std::shared_ptr<device> dev, std::vector<char> buffer) const 
+program_plp(std::shared_ptr<device> dev, std::vector<char> buffer) const
```
stsoe, line 199: Not part of this PR, but still has to be fixed. Do not pass by value, especially buffer looks very inefficient, but also dev should be pass by const reference.

src/runtime_src/core/edge/user/system_linux.cpp
```diff
@@ -85,7 +100,20 @@ glibc_info()
   return _pt;
 }
 
-void 
+static std::string
+machine_info()
+{
+  static const char node[] = MACHINE_NODE_PATH;
```
stsoe, line 107: What's the point of node here? Why not use MACHINE_NODE_PATH directly line 108?

stsoe (changes requested):

src/runtime_src/core/pcie/windows/system_windows.cpp
sonals: I wonder how this crept in. I went ahead and updated the performance critical std::vector reference change `system::program_plp(std::shared_ptr<device> dev, const std::vector<char> &buffer)` 
Leaving out the shared_ptr device change for future.

src/runtime_src/core/pcie/windows/system_windows.cpp
sonals: Removed

src/runtime_src/core/pcie/linux/system_linux.cpp
sonals: Removed my brain dead artifact.

src/runtime_src/core/edge/user/system_linux.cpp
sonals: Done, see above.

src/runtime_src/core/edge/user/system_linux.cpp
sonals: Done, see above.

src/runtime_src/core/pcie/linux/system_linux.cpp
rozumx: Yes, that will work as well.  

Personally for me, I like to use a static const variable to contain this type of information especially if the value is determine via the compiler environment.  Having this in a static const variable also enables one to easily inspect this value via the debugger, if needed in the future.  

Again, not a big deal.  It is just a coding pattern I like to do especially if it is being used as an argument into a function/method that is difficult to step into.

rozumx (approved):
Thank you for making the recommended changes.

src/runtime_src/core/edge/user/system_linux.cpp
AShivangi: Sorry, that's on me. I will take care of this asap

src/runtime_src/core/pcie/linux/system_linux.cpp
stsoe: Sounds like you suggest working around the misuse of #define :-)  
In reality all the #defines should be `constexpr char blah[] = "blahblah"`   
Should we outlaw #defines in C++ code?

PR #4078, opened 2020-09-21
Address various xbmgmt2/xbutil2 EoU issues & bugs

rozumx:
Work done
----------
+ xbutil2 program
.  + Added long discription to xbutil2 program command
.  + Removed -r option (not used)
.  + Added long option names: "--device" and "--program"
+ xbutil2 validate
.  + Added "--output" option
.  + Added "--runs" option that will take in "all", "quick", or a collection of tests
.  + Added "--format" option for the user to select between "text" and "json"
.  + Removed "--json" and "--quick" option
.  + Corrected an issue where the static TestCollection data was being updated with test data and not the instance of the test data
.  + Added support for "--runs" help to show all the available runs
+ xbutil2 examine
.  + Make the option "--examine" hidden
.  + All reports in the "help" section are sort and presented alphabetically
.  + Remove unneeded host-interface report
+ xbutil2 advanced
.  + Removed obsolete "--clock" option (Not needed per XRT PRT)
.  + Fix help error when the "--p2p" option
+ xbmgmt2 program
.  + Hide advance command "--flash-type"
.  + Renamed "--plp" to "--partition"
+ xbmgmt2 status
.  + Remove unused option "--element"
+ xbmgmt2 advanced
.  + Removed unused "--clock" option (Not needed per XRT PRT)
.  + Removed unused "--nifd" option
+ xbmgmt2 config
.  + Rename "--enable_retention" and "--disable_retention" to "--retention [ENABLE | DISABLE]"
.  + Added some helpful user errors when "--device" is not set and "--deamon" is
+ General
.  + Fix bug on how the help messages are produced for report and test options
.  + Fix various spelling errors


src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -1002,8 +1069,59 @@ SubCmdValidate::execute(const SubCmdOptions& _options) const
...
+
+      // Validate all of the test names
+      bool nameFound = false;
+      for (auto &test : testNameDescription) {
+        const std::string testNameLC = boost::algorithm::to_lower_copy(test.first);
+        if (userTestNameLC.compare(testNameLC) == 0) {
+          nameFound = true;
+          break;
+        }
+      }
```
AShivangi, line 1105: Hmm. Does this work for all test names? What if the first test name is valid, but the second one is not? Are we looking for at least one valid test name?

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -1012,22 +1130,46 @@ SubCmdValidate::execute(const SubCmdOptions& _options) const
...
 
-  for(auto const& dev : deviceCollection) {
-    run_test_suite_device(dev, _ptDevCollectionTestSuite, json, quick);
+  for (unsigned index = 0; index < testSuite.size(); ++index) {
+    if (testsToRun[0] == "all") {
+      testObjectsToRun.push_back(&testSuite[index]);
+      continue;
+    }
+
+    if (testsToRun[0] == "quick") {
```
AShivangi, line 1142: else if?

AShivangi (approved):
Looks great! Thanks for making the changes :)

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
rozumx: Correct, the algorithm will search the _entire collection_ looking for 1 valid test name.  If none found an error will be produced.

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
rozumx: Don't believe one is necessary here.   Let me explain further.
This condition is part of a loop of tests that the user has requested to be ran.  The key here is that ALL the possible tests are examined to see if they match one of the tests in the user collection (not visa-versa).  This allows for a more efficient search with the following conditions, especially since that the user set of tests have already been qualified, no need to do it again.   

Besides, the "continue" and "break" remove the need for the "else if" :)

PR #4084, opened 2020-09-21
XRT-573 Support M2M CU and copybo command in new KDS

mamin506:


src/runtime_src/core/common/drv/kds_core.c
```diff
@@ -577,6 +578,14 @@ int kds_add_context(struct kds_sched *kds, struct kds_client *client,
 			return -EINVAL;
 		}
 		++client->virt_cu_ref;
+		/* a special handling for m2m cu :( */
```
maxzhen, line 581: should this special handling be done when the client reserves virtual CU for the first time only?

src/runtime_src/core/common/drv/kds_core.c
```diff
@@ -601,6 +611,14 @@ int kds_del_context(struct kds_sched *kds, struct kds_client *client,
 			return -EINVAL;
 		}
 		--client->virt_cu_ref;
+		/* a special handling for m2m cu :( */
```
maxzhen, line 614: see above comment, done for the last release?

src/runtime_src/core/common/drv/kds_core.c
```diff
@@ -617,6 +635,9 @@ insert_cu(struct kds_cu_mgmt *cu_mgmt, int i, struct xrt_cu *xcu)
 {
 	cu_mgmt->xcus[i] = xcu;
 	xcu->info.cu_idx = i;
+	/* m2m cu */
+	if (xcu->info.intr_id == 128)
```
maxzhen, line 639: need a more descriptive name for const number 128

maxzhen (changes requested):

src/runtime_src/core/common/drv/kds_core.c
mamin506: Sure. we can.

xdavidz (approved):

maxzhen (approved):

PR #4107, opened 2020-09-25
Enable stall trace in new architecture

IshitaGhosh:


jvillarre (approved):
Change looks good to make sure device offload is loaded when checking stall trace.  This change enables both the old architecture to continue working and the new plugins to work as well.

PR #4109, opened 2020-09-25
flash type reported by vsec is not fine-grained

xuhz:
qspi flash type reported by vsec is not fine-grained to tell it is single or not, qspi flasher should rely on the flash type defined in devices.h to get the accurate info


maxzhen:
Shouldn't the VSEC provide accurate info?

houlz0507 (approved):

xuhz:
yes. vsec should provide that info. Has asked Gary and may create a CR tracking it.
before vsec has a fix, this PR can be a solution

mamin506 (approved):

PR #4114, opened 2020-09-28
XMA2 Timeout fix. CR-1072041

sarab96:
CU cmd Timeout fix. timeout of cu hangs or soft kernel crashes.
CR-1072041


stsoe (changes requested):
Please add comments explaining what it going on.  The functions are too long to review without comments.

sarab96:
Done
> Please add comments explaining what it going on. The functions are too long to review without comments.

stsoe (approved):
The code is still too poorly organized to review probably , I think this is legacy code though.

PR #4118, opened 2020-09-28
CR1045828 Clear AER / Device Status Registers on attach

houlz0507:


maxzhen (commented):
do we clear during attach? or we just want to do it after hot reset?

houlz0507:
> do we clear during attach? or we just want to do it after hot reset?
Both. Please see 1576 which clear during probe

maxzhen:
> 
> 
> > do we clear during attach? or we just want to do it after hot reset?
> > Both. Please see 1576 which clear during probe

wow... not expecting a side effect like this in xocl_fill_dsa_priv() whose function name does not give any hint...

maxzhen (approved):

PR #4122, opened 2020-09-29
CR-1077372 zocl panic when xbutil validate

xuhz:
1. enable ert on u30
2. unconfig ert config cmd in ert mode

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/sched_exec.c
+++ b/src/runtime_src/core/edge/drm/zocl/sched_exec.c
@@ -770,10 +770,13 @@ configure(struct sched_cmd *cmd)
 
 	cfg = (struct ert_configure_cmd *)(cmd->packet);
 
-	if (exec->configured != 0) {
-		DRM_WARN("Reconfiguration not supported\n");
-		return 1;
-	}
+	/*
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -770,10 +770,13 @@ configure(struct sched_cmd *cmd)
...
-		DRM_WARN("Reconfiguration not supported\n");
-		return 1;
-	}
+	/*
+	 * In ert mode, no ert config cmd for the same xclbin will be coming
+	 * from host. so always unconfig ert when there is ert config cmd
+	 * arriving
+	 */
+	if (zdev->ert)
+		sched_reset_scheduler(zdev->ddev)
```
xdavidz, line 780: @larry9523  is this change safer for versal too? I think we don't want to support reconfig?

larry9523: xocl is responsible to make sure only one config command is sent per xclbin

larry9523 (approved):

xdavidz (approved):

mamin506 (approved):

PR #4123, opened 2020-09-29
C++ API rst documentation

stsoe:
This PR adds support for C++ API documentation.  This is done through
doxygen and a sphinx plugin 'breathe'.  The documetation step in XRT,
runs doxygen on C++ headers to genetated XML data consumed and
formatted by sphinx with the breathe plugin.

This PR also separates documentation generation from building.  By
default the build.sh script no longer generates documentation.  This
is done because the necessary version of sphinx and plugins are not
available on all platforms.  In order to generate documentation, the
build.sh script must be invoked with the '-docs' option, e.g.

% build.sh -docs

Documentation generation will work if and only if sphinx-build and
breathe are installed along with kernel-doc.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -41,6 +41,7 @@ usage()
     echo "[-edge]                    Build edge of x64.  Turns off opt and dbg"
     echo "[-nocmake]                 Skip CMake call"
     echo "[-noctest]                 Skip unit tests"
+    echo "[-docs]                    Enable documentation generation with sphinx"
     echo "[-j <n>]                   Compile parallel (default: system cores)"
     echo "[-ccache]                  Build using RDI's compile cache"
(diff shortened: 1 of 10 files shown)


uday610 (commented):
@stsoe , the doc build and source code build is a mutually exclusive process? Or we can do both together by some combination of build.sh switch?

stsoe:
@uday, it is mutually exclusive, but changed so that you no longer have to wait for building all source code.

keryell:
s/documetation/documentation/
s/genetated/generated/

While looking at the comments, I wonder whether we could not get rid of these Christmas commenting style with a lot of `*` to fit Fortran66 coding style for the teletype printers, now we have colorful displays... :-)
Also painful when copy-pasting, with a lot of `*` junk. :-(

PR #4127, opened 2020-09-29
CR-1077438 ASTeR TC 7.02 - Intermittent firewall trip during xbutil validate (u50 base4)

chienwei-lan:


src/runtime_src/ert/scheduler/scheduler_v30.cpp
```diff
@@ -1081,7 +1081,8 @@ scheduler_v30_loop()
           auto cqvalue = read_reg(slot.slot_addr);
 
           if (cqvalue & (AP_START|AP_CONTINUE)) {
-            write_reg(slot.slot_addr,0x0); // clear
+            while(read_reg(slot.slot_addr))
+              write_reg(slot.slot_addr,0x0); // clear
```
maxzhen, line 1085: why do we still need this change? shouldn't the one line change below just do it?

maxzhen (changes requested):

mamin506 (approved):

maxzhen (approved):

PR #4129, opened 2020-09-29
3rd attemp to fix SC mismatch warning msg

maxzhen:
The previous fix does not work on U.2 where there is no SC at all!
I have tested on U.2 and U50. @xuhz , pls help to verify on U25/30.


xuhz (commented):
new added code is to set bmc_ver as 0.0.0 if there is no sc?

then why do you remove the code that set exp_bmc_ver as bmc_ver if there is fixed bmc? u25 and u30 will not get exp_bmc_ver anymore

maxzhen:
@xuhz , I think that you're right, I should not remove the original code that set exp_bmc_ver as bmc_ver if there is fixed bmc. I've added them back. Also added some more comments. Please try it once again.

xuhz (approved):
tested on u25. worked fine

PR #4138, opened 2020-09-30
Addressing minibash issues

AShivangi:
- XRT-781 Align ":" in xbutil validate
- XRT-737 hide xbutil examine --element option
- XRT-764 Its clear why test 9 is skipped. Can the messaging be changed to explain why test 10 is skipped?
- XRT-767 Allow the user to produce the latest JSON version for a given XRT release
- XRT-782 xbmgmt 2 advanced config is not hidden
- remove deprecated SubCmdList
- CR-1077416 Tracking CR: xbutil2 testing. Improve messaging when card is not ready
This will also print devices which are visible but are not ready for use
Sample output:
```
$ xbutil --new examine -r firewall
0 devices found
Warning: Due to missing devices, the following reports will not be generated:
         - firewall
```
- CR-1077611 xbmgmt2 testing: does not run if there is no deployment platform installed on the system
Sample output:
```
$ sudo xbmgmt --new status
Device : [0000:b3:00.0]

Flash properties
  Type                 : spi
  Serial Number        : 12809621T512

  Platform             : xilinx_u200_xdma_201830_3
  SC Version           : 4.3.9
  Platform ID          : 0x5ee83d83

  None

WARNING  : No shell is installed on the system.
```


src/runtime_src/core/tools/common/ReportHost.cpp
```diff
@@ -108,6 +108,10 @@ ReportHost::writeReport(const xrt_core::device * _pDevice,
 
   _output << "Devices\n";
   boost::property_tree::ptree& available_devices = _pt.get_child("host.devices", empty_ptree);
+
+  if(available_devices.size() == 0)
```
rozumx, line 112: Since this code is determining if the collection is empty, it would be better to use:
` if(available_devices.empty())`
it is faster and easier to read.

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
```diff
@@ -256,6 +256,7 @@ OO_Config::OO_Config( const std::string &_longName, bool _isHidden)
   m_optionsDescription.add_options()
     ("device,d", boost::program_options::value<decltype(m_devices)>(&m_devices)->multitoken(), "The Bus:Device.Function (e.g., 0000:d8:00.0) device of interest")
     ("retention", boost::program_options::value<decltype(m_retention)>(&m_retention),"Enables / Disables memory retention.  Valid values are: [ENABLE | DISABLE]")
+    ("ddr", boost::program_options::bool_switch(&m_ddr), "Memory type to be enabled/disabled for retention")
```
rozumx, line 259: Why is this option no longer hidden?  Comments didn't see to indicate what this was being done.

If it is needed, then the describing needs to be updated.  For example:
`Enable DDR memory for retention`

src/runtime_src/core/tools/xbmgmt2/ReportPlatform.cpp
```diff
@@ -197,6 +197,8 @@ ReportPlatform::getPropertyTree20202( const xrt_core::device * device,
 static const std::string
 shell_status(bool shell_status, bool sc_status, int multiDSA)
```
rozumx, line 198: The variable `multiDSA `should be rename a name that represents this value.  Something similar to: `numDsaShells`

src/runtime_src/core/tools/xbmgmt2/ReportPlatform.cpp
```diff
@@ -250,6 +252,10 @@ ReportPlatform::writeReport( const xrt_core::device * device,
 
   output << "Flashable partitions installed in system\n"; 
   boost::property_tree::ptree& available_shells = pt.get_child("platform.available_shells");
+  
+  if(available_shells.size() == 0)
```
rozumx, line 256: Please use the `empty()` method.

src/runtime_src/core/tools/xbutil2/SubCmdExamine.cpp
```diff
@@ -163,6 +163,19 @@ SubCmdExamine::execute(const SubCmdOptions& _options) const
           missingReports.push_back(report->getReportName());
       }
       if (!missingReports.empty()) {
+
+        auto dev_pt = XBU::get_available_devices(true);
+        if(dev_pt.size() == 0)
```
rozumx, line 168: Please use the `empty()` method.

rozumx (approved):
Approved with requested minor changes.

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
AShivangi: This is a new feature that was added as a hidden command since it was a new feature. But over time it has matured and ready to be exposed to the users. We exposed it to the users in xbmgmt, but forgot to do so in xbmgmt2

src/runtime_src/core/tools/xbmgmt2/ReportPlatform.cpp
AShivangi: Thanks! I couldn't think of a better variable name at the time :)

dayeh-xilinx:
retest this please

PR #4139, opened 2020-09-30
Support FA descriptor for OCL

stsoe:
Updated OCL xclbin meta data parser to amend argument meta data with
FA specific data.  This code relies on XML kernel attribute
hwControlProtocol set to "fast_adapter" for FA style kernels.

Execution context take custom code path for Fast Adapter protocol, and
uses computed xclbin meta data to offset and populate the command
payload with the fa_descriptor.

Update low level fa_kernel test to add OCL test case alongside the
existing XRT native test case.  Renamed main.cpp -> xrt.cpp and added
ocl.cpp.  The executables are fa_xrt and fa_ocl when built by CMake.

### DIFF
--- a/src/runtime_src/xocl/core/execution_context.cpp
+++ b/src/runtime_src/xocl/core/execution_context.cpp
@@ -17,6 +17,8 @@
 #include "execution_context.h"
 #include "device.h"
 #include "event.h"
+#include "ert.h"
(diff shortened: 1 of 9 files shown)


stsoe:
@mamin506 FYI.  Note that the low level fa_kernel XRT native test was renamed per PR description.

PR #4140, opened 2020-09-30
VART is not supported on Windows

IshitaGhosh:


jvillarre (approved):
Changes look good to stop building vart_plugin on Windows for now.

PR #4142, opened 2020-10-01
CR-1078348 rw_shared support in xrt::kernel opening

stsoe:
Don't bail on shared CU if Runtime.rw_shared is true.


uday610 (approved):
Thank you

PR #4149, opened 2020-10-01
CR-1077617: Adding section to run summary

jvillarre:
Changing run summary file generation in old and new architecture so multiple run summaries generated from a single execution can be distinguished and processed.  The additional run summary section allows tools to find run summary files from the same run and consolidate information from them all.


rozumx (approved):
Hello Jason,

Over all the changes look good.   

In the past year, we have been working hard to remove OS dependencies compile directives from the code base, which is a huge undertaking.  The good news is that we have "nearly" eliminated that from the new versions of xbutil and xbmgmt.

We did so by creating a "common" API and place the implementation in a abstract library for the given OS.  This has worked out rather nicely.   

Going forward, if you are your team are interested in this, please do have them reach out to use and we will be happy to show you what we did.  And yes, it makes supporting the code REALLY easy :).

pgschuey (approved):
changes look good

PR #4151, opened 2020-10-01
CR-1076201 xrt::kernel::<member_function> to get argument offset

stsoe:
Added for both C++ and C


uday610 (changes requested):
Looks good. Only small change in a C function header.

change:  Use with ``read_register()`` and ``write_register()`` if manually
to : Use with `xrtKernelReadRegister` and `xrtKernelWriteRegister` if manually

PR #4152, opened 2020-10-01
Notify host if AP_DONE | AP_IDLE

chienwei-lan:
If the CU is a dataflow CU, KDS and ERT have weak relationship with each other which means KDS doesn't have to wait ERT's response for action, KDS may issue another CU START command if CU marked as AP_DONE.

https://github.com/Xilinx/XRT/pull/4127

### DIFF
--- a/src/runtime_src/ert/scheduler/scheduler.cpp
+++ b/src/runtime_src/ert/scheduler/scheduler.cpp
@@ -1154,8 +1154,14 @@ scheduler_loop()
         if (!cu_status[cuidx])
           continue; // CU is not used
 
+        /* For dataflow kernel, KDS and ERT will check the CUs from both sides.
+         * It's likely that ERT checks the CUs after the CUs are completed by KDS.
+         * So here ERT should check both AP_DONE and AP_IDLE bits or ERT will keep
+         * polling CU status register and never get AP_DONE. It may cause firewall
+         * tripped if we freeze the axi gate of the dynamic region.
+         */
         auto cuvalue = read_reg(cu_idx_to_addr(cuidx));
-        if (!(cuvalue & AP_DONE))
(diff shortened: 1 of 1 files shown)


src/runtime_src/ert/scheduler/scheduler.cpp
```diff
@@ -1155,7 +1155,7 @@ scheduler_loop()
           continue; // CU is not used
 
         auto cuvalue = read_reg(cu_idx_to_addr(cuidx));
-        if (!(cuvalue & AP_DONE))
+        if (!(cuvalue & AP_DONE|AP_IDLE))
```
stsoe, line 1158: Please check https://github.com/Xilinx/XRT/pull/1372.   
I am sorry that the comment on that pull request is not complete enough to explain the change that was made, but I am pretty sure you shouldn't be reverting it.

stsoe (changes requested):
See comment

src/runtime_src/ert/scheduler/scheduler.cpp
maxzhen: The code should be fixed and "()" should be added appropriately.

More importantly, we need to add some comments here to explain why this change is required. Basically, both ERT and KDS are babysitting data flow CUs. By the time when ERT gets to check the CU status, it might have finished the job by KDS and in an idle state. If ERT does not check AP_IDLE, it will keep coming back and check the status until it sees AP_DONE, which might not happen since the application has finished. This is not ideal and will trigger firewall trip if, at the time, an xclbin download is happening.

maxzhen (approved):

maxzhen:
Discussed with @stsoe offline. The change looks OK to him. Merging now.

PR #4154, opened 2020-10-02
disable ert via xbutil through old kds

maxzhen:
xbutil scheduler -k 1 -d \<board\> disables ert.
xbutil scheduler -k 0 -d \<board\> enables ert.

Make sure to run xbutil reset after the change to make sure the new setting takes effect.


mamin506 (approved):
Look good to me

arpitxilinx (approved):

PR #4165, opened 2020-10-05
CR-1077772 ZOCL version number fix

rbramand-xilinx:
> Driver version of zocl is reporting outdated value, fixed to report latest value similar to pcie platforms
> As zocl doesnot follow dkms flow, version.h file is not available at zocl build time
> Instead of reading zocl sysfs node 'version', reading directly XRT_DRIVER_VERSION macro which is available at userspace build time

### DIFF
--- a/src/runtime_src/core/edge/tools/xbutil/base.h
+++ b/src/runtime_src/core/edge/tools/xbutil/base.h
@@ -30,25 +30,13 @@
 
 namespace xcldev {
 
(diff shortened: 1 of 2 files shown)


mamin506 (approved):

chvamshi-xilinx (approved):

rbramand-xilinx:
Closing this pull request as this change already went in with other commit

PR #4167, opened 2020-10-05
CR 1068826: added support for latest u200 2018.3; also AIE profile changes

pgschuey:


hackwa (approved):
looks good. Since aie runs at Ghz rates, it makes sense to sample the data in microseconds intervals. @pgschuey does it make sense to set a minimum interval of 50usec?

PR #4170, opened 2020-10-06
CR-1077997 Minor updates handling the emulation_data path for Versal …

venkatp-xilinx:
Minor updates handling the emulation_data path for Versal DC Platforms. Enabled the hal log for sw_emu with ENV, and other minor tweaks.

### DIFF
--- a/src/runtime_src/core/pcie/emulation/common_em/config.cxx
+++ b/src/runtime_src/core/pcie/emulation/common_em/config.cxx
@@ -71,7 +71,7 @@ namespace xclemulation{
     mPrintErrorsInConsole = true;
     mVerbosity = 0;
     mServerPort = 0;
-    mKeepRunDir=false;
+    mKeepRunDir=true;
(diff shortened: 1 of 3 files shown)


stsoe:
Hi @venkatp-xilinx @chvamshi-xilinx  This PR results in stray files dltmp xmltmp being left by software emulation.  KeepRunDir default was changed to true.

chvamshi-xilinx:
Hi @stsoe , i think this was an intentional change by @venkatp-xilinx due to some request. @venkatp-xilinx , please confirm and share the background.

venkatp-xilinx:
Yes, as Vamshi said, it is done intentionally. Ideally we require this run dir in all the failure scenarios in order to debug the reason for failure. And in sprite regression, for all the success cases, they delete the test output and only outputs are stored for failure tests and having this run dir by default is very useful for initial analysis without rerunning the design.

PR #4175, opened 2020-10-06
Add AIE PLIO event counter profiling support

larry9523:
This PR added
1. PLIO event profiling (so that we have a full support for event counter performance counter option)
2. Update the interface name
xrtStartProfiling --> xrtAIEStartProfiling
xrtReadProfiling --> xrtAIEReadProfiling
xrtStopProfiling --> xrtAIEStopProfiling

### DIFF
--- a/src/runtime_src/core/edge/common/aie_parser.cpp
+++ b/src/runtime_src/core/edge/common/aie_parser.cpp
@@ -31,6 +31,7 @@ namespace pt = boost::property_tree;
 using tile_type = xrt_core::edge::aie::tile_type;
 using rtp_type = xrt_core::edge::aie::rtp_type;
 using gmio_type = xrt_core::edge::aie::gmio_type;
+using plio_type = xrt_core::edge::aie::plio_type;
 using counter_type = xrt_core::edge::aie::counter_type;
 
 inline void
@@ -158,6 +159,31 @@ get_gmio(const pt::ptree& aie_meta)
   return gmios;
 }
 
+std::vector<plio_type>
+get_plio(const pt::ptree& aie_meta)
+{
+  std::vector<plio_type> plios;
+
+  auto plio_nodes = aie_meta.get_child_optional("aie_metadata.PLIOs");
+  if (!plio_nodes)
+    return plios;
+
+  for (auto& plio_node : aie_meta.get_child("aie_metadata.PLIOs")) {
+    plio_type plio;
+
+    plio.id = plio_node.second.get<uint32_t>("id");
+    plio.name = plio_node.second.get<std::string>("name");
+    plio.logical_name = plio_node.second.get<std::string>("logical_name");
+    plio.shim_col = plio_node.second.get<uint16_t>("shim_column");
+    plio.stream_id = plio_node.second.get<uint16_t>("stream_id");
+    plio.is_master = plio_node.second.get<bool>("slaveOrMaster");
+
+    plios.emplace_back(std::move(plio));
+  }
+
+  return plios;
+}
+
 std::vector<counter_type>
 get_profile_counter(const pt::ptree& aie_meta)
 {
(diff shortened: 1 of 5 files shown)


src/runtime_src/core/edge/common/aie_parser.cpp
```diff
@@ -158,6 +159,31 @@ get_gmio(const pt::ptree& aie_meta)
...
 }
 
+std::vector<plio_type>
+get_plio(const pt::ptree& aie_meta)
+{
+  std::vector<plio_type> plios;
+
+  auto plio_nodes = aie_meta.get_child_optional("aie_metadata.PLIOs");
+  if (!plio_nodes)
+    return plios;
```
stsoe, line 169: You could do `return {}`
Then declare your vector before the for loop.

src/runtime_src/core/edge/common/aie_parser.cpp
```diff
@@ -251,6 +277,18 @@ get_gmios(const xrt_core::device* device)
   return ::get_gmio(aie_meta);
 }
 
+std::vector<plio_type>
+get_plios(const xrt_core::device* device)
+{
+  auto data = device->get_axlf_section(AIE_METADATA);
+  if (!data.first || !data.second)
+    return std::vector<plio_type>();
```
stsoe, line 285: `return {} `
should do it.

src/runtime_src/core/edge/user/aie/aie.cpp
```diff
@@ -338,32 +342,50 @@ start_profiling(int option, const std::string& port1Name, const std::string& por
   auto gmio = std::find_if(gmios.begin(), gmios.end(),
             [port1Name](gmio_type it) { return it.name.compare(port1Name) == 0; });
 
-  if (gmio == gmios.end())
+  // For PLIO inside graph, there is no name property.
+  // So we need to match logical name too
+  auto plio = std::find_if(plios.begin(), plios.end(),
+            [port1Name](plio_type it) { return it.name.compare(port1Name) == 0; });
```
stsoe, line 348: `[&port1Name] ` prevents a string copy
and use const plio_type&, e.g. this way:
`[&port1Name] (auto& pt) { return pt.name.compare(port1Name) == 0; }`
Personally I don't like camel case so I'd prefer port1_name but that's just me.

src/runtime_src/core/edge/user/aie/aie.cpp
```diff
@@ -338,32 +342,50 @@ start_profiling(int option, const std::string& port1Name, const std::string& por
...
+  if (plio == plios.end()) {
+    plio = std::find_if(plios.begin(), plios.end(),
+            [port1Name](plio_type it) { return it.logical_name.compare(port1Name) == 0; });
```
stsoe, line 351: ditto

stsoe (changes requested):
Minor suggestions.

src/runtime_src/core/edge/common/aie_parser.cpp
larry9523: done

src/runtime_src/core/edge/common/aie_parser.cpp
larry9523: Let's leave this case as it for now because all other functions like get_rtp, get_gmios, get_tiles etc. are using the same style. We can improve it in a later code clean up.

src/runtime_src/core/edge/user/aie/aie.cpp

src/runtime_src/core/edge/user/aie/aie.cpp

larry9523:
@stsoe review comments are addressed. Please take a look again. Thanks!

src/runtime_src/core/edge/common/aie_parser.cpp
```diff
@@ -158,6 +159,31 @@ get_gmio(const pt::ptree& aie_meta)
   return gmios;
 }
 
+std::vector<plio_type>
+get_plio(const pt::ptree& aie_meta)
+{
+  std::vector<plio_type> plios;
```
stsoe, line 165: Move to line 170

stsoe (approved):
Looks good.  Consider moving the vector definition as noted in comment.

src/runtime_src/core/edge/common/aie_parser.cpp
larry9523: Sorry, forgot about it. It is done now.

PR #4176, opened 2020-10-06
pybind11 API updates based on feedback

schelleg:
bo.write rewritten to use buffers

bo updates constructor and new functions added.

template change to keep up with master

Signed-off-by: Graham Schelle <graham.schelle@xilinx.com>

### DIFF
--- a/src/python/pybind11/src/pyxrt.cpp
+++ b/src/python/pybind11/src/pyxrt.cpp
@@ -93,7 +93,7 @@ py::class_<xrt::run>(m, "run")
        r.set_arg(i, item);
      })
    .def("set_arg", [](xrt::run & r, int i, int & item){
-       r.set_arg<int>(i, item);
+       r.set_arg<int&>(i, item);
      })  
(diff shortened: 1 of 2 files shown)


src/python/pybind11/src/pyxrt.cpp
```diff
@@ -105,7 +105,11 @@ py::class_<xrt::run>(m, "run")
 py::class_<xrt::kernel>(m, "kernel")
     .def(py::init([](xrt::device d, const py::array_t<unsigned char> u, const std::string & n, bool e){
         return new xrt::kernel(d, (const unsigned char*) u.request().ptr, n, e);
```
stsoe, line 107: I am not claiming to understand this file, but where is the corresponding delete?

schelleg: with this return style, Python takes ownership of the object and calls delete() when the Python object's reference counter goes to zero.

sonals:
ok to test

PR #4177, opened 2020-10-06
Callbacks to shutdown XDP AIE trace, profile from user level AIE APIs for profiling

IshitaGhosh:


jvillarre (approved):
Callbacks look good to shut down profiling plugins if they are active

larry9523 (approved):
Looks good. 
Two nits
1) It may be better to put some comment before calling xdpaie::*** so that code reader knows what we are doing
2) Those call back are not expensive at all, if it has been called, right? Looks like we are blindly calling the xdp callback every time xcl profiling api is called. It should only be necessary to called them once,

IshitaGhosh:
@larry9523 , 
1. I can add the comment
2. The callbacks check a map to find an entry and erase it if found. It should not be much expensive. But I can revisit it in future PR.

PR #4180, opened 2020-10-07
Automatically install packages on amazon linux

ksmurakoshi:
When installing on amazon linux, XRT/src/runtime_src/tools/scripts/xrtdeps.sh requires user input (yum install <package>). Fix to allow xrtdeps.sh to install without user input. 

Signed-off-by: Kevin Murakoshi <murakevi@amazon.com>

### DIFF
--- a/src/runtime_src/tools/scripts/xrtdeps.sh
+++ b/src/runtime_src/tools/scripts/xrtdeps.sh
@@ -324,7 +324,7 @@ prep_centos7()
     fi
     echo "Installing cmake3 from EPEL repository..."
     yum install -y cmake3
-    if [ $docker == 0 ]; then 
+    if [ $docker == 0 ]; then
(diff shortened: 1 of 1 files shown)


sonals:
ok to test

stsoe (commented):
@ksmurakoshi Thanks looks good.  If you're from outside of Xilinx, you need "sign-off-by" in your pull request. Please follow the instructions listed here: https://github.com/Xilinx/XRT/blob/master/CONTRIBUTING.rst

ksmurakoshi:
@stsoe I updated my initial pull request. Is that enough? (or do you need me to do something else)?

stsoe:
@ksmurakoshi Can you add signed-off in this PR also?  Maybe update your commit comment with a force push?

stsoe:
Thank you!

PR #4181, opened 2020-10-07
CR-1078587 & CR-1079255 Add new check to handle NO SC case on U2

rajkumar-xilinx:
Satellite controller mode NOSC_MODE used to handle NO SC case on Raptor2 1RP U2 platform.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
@@ -1053,8 +1053,13 @@ static bool xmc_clk_scale_on(struct platform_device *pdev)
 static bool nosc_xmc(struct platform_device *pdev)
 {
 	struct xocl_xmc *xmc = platform_get_drvdata(pdev);
+	struct xmc_status status;
 
 	if (xmc->priv_data && (xmc->priv_data->flags & XOCL_XMC_NOSC))
+		return true;
+
+	safe_read32(xmc, XMC_STATUS_REG, (u32 *)&status);
+	if (status.sc_mode == XMC_SC_NORMAL_MODE_SC_NOT_UPGRADABLE)
 		return true;
 
 	return false;
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
@@ -228,7 +228,8 @@ enum sc_mode {
 	XMC_SC_BSL_MODE_UNSYNCED = 2,
 	XMC_SC_BSL_MODE_SYNCED = 3,
 	XMC_SC_BSL_MODE_SYNCED_SC_NOT_UPGRADABLE = 4,
-	XMC_SC_NORMAL_MODE_SC_NOT_UPGRADABLE = 5
+	XMC_SC_NORMAL_MODE_SC_NOT_UPGRADABLE = 5,
+	XMC_SC_NOSC_MODE = 6
 };
 
 #define	READ_REG32(xmc, off)			\
@@ -1066,7 +1067,7 @@ static bool nosc_xmc(struct platform_device *pdev)
 		return true;
 
 	safe_read32(xmc, XMC_STATUS_REG, (u32 *)&status);
-	if (status.sc_mode == XMC_SC_NORMAL_MODE_SC_NOT_UPGRADABLE)
+	if (status.sc_mode == XMC_SC_NOSC_MODE)
 		return true;
 
 	return false;


chvamshi-xilinx (approved):

rajkumar-xilinx:
Alan Paul from CMC team suggested to use this SC mode for NO SC case. Hence, used this mode to identify if the platform has SC or not.

maxzhen (commented):
I'm not sure if sc-not-upgradeable means no sc. @xdavidz ? We do have SC on U25.

xdavidz:
> I'm not sure if sc-not-upgradeable means no sc. @xdavidz ? We do have SC on U25.

No, it doesn't mean no sc. 
On U25, it means no sc to be changed or flashed.
On U30, it means the sc is not upgradeable, obviously the original meaning.

Once we are using this flag, we have to make sure the core version is greater than XMC_CORE_SUPPORT_NOTUPGRADABLE

chvamshi-xilinx:
Thanks @maxzhen  and @xdavidz . 
@rajkumar-xilinx , lets check with CMC team for the right flag.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
```diff
@@ -1066,7 +1067,7 @@ static bool nosc_xmc(struct platform_device *pdev)
 		return true;
 
 	safe_read32(xmc, XMC_STATUS_REG, (u32 *)&status);
-	if (status.sc_mode == XMC_SC_NORMAL_MODE_SC_NOT_UPGRADABLE)
+	if (status.sc_mode == XMC_SC_NOSC_MODE)
```
xdavidz, line 1070: should we check cmc core version or we don't have to?

rajkumar-xilinx: Hi @xdavidz , we don't need to check for core version register since this CMC FW doesn't define core version register and it is dedicated repo for U2 platform.

CMC FW has new mode defined "NOSC_MODE" in the satellite controller modes list, so we are using this mode to identify if the platform has SC or not

maxzhen: so how do you plan to support old cmc on u.2 which does not support this mode? they are going to be treated as sc is present?

maxzhen (changes requested):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
rajkumar-xilinx: Older means are you referring to u2 legacy shells?

On U2 Legacy shell: xrt driver will continue to use flag XOCL_XMC_NOSC flag which set in device.h 

On Raptor2 1RP U2 shell: It will use this new sc mode to identify it.

Legacy shells can also use this new way, but as per platform team, there is no plan to check-in this new cmc firmware onto legacy shells. Hence, we have to use flag XOCL_XMC_NOSC

maxzhen: By "old" I mean old CMC before this new mode is added. The CMC can be upgraded and downgraded separately from shell. So you can't determine CMC version by looking at shell version.

Also, since you're modifying common CMC code for **all** Alveo boards, we have to consider other old CMC on other boards as well. CMC may have reused some bits in it's register map before and the way to make sure XRT does not mis-interpret any bit when new feature is added is to check core version. So, this is why we always check core version before we start interpreting new SC mode. @xdavidz knows the details.

The fact that the CMC on U.2 does not provide core version means that the CMC on U.2 is not compatible with CMC on other boards. There maybe likely more incompatible registers on U.2 CMC, but I don't know at this moment. However, if we modify common CMC code based on what CMC on U.2 present to XRT, we may break other boards. And with our current test coverage in pipeline, we will not know the breakage before we merge the code. CMC team does not care about compatibility, but we have to.

For now, I can see that we have two options:
1) Fork CMC driver to have a separate CMC driver for U.2. This requires a new endpoint to be defined in dtb for U.2, so that XRT knows to load this specific version of CMC driver.
2) Ask CMC team to implement the CMC on U.2 to be compatible with CMC on other Alveo boards.
To me, option 1 seems more practical considering the engineering resources CMC team has.

rajkumar-xilinx: Hi @maxzhen , thank you for sharing more details to understand the usage of core version check.

CMC driver is separate for U2, so all changes which are specific to U2 will go to this driver (https://gitenterprise.xilinx.com/CMC/CMC/tree/U2/ ).
A separate branch was created for U2 in CMC repo, so any U2 changes will goto this branch.
This CMC firmware will not work on Alveo boards since this CMC driver will not have any support to interact with SC (SC is not present on U2 platform).

Also, On U2, CMC fw is part of bitfile and hence stopping/loading of XMC is skipped from XRT. Platform team confirmed this.

maxzhen: i think you're talking about cmc firmware, not cmc driver in xrt, right? i don't care how cmc firmware is implemented / maintained at all. the register map is the only thing visible to xrt, which we need to care about.

maxzhen:
Closing this. We agreed to fork cmc driver to have a dedicated version for u.2 until it's cmc firmware is merged w/ main branch.

PR #4186, opened 2020-10-08
XRT-709 xbutil query displays all sensors as N/A on U30

xuhz:


xdavidz (approved):

larry9523 (commented):
I am confused by the logic here.
1. we make U30 as autonomous XMC
2. when we decide if sc is ready, we say it should be autonomous XMC but should not be U30.
so why bother? Is it because in other places, U30 should be treated as autonomous but not in is_ready case?

maxzhen:
We should stop adding more flags to autonomous CMC. The existing flags are considered hack for versal and smart nic only. When we move U30 to SSv2, it should be officially data driven without hack.

xuhz:
> I am confused by the logic here.
> 
> 1. we make U30 as autonomous XMC
> 2. when we decide if sc is ready, we say it should be autonomous XMC but should not be U30.
>    so why bother? Is it because in other places, U30 should be treated as autonomous but not in is_ready case?

in other places, eg host need to control cmc, since cmc on u30 is not controlled by host, autonomous is correct, but for sc_ready, the info is from cmc regmap -- which is the only cmc resource that can be accessed also from host. Host need to read the regmap instead of just return. Versal is the same case?

xdavidz:
> We should stop adding more flags to autonomous CMC. The existing flags are considered hack for versal and smart nic only. When we move U30 to SSv2, it should be officially data driven without hack.

Maybe U30 is just not that lucky, but if time is very urgent to make U30 work, I think 1 or 2 flags are the same for us to clean those up in the future. :)

larry9523:
> > I am confused by the logic here.
> > 
> > 1. we make U30 as autonomous XMC
> > 2. when we decide if sc is ready, we say it should be autonomous XMC but should not be U30.
> >    so why bother? Is it because in other places, U30 should be treated as autonomous but not in is_ready case?
> 
> in other places, eg host need to control cmc, since cmc on u30 is not controlled by host, autonomous is correct, but for sc_ready, the info is from cmc regmap -- which is the only cmc resource that can be accessed also from host. Host need to read the regmap instead of just return. Versal is the same case?

OK. So it is because of other places.

larry9523 (approved):

maxzhen:
> 
> 
> > We should stop adding more flags to autonomous CMC. The existing flags are considered hack for versal and smart nic only. When we move U30 to SSv2, it should be officially data driven without hack.
> 
> Maybe U30 is just not that lucky, but if time is very urgent to make U30 work, I think 1 worse and 1+ is even worse. However, maybe we should keep it in mind and file a CR to clean all this up in next release asap?

Cleaning up may need new metadata and this is the right time to ask for it since we're transitioning to SSv2 data driven U30. If we miss this window, it might be difficult to clean up. Let's have an offline talk.

PR #4196, opened 2020-10-08
Add warning messages if aie counter or trace metadat is missing in design but aie_profile/aie_trace is enabled 

IshitaGhosh:


jvillarre (approved):
Changes look good to issue warning and catch edge cases

PR #4198, opened 2020-10-08
bash fixes

AShivangi:
- XRT-765 Exception thrown after help message produced and misspelling
- XRT-771 Xbutil examine -d 0000:3f:00.1 -r verbose show more than one device
- CR-1077429 Improve messaging for BDF specified incorrect or not found
- CR-1077401 xbmgmt2 - xbmgmt --new reset --type soft-kernel reports 'invalid option soft-kernel' but help shows it as a valid option
- XRT-772 Parsing of the BDF value needs to be strict.
Tested in xbutil:
65:00- Valid
65:00.1 - Valid
0000:65:00.1 - Valid
0000:65:00.0 - Invalid (mgmt not user pf)
65 - Invalid (Need at least bus and dev)

### DIFF
--- a/src/runtime_src/core/common/utils.cpp
+++ b/src/runtime_src/core/common/utils.cpp
@@ -23,7 +23,8 @@
 #include <string>
 #include <atomic>
 #include <cstdint>
-#include <boost/tokenizer.hpp>
+#include <limits>
+#include <boost/algorithm/string.hpp>
 
 namespace {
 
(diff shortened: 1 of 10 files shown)


src/runtime_src/core/common/utils.cpp
```diff
@@ -183,24 +185,57 @@ unit_convert(size_t size)
...
+  
+  //check if edge, return the first device
+  uint16_t d = 0;
+  auto device = _inUserDomain ? get_userpf_device(d) : get_mgmtpf_device(d);
+  auto bdf = device_query<query::pcie_bdf>(device);
+  if(std::get<0>(bdf) == 0 && std::get<1>(bdf) == 0 && std::get<2>(bdf) == 0) {
+    return 0;
+  }
+
+  // check if we have 2-3 tokens: domain, bus, device.function
```
rozumx, line 206: Check to see if the tokens.size() is greater then 3 if so, throw the exception now and not later.  Again, qualify the string that is being parse sooner then later.

src/runtime_src/core/common/utils.cpp
```diff
@@ -183,24 +185,57 @@ unit_convert(size_t size)
...
+  // domain is optional
```
rozumx, line 207: Validate that the given string has valid characters.  For example, only the following characters are valid: 0-9, a-f, A-F, ':', '.' 
Any other character should throw an exception.

src/runtime_src/core/common/utils.cpp
```diff
@@ -183,24 +185,57 @@ unit_convert(size_t size)
...
+  if(tokens.size() == 2 || tokens.size() == 3) {
```
rozumx, line 208: This code it too complicated.  This is because the first element (e.g., "domain") is optional.   

If the token vector was reversed (e.g., `std::revers(std::begin(tokens), std::end(tokens));`)  it will become more manageable.  Where:
index 0 is device.function
index 1 is bus
index 2 is domain (if available)

Doing so removes all of this "special" indexing being performed below.

src/runtime_src/core/common/utils.cpp
```diff
@@ -183,24 +185,57 @@ unit_convert(size_t size)
...
+  int radix = 16;
+  uint16_t bus = 0, dev = 0, func = std::numeric_limits<uint16_t>::max();
 
+  //throw an eror if no devices are present
   uint64_t devices = _inUserDomain ? xrt_core::get_total_devices(true).first : xrt_core::get_total_devices(false).first;
+  if(devices == 0) 
+    throw std::runtime_error("No devices found");
+  
+  //check if edge, return the first device
+  uint16_t d = 0;
```
rozumx, line 199: What does 'd' stand for?  It is extremely rare to use 1 character variable names now a days (except for indexes)

src/runtime_src/core/common/utils.cpp
```diff
@@ -183,24 +185,57 @@ unit_convert(size_t size)
...
+    auto cmp_func = [bdf](uint16_t func) 
+    {
+      if (func != std::numeric_limits<uint16_t>::max())
+        return func == std::get<2>(bdf);
+      return true;
+    };
+
+    if (bus == std::get<0>(bdf) && dev == std::get<1>(bdf) && cmp_func(func))
       return i;
   }
```
rozumx, line 241: Great to see that the exception is thrown.
Does this exception get caught and reports the list of available devices?

src/runtime_src/core/common/utils.cpp
```diff
@@ -183,24 +185,57 @@ unit_convert(size_t size)
...
-  if (n == 2)
-    s >> std::hex >> dom >> dummy;
-  s >> std::hex >> b >> dummy >> d >> dummy >> f;
-
-  if ((n != 1 && n != 2) || s.fail())
-    throw std::runtime_error("Bad BDF string '" + bdfstr + "'");
+  std::vector<std::string> tokens; 
+  boost::split(tokens, bdfstr, boost::is_any_of(":")); 
+  int radix = 16;
+  uint16_t bus = 0, dev = 0, func = std::numeric_limits<uint16_t>::max();
```
rozumx, line 191: Please specified each variable separately.

rozumx, line 191: Help me understand, what does it mean to set the "func" value to 0xffff?

I thought it is either '0' or '1' depending on whether or not the mgmt or user drivers are being talked to.   If not defined, should these values be used instead?

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -510,7 +510,7 @@ kernelVersionTest(const std::shared_ptr<xrt_core::device>& _dev, boost::property
 {
   //please append the new supported versions
   const std::vector<std::string> ubuntu_kernel_versions = { "4.4.0", "4.13.0", "4.15.0", "4.18.0", "5.0.0", "5.3.0" };
-  const std::vector<std::string> centos_rh_kernel_versions = { "3.10.0-693", "3.10.0-862", "3.10.0-957", "3.10.0-1062" };
+  const std::vector<std::string> centos_rh_kernel_versions = { "3.10.0-693", "3.10.0-862", "3.10.0-957", "3.10.0-1062", "3.10.0-1127", "4.18.0-147", "4.18.0-193" };
```
rozumx, line 513: Ow! I missing this NASTY string with my earlier reviews.  We need to find a better way to make this data driven.  Too late for this release, but definitely next.

rozumx (changes requested):
bdf2index changes requested.

src/runtime_src/core/common/utils.cpp
AShivangi: I can't report the available devices right now. the function to report the list of available devices is one layer below this file, in XBUtilties.cpp. One option is to move this function in that file. I didn't do that right now since I wasn't sure how much code change would be across the code base. Would you like me o do so?

src/runtime_src/core/common/utils.cpp
AShivangi: This "func" value is what we get from the user. If they pass in a 0 or a 1, we parse it, but if they don;t pass in any function, then the value is -1. The "cmp_func" checks if this variable has a valid value, i.e., not -1 and parses it if that is the case.

src/runtime_src/core/common/utils.cpp
AShivangi: Done

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -510,7 +510,7 @@ kernelVersionTest(const std::shared_ptr<xrt_core::device>& _dev, boost::property
 {
   //please append the new supported versions
   const std::vector<std::string> ubuntu_kernel_versions = { "4.4.0", "4.13.0", "4.15.0", "4.18.0", "5.0.0", "5.3.0" };
```
keryell, line 512: Just curious about why are these versions hard-coded in the source?

AShivangi: Until the new kernels are thoroughly tested, we don't support them officially.

keryell: OK, but it does not prevent XRT to run, I hope?

AShivangi: No, it doesn't. It simply gives a warning saying:
`Kernel verison %s is not officially supported. %s is the latest supported version`

dayeh-xilinx:
retest this please

Himanshu-xilinx:
These changes break xbutil2 on embedded device. Please test embedded device before raising PR.

PR #4214, opened 2020-10-10
xclbinutil: Various Bug Fixes

rozumx:
Work Done
---------
+ Corrected echo alignment of the environment scripts (e.g., setup) for the PYTHONPATH
+ Corrected pointer dereference of a deleted object
+ Updated GROUP section logic to not autogenerate the GROUPs if the are delected in the same session.

### DIFF
--- a/src/runtime_src/tools/scripts/setup.csh
+++ b/src/runtime_src/tools/scripts/setup.csh
@@ -68,4 +68,4 @@ endif
 echo "XILINX_XRT      : $XILINX_XRT"
 echo "PATH            : $PATH"
 echo "LD_LIBRARY_PATH : $LD_LIBRARY_PATH"
-echo "PYTHONPATH     : $PYTHONPATH"
+echo "PYTHONPATH      : $PYTHONPATH"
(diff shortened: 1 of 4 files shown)


rozumx:
@stsoe, This PR addresses the issues you were seeing yesterday.  

During my testing, I was looking at the use-cases and how this data could get out-of-sync with MEM_TOPOLOGY and CONNECTIVITY.   The good news is that when the xclbin image is first created all is good.  

An issue does occur if the user "modifies" the MEM_TOPOLOGY or CONNECTIVITY section in the xclbin image.  In this case, the existing GROUP sections are not modified, resulting in unexpected behavior.  End users might not encounter this use case, but internal folks will if they are debugging these sections.

**_Question_**
Is this a serious enough issue to warrant the following algorithm modifications:
1. Always regenerate the GROUP sections if either the MEM_TOPOLOGY or CONNECTIVITY seconds are added and/or modified (e.g. replace).
2. If MEM_TOPOLOGY or CONNECTIVITY sections are removed, then the GROUP sections are removed.

Of course, the option "--skip-bank-grouping" will disable this functionality.

stsoe:
Ah, interesting case.   I think it ought to be fixed as you describe.   Probably not serious enough for 20.2, but let's get it done when you have the time.   Probably file a CR?

stsoe (approved):

PR #4216, opened 2020-10-12
CR-1079754: Added libhlsmathsim.so path to LD_LIBRARY_PATH

vboggara-xilinx:
Issue : libhlsmathsim.so path is missing in LD_LIBRARY_PATH.
Fix: added libhlsmathsim.so path to LD_LIBRARY_PATH.
Test: Ran some case manually.
Risk: Low


wilderfield:
It doesn't seem ideal to me, changing a user's environment at runtime, especially LD_LIBRARY_PATH, a variable that impacts so many other applications.

chvamshi-xilinx:
> It doesn't seem ideal to me, changing a user's environment at runtime, especially LD_LIBRARY_PATH, a variable that impacts so many other applications.

Yes @wilderfield . I agree with you. There should be loader for the child process and set all this environment.
But, As We are converging on 2020.2, PR filer is going make that change in next release.
@vboggara-xilinx , please file a CR for the next release to clean this up

PR #4217, opened 2020-10-12
Add tracing for HAL APIs in Edge flow

IshitaGhosh:


src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -1523,6 +1523,8 @@ resetAIEArray(drm_zocl_aie_reset &reset)
 unsigned
 xclProbe()
 {
+  PROBE_CB;
```
stsoe, line 1526: Please comment why PROBE_CB is not protected like all the other macro CBs?

stsoe (changes requested):

src/runtime_src/core/edge/user/shim.cpp
IshitaGhosh: The whole "xclProbe" method is under #ifndef __HWEM__ macro. So, there is no need to add #ifndef __HWEM__ macro guard again for PROBE_CB.

stsoe (approved):

jvillarre (approved):
Addition of HAL level callbacks for edge looks good.

PR #4219, opened 2020-10-13
Added environment variable suport to select classic or nextgen tools.

rozumx:
Updated the xbutil/xbmgmt wrapper scripts to select the "classic" (e.g., xbutil/xbmgmt) or "nextgen" (e.g., xbutil2/xbmgmt2) tools.

This selection is done via the environment variable XRT_TOOLS_NEXTGEN, where the default value is false.   The selection algorithm is as follows:
1. If the XRT_TOOLS_NEXTGEN is not set, then xbutil/xbmgmt is used.
2. If the XRT_TOOLS_NEXTGEN is set to false, then xbutil/xbmgmt is used.
3. If the XRT_TOOLS_NEXTGEN is set to true, then xbutil2/xbmgmt is used.

Note: If the user wish to specify that the "nextgen" tool should be used, the "--new" option can be used.


src/runtime_src/core/tools/xbmgmt2/xbmgmt
```diff
@@ -10,6 +10,16 @@ fi
 # Working variables
 XRT_PROG=xbmgmt
 
+# Determine which version of xbmgmt should be used
+# Check to see environment variable is NOT set
+if [ -z ${XRT_TOOLS_NEXTGEN+x} ]; then 
+  XRT_PROG=xbmgmt
+else
+  if [ "${XRT_TOOLS_NEXTGEN,,}" = "true" ]; then
```
stsoe, line 18: isn't this called 'elif' ?

stsoe (commented):
Minor comment regarding else if.

What's the motivation behind this new env var when it defaults to false and when --new switch can as well be used?

src/runtime_src/core/tools/xbmgmt2/xbmgmt
rozumx: Yes, it is....I will update it.

Regarding the new ENV var and --new, if the user specified the "--new" option, it will (by design) take precedence over the variable name.

rozumx: Done

PR #4221, opened 2020-10-13
CR-1075952 Buffer flags in C++ and C

stsoe:
Amend #4195 this time with buffer flags as enum class for C++ APIs.
For C APIs provide XRT_BO_FLAGS equivalent.


dayeh-xilinx:
retest this please

PR #4225, opened 2020-10-13
xclbinutil: Bug Fix - Updated grouping algorithm to remove duplicate groupings

rozumx:
Work Done
---------
+ Bug Fix: Fix an issue where there could be duplicate grouping entries in the GROUP_TOPOLOGY section.  This issue occurred if 2 or more kernel share a common grouping of memories.  The code has been updated to examine the existing memory groupings looking for duplicates.  If a previous one is found, it is used.
+ Bug Fix: Fix an issue where the tag name for the Memory Groups were using the connectivity index values instead of the memory index values.


stsoe (approved):
Looks good.  I will give it try before we merge.

PR #4234, opened 2020-10-14
CR-1080069  XMA should use group topology and group connectivity

stsoe:
This is related to parent CR-1079649. XMA must use group sections in
the xclbin if they are there.


sarab96 (approved):
Looks good. Other checks on 64 entries in topology will be added later on.

PR #4242, opened 2020-10-15
Fixes in handling of DeviceIntf amongst different xdp trace plugins

IshitaGhosh:


jvillarre (approved):
Changes look good and moves interface into a central location.

PR #4243, opened 2020-10-15
Xrt api fix

uday610:
The changes based on recent API fixes

1) Example not to use: auto device = xrt::device(0); 
    Instead use 
    unsigned int dev_index = 0;
    auto device = xrt::device(dev_index);

2) Removed incorrect statement:
    "If the user-provided pointer is not aligned to 4K boundary, XRT internally copies the data to align it at 4K boundary." 

3) Changed all occarance of XCL_BO_* to XRT_BO_*

4) For C++ xrt::bo no buffer flag is used for normal buffer

5) Removed buffer size and offset from xrt::bo::write, xrt::bo::sync, etc, Mention about overloaded options

6) Mention enum  exclusive access xrt::kernel::cu_access_mode::exclusive instead of true 

7) Introduced offset access APIs

8) Mention about std::chrono::milliseconds

### DIFF
--- a/src/runtime_src/doc/toc/xrt_native_apis.rst
+++ b/src/runtime_src/doc/toc/xrt_native_apis.rst
@@ -474,7 +474,7 @@ The API ``xrtPLKernelOpen`` opens a kernel's CU in a shared mode so that the CU 
      
            xrtKernelHandle kernel = xrtPLKernelOpenExclusive(device, xclbin_uuid, "name");
 
-**C++**: When the ``xrt::kernel`` constructor is called with an additional boolean argument set as true, it opens CU in exclusive mode and returns the kernel object.    
+**C++**: In C++, change from here ``xrt::kernel`` constructor is called with an additional boolean argument set as true, it opens CU in exclusive mode and returns the kernel object.    
 
 .. code:: c++
       :number-lines: 39
--- a/src/runtime_src/doc/toc/xrt_native_apis.rst
+++ b/src/runtime_src/doc/toc/xrt_native_apis.rst
@@ -534,7 +534,7 @@ To read and write from the AXI-Lite register space corresponding to a CU, the CU
 
 In the above code block
 
-              - The compute unit named "foo_1" (name syntax: "kernel_name:{cu_name}") is opened exclusively.
+              - The CU named "foo_1" (name syntax: "kernel_name:{cu_name}") is opened exclusively.
               - The Register Read/Write operation is performed. 
               - Closed the kernel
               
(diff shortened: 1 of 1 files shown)


stsoe (approved):

src/runtime_src/doc/toc/xrt_native_apis.rst
```diff
@@ -97,7 +97,8 @@ The above code block shows
 .. code:: c++
       :number-lines: 10
            
-           auto device = xrt::device(0);
+           unsigned int dev_index = 0;
+           auto device = xrt::device(dev_index);
```
keryell, line 101: Why `auto` here?
```c++
xrt::device device(dev_index);
```
or even clearer
```c++
xrt::device device { dev_index };
```

src/runtime_src/doc/toc/xrt_native_apis.rst
```diff
@@ -238,17 +245,17 @@ Code example of transferring data from the host to the device
 .. code:: c++
       :number-lines: 20    
            
-           auto input_buffer = xrt::bo(device, buffer_size_in_bytes, XCL_BO_FLAGS_NONE, bank_grp_idx_0);
+           auto input_buffer = xrt::bo(device, buffer_size_in_bytes, bank_grp_idx_0);
```
keryell, line 248: Do not use `auto` when it makes the code longer...
I thought `bo` was renamed `buffer`.

src/runtime_src/doc/toc/xrt_native_apis.rst
```diff
@@ -550,8 +560,46 @@ In the above code block
...
+
+           // Assume foo has 3 arguments, a,b,c (arg 0, arg 1 and arg 2 respectively) 
+           
+           xrtKernelHandle kernel = xrtPLKernelOpenExclusive(device, xclbin_uuid, "foo:{foo_1}");
+           uint32_t arg_c_offset = xrtKernelArgOffset(kernel, 2);
+ 
+
+**C++**: The equivalent C++ API example
+
+.. code:: c
```
keryell, line 594: Clearly not C

keryell (commented):
I have the feeling I did this review somewhere else already...

src/runtime_src/doc/toc/xrt_native_apis.rst
stsoe: Agreed with {}.  As I said before, personally I prefer auto and types to the right.  The code reads much better that way IMO.  But C++ is inconsistent in this way, which is evident on line 100.

src/runtime_src/doc/toc/xrt_native_apis.rst
stsoe: No, there will be no renaming.

keryell:
```c++
using buffer = xrt::bo;
```
?

stsoe: I considered that but decided against it.    I prefer some consistency and 'BO' for 'buffer object' is a pretty engrained concept in XRT.   There is also the C-APIs that use BO, which would have to be duplicated to Buffer.   It's too much of a mess and leaves inconsistent examples.  In hindsight, maybe buffer should have been used from the onset.  Next generation APIs ...

keryell: OK. Anyway, for me it will be abstracted as a `sycl::buffer`, so the end-user will not see it...

PR #4250, opened 2020-10-16
XRT-791 support switching xclbin which contains soft kernel

houlz0507:


src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -2707,6 +2765,7 @@ ps_ert_submit(struct sched_cmd *cmd)
 	switch (opcode(cmd)) {
 	case ERT_CONFIGURE:
 		SCHED_DEBUG("<- %s (configure)\n", __func__);
+		kill_soft_kernel(cmd);
```
larry9523, line 2768: This is OK. But I think a better place would be in config function itself (around line 779). That is the place we handle reconfiguration in ert mode with comment.

larry9523 (changes requested):
Looks good. A minor change suggestion requested.

src/runtime_src/core/edge/drm/zocl/sched_exec.c
houlz0507: done

larry9523 (approved):

mamin506 (approved):

PR #4254, opened 2020-10-17
More doc updates and cleanups

sonals:
Removed streaming APIs from ToC


src/runtime_src/doc/toc/index.rst
```diff
@@ -33,7 +33,6 @@ FPGA. The key user APIs are defined in ``xrt.h`` header file.
    p2p.rst
    m2m.rst
    sb.rst
-   streaming_data_support.rst
```
keryell, line 36: Is this still referenced from somewhere else?

PR #4255, opened 2020-10-17
Switch ReportAsyncError() to xrt::error APIs

stsoe:
The C++ error APIs propagate errors as exceptions, which gives caller
chance to catch and take corrective action.  The C-APIs, by nature of
C, return error codes after printing the exception message.


AShivangi (approved):
Thank you for making the changes :)

PR #4264, opened 2020-10-18
xbutil2/xbmgmt2 : Fixed various bash reported issues

rozumx:
Work Done
---------
+ Renamed report type "verbose" to "all"
+ Enabled the build flow to "optionally" add "native" reports.  Doing so prevents "native" reports from being options for "non-native" builds.
+ Enabled the build flow to "optionally" add "native" commands.  xbutil validate is now a native command.
+ Removed CMakeLists.txt dead code

### DIFF
--- a/src/runtime_src/core/tools/common/XBHelpMenus.cpp
+++ b/src/runtime_src/core/tools/common/XBHelpMenus.cpp
@@ -545,7 +545,7 @@ XBUtilities::create_suboption_list_string(const VectorPairStrings &_collection)
 
 std::string 
 XBUtilities::create_suboption_list_string( const ReportCollection &_reportCollection, 
-                                           bool _addVerboseOption)
+                                           bool _addAllOption)
 {
   VectorPairStrings reportDescriptionCollection;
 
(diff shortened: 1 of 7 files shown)


rozumx:
### Code Reviewers
Please do examine the xbutil reports and validate that the "NATIVE" (e.g., reports not supported by the Edge targets) is indeed correct set.

chvamshi-xilinx (approved):
Thanks for making these changes, Looks good. 
One question when i discussed with @Himanshu-xilinx is about non-native specific sub commands.
We may need to have another flag which are not applicable for native. Approving this PR. You can create one more PR for this.
example: reportAIe.

AShivangi (approved):
Should host report be a native report as well?

rozumx:
@chvamshi-xilinx and @Himanshu-xilinx, I like the idea of support non-native and in fact the code already support that indirectly via the  ENABLE_NATIVE_SUBCMDS_AND_REPORTS define value (though it isn't obvious and is rather restrictive).  

Let me think about it today and see if a better solution could be obtained.

Himanshu-xilinx (approved):
changes looks good but ReportAIE is embedded only report currently so it should not be in common reports

PR #4265, opened 2020-10-18
refactor-2

sarab96:
class for xma session. refactor for cpp

### DIFF
--- a/src/xma/include/lib/xma_session.hpp
+++ b/src/xma/include/lib/xma_session.hpp
@@ -31,8 +31,8 @@ namespace plg {
  */
 class session
 {
-protected:
-    void*        session_signature;
+private:
+    void*        session_signature{nullptr};
     int32_t         session_id;
     /** Subclass this session is a part of */
     XmaSessionType session_type;
@@ -43,20 +43,22 @@ protected:
     /** Private plugin data attached to a specific kernel session. Allocated
     by XMA prior to calling plugin init() and freed automatically as part of
     close. */
-    void          *plugin_data;
+    void          *plugin_data{nullptr};
     /** Private stats data attached to a specific session. This field is
     allocated and managed by XMA for each session type. */
-    void          *stats;
+    void          *stats{nullptr};
   
-
-  session(int s_id, XmaSessionType s_type, int32_t c_id);
+  ~session();
 
 public:
   int32_t
-  init(int32_t dev_idx, int32_t cu_idx, std::string cu_name);
+  init(int32_t dev_idx, int32_t cu_idx, const std::string& cu_name);
 
   int32_t
   alloc_buf() const;
+
+  session(int s_id, XmaSessionType s_type, int32_t c_id);
+  void destroy();
 
 }; //class session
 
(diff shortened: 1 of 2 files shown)


src/xma/include/lib/xma_session.hpp
```diff
@@ -0,0 +1,64 @@
...
+    /** Private stats data attached to a specific session. This field is
+    allocated and managed by XMA for each session type. */
+    void          *stats;
+  
+
+  session(int s_id, XmaSessionType s_type, int32_t c_id);
+
+public:
+  int32_t
+  init(int32_t dev_idx, int32_t cu_idx, std::string cu_name);
```
stsoe, line 56: const std::string&
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#Rf-in

src/xma/include/lib/xma_session.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+namespace xma_core {
+namespace plg {
+
+/**
+ * class session : abstraction for xma session
+ */
+class session
+{
+protected:
```
stsoe, line 34: I am wary about protected.  But occasionally it is needed and make sense, but I don't have the context here.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#Rnr-protected-data

src/xma/src/xmaplugin/xma_session.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+namespace xma_core {
+namespace plg {
+
+session::session(int s_id, XmaSessionType s_type, int32_t c_id)
+{
+  session_id = s_id;
+  session_type = s_type;
+  channel_id = c_id;
+  session_signature = nullptr;
```
stsoe, line 27: Prefer in class initialization, in this case all ` = nullptr` can be in class.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#Rc-in-class-initializer

src/xma/src/xmaplugin/xma_session.cpp
```diff
@@ -0,0 +1,50 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+#include "lib/xma_session.hpp"
+#include "app/xmaerror.h"
+
+namespace xma_core {
+namespace plg {
+
+session::session(int s_id, XmaSessionType s_type, int32_t c_id)
```
stsoe, line 22: Prefer initialization to assignment:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#Rc-initialize

stsoe (changes requested):

src/xma/include/lib/xma_session.hpp
sarab96: Done

src/xma/src/xmaplugin/xma_session.cpp
sarab96: Agree. Done

src/xma/src/xmaplugin/xma_session.cpp
sarab96: Agree. Done

src/xma/include/lib/xma_session.hpp
sarab96: Thought about it. Use of private and getter/setter is better. Done

stsoe (approved):

PR #4266, opened 2020-10-19
update cmc register without affecting it's previous state & other fields

rajkumar-xilinx:


xdavidz (approved):
Please add comments like:
we only apply 24:0 address, and keep original 31-24 bits while applying 28 and 28 bit?

rajkumar-xilinx:
> Please add comments like:
> we only apply 24:0 address, and keep original 31-24 bits while applying 28 and 28 bit?

Done, thanks.

chvamshi-xilinx (approved):

PR #4267, opened 2020-10-19
Added support for HBM backward compatibility of existing application

saifuddin-xilinx:
- This supports the backward compatibility of the existing applications.
- Application code with EXT pointer will still work with this change. 
- It is advisable not to use EXT pointer in the application.

### DIFF
--- a/src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx
+++ b/src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx
@@ -869,6 +869,14 @@ XclBinUtilities::createMemoryBankGrouping(XclBin & xclbin)
       // Transform and group the memories
       transformMemoryBankGroupingCollections(connectivity, groupTopology, groupConnectivity);
 
+      // Perform some DRC checks on the memory grouping and connectivity produced before marging to group connectivity
+      validateMemoryBankGroupEntries((unsigned int) memTopology.size(), groupTopology, groupConnectivity);
+
+      // Merge connectivity information into the group connectivity
+      for (auto & connection : connectivity) {
+	      groupConnectivity.push_back(connection);
+      }
+
       // Re-create the property tree, create and re-populate the Group Connectivity section, and add it.
       {
         boost::property_tree::ptree ptConnection;
@@ -908,9 +916,6 @@ XclBinUtilities::createMemoryBankGrouping(XclBin & xclbin)
     pGroupTopologySection->readJSONSectionImage(ptTop);
     xclbin.addSection(pGroupTopologySection);
   }
-
-  // Perform some DRC checks on the memory grouping and connectivity produced
-  validateMemoryBankGroupEntries((unsigned int) memTopology.size(), groupTopology, groupConnectivity);
 }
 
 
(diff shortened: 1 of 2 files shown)


stsoe (approved):
Changes are good.  
Before we merge, let me run standard board tests.

stsoe:
Tests are passing just fine.

src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx
```diff
@@ -869,6 +869,14 @@ XclBinUtilities::createMemoryBankGrouping(XclBin & xclbin)
       // Transform and group the memories
       transformMemoryBankGroupingCollections(connectivity, groupTopology, groupConnectivity);
 
+      // Perform some DRC checks on the memory grouping and connectivity produced before marging to group connectivity
+      validateMemoryBankGroupEntries((unsigned int) memTopology.size(), groupTopology, groupConnectivity);
```
rozumx, line 873: The reason why this validation was done at the end was to validate the contents and connections of the two GROUP sections.  Now that is no longer the case.

Why couldn't the validation method be updated?

src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx
```diff
@@ -869,6 +869,14 @@ XclBinUtilities::createMemoryBankGrouping(XclBin & xclbin)
...
+
+      // Merge connectivity information into the group connectivity
+      for (auto & connection : connectivity) {
```
rozumx, line 876: So now we have duplicate entries.

Couldn't find it, how did you make sure that the indexes GROUP memory indexes are correct?  Are you assuming that the memory indexes are at the start.  If so this causes a hard-coded binding which should NOT be there.

Also, could you refer us to the various use-cases that need this data.  

Again, this looks very dangerous.

src/runtime_src/xocl/core/device.cpp
```diff
@@ -729,8 +729,9 @@ get_cu_memidx() const
       for (auto& cu : get_cu_range())
         mask &= cu->get_memidx_intersect();
 
-      // select first common memory bank index if any
-      for (size_t idx=0; idx<mask.size(); ++idx) {
+      // Select first common Group index if present prior to bank index.
+      // Traverse from the higher order of the mask as groups comes in higher order
+      for (int idx=mask.size() - 1; idx >= 0; --idx) {
```
rozumx, line 734: There should not be any order dependencies.

rozumx (changes requested):
There are several array dependencies and DRC checks are being bypassed which greatly concerns me.  

Please do schedule a meeting with Soren and myself to review the use-cases and possible solutions.

rozumx (approved):
After a discussion with Soren, Saif, Vamshi, and Uday we understand why these changes were made.  In addition, where we are in the release and to minimize the release risk, these change will remain as is.  

Post 2020.2, we all committed to re-examining how the grouping is performed and to create a more robust solution.

chvamshi-xilinx (approved):

PR #4272, opened 2020-10-19
Reduce rtp port name search complexity

larry9523:
This PR addressed the following issues
1.
We are currently using vector to store RTP port and when a port is updated, we use
    auto rtp = std::find_if(rtps.begin(), rtps.end(), auto it = rtps.find(port);
            [port](rtp_type it) { return it.name.compare(port) == 0; });
to match a given port name. 

This is time consuming when there are as much as 400 rtp ports used.  In this PR, we changed vector to unordered_map to store rtp ports so that we can theoretically reduce the search complexity to O(1)

2.
For interface xrtGraphWait and xrtGraphWaitDone, we return successfully if the graph is already in suspend state or stop state.

### DIFF
--- a/src/runtime_src/core/edge/user/aie/graph.cpp
+++ b/src/runtime_src/core/edge/user/aie/graph.cpp
@@ -79,7 +79,8 @@ graph_type(std::shared_ptr<xrt_core::device> dev, const uuid_t, const std::strin
       rtp.selector_row += 1;
       rtp.ping_row += 1;
       rtp.pong_row += 1;
-      rtps.insert(std::make_pair(rtp.name, std::move(rtp)));
+      std::string port_name(rtp.name);
+      rtps.emplace(std::move(port_name), std::move(rtp));
     }
 
     state = graph_state::reset;
--- a/src/runtime_src/core/edge/user/aie/graph.cpp
+++ b/src/runtime_src/core/edge/user/aie/graph.cpp
@@ -79,7 +79,7 @@ graph_type(std::shared_ptr<xrt_core::device> dev, const uuid_t, const std::strin
       rtp.selector_row += 1;
       rtp.ping_row += 1;
       rtp.pong_row += 1;
-      rtps.emplace_back(std::move(rtp));
+      rtps.insert(std::make_pair(rtp.name, std::move(rtp)));
     }
 
     state = graph_state::reset;
@@ -193,6 +193,9 @@ void
 graph_type::
 wait_done(int timeout_ms)
 {
+    if (state == graph_state::stop)
+      return;
+
     if (state != graph_state::running)
       throw xrt_core::error(-EINVAL, "Graph '" + name + "' is not running, cannot wait");
 
@@ -201,8 +204,6 @@ wait_done(int timeout_ms)
     /*
      * We are using busy waiting here. Until every tile in the graph
      * is done, we keep polling each tile.
-     *
-     * TODO Will register AIE event for tile done.
      */
     while (1) {
         uint8_t done;
@@ -243,6 +244,9 @@ void
 graph_type::
 wait()
 {
+    if (state == graph_state::stop)
+        return;
+
     if (state != graph_state::running)
         throw xrt_core::error(-EINVAL, "Graph '" + name + "' is not running, cannot wait");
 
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/edge/user/aie/graph.cpp
```diff
@@ -79,7 +79,7 @@ graph_type(std::shared_ptr<xrt_core::device> dev, const uuid_t, const std::strin
       rtp.selector_row += 1;
       rtp.ping_row += 1;
       rtp.pong_row += 1;
-      rtps.emplace_back(std::move(rtp));
+      rtps.insert(std::make_pair(rtp.name, std::move(rtp)));
```
keryell, line 82:
```suggestion
      rtps.emplace(rtp.name, std::move(rtp));
```

src/runtime_src/core/edge/user/aie/graph.cpp
```diff
@@ -398,34 +405,33 @@ void
...
 update_rtp(const std::string& port, const char* buffer, size_t size)
 {
-    auto rtp = std::find_if(rtps.begin(), rtps.end(),
-            [port](rtp_type it) { return it.name.compare(port) == 0; });
-
-    if (rtp == rtps.end())
+    auto it = rtps.find(port);
+    if (it == rtps.end())
       throw xrt_core::error(-EINVAL, "Can't update graph '" + name + "': RTP port '" + port + "' not found");
+    auto& rtp = it->second;
```
keryell, line 411: I love this laziness! :-)

src/runtime_src/core/edge/user/aie/graph.h
```diff
@@ -25,6 +25,7 @@
 #include "core/common/device.h"
 #include <string>
 #include <vector>
+#include <unordered_map>
```
keryell, line 28: Are you using an `unordered_sort` on the headers here? ;-)

keryell (changes requested):

src/runtime_src/core/edge/user/aie/graph.cpp
stsoe: Same code fewer key strokes.  But btw is either construct really safe?  Is there any chance rtp.name could have been moved before it is copied, for sure order of evaluation is not defined here. There is enough doubt in my mind that I think a copy of the string should have been done outside insert or emplace, the another std::move on the copied string.

Better, a std::set should have been used with proper compare defined.

stsoe (changes requested):
See comment on insert/emplace.

src/runtime_src/core/edge/user/aie/graph.cpp
larry9523: Done

stsoe (approved):

src/runtime_src/core/edge/user/aie/graph.h
keryell: It would have been nicer with `#include <unordered_map>` before `#include <vector>`

src/runtime_src/core/edge/user/aie/graph.cpp
keryell: > Same code fewer key strokes. But btw is either construct really safe? Is there any chance rtp.name could have been moved before it is copied, for sure order of evaluation is not defined here. There is enough doubt in my mind that I think a copy of the string should have been done outside insert or emplace, then another std::move on the copied string.

Ouch, you are right!

> Better, a std::set should have been used with proper compare defined.

Brilliant.

src/runtime_src/core/edge/user/aie/graph.h
larry9523: Point taken. Will sort it in my next PR.

stsoe: Never thought about sorting #includes but actually that's a good idea.  I will follow that.
Do make sure all XRT local includes are grouped before system includes though.

keryell: Sorry for using too much convoluted humor too often...
Sorting headers is common in a lot of projects and coding styles to have a quicker idea about what are the headers in use.

PR #4286, opened 2020-10-21
skip scheduler init removed

sarab96:
removed as stress test fail in multiple process case. As no lock into load_xclbin and scheduler::init in user space


maxzhen (changes requested):
you also want to remove int shim::get_dev_xclbin_uuid(xuid_t out) and the include of "core/include/experimental/xclbin_util.h" which were added by your previous change.

maxzhen (approved):

maxzhen:
Travis CI seems getting stuck. Merging it now since it passed pipeline. This fix is critical for a few regressions reported by DSV.

PR #4288, opened 2020-10-21
allow enable ert only when ert subdev exist

mamin506:
Since ERT is by default enable in xrt.ini and U.2 shell doesn't support ERT, it would cause kernel panic without this fix.


maxzhen (changes requested):
please investigate why we don't have ert enabled on u.2. it does not sound right.

houlz0507 (approved):

mamin506:
> please investigate why we don't have ert enabled on u.2. it does not sound right.

This is because "XOCL_DEVINFO_ERT_USER" is missed in USER_RES_DSA52_U2. I added it and test on Raptor2 and legacy shell.

maxzhen (approved):

PR #4289, opened 2020-10-21
CR-1080069: use HBM mem group as default when present

sarab96:
use HBM mem group as default when present

### DIFF
--- a/src/xma/src/xmaapi/xmaxclbin.cpp
+++ b/src/xma/src/xmaapi/xmaxclbin.cpp
@@ -286,7 +286,7 @@ static int get_xclbin_mem_topology(const char *buffer, XmaXclbinInfo *xclbin_inf
         throw std::runtime_error("Could not find MEM TOPOLOGY in xclbin file");
     }
     const char *data3 = &buffer[mem_hdr->m_sectionOffset];
-    const mem_topology *mem_topo1 = reinterpret_cast<const mem_topology *>(data3);
+    auto mem_topo1 = reinterpret_cast<const mem_topology *>(data3);
     mem_topology *mem_topo2 = const_cast<mem_topology*>(mem_topo1);
     xclbin_info->has_mem_groups = false;
 
--- a/src/xma/src/xmaapi/xmaxclbin.cpp
+++ b/src/xma/src/xmaapi/xmaxclbin.cpp
@@ -286,6 +286,11 @@ static int get_xclbin_mem_topology(const char *buffer, XmaXclbinInfo *xclbin_inf
         const char *data = &buffer[ip_hdr->m_sectionOffset];
         const mem_topology *mem_topo = reinterpret_cast<const mem_topology *>(data);
         auto& xma_mem_topology = xclbin_info->mem_topology;
+        const axlf_section_header *mem_grp_hdr = xclbin::get_axlf_section(xclbin, ASK_GROUP_TOPOLOGY);
+        xclbin_info->has_mem_groups = false;
+	if (mem_grp_hdr) {
+            xclbin_info->has_mem_groups = true;
+	}
 
         xclbin_info->number_of_mem_banks = mem_topo->m_count;
         xma_logmsg(XMA_DEBUG_LOG, XMAAPI_MOD, "MEM TOPOLOGY - %ud banks ",xclbin_info->number_of_mem_banks);
@@ -375,19 +380,30 @@ int xma_xclbin_info_get(const char *buffer, XmaXclbinInfo *info)
     return XMA_SUCCESS;
 }
 
-int xma_xclbin_map2ddr(uint64_t bit_map, int32_t* ddr_bank)
+int xma_xclbin_map2ddr(uint64_t bit_map, int32_t* ddr_bank, bool has_mem_grps)
 {
     //64 bits based on MAX_DDR_MAP = 64
     int ddr_bank_idx = 0;
-    while (bit_map != 0)
-    {
-        if (bit_map & 1)
-        {
-            *ddr_bank = ddr_bank_idx;
-            return XMA_SUCCESS;
+    if (!has_mem_grps) {
+        while (bit_map != 0) {
+            if (bit_map & 1) {
+                *ddr_bank = ddr_bank_idx;
+                return XMA_SUCCESS;
+            }
+            ddr_bank_idx++;
+            bit_map = bit_map >> 1;
         }
-        ddr_bank_idx++;
-        bit_map = bit_map >> 1;
+    } else {//For Memory Groups use last group as default memory group; For HBM groups
+        ddr_bank_idx = 63;
+        uint64_t tmp_int = 1ULL << 63;
+        while (bit_map != 0) {
+            if (bit_map & tmp_int) {
+                *ddr_bank = ddr_bank_idx;
+                return XMA_SUCCESS;
+            }
+            ddr_bank_idx--;
+            bit_map = bit_map << 1;
+        }
     }
     *ddr_bank = -1;
     return XMA_ERROR;
--- a/src/xma/src/xmaapi/xmaxclbin.cpp
+++ b/src/xma/src/xmaapi/xmaxclbin.cpp
@@ -315,7 +315,8 @@ static int get_xclbin_mem_topology(const char *buffer, XmaXclbinInfo *xclbin_inf
         temp_mem_topology.m_size = mem_topo->m_mem_data[i].m_size;
         temp_mem_topology.m_base_address = mem_topo->m_mem_data[i].m_base_address;
         //m_tag is 16 chars
-        temp_mem_topology.m_tag = std::string((char*)mem_topo->m_mem_data[i].m_tag);
+        auto tmp_ptr = mem_topo->m_mem_data[i].m_tag;
+        temp_mem_topology.m_tag = std::string(tmp_ptr, std::find(tmp_ptr, tmp_ptr+16, '\0'));//magic 16 from xclbin.h
         xma_logmsg(XMA_DEBUG_LOG, XMAAPI_MOD, "index=%d, tag=%s, type = %ud, used = %ud, size = %lx, base = %lx ",
                i,temp_mem_topology.m_tag.c_str(), temp_mem_topology.m_type, temp_mem_topology.m_used,
                temp_mem_topology.m_size, temp_mem_topology.m_base_address);
--- a/src/xma/src/xmaapi/xmaxclbin.cpp
+++ b/src/xma/src/xmaapi/xmaxclbin.cpp
@@ -293,7 +293,7 @@ static int get_xclbin_mem_topology(const char *buffer, XmaXclbinInfo *xclbin_inf
     const axlf_section_header *mem_grp_hdr = xrt_core::xclbin::get_axlf_section(xclbin, ASK_GROUP_TOPOLOGY);
     if (mem_grp_hdr) {
         const char *data2 = &buffer[mem_grp_hdr->m_sectionOffset];
-        const mem_topology *mem_grp_topo = reinterpret_cast<const mem_topology *>(data2);
+        auto mem_grp_topo = reinterpret_cast<const mem_topology *>(data2);
         if (mem_grp_topo->m_count > mem_topo1->m_count) {
             xclbin_info->has_mem_groups = true;
             mem_topo2 = const_cast<mem_topology*>(mem_grp_topo);
--- a/src/xma/src/xmaapi/xmaxclbin.cpp
+++ b/src/xma/src/xmaapi/xmaxclbin.cpp
@@ -280,43 +280,47 @@ static int get_xclbin_mem_topology(const char *buffer, XmaXclbinInfo *xclbin_inf
 {
     const axlf *xclbin = reinterpret_cast<const axlf *>(buffer);
 
-    const axlf_section_header *ip_hdr = xrt_core::xclbin::get_axlf_section(xclbin, ASK_GROUP_TOPOLOGY);
-    if (ip_hdr)
-    {
-        const char *data = &buffer[ip_hdr->m_sectionOffset];
-        const mem_topology *mem_topo = reinterpret_cast<const mem_topology *>(data);
-        auto& xma_mem_topology = xclbin_info->mem_topology;
-        const axlf_section_header *mem_grp_hdr = xclbin::get_axlf_section(xclbin, ASK_GROUP_TOPOLOGY);
-        xclbin_info->has_mem_groups = false;
-	if (mem_grp_hdr) {
-            xclbin_info->has_mem_groups = true;
-	}
+    //const axlf_section_header *ip_hdr = xrt_core::xclbin::get_axlf_section(xclbin, ASK_GROUP_TOPOLOGY);
+    const axlf_section_header *mem_hdr = xclbin::get_axlf_section(xclbin, MEM_TOPOLOGY);
+    if (!mem_hdr) {
+        xma_logmsg(XMA_ERROR_LOG, XMAAPI_MOD, "Could not find MEM TOPOLOGY in xclbin ");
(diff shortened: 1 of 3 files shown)


src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -286,6 +286,11 @@ static int get_xclbin_mem_topology(const char *buffer, XmaXclbinInfo *xclbin_inf
         const char *data = &buffer[ip_hdr->m_sectionOffset];
         const mem_topology *mem_topo = reinterpret_cast<const mem_topology *>(data);
         auto& xma_mem_topology = xclbin_info->mem_topology;
+        const axlf_section_header *mem_grp_hdr = xclbin::get_axlf_section(xclbin, ASK_GROUP_TOPOLOGY);
```
stsoe, line 289: This is always going to return a section.  For old pre 2020.2 xclbins it will return MEM_TOPOLOGY, for 2020.2 xclbins it will return the GROUP_TOPOLOGY section which always exists whether grouping it used or not.

src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -286,6 +286,11 @@ static int get_xclbin_mem_topology(const char *buffer, XmaXclbinInfo *xclbin_inf
...
+        xclbin_info->has_mem_groups = false;
+	if (mem_grp_hdr) {
```
stsoe, line 291: This will always be true.  The key is you don't really know if the xclbin is grouping banks or not.  Not unless you iterate all mem_data entries and search for the bank 'tag' that is used for the groups; and doing this is not really a good idea.

The xclbin produced in 2020.2 is composed by xclbinutil and will always have a GROUP_TOPOLOGY section, but this section may not have groups and could be just a copy of MEM_TOPOLOGY.

src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -375,19 +380,30 @@ int xma_xclbin_info_get(const char *buffer, XmaXclbinInfo *info)
...
+    } else {//For Memory Groups use last group as default memory group; For HBM groups
+        ddr_bank_idx = 63;
+        uint64_t tmp_int = 1ULL << 63;
+        while (bit_map != 0) {
+            if (bit_map & tmp_int) {
+                *ddr_bank = ddr_bank_idx;
+                return XMA_SUCCESS;
+            }
+            ddr_bank_idx--;
+            bit_map = bit_map << 1;
```
stsoe, line 405: Ideally you should always search from the back, but given that I have seen a decoder example where the xclbin was modified to manually expand the *first* bank entry, searching from back would break this use case whether we deem it valid or not.

stsoe (approved):
Hi.  Please check my comments.  The implementation doesn't end up doing what you have in mind.

stsoe (changes requested):
See comments.  The implementation doesn't end up doing what you have in mind.

src/xma/src/xmaapi/xmaxclbin.cpp
sarab96: Wrong. This is from xclbin namespace and not from xrt::xclbin

src/xma/src/xmaapi/xmaxclbin.cpp
sarab96: ok. Better would be if xlcbinutil doesn't make a group_topology which is just a copy of mem_topology. That copy doesn't serve any new purpose. I will make a change in this.

src/xma/src/xmaapi/xmaxclbin.cpp
stsoe: Ok, I see, but I am not entirely wrong :-)   ASK_GROUP_TOPOLOGY will always be present in 2020.2 xclbins regardless of grouping or not, so it can not really be used to determine if you should search from back or not.   My point is that you should probably always search from back.

stsoe: Also using xclbin:: directly is not super cool either, since xrt:;xclbin does a check on xrt.ini to see if any grouping (present or not) should be ignored, in which case it returns MEM_TOPOLOGY and CONNECTIVITY instead of GROUP_TOPOLOGY and GROUP_CONNECTIVITY.

src/xma/src/xmaapi/xmaxclbin.cpp
sarab96: Modified

sarab96:
Modified to handle this

src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -280,38 +280,47 @@ static int get_xclbin_mem_topology(const char *buffer, XmaXclbinInfo *xclbin_inf
...
+    if (!mem_hdr) {
+        xma_logmsg(XMA_ERROR_LOG, XMAAPI_MOD, "Could not find MEM TOPOLOGY in xclbin ");
+        throw std::runtime_error("Could not find MEM TOPOLOGY in xclbin file");
+    }
+    const char *data3 = &buffer[mem_hdr->m_sectionOffset];
+    const mem_topology *mem_topo1 = reinterpret_cast<const mem_topology *>(data3);
+    mem_topology *mem_topo2 = const_cast<mem_topology*>(mem_topo1);
+    xclbin_info->has_mem_groups = false;
+
+    const axlf_section_header *mem_grp_hdr = xclbin::get_axlf_section(xclbin, ASK_GROUP_TOPOLOGY);
```
stsoe, line 294: Please use xrt_core::xclbin::get_axlf_section such that xrt.ini options are respected.  Your other code will still work without changes, because asking for GROUP_TOPOLOGY will return MEM_TOPOLOGY if there is not GROUP_TOPOLOGY and asking for MEM_TOPOLOGY will indeed always return MEM_TOPOLOGY.

stsoe (changes requested):

src/xma/src/xmaapi/xmaxclbin.cpp
sarab96: Done

stsoe (approved):

src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -280,38 +280,46 @@ static int get_xclbin_mem_topology(const char *buffer, XmaXclbinInfo *xclbin_inf
...
-                   i,temp_mem_topology.m_tag.c_str(), temp_mem_topology.m_type, temp_mem_topology.m_used,
-                   temp_mem_topology.m_size, temp_mem_topology.m_base_address);
-            xma_mem_topology.emplace_back(std::move(temp_mem_topology));
+    const axlf_section_header *mem_hdr = xrt_core::xclbin::get_axlf_section(xclbin, MEM_TOPOLOGY);
+    if (!mem_hdr) {
+        xma_logmsg(XMA_ERROR_LOG, XMAAPI_MOD, "Could not find MEM TOPOLOGY in xclbin ");
+        throw std::runtime_error("Could not find MEM TOPOLOGY in xclbin file");
+    }
+    const char *data3 = &buffer[mem_hdr->m_sectionOffset];
+    const mem_topology *mem_topo1 = reinterpret_cast<const mem_topology *>(data3);
```
keryell, line 289: DRY:
```suggestion
    auto mem_topo1 = reinterpret_cast<const mem_topology *>(data3);
```

src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -280,38 +280,46 @@ static int get_xclbin_mem_topology(const char *buffer, XmaXclbinInfo *xclbin_inf
...
+    mem_topology *mem_topo2 = const_cast<mem_topology*>(mem_topo1);
```
keryell, line 290: DRY:
```suggestion
    auto mem_topo2 = const_cast<mem_topology*>(mem_topo1);
```

src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -280,38 +280,46 @@ static int get_xclbin_mem_topology(const char *buffer, XmaXclbinInfo *xclbin_inf
...
+    xclbin_info->has_mem_groups = false;
+
+    const axlf_section_header *mem_grp_hdr = xrt_core::xclbin::get_axlf_section(xclbin, ASK_GROUP_TOPOLOGY);
```
keryell, line 293: Here it looks less than a DRY issue, so probably this writing is fine

src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -280,38 +280,46 @@ static int get_xclbin_mem_topology(const char *buffer, XmaXclbinInfo *xclbin_inf
...
+    if (mem_grp_hdr) {
+        const char *data2 = &buffer[mem_grp_hdr->m_sectionOffset];
+        const mem_topology *mem_grp_topo = reinterpret_cast<const mem_topology *>(data2);
```
keryell, line 296: DRY

src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -280,38 +280,46 @@ static int get_xclbin_mem_topology(const char *buffer, XmaXclbinInfo *xclbin_inf
...
+        if (mem_grp_topo->m_count > mem_topo1->m_count) {
+            xclbin_info->has_mem_groups = true;
+            mem_topo2 = const_cast<mem_topology*>(mem_grp_topo);
         }
     }
-    else
-    {
-        xma_logmsg(XMA_ERROR_LOG, XMAAPI_MOD, "Could not find MEM TOPOLOGY in xclbin ip_hdr=%p ", ip_hdr);
-        throw std::runtime_error("Could not find MEM TOPOLOGY in xclbin file");
+    const mem_topology *mem_topo = reinterpret_cast<const mem_topology *>(mem_topo2);
```
keryell, line 302: DRY

src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -280,38 +280,46 @@ static int get_xclbin_mem_topology(const char *buffer, XmaXclbinInfo *xclbin_inf
...
+    auto& xma_mem_topology = xclbin_info->mem_topology;
```
keryell, line 303: Ah, if you think about `auto` here probably there could be more `auto` above then.

src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -280,38 +280,46 @@ static int get_xclbin_mem_topology(const char *buffer, XmaXclbinInfo *xclbin_inf
...
+        throw std::runtime_error("XMA supports max of only " + std::to_string(MAX_DDR_MAP) + " mem banks");
+    }
+    for (int i = 0; i < mem_topo->m_count; i++) {
+        XmaMemTopology temp_mem_topology;
+        temp_mem_topology.m_type = mem_topo->m_mem_data[i].m_type;
+        temp_mem_topology.m_used = mem_topo->m_mem_data[i].m_used;
+        temp_mem_topology.m_size = mem_topo->m_mem_data[i].m_size;
+        temp_mem_topology.m_base_address = mem_topo->m_mem_data[i].m_base_address;
+        //m_tag is 16 chars
+        temp_mem_topology.m_tag = std::string((char*)mem_topo->m_mem_data[i].m_tag);
```
keryell, line 318:
```suggestion
        temp_mem_topology.m_tag = static_cast<char*>(mem_topo->m_mem_data[i].m_tag);
```
?

src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -375,19 +383,30 @@ int xma_xclbin_info_get(const char *buffer, XmaXclbinInfo *info)
...
 {
     //64 bits based on MAX_DDR_MAP = 64
     int ddr_bank_idx = 0;
-    while (bit_map != 0)
-    {
-        if (bit_map & 1)
-        {
-            *ddr_bank = ddr_bank_idx;
-            return XMA_SUCCESS;
+    if (!has_mem_grps) {
```
keryell, line 390: Often XRT lacks some high-level comments...
Not clear if this is correct since there is no explanation to compare with.
It looks like your are parsing a bitset here to find the number of trailing 0...
You need some C++20 here https://en.cppreference.com/w/cpp/numeric/countr_zero

src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -375,19 +383,30 @@ int xma_xclbin_info_get(const char *buffer, XmaXclbinInfo *info)
...
+        while (bit_map != 0) {
+            if (bit_map & 1) {
+                *ddr_bank = ddr_bank_idx;
+                return XMA_SUCCESS;
+            }
+            ddr_bank_idx++;
+            bit_map = bit_map >> 1;
+        }
+    } else {//For Memory Groups use last group as default memory group; For HBM groups
```
keryell, line 399: Oh, a comment!!!
The readers will be so happy!

src/xma/src/xmaapi/xmaxclbin.cpp
keryell: But why your wisdom is not shown in some comments???

keryell (changes requested):
Give the world some comments :-)

src/xma/src/xmaapi/xmaxclbin.cpp
sarab96: Agree. Will do

src/xma/src/xmaapi/xmaxclbin.cpp
sarab96: Correct. Leading and trailing zeros are checked in two mem_topologies. C++20 is not yet supported by XRT. Good suggestion.

src/xma/src/xmaapi/xmaxclbin.cpp
sarab96: Understood. Will be better. This is old code and doing bug fix in this. Refactoring into proper cpp code is in progress. Will add you in those for review.

src/xma/src/xmaapi/xmaxclbin.cpp
sarab96: Prefer to keep std::string. Want to use const char* internally and so std::string. Also std::string has better protection against invalid string in input char*

src/xma/src/xmaapi/xmaxclbin.cpp
sarab96: Updated

src/xma/src/xmaapi/xmaxclbin.cpp
sarab96: Updated

src/xma/src/xmaapi/xmaxclbin.cpp
keryell: I do not want you to change the type of `temp_mem_topology.m_tag` which remains I guess a `std::string`.
My suggestion is to use first a C++ cast instead of old-school C cast and to remove useless code with a `std::string` constructor here, since there is an implicit constructor from `char *` to `std::string`.
Think Python, think zen. :-)

stsoe: A static analysis tool had better complain about the blind construction from something that is implicitly assumed to null terminated.  So even though m_tag is documented to be null terminated, I think you ought to construct the std::string copy as  `temp_mem_topology.m_tag = {...m_tag, strnlen(...m_tag, 16)}; // magic 16 from xclbin.h`

sarab96: Modified

sarab96: Didn't use strnlen_s as avoiding use of LIB_EXT1 stuff

PR #4295, opened 2020-10-22
disable interrupt if enable_flat is true

Himanshu-xilinx:
enable_flat flag is added for embedded platforms where it load full bitstream after boot. This feature does not support interrupt mode as interrupt controller exist in pl and is configured at boot time. So if enable_flat is true, polling mode should be enabled by default.

### DIFF
--- a/src/runtime_src/core/common/config_reader.h
+++ b/src/runtime_src/core/common/config_reader.h
@@ -355,6 +355,13 @@ get_xma_cpu_mode()
   return value;
 }
 
+inline bool
+get_enable_flat()
+{
+  static unsigned int value = detail::get_bool_value("Runtime.enable_flat",false);
+  return value;
+}
+
 /**
  * Enable / Disable kernel driver scheduling when running in hardware.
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx (approved):

src/runtime_src/core/common/config_reader.h
```diff
@@ -355,6 +355,13 @@ get_xma_cpu_mode()
   return value;
 }
 
+inline bool
+get_enable_flat()
+{
+  static unsigned int value = detail::get_bool_value("Runtime.enable_flat",false);
```
stsoe, line 361: let's do `static bool`  not that it changes behavior

stsoe (changes requested):
Looks good, just change the one line I pointed at.

src/runtime_src/core/common/config_reader.h
Himanshu-xilinx: fixed

chvamshi-xilinx:
@stsoe , merging this change as requested change is made.

PR #4303, opened 2020-10-23
CR-1081090 Memory leak happens in hardware for AIE GMIO design.

larry9523:
This PR fixed a GMIO memory leak issue.

The reason of the leak is currently, in non block GMIO case, we rely on xrtGMIOWait to be called to detach DMA Buf from aie driver. But xrtGMIOWait may not always be called. So a reference count of the buffer could be hold by aie driver.

The fix is: once we have done the BD setting, we can detach the DMA Buf so that the detachment will be guaranteed to be called.

### DIFF
--- a/src/runtime_src/core/edge/user/aie/aie.cpp
+++ b/src/runtime_src/core/edge/user/aie/aie.cpp
@@ -258,7 +258,7 @@ submit_sync_bo(xrtBufferHandle bo, std::vector<gmio_type>::iterator& gmio, enum 
   /*
    * We are safe to release the exported BO DMA Buf here.
    * 1. The physical address of BO is set to AIE register already;
-   * 2. The reference count of the buffer is hold by BO itself.
+   * 2. The reference count of the buffer is held by BO itself.
    */
   clear_bd(bd);
 }
--- a/src/runtime_src/core/edge/user/aie/aie.cpp
+++ b/src/runtime_src/core/edge/user/aie/aie.cpp
@@ -226,7 +226,6 @@ submit_sync_bo(xrtBufferHandle bo, std::vector<gmio_type>::iterator& gmio, enum 
     /* Pending BD is completed by order per Shim DMA spec. */
     for (int i = 0; i < num_comp; ++i) {
       BD bd = dmap->dma_chan[chan].pend_bds.front();
-      clear_bd(bd);
       dmap->dma_chan[chan].pend_bds.pop();
       dmap->dma_chan[chan].idle_bds.push(bd);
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/edge/user/aie/aie.cpp
```diff
@@ -254,6 +254,13 @@ submit_sync_bo(xrtBufferHandle bo, std::vector<gmio_type>::iterator& gmio, enum
   /* Enqueue BD */
   XAie_DmaChannelPushBdToQueue(devInst, shim_tile, pchan, gmdir, bd.bd_num);
   dmap->dma_chan[chan].pend_bds.push(bd);
+
+  /*
+   * We are safe to release the exported BO DMA Buf here.
+   * 1. The physical address of BO is set to AIE register already;
+   * 2. The reference count of the buffer is hold by BO itself.
+   */
+  clear_bd(bd);
```
stsoe, line 263: I am always concerned when I see the pattern:

```
doSomething();
cleanup();
```

What if doSomething() throws an exception?

larry9523: Good point. This PR is actually improved as the aspect of this concern. We have a much shorter path of doSomthing(). :)

So @stsoe , what is your suggestion to address this? In this particular case, prepare_bd and clear_bd can not be part of BD's constructor and destructor because bd has a longer life cycle than the DMA buffer. So, should I create a class to wrap the dma buf so that clear_bd can be called even something goes wrong in between of prepare_bd <--> clear_bd?

stsoe: Yeah, not quite sure I follow how these BDs are used, pushed, and popped on/off idle and pend queues, but you'd have to scope the BD object create in line 234 such that line 263 can be removed.   So as you suggest maybe `struct bd_scoped {...}` is one way of doing this.

larry9523: @stsoe Added a BD_scope struct as suggested. If error happens in the BD_scope destructor, clear_bd will throw exception. The host app will look like
"
[ 1951.221991] xilinx-ai-engine aiepart_0_50: failed to get dmabuf 0.
terminate called after throwing an instance of 'xrt_core::error'
  what():  Sync AIE Bo: fail to detach DMA buf.
[ 1951.282719] [drm] Pid 834 closed device
Aborted
"
I think this is accecptible since something bad happened in the aie kernel driver. We should not move on.

stsoe (approved):

PR #4306, opened 2020-10-23
xbutil host fix + cleanup

Himanshu-xilinx:


chvamshi-xilinx (approved):

src/runtime_src/core/common/ishim.h
```diff
@@ -72,6 +72,9 @@ struct ishim
   virtual void
   get_bo_properties(xclBufferHandle boh, struct xclBOProperties *properties) const = 0;
 
+  virtual void
+  get_device_info(struct xclDeviceInfo2 *info) const = 0;
+
```
stsoe, line 77: I don't think this is needed for the changes you've made

src/runtime_src/core/common/ishim.h
```diff
@@ -271,6 +274,13 @@ struct shim : public DeviceType
       throw error(ret, "failed to get BO properties");
   }
 
+  virtual void
+  get_device_info(struct xclDeviceInfo2 *info) const
+  {
+    if (auto ret = xclGetDeviceInfo2(DeviceType::get_device_handle(), info))
+      throw error(ret, "failed to get device info");
+  }
+
```
stsoe, line 283: Not needed, device_linux is part of libxrt_core and can call xclGetDeviceInfo2 directly.

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -88,6 +88,8 @@ struct devInfo
   get(const xrt_core::device* device,key_type key)
   {
     auto edev = get_edgedev(device);
+    device->get_device_info(&deviceInfo);
```
stsoe, line 91: This can be done by calling xclGetDeviceInfo2 directly:
`xclGetDeviceInfo2(device->get_device_handle(), &deviceInfo);`

The ishim interface is only for user space code that doesn't directly link with libxrt_core.  But here we are in the core library itself so we can call anything from libxrt_core directly.

stsoe (changes requested):

src/runtime_src/core/common/ishim.h
Himanshu-xilinx: true this is not needed for that changes i made but since we have exposed all other shim api's we exposed this also. this can be used at other places also in future. 
@stsoe,  any specific reason of not adding this in ishim?

src/runtime_src/core/common/ishim.h
Himanshu-xilinx: removed

src/runtime_src/core/edge/user/device_linux.cpp
Himanshu-xilinx: did the changes as suggested

src/runtime_src/core/common/ishim.h
stsoe: Yes, I would like to see when this is needed and why.  I think device info should come from individual query requests.  For example windows does not truly populate this struct but does provide query requests.

stsoe (approved):

PR #4308, opened 2020-10-23
Fixed a issue for backward compatibility  changes

saifuddin-xilinx:
-- This is missed for backward compatibility fix
-- As group connectivity section contains both group and bank index, we need to traverse from higher-order to give priority group over bank index
-- We hit this issue CR-1077272 "Multi kernel with port sharing hung in hw_emu"

### DIFF
--- a/src/runtime_src/xocl/core/device.cpp
+++ b/src/runtime_src/xocl/core/device.cpp
@@ -730,6 +730,13 @@ get_cu_memidx() const
       for (auto& cu : get_cu_range())
         mask &= cu->get_memidx_intersect();
 
+      // If all bits are set i.e. No Banks are specified. Hence return the
+      // default bank i.e. Bank 0.
+      if (mask.all()) {
+        m_cu_memidx = 0;
+	return m_cu_memidx;
+      }
+
       // Select first common Group index if present prior to bank index.
       // Traverse from the higher order of the mask as groups comes in higher order
       for (int idx=mask.size() - 1; idx >= 0; --idx) {
--- a/src/runtime_src/xocl/core/device.cpp
+++ b/src/runtime_src/xocl/core/device.cpp
@@ -734,7 +734,7 @@ get_cu_memidx() const
       // default bank i.e. Bank 0.
       if (mask.all()) {
         m_cu_memidx = 0;
-	return m_cu_memidx;
+        return m_cu_memidx;
       }
 
       // Select first common Group index if present prior to bank index.
--- a/src/runtime_src/xocl/core/device.cpp
+++ b/src/runtime_src/xocl/core/device.cpp
@@ -730,13 +730,6 @@ get_cu_memidx() const
       for (auto& cu : get_cu_range())
         mask &= cu->get_memidx_intersect();
 
-      // If all bits are set i.e. No Banks are specified. Hence return the
-      // default bank i.e. Bank 0.
-      if (mask.all()) {
-        m_cu_memidx = 0;
-        return m_cu_memidx;
-      }
-
       // Select first common Group index if present prior to bank index.
       // Traverse from the higher order of the mask as groups comes in higher order
       for (int idx=mask.size() - 1; idx >= 0; --idx) {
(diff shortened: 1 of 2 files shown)


src/runtime_src/xocl/core/memory.cpp
```diff
@@ -333,7 +333,9 @@ get_memidx_nolock(const device* dev, memory::memidx_type subidx) const
   if (mset.none())
     throw std::runtime_error("No matching memory index");
 
-  for (size_t idx=0; idx<mset.size(); ++idx) {
+  // As connectivity section contains both group and bank index. Traverse from
+  // the higher order to give priority on group index over bank index
+  for (int idx=mset.size() - 1; idx >= 0; --idx) {
```
stsoe, line 338: Argh.  You should check if there are more places like this.  I see one in update_memidx_nolock(...) though I am not sure where and how it is used.

stsoe (commented):
Good catch.  See comment.

src/runtime_src/xocl/core/memory.cpp
saifuddin-xilinx: Hi @stsoe ,
The other place which has a similar code i.e. default_allocation_message()
  ```
auto mask = device->get_boh_memidx(boh);
  xocl::device::memidx_type memidx = 0;
  for (size_t idx=0; idx<mask.size(); ++idx) {
    if (mask.test(idx)) {
      memidx = idx;
      break;
    }
  }
```
But I believe, this need not required any changes. For default memory bank we should give priority to the Banks.
Please share your comments.

File: src/runtime_src/xocl/core/device.cpp [hunk 2]
```diff
@@ -730,6 +730,13 @@ get_cu_memidx() const
       for (auto& cu : get_cu_range())
         mask &= cu->get_memidx_intersect();
 
+      // If all bits are set i.e. No Banks are specified. Hence return the
+      // default bank i.e. Bank 0.
+      if (mask.all()) {
+        m_cu_memidx = 0;
+	return m_cu_memidx;
+      }
+
```
saifuddin-xilinx, line 739: These changes are required because when there are no Banks are specified we need to select the default bank. 
Earlier, When No Banks specified, all bits are set in the "mask" and the "**for**" loop is returning the Bank0. But now as we are traversing from the higher-order we need to hardcode to Bank 0 explicitly.

src/runtime_src/xocl/core/memory.cpp
saifuddin-xilinx: Hi @stsoe ,
I have updated the same for other places. Apart from that, there is another place which has a similar code i.e. default_allocation_message()

auto mask = device->get_boh_memidx(boh);
xocl::device::memidx_type memidx = 0;
for (size_t idx=0; idx<mask.size(); ++idx) {
  if (mask.test(idx)) {
    memidx = idx;
    break;
  }
}
But I believe, this need not required any changes. For default memory bank we should give priority to the Banks.
Please share your comments.

dayeh-xilinx:
retest this please

src/runtime_src/xocl/core/device.cpp
stsoe: I am not understanding what triggered this change?

chvamshi-xilinx: Hi @stsoe , Its again due to backward compatibility.  XRT was allocating memory in the first bank If there are no banks used. Ideally this should have not been supported earlier as well, we should have thrown a error. 
To support backward compatibility, we are doing same here. If no banks are specified, allocate in first bank.
We shouldn't break the backward compatibility at this stage. We can remove this check in early cycle of 2021.1 and ask all the app developers to remove/update their test.

@saifuddin-xilinx ,
Please add a clear comment explaining above.

keryell:
I have the feeling that this repeated code should go in a function.

src/runtime_src/xocl/core/device.cpp
chvamshi-xilinx: Hi @saifuddin-xilinx , As discussed , lets remove this change as there are couple of CRs depends on this change.
As we already having this discussion with @stsoe and @maxzhen , lets get a conclusion there about this change.
We can send a separate PR if we need this change.

saifuddin-xilinx: Remove already

stsoe (approved):

chvamshi-xilinx (approved):

PR #4312, opened 2020-10-23
XRT_BBP 1016679: Dont trust size given by userpf

chienwei-lan:


maxzhen (changes requested):
Per our offline discussion, pls make sure:

1. when allocating response buffer, we shall use remote_entry_size * min(local_num_entries, remote_num_entries), and check the final total buffer size.
2. when filling up each entry, we should use min(local_entry_size, remote_entry_size)
3. when moving to next entry, we should use remote_entry_size as step size.

maxzhen (approved):

PR #4321, opened 2020-10-26
Addressing XDP Coverity issues 212793 and 211146

jvillarre:
Fixes issue with uninitialized variable in device offload plugin base class and fixes issue with possible loss of data when storing result of uint32_t multiplication into a uint64_t.


dayeh-xilinx:
retest this please

IshitaGhosh (approved):
Changes look good

PR #4324, opened 2020-10-27
refactor-1

sarab96:
encoder session class to handle encoder video ffmpeg plugins
to use with raw & encoded frames to pass to plugin functions


stsoe (approved):
I have no clue what I am looking at?

PR #4327, opened 2020-10-27
Adds missing ret value for out0

syed-ahmed:
Signed-off-by: Syed Ahmed <stahmed@seas.upenn.edu>

I got "User buffer is not physical contiguous" when using cl_mem_alloc_host_ptr or cl_mem_use_host_ptr with cl_read_only/cl_write_only. While reading this issue: https://github.com/Xilinx/XRT/pull/2164, I learned that I could handle the warning from the user side? However reading the code, it looks like the set `ret` value is not returned in the `out0` label.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
@@ -460,6 +460,7 @@ zocl_userptr_bo_ioctl(struct drm_device *dev, void *data, struct drm_file *filp)
 
 out0:
 	kvfree(pages);
+	return ret;
 out1:
 	zocl_free_userptr_bo(&bo->cma_base.base);
 	DRM_DEBUG("handle creation failed\n");


sonals:
ok to test

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -460,6 +460,7 @@ zocl_userptr_bo_ioctl(struct drm_device *dev, void *data, struct drm_file *filp)
 
 out0:
 	kvfree(pages);
+	return ret;
```
larry9523, line 463: We can not return here without tear down the userptr_bo.

larry9523 (changes requested):
@syed-ahmed, thanks for the patch. But I am not quite sure why do we need to return in out0 lable? The out0 label will fall through to out1 so that ret will be returned.

syed-ahmed:
> @syed-ahmed, thanks for the patch. But I am not quite sure why do we need to return in out0 lable? The out0 label will fall through to out1 so that ret will be returned.

Gotcha. That makes sense. I don't have DRM_DEBUG enabled and overlooked the fall through to out1. Thanks for pointing that!

PR #4328, opened 2020-10-27
refactor-2

sarab96:
scaler session for plugins
For scaler session handling from ffmpeg plugins


stsoe (approved):
These code changes lack comments.  Pointers and references are passed around  without any clue of ownership.  I have a feeling some arguments are missing const, for example should send_frame take a const XmaFrame*?   What is the size of frame_list, etc.

sarab96:
Made changes for const arguments. Size of frame_list is fixed and decided by num_ouput parameter in scaler properties member. This part is same as in current XMA implementation.

PR #4332, opened 2020-10-28
xbutil inside docker mistakenly sees unassigned devices

maxzhen:


xuhz (commented):
Looks like (even the existing)code is using /dev/xfpga to determine whether in container or not.
This is working in some cases because by default the /dev/xfpga is not available within container.
However, this is not correct, especially with the workaround we introduced to make the profiling subdevice accessible with container. -- in that case, the /dev/xfgpa is mapped from host to container, and get_subdev_path() will return true.

Using /dev/dri/renderxxx and /dev/xclmgmtxxx is the correct way

maxzhen:
> 
> 
> Looks like (even the existing)code is using /dev/xfpga to determine whether in container or not.
> This is working in some cases because by default the /dev/xfpga is not available within container.
> However, this is not correct, especially with the workaround we introduced to make the profiling subdevice accessible with container. -- in that case, the /dev/xfgpa is mapped from host to container, and get_subdev_path() will return true.
> 
> Using /dev/dri/renderxxx and /dev/xclmgmtxxx is the correct way

The test for /dev/xfpga code is for xoclv2 driver, which probably does not work. I'll need to work out a proper fix. I tested the case where /dev/xfpga is also mapped to the docker container and existing driver works fine. I'll merge it now since this will fix this regression. I'll work on the solution for the new driver in next release.

PR #4335, opened 2020-10-28
CR-1081796: added signal to terminate aied thread

Himanshu-xilinx:


src/runtime_src/core/edge/user/aie/aied.cpp
```diff
@@ -32,28 +32,35 @@ namespace zynqaie {
 Aied::Aied(xrt_core::device* device): mCoreDevice(device)
 {
   done = false;
-  mPollingThread = std::thread(&Aied::pollAIE, this);
+  pthread_create(&ptid, NULL, &Aied::pollAIE, this);
```
stsoe, line 35: Why can't the thread be stopped in the destructor?  Bare pthread should be avoided.
Killing threads is usually a bad idea, which is why std::thread doesn't support it.
This PR lacks comments explaining the reasoning.

stsoe (changes requested):

src/runtime_src/core/edge/user/aie/aied.cpp
Himanshu-xilinx: This thread is sleeping in zocl. it can be awaked  only in 3 condition.
1. waiting condition is met. 
2. if timeout happen. it happen every 500ms but its huge time. because of which clReleasecontext is taking huge time.
3. when signal is received and it wake the thread. this is the reason we shifted to pthread as we found no way to pass signal from parent to child.

larry9523: Some background:
1. This thread is severed as a daemon. Most of the time, it sends an ioctl to zocl driver which is blocked at wait_event_interruptible_timeout(..., 500ms) in kernel space. In worst case, it takes 500ms for this thread to wake up and goes back to userspace.
2. If no signal is sent to the thread in the ~aied(), where mPollingThread.join() is called, it takes 500ms (worst case) here. That is why QoR observed that clRelaseContext takes 300ms (average) to execute
3. So if we can send a signal to this thread, it will wake up immediately from wait_event_interruptible. (I don't like wait_event_interruptible_timeout, but it is a different story)
4. I did some research that std::thread has no way to send signal to a single thread. Correct me if I am wrong. So a suggestion I gave is to use pthread so that we can send a signal here.

larry9523 (approved):
The only comment I have is wait_event_interruptible_timeout can be changed to wait_event_interruptible(). But this can be revisited in next release.

src/runtime_src/core/edge/user/aie/aied.cpp
stsoe: Can the thread run detached?

Himanshu-xilinx: no, because once userspace data is gone it make no sense to keep this.
this complete daemon will go off once in 2021.1 aie driver start supporting sysfs nodes for aie inforamtion.

stsoe: You can still make the detached thread stop by setting done.
Other user space processes will see the thread alive for maybe 500ms extra, but that hardly matters.

The detached thread will die once the process that started it exists so the xbutil case should be an issue since after process exit there will be no further calls into the driver.

The only problem with detaching is to make sure that the driver handles restarting the of main program (and thereby the 'daemon') while the previous wait_event_... has not yet returned from timeout.

stsoe: @Himanshu-xilinx .  Chatting with Larry, I do think it is safer to signal the thread and join properly as you are doing.  Let's keep as is.

stsoe (approved):

chvamshi-xilinx (approved):

PR #2466, opened 2019-11-20
2019.2

jnicholsXilinx:
Remove compatibility check of DSA in YAML configuration against DSA loaded on accelerator. This check is superfluous since the xclbin will not be loaded if there is a mismatch. More importantly, this change allows multiple cards types to be supported on the same server.

PR #4337, opened 2020-10-28
Cherry pick check to ignore DSA check on XMA Legacy from 2019.2 branch (#2466)

rathaval:
This was merged into XRT/2019.2 but not into XRT/master. See commit 5843d44 for the original change.
However, with U30 Raptor bring up, this change helps quickly verify U30 use-cases on XMA legacy.

Requesting to be merged into XRT/master as xma_legacy is not yet End-of-Life.

* Remove DSA check and make DSA in config optional

* Remove #ifdef of DSA compatibility check


sarab96 (approved):
Approved as required for Raptor checks. For general use XMA_legacy is discouraged and no new changes are allowed.

PR #4346, opened 2020-10-29
Check kernel configuration for verify_pkcs7_signature() function

mclayton1000:
Check that the option SYSTEM_DATA_VERIFICATION is enabled in current kernel configuration.
That configuration option is mandatory to invoke verify_pkcs7_signature() function.


maxzhen:
ok to test

PR #4350, opened 2020-10-29
XRT-787 SC flash support on u30

xuhz:
1. xbmgmt2 is not supported at this time. will file a CR tracking it
2. sc flash for u30 will not do a hot(card) reset. It is doing a ps only reset on the FPGA with fixed sc so far. This is just to workaround https://jira.xilinx.com/browse/CR-1076187
this makes the sc flash takes longer. Once that CR is fixed by cmc, we can remove the ert reset to save time
3. xmc driver set mailbox offset during probe(). For some reason even Alan doesn't know, sometimes, at that time the register doesn't return the value(it returns 0). So we set a default value once in that case. When I use xrt userlevel xmc, since the register is read when the xbmgmt is running, the register never returns 0

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
@@ -318,6 +318,7 @@ struct xmc_pkt_hdr {
 	u32 op		: 8;
 };
 
+#define XMC_MBX_DEF_OFFSET 0x1000
 /* We have a 4k buffer for xmc mailbox */
 #define	XMC_PKT_MAX_SZ	1024 /* In u32 */
 #define	XMC_PKT_MAX_PAYLOAD_SZ	\
@@ -3201,7 +3202,7 @@ static void xmc_enable_mailbox(struct xocl_xmc *xmc)
 
 	xmc->mbx_enabled = true;
 	safe_read32(xmc, XMC_HOST_MSG_OFFSET_REG, &val);
-	xmc->mbx_offset = val;
+	xmc->mbx_offset = val ? val : XMC_MBX_DEF_OFFSET;
 	xocl_info(&xmc->pdev->dev, "XMC mailbox offset: 0x%x", val);
 }
 
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
@@ -318,7 +318,6 @@ struct xmc_pkt_hdr {
 	u32 op		: 8;
 };
 
-#define XMC_MBX_DEF_OFFSET 0x1000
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
```diff
@@ -3201,7 +3202,7 @@ static void xmc_enable_mailbox(struct xocl_xmc *xmc)
 
 	xmc->mbx_enabled = true;
 	safe_read32(xmc, XMC_HOST_MSG_OFFSET_REG, &val);
-	xmc->mbx_offset = val;
+	xmc->mbx_offset = val ? val : XMC_MBX_DEF_OFFSET;
```
xdavidz, line 3205: Is this temporary workaround or permanent change? I would think for some reason the data in the mailbox is corrupted or temporary unavailable.

src/runtime_src/core/pcie/tools/xbmgmt/cmd_flash.cpp
```diff
@@ -204,7 +293,30 @@ static int updateSC(unsigned index, const char *file)
...
     }
-    
+    std::string vbnv;
+    mgmt_dev->sysfs_get( "rom", "VBNV", errmsg, vbnv );
+    if (!errmsg.empty()) {
+        std::cerr << errmsg << std::endl;
+        return -EINVAL;
+    }
+    //don't trigger reset for u30. let python helper handle everything
+    if (vbnv.find("_u30_") != std::string::npos) {
```
xdavidz, line 303: Again, please confirm this is permanent&stable change or temporary workaround.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
maxzhen: > 
> 
> Is this temporary workaround or permanent change? I would think for some reason the data in the mailbox is corrupted or temporary unavailable.

had a offline chat with @xuhz. he will remove this hard-code offset and do it differently.

xuhz: it is a common issue for all cases where cmc is running on ps. so far there is no dependency for xmc driver loading on host and cmc running on ps. So when xmc is probed, the cmc may not be ready yet.
old u30 doesn't have POR, so after reset, the cmc is up and running pretty fast. On new u30, it takes ~1min for the ps up, before that xmc on host may have been probed and a 0 is saved as mailbox offset.
Don't know versal has same case or not

xdavidz: Versal boot time is around 10s as of today. I think return EIO and user retry the flash is safer.

maxzhen (approved):

xdavidz (approved):

AShivangi (approved):

PR #4356, opened 2020-10-29
CR 1081612 : fix crash when aie_profile is enabled for design without any aie counters 

IshitaGhosh:


jvillarre (approved):
Changes look good.

PR #4359, opened 2020-10-30
CR-1081983: Enables LOP trace and other profiling plugins to work without AIE

jvillarre:
This pull request breaks the inherent dependency on Edge between the xdp_core library and the xrt_core library that existed when building for AIE offload.  It put the AIE specific portions into the plugins and makes xdp_core only dependent on xrt_coreutil again.


IshitaGhosh (approved):
Changes look good and AIE trace, profile works with it.

maxzhen:
retest this please

PR #4364, opened 2020-10-30
Replace boost/detail/endian.hpp with

ckalluri73:
boost/predef/other/endian.h

The use of BOOST_*_ENDIAN and BOOST_BYTE_ORDER is deprecated. Include
<boost/predef/other/endian.h> and use __BYTE_ORDER and1234

Signed-off-by: Sai Hari Chandana Kalluri <chandana.kalluri@xilinx.com>

### DIFF
--- a/src/runtime_src/xocl/core/kernel.cpp
+++ b/src/runtime_src/xocl/core/kernel.cpp
@@ -21,7 +21,7 @@
 #include "compute_unit.h"
 #include "core/common/xclbin_parser.h"
 
-#include <boost/detail/endian.hpp>
+#include <boost/predef/other/endian.h>
 
 #include <sstream>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx:
ok to test

ckalluri73:
A little background for the patch:
I am currently working on upgrading Yocto version to gatesgarth which uses boost 1.74.0 version.
The latest version of boost has deprecated some macros resulting in build failures for xrt in Yocto.
The above patch fixes the compilation errors.

The build pipeline is using petalinux to test and this is based on current Yocto Zeus version which is using boost 1.70.0. 
This patch is not for 2020.2 release but 2021.1

stsoe (commented):
Thanks for the suggestion.  Unfortunately the changes is not compatible with older versions of Boost.  I will close this PR and remove the check as was recently done else where.

keryell:
When I say hardware people should use modern C++... :-) https://en.cppreference.com/w/cpp/types/endian

stsoe:
Wish life was that simple.

PR #4370, opened 2020-10-31
refactor-1

sarab96:
refactor for xma

### DIFF
--- a/src/xma/include/lib/xma_singleton.hpp
+++ b/src/xma/include/lib/xma_singleton.hpp
@@ -1,0 +1,64 @@
+/**
+ * Copyright (C) 2020 Xilinx, Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may
+ * not use this file except in compliance with the License. A copy of the
+ * License is located at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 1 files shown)


stsoe (commented):
There is typically no need for a singleton class exposed through as a public class.

In this case it looks like the singleton is added to provide access to single instance of num_of_blah_blah().  Consider moving all of this to a .cpp file, where the accessors are provided through functions within a namespace.  In the cpp file, you can have the atomics, mutexes, and bools as static globals with the namespace.

client code would look like:

`auto decoders = xma_core::app::num_decoders();`

where the implementation of num_decoders() in the cpp file would simply provide safe access to the static global atomic declared and defined in same .cpp file.

sarab96:
I'm avoiding multiple static global variables as those can run into initialization order problems if not handled properly. So using cpp class is better way to go. Singleton is not used for any performance critical areas or to manage execution flow of host app

stsoe:
I think you should move the singleton class to the .cpp file and provide accessors, which under the hood calls instance() like you already do.  There is no point in having client code know that there is a singleton class involved here, it complicates the upstream code unnecessarily.    But maybe you intend for client code to directly set the variables managed by the singleton?   I am not sure, because the usage of the singleton is not commented, I don't know the intent.

sarab96:
This is library internal class and is not exposed to user space. Getter/Setter layer is not required here. These are static globals for XMA2 library internal code. Nothing fancy in this part of library. So keeping it as singleton class is a good safe way. And keeping it in .hpp avoids extra layer of setter/getter which is not required in this part.

stsoe:
retest this please

dayeh-xilinx:
retest this please

PR #4371, opened 2020-11-01
CR-1082322: kernel argument fix in cpp api

sarab96:
kernel argument check added
This fix addresses below problem.

If extra argument was passed to kernel:
    - No error was given
    - Kernel output was incorrect

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -1053,10 +1053,8 @@ public:
     // get kernel arguments from xml parser
     // compute regmap size, convert to typed argument
     for (auto& arg : xrt_core::xclbin::get_kernel_arguments(xml_section.first, xml_section.second, name)) {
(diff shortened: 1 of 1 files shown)


stsoe:
I am not comfortable with this fix without a corresponding CR.  While the fix may be exactly what is needed, please associate it with a CR.

sarab96:
Added CR #: CR-1082322. With this fix extra kernel argument throws out of range exception. Like argument 1 is out of range from range size of one.

stsoe:
Thanks @sarab96.  I will need the non-indexed args stored in the args array for supporting OpenCL through the native APIs.
Lets change this PR to simply throw out-of-range in kernel::get_arg(size_t argidx) when argidx is different from the arguments index.  Either that or ignore the trailing arguments that are out-of-range (as the c-style va_list does).

sarab96:
Updated.  It will throws runtime error of bad argument index

stsoe (approved):

stsoe:
retest this please

PR #4373, opened 2020-11-02
zocl git hash version

sarab96:
Added git hash, hash date, branch, etc to zocl driver sysfs
In edge systems it is common to replace zocl.ko directly
Full XRT edge build and boot.bin is not always used
Flash is not update always and instead zocl.ko is directly replaced

So adding version info to zocl sysfs as xbutil version is not sufficient for edge platforms

### DIFF
--- a/src/runtime_src/core/edge/drm/Makefile
+++ b/src/runtime_src/core/edge/drm/Makefile
@@ -6,6 +6,12 @@ KBUILD_OPTS := M=$(CURDIR)/$(DRIVER)
 KBUILD_OPTS += ARCH=$(TARGET)
 KBUILD_OPTS += CROSS_COMPILE=$(CROSS_COMPILE)
 
+GIT_HASH := `git -C /proj/rdi-xsj/staff/sarabjee/xrt_git/sarab/XRT/src rev-parse --verify HEAD`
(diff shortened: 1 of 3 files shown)


mamin506 (approved):

maxzhen (approved):

sarab96:
retest this please

src/runtime_src/core/edge/drm/Makefile
```diff
@@ -6,6 +6,12 @@ KBUILD_OPTS := M=$(CURDIR)/$(DRIVER)
 KBUILD_OPTS += ARCH=$(TARGET)
 KBUILD_OPTS += CROSS_COMPILE=$(CROSS_COMPILE)
 
+GIT_HASH := `git -C /proj/rdi-xsj/staff/sarabjee/xrt_git/sarab/XRT/src rev-parse --verify HEAD`
```
chvamshi-xilinx, line 9: @sarab96 , Can you please explain why hardcoding the path here? 
And xrt-recipe in petalinux doesnt use this make file. It uses the makefile which is inside zocl.
https://gitenterprise.xilinx.com/Yocto/meta-xilinx/blob/2020/meta-xilinx-bsp/recipes-xrt/zocl/zocl_git.bb

PR #4374, opened 2020-11-02
zocl driver version

sarab96:
zocl version update for 2020
This version is used by lspci, etc

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_drv.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_drv.c
@@ -37,8 +37,8 @@
 
 #define ZOCL_DRIVER_NAME        "zocl"
 #define ZOCL_DRIVER_DESC        "Zynq BO manager"
-#define ZOCL_DRIVER_DATE        "20180313"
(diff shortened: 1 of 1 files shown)


maxzhen (changes requested):
do we expect any change to zocl to update driver date? when do we want to update driver major?

sarab96:
zocl had large changes for soft kernel support and I think date was not bumped up for that. Now for U30 there are some changes.
So I think driver date can be changed now. Release, major, minor can be changed now to match current XRT release as it was not changed in earlier releases.

maxzhen:
driver's major is 2020, minor is 2 and patch leve is 1, how does it match XRT's release version, which is something like 2.9.0 here? Can we make it updated automatically as xocl?

larry9523:
I think we have agreement that we should report zocl vsersion as "<xrt version>:<commit number>"

larry9523 (commented):
see comment

sarab96:
Closing this one. New changes made in new PR: https://github.com/Xilinx/XRT/pull/4398

PR #4377, opened 2020-11-02
Load legacy scheduler bin (sched.bin) when the firmware branch name is legacy

rajkumar-xilinx:

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/xocl_fdt.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/xocl_fdt.c
@@ -1806,8 +1806,14 @@ const char *xocl_fdt_get_ert_fw_ver(xdev_handle_t xdev_hdl, void *blob)
 			break;
 		}
 	}
-	if (fw_ver)
+	if (fw_ver) {
 		xocl_xdev_info(xdev_hdl, "Load embedded scheduler firmware %s", fw_ver);
+		/* if firmware_branch_name is "legacy", XRT loads the sched.bin */
+		if (!strcmp(fw_ver, "legacy")) {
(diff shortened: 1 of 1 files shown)


maxzhen (changes requested):
Where does this "legacy" come from and why it can't be "sched.bin"?

chvamshi-xilinx:
> Where does this "legacy" come from and why it can't be "sched.bin"?

Hi @maxzhen , There is no way to make firmware_branch_version to empty string in DTB. When this value is "empty string" in partition.dts, it is becoming some junk characters in the dtb.   Thatswhy Platform team is decided to add 
"legacy" for the firmware_branch_version in the partition metadata so that XRT can load the sched.bin.

We never hit this case as XRT tries to load scheduler which is in XRT installation area first.  If it doesn't find proper scheduler in XRT installation area,  it loads a scheduler which is part of xsabin. As this xsabin is generated for 0RP shell as there is no scheduler inside xsabin.

maxzhen (approved):

maxzhen:
> 
> 
> > Where does this "legacy" come from and why it can't be "sched.bin"?
> 
> Hi @maxzhen , There is no way to make firmware_branch_version to empty string in DTB. When this value is "empty string" in partition.dts, it is becoming some junk characters in the dtb. Thatswhy Platform team is decided to add
> "legacy" for the firmware_branch_version in the partition metadata so that XRT can load the sched.bin.
> 
> We never hit this case as XRT tries to load scheduler which is in XRT installation area first. If it doesn't find proper scheduler in XRT installation area, it loads a scheduler which is part of xsabin. As this xsabin is generated for 0RP shell as there is no scheduler inside xsabin.

OK, I see. Thanks for explaining.

chienwei-lan (approved):

chvamshi-xilinx (approved):

PR #4381, opened 2020-11-02
CR-1081642 After fuzzing shell and running validate, platform gets st…

xdavidz:
…uck in errored state

Based on the comments from the CR, we should explicitly re-set the u64 size for kdma cmd.

### DIFF
--- a/src/runtime_src/core/include/ert.h
+++ b/src/runtime_src/core/include/ert.h
@@ -180,7 +180,7 @@ struct ert_start_copybo_cmd {
   uint32_t state:4;          /* [3-0], must be ERT_CMD_STATE_NEW */
   uint32_t unused:6;         /* [9-4] */
   uint32_t extra_cu_masks:2; /* [11-10], = 3 */
-  uint32_t count:12;         /* [22-12], = 16, exclude 'arg' */
+  uint32_t count:11;         /* [22-12], = 16, exclude 'arg' */
   uint32_t opcode:5;         /* [27-23], = ERT_START_COPYBO */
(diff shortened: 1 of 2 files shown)


houlz0507 (approved):
I assume the size will never beyond 32-bit, correct?

src/runtime_src/core/include/ert.h
```diff
@@ -180,7 +180,7 @@ struct ert_start_copybo_cmd {
   uint32_t state:4;          /* [3-0], must be ERT_CMD_STATE_NEW */
   uint32_t unused:6;         /* [9-4] */
   uint32_t extra_cu_masks:2; /* [11-10], = 3 */
-  uint32_t count:11;         /* [22-12], = 15, exclude 'arg' */
+  uint32_t count:12;         /* [22-12], = 16, exclude 'arg' */
```
stsoe, line 183: huh?  4+6+2+12+5+4 = 33  
count has to match ert_packet, I am not following this change?

stsoe (changes requested):

xdavidz:
> I assume the size will never beyond 32-bit, correct?

This is a quick fix that existing code will only pass in 32-bit size. If we change the existing interface to 64bit, there will be too many other code changes. I just want to minimize the code change scope for this time before the 2020.2 release.

In the future, we should consider switching the interface to 64-bit.

src/runtime_src/core/include/ert.h
xdavidz: sorry, reverted back.

stsoe (approved):

xdavidz:
retest this please

mamin506 (approved):

PR #4385, opened 2020-11-02
error message clear when buffer is not resident yet migrating from device to host

uday610:
This CR https://jira.xilinx.com/browse/CR-1067158
This came from customer when they could not understand the error message. 
The error comes only when buffer is not resident but still we are trying to migrate from device->host
Making the error message clear.


stsoe (approved):

stsoe:
ok test test

PR #4393, opened 2020-11-03
CR-1082277 M2M test fails during xbutil validate (if host_mem has bee…

xdavidz:
…n requested)

Per request from the CR, disable m2m test for HOST memory.

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
+++ b/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
@@ -1509,6 +1509,11 @@ int xcldev::device::getXclbinuuid(uuid_t &uuid) {
     return 0;
 }
 
+bool xcldev::device::isHostMem(const char *m_tag)
+{
+    return (!strncmp(m_tag, "HOST", 4));
+}
+
 int xcldev::device::kernelVersionTest(void)
 {
     if (getenv_or_null("INTERNAL_BUILD")) {
@@ -2317,6 +2322,9 @@ int xcldev::device::testM2m()
     }
 
     for(int32_t i = 0; i < map->m_count; i++) {
+        if (isHostMem((const char *)map->m_mem_data[i].m_tag))
+            continue;
+
         if(map->m_mem_data[i].m_used &&
             map->m_mem_data[i].m_size * 1024 >= m2mBoSize) {
             /* use u8 m_type field to as bank index */
--- a/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
+++ b/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
@@ -1511,7 +1511,9 @@ int xcldev::device::getXclbinuuid(uuid_t &uuid) {
 
 bool xcldev::device::isHostMem(const char *m_tag)
 {
-    return (!strncmp(m_tag, "HOST", 4));
+    std::string str(m_tag);
+
+    return (!str.compare(0, 4, "HOST"));
 }
 
 int xcldev::device::kernelVersionTest(void)
--- a/src/runtime_src/core/pcie/tools/xbutil/xbutil.h
+++ b/src/runtime_src/core/pcie/tools/xbutil/xbutil.h
@@ -1727,7 +1727,7 @@ public:
             if(map->m_mem_data[i].m_type == MEM_STREAMING)
                 continue;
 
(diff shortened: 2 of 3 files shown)


src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -2317,6 +2322,9 @@ int xcldev::device::testM2m()
     }
 
     for(int32_t i = 0; i < map->m_count; i++) {
+        if (isHostMem((const char *)map->m_mem_data[i].m_tag))
```
AShivangi, line 2325: why not use std::compare to compare the string instead of converting it to const char* first?

AShivangi (changes requested):
Can you please add these changes for xbutil2 as well. In "src\runtime_src\core\tools\xbutil2\SubCmdValidate.cpp"

L718, add this check for DMA test
L838 for m2m test

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
xdavidz: This is a C interface as "unsigned char". If we use CPP compare, we have to add more code like std::string(m_tag) then compare. 

I am ok to make the changes.

AShivangi: Yes, let's use cpp :)

maxzhen (approved):

chienwei-lan (approved):

NaAl-Xlnx:
retest this please

src/runtime_src/core/pcie/tools/xbutil/xbutil.h
```diff
@@ -1727,7 +1727,7 @@ class device {
             if(map->m_mem_data[i].m_type == MEM_STREAMING)
                 continue;
 
-            if(!strncmp((const char*)map->m_mem_data[i].m_tag, "HOST", 4))
+            if(isHostMem((const char*)map->m_mem_data[i].m_tag))
```
stsoe, line 1730: No C-style casts please.  Not something you changed :-) but maybe you can fix it?

xdavidz:
retest this please

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -835,6 +835,10 @@ m2mTest(const std::shared_ptr<xrt_core::device>& _dev, boost::property_tree::ptr
   auto mem_topo = reinterpret_cast<const mem_topology*>(membuf.data());
 
   for (auto& mem : boost::make_iterator_range(mem_topo->m_mem_data, mem_topo->m_mem_data + mem_topo->m_count)) {
+    std::string str((char *)mem.m_tag);
```
stsoe, line 838: No C-style cast and don't assume null terminated char* when constructing std::string, use either
`auto c_str = reinterpret_cast<const char*>(mem.m_tag);`
`std::string str{c_str, std::min(strlen(c_str), 16)};  // magic 16 from xclbin.h`
or
`std::string str{c_str, std::find(c_str, c_str + 16, 0)};`  // magic 16 from xclbin.h

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
stsoe: Do not convert unnecessarily to std::string.  I believe isHostMem is wasteful (const char* -> string -> const char*) strncmp would be better in this case

stsoe (commented):
These comments are not serious, just recommendations.

xdavidz:
@stsoe @AShivangi so what is your suggestion here, using strncmp as before or compare? This is the last minute fix for the issue, I want to get clear call what coding style we are going to use here. Maybe we could improve her later, but I did see lots of exceptions  in this xbutil.h and xbuti.cpp as well.

stsoe:
Code is correct though not efficient.   Don't make further changes at this point.

stsoe (approved):

xdavidz:
> Code is correct though not efficient. Don't make further changes at this point.

Thanks, I've created a CR for coding style clean up and assigned it to myself. 
CR-1082484 cpp coding style cleaning up for xbuti

xdavidz:
retest this please

xilinxgitops:
retest this please

PR #4395, opened 2020-11-04
regmap size check updated

sarab96:
regmap size check updated.
Change is based on another PR for cpp XRT kernel APIs
No change in functionality

### DIFF
--- a/src/xma/src/xmaapi/xmaxclbin.cpp
+++ b/src/xma/src/xmaapi/xmaxclbin.cpp
@@ -122,6 +122,8 @@ static int get_xclbin_iplayout(const char *buffer, XmaXclbinInfo *xclbin_info)
             using xarg = xrt_core::xclbin::kernel_argument;
             static constexpr size_t no_index = xarg::no_index;
             auto args = xrt_core::xclbin::get_kernel_arguments(xclbin, temp_ip_layout.kernel_name);
+            //Note: args are sorted by index; Assuming that offset will increase with arg index
+            //This is fair assumption; v++ or HLS decides on these offset values
             std::vector<xarg> args2;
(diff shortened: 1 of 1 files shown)


src/xma/src/xmaapi/xmaxclbin.cpp
```diff
@@ -119,12 +119,17 @@ static int get_xclbin_iplayout(const char *buffer, XmaXclbinInfo *xclbin_info)
...
+            std::copy_if(args.begin(), args.end(), std::back_inserter(args2), 
+                  [](const xarg& y){return y.index != no_index;});
             temp_ip_layout.arg_start = -1;
             temp_ip_layout.regmap_size = -1;
-            if (args.size() > 0) {
-                temp_ip_layout.arg_start = args[0].offset;
-                auto& last = args.back();
+            if (args2.size() > 0) {
+                temp_ip_layout.arg_start = args2[0].offset;
+                auto& last = args2.back();
```
stsoe, line 132: Appears to assume that offset is  increasing with argument.  Not sure if that is a safe assumption, but it is definitely an implicit assumption that should be commented.

sarab96: I see sort in get_kernel_arguments function. Line# 606. So this should be fine.

sarab96: On offset increasing by arg index. Yes that is an assumption. But that is fair assumption. It probably can be confirmed by v++ or hls team. Also I have not seen any place where this isn't true.

sarab96: Comment on this included in the code

PR #4397, opened 2020-11-04
Update metadata to fix host compilation issue on sprite

haeseunglee:
No source code change. Just update metadata to resolve host compilation issue on sprite.


stsoe (changes requested):
None of the low level XRT tests should link with xrt_core, please check CMakeLists.txt for required link libraries.

Also, why the ABI flag?

haeseunglee:
I removed the xrt_core library from metadata, except 100_ert_ncu.
The ABI flag is due to the sprite's g++ setting, host compilation fails without the ABI flag on the sprite.

stsoe (approved):

PR #4400, opened 2020-11-04
Memory Grouping : Work around for U280 grouping issue and bug fix for multiple groups

rozumx:
Issue:
  The U280 platform's memory types are incorrectly set (e.g., MEM_DDR4, MEM_DRAM, etc.).  Because the underlying grouping sorting algorithm grouped according to type, the result was two separate groups for each type.  A testcase currently not supported (and discovered by xclbinutil's DRC check).

Resolution:
The grouping algorithm has been updated to:
  1. Ignore the memory type
  2. Detect if there are more then 1 group for a given IP / argument pair.  If detected, no groupings will be performed for that IP / argument pair.

CRs:
  1. CR-1072704 - hbm_simple working for 2020.1_daily_latest but failing for 2020.2_daily_latest
  2. CR-1081877 - [u280]: Invalid memory grouping for L3 HBM designs during hw_emu (2020.2_daily_latest)

### DIFF
--- a/src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx
+++ b/src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx
@@ -602,6 +602,7 @@ typedef struct {
   std::string memType;          // Type of memory being indexed
   uint64_t baseAddress;         // Base address of the memory
   uint64_t size;                // Size of the memory
+  bool canGroup;                // Indicates if this entry can be grouped
 } WorkingConnection;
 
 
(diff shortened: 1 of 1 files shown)


stsoe (approved):

chvamshi-xilinx:
Marking "do not merge". @saifuddin-xilinx  is going to verify couple of grouping example. Will merge the changes once those passes.

saifuddin-xilinx:
Basic tests are passed. Verified both U280 and U50 platform.

PR #4413, opened 2020-11-05
Add warning message for trace buffer full in HAL Device Offload Plugin

IshitaGhosh:


src/runtime_src/core/common/config_reader.h
```diff
@@ -145,8 +145,7 @@ get_container()
 inline std::string
 get_data_transfer_trace()
 {
-  static std::string value = 
-    detail::get_string_value("Debug.data_transfer_trace","off");
+  std::string value = detail::get_string_value("Debug.data_transfer_trace","off");
```
stsoe, line 148: Why was static removed?

stsoe (changes requested):

src/runtime_src/core/common/config_reader.h
IshitaGhosh: I have shared details over email.

pgschuey (approved):
changes look good

stsoe (approved):

PR #4419, opened 2020-11-06
CR-1080843 xbutil validate SEG fault with hal profiling

stsoe:
Change xbutil to dynamic link with xrt_core and xrt_coreutil

The static linking caused problems for hal level profiling, which
run-time loads plugin libraries that are dynamically linked with
xrt_core and xrt_coreutil.  Static global data from xrt_core was
duplicated in both xbutil and the plugin library.


larry9523 (approved):
I did notice this static link before. When I debug some xbutil issue, I replace the xrt_core.so and it never took effect. :)

IshitaGhosh (approved):
Changes look good.

PR #4425, opened 2020-11-09
CR-1082484 cpp coding style cleaning up for xbutil isHostMem

xdavidz:
Make the code clean again.

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
+++ b/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
@@ -1509,11 +1509,9 @@ int xcldev::device::getXclbinuuid(uuid_t &uuid) {
     return 0;
 }
 
-bool xcldev::device::isHostMem(const char *m_tag)
+bool xcldev::device::isHostMem(const unsigned char *m_tag)
 {
-    std::string str(m_tag);
-
-    return (!str.compare(0, 4, "HOST"));
+    return (!strncmp((const char *)m_tag, "HOST", 4));
 }
 
 int xcldev::device::kernelVersionTest(void)
@@ -2324,7 +2322,7 @@ int xcldev::device::testM2m()
     }
 
     for(int32_t i = 0; i < map->m_count; i++) {
-        if (isHostMem((const char *)map->m_mem_data[i].m_tag))
+        if (isHostMem(map->m_mem_data[i].m_tag))
             continue;
 
         if(map->m_mem_data[i].m_used &&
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -1509,11 +1509,9 @@ int xcldev::device::getXclbinuuid(uuid_t &uuid) {
     return 0;
 }
 
-bool xcldev::device::isHostMem(const char *m_tag)
+bool xcldev::device::isHostMem(const unsigned char *m_tag)
 {
-    std::string str(m_tag);
-
-    return (!str.compare(0, 4, "HOST"));
+    return (!strncmp((const char *)m_tag, "HOST", 4));
```
AShivangi, line 1514: please use `reinterpret_cast<const char*>`

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -835,7 +835,8 @@ m2mTest(const std::shared_ptr<xrt_core::device>& _dev, boost::property_tree::ptr
   auto mem_topo = reinterpret_cast<const mem_topology*>(membuf.data());
 
   for (auto& mem : boost::make_iterator_range(mem_topo->m_mem_data, mem_topo->m_mem_data + mem_topo->m_count)) {
-    std::string str((char *)mem.m_tag);
+    auto c_str = reinterpret_cast<const char*>(mem.m_tag);
+    std::string str(c_str, std::min<size_t>(strlen(c_str), 16)); // max len is 16 from xclbin.h
     if (!str.compare(0, 4, "HOST"))
         continue;
```
stsoe, line 841: What is wrong with using strcmp?
`strcmp(c_str,"HOST");`
Avoid unnecessary conversion to std::string.  If the std::string is needed,  for example the tag name is stored some place, then it would be fine, but if not needed it is just a waste.

AShivangi (approved):

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
stsoe: Why is this a class member function?  It is not part the device interface.  It would be better to have this be a namespace scoped free function.
```
namespace xcldev {
inline bool
isHostMem(const char* tag)
{
  return strcmp(tag,"HOST")==0;
}

inline bool
isHostMem(const unsigned char* tag)
{
  return isHostMem(reinterpret_cast<const char*>(tag));
}
}
```

stsoe (changes requested):

stsoe (approved):

PR #4427, opened 2020-11-09
CR-1082801 DDR self-refresh is not working on xilinx_u250_gen3x16_xdma_3_1_202020_1

chienwei-lan:


src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_ioctl.c
```diff
@@ -292,6 +292,37 @@ static bool xclbin_downloaded(struct xocl_dev *xdev, xuid_t *xclbin_id)
...
+	size_t i, j = 0;
+
+	if (mem_topo->m_count != new_topo->m_count)
+		return ret;
+
+	for (i = 0; i < mem_topo->m_count; ++i) {
+		if (IS_HOST_MEM(mem_topo->m_mem_data[i].m_tag))
+			continue;
+		for (j = 0; j < new_topo->m_count; ++j) {
+			if (IS_HOST_MEM(new_topo->m_mem_data[j].m_tag))
```
maxzhen, line 307: shouldn't you compare the m_tag here? we don't want to assume the order of mem entries are the same b/w these two xclbins, right?

maxzhen (changes requested):

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_ioctl.c
maxzhen: talked to Chien-Wei offline. the code here is not efficient, but functionality is OK. approving it now.

maxzhen (approved):

maxzhen (commented):
why do you make this change? do you plan to update the code in master?

chienwei-lan:
> why do you make this change? do you plan to update the code in master?

Yes, I did have a PR for master.

maxzhen:
Approved by Sumit.

PR #4433, opened 2020-11-10
Ini api

uday610:
API for setting ini options for [Runtime] group

### DIFF
--- a/src/runtime_src/doc/toc/xrt_ini.rst
+++ b/src/runtime_src/doc/toc/xrt_ini.rst
@@ -46,9 +46,7 @@ Example
 .. code-block:: ini
 
     xrt::ini::set("Runtime.runtime_log", "console");
-    xrt::ini::set("Runtime.exclusive_cu_context", true);
-
-
+    xrt::ini::set("Runtime.verbosity", 5);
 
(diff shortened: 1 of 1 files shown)


src/runtime_src/doc/toc/xrt_ini.rst
```diff
@@ -30,6 +30,23 @@ The following is a simple example that turns on profile timeline trace and sends
    timeline_trace = true
 
 
+**API Support**: From 2020.2 release the runtime configuration options can also be provided through APIs. These APIs are applicable for ``[Runtime]`` group options, not for ``[Debug]`` or ``[Emulation]`` related options.
```
stsoe, line 33: This is not true, sorry misunderstood what you were asking.   The APIs for changing ini values support any group, it's all about the string used as key for the API.

stsoe (changes requested):

src/runtime_src/doc/toc/xrt_ini.rst
uday610: Ok, removed that line. Thanks

stsoe (approved):

PR #4436, opened 2020-11-11
Fix instantiation of logger objects for HAL APIs like xclOpen, xclProbe, xclOpenContext

IshitaGhosh:


xilinxgitops:
retest this please

jvillarre (approved):
Changes look good to instantiate objects instead of declaring functions.

PR #4437, opened 2020-11-11
fixing petalinux related builds due to bsp path

Himanshu-xilinx:
* its a temporary fix for pipeline as petalinx linux is in release phase and 1109_1 bsp's are not available

### DIFF
--- a/build/build_edge.sh
+++ b/build/build_edge.sh
@@ -138,6 +138,7 @@ fi
 if [ -f $SETTINGS_FILE ]; then
     source $SETTINGS_FILE
 fi
+PETALINUX="/proj/petalinux/2020.2/petalinux-v2020.2_1108_1/tool/petalinux-v2020.2-final"
 source $PETALINUX/settings.sh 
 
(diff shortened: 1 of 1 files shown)


build/build_edge.sh
```diff
@@ -138,6 +138,7 @@ fi
 if [ -f $SETTINGS_FILE ]; then
     source $SETTINGS_FILE
 fi
+PETALINUX="/proj/petalinux/2020.2/petalinux-v2020.2_1108_1/tool/petalinux-v2020.2-final"
```
maxzhen, line 141: Can we ask PetaLinux team to publish new path and keep old one as a sym link for some days, so that we can get a window to fix our reference?

chvamshi-xilinx:
Yes @maxzhen , I have already asked them to do that. This happened due to following reason. Petalinux is frozen, but Vivado is not frozen, they want to respin the build on same area with new Vivado. We have a problem while that build is happening.
Currently, we are using one day older build to work-around that issue. its not clean, but till we get a sym-link from petalinux, we can go with this.

PR #4443, opened 2020-11-11
workaround petalinux build issue

chvamshi-xilinx:


chvamshi-xilinx:
This needs to be merged first as other PRs may fail due to the petalinux build issue. Merging as there is no impact.

PR #4444, opened 2020-11-11
Fix verison -> version typos

keryell:
Thanks to the spell-checker of Slack for finding these!


maxzhen:
ok to test

AShivangi (approved):
Thanks for catching these!

PR #4455, opened 2020-11-12
Fix issue with reading AIE counter and GMIO metadata when PL device trace offload is enabled : CR 1083019

IshitaGhosh:


jvillarre (approved):
Changes look good in 2020.2 branch

PR #4457, opened 2020-11-13
prompt warm reboot after configuring P2P

houlz0507:


maxzhen (approved):

AShivangi (approved):

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1166,10 +1166,18 @@ int shim::p2pEnable(bool enable, bool force)
...
 
     if (mDev == nullptr)
         return -EINVAL;
 
+    ret = check_p2p_config(mDev, err);
+    if (ret == P2P_CONFIG_ENABLED && enable) {
+        throw std::runtime_error("P2P is already enabled");
+    } else if (ret == P2P_CONFIG_DISABLED && !enable) {
+        throw std::runtime_error("P2P is already disabled");
+    }
```
stsoe, line 1179: This will not work as expected.  Who is catching these exceptions when called from xclP2PEnable via xrt-next.h in C code?  All internal usage would be okay, but care must be taken when we expose these APIs as C-APIs also.
Fix in https://github.com/Xilinx/XRT/pull/4462

PR #4459, opened 2020-11-14
Use PRIVATE link dependencies

stsoe:
target_link_libraries should use PRIVATE link dependencies by default.

This controls what libraries dependents are linked with.  By default dependents must specify their explicit dependencies and not rely on implicit dependencies from link library.   Only use PUBLIC for dependencies used in header files.

Libraries and targets following PUBLIC are linked to, and are made part of the link interface. Libraries and targets following PRIVATE are linked to, but are not made part of the link interface.


stsoe:
@chvamshi-xilinx Please review if this change by any chance could impact petalinux.  I don't think it does, but I don't have full visibility into how emulation libraries are link in terms of resolving dependencies.

chvamshi-xilinx (approved):

chvamshi-xilinx:
> @chvamshi-xilinx Please review if this change by any chance could impact petalinux. I don't think it does, but I don't have full visibility into how emulation libraries are link in terms of resolving dependencies.

Changes are good @stsoe . There wont be any impact on petalinux builds.

PR #4460, opened 2020-11-14
XRT Graph c++ API

larry9523:
This PR added XRT Graph C++ API. (AIE level will C++ API will be followed). An example usage of Graph C++ API looks like

auto device = xrt::device(0);
auto uuid = device.load_xclbin(xclbin_file);
auto graph = xrt::graph(device, uuid, "mygraph");
graph.reset();
graph.run(3);
graph.update("mygraph.k1.in[2]", inputVector);
graph.wait();
graph.read("mygraph.k2.inout[2]", outputVectorAsync[0]);
graph.update("mygraph.k1.in[2]", inputVector); 
graph.run(3);
graph.end();

### DIFF
--- a/src/runtime_src/core/common/api/aie/xrt_graph.cpp
+++ b/src/runtime_src/core/common/api/aie/xrt_graph.cpp
@@ -108,19 +108,7 @@ public:
   }
 
   void
-  update_rtp(const std::string& port, const char* buffer, size_t size)
-  {
-    device->update_graph_rtp(handle, port, buffer, size);
-  }
-
-  void
   read_rtp(const char* port, char* buffer, size_t size)
-  {
-    device->read_graph_rtp(handle, port, buffer, size);
-  }
-
-  void
-  read_rtp(const std::string& port, char* buffer, size_t size)
   {
     device->read_graph_rtp(handle, port, buffer, size);
   }
@@ -316,14 +304,14 @@ void
 graph::
 update_port(const std::string& port_name, const void* value, size_t bytes)
 {
-  handle->update_rtp(port_name, reinterpret_cast<const char*>(value), bytes);
+  handle->update_rtp(port_name.c_str(), reinterpret_cast<const char*>(value), bytes);
 }
 
 void
 graph::
 read_port(const std::string& port_name, void* value, size_t bytes)
 {
-  handle->read_rtp(port_name, reinterpret_cast<char *>(value), bytes);
+  handle->read_rtp(port_name.c_str(), reinterpret_cast<char *>(value), bytes);
 }
 
 } // namespace xrt
--- a/src/runtime_src/core/common/api/aie/xrt_graph.cpp
+++ b/src/runtime_src/core/common/api/aie/xrt_graph.cpp
@@ -249,22 +249,26 @@ run(uint32_t iterations)
 
 void
 graph::
-wait(int timeout_ms)
+wait(std::chrono::milliseconds timeout_ms)
 {
-  if (timeout_ms == 0)
+  if (timeout_ms.count() == 0)
     handle->wait(static_cast<uint64_t>(0));
   else
-    handle->wait(timeout_ms);
+    handle->wait(static_cast<int>(timeout_ms.count()));
 }
 
 void
 graph::
-suspend(uint64_t cycles)
+wait(uint64_t cycles)
 {
-  if (cycles == 0)
-    handle->suspend();
-  else
-    wait(cycles);
+  handle->wait(cycles);
+}
+
+void
+graph::
+suspend()
+{
+  handle->suspend();
 }
 
 void
--- a/src/runtime_src/core/common/api/aie/xrt_graph.cpp
+++ b/src/runtime_src/core/common/api/aie/xrt_graph.cpp
@@ -108,7 +108,19 @@ public:
   }
 
   void
+  update_rtp(const std::string& port, const char* buffer, size_t size)
+  {
+    device->update_graph_rtp(handle, port, buffer, size);
+  }
+
+  void
   read_rtp(const char* port, char* buffer, size_t size)
+  {
+    device->read_graph_rtp(handle, port, buffer, size);
+  }
+
+  void
+  read_rtp(const std::string& port, char* buffer, size_t size)
   {
     device->read_graph_rtp(handle, port, buffer, size);
   }
@@ -127,6 +139,15 @@ open_graph(xrtDeviceHandle dhdl, const uuid_t xclbin_uuid, const char* graph_nam
 {
   auto device = xrt_core::device_int::get_core_device(dhdl);
   auto handle = device->open_graph(xclbin_uuid, graph_name);
+  auto ghdl = std::make_shared<xrt::graph_impl>(device, handle);
+  return ghdl;
+}
+
+static std::shared_ptr<xrt::graph_impl>
+open_graph(xclDeviceHandle dhdl, const xrt::uuid& xclbin_id, const std::string& name)
+{
+  auto device = xrt_core::get_userpf_device(dhdl);
+  auto handle = device->open_graph(xclbin_id.get(), name.c_str());
   auto ghdl = std::make_shared<xrt::graph_impl>(device, handle);
   return ghdl;
 }
@@ -203,6 +224,110 @@ send_exception_message(const char* msg)
 }
 
 }
+
+//////////////////////////////////////////////////////////////
+// xrt_graph C++ API implementations (xrt_graph.h)
+//////////////////////////////////////////////////////////////
+namespace xrt {
+
+graph::
+graph(const xrt::device& device, const xrt::uuid& xclbin_id, const std::string& name)
+  : handle(open_graph(device, xclbin_id, name))
+{}
+
+void
+graph::
+reset() const
+{
+  handle->reset();
+}
+
+uint64_t
+graph::
+get_timestamp() const
+{
+  return (handle->get_timestamp());
+}
+
+void
+graph::
+run()
+{
+  handle->run(0);
+}
+
+void
+graph::
+run(uint32_t iterations)
+{
+  handle->run(iterations);
+}
+
+void
+graph::
+wait_done(int timeout_ms)
+{
+  handle->wait(timeout_ms);
+}
+
+void
+graph::
+wait()
+{
+  handle->wait(0);
+}
+
+void
+graph::
+wait(uint64_t cycles)
+{
+  handle->wait(cycles);
+}
+
+void
+graph::
+suspend()
+{
+  handle->suspend();
+}
+
+void
+graph::
+resume()
+{
+  handle->resume();
+}
+
+void
+graph::
+end()
+{
+  handle->end(0);
+}
+
+void
+graph::
+end(uint64_t cycles)
+{
+  handle->end(cycles);
+}
+
+void
+graph::
+update_port(const std::string& port_name, const void* value, size_t bytes)
+{
+  handle->update_rtp(port_name, reinterpret_cast<const char*>(value), bytes);
+}
+
+void
+graph::
+read_port(const std::string& port_name, void* value, size_t bytes)
+{
+  handle->read_rtp(port_name, reinterpret_cast<char *>(value), bytes);
+}
+
+} // namespace xrt
+
 
 ////////////////////////////////////////////////////////////////
 // xrt_aie API implementations (xrt_aie.h, xrt_graph.h)
--- a/src/runtime_src/core/common/api/aie/xrt_graph.cpp
+++ b/src/runtime_src/core/common/api/aie/xrt_graph.cpp
@@ -21,6 +21,7 @@
 // core/include/experimental/xrt_graph.h -- end user APIs
 // core/include/xcl_graph.h -- shim level APIs
 #include "core/include/experimental/xrt_aie.h"
+#include "core/include/experimental/xrt_bo.h"
 #include "core/include/xcl_graph.h"
 
 #include "core/include/experimental/xrt_device.h"
@@ -160,7 +161,8 @@ static void
 sync_aie_bo(xrtDeviceHandle dhdl, xrtBufferHandle bohdl, const char *gmio_name, xclBOSyncDirection dir, size_t size, size_t offset)
 {
   auto device = xrt_core::device_int::get_core_device(dhdl);
-  device->sync_aie_bo(bohdl, gmio_name, dir, size, offset);
+  auto bo = xrt::bo(bohdl);
+  device->sync_aie_bo(bo, gmio_name, dir, size, offset);
 }
 
 static void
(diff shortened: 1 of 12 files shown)


src/runtime_src/core/common/api/aie/xrt_graph.cpp
```diff
@@ -204,6 +225,110 @@ send_exception_message(const char* msg)
...
+// xrt_graph C++ API implementations (xrt_graph.h)
+//////////////////////////////////////////////////////////////
+namespace xrt {
+
+graph::
+graph(const xrt::device& device, const xrt::uuid& xclbin_id, const std::string& name)
+  : handle(open_graph(device, xclbin_id, name))
+{}
+
+void
```
keryell, line 238: Perhaps for such simple things you could have 4 lines in 1 line.
Also I am wondering whether all these simple functions could no be placed directly in a `.hpp` instead to allow inlining. Or do you want to have the `handle` hidden from the public headers?

src/runtime_src/core/common/api/aie/xrt_graph.cpp
```diff
@@ -204,6 +225,110 @@ send_exception_message(const char* msg)
...
+reset() const
+{
+  handle->reset();
+}
+
+uint64_t
+graph::
+get_timestamp() const
+{
+  return (handle->get_timestamp());
```
keryell, line 249:
```suggestion
  return handle->get_timestamp();
```

src/runtime_src/core/common/api/aie/xrt_graph.cpp
```diff
@@ -204,6 +225,110 @@ send_exception_message(const char* msg)
...
+void
+graph::
+run()
+{
+  handle->run(0);
+}
+
+void
+graph::
+run(uint32_t iterations)
```
keryell, line 261: Put in the header
```suggestion
run(std::uint32_t iterations = 0)
```
and remove the previous overload?

src/runtime_src/core/common/api/aie/xrt_graph.cpp
```diff
@@ -204,6 +225,110 @@ send_exception_message(const char* msg)
...
+{
+  handle->run(iterations);
+}
+
+void
+graph::
+wait_done(int timeout_ms)
```
keryell, line 268: `wait_done` sounds a little bit strange.

src/runtime_src/core/common/api/aie/xrt_graph.cpp
```diff
@@ -204,6 +225,110 @@ send_exception_message(const char* msg)
...
+void
+graph::
+wait()
+{
+  handle->wait(0);
+}
+
+void
+graph::
+wait(uint64_t cycles)
```
keryell, line 282: Add in the header
```suggestion
wait(std::uint64_t cycles = 0)
```
and remove the previous overload.

src/runtime_src/core/common/api/aie/xrt_graph.cpp
```diff
@@ -204,6 +225,110 @@ send_exception_message(const char* msg)
...
+void
+graph::
+end()
+{
+  handle->end(0);
+}
+
+void
+graph::
+end(uint64_t cycles)
```
keryell, line 310: Idem...

src/runtime_src/core/common/api/aie/xrt_graph.cpp
keryell: Furthermore I am a little worried by `_ms` here  that does not smell very C++-ish.
Could we use something similar to https://en.cppreference.com/w/cpp/thread/sleep_for instead?

src/runtime_src/core/common/api/aie/xrt_graph.cpp
```diff
@@ -204,6 +225,110 @@ send_exception_message(const char* msg)
...
+run(uint32_t iterations)
+{
+  handle->run(iterations);
+}
+
+void
+graph::
+wait_done(int timeout_ms)
+{
+  handle->wait(timeout_ms);
```
keryell, line 270: And the conversion to `ms` would be done here.

src/runtime_src/core/include/experimental/xrt_graph.h
```diff
@@ -26,6 +26,205 @@
 
 typedef void *xrtGraphHandle;
```
keryell, line 27:
```suggestion
using xrtGraphHandle = void*;
```
but actually you should have something more type-safe like:
```c++
struct xrtGraphHandle {
  void* handle;

  // Some accessors or converters...
};
```

keryell (changes requested):
That seems interesting!
Is there already a documentation in this project about it?
I am curious about the meaning of `3` for example.
Does `"mygraph.k2.inout[2]"` means a decoration or is this a kind of Python which is parsed inside the implementation?

A few nitpicks otherwise.

larry9523:
> That seems interesting!
> Is there already a documentation in this project about it?
> I am curious about the meaning of `3` for example.
> Does `"mygraph.k2.inout[2]"` means a decoration or is this a kind of Python which is parsed inside the implementation?
> 
> A few nitpicks otherwise.

"mygraph.k2.inout[2]" is simply an example of RTP port name from a real design. It was generated by aietool based on the graph code in that design.

src/runtime_src/core/common/api/aie/xrt_graph.cpp
stsoe: I am left leaning and so it shall remain :-)
I believe open_graph is internal.

src/runtime_src/core/common/api/aie/xrt_graph.cpp
stsoe: Yes,  should use std::chrono::milliseconds, but do keep wait(unsinged int timeout_ms) as inline (see xrt_kernel.h).  I actually thing the timeout_ms is simpler than chrono, but agree we should have the chrono one also

src/runtime_src/core/common/api/aie/xrt_graph.cpp
stsoe: Maybe just uint64_t and a wholesale change if <cstdint> can't be counted on to also put the types in global namespace, which all current implementations do.

src/runtime_src/core/common/api/aie/xrt_graph.cpp
stsoe: what does Idem... mean?

src/runtime_src/core/include/experimental/xrt_graph.h
stsoe: This is a C header as well as a C++ header, leave as typedef.

src/runtime_src/core/common/api/aie/xrt_graph.cpp
```diff
@@ -107,11 +107,23 @@ class graph_impl {
...
   read_rtp(const char* port, char* buffer, size_t size)
   {
     device->read_graph_rtp(handle, port, buffer, size);
   }
+
+  void
+  read_rtp(const std::string& port, char* buffer, size_t size)
+  {
+    device->read_graph_rtp(handle, port, buffer, size);
+  }
```
stsoe, line 126: Is there an reason to add the shim read_graph_rtp for the std::string version?   For convenience it may make sense to keep read_rtp for std::string, but maybe it should simple call read_rtp(port.c_str(), buffer, size)?   

If on the other hand the const char* version of read_rtp ends up converting the port to a std::string at a lower level then isn't read_rtp(const std::stirng&, ...) sufficient?

src/runtime_src/core/common/ishim.h
```diff
@@ -142,9 +142,15 @@ struct ishim
   virtual void
   update_graph_rtp(xclGraphHandle handle, const char* port, const char* buffer, size_t size) = 0;
 
+  virtual void
+  update_graph_rtp(xclGraphHandle handle, const std::string& port, const char* buffer, size_t size) = 0;
+
```
stsoe, line 147: Are both needed?

src/runtime_src/core/common/ishim.h
```diff
@@ -142,9 +142,15 @@ struct ishim
...
   virtual void
   read_graph_rtp(xclGraphHandle handle, const char* port, char* buffer, size_t size) = 0;
 
+  virtual void
+  read_graph_rtp(xclGraphHandle handle, const std::string& port, char* buffer, size_t size) = 0;
+
```
stsoe, line 153: Both needed?

src/runtime_src/core/common/api/aie/xrt_graph.cpp
larry9523: It is on my todo list to use char* to replace std::string& for port name in the rtp_type unordered map and function in xcl level rtp update/read to increase rtp performance by skipping the string allocator. That is why I keep read_trp(const char * ...). I have thought since the c++ api already take std::string (C api takes char *), we should keep using std::string.

But rethink about it, I agree with you, read_rtp(port.c_str(),...) should just work. I will remove the read_rtp(std::string&...) version

src/runtime_src/core/common/api/aie/xrt_graph.cpp
larry9523: done

src/runtime_src/core/common/api/aie/xrt_graph.cpp
larry9523: Providing two interfaces for run, wait and end, with/without iterations and cycles, are to matching the aietool API so that we have similar concept about run() and run(iterations), etc.

src/runtime_src/core/common/api/aie/xrt_graph.cpp
larry9523: I put this onto my todo list to add std::chrono::milliseconds interface. Will implement it in next PR.

src/runtime_src/core/common/api/aie/xrt_graph.cpp
larry9523: std::string& functions are removed.

src/runtime_src/core/common/api/aie/xrt_graph.cpp
sonals: I would suggest the same as @keryell. run() can have default of 0.

src/runtime_src/core/include/experimental/xrt_graph.h
```diff
@@ -26,6 +26,205 @@
...
+   * graph immediately.
+   *
+   * This API with non-zero AIE cycle is for grahp that is running forever
+   * or graph that has multi-rate core(s); zero AIE cycle means busy wait
+   * until graph is done. It has the same effect as wait() API.
+   *
+   * The current thread will block until graph run completes or is paused.
+   */
+  void
+  wait(uint64_t cycles);
```
sonals, line 132: I wonder if we need only one wait with overloaded argument types:
```
void wait() // runt to finish
void wait(std::chrono::milliseconds timeout) // run to finish or timeout
void wait(uint64_t cycles) //run to finish or run for specified cycles
```

src/runtime_src/core/common/api/aie/xrt_graph.cpp
keryell: Sorry for being obscure.
I meant:
```c++
end(std::uint64_t cycles = 0)
```
and remove the plain `end()` overload at the same time.

src/runtime_src/core/include/experimental/xrt_graph.h
keryell: OK, then forgot about it (and type-safety...).

src/runtime_src/core/include/experimental/xrt_graph.h
keryell: I have the feeling that with @stsoe we agreed on this, even having the first one redundant by using an `= 0` default argument.

stsoe: Yes indeed :-)

uday610 (commented):
because suspend API does things forcefully. So it is better to have two version of suspend, immediately or after specified number of AIE cycles. 

In summary, all I am thinking of doing wait API for wait, suspend API for suspend. Current API design in C is very mixed and not intuitive. But we can make it better in C++ by changing underlying implementation.

uday610:
Larry 

Regarding graph C++ APIs (which are for 2021.1) I am thinking we can do little bit of simplification of few of the APIs. 

Specially, the APIs xrtGraphWait, xrtGraphWaitDone, xrtGraphSuspend are all used differently and not intuitive from the coding aspect. 

For example, xrtGraphWait(graph,0) waits indefinitely till the graph is ending, it does not do anything forcefully. 

But xrtGraphWait(graphHandle,3000); suspends the graph forcefully, which seems odd, because we have an API xrtGraphSuspend already for suspending task. So why not using xrtGraphSuspend for all suspending task (by overloading)? 

Also in the same manner xrtGraphWaitDone is very odd name. All it does it waits for some specified time, and if graph is not finished, it quits. So this is just a timeout version of xrtGraphWait(graph,0). So why not just obsolete this xrtGraphWaitDone function and overload it with xrtGraphWait? 

I know for C the APIs are already in use (though the names are not meaningful and very less intuitive), but as C++ APIs are new (and never used) can we make them more meaningful? 

In my opinion, it is should not be a requirement to have exact C and C++ matching. 

So I propose the following. 


•	xrt::graph::wait(graph_handle)     This is equivalent to  xrtGraphWait(graph,0) , which waits indefinitely. 
•	xrt::graph::wait(graph_handle, timeout_ms)  This should equivalent to xrtGraphWaitDone(graphHandle, timeout)
o	Naming wise this makes more sense because xrt::graph::wait is not doing anything forcefully it simply waits for sometime 
o	We are getting rid of that very ODDLY naming API xrt::graph::graphWaitDone 
•	xrt::graph::suspend(graph_handle)  Suspend the graph immediately.
•	xrt::graph::suspend(graph_handle, timeout)  This is equivalent to xrtGraphWait(graphHandle,3000) 
o	Naming wise makes more sense

src/runtime_src/core/common/api/aie/xrt_graph.cpp
larry9523: Does it mean we should explicitly remove run() API? It seems we need a clearly definition of 
run(0)
wait(0)
end(0)

src/runtime_src/core/include/experimental/xrt_graph.h
```diff
@@ -182,4 +381,10 @@ xrtGraphUpdateRTP(xrtGraphHandle gh, const char *hierPathPort, const char *buffe
 int
 xrtGraphReadRTP(xrtGraphHandle gh, const char *hierPathPort, char *buffer, size_t size);
 
+/// @endcond
+
+#ifdef __cplusplus
+}
+#endif
+
```
AlphaBu, line 389: For the API's for user, is it good to keep consistence between C version and C++ version?
xrtGraphOpen  ->  open  //May have
xrtGraphClose ->  close
xrtGraphReset ->  reset
xrtGraphTimeStamp -> timeStamp
xrtGraphRun -> run
xrtGraphWaitDone -> waitDone
xrtGraphWait -> wait
xrtGraphSuspend -> suspend
xrtGraphResume -> resume
xrtGraphEnd -> end
xrtGraphUpdateRTP -> updateRTP
xrtGraphReadRTP -> readRTP

keryell: Who is your typical ser who will use this API both in C and in C++ at the same time?
Even just 1 user using just 1 API seems already rare and anyway will require to look at the documentation.
For me, I will never use the C API, maybe the C++ API...
So, yes, improve the C++ API. :-)

stsoe (approved):

PR #4461, opened 2020-11-14
CR-1079673 XRT build failure with cmake 3.17 and boost 1.65.1

stsoe:
Fix code to work-around boost::process rdbuf() conversion error.
Add proper Boost version check to test if boost::process is available.

### DIFF
--- a/src/CMake/cpackLin.cmake
+++ b/src/CMake/cpackLin.cmake
@@ -28,15 +28,6 @@ endif()
 
 include (CMake/glibc.cmake)
 
-# Trick to get the Boost Version string and one version greater
-SET(Boost_MINOR_VERSION_ONEGREATER "${Boost_MINOR_VERSION}")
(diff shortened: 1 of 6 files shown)


AShivangi (commented):
Thanks for making these changes :)

rozumx (commented):
Approve changes.

keryell:
Curious to use a Boost from 2017 and a CMake from 2020 at the same time...
Why not just using modern software? :-)

stsoe:
Hi @keryell   Unfortunately we have to support Ubuntu 16.04, 18.04, 20.04, ... Centos 7.x, Centos 8.x .  System libraries, including Boost and cmake, are what is default installed on these platforms.

keryell:
Sure... But usually you have everything new or everything old... Here this is a strange cocktail...

PR #3279, opened 2020-04-25
CR-1062988 clPollStreams fix and improvement

houlz0507:

PR #4462, opened 2020-11-14
Prevent exceptions from escaping through C API calls

stsoe:
Amend #4457, #3279
Catch expcetions before returning.


houlz0507 (approved):

keryell:
Do you need also a `catch (...)`?

PR #4467, opened 2020-11-15
CR-1077647: soft kernel verbosity n logging support

sarab96:
soft kernel verbosity n logging support

### DIFF
--- a/src/runtime_src/core/edge/skd/sk_daemon.cpp
+++ b/src/runtime_src/core/edge/skd/sk_daemon.cpp
@@ -31,7 +31,7 @@
 
 xclDeviceHandle devHdl;
 
-unsigned int getHostBO(unsigned long paddr, size_t size)
+static unsigned int getHostBO(unsigned long paddr, size_t size)
 {
   unsigned int boHandle;
 
@@ -40,41 +40,36 @@ unsigned int getHostBO(unsigned long paddr, size_t size)
   return boHandle;
 }
 
-void *mapBO(unsigned int boHandle, bool write)
+static void *mapBO(unsigned int boHandle, bool write)
 {
   void *buf;
   buf = xclMapBO(devHdl, boHandle, write);
   return buf;
 }
 
-void freeBO(unsigned int boHandle)
+static void freeBO(unsigned int boHandle)
 {
   xclFreeBO(devHdl, boHandle);
 }
 
-int logMsg(xrtLogMsgLevel level, const char* tag,
+static int logMsg(xrtLogMsgLevel level, const char* tag,
 		       const char* format, ...)
 {
   static auto verbosity = xrt_core::config::get_verbosity();
-  if (level <= verbosity) {
-    va_list args;
-    va_start(args, format);
-    int ret = -1;
-    ZYNQ::shim *drv = ZYNQ::shim::handleCheck(devHdl);
-    ret = drv ? drv->xclLogMsg(level, tag, format, args) : -ENODEV;
-    va_end(args);
-
-    return ret;
+  if (level > verbosity) {
+    return 0;
   }
+  va_list args;
+  va_start(args, format);
+  int ret = -1;
+  ZYNQ::shim *drv = ZYNQ::shim::handleCheck(devHdl);
+  ret = drv ? drv->xclLogMsg(level, tag, format, args) : -ENODEV;
+  va_end(args);
 
-  return 0;
-
-
-//sarab
-//  return xclLogMsg(devHdl, level, tag, format, (char*) "test");
+  return ret;
 }
 
-int getBufferFd(unsigned int boHandle)
+static int getBufferFd(unsigned int boHandle)
 {
     return xclExportBO(devHdl, boHandle);
 }
@@ -118,7 +113,7 @@ static int destroySoftKernel(unsigned int boh, void *mapAddr)
  * This function calls XRT interface to notify a soft kenel is idle
  * and wait for next command.
  */
-int waitNextCmd(uint32_t cu_idx)
+static int waitNextCmd(uint32_t cu_idx)
 {
   return xclSKReport(devHdl, cu_idx, XRT_SCU_STATE_DONE);
 }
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/edge/skd/sk_daemon.cpp
```diff
@@ -19,13 +19,15 @@
...
 #include <string.h>
+#include <cstdarg>
 #include <sys/wait.h>
 #include <sys/mman.h>
 #include <sys/prctl.h>
 
 #include "sk_types.h"
 #include "sk_daemon.h"
-#include "xclhal2_mpsoc.h"
+#include "core/common/config_reader.h"
```
stsoe, line 29: I happen to be touching sdk/CMakeLists.txt currently.   Adding and using symbols from this header files implies that you must explicitly link with xrt_coreutil.  Please add that link dependency to this pull request.

stsoe (changes requested):

src/runtime_src/core/edge/skd/sk_daemon.cpp
```diff
@@ -50,6 +52,28 @@ void freeBO(unsigned int boHandle)
   xclFreeBO(devHdl, boHandle);
 }
 
+int logMsg(xrtLogMsgLevel level, const char* tag,
```
sonals, line 55: `static int logMsg(xrtLogMsgLevel level, const char* tag,`

src/runtime_src/core/edge/skd/sk_daemon.cpp
```diff
@@ -50,6 +52,28 @@ void freeBO(unsigned int boHandle)
...
+		       const char* format, ...)
+{
+  static auto verbosity = xrt_core::config::get_verbosity();
+  if (level <= verbosity) {
```
sonals, line 59: Please use _early exit_. Invert the if logic so the core code in not inside indentation.
```
if (level > verbosity)
  return 0
```

sonals (changes requested):
Shouldn't we change all places where we are calling syslog directly to use the new logging API?

src/runtime_src/core/edge/skd/sk_daemon.cpp
```diff
@@ -50,6 +52,28 @@ void freeBO(unsigned int boHandle)
...
+    ret = drv ? drv->xclLogMsg(level, tag, format, args) : -ENODEV;
+    va_end(args);
+
+    return ret;
+  }
+
+  return 0;
+
+
+//sarab
```
larry9523, line 73: remove?

larry9523 (changes requested):
Any reason we want to replicate xclLogMsg code here instead of calling xclLogMsg directly?

src/runtime_src/core/edge/skd/sk_daemon.cpp
sarab96: done

src/runtime_src/core/edge/skd/sk_daemon.cpp
sarab96: done

src/runtime_src/core/edge/skd/sk_daemon.cpp
sarab96: done

src/runtime_src/core/edge/skd/sk_daemon.cpp
sarab96: done

sarab96:
Will change syslog call in separate PR after separate test of that
> Shouldn't we change all places where we are calling syslog directly to use the new logging API?

sarab96:
soft kernels will not link to xrt_core/shim, so separate func in sk_types.h. And xclLogMsg doesn't take va_list as argument
> Any reason we want to replicate xclLogMsg code here instead of calling xclLogMsg directly?

src/runtime_src/core/edge/skd/sk_daemon.cpp
```diff
@@ -38,19 +40,36 @@ unsigned int getHostBO(unsigned long paddr, size_t size)
...
+    return 0;
+  }
+  va_list args;
+  va_start(args, format);
+  int ret = -1;
+  ZYNQ::shim *drv = ZYNQ::shim::handleCheck(devHdl);
+  ret = drv ? drv->xclLogMsg(level, tag, format, args) : -ENODEV;
+  va_end(args);
+
+  return ret;
```
stsoe, line 69: How about consolidating all these logMsg functions into xrt_core::message (core/common/message.h)?  Add the missing va_list function there.   

shim::xclLogMsg should never have been a member function of the shim class in the first place.  You can ignore this comment now, since it is a little bit out of scope.  But this is one thing we should change at some point.

larry9523:
> soft kernels will not link to xrt_core/shim, so separate func in sk_types.h. And xclLogMsg doesn't take va_list as argument
> 
> > Any reason we want to replicate xclLogMsg code here instead of calling xclLogMsg directly?

1) I am not saying calling xclLogMsg in softkernel.
2) My preference is to provide a shim level log function with va_list. But I am not going to insist this since we will clean up shim::xclLogMsg later.

The idea when skd was initially written was make use of libxrt_core but not implementing one.

src/runtime_src/core/edge/skd/sk_daemon.cpp
sarab96: Agree. Will look into this separately

sarab96:
Agree. Separately will look at adding it in xrt::message namespace and then using it in all other places

> > soft kernels will not link to xrt_core/shim, so separate func in sk_types.h. And xclLogMsg doesn't take va_list as argument
> > > Any reason we want to replicate xclLogMsg code here instead of calling xclLogMsg directly?
> 
> 1. I am not saying calling xclLogMsg in softkernel.
> 2. My preference is to provide a shim level log function with va_list. But I am not going to insist this since we will clean up shim::xclLogMsg later.
> 
> The idea when skd was initially written was make use of libxrt_core but not implementing one.

larry9523 (approved):

PR #4468, opened 2020-11-16
updated cmd parser to fix coverity issue

vishnuchebrolu:


tests/validate/common/includes/cmdparser/cmdlineparser.h
```diff
@@ -38,7 +38,15 @@ class CmdLineParser {
...
+    CmdSwitch() {
+      key = "--help";
+      shortcut = "-h";
+      default_value = "0";
+      value = "0";
+      desc = "Command Line Parser";
+      istoggle = true;
+      isvalid = true;
+    }
     CmdSwitch(const CmdSwitch &rhs) { copyfrom(rhs); }
```
stsoe, line 50: The copy constructor and the copy assignment operators are not needed.

stsoe (changes requested):

tests/validate/common/includes/cmdparser/cmdlineparser.h
```diff
@@ -38,7 +38,15 @@ class CmdLineParser {
...
-    CmdSwitch() {}
+    CmdSwitch() {
+      key = "--help";
+      shortcut = "-h";
+      default_value = "0";
+      value = "0";
+      desc = "Command Line Parser";
+      istoggle = true;
+      isvalid = true;
+    }
```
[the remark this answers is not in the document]
vishnuchebrolu: updated.

tests/validate/common/includes/cmdparser/cmdlineparser.h
vishnuchebrolu: updated.

vishnuchebrolu:
@stsoe Can you please review the pull request.

tests/validate/common/includes/cmdparser/cmdlineparser.h
```diff
@@ -88,7 +69,7 @@ class CmdLineParser {
   /*!
    * retrieve value using a key
    */
-  string value(const char *key);
+  std::string value(const char *key);
```
stsoe, line 72: Looking at the data members in this class they are all std::string, so presumably 'key' will converted to a std::string when looking up in one of the maps,  The conversion should be done at the API level so the signature should be `std::string value(const std::string& key)`.  This makes the API more flexible as it can be called with both a char* and a std::string.

tests/validate/common/includes/cmdparser/cmdlineparser.h
```diff
@@ -88,7 +69,7 @@ class CmdLineParser {
...
 
   int value_to_int(const char *key);
```
stsoe, line 74: Same comment as above, though I am not sure what this function does.

tests/validate/common/includes/cmdparser/cmdlineparser.h
```diff
@@ -110,17 +91,18 @@ class CmdLineParser {
    */
   CmdSwitch *getCmdSwitch(const char *key);
 
-  bool token_to_fullkeyname(const string &token, string &fullkey);
+  bool token_to_fullkeyname(const std::string &token, std::string &fullkey);
```
stsoe, line 94: Little awkward to have the function pass its output in fullkey rather than return the full key.  I understand that bool return value indicates if the fullkey was there, but couldn't an empty string do the same?

You may also want to comment the API.

stsoe (commented):
These are minor comments.   Let me know if you plan on changing.  I am okay to merge as is if you prefer.

PR #4472, opened 2020-11-17
CR-1083074: changed linking from xdp to xdp_core

pgschuey:


pgschuey:
Please re-test

stsoe:
The overloaded va_list send method appears to be conflicting with the already present inline method.  This begs the question why we need the va_list one in the first place?  If you want / need it, at least rename it.

dayeh-xilinx:
retest this please

jvillarre (approved):
Changes look good and set up the right dependencies for the plugin

dayeh-xilinx:
retest this please

PR #4478, opened 2020-11-17
XRT-813 Use multiple threads to test IOPS

mamin506:
Compile:
make xcl_api_iops

Usage:
./xcl_api_iops -d -k /opt/xilinx/.../test/verify.xclbin -t 3 -l 64 -a 1000000

Use 3 submitted threads could get best overall performance. 

xrt_api_iops test is updated as well.

### DIFF
--- a/tests/xrt/perf_IOPS/xcl_api_iops.cpp
+++ b/tests/xrt/perf_IOPS/xcl_api_iops.cpp
@@ -226,7 +226,7 @@ int testMultiThreads(xclDeviceHandle handle, xuid_t uuid, int bank,
         arg[i].handle = handle;
         arg[i].queueLength = queueLength;
         arg[i].total = total;
-        threads[i] = std::move(std::thread(runTestThread, std::ref(arg[i])));
+        threads[i] = std::thread([&](int i){ runTestThread(arg[i]); }, i);
     }
 
     /* Wait threads to prepare to start */
--- a/tests/xrt/perf_IOPS/xcl_api_iops.cpp
+++ b/tests/xrt/perf_IOPS/xcl_api_iops.cpp
@@ -4,19 +4,17 @@
 #include <vector>
 #include <memory>
 #include <chrono>
+#include <thread>
 #include <sys/mman.h>
 #include <getopt.h>
 
+#include "xilutil.hpp"
 #include "xrt.h"
 #include "ert.h"
 #include "xclbin.h"
 
 using ms_t = std::chrono::microseconds;
 using Clock = std::chrono::high_resolution_clock;
-
-bool start = false;
-bool stop = false;
-pthread_barrier_t barrier;
 
 struct task_info {
     unsigned                boh;
@@ -34,9 +32,24 @@ typedef struct task_args {
     Clock::time_point end;
 } arg_t;
 
-void usage()
+bool start = false;
+bool stop = false;
+barrier barrier;
+
+static void usage(char *prog)
 {
-    printf("Usage: test -k <xclbin>\n");
+    std::cout << "Usage: " << prog << " -k <xclbin> -d <dev id> [options]\n"
+              << "options:\n"
+              << "    -t       number of threads\n"
+              << "    -l       length of queue (send how many commands without waiting)\n"
+              << "    -a       total amount of commands per thread\n"
+              << std::endl;
+}
+
+static void usage_and_exit(char *prog)
+{
+    usage(prog);
+    exit(0);
 }
 
 static std::vector<char>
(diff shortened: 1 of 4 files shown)


sonals (changes requested):
std::thread instead of pthread?

mamin506:
> std::thread instead of pthread?

Sure. Would use std::thread instead.

tests/xrt/perf_IOPS/xcl_api_iops.cpp
```diff
@@ -177,79 +189,76 @@ int testSingleThread(xclDeviceHandle handle, xuid_t uuid, int bank)
...
-        arg[i]->handle = handle;
-        arg[i]->queueLength = queueLength;
-        arg[i]->total = total;
-        pthread_create(&tids[i], NULL, runTestThread, arg[i]);
+        arg[i].thread_id = i;
+        arg[i].bank = bank;
+        arg[i].handle = handle;
+        arg[i].queueLength = queueLength;
+        arg[i].total = total;
+        threads[i] = std::move(std::thread(runTestThread, std::ref(arg[i])));
```
keryell, line 229: I guess the `std::move()` is redundant.
Also, in modern C++ you do not need these `std::ref` and old API.
You can just use a lambda and a normal API.
```c++
threads[i] = std::move { [&] { runTestThread(i, bank, handle, queueLength, total); } };
```
Modern C++ is more like Python than C and `void**`. :-)

tests/xrt/perf_IOPS/xilutil.hpp
```diff
@@ -0,0 +1,59 @@
...
+#ifndef XIL_UTIL_HPP
+#define XIL_UTIL_HPP
+
+#include <mutex>
+#include <condition_variable>
+
+/* std::barrier can be used since c++20
+ * Thanks boost library. Modified a little bit with pure c++11 code
+ */
+class barrier
```
keryell, line 12: Why not using Boost (or C++20...) at the first place instead of reimplementing libraries?

tests/xrt/perf_IOPS/xrt_api_iops.cpp
```diff
@@ -59,29 +90,127 @@ int testSingleThread(const xrt::device& device, const xrt::uuid& uuid)
...
+
+    auto hello = xrt::kernel(device, uuid.get(), "hello");
+
+    barrier.init(threadNumber + 1);
+
+    for (int i = 0; i < threadNumber; i++) {
+      arg[i].thread_id = i;
+      arg[i].queueLength = queueLength;
+      arg[i].total = total;
+      threads[i] = std::move(std::thread(runTestThread, std::ref(device), std::ref(hello), std::ref(arg[i])));
```
keryell, line 133: We are no longer in 1998... :-)
```suggestion
      threads[i] = std::thread { [&] { runTestThread(device, hello, arg[i]) } };
```
Perhaps
```c++
threads[i] = { [&] { runTestThread(device, hello, arg[i]) } };
```
could work too?

tests/xrt/perf_IOPS/xilutil.hpp
mamin506: Not every compiler can compile c++20. We have a machine that cannot compile c++14.

tests/xrt/perf_IOPS/xrt_api_iops.cpp
mamin506: Thanks @keryell . Use lambda function with std::thread is a great idea. Learn something new today.

Your suggestion missed that 'i' is local variable in the loop. Below line works.
`threads[i] = std::thread([&](int i){ runTestThread(device, hello, arg[i]); }, i);`

tests/xrt/perf_IOPS/xcl_api_iops.cpp
mamin506: Good suggestion. :)

tests/xrt/perf_IOPS/xrt_api_iops.cpp
keryell: Yes, this is a new language called... modern C++! ;-) I will invite you to #cpp on Slack.
Too bad C++ is not well known at Xilinx...
Anyway, I missed that if we capture `i` per reference we will be in trouble... :-(
Good to see you figured out a clever way to handle it!
But there is a simpler and official way to do this according to
https://en.cppreference.com/w/cpp/language/lambda
Could you try for me:
```c++
threads[i] = { [&, i] { runTestThread(device, hello, arg[i]); } };
```

maxzhen (approved):

stsoe (changes requested):
Let's take this off-line.  Given that we plan to publish these tests on GitHub, let's rewrite them with consistent naming convention, lots of comments, OS agnostic, and building with CMake.

maxzhen:
Closing for now.

PR #4479, opened 2020-11-18
CR-1076108 ASTeR TC 4.01 Firewall test causes server reboot

xdavidz:
I hope this won't cause u30 regression and still can catch the 0xDEADFA11 earlier.

This will prevent later read32 panic the system before heath check thread could find the firewall trip and reset.
Detailed reason listed in the CR.
https://jira.xilinx.com/browse/CR-1076108

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
@@ -3201,7 +3201,8 @@ static void xmc_enable_mailbox(struct xocl_xmc *xmc)
 
 	xmc->mbx_enabled = true;
 	safe_read32(xmc, XMC_HOST_MSG_OFFSET_REG, &val);
-	xmc->mbx_offset = val;
+	/* If val is not 32bit aligned, set back to default value */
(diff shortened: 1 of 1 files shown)


maxzhen (changes requested):
please create a function for reading mailbox offset and hide this ugly check inside.

xuhz (commented):
I am wondering whether this really can fix the issue if this issue needs to be fixed or worthy of fixing.
there are other possible register writes other than the mailbox write after firewall.
how about check the word alignment in io_write32 and don't write if not aligned?

maxzhen:
On a second thought, I agree with @xuhz that this fix is bogus. A firewall trip on **mgmt** pf is never OK. There is no way for driver to stay up and running when this happens. If firewall trip does happen on mgmt pf, it's a HW bug and must be fixed in HW. We can't workaround it inside drive. I'm closing this pull request and the CR.

xdavidz:
> I am wondering whether this really can fix the issue if this issue needs to be fixed or worthy of fixing.
> there are other possible register writes other than the mailbox write after firewall.
> how about check the word alignment in io_write32 and don't write if not aligned?

Please check the comments in the CR. I will include you and Max into the email loop with Niel.

PR #4486, opened 2020-11-19
Fix broken edge build

stsoe:
after. Since #4481 breaks #4467 the build is currently broken.
This PR fixes the problem.


stsoe:
retest this please

PR #4490, opened 2020-11-21
checks added, frame size

sarab96:
Frame size check added
null checks added


stsoe (approved):
Thanks for using nullptr, now that the file is open maybe you want to replace NULL with nullptr in the rest of the file.

sarab96:
Done
> Thanks for using nullptr, now that the file is open maybe you want to replace NULL with nullptr in the rest of the file.

PR #4492, opened 2020-11-23
Capture soft kernel return code - 1

sarab96:
Capture soft kernel return code
Capturing it in upper 8 bits of reg-0
Other changes for scheduler/ert to be done separately

### DIFF
--- a/src/runtime_src/core/edge/skd/sk_daemon.cpp
+++ b/src/runtime_src/core/edge/skd/sk_daemon.cpp
@@ -150,7 +150,7 @@ static void softKernelLoop(char *name, char *path, uint32_t cu_idx)
   kernel_t kernel;
   struct sk_operations ops;
   uint32_t *args_from_host;
-  uint32_t kernel_return;
+  int32_t kernel_return;
   unsigned int boh;
   int ret;
 
@@ -207,9 +207,7 @@ static void softKernelLoop(char *name, char *path, uint32_t cu_idx)
 
     /* Start run the soft kernel. */
     kernel_return = kernel(&args_from_host[1], &ops);
-    kernel_return &= 0xFF000000; //Keep only upper 8 bits
-    args_from_host[0] &= 0xFFFFFF; //Clear upper 8 bits
-    args_from_host[0] |= kernel_return;
+    args_from_host[1] = (uint32_t)kernel_return;
   }
 
   dlclose(sk_handle);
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/edge/skd/sk_daemon.cpp
```diff
@@ -201,11 +202,14 @@ static void softKernelLoop(char *name, char *path, uint32_t cu_idx)
...
     /* Reg file indicates the kernel should not be running. */
-    if (args_from_host[0] != 0x1)
-      continue;
+    if (!(args_from_host[0] & 0x1))
+      continue; //AP_START bit is not set; New Cmd is not available
 
     /* Start run the soft kernel. */
-    kernel(&args_from_host[1], &ops);
+    kernel_return = kernel(&args_from_host[1], &ops);
+    kernel_return &= 0xFF000000; //Keep only upper 8 bits
```
larry9523, line 210: I did not quite understand this "Keep only upper 8 bits". The signature of soft kernel function looks like
int soft_kernel(void *args, struct sk_operations *ops);
so it returns an integer. e.g. -EINVAL (-22, 0xFFFFFFEA). but after 
kernel_return &= 0xFF000000;
we got kernel_return = 0xFF00000000. We don't get any information from it, do we?

larry9523 (commented):
Two comments
1) We should use 32 bits instead of 8 bits to carry the soft kernel return code. a) We can reuse the args_from_host[1] to capture the return value after soft kernel returns; b) we can put the return value to last 4 bytes of the BO (which is bytes 4092~4095) and limits the reg file to 4088
2) I don't see this return value is populated back to host. Are we going to do it? Or otherwise, how should user check the return value of softkernel?

src/runtime_src/core/edge/skd/sk_daemon.cpp
sarab96: changing it to save 32 bit int in regamp[1]
Host side change will be in separate PR

sarab96:
Agree.
changed it to save 32 bit int in regmap[1]
Host side change will be in separate PR

> Two comments
> 
> 1. We should use 32 bits instead of 8 bits to carry the soft kernel return code. a) We can reuse the args_from_host[1] to capture the return value after soft kernel returns; b) we can put the return value to last 4 bytes of the BO (which is bytes 4092~4095) and limits the reg file to 4088
> 2. I don't see this return value is populated back to host. Are we going to do it? Or otherwise, how should user check the return value of softkernel?

larry9523 (approved):

sarab96:
retest this please

dayeh-xilinx:
retest this please

dayeh-xilinx:
retest this please

PR #4480, opened 2020-11-18
Add support for cmake's find_package on linux

wilderfield:
The purpose of this PR is to enable consumer's of XRT who build their projects with cmake to use find_package(XRT) to import XRT targets into their project.

"Exporting a target allows reusing it in other CMake projects, just as if it were defined in the current project."

At the moment this is only supported on native linux builds. Embedded, and Windows will take more work.

To test this:

1) Modify tests/xrt/CMakeLists.txt
       add the following around line 14:
          find_package(XRT)

2) Modify tests/xrt/56_xclbin/CMakeLists.txt
       instead of `xrt_coreutil_LIBRARY` use `XRT_COREUTIL_LIBRARIES`

Build the tests, and run as normal.

Note, you can perform step 2) on any of your preferred tests.

The find_package(XRT) command will make the following variables available to consumers:
  
Input:
```
+find_package(XRT)
+
+message(${XRT_FOUND})
+message(${XRT_CMAKE_DIR})
+message(${XRT_INCLUDE_DIRS})
+message(${XRT_CORE_LIBRARIES})
+message(${XRT_COREUTIL_LIBRARIES})
+message(${XRT_OPENCL_LIBRARIES})
+message(${XRT_SWEMU_LIBRARIES})
+message(${XRT_HWEMU_LIBRARIES})
```
  
Output:
```
1                                                             
/home/bryanloz/DEEPXILINX/LXRT/XRT/build/Release/opt/xilinx/xrt/share/cmake/XRT
/home/bryanloz/DEEPXILINX/LXRT/XRT/build/Release/opt/xilinx/xrt/include        
XRT::xrt_coreXRT::xrt_coreutil                                                 
XRT::xrt_coreutil                                                              
XRT::xilinxopenclXRT::xrt++                                                    
XRT::xrt_swemu                                                                 
XRT::xrt_hwemu  
```

PR #4499, opened 2020-11-25
Add support for cmake's find_package on linux

stsoe:
Amend  #4480: Create relocatable export files

Do not use CMAKE_INSTALL_PREFIX as in install targets.  This variable
is an absolute path and makes XRT export paths non-relocatable.


stsoe:
Use original repo:branch for updates.

PR #4509, opened 2020-11-30
XRT-717 QSFP phase2: Add support of taking input data via sysfs node.

xdavidz:
This is the effort for enabling write into qsfp control nodes.
Spec is here: https://confluence.xilinx.com/display/DCG/Alveo+QSFP+Module+Management+and+Support

Test Results:
```
T1: read io_config*\
cp /sys/bus/pci/devices/0000\:5e\:00.0/xmc.m.18874373/qsfp0_io_config a
hexdump -n 1 a

T2: write io_config
head -c 4 a > b
cat b > /sys/bus/pci/devices/0000\:5e\:00.0/xmc.m.18874373/qsfp0_io_config

Tested sudo permission works as well.
/usr/bin/sudo cp /sys/bus/pci/devices/0000\:5e\:00.0/xmc.m.18874371/qsfp0_io_config a
/usr/bin/sudo hexdump a
0000000 0003



```

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
@@ -338,7 +338,6 @@ struct xmc_pkt_hdr {
 
 #define CMC_OP_QSFP_DIAG_OFFSET 	0x14
 #define CMC_OP_QSFP_IO_OFFSET           0x8
-#define CMC_MAX_QSFP_READ_SIZE          128
 
 #define BDINFO_MAC_DYNAMIC              0x4B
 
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
```diff
@@ -2577,28 +2600,66 @@ QSFP_BIN_ATTR(3);
...
+		dev_get_drvdata(container_of(kobj, struct device, kobj));       \
+	return xmc_qsfp_write(xmc, PORT, buffer, off, count);                   \
+}
+
+QSFP_CONTROL(0)
+QSFP_CONTROL(1)
+QSFP_CONTROL(2)
+QSFP_CONTROL(3)
+
+static BIN_ATTR_WO(qsfp0_control, 0);
```
maxzhen, line 2619: is sysfs nodes created by BIN_ATTR_WO() allows sudo access or do you have to su to root to write to it?

xdavidz: I suppose those are defined by the linux API, https://elixir.bootlin.com/linux/latest/source/include/linux/sysfs.h#L227
aka. RW as 0644, RO as 0444, WO as 02000. 

If we should customize the permission. We will need to not use BIN_ATTR_WO macro, instead we can define the bin_attribute directly as the same as cmc_image_write

maxzhen: 02000 means that only root can write to it. but i'd think that we would like to allow sudoers to write to it, right? so we can't use BIN_ATTR_WO here?

houlz0507 (approved):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
xdavidz: adjust RW to 0600 which is 2(W)+4(R)
adjust W to 0200 which is 2(W)

verified sudoers works.

maxzhen (approved):

PR #4511, opened 2020-11-30
soft kernel in XRT native-1

sarab96:
Step-1: For adding soft kernel in native XRT
Adding arg setter for soft kernels

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -959,7 +959,7 @@ class kernel_impl
     // assert ( ips.size() >= 1);
     auto ctrl = IP_CONTROL((ips[0]->properties & IP_CONTROL_MASK) >> IP_CONTROL_SHIFT);
     for (size_t idx = 1; idx < ips.size(); ++idx)
-      if (IP_CONTROL((ips[0]->properties & IP_CONTROL_MASK) >> IP_CONTROL_SHIFT) != ctrl)
+      if (IP_CONTROL((ips[idx]->properties & IP_CONTROL_MASK) >> IP_CONTROL_SHIFT) != ctrl)
         throw std::runtime_error("CU control protocol mismatch");
 
     return ctrl;
@@ -1214,6 +1214,24 @@ class run_impl
       desc_entry->arg_size = arg.size();
       auto count = std::min(arg.size() / sizeof(uint32_t), value.size());
       std::copy_n(value.begin(), count, desc_entry->arg_value);
+    }
+  };
+
+  //Soft Kernel Arg setter
+  struct sk_arg_setter : arg_setter
+  {
+    sk_arg_setter(uint32_t* data)
+      : arg_setter(data)
+    {}
+
+    //Soft kernel regmap size and arguments are unknown; xclbin doesn't have this info
+    virtual void
+    set_arg_value(const uint8_t* args, const uint32_t size) //sizein bytes from 0x0
+    {
+      if (size > 4092) {//From user max is 4092; First 32 bits are exclusive for SKD
+        throw xrt_core::error(-EINVAL, "Max soft kernel regamp size is 4092 bytes");
+      }
+      std::memcpy((uint8_t*)data, args, size);
     }
   };
 
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -852,6 +852,10 @@ class kernel_impl
   size_t fa_output_entry_bytes = 0;    // Fast adapter output desc bytes
   size_t num_cumasks = 1;              // Required number of command cu masks
   uint32_t protocol = 0;               // Default opcode
+  //TODO For soft kernel; kernel type variable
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1217,6 +1217,24 @@ class run_impl
     }
   };
 
+  //Soft Kernel Arg setter
+  struct sk_arg_setter : arg_setter
```
stsoe, line 1221: Does not appear to be used.  Please complete this suggested change.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1217,6 +1217,24 @@ class run_impl
...
+  {
+    sk_arg_setter(uint32_t* data)
+      : arg_setter(data)
+    {}
+
+    //Soft kernel regmap size and arguments are unknown; xclbin doesn't have this info
+    virtual void
+    set_arg_value(const uint8_t* args, const uint32_t size) //sizein bytes from 0x0
```
stsoe, line 1229: How and where will this function be called, it doesn't match the signature of set_arg_value?

stsoe (changes requested):

src/runtime_src/core/common/api/xrt_kernel.cpp
sarab96: Added more info & code.

src/runtime_src/core/common/api/xrt_kernel.cpp
sarab96: See more info & code. For soft kernels args vector will be of size zero. For running kernel only full regmap setting will be allowed.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1297,6 +1346,14 @@ class run_impl
     arg.set(arg_setter.get(), args);
   }
 
+  void
+  set_sk_args(const uint8_t* args, const uint32_t size)
+  {
+    //TODO This is allowed only for soft kernels
+    arg_setter->set_sk_args(args, size);
+  }
```
stsoe, line 1354: Who will call this and how?  Soft kernels will not be a concept exposed at the API level so I am curious how this function will end being called?

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -852,6 +852,10 @@ class kernel_impl
   size_t fa_output_entry_bytes = 0;    // Fast adapter output desc bytes
   size_t num_cumasks = 1;              // Required number of command cu masks
   uint32_t protocol = 0;               // Default opcode
+  //TODO For soft kernel; kernel type variable
+  // OR use new protocol type for soft kernels
+  //bool is_soft_kernel;
+  // args will be zero size vector for soft kernel
```
stsoe, line 858: Remove.   This has be encoded in protocol.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1046,6 +1050,10 @@ class kernel_impl
 
     // set kernel protocol
     protocol = get_ip_control(ips);
+    //TODO for soft kernel
+    //either based on protocol type or kernel type variable
+    //For soft kernel args vector is of size zero
+    //Only allow setting of full regmap
```
stsoe, line 1056: Remove

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1180,6 +1188,9 @@ class run_impl
 
     virtual void
     set_arg_value(const argument& arg, const arg_range<uint32_t>& value) = 0;
+
+    virtual void
+    set_sk_args(const uint8_t* args, const uint32_t size) = 0;
```
stsoe, line 1193: I don't think this is necessary.  The current set_arg_value should do just fine.  If not it should be set_arg_value with a arg_range instantiated for uint8.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1189,6 +1200,11 @@ class run_impl
       : arg_setter(data)
     {}
 
+    virtual void
+    set_sk_args(const uint8_t* args, const uint32_t size)
+    {
+      throw xrt_core::error("Invalid arguments. Reserved for soft kernels");
+    }
```
stsoe, line 1207: If the base class has multiple virtual functions some of which are not supported by all derived classes, then don't make the pure virtual, instead default their implementation (throw) in the base class.   However I don't think a new function is needed for soft kernels (see other comments)

stsoe (changes requested):
I don't think it is necessary for the sk_arg_setter to have different API compared to base class.  Nor should set_sk_args be necessary.    While sk_arg_setter will be needed, I think maybe you can provide some more context, and I can help integrate support for soft kernels without requiring changes beyond adding the setter.

sarab96:
soft kernel plan changed to make them similar to pl kernels in xclbin
Hence closing this PR

PR #4540, opened 2020-12-08
soft kernel return code not available with old zocl

sarab96:
If PS has old zocl then ensure cmd completion and don't expect return_code


larry9523 (commented):
Looks good. A couple of nits
1) Please try to use C89 comments style other than C99 in Linux kernel code whenever possible
2) It is better to replace 'old zocl' by 'shells that can not report command state in CQ'

sarab96:
Updated
> Looks good. A couple of nits
> 
> 1. Please try to use C89 comments style other than C99 in Linux kernel code whenever possible
> 2. It is better to replace 'old zocl' by 'shells that can not report command state in CQ'

larry9523 (approved):

PR #4545, opened 2020-12-09
CR-1084965 xbmgmt flash --factory_reset report:Failed to install bitstream guard: -22

maxzhen:
Some of the old QSPI IP on old boards does not support the QSPI_CMD_EXT_QUAD_WRITE command. We'll have to fall back to use QSPI_CMD_QUAD_WRITE for them.


maxzhen:
tested on u50/u250/u.2

cellery (approved):

PR #4534, opened 2020-12-07
Get xocl driver compiling on Linux 5.9

sonals:

PR #4555, opened 2020-12-12
Fix zocl build for Linux kernel 5.9

matwey:


maxzhen:
ok to test

larry9523 (approved):
Thanks for the patch. It looks good.
The only nit I have is it might be better if we are consistent on
LINUX_VERSION_CODE >= KERNEL_VERSION(5.9.0) and KERNEL_VERSION(5.9.0) >= KERNEL_VERSION(5.9.0)
I prefer the former.  ( I know we are not yet consistent across all zocl code :))

matwey:
I see that there is already fix for this issue in: #4534

larry9523:
> I see that there is already fix for this issue in: #4534

that is for xocl only, isn't it?

matwey:
Ah, you are right... I will account your comments tomorrow.

mclayton1000:
Hi @matwey  , hi @maxzhen , hi @larry9523 ,
as in #4534 , the patch should check for Linux >= 5.7, because those changes were introduced from kernel 5.7.
```
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,7,0)
+       .gem_free_object_unlocked       = xocl_free_object,
+#else
        .gem_free_object                = xocl_free_object,
+#endif
```
and:
```
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,7,0)
+       #define XOCL_DRM_GEM_OBJECT_PUT_UNLOCKED drm_gem_object_put
+       #define XOCL_DRM_GEM_OBJECT_GET drm_gem_object_get
+#elif LINUX_VERSION_CODE >= KERNEL_VERSION(4,12,0)
```
Best regards.

PR #4556, opened 2020-12-14
Use PETALINUX_VER variable

rbramand-xilinx:
> using PETALINUX_VER variable instead of hardcoded path


chvamshi-xilinx (approved):
Looks good.

PR #4559, opened 2020-12-14
Fix missed std:: for streampos

matwey:
`streampos` is a member of `std::` and that should be properly specified in order
to comply the standard.


maxzhen:
ok to test

AShivangi (approved):

stsoe (changes requested):
Let be more modern and use `auto` here; the type really is pretty clear from tellg().

PR #4562, opened 2020-12-15
CR-1081174 increase tx timeout to 6 sec for hot reset mailbox command

rajkumar-xilinx:
Increased transfer timmeout value to 6 sec for Hot Reset request message in mailbox driver.
It is observed this request is getting timed out in winodws host <-> linux vm case.
There are couple of reasons like wireserver in windows & udev events received by mpd,
causing this timeout issue.
Right before the reset mailbox msg is sent, xocl driver does an offline/online of the mailbox subdev.
This triggers async udev event in mpd. Then when the reset mailbox msg is sent by xocl driver,
the online event may not have been handled yet (the destroyed thread has not been recreated yet),
so nobody will handle the msg – that causes the tx timeout.

Signed-off-by: Brian Xu <brianx@xilinx.com>

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
@@ -137,7 +137,7 @@ static void xocl_mig_cache_read_from_peer(struct xocl_dev *xdev)
 	memcpy(mb_req->data, &subdev_peer, data_len);
 
 	ret = xocl_peer_request(xdev,
-		mb_req, reqlen, mig_ecc, &resp_len, NULL, NULL, 0);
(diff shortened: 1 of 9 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
```diff
@@ -364,14 +364,14 @@ int xocl_hot_reset(struct xocl_dev *xdev, u32 flag)
 			return 0;
 
 		xocl_peer_request(xdev, &mbreq, sizeof(struct xcl_mailbox_req),
-			&ret, &resplen, NULL, NULL, 0);
+			&ret, &resplen, NULL, NULL, 0, 6);
```
rajkumar-xilinx, line 367: Increasing timeout to 6 seconds in hot reset case

xuhz (approved):

PR #4571, opened 2020-12-17
CR#1085436: buffer size check

sarab96:
soft kernel should not crash for buffer size smaller than 4096 Bytes

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/Makefile
+++ b/src/runtime_src/core/edge/drm/zocl/Makefile
@@ -22,7 +22,7 @@ endif
 
 endif
 
-ccflags-y := -I$(src)/include -I$(src)/../../include -I$(src)/../../../include -I$(src)/../../../common/drv/include -I/scratch/sarab/xrt_git/build/aarch64/build/tmp/work-shared/zynqmp-generic/kernel-source/include -DXRT_HASH="\"$(GIT_HASH)\"" -DXRT_HASH_DATE="\"$(GIT_HASH_DATE)\"" -DXRT_BRANCH="\"$(GIT_BRANCH)\"" -DXRT_MODIFIED_FILES="\"$(GIT_MODIFIED_FILES)\"" -DXRT_DATE="\"$(DATE_NOW)\"" -DXRT_DRIVER_VERSION="\"$(XRT_DRIVER_VERSION)\""
+ccflags-y := -I$(src)/include -I$(src)/../../include -I$(src)/../../../include -I$(src)/../../../common/drv/include -DXRT_HASH="\"$(GIT_HASH)\"" -DXRT_HASH_DATE="\"$(GIT_HASH_DATE)\"" -DXRT_BRANCH="\"$(GIT_BRANCH)\"" -DXRT_MODIFIED_FILES="\"$(GIT_MODIFIED_FILES)\"" -DXRT_DATE="\"$(DATE_NOW)\"" -DXRT_DRIVER_VERSION="\"$(XRT_DRIVER_VERSION)\""
 
 #flags passed from xrt_git.bb file are added below
 ifneq ($(cflags_zocl),)
--- a/src/runtime_src/core/edge/drm/zocl/Makefile
+++ b/src/runtime_src/core/edge/drm/zocl/Makefile
@@ -22,7 +22,7 @@ endif
 
 endif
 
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/edge/drm/zocl/Makefile
```diff
@@ -22,7 +22,7 @@ endif
 
 endif
 
-ccflags-y := -I$(src)/include -I$(src)/../../include -I$(src)/../../../include -I$(src)/../../../common/drv/include -DXRT_HASH="\"$(GIT_HASH)\"" -DXRT_HASH_DATE="\"$(GIT_HASH_DATE)\"" -DXRT_BRANCH="\"$(GIT_BRANCH)\"" -DXRT_MODIFIED_FILES="\"$(GIT_MODIFIED_FILES)\"" -DXRT_DATE="\"$(DATE_NOW)\"" -DXRT_DRIVER_VERSION="\"$(XRT_DRIVER_VERSION)\""
+ccflags-y := -I$(src)/include -I$(src)/../../include -I$(src)/../../../include -I$(src)/../../../common/drv/include -I/scratch/sarab/xrt_git/build/aarch64/build/tmp/work-shared/zynqmp-generic/kernel-source/include -DXRT_HASH="\"$(GIT_HASH)\"" -DXRT_HASH_DATE="\"$(GIT_HASH_DATE)\"" -DXRT_BRANCH="\"$(GIT_BRANCH)\"" -DXRT_MODIFIED_FILES="\"$(GIT_MODIFIED_FILES)\"" -DXRT_DATE="\"$(DATE_NOW)\"" -DXRT_DRIVER_VERSION="\"$(XRT_DRIVER_VERSION)\""
```
stsoe, line 25: oops, don't think you mean to add this include search path.

sarab96: :) updated

maxzhen (commented):
what if buffer size > 4k, but not aligned at 4k boundary or not multiple of 4k? trying to understand if size < 4k is good enough...

mamin506 (approved):

larry9523:
> what if buffer size > 4k, but not aligned at 4k boundary or not multiple of 4k? trying to understand if size < 4k is good enough...

I believe the requirement is page aligned. 
https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/drm_gem.c#L153

sarab96:
Updated for alignment check

> > what if buffer size > 4k, but not aligned at 4k boundary or not multiple of 4k? trying to understand if size < 4k is good enough...
> 
> I believe the requirement is page aligned.
> https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/drm_gem.c#L153

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -808,6 +808,11 @@ int zocl_get_hbo_ioctl(struct drm_device *dev, void *data,
 		DRM_ERROR("Buffer at out side of reserved memory region\n");
 		return -ENOMEM;
 	}
+	if ((args->size & ZOCL_PAGE_SIZE_M1) != 0) {
+		//DRM requirement
+		DRM_ERROR("Buffer size must be multiple of 4096\n");
+		return -ENOMEM;
+	}
```
larry9523, line 815: 1) I would suggest using system macro instead of define our own. PAGE_ALIGEND might be your best candidate
2) We probably want to check the start address and size are both PAGE_ALIGEND and also the size should not be zero, if it is not checked somewhere else.
3) The return error code should be -EINVAL not -ENOMEM

larry9523 (changes requested):

larry9523 (changes requested):
See comments

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -808,6 +808,11 @@ int zocl_get_hbo_ioctl(struct drm_device *dev, void *data,
 		DRM_ERROR("Buffer at out side of reserved memory region\n");
 		return -ENOMEM;
 	}
+	if ((args->size & ZOCL_PAGE_SIZE_M1) != 0) {
+		//DRM requirement
```
larry9523, line 812: use /* */ please

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
sarab96: done

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
sarab96: updated

larry9523 (approved):

maxzhen (approved):

PR #4572, opened 2020-12-17
CR#1083100: dma check for size zero

sarab96:
dma check for size zero


maxzhen (commented):
what will happen if driver sees zero size IO request? if it'll cause driver to hang/crash/etc, then we have to fix this in driver. Otherwise, a non-privileged user can send DMA request directly down to driver without going through xclSyncBO() and will harm the whole system. if driver can handle this gracefully, then there is no need to fix anything in xclSyncBO(), right?

sarab96:
xclSyncBO gives error. error code -5 is returned. So I think xclSyncBO level change should be fine.
> what will happen if driver sees zero size IO request? if it'll cause driver to hang/crash/etc, then we have to fix this in driver. Otherwise, a non-privileged user can send DMA request directly down to driver without going through xclSyncBO() and will harm the whole system. if driver can handle this gracefully, then there is no need to fix anything in xclSyncBO(), right?

maxzhen (approved):

PR #4573, opened 2020-12-17
refine iops test print result

mamin506:
1. Remove empty line
2. Format "Maximum IOPS: %d"


maxzhen:
can we change iops in Format "Maximum iops: %d" to be upper case?

maxzhen (approved):

PR #4574, opened 2020-12-17
cr-1085563 ASTeR TC1.05 - xbutil reset hangs when ran on u50c recover…

houlz0507:
…y image


xdavidz (approved):

maxzhen:
should we consider to not allow xocl to attach on user pf on recovery image shell? so that user will not even try to hot reset it, validate it, and we don't get to see all related CRs any more....

PR #4580, opened 2020-12-20
signal handler before child

sarab96:
zombie process seen on PS after soft kernel errors
signal handler is needed before creation of soft kernel child process

### DIFF
--- a/src/runtime_src/core/edge/skd/sk_daemon.cpp
+++ b/src/runtime_src/core/edge/skd/sk_daemon.cpp
@@ -338,9 +338,8 @@ void configSoftKernel(xclSKCmd *cmd)
      * kernel image.
      */
     pid = fork();
-    if (pid > 0) {
+    if (pid > 0)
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/edge/skd/sk_daemon.cpp
```diff
@@ -338,6 +338,10 @@ void configSoftKernel(xclSKCmd *cmd)
      * kernel image.
      */
     pid = fork();
+    if (pid > 0) {
```
larry9523, line 341: I think the rule is no "{" "}" for one line statement.

larry9523 (changes requested):
Looks good. Just minor nit

src/runtime_src/core/edge/skd/sk_daemon.cpp
sarab96: done

larry9523 (approved):

PR #4584, opened 2020-12-23
xbutil to check if soft kernel process is alive

sarab96:
xbutil cu_stat cmd to check if soft kernel process is alive

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/sched_exec.c
+++ b/src/runtime_src/core/edge/drm/zocl/sched_exec.c
@@ -1119,6 +1119,7 @@ cu_stat(struct sched_cmd *cmd)
 	}
 
 	/* indevidual SK CU status */
+	mutex_lock(&sk->sk_lock);
 	for (i = 0; i < sk->sk_ncus && pkt_idx < max_idx; ++i) {
 		p_tmp = NULL;
 		if (sk->sk_cu[i]) {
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/edge/drm/zocl/sched_exec.c
```diff
@@ -1119,10 +1120,16 @@ cu_stat(struct sched_cmd *cmd)
...
 	/* indevidual SK CU status */
 	for (i = 0; i < sk->sk_ncus && pkt_idx < max_idx; ++i) {
-		if (sk->sk_cu[i])
-			pkg->data[pkt_idx++] = (exec->scu_status[cu_mask_idx(i)] &
-				(1 << (i % sizeof(exec->scu_status[0])))) ? 1 : 0;
-		else
+		p_tmp = NULL;
+		if (sk->sk_cu[i]) {
+			/* xbutil to check if process is alive */
+			p_tmp = find_get_pid(sk->sk_cu[i]->sc_pid);
```
larry9523, line 1126: I think this could be a temporary solution for now. Let's revisit this in the next couple of weeks. I think a more robust solution should be introducing softkernel context in zocl. Once the context is gone, sk_cu[i] should be NULL.

However, I do have some concern here, since we are not holding sk_lock, do we have a crash window that sk_cu[i] was just released after line 1124?

larry9523 (changes requested):
Seems there is a race condition that could probably cause kernel crash

larry9523 (approved):

src/runtime_src/core/edge/drm/zocl/sched_exec.c
sarab96: Updated for sk_lock. Checking process alive is still better as xbutil query can be used anytime even after end of an application. context may have dependency on a running application

larry9523: I mean softkernel process context on device side but not the host side application.

PR #4586, opened 2020-12-28
Get XRT drivers 'compiling' on 5.10 kernel. Note that P2P requires mo…

sonals:
…re work to become functional on 5.10


houlz0507 (approved):

chienwei-lan (approved):

Himanshu-xilinx (approved):
looks good

PR #4591, opened 2021-01-05
50% performance gain by switching to memcpy in MicroBlaze

chienwei-lan:
calling write_reg in configure_cu => perf_iops:33k
switching to memcpy in the same function => perf_iops:49k

### DIFF
--- a/src/runtime_src/ert/scheduler/scheduler.cpp
+++ b/src/runtime_src/ert/scheduler/scheduler.cpp
@@ -1225,7 +1225,6 @@ scheduler_loop()
           continue; // CU is not used
 
         auto cuvalue = read_reg(cu_idx_to_addr(slot.cu_idx));
-        //DMSGF("cuidx %d slot_idx %d, cuvalue(0x%x)\r\n",slot.cu_idx,slot_idx,cuvalue);
         if (!(cuvalue & (AP_DONE|AP_IDLE)))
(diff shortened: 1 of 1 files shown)


mamin506 (approved):

src/runtime_src/ert/scheduler/scheduler.cpp
```diff
@@ -623,8 +633,10 @@ configure_cu(addr_type cu_addr, addr_type regmap_addr, size_type regmap_size)
 {
   // write register map, starting at base + 0x10
   // 0x4, 0x8, 0xc used for interrupt, which is initialized in setup
-  for (size_type idx = 4; idx < regmap_size; ++idx)
-    write_reg(cu_addr + (idx << 2), read_reg(regmap_addr + (idx << 2)));
+  uint32_t *addr_ptr = (uint32_t *)(uintptr_t)cu_addr;
+  uint32_t *regmap_ptr = (uint32_t *)(uintptr_t)regmap_addr;
+
+  memcpy(addr_ptr+4, regmap_ptr+4, regmap_size);
```
maxzhen, line 639: @chienwei-lan , will this break hardware emulation? write_reg() is emulated, but memcpy is not. I'm wondering if we should #ifdef ERT_HW_EMU and keep emulation code as is?

PR #4593, opened 2021-01-05
Solve ulite driver crash while unload xclmgmt driver or reset

chienwei-lan:


xuhz (commented):
I am not sure whether the change can fix the issue.
When I saw the crash, I am sure the console was opened just once.

a quick look at the assembly code, seems it crashed in getting ops of the port -- there is null reference.

a quick look at the src, in _remove(),
{
        pdata->console_opened = 0;
	ret = uart_remove_one_port(pdata->xcl_ulite_driver, port);
}

and in ulite_thread()
{
       while (pdata->console_opened) {
		ulite_worker(port);
}
there is no lock protection. potential race condition exists.

xuhz:
I made the following comments hrs ago. But for some reason, it was not shown.

   1. if nobody ever opens a console, pdata->thread would be null, i don't know what will happen to kthread_stop() with a null arg
    2. this kthread_stop() may not cause the thread exit because the thread loop is not checking
    like
    while (!kthread_should_stop()) {
    }
    it checks the console_opened, but that is not updated yet. if you want to use console_opened to check without lock, it should be an atomic type

you move the 'console_opened = 0' before you try to stop the thread. I believe an atomic variable should be used.
and also add while (!kthread_should_stop()) would be better. pls refer to other kernel thread code in xrt
 
    3. BTW, when is the memory of pdata freed? memory leak?

xuhz (approved):

PR #4595, opened 2021-01-05
qdma streamming memory leak

houlz0507:


keryell (commented):
I guess you are not adding a memory leak but are trying to remove one?
If so, the commit messages could be something like `Fix a QDMA streamming memory leak`?

PR #4602, opened 2021-01-07
Don't pass object by value

stsoe:
Very minor tweak


stsoe:
retest this please  -- not reproducible when running on xcosda83

PR #4605, opened 2021-01-08
CR-1081174 increase tx timeout to 6 sec for hot reset mailbox command

rajkumar-xilinx:
Increased transfer timmeout value to 6 sec for Hot Reset request message in mailbox driver.
It is observed this request is getting timed out in winodws host <-> linux vm case.
There are couple of reasons like wireserver in windows & udev events received by mpd,
causing this timeout issue.
Right before the reset mailbox msg is sent, xocl driver does an offline/online of the mailbox subdev.
This triggers async udev event in mpd. Then when the reset mailbox msg is sent by xocl driver,
the online event may not have been handled yet (the destroyed thread has not been recreated yet),
so nobody will handle the msg – that causes the tx timeout.

Signed-off-by: Brian Xu brianx@xilinx.com


rajkumar-xilinx:
Dupe of https://github.com/Xilinx/XRT/pull/4606

PR #4622, opened 2021-01-12
CR-1085740 U30 cards bricked after running restore to golden

AShivangi:
Output:
```
# xbmgmt flash --factory_reset --card 0000:66:00.0
Factory reset is not currently supported on U30.

#
```


maxzhen (approved):
After we remove this hack, we'll enable factory_reset for all u30 board again, including this version of U30. Do we know if this is OK with platform team?

AShivangi:
I checked with Brian and he said that this is ok

xuhz (approved):

PR #4632, opened 2021-01-15
Revert memcpy in ERT ssv2

chienwei-lan:


src/runtime_src/ert/scheduler/scheduler.cpp
```diff
@@ -1229,7 +1222,7 @@ scheduler_loop()
           continue; // CU is not used
 
         auto cuvalue = read_reg(cu_idx_to_addr(slot.cu_idx));
-        if (!(cuvalue & (AP_DONE|AP_IDLE)))
```
maxzhen, line 1232: not sure why revert this line?

chienwei-lan: It's a bug I introduced in previous ssv2 data flow PR 

if we don't remove AP_IDLE and we somehow do not kick start CU. 

Once ERT check it's AP_IDLE, it will notify host as a false CU completion

PR #4633, opened 2021-01-19
CR-1086697 : Added the new kernel version check for xbutil/xbutil2 tool

saifuddin-xilinx:
CR-1086697 : xbutil scan complains that CentOS 7.5 standard kernel is not supported
 -- Added the new kernel version check


saifuddin-xilinx:
Resolved the conflicts.

chvamshi-xilinx (approved):
This is ok this time as this is for U30, but we cannot add supported kernel versions without full DSV verification.

maxzhen (approved):

PR #4644, opened 2021-01-21
CR-1084638

AShivangi:
**Final fix**: The error still shows up for another device that is being reset and not being queried, but it is directed to stderr and can be ignored.


maxzhen:
this is xbutil2 change, not impacting u30.

maxzhen (approved):

PR #4646, opened 2021-01-21
Potential issue of handling soft cu completion

larry9523:
While we are reusing cu_mask field in ert_start_kernel_cmd, we copy both header field and cu_mask (return_code) back to command queue. Since the state field in ert command header is the only flag we check to see if a new command is coming in, this copy will make ert (zocl) thought there is a new command and start to handle it. This could cause serious problem

1) If return_code is zero (cu_mask for the extra command), the extra command may not be executed because we can not find valid CU. But it will sent an extra interrupt to host in the cq slot of previous valid command. This could be a race if host just sent a new command to that slot and will get this false interrupt.
2) If return_code is not zero, the extra command could be falsely executed and may corrupt device memory
3) It will double the command number in ert(zocl).

To fix it, we just need to copy the return_code (cu_mask) but not the header.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/sched_exec.c
+++ b/src/runtime_src/core/edge/drm/zocl/sched_exec.c
@@ -2,7 +2,7 @@
 /*
  * A GEM style device manager for MPSoC based OpenCL accelerators.
  *
- * Copyright (C) 2017-2020 Xilinx, Inc. All rights reserved.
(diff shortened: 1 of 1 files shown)


sarab96:
This PR change is incorrect. We need to discuss this.

larry9523:
> This PR change is incorrect. We need to discuss this.

Yes, please let me know what is incorrect.

larry9523:
Close for now. Will come back with a more completed fix.

PR #4648, opened 2021-01-21
CR-1086282 system crash when run reset test for xilinx_u250_gen3x16_xdma_shell_2_1

mamin506:
Before this change, the KDMA resource would be mapped in iores and cu sub-devices.
No one really access KDMA through the ops that iores sub-device provide. The idea is to remove KDMA in iores.

But feature_rom sub-device would need to know the physical address of kdma. It get kdma resource via xocl_subdev_get_resource(), which would search resource in other sub-device and return the details. Then this make feature_rom depends on iores sub-device.
I change the implementation here to directly get kdma resource from dtb blob.

In NoDMA shell, there is no kdma CU. The m2m sub-device is the owner of kdma resource.

Pranjal has verified this on his machine, no more resource overlap error and m2mtest pass as well.


dayeh-xilinx:
retest this please

src/runtime_src/core/pcie/driver/linux/xocl/subdev/feature_rom.c
```diff
@@ -578,23 +578,24 @@ static int get_header_from_peer(struct feature_rom *rom)
...
 	memcpy(&rom->header, header, sizeof(*header));
 
-	xocl_xdev_info(xdev, "Searching CMC in dtb.");
-	ret = xocl_subdev_get_resource(xdev, RESNAME_KDMA,
-			IORESOURCE_MEM, &res);
-	if (!ret) {
+	xocl_xdev_info(xdev, "Searching CDMA in dtb.");
+	offset = xocl_fdt_path_offset(xdev, XDEV(xdev)->fdt_blob, "/" NODE_ENDPOINTS "/" RESNAME_KDMA);
+	if (offset) {
+		io_off = xocl_fdt_getprop(xdev, XDEV(xdev)->fdt_blob, offset, PROP_IO_OFFSET, NULL);
```
pranjalv-xilinx, line 593: Minor Nit ....might be good to add error check for return value of xocl_fdt_getprop() just incase this node does not have IO offset property.

dayeh-xilinx:
retest this please

src/runtime_src/core/pcie/driver/linux/xocl/subdev/feature_rom.c
mamin506: Done.

houlz0507 (approved):

pranjalv-xilinx (approved):

PR #4661, opened 2021-01-27
Doc update

imrickysu:
- OpenCL requires C++11
- Also updated embedded platform XRT requirements


src/runtime_src/doc/toc/system_requirements.rst
```diff
@@ -16,7 +16,10 @@ Supported Xilinx® Accelerator Cards are listed in :ref:`platforms.rst`.
 XRT Software Stack for PCIe Accelerator Cards
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-XRT and OpenCL runtime require Linux kernel 3.10+ and GCC with C++14 features. 
+XRT software stack requires Linux kernel 3.10+. 
+Compiling host code with XRT OpenCL API requires GCC with C++11 features; Compiling host code with XRT native API requires GCC with C++14 features.
```
stsoe, line 20: There really is no requirement when compiling OpenCL applications.  For example host code may be C code and `-std=c++zz` would make no sense.   The `-std=c++zz` is a requirement only if host code is C++ and is using features from c++zz.  For OpenCL we simply shouldn't mention it.

When using native C++ APIs, host code is including XRT header files that use c++14 features, so the statement to that effect is correct.

stsoe (changes requested):

keryell:
This PR is unclear.
Does this mean that to use XRT software we need to use C++11 (10 year old) or C++14 (7 year old) and anything more modern is forbidden?

imrickysu:
> 
> 
> This PR is unclear.
> Does this mean that to use XRT software we need to use C++11 (10 year old) or C++14 (7 year old) and anything more modern is forbidden?

Hi @keryell ,

It says it requires C++14 feature. This phrase was used in the original doc. It should imply c++14 or newer c++ std should be used. 

Do you suggest we say it explicitly like "Compiling host code with XRT native C++ API requires GCC standard with -std=c++14 or newer."?

imrickysu:
Removed c++ std dialect requirement in system requirement page because the GCC version on the supported OS can always be supported. The API page describes the detailed dialect std requirements.

stsoe:
Sure, but you cannot compile C++ host code that use specific C++ features without g++ without specifying a -std flag.  For native C++ APIs the minimum is -std=c++14, it of course doesn't mean you can't use -std=c++20 if compiler supports that.

keryell:
> Do you suggest we say it explicitly like "Compiling host code with XRT native C++ API requires GCC standard with -std=c++14 or newer."?

Yes, it seems clearer and more positive.
Also I am unsure it requires GCC. I hope another compiler could work.

imrickysu:
Added API compile requirement to API pages: 

- XRT Native: Compiling host code with XRT native C++ API requires C++ standard with -std=c++14 or newer. On GCC version older than 4.9.0, please use -std=c++1y instead because -std=c++14 is introduced to GCC from 4.9.0.
- OpenCL: Please follow the general OpenCL guidelines to compile host applications that uses XRT OpenCL API with Xilinx OpenCL extensions. Xilinx OpenCL extension doesn't require additional compiler features. Normally C99 or C++11 would be good.

PR #4625, opened 2021-01-13
Windows flashing

AShivangi:
Tested flashing flows:
- 1RP -> golden
- Golden -> 1RP
- 2RP -> 2RP

_NOTE: This change requires windows driver with git hash [9892023db1ce503427e75647a30f761339c7c2a8 ](https://github.com/Xilinx/MS_XCLMGMT/commit/9892023db1ce503427e75647a30f761339c7c2a8)or later (Dec 14, 2020)_
_This version of the windows driver is required since the pcie_info struct has been decoupled from the ioc_device_info struct. Doing so prevents commands run in the golden image from crashing._

PR #4664, opened 2021-01-27
Coverity fix

AShivangi:
introduced by https://github.com/Xilinx/XRT/pull/4625

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/ReportPlatform.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/ReportPlatform.cpp
@@ -111,25 +111,20 @@ ReportPlatform::getPropertyTree20202( const xrt_core::device * device,
   pt_platform.put("hardware.serial_num", info.mSerialNum);
 
   //Flashable partition running on FPGA
-  // the vectors are being populated by empty strings which need to be removed
-  auto remove_empty_strings = [](std::vector<std::string> uuids) {
-    auto it = uuids.begin();
-    while(it != uuids.end()) {
-      if((*it).empty())
-        uuids.erase(it);
-      else
-        it++;
-    }
-  };
-  
+
   std::vector<std::string> logic_uuids, interface_uuids;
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/tools/xbmgmt2/ReportPlatform.cpp
```diff
@@ -111,25 +111,20 @@ ReportPlatform::getPropertyTree20202( const xrt_core::device * device,
...
+    logic_uuids.erase(
+      std::remove_if(logic_uuids.begin(), logic_uuids.end(),	
+                      [](const std::string& s) { return s.empty(); }));
   } catch (...) {}
   try {
     interface_uuids = xrt_core::device_query<xrt_core::query::interface_uuids>(device);
-    remove_empty_strings(interface_uuids);
+    interface_uuids.erase(
+      std::remove_if(interface_uuids.begin(), interface_uuids.end(),	
+                  [](const std::string& s) { return s.empty(); }));
```
stsoe, line 127: This erases one element only and the other moved elements are bogus.  Make sure to add the end() iterator also. See https://en.cppreference.com/w/cpp/container/vector/erase.

std::remove_if returns an iterator to the new logic end of the vector, erase should erase the range [itr, end[ where itr is the new logical end returned by remove_if and end is the end() of the vector itself.
`vec.erase(std::remove_if(b,e,cond), vec.end());`

stsoe (changes requested):

src/runtime_src/core/tools/xbmgmt2/ReportPlatform.cpp
AShivangi: Oh, that is what was missing from my initial implementation. Thanks for the clarification. I couldn't figure out why my logic wasn't working earlier. It was missing the end() :)

stsoe (approved):

AShivangi:
retest this please

maxzhen:
this is xbmgmt2 change. should have no impact on u30.

PR #4671, opened 2021-01-28
CR#1086564: fix for upto 128 soft kernels

sarab96:
fix to support upto 128 CUs


stsoe (commented):
While the code change is safe, I have no clue why this change is made the way it is and why dev_tmp1 doesn't simply store the bitset rather than dividing it into 4 explicitly managed words?

The changed function is much too long and has no comments at all .

sarab96:
@stsoe : bitset<128> overflows in to_ulong call if more than 32 CUs. cu_mask value is used in ert pkt and saved to support upto 128 soft kernels

stsoe (approved):

PR #4673, opened 2021-01-28
Enhance noop emulation to insert completion delay

stsoe:
Minor tweaks based on profiling cost of start() and wait().
Remove xclProbe from xrt iops test case.
Fix xcl iops test case calling unmap.
Ensure both test create same number of commands.

### DIFF
--- a/tests/xrt/perf_IOPS/xcl_api_iops.cpp
+++ b/tests/xrt/perf_IOPS/xcl_api_iops.cpp
@@ -40,18 +40,26 @@ load_file_to_memory(const std::string& fn)
     return bin;
 }
 
-double runTest(xclDeviceHandle handle, std::vector<std::shared_ptr<task_info>>& cmds,
-               unsigned int total)
+int
+startCmd(xclDeviceHandle handle, task_info* cmd)
+{
+  cmd->ecmd->state = ERT_CMD_STATE_NEW;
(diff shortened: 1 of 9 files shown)


stsoe:
@mamin506 Please check iops test case changes.  Do not worry about the other changes, which are mostly to ease profiling.

mamin506 (approved):

tests/xrt/perf_IOPS/xcl_api_iops.cpp
```diff
@@ -40,18 +40,26 @@ load_file_to_memory(const std::string& fn)
     return bin;
 }
 
-double runTest(xclDeviceHandle handle, std::vector<std::shared_ptr<task_info>>& cmds,
-               unsigned int total)
+int
+startCmd(xclDeviceHandle handle, task_info* cmd)
+{
+  cmd->ecmd->state = ERT_CMD_STATE_NEW;
```
mamin506, line 46: KDS would set command state to ERT_CMD_STATE_NEW just before the command added to pending list.

stsoe: Yes, but that is specific to KDS and doesn't work in some flows that don't use KDS.

stsoe:
Replaced by #4676

PR #4680, opened 2021-02-02
Fix zocl fall back to polling mode in case of free running kernels

rbramand-xilinx:
> This PR adds a check to skip zocl falling back to polling mode when free running kernel is encountered

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
@@ -1019,6 +1019,9 @@ zocl_xclbin_cus_support_intr(struct drm_zocl_dev *zdev)
 
 	for (i = 0; i < zdev->ip->m_count; ++i) {
 		ip = &zdev->ip->m_ip_data[i];
+		if(xclbin_protocol(ip->properties) & 0x2) {
+			continue;
+		}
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -1019,6 +1019,9 @@ zocl_xclbin_cus_support_intr(struct drm_zocl_dev *zdev)
 
 	for (i = 0; i < zdev->ip->m_count; ++i) {
 		ip = &zdev->ip->m_ip_data[i];
+		if(xclbin_protocol(ip->properties) & 0x2) {
```
larry9523, line 1022: Suggest to add a space after "if", replace & with "==" and use AP_CTRL_NONE instead of 2
if (xclbin_protocol(ip->properties) == AP_CTRL_NONE) {

larry9523 (changes requested):
Please see comment in line

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
rbramand-xilinx: fixed as per comments

larry9523 (approved):

chvamshi-xilinx (approved):

PR #4681, opened 2021-02-02
show truncated cu status

sarab96:
show truncated cu status when many soft kernels rather than showing no info

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
@@ -4976,12 +4976,11 @@ kds_custat_show(struct device *dev, struct device_attribute *attr, char *buf)
 out:
 	tembuf[sz++] = 0;
 	if (sz >= (ssize_t)PAGE_SIZE) {
-		sz = PAGE_SIZE-50; //save truncated CU status
+		sz = PAGE_SIZE - 30; /* save truncated CU status */
 		truncated = true;
 	}
 	if (truncated) {
 		sz += sprintf(tembuf+sz, "\n.. TRUNCATED ..\n");
-		tembuf[sz++] = 0;
 		tembuf[sz++] = 0;
 	}
 	tembuf[PAGE_SIZE-1] = 0;
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
@@ -4863,8 +4863,10 @@ kds_custat_show(struct device *dev, struct device_attribute *attr, char *buf)
 	struct xocl_ert *xert = exec_is_ert(exec) ? exec->ert : NULL;
 	unsigned int idx = 0;
 	ssize_t sz = 0;
+	ssize_t sz_tmp = 0;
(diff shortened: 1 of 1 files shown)


maxzhen:
@mamin506 , does this change apply to new kds?

src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
```diff
@@ -4940,15 +4967,25 @@ kds_custat_show(struct device *dev, struct device_attribute *attr, char *buf)
...
+		sz = sz_tmp;
+		truncated = true;
+		goto out;
+	}
 
 out:
-	if (sz)
+	tembuf[sz++] = 0;
+	if (sz >= (ssize_t)PAGE_SIZE) {
+		sz = PAGE_SIZE-50; //save truncated CU status
```
larry9523, line 4979: 1. what is the magic number 50 for?
2. please use /* */ instead of //
3. better have a space like sz = PAGE_SIZE - 50

src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
```diff
@@ -4940,15 +4967,25 @@ kds_custat_show(struct device *dev, struct device_attribute *attr, char *buf)
...
+	}
+	if (truncated) {
+		sz += sprintf(tembuf+sz, "\n.. TRUNCATED ..\n");
 		tembuf[sz++] = 0;
-
-	if (sz >= (ssize_t)PAGE_SIZE)
-		sz = 0;
-	else
-		memcpy(buf, tembuf, sz);
+		tembuf[sz++] = 0;
```
larry9523, line 4985: why do we need to 0 here?

larry9523 (changes requested):
Generally the fix looks good.

One thing I'd like to point out is, to support display full PS kernel name, I plan to remove the SCU name in this sysfs node by cache an softkernel one inside icap (just like what PL kernel is doing). So that we will not run into this overflow situation. My suggestion is we can even remove the tembuf we prepared for the  last urgent release once the softkenel sysfs node is in place. So don't really need this truncate stuff.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
sarab96: Updated. 50 is just some space to add "TRUNCATED" to the end. 50 changed to 30.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/mb_scheduler.c
sarab96: updated

sarab96:
As discussed more changes can happen after upcoming changes related to sysfs code

> Generally the fix looks good.
> 
> One thing I'd like to point out is, to support display full PS kernel name, I plan to remove the SCU name in this sysfs node by cache an softkernel one inside icap (just like what PL kernel is doing). So that we will not run into this overflow situation. My suggestion is we can even remove the tembuf we prepared for the last urgent release once the softkenel sysfs node is in place. So don't really need this truncate stuff.

mamin506 (approved):
We will need to add similar structure in new KDS to support the same feature.

larry9523 (approved):

PR #4683, opened 2021-02-02
CR-1086077: Fixing exception thrown at end of execution when memory offload used for trace

jvillarre:
Explicitly unmapping the trace buffer after syncing so the XRT side device does not retain the pointer


IshitaGhosh (approved):

dayeh-xilinx:
retest this please

PR #4684, opened 2021-02-02
Refactoring of OpenCL profiling

jvillarre:
This pull request is a major change to how OpenCL level profiling is implemented, and separates out the profile summary generation/counter information from the trace generation into separate plugins.  The old architecture libraries are no longer built or used, although the code remains as a reference at this point.  This pull request also introduces new xrt.ini switches to independently control each aspect of profiling and deprecates some of the old switches (which are still supported at this time).

As part of this submission, the xocl to xdp side interface was changed, breaking the dependency between the xdp library and the internal xocl structures.  All callbacks to the plugins now transfer C objects and all xocl data structure manipulation is done on the xocl side (in the xocl/api/plugin/xdp directory).  The actual callback points in the xocl library have not changed, although some of the function signatures are different.


src/runtime_src/core/common/device.cpp
```diff
@@ -92,8 +92,7 @@ get_xclbin_uuid() const
   catch (const query::no_such_key&) {
   }
 
-  // Emulation mode likely, just return m_xclbin_uuid which reflects
-  // the uuid of the xclbin loaded by this process.
+  // Emulation mode likely
```
stsoe, line 95: Is this reverting a comment that was recently added.   I'd like to preserve the comment.

src/runtime_src/xdp/profile/database/database.h
```diff
@@ -40,7 +40,7 @@ namespace xdp {
   {
   public:
     // For messages sent to specific plugins
-    enum MessageType { } ;
+    enum MessageType { READ_COUNTERS } ;
```
stsoe, line 43: enum **class** with lower case enumeration values is preferred.

stsoe (commented):
Looks good to me.  Just a request to preserve one comment.

src/runtime_src/core/common/device.cpp
jvillarre: Restored the comment to the same as master in the latest commits.

stsoe (approved):

IshitaGhosh (approved):
The support for OpenCL profiling using new architecture in XDP looks good.

dayeh-xilinx:
retest this please - after failing spooling of results for u30 due to python upgrade

xilinxgitops:
retest this please

PR #4685, opened 2021-02-02
fixed xbutil reset exit too early on Azure

houlz0507:


src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
```diff
@@ -1513,6 +1503,23 @@ int xocl_userpf_probe(struct pci_dev *pdev,
...
+		xocl_err(&pdev->dev, "failed to init sysfs");
+		goto failed;
+	}
+	ret = xocl_init_persist_sysfs(xdev);
+	if (ret) {
+		xocl_err(&pdev->dev, "failed to init persist sysfs");
+		goto failed;
+	}
+
 	return 0;
```
maxzhen, line 1523: it's actually ok for sysfs node to be initialized before device node is created as long as we returns 'offline' through the sysfs until driver is fully initialized. i'm wondering if we should set a flag right before we return 0 here so that the sysfs node can return 'online'?

PR #4694, opened 2021-02-04
For u30 card this validation required

saifuddin-xilinx:
For U30 device while reading 0xdeadfa11 as  0xdeadfa11, which is a invalid value. 
But validation is done only for value "0".
Added 0xdeadfa11 also as a check.


maxzhen (approved):

chvamshi-xilinx (approved):
This may be required for PU1 as well. Please make sure that you cherry pick this change to PU1 aswell.

PR #4701, opened 2021-02-07
Loader

sonals:


sonals:
Hmm I just learned that RHEL 7.X uses bash 4.2 which does not support `-n` named references :-(

So will find another solution.

dayeh-xilinx:
retest this please

sonals:
retest this please

AShivangi (commented):
This looks pretty good!
What happens when the user types "xbmgmt" without any trailing args? Does it print xbmgmt or xbmgmt2 help?

src/runtime_src/core/tools/xbmgmt2/xbmgmt
```diff
@@ -1,29 +1,84 @@
...
+        ["-d"]="xbmgmt2" \
+        ["-r"]="xbmgmt2" \
+        ["-h"]="xbmgmt2" \
+    )
+
+# In the following table "reset" points to a second table by name
+# Note that reset is valid both for classic and next generation tool
+# and hence requires a second lookup to resolve between xbmgmt and xbmgmt2
+declare -A COMMANDS=( \
+        ["help"]="XBMGMT" \
```
AShivangi, line 41: shouldn't this default to xbmgmt2 as well? Maybe we should come up with a better way to differentiate between xbmgmt v1 and v2 help. It might be confusing as to why "--" prints a different help

sonals: The default is still xbmgmt and I was hoping that the current users and scripts do not see any change. So `xbgmt help` should continue to work as before.
`DEFAULT=xbmgmt`

AShivangi: Ok. That sounds good.

src/runtime_src/core/tools/xbmgmt2/xbmgmt
```diff
@@ -1,29 +1,84 @@
...
   "$0.bat" "$@"
   exit $?
 fi
 
+DEFAULT=xbmgmt
+XBMGMT=xbmgmt
+XBMGMT2=xbmgmt2
+
+# Table to resolve reset sub-command based on switch passed to reset
+declare -A RSTOPT=( \
```
rozumx, line 21: I like what you did with the COMMANDS map below where you used the XBMGMT, XBMGM2, XBUTIL, and XBUTIL2 variables instead of the actual values like in RSTOPT, for focuses on the variable's name and not value (like this mapping does).  

While it is highly unlikely that the underlying command name will change in the future, by using the variable's name and not value does make for a more scalable solution.

rozumx (approved):
Overall a simple and effective solution.  Thank you for doing these changes.

**_Side Note_**
There is one coding style inconsistency that was noticed, but it isn't large enough to prevent approving this pull request.

AShivangi (approved):

src/runtime_src/core/tools/xbmgmt2/xbmgmt
sonals: Updated

PR #4707, opened 2021-02-09
Unify the host side code of new kds flow

chienwei-lan:


xdavidz:
@chienwei-lan should you or me refactor the ert_versal code? Given it is almost copied from ert_user but has some versal specific handling.

maxzhen:
> 
> 
> @chienwei-lan should you or me refactor the ert_versal code? Given it is almost copied from ert_user but has some versal specific handling.

can we leave versal one as a standalone subdev? how bad is it? @xdavidz

xdavidz:
> > @chienwei-lan should you or me refactor the ert_versal code? Given it is almost copied from ert_user but has some versal specific handling.
> 
> can we leave versal one as a standalone subdev? how bad is it? @xdavidz

We can leave it alone. The bad thing is that 99% code in it is dup with ert_user.c. I can do the clean up.

chienwei-lan:
> > > @chienwei-lan should you or me refactor the ert_versal code? Given it is almost copied from ert_user but has some versal specific handling.
> > 
> > 
> > can we leave versal one as a standalone subdev? how bad is it? @xdavidz
> 
> We can leave it alone. The bad thing is that 99% code in it is dup with ert_user.c. I can do the clean up.

we can have a discuss about how to integrate ert_versal too

PR #4709, opened 2021-02-10
mpd: restart mpd services when device's serial number not found during mpd init

rajkumar-xilinx:
MPD shall be restarted when it unable to read serial number of one or more
fpga devices during it's init.


chvamshi-xilinx (approved):
Looks good .

rajkumar-xilinx:
https://github.com/Xilinx/XRT/pull/4725

PR #4712, opened 2021-02-11
XRT-848 thread-safe xrt_iops test

mamin506:
The s_cnt had no lock protect. This would be an issue when user application use multiple threads to call xclExecBuf() and share device fd.

But s_cnt is on the critical path, use lock would has negative performance impact. I use percpu variable to fix this issue as Max suggested.

**The IOPS test result** (xsjminm50 + U50 (package 4-3102127)):
$ xbutil scheduler --test xcl_iops --args "-t 1 -l 128 -a 1000000"
Overall Commands: 1000000 IOPS: 1255515

$ xbutil scheduler --test xcl_iops --args "-t 4 -l 128 -a 1000000"
Overall Commands: 4000000 IOPS: 2627524

**Without this change** (Same hardware + XRT 2.10.27)
$ xbutil scheduler --test xcl_iops --args "-t 1 -l 128 -a 1000000"
Overall Commands: 1000000 IOPS: 1212855

$ xbutil scheduler --test xcl_iops --args "-t 4 -l 128 -a 1000000"
Overall Commands: 4000000 IOPS: 2536824

### DIFF
--- a/src/CMake/dkms-edge.cmake
+++ b/src/CMake/dkms-edge.cmake
@@ -97,6 +97,7 @@ SET (XRT_DKMS_COMMON_XRT_DRV_INCLUDES
   common/drv/include/kds_client.h
   common/drv/include/xrt_cu.h
   common/drv/include/xrt_xclbin.h
+  common/drv/include/kds_stat.h
   )
 
 SET (XRT_DKMS_CORE_EDGE_INCLUDES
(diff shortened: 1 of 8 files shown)


maxzhen (approved):

maxzhen:
retest this please - host cold reboot time out caused by infra issue.

PR #4693, opened 2021-02-04
CR-1087204 fix for p2p_fpga2fpga data transfer

madhusud-xilinx:
fix for p2p_fpga2fpga data transfer

PR #4721, opened 2021-02-12
CR-1087204 fix for p2p_fpga2fpga data transfer (#4693)

madhusud-xilinx:
* CR-1087204 fix for p2p_fpga2fpga data transfer

* CR-1087204 incorporated review comments


chvamshi-xilinx:
ok to test

PR #4722, opened 2021-02-12
CR-1083805 Added then API xclExecBufWithWaitList even for hw emu to s…

venkatp-xilinx:
CR-1083805 Added then API xclExecBufWithWaitList even for hw emu to support XMA designs.


venkatp-xilinx:
travis build failed with unrelated error, retest this please.

PR #4723, opened 2021-02-12
- XRT update for XSS

naveenis-xilinx:
- change in xocl_bo.c file : it is for a XRT crash fix in different block size case.
- change in xocl_kds.c file : It is a typo fix for getting correct error status.


chvamshi-xilinx:
ok to test

mamin506 (approved):

maxzhen:
retest this please - last test result is lost

PR #4724, opened 2021-02-12
Fix for Stream Data Transfer row in opencl_summary

IshitaGhosh:
Fix for CR 1087895, 1086643


jvillarre (approved):
Changes look good to use the correct index into the counters

Issue #73, opened 2018-07-13
Add build id with git hash (and timestamp) to every build


sonals:
Every time git hash changes we should embed git hash (and timestamp) into all our binaries:

xbutil (xbsak)
xbflash
libxrtcore.so
libxilinxopencl.so
HAL libs

The utilities should display the build id when they start.
The libraries could return the build id as part of an API.

uday610:
@sonals , is this still an issue?

stsoe:
No longer relevant.  version.h has information.

PR #4725, opened 2021-02-12
xocl: issue xbutil reset only if device is in ready state

rajkumar-xilinx:
Issue xbutil reset command only when device is in ready state. Also, read return status of xocl peer requests and act accordingly.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
@@ -338,12 +338,12 @@ failed:
  *  a) If xocl unable to communicate to mgmt/mpd
  *     - xocl should reenable all the sub-devices and mark the device online/ready.
  *  b) If reset Channel is disabled
- *	   - xocl should reenable all the sub-devices and mark the device online/ready.
- *	c) Reset is issued to mpd, but mpd doesn’t have serial number of requested device
- *	   - MPD returns E_EMPTY serial number error code to xocl
- *	   - xocl should reenable all the sub-devices and mark the device online/ready.
- *	d) Reset is issued to mgmt/mpd, but mgmt/mpd unable to reset properly
- *	   - xocl gets a ESHUTDOWN response from mgmt/mpd,
+ *     - xocl should reenable all the sub-devices and mark the device online/ready.
+ *  c) Reset is issued to mpd, but mpd doesn’t have serial number of requested device
+ *     - MPD returns E_EMPTY serial number error code to xocl
+ *     - xocl should reenable all the sub-devices and mark the device online/ready.
+ *  d) Reset is issued to mgmt/mpd, but mgmt/mpd unable to reset properly
+ *     - xocl gets a ESHUTDOWN response from mgmt/mpd,
  *     - xocl assumes that reset is successful,
  *     - xbutil waits on the device ready state in a loop.
  *     - xbutil reset would be in waiting state forever.
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
@@ -327,6 +327,31 @@ failed:
 	return ret;
 }
 
+/*
+ * Reset command should support following cases
+ * case 1) When device is not in ready state
+ *  - xbutil should not send any request to the xocl.
+ *  - It should just return fail status from userspace itself
+ * case 2) When device is ready & device offline status is true
+ *  - Need to check when we hit this case
+ * case 3) When device is ready & online
+ *  a) If xocl unable to communicate to mgmt/mpd
+ *     - xocl should reenable all the sub-devices and mark the device online/ready.
+ *  b) If reset Channel is disabled
+ *	   - xocl should reenable all the sub-devices and mark the device online/ready.
+ *	c) Reset is issued to mpd, but mpd doesn’t have serial number of requested device
+ *	   - MPD returns E_EMPTY serial number error code to xocl
+ *	   - xocl should reenable all the sub-devices and mark the device online/ready.
+ *	d) Reset is issued to mgmt/mpd, but mgmt/mpd unable to reset properly
+ *	   - xocl gets a ESHUTDOWN response from mgmt/mpd,
+ *     - xocl assumes that reset is successful,
+ *     - xbutil waits on the device ready state in a loop.
+ *     - xbutil reset would be in waiting state forever.
+ *     - Need to handle this case to exit xbutil reset gracefully.
+ *  e) Reset is issued to mgmt/mpd, but mgmt/mpd reset properly
+ *     - xocl gets a ESHUTDOWN response from mgmt/mpd,
+ *     - Device becomes ready and xbutil reset successful.
+ */
 int xocl_hot_reset(struct xocl_dev *xdev, u32 flag)
 {
 	int ret = 0, mbret = 0;
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
@@ -366,11 +366,9 @@ int xocl_hot_reset(struct xocl_dev *xdev, u32 flag)
 
 		mbret = xocl_peer_request(xdev, &mbreq, sizeof(struct xcl_mailbox_req),
 			&ret, &resplen, NULL, NULL, 0, 6);
-		if (mbret)
-			xocl_reset_notify(xdev->core.pdev, false);
 		/* userpf will back online till receiving mgmtpf notification */
 
-		return 0;
+		return mbret;
 	}
 
 	mbret = xocl_peer_request(xdev, &mbreq, sizeof(struct xcl_mailbox_req),
@@ -484,6 +482,7 @@ static void xocl_work_cb(struct work_struct *work)
 	struct xocl_dev *xdev = container_of(_work,
 			struct xocl_dev, core.works[_work->op]);
 	struct xocl_dev *buddy_xdev = xdev;
+	int ret;
 
 	if (XDEV(xdev)->shutdown && _work->op != XOCL_WORK_ONLINE) {
 		xocl_xdev_info(xdev, "device is shutdown please hotplug");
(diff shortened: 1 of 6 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
```diff
@@ -364,8 +364,10 @@ int xocl_hot_reset(struct xocl_dev *xdev, u32 flag)
 		if (flag & XOCL_RESET_NO)
 			return 0;
 
-		xocl_peer_request(xdev, &mbreq, sizeof(struct xcl_mailbox_req),
+		mbret = xocl_peer_request(xdev, &mbreq, sizeof(struct xcl_mailbox_req),
 			&ret, &resplen, NULL, NULL, 0, 6);
+		if (mbret)
+			xocl_reset_notify(xdev->core.pdev, false);
```
rajkumar-xilinx, line 370: Also, we need to set any available sysfs node to indicate this error to xbutil/shim layer.
Or, use xocl_drvinst_set_offline(<pass error code>), so that xbutil reads this error code and exit from while (dev_offline) loop in xclReset() shim api.

rajkumar-xilinx:
With this change, xbutil reset terminates gracefully with below error message on dmesg log
**[  +0.000010] xocl 0101:00:00.0: mailbox.u.9437184 ffff899ab534b010 mailbox_request: req 5 is received on disabled channel, err: -14**

chvamshi-xilinx (approved):

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -1766,6 +1766,10 @@ int xcldev::xclReset(int argc, char *argv[])
 
     std::string vbnv, errmsg;
     auto dev = pcidev::get_dev(index);
+    if (!dev->is_ready) {
```
maxzhen, line 1769: should we also check if device is offline? offline means a previous reset has not finished or driver has not finished probing, so not ready for handling reset?

rajkumar-xilinx: If the previous reset has issue or driver probe failed, then device is not in ready state. So, I think one more "dev_offline" check may not be required. Please let me know if this is not the case, I will add dev_offline check as well here.

rajkumar-xilinx:
**#xbutil list**

> INFO: Found total 1 card(s), 0 are usable
> WARNING: The xbutil sub-command list has been deprecated. Please use xbutil scan for equivalent functionality.
> 
> ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> System Configuration
> OS name:        Linux
> Release:        5.4.0-65-generic
> Version:        #73~18.04.1-Ubuntu SMP Tue Jan 19 09:02:24 UTC 2021
> Machine:        x86_64
> Model:          Virtual Machine
> CPU cores:      8
> Memory:         16036 MB
> Glibc:          2.27
> Distribution:   Ubuntu 18.04.5 LTS
> Now:            Mon Feb 15 21:34:13 2021 GMT
> ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> XRT Information
> Version:        2.10.0
> Git Hash:       0dc24e88ab155c6fbc6e4f9c655f726e6e498be9
> Git Branch:     12feb_master
> Build Date:     2021-02-16 02:59:43
> XOCL:           2.10.0,0dc24e88ab155c6fbc6e4f9c655f726e6e498be9
> XCLMGMT:        2.10.0,0dc24e88ab155c6fbc6e4f9c655f726e6e498be9
> ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> *[0] 0101:00:00.0 xilinx_u250_gen3x16_base_2 user(inst=128)
> WARNING: card(s) marked by '*' are not ready.

**#xbutil reset**

> device [0] is not ready, reset command exiting

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
maxzhen: @rajkumar-xilinx , thanks for clarifying. do we know why device is not in ready state if the previous reset has failed? I'm wondering if we should also add dev_offline as one of the condition in ready sysfs node? Should we consider device as ready if it's not online, @houlz0507 @xuhz ?

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
```diff
@@ -364,11 +364,11 @@ int xocl_hot_reset(struct xocl_dev *xdev, u32 flag)
 		if (flag & XOCL_RESET_NO)
 			return 0;
 
-		xocl_peer_request(xdev, &mbreq, sizeof(struct xcl_mailbox_req),
+		mbret = xocl_peer_request(xdev, &mbreq, sizeof(struct xcl_mailbox_req),
 			&ret, &resplen, NULL, NULL, 0, 6);
 		/* userpf will back online till receiving mgmtpf notification */
 
-		return 0;
+		return mbret;
```
xuhz, line 371: this is not correct. 
both mbret and ret should be checked. and ret of ESHUTDOWN is the errno returned in mbx response when reset is issued successfully.
pls refer to the code in same function.
also pls try to make the code in this function neat -- current code looks really messy.

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
```diff
@@ -498,11 +499,16 @@ static void xocl_work_cb(struct work_struct *work)
...
-		(void) xocl_hot_reset(xdev, XOCL_RESET_FORCE |
+		ret = xocl_hot_reset(xdev, XOCL_RESET_FORCE |
 			       	XOCL_RESET_SHUTDOWN);
 		mutex_unlock(&xocl_reset_mutex);
+		if (ret) {
+			xocl_xdev_err(xdev, "reset failed with error %d", ret);
+			xocl_reset_notify(xdev->core.pdev, false);
+			xocl_drvinst_set_offline(xdev->core.drm, false);
+		}
 		break;
```
xuhz, line 512: if you handle the failure case in xocl_hot_reset(), no need to check the return code here. otherwise, you may need to check the return code of xocl_hot_reset() at other spots also
It'd better to introduce another sysfs node to save the reset status, and tell the user whether the reset succeeds or not with this sysfs node.

src/runtime_src/core/pcie/tools/cloud-daemon/mpd.cpp
```diff
@@ -277,8 +277,11 @@ void Mpd::start()
             return;
         }
         int ret = (*plugin_init)(&plugin_cbs);
-        if (ret != 0)
-            syslog(LOG_ERR, "mpd plugin_init failed: %d", ret);
+        if (ret != 0) {
+            syslog(LOG_ERR, "mpd plugin_init failed: %d, restart the mpd service", ret);
+            stop();
+            throw std::runtime_error("can't start due to plugin_init failed");
```
xuhz, line 283: stop() seems not required.

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
xuhz: seems the device is always online when it is ready. again, offline caused by a failed reset is a regression.

xuhz (changes requested):

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
maxzhen: The condition we check for ready does not include online/offline today. The 'ready' sysfs node is created before device is set to online (which is the very last step in probe() routine). So there is a small window when the 'ready' sysfs node says 1, but device is not online yet. @houlz0507 can correct me, if I'm wrong.

xuhz:
xocl_drvinst_set_offline(xdev, false);
called twice
I am not sure whether this is deliberate.
1st time, before xocl_mb_connect() is called, it is called already. so before xocl gets the ready state info from mgmt, it is set as online already

maxzhen:
> 
> 
> xocl_drvinst_set_offline(xdev, false);
> called twice
> I am not sure whether this is deliberate.
> 1st time, before xocl_mb_connect() is called, it is called already. so before xocl gets the ready state info from mgmt, it is set as online already

The first call sounds like a bug to me.

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
rajkumar-xilinx: Hi @xuhz , when I checked both mbret and ret values, I see only mbret has some error code when hot reset request is getting timed out.

> [  +0.000380] xocl 0101:00:00.0: mailbox.u.9437184 ffff92bbde687810 mailbox_request: sending request: 5 via SW
> [  +0.012797] xocl 0101:00:00.0: mailbox.u.9437184 ffff92bbde687810 mailbox_read: Software TX msg is too big
> [  +0.950709] xocl 0101:00:00.0: mailbox.u.9437184 ffff92bbde687810 xocl_drvinst_close: CLOSE 2
> [  +0.000007] xocl 0101:00:00.0: mailbox.u.9437184 ffff92bbde687810 xocl_drvinst_close: NOTIFY 00000000f61ca0e9
> [Feb17 22:07] xocl 0101:00:00.0: mailbox.u.9437184 ffff92bbde687810 timeout_msg: found waiting msg time'd out
> [  +0.000147] xocl 0101:00:00.0:  ffff92ba3319a0b0 xocl_hot_reset: +++mbret: -62, ret: 0
> [  +0.000007] xocl 0101:00:00.0:  ffff92ba3319a0b0 xocl_work_cb: reset failed with error -62

xuhz: if the mailbox msg doesn't time out, eg. the S/N missing case, the mpd plugin will return -2040 that would be in ret (response)

xuhz: and if the reset is sent to wireserver, the mailbox msg response will be ESHUTDOWN (in ret too), which tells the xocl that the whole xocl would be gone. In this case, reset will be waiting in userland till the xocl probe set the dev online

rajkumar-xilinx: yes, I will add code to check ret value also.

src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.cpp
```diff
@@ -73,17 +73,42 @@ int init(mpd_plugin_callbacks *cbs)
...
+
         int curlInit = curl_global_init(CURL_GLOBAL_ALL);
         if (curlInit != 0)
             syslog(LOG_ERR, "mpd cannot initalize curl: %d", curlInit);
         std::string private_ip = AzureDev::get_wireserver_ip();
         if (!private_ip.empty())
             restip_endpoint = private_ip;
         syslog(LOG_INFO, "azure restserver ip: %s\n", restip_endpoint.c_str());
-        fpga_serial_number = AzureDev::get_serial_number();
+        while (retry > 0) {
```
rajkumar-xilinx, line 90: remove this code

chvamshi-xilinx:
Hi @rajkumar-xilinx , As Sonal Suggested, please add all the combinations which are getting fixed with this change.
Also, add if any of the case is not handled (like xocl/mpd race condition)

chvamshi-xilinx:
image

chvamshi-xilinx (approved):

maxzhen (approved):

houlz0507 (approved):

xuhz (approved):

chvamshi-xilinx:
chery pick this change to 2020.2 as Azure is using 2020.2

PR #4727, opened 2021-02-12
Profiling fixes in the refactored code base

jvillarre:
This pull request addresses the following issues in the profiling architecture:

- Adding a mutex to address a race condition in the generation of events
- Adding the ability to exclude device level profiling on xclbins generated from older tools

### DIFF
--- a/src/runtime_src/xdp/profile/database/events/creator/device_event_from_trace.cpp
+++ b/src/runtime_src/xdp/profile/database/events/creator/device_event_from_trace.cpp
@@ -34,21 +34,12 @@ namespace xdp {
     xclbin = (db->getStaticInfo()).getCurrentlyLoadedXclbin(devId) ;
 
     auto numAM = (db->getStaticInfo()).getNumAM(deviceId, xclbin) ;
+    traceIDs.resize(numAM);
+    cuStarts.resize(numAM);
+    amLastTrans.resize(numAM);
 
-    if (numAM > 0) {
-      traceIDs.resize(numAM) ;
-      cuStarts.resize(numAM) ;
-      amLastTrans.resize(numAM) ;
-    }
-
-    auto numAIM = (db->getStaticInfo()).getNumAIM(deviceId, xclbin) ;
-    if (numAIM > 0) {
-      aimLastTrans.resize(numAIM);
-    }
-    auto numASM = (db->getStaticInfo()).getNumASM(deviceId, xclbin) ;
-    if (numASM > 0) {
-      asmLastTrans.resize(numASM);
-    }
+    aimLastTrans.resize((db->getStaticInfo()).getNumAIM(deviceId, xclbin)) ;
+    asmLastTrans.resize((db->getStaticInfo()).getNumASM(deviceId, xclbin)) ;
   }
 
   void DeviceEventCreatorFromTrace::createDeviceEvents(xclTraceResultsVector& traceVector)
--- a/src/runtime_src/xdp/profile/database/events/creator/device_event_from_trace.cpp
+++ b/src/runtime_src/xdp/profile/database/events/creator/device_event_from_trace.cpp
@@ -33,12 +33,22 @@ namespace xdp {
 
     xclbin = (db->getStaticInfo()).getCurrentlyLoadedXclbin(devId) ;
 
-    traceIDs.resize(db->getStaticInfo().getNumAM(deviceId, xclbin));
-    cuStarts.resize(db->getStaticInfo().getNumAM(deviceId, xclbin));
-    amLastTrans.resize(db->getStaticInfo().getNumAM(deviceId, xclbin));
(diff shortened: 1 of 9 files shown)


src/runtime_src/xdp/profile/database/static_info_database.cpp
```diff
@@ -151,6 +151,63 @@ namespace xdp {
...
+      boost::property_tree::ptree pt;
+      boost::property_tree::read_json(ss, pt);
+
+      std::string version =
+	pt.get<std::string>("build_metadata.xclbin.generated_by.version", "") ;
+
+      if (version == "" || version.find(".") == std::string::npos)
+	return false ;
+
+      auto firstDotPosition = version.find(".") ;
```
IshitaGhosh, line 178: Instead of parsing the version string and then converting major and minor value separately into "int", we can use "strtod" function to convert the version string to a "double" value and compare it with 2019.2 double value.

IshitaGhosh (changes requested):

src/runtime_src/xdp/profile/plugin/device_offload/hal/hal_device_offload_plugin.cpp
```diff
@@ -185,6 +186,13 @@ namespace xdp {
     void* ownedHandle = deviceIdToHandle[deviceId] ;
   
     clearOffloader(deviceId); 
+
+    if (!(db->getStaticInfo()).validXclbin(userHandle)) {
+      xrt_core::message::send(xrt_core::message::severity_level::warning,
+			      "XRT",
+			      "Device profiling is no longer supported on this xclbin.  to enable dvice profiling please rebuild with the current version of the tools.") ;
```
IshitaGhosh, line 193: I think we should mention the tool version till which we are supporting the xclbins i.e. currently 2019.2. We can add a "double" type const member in Static Info Database and use it while checking valid/supported xclbin and also while reporting this warning. If there is a change in supported version in future, we can just update the const member to the supported version double value.

src/runtime_src/xdp/profile/plugin/device_offload/opencl/opencl_device_offload_plugin.cpp
```diff
@@ -220,6 +221,13 @@ namespace xdp {
 
     clearOffloader(deviceId);
 
+    if (!(db->getStaticInfo()).validXclbin(device->get_xcl_handle())) {
```
IshitaGhosh, line 224: Please add the version till which xclbin is supported.

IshitaGhosh (changes requested):

src/runtime_src/xdp/profile/plugin/device_offload/hal/hal_device_offload_plugin.cpp
IshitaGhosh: Actually, a method in VPStaticDatabase
double VPStaticDatabase::getOldestSupportedToolVersion { return double(2019.2); }
would be better.

src/runtime_src/xdp/profile/database/events/creator/device_event_from_trace.cpp
```diff
@@ -33,12 +33,22 @@ namespace xdp {
...
     xclbin = (db->getStaticInfo()).getCurrentlyLoadedXclbin(devId) ;
 
-    traceIDs.resize(db->getStaticInfo().getNumAM(deviceId, xclbin));
-    cuStarts.resize(db->getStaticInfo().getNumAM(deviceId, xclbin));
-    amLastTrans.resize(db->getStaticInfo().getNumAM(deviceId, xclbin));
+    auto numAM = (db->getStaticInfo()).getNumAM(deviceId, xclbin) ;
 
-    aimLastTrans.resize(db->getStaticInfo().getNumAIM(deviceId, xclbin));
-    asmLastTrans.resize(db->getStaticInfo().getNumASM(deviceId, xclbin));
+    if (numAM > 0) {
```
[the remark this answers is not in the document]
jvillarre: This code was introduced while I was tracking down memory corruption.  I have reverted it in the latest commit

src/runtime_src/xdp/profile/database/static_info_database.cpp
jvillarre: After verifying that std::stod works on strings with single decimal points and multiple periods, I have converted this code to a call to std::stod

src/runtime_src/xdp/profile/plugin/device_offload/hal/hal_device_offload_plugin.cpp
jvillarre: Added function to the static database and used it in the construction of the message.

IshitaGhosh (approved):
Changes look good

PR #4728, opened 2021-02-12
CR-1088338 and CR-1086561 fixes

jvillarre:
Fixes that rename xclbin.ex.run_summary and the output unit on the data transfer tables in the summary.


IshitaGhosh (approved):
Changes look good

PR #4732, opened 2021-02-15
CR-1088262: fix grp index issue for use host mem ptr in edge

Himanshu-xilinx:
* Recent changes in XRT userspace is relying on driver to get group/bank index.
* xocl userspace is getting the group/bank index from the flags which are returned by driver.
* xclAllocBO is storing the bo flags and returning them whenever getBOproperties get called.
* xclAllocUserPtrBo is not handling this in edge driver.
* Fix: made changes in xclAllocUserPtrBo also.


chvamshi-xilinx (approved):
Looks good.

stsoe (changes requested):
Please update description with what was the root cause and why the fixes in this PR fixes it.

larry9523 (approved):
The only thing is missing is the copyright year should be updated. But if this is urgent, we can fix it with a later push.

stsoe (approved):

PR #4733, opened 2021-02-16
CR-1087409 Provide IOPS test host code for new validate kernels

mamin506:
What are added:
1. Support IOPS test case in new 'xbutil validate' flow.
2. Show CU name at the end of IOPS test result
    - Old flow, kernel name 'hello' and has 7 arguments.
    - New flow, kernel name 'verify' and has 1 argument.
3. Add internal API, xclOpenByBDF();

Test result:
U30
```
$xbutil validate -d 0
...
INFO: == Starting IOPS test:
Maximum IOPS: 94838 (verify)
INFO: == IOPS test PASSED
...
```
U50
```
...
INFO: == Starting IOPS test: 
Maximum IOPS: 184314 (hello)
INFO: == IOPS test PASSED
...
```

Hi @vishnuchebrolu, I have one change in cmdlineparser.cpp. In before, for a toggle switch, the default value is true. I thought this is a bug. Please confirm. I have a toggle switch '-v' to turn verbose mode on/off.

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
+++ b/src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
@@ -1174,7 +1174,9 @@ int xcldev::device::runTestCase(const std::string& py,
         static const std::map<std::string, std::string> test_map = {
             { "22_verify.py",             "validate.exe"    },
             { "23_bandwidth.py",          "kernel_bw.exe"   },
-            { "host_mem_23_bandwidth.py", "slavebridge.exe" }
+            { "host_mem_23_bandwidth.py", "slavebridge.exe" },
+            { "xrt_iops_test.exe",        "xrt_iops_test.exe" },
+            { "xcl_iops_test.exe",        "xcl_iops_test.exe" }
         };
         
         if (test_map.find(py) == test_map.end())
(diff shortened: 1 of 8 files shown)


vishnuchebrolu:
@mamin506 , We had the following as part of commandline parser in Vitis Git Examples : 
https://gitenterprise.xilinx.com/SDxGitExamples/Vitis_Accel_Examples/blob/master/common/includes/cmdparser/cmdlineparser.cpp#L106
Can you explain why the default is changed to false?

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
```diff
@@ -1191,7 +1193,12 @@ int xcldev::device::runTestCase(const std::string& py,
         auto device = pcidev::get_dev(m_idx);
         std::string bdf = boost::str(boost::format("%04x:%02x:%02x.%01x") % device->domain % device->bus % device->dev % device->func);
 
-        cmd = xrtTestCasePath + " " + xclbinPath + " -d " + bdf;
+        // Other new flow test cases are using OpenCL APIs, which can find device by BDF.
+        // If XRT and XCL APIs support open device by BDF, we can unify below two cmds.
+        if (py.find("xrt_iops_test") || py.find("xcl_iops_test"))
+            cmd = xrtTestCasePath + " " + xclbinPath + " -d " + std::to_string(m_idx) + " " + args;
```
maxzhen, line 1199: we have to support opening device by bdf, m_idx may point to a different device in iops test process.

maxzhen (changes requested):

mamin506:
> @mamin506 , We had the following as part of commandline parser in Vitis Git Examples :
> https://gitenterprise.xilinx.com/SDxGitExamples/Vitis_Accel_Examples/blob/master/common/includes/cmdparser/cmdlineparser.cpp#L106
> Can you explain why the default is changed to false?

In addSwitch(), the origin code is "cmd.isvalid = true". In parse(), it uses "pcmd->isvalid = true;". So, no matter the toggle switch applied or not, isValid() would always return true.

AShivangi (approved):
This looks good. 

@maxzhen, we don't have iops test support in xbutil2. Do you think we can skip adding this for the Beta release?

maxzhen:
> 
> 
> This looks good.
> 
> @maxzhen, we don't have iops test support in xbutil2. Do you think we can skip adding this for the Beta release?

Yes, we don't need this for U30 Beta.

mamin506:
@stsoe As per our discussion, add shim_int.h for an internal shim API xclOpenByBDF() for Linux PCIe platform.

src/runtime_src/core/pcie/tools/xbutil/xbutil.cpp
mamin506: Fixed.

maxzhen (approved):

stsoe (approved):

PR #4738, opened 2021-02-16
continuous writing in xdp plugins

hackwa:
Currently only opencl and hal device trace are written continuously. 
Opencl host, lop data can be enabled in future.


jvillarre (approved):
Finer-grained locking and additions for continuous writing thread look good.

PR #4739, opened 2021-02-16
NSPR-9558 Reconcile MEM TOPOLOGY for U30/VCK5000

larry9523:
This PR addresses the following potential problem by comparing mem_topology with reserved_ps_mem in xsabin.

Reserving memory at PS Petalinux device tree is to let Petalinux know the memory region is reserved by PS/PL kernel so that OS won't touch it. But this information is not visible by host XRT driver. Host xocl will relies on information in XCLBIN to allocate memory. So if memory region in XCLBIN does not match reserved memory in PS Petalinux, we will have problem

Host will allocate memory that can not be used by PS/PL kernel that causes PS/PL kernel fail/hang (if there is enough hardware protection to prevent PS/PL kernel accessing illegal memory region)
Serious security issue if there is not enough hardware isolation
1) Crash Petalinux
2) Hacking XCBLIN can allow host application to access any portion of device memory

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
@@ -1,7 +1,7 @@
 /*
  * A GEM style device manager for PCIe based OpenCL accelerators.
  *
- * Copyright (C) 2016-2019 Xilinx, Inc. All rights reserved.
+ * Copyright (C) 2016-2021 Xilinx, Inc. All rights reserved.
  *
  * Authors: Jan Stephan <j.stephan@hzdr.de>
  *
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/xocl_drm.h
```diff
@@ -25,6 +25,8 @@
 typedef void (*xocl_execbuf_callback)(unsigned long data, int error);
 
 #define IS_HOST_MEM(m_tag)	(!strncmp(m_tag, "HOST[0]", 7))
+#define IS_PLRAM(m_tag)		(!strncmp(m_tag, "PLRAM[0]", 8))
```
maxzhen, line 28: we may have more than one PLRAM, right? should we just compare "PLRAM["?

larry9523: That was my initial thought. But I am not sure why we are using HOST[0]? Maybe HOST[1] means differently?

maxzhen: XRT currently can only support one slave port, which is "HOST[0]". "HOST[x]" is not expected...

larry9523: done

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
```diff
@@ -188,6 +188,64 @@ static int xocl_mmap(struct file *filp, struct vm_area_struct *vma)
...
+
+	/* PLRAM does not have to be accessed by PS */
+	if (IS_PLRAM(mem_data->m_tag))
+		return true;
+
+	blob = XDEV(xdev)->fdt_blob;
+	if (!blob)
+		return true;
+
+	for (offset = fdt_next_node(blob, -1, NULL);
```
houlz0507, line 213: Maybe move this part to xocl_fdt.c? Trying to hide all fdt_* calls in xocl_fdt.c

maxzhen (approved):

houlz0507 (approved):

chienwei-lan (approved):

PR #4759, opened 2021-02-18
CR-1088591

AShivangi:
Based on the discussion with PRT:

BLP: --base, -b
PLP:--shell, -s
ULP: --user, -u


rozumx (approved):
Thanks for making these changes.  

Note: Normally unused code should be removed instead of commented out.  That said, I do understand that in the near future you will be adding these options...so it is O.K. to to just comment them out in this case.

PR #4762, opened 2021-02-19
CR-1077304: Displaying copy buffer events in the OpenCL timeline trace

jvillarre:
Reintroducing the "Copy" category in the opencl_trace file for copy buffer events.  This was missing in the refactored code and copy buffer events were being put in the general API bucket.


IshitaGhosh (approved):
Changes look good.

PR #4764, opened 2021-02-19
CR-1089448 Add xbmgmt2 ulp program

AShivangi:
Sample output:
```
$ xbmgmt -new program --user /opt/xilinx/firmware/u30/gen3x4/base/test/verify.xclbin -d 17:00
Downloading xclbin on device [0000:17:00.0]...
INFO: Successfully downloaded xclbin

$
```

### DIFF
--- a/src/runtime_src/core/common/device.h
+++ b/src/runtime_src/core/common/device.h
@@ -292,6 +292,8 @@ public:
 
   virtual void reset(query::reset_type&) const {}
 
+  virtual void load_xclbin(std::vector<char> &buffer) const{}
+
   /**
    * open() - opens a device with an fd which can be used for non pcie read/write
    * xospiversal and xspi use this
--- a/src/runtime_src/core/common/device.h
+++ b/src/runtime_src/core/common/device.h
@@ -45,6 +45,7 @@
 namespace xrt_core {
 
 using device_collection = std::vector<std::shared_ptr<xrt_core::device>>;
+using MemoryBuffer = const std::vector<char>;
 
 /**
  * class device - interface to support OS agnositic operations on a device
@@ -292,7 +293,7 @@ public:
 
   virtual void reset(query::reset_type&) const {}
 
-  virtual void load_xclbin(std::vector<char> &) const{}
+  virtual void load_xclbin(MemoryBuffer &) const{}
 
   /**
    * open() - opens a device with an fd which can be used for non pcie read/write
--- a/src/runtime_src/core/common/device.h
+++ b/src/runtime_src/core/common/device.h
@@ -45,7 +45,7 @@
 namespace xrt_core {
 
 using device_collection = std::vector<std::shared_ptr<xrt_core::device>>;
-using MemoryBuffer = const std::vector<char>;
+using MemoryBuffer = std::vector<char>;
 
 /**
  * class device - interface to support OS agnositic operations on a device
@@ -293,7 +293,7 @@ public:
 
   virtual void reset(query::reset_type&) const {}
 
-  virtual void load_xclbin(MemoryBuffer &) const{}
+  virtual void load_xclbin(const MemoryBuffer &) const{}
 
   /**
    * open() - opens a device with an fd which can be used for non pcie read/write
--- a/src/runtime_src/core/common/device.h
+++ b/src/runtime_src/core/common/device.h
@@ -292,7 +292,7 @@ public:
 
   virtual void reset(query::reset_type&) const {}
 
-  virtual void load_xclbin(std::vector<char> &buffer) const{}
+  virtual void load_xclbin(std::vector<char> &) const{}
 
   /**
    * open() - opens a device with an fd which can be used for non pcie read/write
--- a/src/runtime_src/core/pcie/linux/device_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/device_linux.cpp
@@ -514,8 +514,11 @@ close(int dev_handle) const
 void
 device_linux::
 load_xclbin(std::vector<char> &buffer) const {
+  static const xclbin_magic_value = 7;
+  std::string xclbin_magic_str(buffer.data(), buffer.data() + xclbin_magic_value);
+  if (xclbin_magic_str != "xclbin2")
+    throw xrt_core::error(boost::str(boost::format("Bad binary version '%s'") % xclbin_magic_str));
   int ret = 0;
-
   try {
     xrt_core::scope_value_guard<int, std::function<void()>> fd = file_open("", O_RDWR);
     xclmgmt_ioc_bitstream_axlf obj = { reinterpret_cast<axlf *>(buffer.data()) };
--- a/src/runtime_src/core/pcie/linux/device_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/device_linux.cpp
@@ -514,7 +514,7 @@ close(int dev_handle) const
 void
 device_linux::
 load_xclbin(std::vector<char> &buffer) const {
-  static const xclbin_magic_value = 7;
+  static const int xclbin_magic_value = 7;
   std::string xclbin_magic_str(buffer.data(), buffer.data() + xclbin_magic_value);
   if (xclbin_magic_str != "xclbin2")
     throw xrt_core::error(boost::str(boost::format("Bad binary version '%s'") % xclbin_magic_str));
--- a/src/runtime_src/core/pcie/linux/device_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/device_linux.cpp
@@ -513,9 +513,11 @@ close(int dev_handle) const
 
 void
 device_linux::
-load_xclbin(MemoryBuffer &buffer) const {
+load_xclbin(const MemoryBuffer &buffer) const {
   //resolves to xclbin2
   const char xclbin_magic_str[] = { 0x78, 0x63, 0x6c, 0x62, 0x69, 0x6e, 0x32 };
+  if (buffer.size() < sizeof(xclbin_magic_str))
+    throw xrt_core::error("Xclbin is smaller than expected");
   if (std::memcmp(buffer.data(), xclbin_magic_str, sizeof(xclbin_magic_str)) != 0)
     throw xrt_core::error(boost::str(boost::format("Bad binary version '%s'") % xclbin_magic_str));
   int ret = 0;
--- a/src/runtime_src/core/pcie/linux/device_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/device_linux.cpp
@@ -521,8 +521,7 @@ load_xclbin(MemoryBuffer &buffer) const {
   int ret = 0;
   try {
     xrt_core::scope_value_guard<int, std::function<void()>> fd = file_open("", O_RDWR);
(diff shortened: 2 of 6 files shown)


src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -730,6 +733,36 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
...
+    std::vector<char> raw(size);
+    stream.read(raw.data(), size);
+    std::string v(raw.data(),raw.data()+7);
+    if (v != "xclbin2")
+      throw xrt_core::error(boost::str(boost::format("Bad binary version '%s'") % v));
+    
+    auto bdf = xrt_core::query::pcie_bdf::to_string(xrt_core::device_query<xrt_core::query::pcie_bdf>(dev));
+    std::cout << "Downloading xclbin on device [" << bdf << "]..." << std::endl;
+    dev->load_xclbin(raw);
+    std::cout << boost::str(boost::format("INFO: Successfully downloaded xclbin \n\n"));
```
rozumx, line 761: Boost format knows out to write to an istream.  Therefore, there is no need to convert it to a string.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -730,6 +733,36 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
...
+    if (!stream)
+      throw xrt_core::error(boost::str(boost::format("Could not open %s for reding") % xclbin));
+
+    stream.seekg(0,stream.end);
+    ssize_t size = stream.tellg();
+    stream.seekg(0,stream.beg);
+
+    std::vector<char> raw(size);
+    stream.read(raw.data(), size);
+    std::string v(raw.data(),raw.data()+7);
```
rozumx, line 754: Magic number '7'....I know it is lucky but why is it here :^).

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -730,6 +733,36 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
...
+  if(!xclbin.empty()) {
+    XBU::verbose(boost::str(boost::format("  xclbin: %s") % xclbin));
+    //only 1 card and name
+    if(deviceCollection.size() > 1)
+      throw xrt_core::error("Please specify a single device");
+    auto dev = deviceCollection.front();
+
+    std::ifstream stream(xclbin, std::ios::binary);
+    if (!stream)
+      throw xrt_core::error(boost::str(boost::format("Could not open %s for reding") % xclbin));
```
rozumx, line 746: What does the color "red" have to do with reading in the file....LOL

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -730,6 +733,36 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
...
+
+    stream.seekg(0,stream.end);
+    ssize_t size = stream.tellg();
+    stream.seekg(0,stream.beg);
+
+    std::vector<char> raw(size);
```
rozumx, line 752: Instead of calling the buffer "raw," which is rather generic, it might be best to call it something like "xclbinBuffer"

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
rozumx, line 754: The variable "v" should be more descriptive, for example: xclbinMagicValue.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -730,6 +733,36 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
...
+    stream.read(raw.data(), size);
+    std::string v(raw.data(),raw.data()+7);
+    if (v != "xclbin2")
+      throw xrt_core::error(boost::str(boost::format("Bad binary version '%s'") % v));
```
rozumx, line 756: The Error message while valid is:

1. Not very helpful.
2. Will produce garbage if pointing to a binary file.

Q: Why is this check being done here as opposed to in the "load_xclbin" method?

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -510,4 +511,22 @@ close(int dev_handle) const
   pcidev::get_dev(get_device_id(), false)->close(dev_handle);
 }
 
+void
+device_linux::
+load_xclbin(std::vector<char> &buffer) const {
+  int ret = 0;
+
+  try {
```
rozumx, line 519: Magic "binary value" check (e.g., "xclbin2" should be done here.  This this check should be a BINARY check and NOT text.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -510,4 +511,22 @@ close(int dev_handle) const
   pcidev::get_dev(get_device_id(), false)->close(dev_handle);
 }
 
+void
+device_linux::
+load_xclbin(std::vector<char> &buffer) const {
```
rozumx, line 516: Being that std::vector<char> is being used in a lot of places and, have you thought about defining it in a "using" statement.  Something like:
`using MemoryBuffer=std::vector<char>;`
and then changing this method signature to:
`load_xclbin(const MemoryBuffer &buffer) const {`

And yes, the buffer should be declared `const `for it isn't being changed.

rozumx (changes requested):
Not needed for this Pull Request, but for future pull requests the method:
`SubCmdProgram::execute()`
Is too large and needs refactored into multiple smaller sub methods().

src/runtime_src/core/pcie/linux/device_linux.cpp
AShivangi: Hmm. I think leaving it as a vector should be fine. `std::vector<char>` isn't verbose and is being used 1 or 2 times per file. 

I can't define the buffer as const since we are casting it to `xclmgmt_ioc_bitstream_axlf` in L 521

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
AShivangi: Hehe :P

src/runtime_src/core/pcie/linux/device_linux.cpp
AShivangi: Moved. I'm not sure how to do a binary check against a string...

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
AShivangi: Done

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
AShivangi: What would be a more helpful error in your opinion? `v` is just the xclbin magic string, i.e., `xclbin2` so even if it produces garbage, it will only print a few beginning characters which might be helpful

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
AShivangi: Yes. Thanks :)

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
AShivangi: done

src/runtime_src/core/pcie/linux/device_linux.cpp
rozumx: Hmm, that last time I checked, this buffer is not being modified, am I mistakened?

If not, then this buffer should be a const and xclbmgmt_ioc_bistream_axlf should be updated to take a const value.   

I understand that that might be be possible to update ALL of the locations in a timely manner, as such, it might be best to use a const_cast() of the buffer when calling this method.

As for the std::vector<char> it would be better it is was aliased.   You can consider this as part of the C++ coding standard that we have started to adopt.  In this case, the alias is used to simplify notation and hide implementation details.

src/runtime_src/core/pcie/linux/device_linux.cpp
rozumx: Then don't use a string :^).  The underlying issue is that "by chance" the magic xclbin image value is "xclbin2", which is all ascii, imaging instead it is "0x78636c62696e32"  How would you validate these value?

Fortunately, there are memory functions that can help (e.g., std::memcmp).

**_Note_**
This code does work in this case, but it is something that I would recommend against.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
rozumx: Also, what happens if the file to be read in is only 2 bytes in size.  In this case, you will be going beyond the buffer and the system "could" crash.  And yes, coverity will catch this issue.

xilinxgitops:
retest this please

rozumx (approved):

AShivangi:
retest this please

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -730,6 +733,33 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
...
+    std::ifstream stream(xclbin, std::ios::binary);
+    if (!stream)
+      throw xrt_core::error(boost::str(boost::format("Could not open %s for reading") % xclbin));
+
+    stream.seekg(0,stream.end);
+    ssize_t size = stream.tellg();
+    stream.seekg(0,stream.beg);
+
+    std::vector<char> xclbin_buffer(size);
+    stream.read(xclbin_buffer.data(), size);
```
stsoe, line 753: Please use xrt::xclbin to read the xclbin.
Let's rename xrt_core::load_xclbin() to reflect different loading different from the 'normal' xclbin loading.

src/runtime_src/core/common/device.h
```diff
@@ -292,6 +293,8 @@ class device : public ishim
 
   virtual void reset(query::reset_type&) const {}
 
+  virtual void load_xclbin(const MemoryBuffer &) const{}
```
stsoe, line 296: This is not commented.  What is this?   xclbin loading is done with xrt::xclbin objects after 4745 pull request.  It looks like this PR is bypassing that.    This function may be fine, but it is confusing that the function seems to be loading xclbins but in a different way than other code.  What does it mean to load an xclbin with this function?   Should the function reflect that loading is via xclmgmt servicing something different?

Should the default behavior be silence?  What does that mean?  Is an exception better?

It looks like std:;vector<char> is not  really used, so it should be const char* and caller should call the function with vector.data().

PR #4765, opened 2021-02-21
xocl: issue xbutil reset only if device is in ready state (#4725)

rajkumar-xilinx:
Issue xbutil reset command only when device is in ready state. Also, read return status of xocl peer requests and act accordingly.


rajkumar-xilinx:
retest this please

xilinxgitops:
retest this please

PR #4529, opened 2020-12-04
get soft kernel return code

sarab96:
get soft kernel return code to host

PR #4471, opened 2020-11-16
XRT-812 command queue interrupt support on ps for u30

xuhz:
tested on u30 and u250 xdma 3.1, MB/PS cq can change between polling and interrupt mode.
also ran stress and performance comparison test on u30

PR #4767, opened 2021-02-21
Porting recent U30 changes to new KDS

mamin506:
Changes:
1. Support xbutil query PS kernel with new KDS
2. Host to PS CQ interrupt (porting #4471 )
3. Support start PS kernel command return code (#4529  & #4540 )
4. Construct and send PS kernel config command to ERT in xclbin download.

@AShivangi  please review xbutil.h changes.
@larry9523 please take a look and provide feedback for all of the changes.
@chienwei-lan please review ert_user.c changes.

### DIFF
--- a/src/runtime_src/core/common/drv/kds_core.c
+++ b/src/runtime_src/core/common/drv/kds_core.c
@@ -66,11 +66,6 @@ ssize_t show_kds_custat_raw(struct kds_sched *kds, char *buf)
 	}
 	mutex_unlock(&cu_mgmt->lock);
 
-	if (sz < PAGE_SIZE - 1)
-		buf[sz++] = 0;
-	else
-		buf[PAGE_SIZE - 1] = 0;
-
 	return sz;
 }
 
@@ -102,11 +97,6 @@ ssize_t show_kds_scustat_raw(struct kds_sched *kds, char *buf)
 	}
 	mutex_unlock(&scu_mgmt->lock);
 
-	if (sz < PAGE_SIZE - 1)
-		buf[sz++] = 0;
-	else
-		buf[PAGE_SIZE - 1] = 0;
-
 	return sz;
 }
 
@@ -125,8 +115,6 @@ ssize_t show_kds_stat(struct kds_sched *kds, char *buf)
 			kds->cu_intr_cap);
 	sz += scnprintf(buf+sz, PAGE_SIZE - sz, "Interrupt mode: %s\n",
 			(kds->cu_intr)? "cu" : "ert");
-	sz += scnprintf(buf+sz, PAGE_SIZE - sz, "Configured: %d\n",
-			cu_mgmt->configured);
 	sz += scnprintf(buf+sz, PAGE_SIZE - sz, "Number of CUs: %d\n",
 			cu_mgmt->num_cus);
 	for (i = 0; i < cu_mgmt->num_cus; ++i) {
@@ -137,11 +125,6 @@ ssize_t show_kds_stat(struct kds_sched *kds, char *buf)
 				(cu_mgmt->cu_intr[i])? "enable" : "disable");
 	}
 	mutex_unlock(&cu_mgmt->lock);
-
-	if (sz < PAGE_SIZE - 1)
-		buf[sz++] = 0;
-	else
-		buf[PAGE_SIZE - 1] = 0;
 
 	return sz;
 }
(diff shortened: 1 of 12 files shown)


src/runtime_src/core/common/drv/kds_core.c
```diff
@@ -74,6 +74,42 @@ ssize_t show_kds_custat_raw(struct kds_sched *kds, char *buf)
...
+	 * So, this separate kds_scustat_raw is better.
+	 *
+	 * But in the worst case, this is still not good enough.
+	 */
+	mutex_lock(&scu_mgmt->lock);
+	for (i = 0; i < scu_mgmt->num_cus; ++i) {
+		sz += scnprintf(buf+sz, PAGE_SIZE - sz, cu_fmt, i,
+				scu_mgmt->name[i], scu_mgmt->status[i],
+				scu_mgmt->usage[i]);
+	}
```
larry9523, line 102: If we overflow, the memory will already be corrupted, right?

src/runtime_src/core/common/drv/kds_core.c
```diff
@@ -159,6 +195,39 @@ kds_cu_config(struct kds_cu_mgmt *cu_mgmt, struct kds_command *xcmd)
...
+	int i, j;
+
+	scmd = (struct ert_configure_sk_cmd *)xcmd->execbuf;
+	for (i = 0; i < scmd->num_image; i++) {
+		struct config_sk_image *cp = &scmd->image[i];
+
+		for (j = 0; j < cp->num_cus; j++) {
+			int scu_idx = j + cp->start_cuidx;
+
+			/* TODO: Why continue? */
```
larry9523, line 211: Historical reason. Config PS kernel ert command was not atomic. There could be multiple configs from different processes. So we continue if the name has been cached. But I believe it can be removed now.

src/runtime_src/core/pcie/tools/xbutil/xbutil.h
```diff
@@ -520,6 +526,40 @@ class device {
...
+                std::cout << "WARNING: 'kds_scustat_raw' has no expect tokens, stop parsing.\n";
+                break;
+            }
+
+            tokenizer::iterator tok_it = tokens.begin();
+            cu_idx = std::stoi(std::string(*tok_it++));
+            name = std::string(*tok_it++);
+            status = std::stoul(std::string(*tok_it++), nullptr, radix);
+            usage = std::stoul(std::string(*tok_it++));
+            // TODO: Let's avoid this special handling for PS kernel name
```
larry9523, line 552: Once we have PS kernel ip_layout section in xclbin.

larry9523 (changes requested):
Seems to me, there is potential memory corruption. Could you please confirm?

src/runtime_src/core/common/drv/kds_core.c
mamin506: Ok. I would remove this in next commit which send PS kernel config command to ERT in xclbin download. This can make sure it is atomic.

src/runtime_src/core/common/drv/kds_core.c
mamin506: Buf would not over flow. The second argument of scnprintf() is the maximum bytes can be written.
Use scnprintf() is a very simple implementation of a device attribute (see https://github.com/torvalds/linux/blob/master/Documentation/filesystems/sysfs.rst)

BTW, this code cannot show more than PAGE_SIZE. This solution doesn't work well when we have 128 long name PS kernels and with large usage value.

larry9523: > Buf would not over flow. The second argument of scnprintf() is the maximum bytes can be written.
> Use scnprintf() is a very simple implementation of a device attribute (see https://github.com/torvalds/linux/blob/master/Documentation/filesystems/sysfs.rst)
> 
> BTW, this code cannot show more than PAGE_SIZE. This solution doesn't work well when we have 128 long name PS kernels and with large usage value.

OK. Sounds good.

larry9523 (approved):

xilinxgitops:
retest this please

maxzhen (approved):

AShivangi (approved):
Looks good. Approving this for now but these changes need to be ported to xbutil2 as well.

chienwei-lan (approved):

PR #4768, opened 2021-02-22
ERT bist microblaze and driver part

chienwei-lan:


src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
```diff
@@ -233,12 +235,98 @@ static ssize_t ert_intr_store(struct device *dev,
...
+	if (kstrtou32(buf, 10, &go_sleep) == -EINVAL || go_sleep > 2) {
+		xocl_err(&to_platform_device(dev)->dev,
+			"usage: echo 0 or 1 > mb_sleep");
+		return -EINVAL;
+	}
+
+	if (go_sleep)
+		ert_user_gpio_cfg(pdev, MB_SLEEP);
+	else
+		ert_user_gpio_cfg(pdev, MB_WAKEUP);
```
mamin506, line 253: Do we need protect here? If this node is only for internal debug purpose, then without lock is okay.
Otherwise, we needs to deal with concurrently write.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
```diff
@@ -420,15 +508,13 @@ ert_release_slot(struct xocl_ert_user *ert_user, struct ert_user_command *ecmd)
 static void
 ert_post_process(struct xocl_ert_user *ert_user, struct ert_user_command *ecmd)
 {
-	struct ert_mb_validate_cmd cmd = {0}; 
-
-	if (unlikely(ert_special_cmd(ecmd)))
+	if (likely(!ert_special_cmd(ecmd)))
```
mamin506, line 511: In ert_post_process(), you copy some info from CQ slot 0 to host. But it is called in process_ert_cq(). My concern is this logic required that process_ert_sq() doesn't reuse slot 0 for other command. Right?

mamin506 (changes requested):
Please see my comments. This PR would has some conflicts with #4767, one of us need to resolve it ;)

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
chienwei-lan: post_process() called before we release the slot, at that moment the slot 0 is marked as busy. 

So no one can use slot 0 before this command processed and slot been released. 

And we only have one thread to process ert_process_cq and ert_process_rq, so before we have done processing all the cmd in cq, no body can cut the line to submit another slot 0 command

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
chienwei-lan: I added lock in ert_user_gpio_cfg to protect this section

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
mamin506: I see. Sounds good.

mamin506 (approved):

PR #4698, opened 2021-02-05
XRT-845 New kds stress test

mamin506:
Use this PR to fix new kds, use kds mode, stress test issues.

PR #4769, opened 2021-02-22
[Edge] remove lock which leads to deadlock

mamin506:
This lock is added in #4698 without testing on edge platform. 
Himanshu points out that zocl_context_ioctl() -> zocl_lock_bitstream() -> acquire the same lock.

After review this part, it looks like just simply remove the lock here is enough.
The reason is commented in the code.


chvamshi-xilinx (approved):

Himanshu-xilinx (approved):
approved

PR #4770, opened 2021-02-22
CR-1089519 configure clock freqs for flat shell also

rajkumar-xilinx:
Gating logic is not available on the flat shells. So, do not call axigate freeze/free apis for these shells.
Updating clock freqs without calling axigate apis.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/clock.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/clock.c
@@ -875,7 +875,10 @@ done:
  *   4) wait for hbm calibration done
  *
  * Note:
- * Violate this flow will cause random firewall trip.
+ * 1) Violate this flow will cause random firewall trip.
+ * 2) Flat shell: gating logic is not available on these shells.
+ *    - So, do not call axigate freeze/free apis for these shells.
+ *    - Updating clock freqs without calling axigate apis.
  */
 static int clock_ocl_freqscaling(struct clock *clock, bool force, int level,
                                  bool freeze)
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/clock.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/clock.c
@@ -877,7 +877,8 @@ done:
  * Note:
  * Violate this flow will cause random firewall trip.
  */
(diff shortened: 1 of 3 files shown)


chvamshi-xilinx (approved):
Approving as it is going to affect only flat shells and with an assumption that gating is not required to update clocks.We may need to revisit this check in future. 
U,2 is the only flat shell today and they confirmed behavior from their regressions tests.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/clock.c
```diff
@@ -886,10 +890,12 @@ static int clock_ocl_freqscaling(struct clock *clock, bool force, int level)
 	for (i = 0; i < CLOCK_MAX_NUM_CLOCKS; i++)
 		curr[i] = clock_get_freq_impl(clock, i);
 
-	err = clock_freeze_axi_gate(clock, level);
+	if (freeze)
```
houlz0507, line 893: please check the return value. if err==-ENODEV, means there is not axigate device. Then you do not have to add 'freeze' parameter.

maxzhen (approved):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/clock.c
chvamshi-xilinx: Hi @houlz0507 , we thought of adding that check, but its going to affect platforms which doesn't have axi-gate (ideally there should not be any platform). But, as we are at edge of PU1, just wanted to make sure we are not touching any other flows.

houlz0507: ok. Please consider this in master branch.

houlz0507 (approved):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/clock.c
chvamshi-xilinx: Yes @houlz0507 , we should make that change in master branch. Thanks

rajkumar-xilinx: I have taken care of this comment in master PR https://github.com/Xilinx/XRT/pull/4785

PR #4771, opened 2021-02-22
CR 1087523 : Organize Read, Write Channel for each AXI MM connection for a CU in timeline trace

IshitaGhosh:


jvillarre (approved):
Changes look good and the generated CSV looks cleaner.

PR #4774, opened 2021-02-22
Move ps_kernel.h out of driver specific header folder

larry9523:
xbutil and xbutil2 is generic user space code and can not include driver specific header directory.

This PR moves ps_kernel.h to core/include from core/pcie/driver/linux/include so that xbutil/xbutil2 can use.


AShivangi (approved):
Thanks for making these changes :)

stsoe (approved):

PR #4775, opened 2021-02-23
CR-1073379 CMC / XRT Interface should be able to detect if CMC sensor…

xdavidz:
… updates have stopped - XRT


xdavidz:
Tested on U50. This will print warning when CMC heartbeat count stops increasing with error code reason. See https://jira.xilinx.com/browse/CR-1073379 for more info.

maxzhen (approved):

houlz0507 (approved):

PR #4776, opened 2021-02-23
Added a xbutil validate test case for the VCU softkernel validation for decoder.

saifuddin-xilinx:
Added a xbutil validate test case for the VCU softkernel validation for decoder.
The test application will run with XRT API.
Please find the details of the application:

Sending a single frame as an input buffer array.
Input buffer is hardcoded into a header file as a hex array
Expected(Golden) output is also saved on the header file.
Hard-coded gstreamer specific filed i.e. video information based on the input frame we are using for this test.
Verify the output for correctness
Code is placed under XRT/tests/xrt/vcu_test directory.


saifuddin-xilinx:
Will create a new one

PR #4777, opened 2021-02-23
CR-1089476 RTL kernel with k2k streaming hung right after loading xclbin on board for 3 platforms

mamin506:
Not sure why memcpy() sometime would not be able to copy all of the bytes to CU. All of the input argument is as expected.

Need more efforts to investigate why memcpy() is not working properly.

For now, I have to use for loop to copy data to CU.


maxzhen:
can we use memcpy to copy the first aligned portion of arguments, then copy the last word separately? so that in most cases, just memcpy is needed.

mamin506:
> can we use memcpy to copy the first aligned portion of arguments, then copy the last word separately? so that in most cases, just memcpy is needed.

I tried that as well. But the issue is I don't know what to 'align'.
In my test, the behavior can change even use same value but one is a variable and one is constant.

chienwei-lan:
This change requires roughly 2% more cycles to complete, compare to memcpy

chienwei-lan (approved):

PR #4782, opened 2021-02-23
coverity fixes,event ID atomic, finer grained dynamic locks

hackwa:


jvillarre (approved):
Fixes to the locking and other small Coverity issue fixes look good

PR #4783, opened 2021-02-23
xbutil2 bist second half

chienwei-lan:


src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -1031,14 +1091,21 @@ pretty_print_test_run(const boost::property_tree::ptree& test,
...
+  std::string prev_tag = "";
   auto color = EscapeCodes::FGC_PASS;
   bool warn = false;
   bool error = false;
 
   try {
     for (const auto& dict : test.get_child("log")) {
       for (const auto& kv : dict.second) {
-        _ostream<< boost::format("    %-24s: %s\n") % kv.first % kv.second.get_value<std::string>();
+        if (kv.first.compare(prev_tag)) {
```
AShivangi, line 1102: Thanks :)

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -966,14 +1004,36 @@ hostMemBandwidthKernelTest(const std::shared_ptr<xrt_core::device>& _dev, boost:
 void
 bistTest(const std::shared_ptr<xrt_core::device>& _dev, boost::property_tree::ptree& _ptTest)
 {
+  /* We can only test ert validate on SSv3 platform, skip if it's not a SSv3 platform */
+  int32_t ert_cfg_gpio = 0;
+  try {
+   ert_cfg_gpio = xrt_core::device_query<xrt_core::query::ert_sleep>(_dev);
+  } catch(...) {
+      logger(_ptTest, "Warning", "ERT validate is not available");
```
AShivangi, line 1012: Should this be a warning or simply a detail? Like in P2P

src/runtime_src/core/common/query_requests.h
```diff
@@ -200,9 +200,15 @@ enum class key_type
   aie_metadata,
   graph_status,
   mailbox_metrics,
+
   clock_timestamp,
-  noop
+  ert_sleep,
+  ert_cq_write,
```
AShivangi, line 206: What does cq stand for?

AShivangi (approved):
The code is very clean and easy to read!

src/runtime_src/core/common/query_requests.h
chienwei-lan: cq means command queue. 

We measure the clock cycle while read/write command queue and CU

AShivangi: Got it

PR #4789, opened 2021-02-24
CR-1087576 corrected soft kernel name in xbutil query

sarab96:
corrected soft kernel name in xbutil query

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
@@ -2731,8 +2731,12 @@ static int icap_cache_ps_kernel_axlf_section(const struct axlf *xclbin,
 		    (struct soft_kernel *)&blob[header->m_sectionOffset];
 		char *begin = (char *)sp;
 
+                /* save mpo_name for use with xbutil query cmd
+                 * mpo_name is unique to a xclbin
+                 * symbol name is for use inside skd
+                 */
 		strncpy(pnode->pkn_data[idx].pkd_sym_name,
-		    begin + sp->mpo_symbol_name,
+		    begin + sp->mpo_name,
 		    PS_KERNEL_NAME_LENGTH - 1);
 		pnode->pkn_data[idx].pkd_num_instances =
 		    sp->m_num_instances;
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
@@ -2736,6 +2736,9 @@ static int icap_cache_ps_kernel_axlf_section(const struct axlf *xclbin,
                  * symbol name is for use inside skd
                  */
 		strncpy(pnode->pkn_data[idx].pkd_sym_name,
+		    begin + sp->mpo_symbol_name,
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
```diff
@@ -2731,8 +2731,12 @@ static int icap_cache_ps_kernel_axlf_section(const struct axlf *xclbin,
 		    (struct soft_kernel *)&blob[header->m_sectionOffset];
 		char *begin = (char *)sp;
 
+                /* save mpo_name for use with xbutil query cmd
+                 * mpo_name is unique to a xclbin
+                 * symbol name is for use inside skd
+                 */
 		strncpy(pnode->pkn_data[idx].pkd_sym_name,
-		    begin + sp->mpo_symbol_name,
+		    begin + sp->mpo_name,
```
larry9523, line 2739: Hmm, I am not sure what name we should display here. To me, symbol_name is the one that matters. It is OK if mpo_name is wrong but if symbol_name is wrong, the PS kernel can not be launched.

larry9523 (commented):
While we can discuss what name should be displayed here, it might be a good idea to put both mpo_name and mpo_symbol_name together to the sysfs node. So that user utility can consume whatever it wants

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
sarab96: symbol name can be same like "process_frame" for decoder as well as encoder kernels. mpo_name is unique per xclbin. If symbol name is incorrect then xclbin is wrong and load_xclbin will fail

sarab96:
done
> While we can discuss what name should be displayed here, it might be a good idea to put both mpo_name and mpo_symbol_name together to the sysfs node. So that user utility can consume whatever it wants

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
larry9523: I am not convinced we should use name instead of symbol_name (which is actually PS kernel function name).

As we evolve to v++ flow, PS kernel name (which will be displayed in xbutil) will be from ip_layout alike section. And the source of PS kernel name will be the function name. 

Take PL kernel as an example. Today, if we have a PL kernel.cl
__attribute__ ((reqd_work_group_size(1, 1, 1)))                                                          |clean:
kernel void simple(global int *restrict s1,                                                              |        -$(RM) $(EMCONFIG_FILE) $(HOST_EXE)  $(XCLBIN_1)   $(XOS_1)
                   global const int *s2,                                                                 |
                   int len) 
              
  And in make file we have 
  v++ -k simple
  
So eventually we will use the function name as the kernel name and then displayed to xbutil query.

larry9523: Also this is already the case in Alpha release. We cache the name in scu config command. That name is symbol name but not name.

xilinxgitops:
retest this please

sarab96:
Change of plan. Will make requirement for symbol name to be unique in xclbin

PR #4795, opened 2021-02-25
Do not use memcpy to configure CU in ERT

mamin506:
See the bug we fixed in #4777 
We have to accept about 2% performance lost.


xilinxgitops:
retest this please

xilinxgitops:
retest this please

chienwei-lan (approved):

PR #4801, opened 2021-02-25
CRs and qspi report for U30

AShivangi:
### Work Done:
- **CR-1088239** ASTeR tc_09_01_to_09_35_low_power_query_sensor - xbutil2: sensors not reported
- **CR-1089315** ASTeR TS21 - U30 - 'xbmgmt --new advanced --config --showx' does not show status correctly
- Boost process to use python3 instead of python (which could be pointing to puthon2)
- QSPI write protection status report in xbutil2

### DIFF
--- a/src/runtime_src/core/pcie/linux/device_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/device_linux.cpp
@@ -81,6 +81,44 @@ struct kds_cu_info
   }
 };
 
+/**
+ * qspi write protection status
+ * byte 0:
+ *   0: status not available, 1: status available
+ * byte 1 primary qspi(if status available):
+ *   1: write protect enable, 2: write protect disable
+ * byte 2 recovery qspi(if status available):
+ *   1: write protect enable, 2: write protect disable
+ */
+struct qspi_status
+{
+  using result_type = query::xmc_qspi_status::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type)
+  {
+    auto pdev = get_pcidev(device);
+  
+    std::string status_str, errmsg;
+    pdev->sysfs_get("xmc", "xmc_qspi_status", errmsg, status_str);
+    if (!errmsg.empty())
+      throw std::runtime_error(errmsg);
+
+    std::string primary, recovery;
+    for (auto x : status_str) {
+      int status_byte = x - '0';
+      if(status_byte == 0)
+        return std::pair<std::string, std::string>("N/A", "N/A");
+
+      if(primary.empty())
+        primary = status_byte == 1 ? "Enabled": status_byte == 2 ? "Disabled" : "Invalid";
+      else
+        recovery = status_byte == 1 ? "Enabled": status_byte == 2 ? "Disabled" : "Invalid";
+    }
+    return std::pair<std::string, std::string>(primary, recovery);
+  }
+};
+
 struct mac_addr_list
 {
   using result_type = query::mac_addr_list::result_type;
@@ -388,6 +426,7 @@ initialize_query_table()
   emplace_sysfs_get<query::mac_contiguous_num>                 ("xmc", "mac_contiguous_num");
   emplace_sysfs_get<query::mac_addr_first>                     ("xmc", "mac_addr_first");
   emplace_sysfs_get<query::oem_id>                             ("xmc", "xmc_oem_id");
+  emplace_func0_request<query::xmc_qspi_status,                qspi_status>();
   emplace_func0_request<query::mac_addr_list,                  mac_addr_list>();
 
   emplace_sysfs_get<query::firewall_detect_level>             ("firewall", "detected_level");
--- a/src/runtime_src/core/pcie/linux/device_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/device_linux.cpp
@@ -84,11 +84,11 @@ struct kds_cu_info
 /**
  * qspi write protection status
  * byte 0:
- *   0: status not available, 1: status available
+ *   '0': status not available, '1': status available
  * byte 1 primary qspi(if status available):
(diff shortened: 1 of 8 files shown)


xilinxgitops:
retest this please

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -81,6 +81,44 @@ struct kds_cu_info
   }
 };
 
+/**
+ * qspi write protection status
+ * byte 0:
+ *   0: status not available, 1: status available
```
rozumx, line 87: When I first saw this document, I thought the value was 0x00 and 0x01.  Looking at the code shows that these values are actually '0' and '1'.

To help others, please update the comment to reflex characters.  For example:
`'0': status not available, '1': status available`

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -81,6 +81,44 @@ struct kds_cu_info
...
+    auto pdev = get_pcidev(device);
+  
+    std::string status_str, errmsg;
+    pdev->sysfs_get("xmc", "xmc_qspi_status", errmsg, status_str);
+    if (!errmsg.empty())
+      throw std::runtime_error(errmsg);
+
+    std::string primary, recovery;
+    for (auto x : status_str) {
+      int status_byte = x - '0';
```
rozumx, line 109: What happens if 'x' is less then '0' (48) or greater then '9' (57).  

Wouldn't it be better if this code simply use the actual characters for comparison?  For example:

```
 if(status_byte == '0')
        return std::pair<std::string, std::string>("N/A", "N/A");

      if(primary.empty())
        primary = status_byte == '1' ? "Enabled": status_byte == '2' ? "Disabled" : "Invalid";
      else
        recovery = status_byte == '1' ? "Enabled": status_byte == '2' ? "Disabled" : "Invalid"
```

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
```diff
@@ -142,37 +142,33 @@ show_device_conf(xrt_core::device* device)
...
 
   try {
-    std::cout << "\tRuntime clock scaling enabled status: ";
     auto scaling_enabled = xrt_core::device_query<xrt_core::query::xmc_scaling_enabled>(device);
-    std::cout << scaling_enabled << "\n";
+    std::cout << boost::format("  %-33s: %s\n") % "Runtime clock scaling enabled" % scaling_enabled;
   }
   catch (const std::exception& ex) {
-    std::cout << ex.what() << "\n";
+    //safe to ignore. These sysfs nodes are not present for u30
```
rozumx, line 157: Hmm...this comment suggests that the code is U30 aware.  

Is this safe to ignore for ALL cards?

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
```diff
@@ -142,37 +142,33 @@ show_device_conf(xrt_core::device* device)
...
 
   try {
-    std::cout << "\tScaling threshold power override: ";
     auto scaling_override = xrt_core::device_query<xrt_core::query::xmc_scaling_override>(device);
-    std::cout << scaling_override << "\n";
+    std::cout << boost::format("  %-33s: %s\n") % "Scaling threshold power override" % scaling_override;
   }
   catch (const std::exception& ex) {
-    std::cout << ex.what() << "\n";
+    //safe to ignore. These sysfs nodes are not present for u30
```
rozumx, line 165: Hmm... See earlier comment.

rozumx (changes requested):
Some minor changes requested.

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
AShivangi: @maxzhen, can you chime in on this? Is there a scenario where the card requires "xmc_scaling_enabled" sysfs node but the driver fails to populate it? Is it safe to ignore this catch?

src/runtime_src/core/pcie/linux/device_linux.cpp
AShivangi: Ah yes. This is so much better! Thank you

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
maxzhen: in that case, it's a driver bug and should be fixed by driver. you should only ignore the case when you don't find this sysfs node. if you find it but can't do anything IO on it, then that is an error that you should not ignore.

rozumx: If I understand correctly, if the catch is tripped is because of a driver bug.   So in this case, we should at least report the exception.

maxzhen: can we differentiate if the error is because the sysfs node is not found or something else? it is not a bug if the node is not found and we should ignore it.

AShivangi: No, we only throw one kind of exception. We can't differentiate it

maxzhen: Then we have to ignore it.

AShivangi:
retest this please

AShivangi:
u280 didn't come back after reboot

src/runtime_src/core/tools/xbmgmt2/OO_Config.cpp
rozumx: Bummer :^(

rozumx (approved):
Approved.  Thanks for making these changes.

PR #4802, opened 2021-02-25
reset is_configured flag properly

chienwei-lan:
We should reset the is_configured flag if the new config command coming

and set it as true once the config command is completed

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
@@ -647,9 +647,6 @@ static inline void process_ert_sq(struct xocl_ert_user *ert_user)
 	if (!ert_user->num_sq)
 		return;
 
-	if (ert_user->polling_mode)
-		return;
-
 	ev_client = first_event_client_or_null(ert_user);
 
 	list_for_each_entry_safe(ecmd, next, &ert_user->sq, list) {
@@ -758,9 +755,6 @@ static inline void process_ert_sq_polling(struct xocl_ert_user *ert_user)
 	unsigned int tick;
 
 	if (!ert_user->num_sq)
-		return;
-
-	if (!ert_user->polling_mode)
 		return;
 
 	for (section_idx = 0; section_idx < 4; ++section_idx) {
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
```diff
@@ -1009,12 +1032,10 @@ static inline int process_ert_rq(struct xocl_ert_user *ert_user)
...
-		if (cmd_opcode(ecmd) == OP_CONFIG && !ert_user->polling_mode) {
-			for (i = 0; i < ert_user->num_slots; i++) {
-				xocl_intc_ert_request(xdev, i, ert_user_isr, ert_user);
-				xocl_intc_ert_config(xdev, i, true);
-			}
-
+		if (cmd_opcode(ecmd) == OP_CONFIG) {
+			ert_user->is_configured = false;
+			if (!ert_user->polling_mode)
+				ert_intc_enable(ert_user, true);
```
maxzhen, line 1038: intr is disabled in _remove routine, why intr is not enabled in _probe()?

xilinxgitops:
retest this please

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
chienwei-lan: We enable interrupt and register related callback function if polling_mode = true. This flag is as part of OP_CONFIG command which  is given by xrt.ini for now and kds in the future. 

So we should disable and unregister in the same path.

And we also have to clean it up in "ert_user_exit" function

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
```diff
@@ -998,12 +1019,12 @@ static inline int process_ert_rq(struct xocl_ert_user *ert_user)
...
-				xocl_intc_ert_request(xdev, i, ert_user_isr, ert_user);
-				xocl_intc_ert_config(xdev, i, true);
-			}
-
+		if (cmd_opcode(ecmd) == OP_CONFIG) {
+			ert_user->is_configured = false;
+			if (ert_user->polling_mode)
+				ert_intc_enable(ert_user, false);
+			else
+				ert_intc_enable(ert_user, true);
```
maxzhen, line 1027: why these are not done in ert_post_process() in CONFIG op code case?

chienwei-lan: it could be, the configuration should be activated once the config cmd is completed and we should change the interrupt and it's rountine setting at the point.

chienwei-lan: config the interrupt and isr in post_process => we use the previous config cmd setting to see if the config cmd is completed or not, even it might change

config the interrupt and isr in process_rq => The setting affects immediately

maxzhen (approved):

mamin506 (approved):

PR #4803, opened 2021-02-26
do NOT use hardcode cu index, get index from IP name

mamin506:
Use hard coding cu index is not a good idea. If verify.xclbin change, this would lead to unexpected behavior.


maxzhen (approved):

xilinxgitops:
retest this please

PR #4815, opened 2021-03-01
Incoming ert command inspection

chienwei-lan:


src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
```diff
@@ -976,18 +1003,14 @@ static inline int process_ert_rq(struct xocl_ert_user *ert_user)
...
-				ecmd->status = KDS_ABORT;
-				list_move_tail(&ecmd->list, &ert_user->cq);
-				--ert_user->num_rq;
-				++ert_user->num_cq;
-				continue;
-			}
-		} else if (cmd_opcode(ecmd) == OP_START)
-			BUG_ON(ert_user->ctrl_busy || !ert_user->config);
+		if (ert_cmd_inspection(ert_user, ecmd)) {
+			ERTUSER_ERR(ert_user, "%s bad cmd, opcode: %d\n", __func__, cmd_opcode(ecmd));
```
mamin506, line 1007: Not sure if this should be a debug message.

mamin506 (approved):
Overall, this looks good to me.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
chienwei-lan: once there is any cmd its opcode we can't recognize, we should have some clues in dmesg

or does kds already report something?

mamin506: KDS doesn't report something. I think we can print error here. This is okay.

PR #4816, opened 2021-03-01
CR 1090162 : Device trace layout improvements

IshitaGhosh:
Remove CU name from AXI MM connections, Use stream connection name in section header for AXI Stream connection


jvillarre (approved):
Changes look good

PR #4818, opened 2021-03-01
continuous trace enhancements

hackwa:
Unique trace ID for each writer
No trace file is created when there's no trace available in ocl, lop, device plugins
Turn off lop trace when opencl trace is enabled


jvillarre (approved):
Changes look good

PR #4821, opened 2021-03-01
Don't run bist test with the xbutil validate "all" testSuite

AShivangi:
Bist test is for internal use only, hence it shouldn't be run when `xbutil validate -r all` is run. Introducing `is_internal` flag to determine which user-facing tests should be run for the customers.


maxzhen:
BIST should be run by user explicitly. It should never be run as part of some other activities. I'm not sure if this is what we're doing here?

AShivangi:
Yes, Earlier it got added as a part of "xbutil validate" so it would run whenever anyone ran the full xbutil validate. This PR is to extract bist test from the test pool. The user will have to run bist manually now

chienwei-lan:
> BIST should be run by user explicitly. It should never be run as part of some other activities. I'm not sure if this is what we're doing here?

BIST should only be run if user call "xbutil2 validate --run bist" explicitly,

rozumx:
Hmm...I understand what we are trying to do here.  Unfortunately, using an internal flag, while addressing the issue doesn't scale well.  Shrivangi, lets chat about this tomorrow to see how we might better classify this test along with how this solution can be scaled.

chienwei-lan (approved):

rozumx (approved):

PR #4822, opened 2021-03-01
Exclude unsupported device: u30

haeseunglee:
No source code change. Just metadata update for 100_ert to exclude u30


maxzhen:
ok to test

PR #4828, opened 2021-03-02
CR-1087396 fix for xclLog tag

sarab96:
log message in PS side were not printing properly as tag was passed in api calls but tag was missing from function signature
messages were truncated as shown below (in PS):
..
[Fri Oct 6 15:52:55 2102 GMT] [XRT] Tid: 548090720272, DEBUG: XR
[Fri Oct 6 15:52:55 2102 GMT] [XRT] Tid: 548090720272, INFO: XR
[Fri Oct 6 15:52:55 2102 GMT] [XRT] Tid: 548090720272, INFO: XR

### DIFF
--- a/src/runtime_src/core/edge/user/shim.cpp
+++ b/src/runtime_src/core/edge/user/shim.cpp
@@ -57,10 +57,10 @@ namespace {
 
 template <typename ...Args>
 void
-xclLog(xrtLogMsgLevel level, const char* format, Args&&... args)
+xclLog(xrtLogMsgLevel level, const char* tag, const char* format, Args&&... args)
 {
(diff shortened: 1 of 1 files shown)


stsoe (changes requested):
I see this problem for sure.  But couldn't we change the usage of xclLog to not pass the tag?  The tag is "XRT" always after all, so just remove "XRT" from all places that call this xclLog function and let xclLog set the tag as it currently does.

sarab96:
Done
> I see this problem for sure. But couldn't we change the usage of xclLog to not pass the tag? The tag is "XRT" always after all, so just remove "XRT" from all places that call this xclLog function and let xclLog set the tag as it currently does.

stsoe (approved):
Thanks!

PR #4829, opened 2021-03-02
removed internal options from xbutil2/xbmgmt2

uday610:


keryell:
Does it mean there will be 2 versions of XRT flying around or just that the options are deprecated?

PR #4830, opened 2021-03-02
Metadata update only to disable 56_xclbin from sprite

haeseunglee:
Metadata update only. No source code change. 
This will disable 56_xclbin from sprite run


maxzhen:
ok to test

PR #4832, opened 2021-03-02
CR-1089903 fix for JVM issue

sarab96:
fix for JVM issue
signal handlers disabled


stsoe (changes requested):
Can we remove the dead code rather than commenting it out?   The commit history preserves old deleted code.  The description of this commit should be more descriptive as to what the problem was that lead to removal of the signal handlers.

sarab96:
Updated.
JVM requires signals to be detected by JVM. Application was seen to hang without this fix

stsoe (approved):

PR #4835, opened 2021-03-03
Fix MPD crash when reading DRM major:minor number

pranjalv-xilinx:
Backport to 2020.2


xuhz (commented):
why do you remove the empty and file existing check?

PR #4840, opened 2021-03-03
CR-1089906 throw for unexpected cmd state

sarab96:
throw for unexpected cmd state
Else log sees too many error messages and XMA plugin gets timeout error and may keep retrying

### DIFF
--- a/src/xma/src/xmaapi/xma_utils.cpp
+++ b/src/xma/src/xmaapi/xma_utils.cpp
@@ -569,6 +569,8 @@ int32_t check_all_execbo(XmaSession s_handle) {
                     	       cu_cmd->state >= ERT_CMD_STATE_NORESPONSE) {
                         //Check for invalid/error execo state
                         xma_logmsg(XMA_ERROR_LOG, XMAUTILS_MOD, "Session id: %d, type: %s, Unexpected ERT_CMD_STATE. state=%s", s_handle.session_id, xma_core::get_session_name(s_handle.session_type).c_str(), xma_core::get_cu_cmd_state(cu_cmd).c_str());
+                        std::cout << "XMA FATAL: Unexpected error ERT_CMD_STATE. Cmd state: " << xma_core::get_cu_cmd_state(cu_cmd) << std::endl;
+                        throw std::runtime_error("Unexpected ERT Cmd State");
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


src/xma/src/xmaapi/xma_utils.cpp
```diff
@@ -569,6 +569,8 @@ int32_t check_all_execbo(XmaSession s_handle) {
                     	       cu_cmd->state >= ERT_CMD_STATE_NORESPONSE) {
                         //Check for invalid/error execo state
                         xma_logmsg(XMA_ERROR_LOG, XMAUTILS_MOD, "Session id: %d, type: %s, Unexpected ERT_CMD_STATE. state=%s", s_handle.session_id, xma_core::get_session_name(s_handle.session_type).c_str(), xma_core::get_cu_cmd_state(cu_cmd).c_str());
+                        std::cout << "XMA FATAL: Unexpected error ERT_CMD_STATE. Cmd state: " << xma_core::get_cu_cmd_state(cu_cmd) << std::endl;
+                        throw std::runtime_error("Unexpected ERT Cmd State");
```
stsoe, line 573: Maybe the exception message should be the same as what you print above?
It's a little odd that the same message is repeated 3 times (xma_log, std::cout, exception), maybe the std::cout should be removed?

stsoe (commented):
Added comment.

src/xma/src/xmaapi/xma_utils.cpp
sarab96: Updated
Exception message is updated with details.
Keeping xma_logmsg as that may go to syslog or file based on ini settings

src/xma/src/xmaapi/xma_utils.cpp
```diff
@@ -569,6 +569,13 @@ int32_t check_all_execbo(XmaSession s_handle) {
                     	       cu_cmd->state >= ERT_CMD_STATE_NORESPONSE) {
                         //Check for invalid/error execo state
                         xma_logmsg(XMA_ERROR_LOG, XMAUTILS_MOD, "Session id: %d, type: %s, Unexpected ERT_CMD_STATE. state=%s", s_handle.session_id, xma_core::get_session_name(s_handle.session_type).c_str(), xma_core::get_cu_cmd_state(cu_cmd).c_str());
+                        std::string err_tmp("XMA FATAL: Session id: ");
+                        err_tmp.append(std::to_string(s_handle.session_id));
+                        err_tmp.append(", type: ");
+                        err_tmp.append(xma_core::get_session_name(s_handle.session_type));
+                        err_tmp.append("; Unexpected ERT_CMD_STATE. state=");
+                        err_tmp.append(xma_core::get_cu_cmd_state(cu_cmd));
```
stsoe, line 577: boost::format might be easier here.  but no worries.

stsoe (approved):

PR #4841, opened 2021-03-03
Mb sched off chk

chienwei-lan:
U25 has ert in its shell, however due to bug, we have a flag MB_SCHED_OFF_CHK as workaround.

Once the bit is set, we should not probe ert then it will fall back to KDS mode

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
@@ -1303,8 +1303,7 @@ static int ert_user_probe(struct platform_device *pdev)
 	int err = 0;
 	xdev_handle_t xdev = xocl_get_xdev(pdev);
 	struct xocl_ert_sched_privdata *priv = NULL;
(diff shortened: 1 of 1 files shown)


xuhz (approved):

houlz0507 (approved):

mamin506:
@maxzhen we need revert this change for now.

I see on U30 and Versal, xocl_mb_sched_on() would return false.
This change disabled PS ERT then PS kernel would not work.
We will need more thought about this and might need clean up our design and avoid check if device is MPSOC or Versal.

@larry9523 can add more comments.

chienwei-lan:
> @maxzhen we need revert this change for now.
> 
> I see on U30 and Versal, xocl_mb_sched_on() would return false.
> This change disabled PS ERT then PS kernel would not work.
> We will need more thought about this and might need clean up our design and avoid check if device is MPSOC or Versal.
> 
> @larry9523 can add more comments.

or should I add it as a workaround before the more comprehensive solution happen?
	bool ert = (XOCL_DSA_IS_VERSAL(xdev) || XOCL_DSA_IS_MPSOC(xdev)) ? 1 :
	    xocl_mb_sched_on(xdev);

PR #4844, opened 2021-03-04
XRT-861 Check CU‘s metadata and safely ignore unrecognizable CU

mamin506:
This change is U30 specific. It would allow incorrect CU address in xclbin and safely handle it.
This change would not break existed cases.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
@@ -1521,12 +1521,21 @@ static int icap_create_subdev_cu(struct platform_device *pdev)
 	for (i = 0; i < ip_layout->m_count; ++i) {
 		struct xocl_subdev_info subdev_info = XOCL_DEVINFO_CU;
 		struct ip_data *ip = &ip_layout->m_ip_data[i];
+		u32 range;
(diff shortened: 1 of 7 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
```diff
@@ -1521,13 +1521,22 @@ static int icap_create_subdev_cu(struct platform_device *pdev)
...
+		u32 range;
 
 		if (ip->m_type != IP_KERNEL)
 			continue;
 
 		if (ip->m_base_address == 0xFFFFFFFF)
 			continue;
 
+		range = subdev_info.res[0].end - subdev_info.res[0].start;
+		/* If CU's address is not aligned, remove CU's resource */
```
maxzhen, line 1533: i thought we want to use cu type not address, no?

mamin506: This is U30 branch specific. There would be no CU type change in the xclbin.
For master, I would provide a different change.

mamin506:
The latest U30 released xclbin has correct address. So, we think this is not needed in U30 branch.
We would have a proper solution in master branch.

PR #4859, opened 2021-03-05
sw emu - AP_START bit set in swscheduler 

vboggara-xilinx:
Issue : AP_CTRL register has stale AP_START bit set, so when swsscheduler writes the register map to the CU, it writes the entire register map including AP_CTRL register in one xclWrite call, this ends up starting the CU before the complete register map is written.
Fix : Madee fix in swsscheduler to clear regmap[0] prior to writing the register map.


vboggara-xilinx:
can someone merge this change please?

PR #4864, opened 2021-03-05
CR-1090745: Fix issues in application debug

jvillarre:
This pull request fixes the python error message reporting by printing out the raised ValueError objects directly as opposed to accessing the message member which no longer exists.
It also refactors the loading of all the OpenCL plugins into a single location and reintroduces the loading of the application debug plugin, which was missed in the original OpenCL profiling refactoring submission.


IshitaGhosh (approved):

hackwa (approved):
python changes looks good

PR #4865, opened 2021-03-06
Add graph open context

larry9523:
This PR addresses the following features
1) Added zocl graph context, a linked list to each sched_exec_client
2) Added new API (C) to open different Graph handler with given context (shared and exclusive) and the default context is primary
3) Added check in most graph operation (run, end, syncrtp etc.) to return -EPERM if the operation is not allowed for its context
4) Added check context open. There can only be one exclusive context. There can be on primary with multiple shared context.


src/runtime_src/core/edge/common/aie_parser.cpp
```diff
@@ -263,6 +277,18 @@ get_tiles(const xrt_core::device* device, const std::string& graph_name)
   return ::get_tiles(aie_meta, graph_name);
 }
 
+int
+get_graph_id(const xrt_core::device* device, const std::string& graph_name)
+{
+  auto data = device->get_axlf_section(AIE_METADATA);
+  if (!data.first || !data.second)
+    return -1;
```
stsoe, line 285: Is this a valid return id?  I assume it is not an error, but what does -1 indicate?  Should it be a symbolic constant, or at least document in aie_parser.h:get_grap_id().

stsoe (approved):
Looks good to me.

xdavidz (approved):

larry9523:
retest this please

src/runtime_src/core/edge/drm/zocl/zocl_aie.c
```diff
@@ -430,6 +431,137 @@ zocl_aie_reset(struct drm_zocl_dev *zdev)
...
+	list_for_each(cptr, &zdev->exec->ctx_list) {
+		ctx = list_entry(cptr, struct sched_client_ctx, link);
+
+		spin_lock_irqsave(&ctx->graph_list_lock, graph_flags);
+
+		list_for_each(gptr, &ctx->graph_list) {
+			gnode = list_entry(gptr, struct zocl_graph_ctx_node,
+			    link);
+
+			if (gnode->gid == gid) {
```
sonals, line 455: Do you want to invert the check which will remove the deep indentation?
```
if (gnode-->gid != gid)
    continue;
```

sonals (changes requested):
Code control flow suggestion for early exit

src/runtime_src/core/edge/drm/zocl/zocl_aie.c
larry9523: Done.

src/runtime_src/core/edge/common/aie_parser.cpp
larry9523: Defined a symbolic constant,
xrt_core::edge::aie::NON_EXIST_ID
Does it look good?

mamin506 (approved):

sonals (approved):

jerryhsu-xilinx (approved):

PR #4868, opened 2021-03-07
Update shim.cpp copyright

larry9523:
This is a pure test of pipeline during weekend when pipeline is idle.

Because PR https://github.com/Xilinx/XRT/pull/4865 is failing by the following error, where the code is not touched at all by the changes, this PR is to test if the same error can be caught by pipe line too.

10:46:20 /opt/xrt/src/runtime_src/core/edge/user/shim.cpp: In member function 'int ZYNQ::shim::xclOpenContext(const unsigned char*, unsigned int, bool)':
10:46:20 /opt/xrt/src/runtime_src/core/edge/user/shim.cpp:844:3: sorry, unimplemented: non-trivial designated initializers not supported
10:46:20    };
10:46:20    ^


larry9523:
root caused

PR #4883, opened 2021-03-10
CR-1091236 Mailbox message with Op code XCL_MAILBOX_REQ_PROGRAM_SHELL…

xuhz:
…(14) with message req id INVALID_REQ_ID(0xffffffffffffffff) cause kernel panic


maxzhen (commented):
we are dropping the response to the bad msg, right? can we just drop this msg with bad msg id upon receiving?

PR #4884, opened 2021-03-10
CR-1091201: Fixing user event table in OpenCL summary

jvillarre:
Changed the statistics database to store user event information based off of strings instead of const character pointers.  This ensures that the XDP side will always have a copy of the string to print out, even if the user originally passed a normal character pointer that was cleaned up before the summary file is generated.


pgschuey (approved):
Conversion to std::string looks good to me.

IshitaGhosh (approved):
Change looks good

PR #4888, opened 2021-03-10
CR-1091194 fix mem leak in probe

Himanshu-xilinx:


src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -1545,12 +1545,19 @@ xclProbe()
...
     return 0;
+  }
 
   result = std::strncmp(version.name, "zocl", 4);
   close(fd);
-
+  delete []version.name;
+  delete []version.desc;
+  delete []version.date;
   return (result == 0) ? 1 : 0;
```
stsoe, line 1561: Can we use std::string and populate the drm_version struct with name.c_str(), version.c_str() ?
The std::string can be constructed and initialized with `n` number of characters, e.g. `std::string name{128,0}`;
Raw new and delete should be avoided for exception safety reasons and to avoid memory leaks.

stsoe (changes requested):

src/runtime_src/core/edge/user/shim.cpp
Himanshu-xilinx: done

dayeh-xilinx:
retest this please. Cancelled earlier to give priority to u30 branch.

stsoe (approved):

Issue #4131, opened 2020-09-30
XRT Build Failure Caused By Deprecated Protobuf API Calls


jo7ueb:
I encountered following error when running `build.sh`.
Deprecated Protobuf API is called many times.

```
[ 19%] Building CXX object xma/src/xmaapi/CMakeFiles/xma2api.dir/xmalogger.cpp.o
In file included from /home/yasu/xilinx/XRT/src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.h:28,
                 from /home/yasu/xilinx/XRT/src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/debug.cxx:35:
/home/yasu/xilinx/XRT/src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/debug.cxx: In member function ‘void xclhwemhal2::HwEmShim::fetchAndPrintMessages()’:
/home/yasu/xilinx/XRT/src/runtime_src/core/include/xcl_api_macros.h:35:38: error: ‘int google::protobuf::MessageLite::ByteSize() const’ is deprecated: Please use ByteSizeLong() instead [-Werror=deprecated-declarations]
   35 |      unsigned c_len = c_msg.ByteSize(); \
      |                                      ^
/home/yasu/xilinx/XRT/src/runtime_src/core/include/xcl_api_macros.h:529:3: note: in expansion of macro ‘SERIALIZE_AND_SEND_MSG’
  529 |   SERIALIZE_AND_SEND_MSG(func_name) \
      |   ^~~~~~~~~~~~~~~~~~~~~~
/home/yasu/xilinx/XRT/src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/debug.cxx:186:4: note: in expansion of macro ‘xclGetDebugMessages_RPC_CALL’
  186 |    xclGetDebugMessages_RPC_CALL(xclGetDebugMessages,ack,force,displayMsgs,logMsgs,stopMsgs);
      |    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/local/include/google/protobuf/generated_enum_util.h:36,
                 from /usr/local/include/google/protobuf/map.h:55,
                 from /usr/local/include/google/protobuf/generated_message_table_driven.h:34,
                 from /home/yasu/xilinx/XRT/build/Debug/runtime_src/core/pcie/emulation/hw_em/../common_em/rpc_messages.pb.h:26,
[...]

PR #4895, opened 2021-03-10
Use new protobuf API instead of deprecated one

keryell:
This fixes compilation failures like
```
  error: ‘int google::protobuf::MessageLite::ByteSize() const’ is deprecated: Please use ByteSizeLong() instead [-Werror=deprecated-declarations]
```
This should fix https://github.com/Xilinx/XRT/issues/4131

### DIFF
--- a/src/runtime_src/core/include/xcl_api_macros.h
+++ b/src/runtime_src/core/include/xcl_api_macros.h
@@ -32,22 +32,22 @@ mtx.unlock();
     AQUIRE_MUTEX()
 
 #define SERIALIZE_AND_SEND_MSG(func_name)\
-     unsigned c_len = c_msg.ByteSize(); \
+     unsigned c_len = c_msg.ByteSizeLong(); \
     buf_size = alloc_void(c_len); \
     bool rv = c_msg.SerializeToArray(buf,c_len); \
     if(rv == false){std::cerr<<"FATAL ERROR:protobuf SerializeToArray failed"<<std::endl;exit(1);} \
     \
     ci_msg.set_size(c_len); \
     ci_msg.set_xcl_api(func_name##_n); \
-    unsigned ci_len = ci_msg.ByteSize(); \
+    unsigned ci_len = ci_msg.ByteSizeLong(); \
     rv = ci_msg.SerializeToArray(ci_buf,ci_len); \
     if(rv == false){std::cerr<<"FATAL ERROR:protobuf SerializeToArray failed"<<std::endl;exit(1);} \
     \
     _s_inst->sk_write(ci_buf,ci_len); \
     _s_inst->sk_write(buf,c_len); \
     \
-    _s_inst->sk_read(ri_buf,ri_msg.ByteSize()); \
-    rv = ri_msg.ParseFromArray(ri_buf,ri_msg.ByteSize()); \
+    _s_inst->sk_read(ri_buf,ri_msg.ByteSizeLong()); \
+    rv = ri_msg.ParseFromArray(ri_buf,ri_msg.ByteSizeLong()); \
     assert(true == rv);\
     buf_size = alloc_void(ri_msg.size()); \
     _s_inst->sk_read(buf,ri_msg.size()); \
--- a/src/runtime_src/core/include/xcl_api_macros.h
+++ b/src/runtime_src/core/include/xcl_api_macros.h
@@ -32,14 +32,14 @@ mtx.unlock();
     AQUIRE_MUTEX()
 
 #define SERIALIZE_AND_SEND_MSG(func_name)\
-     unsigned c_len = c_msg.ByteSizeLong(); \
+    auto c_len = c_msg.ByteSizeLong(); \
     buf_size = alloc_void(c_len); \
     bool rv = c_msg.SerializeToArray(buf,c_len); \
     if(rv == false){std::cerr<<"FATAL ERROR:protobuf SerializeToArray failed"<<std::endl;exit(1);} \
     \
     ci_msg.set_size(c_len); \
     ci_msg.set_xcl_api(func_name##_n); \
-    unsigned ci_len = ci_msg.ByteSizeLong(); \
+    auto ci_len = ci_msg.ByteSizeLong(); \
     rv = ci_msg.SerializeToArray(ci_buf,ci_len); \
     if(rv == false){std::cerr<<"FATAL ERROR:protobuf SerializeToArray failed"<<std::endl;exit(1);} \
     \
--- a/src/runtime_src/core/include/xcl_api_macros.h
+++ b/src/runtime_src/core/include/xcl_api_macros.h
@@ -31,28 +31,55 @@ mtx.unlock();
     func_name##_response r_msg; \
     AQUIRE_MUTEX()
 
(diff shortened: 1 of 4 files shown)


keryell:
Anyone to improve the CI so we can get the reason of error, like compilation output?

keryell:
Curious. How did the ARM64 ever pass?

keryell:
Note that some scripts run in verbose mode to have some feedback from the CI.
I do not know whether this is requested by normal users.
Perhaps for the curious people?
Otherwise we could add a verbose option to be used at least by the CI.

stsoe (changes requested):
@keryell Please comment on the build.sh and xrtdeps.sh changes and why the removed code is not necessary.

build/build.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -vx
```
rozumx, line 1: Is this something we wish to do?
```
   -v        Print shell input lines as they are read.
   -x        Print commands and their arguments as they are executed.

```

src/runtime_src/core/include/xcl_api_macros.h
```diff
@@ -32,22 +32,22 @@ mtx.unlock();
     AQUIRE_MUTEX()
 
 #define SERIALIZE_AND_SEND_MSG(func_name)\
-     unsigned c_len = c_msg.ByteSize(); \
+     unsigned c_len = c_msg.ByteSizeLong(); \
```
rozumx, line 35: Now that the return type is now known, shouldn't the variable type also be updated.  

unsigned is unfortunately not a good 'type', shouldn't it be instead unsigned long.  For example:
```suggestion
     unsigned long c_len = c_msg.ByteSizeLong(); \
```

rozumx (changes requested):
Question regarding:
1. The use of "-vx" for the bash scripts.
2. Not using "unsigned" type.

src/runtime_src/core/include/xcl_api_macros.h
keryell: The code is probably wrong if the message size does not fit in an `int` like it was anyway wrong before if the message size was not fitting in the deprecated `ByteSize()`.
I do not have the big picture of the code so I do not know if it can happen or not.
Probably the code should get rid of macros, use more `auto` and functions/lambda with type-safety guarantees. But this sounds outside the scope of this...

build/build.sh
keryell: Just a `--verbose` doing an equivalent `set`?

stsoe: I failed to notice -vx.  I don't think we want that in our build script.   You can use -verbose to see the actual compile commands, no need for more verbosity.

keryell:
> @keryell Please comment on the build.sh and xrtdeps.sh changes and why the removed code is not necessary.

The code looks like some work around for limitation in a 3 year old Linux distribution.
Should we keep a comment in the code about something that should not have happened at the first place?
Or do you mean a comment in the log message?

stsoe:
> > @keryell Please comment on the build.sh and xrtdeps.sh changes and why the removed code is not necessary.
> 
> The code looks like some work around for limitation in a 3 year old Linux distribution.
> Should we keep a comment in the code about something that should not have happened at the first place?
> Or do you mean a comment in the log message?

We still have to support > 3 year old Linux distributions.  Unless it can proven that the check is completely obsolete for all the OSes Xilinx has to support, then let's leave it.

keryell:
But something is very curious: it looks like the CI on the main branch of this repository has a completely different behavior from the CI I have on my branch.
So the same master branch is compiling on ARM64 Ubuntu bionic on the CI here while it would not compile on the CI of https://github.com/keryell/XRT
This is why I had to fix the ARM64 compilation... :-(
Am I missing something obvious here?

stsoe:
> But something is very curious: it looks like the CI on the main branch of this repository has a completely different behavior from the CI I have on my branch.
> So the same master branch is compiling on ARM64 Ubuntu bionic on the CI here while it would not compile on the CI of https://github.com/keryell/XRT
> This is why I had to fix the ARM64 compilation... :-(
> Am I missing something obvious here?
Why is https://github.com/keryell/XRT 3939 commits behind master?

keryell:
> Why is [keryell/XRT](https://github.com/keryell/XRT) 3939 commits behind master?

Because I do not use it? :-) I just push my branches there...
So I do not think this is the problem.

Look for someone else too: https://travis-ci.org/github/Xilinx/XRT/jobs/762502204
What is going on?

keryell:
> We still have to support > 3 year old Linux distributions. Unless it can proven that the check is completely obsolete for all the OSes Xilinx has to support, then let's leave it.

I agree.
I made this "fix" to have ARM64 working and just understanding from the CI that it was not working at the first place, so there was no real support breakage.
But the current CI seems completely fishy now and thus the last 2 commits of this PR are just insanity caused by current CI state I do not understand...

stsoe:
@keryell Can you change this PR and remove the changes to build.sh?   Whether the changes to build.sh are warranted or not, separating the two changes is preferred.

keryell:
I completely forgot about this up to I tried to install the latest XRT today and it failed...
Here is the minimal change to have it to compile on a modern system.

src/runtime_src/core/include/xcl_api_macros.h
keryell: I just put `auto` to implement your suggestion.

stsoe:
ok to test

stsoe (approved):

stsoe:
@keryell There are build errors, please check link per slacked message

keryell:
OK, there were still 2 function calls to the deprecated ProtoBuf API.
Let's try to build again...

keryell:
But what is curious is that it was compiling fine on our Ubuntu 20.10 machines...
Be ready for Ubuntu 21.04 coming soon! :-)

stsoe:
> Be ready for Ubuntu 21.04 coming soon! :-)
We will of course make sure we can build on 21.04, but I believe official support is  LTS only.

keryell:
So, according to https://github.com/eddelbuettel/rprotobuf/issues/70 it looks like the problem is more complex as `protoc` was changed with the safer behavior starting with version 3.6.
https://github.com/protocolbuffers/protobuf/blob/master/CHANGES.txt
Thus, using the 2 versions according to the version of the ProtoBuf used.
Last try before forking this repository and leaving this to the Jurassix world...

chvamshi-xilinx (approved):
Looks good. @venkatp-xilinx / @hemantk-xilinx/ @vboggara-xilinx. Can you please make this change in C-models/genericpcie model side as well

chvamshi-xilinx:
Adding "don't merge" tag as this change required in device side as well. Ideally things should work, but better to verify both sw_emu and hw_emu before merging this change.

keryell:
If someone spends some time to understand this, it might be an opportunity to check also what is going on in https://github.com/Xilinx/XRT/issues/5081

chvamshi-xilinx:
Removed "donot merge" tag As @vboggara-xilinx & @hemantk-xilinx verified both sw_emu and hw_emu

PR #4916, opened 2021-03-12
Added the new ENV to enable the DDR Self refresh feature

venkatp-xilinx:
Added the new ENV to enable the DDR Self refresh feature


dayeh-xilinx:
retest this please. Cancelled earlier to give priority to u30 branch.

PR #4922, opened 2021-03-12
set device to ready when done updating

pgschuey:


jvillarre (approved):

hackwa (approved):
looks good. the aie profile should work with this change

PR #4923, opened 2021-03-12
VITIS-1667 AIE - Multi Process Support

larry9523:
This PR addresses the following features

1. Added zocl AIE context field in sched_exec_client
2. Added new API (C) to open different AIE context (exclusive, primary, shared). To support backward compatibility, if no xrtAIEOpenXXX (context) is called, by default, we will try to acquire primary context once the user application tries to access AIE array.
3. Added check in AIE operations (GMIO, Profiling, Reset) to return -EPERM if its context mode does not allow those operations
4. Added check when opening AIE context. There can only be one exclusive context. There can be one primary with multiple shared contexts
5. AIE context support will be added to new KDS once it is default scheduler on edge


mamin506 (approved):

src/runtime_src/core/include/experimental/xrt_aie.h
```diff
@@ -113,6 +115,68 @@ extern "C" {
...
+ * @handle:          Handle to the device
+ * @xclbinUUID:      UUID of the xclbin withe specified AIE
+ *
+ * Return:          0 on success, or appropriate error number.
+ *
+ * Note: After close context, any access to AIE, we will try to acquire
+ *       primary context by default.
+ */
+int
+xrtAIECloseContext(xrtDeviceHandle handle, const xuid_t xclbinUUID);
```
stsoe, line 178: Is this really to close a context?     Isn't it more close of an aie kind of handle?  In other words, should xrtAIEOpenBlahBlah() return a handle, that should then be closed with xrtAIEClose(handle) ?  What is the C++ equivalent , once implemented?   Won't the C++ code create an xrt::aie::device (akin to a handle) which is automatic (reference counted) and when it goes out of scope the context is automatically closed unbeknownst to the user?

stsoe (changes requested):
Looks good.  Just few style comments, and one design question related to exposing of close context.

src/runtime_src/core/include/experimental/xrt_aie.h
larry9523: In C APIs, AIE related operations are using device handle rather than AIE specific handle
xrtAIESyncBO(xrtDeviceHandle handle, xrtBufferHandle bohdl
xrtAIEResetArray(xrtDeviceHandle handle);

We can return a handle to xrtAIEOpen_context_, but the only place this handle will be used is xrtCloseContext. I am not sure if we want to introduce a handle here. It would have been more straightforward if we have used an AIE handler in xrt AIE operation, like we have xrtGraphOpen, xrtPLKernelOpen.

src/runtime_src/core/edge/user/aie/aie.cpp
```diff
@@ -130,13 +130,53 @@ XAie_DevInst* Aie::getDevInst()
   return devInst;
 }
 
+void
+Aie::
+openContext(const xrt_core::device* device, const uuid_t xclbin_uuid, xrt::aie::access_mode am)
```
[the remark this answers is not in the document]

src/runtime_src/core/edge/user/aie/aie.h
```diff
@@ -100,6 +100,15 @@ class Aie {
 
     XAie_DevInst *getDevInst();
 
+    void
+    openContext(const xrt_core::device* device, const uuid_t xclbin_uuid, xrt::aie::access_mode am);
```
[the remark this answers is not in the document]

xdavidz (approved):

src/runtime_src/core/include/experimental/xrt_aie.h
stsoe: I realize that xrt::aie::device is just adding additional APIs on top of xrt::device, but maybe it should be more than just API extensions?    And maybe xrtAIEDeviceOpen should be corresponding to creating xrt::aie::device so that both work with same aie specific handle that can manage the context for the user?   It's odd that in C++ the user has to construct an xrt::aie::device but that in C there is no corresponding xrtAIEDeviceOpen().

larry9523: I made the related change based on our discussion. The following APIs are removed
xrtAIEOpenExcluseiv
xrtAIEOpenPrimary
xrtAIEOpenShared
xrtAIECloseContext

The following APIs are added
xrtAIEOpenDevice
xrtAIEOpenDeviceExclusive
xrtAIEOpenDeviceShared

Please take another look. Thanks

stsoe (approved):
Looks good.

PR #4924, opened 2021-03-13
CR-1090650: Fixing XDP issue where kernel reads had start events but no end events in device trace

jvillarre:


hackwa (approved):
now we ignore any pending starts when we receive a pulse event.

PR #4926, opened 2021-03-13
enable mailbox polling mode by default

maxzhen:
Enable polling mode by default for mailbox subdev driver. @xuhz , seems that we don't need to check mailbox transfer speed any more with polling (at most 1000/s times)?

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mailbox.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mailbox.c
@@ -204,7 +204,7 @@
 #include <linux/sched/clock.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


houlz0507 (approved):

xuhz (approved):
checking receiving rate may not be required in polling mode since it will not be triggered in polling mode.
if interrupt mode support is removed later, probably that can be removed too.

maxzhen:
> 
> 
> checking receiving rate may not be required in polling mode since it will not be triggered in polling mode.
> if interrupt mode support is removed later, probably that can be removed too.

Thanks for confirming. Let's keep the interrupt mode and the check for now just in case we need to debug some issue specific to polling.

PR #4935, opened 2021-03-16
CR-1088964 too many printk cause hang in Azure VM. (#4730)

houlz0507:


src/runtime_src/core/pcie/driver/linux/xocl/lib/libxdma.c
```diff
@@ -30,6 +30,8 @@
 
 extern unsigned int desc_blen_max;
 
+#define xocl_pr_info(fmt, args...)				\
+	printk(KERN_DEBUG pr_fmt(fmt), ##args)
```
maxzhen, line 34: can we use pr_debug() which preserves the msg format and also do KERN_DEBUG?

houlz0507: pr_debug() will equivalent to printk only when DEBUG is defined.

maxzhen (approved):

PR #4951, opened 2021-03-17
CR-1090787 [xdma] ERROR: unable to sync BO on VCK5000 PLRAM, xocl:xdm…

xdavidz:
One of the issue of this bug is:

xbutil mem --write -a 0x20204000000 -i 256 -e 0x5a
failed, but return 0 back.


xdavidz:
wrong comment

PR #4953, opened 2021-03-18
XRT-860 Don't map 64KB hardcoding register space

mamin506:
1. Get kernel register range from xml metadata
2. base_paddr and size sysfs node per CU
3. xclRegRead() and xclRegWrite() would map register range per CU

### DIFF
--- a/src/runtime_src/core/pcie/linux/shim.h
+++ b/src/runtime_src/core/pcie/linux/shim.h
@@ -181,8 +181,7 @@ private:
      * Mapped CU register space for xclRegRead/Write(). We support at most
      * 128 CUs and each map is of 64k bytes.
      */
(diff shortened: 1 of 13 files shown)


maxzhen:
retest this please - infra issue (DHCP does not work).

src/runtime_src/core/pcie/linux/shim.h
```diff
@@ -181,8 +181,7 @@ class shim
      * Mapped CU register space for xclRegRead/Write(). We support at most
      * 128 CUs and each map is of 64k bytes.
```
maxzhen, line 182: This comment still mentioned hard-coded 64k range. I think we can remove it now.

maxzhen (approved):

PR #4954, opened 2021-03-18
Trace Changes

hackwa:
1. Enable circular buffer for DMA offload
2. Disable continuous offload, circular buffer for FIFO offload
3. Enable multiple files for lop, opencl plugins


jvillarre (approved):
Changes look good.

PR #4957, opened 2021-03-18
CR-1092167: Adding emulation files in the run summary only when they exist

jvillarre:
With this change, XDP now checks the Emulation.debug_mode setting in the ini file.  If the setting is either batch or gui, then the WDB, WCFG, and kernel internal files will be copied to the local directory and only then should be added to the run summary.

Note that the environment variables are still necessary as they have the names of the files, but they are not sufficient as the environment variables are set even in cases where the files are not copied.

Also note, XDP is going off of the debug_mode flag as the launch_waveform flag has been deprecated.  Any tests or applications that set launch_waveform will not see the emulation files in the run summary.

### DIFF
--- a/src/runtime_src/core/common/config_reader.h
+++ b/src/runtime_src/core/common/config_reader.h
@@ -580,6 +580,13 @@ get_launch_waveform()
 }
 
 inline std::string
+get_debug_mode()
+{
+  static std::string value = detail::get_string_value("Emulation.debug_mode", "off");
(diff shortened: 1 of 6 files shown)


chvamshi-xilinx (commented):
Hi @jvillarre , This check is good, but this check should happen inside emulation drivers itself. 
@venkatp-xilinx , Can you please check.

PR #4958, opened 2021-03-18
CR-1091338: Fixing device side events that had multiple end events match a single start event

jvillarre:
When handling pulses on the device side (events that have a start and stop event on the same cycle), we are creating dummy start events to match with the end events.  We were erroneously marking these dummy events as start events to be matched later on by adding them to the database.  This pull request removes the marking of those dummy events in the database.


IshitaGhosh (approved):
Change looks good

PR #4959, opened 2021-03-18
Add approximation for missing end events on ASM connection based on CU end event

IshitaGhosh:


jvillarre (approved):
Additions look good to approximate the end of any stream transactions where we don't have end events.

PR #4962, opened 2021-03-19
VITIS-1114 XRT caching framework

larry9523:
This PR added the following feature
1) Added a new xrt level C api to use xclbin loaded onto the system by specify xclbin UUID
    int xrtDeviceLoadXclbinUUID(xrtDeviceHandle dhdl, const xuid_t uuid);
   This API, now, only works on VCK190 platform to support AIE multi-processes; Calling this API on other platform will get error 
    "load xclbin by uuid is not supported"
   It will be extended to support other edge platform and DC in future PRs
2) In zocl driver, we will error out when shared AIE context tries to load AIE only xclbin because shared context does not allow to change the hardware
3) When valid context loads AIE only xclbin, the xclbin blob will be cached into a sysfs node. So that any AIE context can load this cached xclbin by its UUID. (because it does not change hardware)

### DIFF
--- a/src/runtime_src/core/common/api/xrt_device.cpp
+++ b/src/runtime_src/core/common/api/xrt_device.cpp
@@ -415,6 +415,26 @@ xrtDeviceLoadXclbinHandle(xrtDeviceHandle dhdl, xrtXclbinHandle xhdl)
 }
 
 int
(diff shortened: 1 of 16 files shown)


stsoe (approved):
Look good.

mamin506 (approved):

PR #4968, opened 2021-03-22
xma header changes for return code, etc

sarab96:
xma header changes for return code, etc
Multiple filter outputs
Cmd object with cmd_state & return code
Version change as change in header files

### DIFF
--- a/src/xma/include/app/xmafilter.h
+++ b/src/xma/include/app/xmafilter.h
@@ -73,9 +73,9 @@ typedef struct XmaFilterProperties
     XmaFilterPortProperties  input;
     /** output data port instance */
     XmaFilterPortProperties  output;
-    //XmaFilterPortProperties  output[MAX_FILTER_OUPUTS]; To be changed in 2020.2_pu1
+    XmaFilterPortProperties  outputs[MAX_FILTER_OUTPUTS];
     /** number of actual filter outputs */
-    //int32_t                   num_outputs; Headers to be changed in 2020.2_pu1 when it will be implemented
+    int32_t                   num_outputs;
     /* user-defineable properties */
     /** array of custom parameters for port */
     XmaParameter             *params;
--- a/src/xma/include/app/xmahw.h
+++ b/src/xma/include/app/xmahw.h
@@ -33,6 +33,17 @@ typedef struct XmaHwSession
     uint32_t         reserved[4];
 } XmaHwSession;
 
+enum xma_cmd_state {
+  QUEUED = 1, //Submitted to XMA -> XRT
+  COMPLETED = 2, //Cmd has finished
+  ERROR = 3, //XMA or XRT error during submission of cmd
+  ABORT = 4, //XRT aborted the cmd; CU may or may not have received the cmd
+  TIMEOUT = 5, //XMA or XRT timeout waiting for cmd to finish
+  PSK_ERROR = 6, //PS Kernel cmd completed but with error return code
+  PSK_CRASHED = 7, //PS kernel has crashed
+  MAX // Always the last one
+};
+
 
 #ifdef __cplusplus
 }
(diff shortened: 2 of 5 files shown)


src/xma/include/app/xmafilter.h
```diff
@@ -73,9 +73,9 @@ typedef struct XmaFilterProperties
     XmaFilterPortProperties  input;
     /** output data port instance */
     XmaFilterPortProperties  output;
-    //XmaFilterPortProperties  output[MAX_FILTER_OUPUTS]; To be changed in 2020.2_pu1
```
stsoe, line 76: Please remove dead code

src/xma/include/app/xmahw.h
```diff
@@ -33,6 +33,17 @@ typedef struct XmaHwSession
...
+enum xma_cmd_state {
+  QUEUED = 1, //Submitted to XMA -> XRT
+  COMPLETED = 2, //Cmd has finished
+  ERROR = 3, //XMA or XRT error during submission of cmd
+  ABORT = 4, //XRT aborted the cmd; CU may or may not have received the cmd
+  TIMEOUT = 5, //XMA or XRT timeout waiting for cmd to finish
+  PSK_ERROR = 6, //PS Kernel cmd completed but with error return code
+  PSK_CRASHED = 7, //PS kernel has crashed
+  MAX // Always the last one
+};
```
stsoe, line 45: If this is a C++ header (not sure about that), then use enum class and lower case enumerators.  If this is a C header, then please prefix all enums so as not to conflict with braindead #defines in included header files.

Additionally if the enum is needed for end-user C-code, then  I strongly recommend having an enum class mirroring the classic enum.  E.g. within #ifdef __cplusplus have 
```
enum class xma_cmd_state : uint8_t {
queued = XMA_QUEUED;
completed = XMA_COMPLETED;
...
};
```
Write your implementation code, which is all C++, using this enum class.
Rename the classic C enum to `enum XmaCmdState` would of course be needed.

stsoe (changes requested):
Please consider the comments I added.

src/xma/include/app/xmafilter.h
sarab96: removed

src/xma/include/app/xmahw.h
sarab96: Prefixed for end-user c-code and added enum class for internal use in c++ code

src/xma/include/lib/xmahw_lib.h
```diff
@@ -61,6 +61,17 @@ constexpr std::uint64_t signature = 0xF42F1F8F4F2F1F0F;
...
 
+enum class xma_cmd_state_priv: std::int32_t {
+  QUEUED = 1, //Submitted to XMA -> XRT
+  COMPLETED = 2, //Cmd has finished
+  ERROR = 3, //XMA or XRT error during submission of cmd
+  ABORT = 4, //XRT aborted the cmd; CU may or may not have received the cmd
+  TIMEOUT = 5, //XMA or XRT timeout waiting for cmd to finish
+  PSK_ERROR = 6, //PS Kernel cmd completed but with error return code
+  PSK_CRASHED = 7, //PS kernel has crashed
+  MAX = 8 // Always the last one
```
stsoe, line 72: It's kind of magic that the numbers must match the external classic enum.

You should (1) use lower case enumerators, otherwise you have the same problem as to begin with in that some braindead code could #define MAX, ABORT, ERROR, ... to something.  (2) initialize the enumerators from corresponding external enumerators
```
queued = XMA_QUEUED,
completed = XMA_COMPLETED,
etc. 
```
When I look at the xmahw_lib.h external header file, it is using CamelCase for C facing types, which is good, so I think it should have used XmaCmdState to match the C naming convention (see XmaHwSession for example).  Your C++ facing enum class could then have used just xma_cmd_state.

The XRT naming convention is foo_underscore_bar for C++ types, functions, and variables, and CamelCase or C stuff; it looks like XMA agrees with this in general.

sarab96: Updated

stsoe (approved):
Thanks for making the changes

PR #4969, opened 2021-03-22
Trace enhancements

hackwa:
1. Use a vector to offload trace instead of hard coded struct
2. Do one sync per offload
3. Don't offload trace in middle of continuous offload if there's too less trace (512 Bytes or 64 packets)


jvillarre (approved):
Changes look good

IshitaGhosh (approved):
Changes look good.

PR #4975, opened 2021-03-23
OpenCL-level profiling trace fix: Ignore write buffer events of buffer size 0

jvillarre:
Fixing issue where stray WRITE_BUFFER events were being generated incorrectly


hackwa (approved):
ignore end buffer events without start and 0 size

PR #4978, opened 2021-03-24
coverity fixes

sarab96:
coverity fixes


maxzhen (approved):
Could we create a function for these duplicated code so that we can fix one place instead of 4 places?

sarab96:
Done
> Could we create a function for these duplicated code so that we can fix one place instead of 4 places?

PR #4981, opened 2021-03-24
don't create CU subdev if metadata is missing

mamin506:
Revert workaround in #4977. Should not create CU subdevice if metadata is missing.


maxzhen (approved):
You did verify that this will not reintroduce the hang issue fixed by #4977?

mamin506:
> You did verify that this will not reintroduce the hang issue fixed by #4977?

Yes. I verified that. Icap would said "**** metadata missing, skip" when create CU sub-device.

PR #4982, opened 2021-03-24
EoU u30 CRs

AShivangi:
### CRs addressed:
- CR-1091797 xbutil examine CU status gives a false negative
- CR-1091660 Xbutil examine shows bug that xbutil query does not
- CR-1091505 Modify verbiage to keep cards vs devices straight during programming
- CR-1091489 xbmgmt examine -r host is reporting in user space
- CR-1091093 xbmgmt2 programming command error references an undocumented --update switch
- CR-1091091 Please have programming operations work for all u30 devices

### DIFF
--- a/src/runtime_src/core/tools/common/ReportCu.cpp
+++ b/src/runtime_src/core/tools/common/ReportCu.cpp
@@ -22,13 +22,7 @@
 #include "core/common/device.h"
 #include "core/common/utils.h"
 
-#include "core/pcie/linux/shim.h"
-#include "xrt.h"
-
 namespace qr = xrt_core::query;
-
-/* exposed by shim */
-int xclUpdateSchedulerStat(xclDeviceHandle);
 
 enum class cu_type {
   PL,
@@ -86,8 +80,7 @@ schedulerUpdateStat(xrt_core::device *device)
     auto at_exit = [] (auto device, auto uuid) { device->close_context(uuid.get(), -1); };
     xrt_core::scope_guard<std::function<void()>> g(std::bind(at_exit, device, uuid));
     
-    auto hdl = device->get_device_handle();
-    xclUpdateSchedulerStat(hdl);
+    device->update_scheduler_status();
   }
   catch (const std::exception&) {
     // xclbin_lock failed, safe to ignore
--- a/src/runtime_src/core/tools/common/ReportCu.cpp
+++ b/src/runtime_src/core/tools/common/ReportCu.cpp
@@ -22,7 +22,13 @@
 #include "core/common/device.h"
 #include "core/common/utils.h"
 
+#include "core/pcie/linux/shim.h"
+#include "xrt.h"
+
 namespace qr = xrt_core::query;
+
+/* exposed by shim */
+int xclUpdateSchedulerStat(xclDeviceHandle);
 
 enum class cu_type {
   PL,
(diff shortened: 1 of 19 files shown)


src/runtime_src/core/tools/common/ReportCu.cpp
```diff
@@ -70,9 +74,28 @@ get_cu_status(uint32_t cu_status)
   return pt;
 }
 
+static void
+schedulerUpdateStat(xrt_core::device *device)
+{
+  try {
+    // lock xclbin
+    auto uuid = xrt::uuid(xrt_core::device_query<xrt_core::query::xclbin_uuid>(device));
+    device->open_context(uuid.get(), static_cast<unsigned int>(-1), true);
```
rozumx, line 83: Please don't use -1 to initialize an unsigned int to all 1s.   Instead please use:
`std::numeric_limits<unsigned int>::max()`

src/runtime_src/core/tools/common/ReportCu.cpp
```diff
@@ -70,9 +74,28 @@ get_cu_status(uint32_t cu_status)
...
 }
 
+static void
+schedulerUpdateStat(xrt_core::device *device)
+{
+  try {
+    // lock xclbin
+    auto uuid = xrt::uuid(xrt_core::device_query<xrt_core::query::xclbin_uuid>(device));
+    device->open_context(uuid.get(), static_cast<unsigned int>(-1), true);
+    auto at_exit = [] (auto device, auto uuid) { device->close_context(uuid.get(), static_cast<unsigned int>(-1)); };
```
rozumx, line 84: Please use: `std::numeric_limits<unsigned int>::max()`

src/runtime_src/core/tools/common/ReportCu.cpp
```diff
@@ -70,9 +74,28 @@ get_cu_status(uint32_t cu_status)
...
+  }
+  catch (const std::exception&) {
+    // xclbin_lock failed, safe to ignore
+  }
+}
+
 boost::property_tree::ptree
 populate_cus(const xrt_core::device *device)
 {
+  if (!std::getenv("XCL_SKIP_CU_READ"))
```
rozumx, line 97: Could you help me understand why we are looking at environment variables.  Unfortunately, this is a bad practice and should be avoided whenever possible.

src/runtime_src/tools/xclbinutil/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ endif()
 # -----------------------------------------------------------------------------
 # Include the generated header files (e.g., version.h)
 include_directories(${CMAKE_BINARY_DIR}/gen)
-
+RETURN()
```
rozumx, line 12: Hmm...did you mean to do this?

rozumx (changes requested):
Some minor changes requested.

Double Check Request:  Since we are rebranding the "Shell" to "Base" and "Card" to "Device", please do a string search to see if you got all of the places they are used :-).

src/runtime_src/tools/xclbinutil/CMakeLists.txt
AShivangi: Nope!

src/runtime_src/core/tools/common/ReportCu.cpp
AShivangi: fixed

src/runtime_src/core/tools/common/ReportCu.cpp
AShivangi: fixed

src/runtime_src/core/tools/common/ReportCu.cpp
AShivangi: Not needed. removed

rozumx (approved):

PR #4983, opened 2021-03-24
CR-1092100 enabling kds2.0 blocks xbutil validate on vck5000

xdavidz:
Just simply migrate the code from mb_scheduler.c to new ert_user.c seems working well. xbutil validate passed without issues.


src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
```diff
@@ -1036,6 +1090,11 @@ static void ert_intc_enable(struct xocl_ert_user *ert_user, bool enable){
 	uint32_t i;
 	xdev_handle_t xdev = xocl_get_xdev(ert_user->pdev);
 
+	if (XOCL_DSA_IS_VERSAL(xdev)) {
+		xocl_mailbox_versal_request_intr(xdev, ert_versal_isr, ert_user);
+		return;
+	}
+
```
chienwei-lan, line 1097: you don't need disable in versal?

xdavidz: The mailbox_versal subdev will disable and enable the interrupt. This api only register the isr onto it.

dayeh-xilinx:
retest this please

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
```diff
@@ -1036,6 +1090,14 @@ static void ert_intc_enable(struct xocl_ert_user *ert_user, bool enable){
 	uint32_t i;
 	xdev_handle_t xdev = xocl_get_xdev(ert_user->pdev);
 
+	if (XOCL_DSA_IS_VERSAL(xdev)) {
```
mamin506, line 1093: We would like to avoid check device type in the code path. Can we avoid this?
If ert_user have different behavior on different device, we can check device in prob().

@chienwei-lan is more familiar with ert_user.c, I'm okay to refine this in another pull request.

mamin506 (approved):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
maxzhen: I agree. Adding is_versal() check here and there is not ideal. I'd suggest we define a callback in ert's softstate data structure. During probe(), we do the check in one place and set the callback to different function pointer for versal and alveo.

xdavidz: I think we talked about this before and I'd fix this this time, but I realized that there are 4 more places in ert_user.c which have "XOCL_DSA_IS_MPSOC and XOCL_DSA_IS_VERSAL". We'd fix all of them together.

Would @chienwei-lan  or me should fix that. I would like chienwei to think all the other places and get this if else fixed perfectly.

mamin506:
@xdavidz please remove "../subdev/ert_versal.o \" from userpf/Makefile.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
maxzhen: We can merge it now. Could you file a XRT task for @chienwei-lan to clean it up later?

maxzhen (approved):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_user.c
xdavidz: https://jira.xilinx.com/browse/XRT-868 created.

chienwei-lan (approved):

PR #4990, opened 2021-03-25
Disable boost::bind warnings

stsoe:
Later versions of boost spew warnings regarding bind placeholders in
global namespace.  This is a bug in boost itself since 1.73.0 (?)
(boost::property_tree), which hopefully will be fixed in later
versions of boost.

Embedded builds uses 1.74.0 by default so use
BOOST_BIND_GLOBAL_PLACEHOLDERS to retain the old behavior.


dayeh-xilinx:
retest this please. Build failed due to emulation test issue.

PR #4992, opened 2021-03-25
CR-1091928 ASTeR TC7.01 - xbutil2 validate fails at xcl_IOPs test

mamin506:
xbutil2 would try to find "PASS" in the output. Only if it exists, the test case passed.


dayeh-xilinx:
retest this please. Build failed due to emulation test issue.

PR #5004, opened 2021-03-29
xclbinutil: Added JSON lint checker and JSON to CBOR data driven flow

rozumx:
Work Done
--------
+ Added the JSON schema lint check support
+ Added a JSON schema for the smartnic section
+ Enable lint checking of the smartnic section
+ Enhance code base to data drive setting the JSON format primative from the JSON Schema.  Note: This was needed for boost property trees always write out JSON values as strings.
+ Create additional regression tests
+ Enhance python scripts to indicate when the various sub tests start.
+ Fixed: Renamed SectionAIEResources extention from "cpp" to "cxx"
+ Fixed: Integers greater then 32-bits were not being encoded.
+ Fixed: Node names for files were not being renamed if they were contained directly in an array.
+ Fixed" Rename node "softhub_connections" to "subhubs"
+ Cleaned-up and refactored RapidJsonUtilities, CBOR, and SectionSmartNic sources.

Note: There is still much more refactoring that needs to be done (e.g., moving away from iterators, JSON schema navigation, further code encapsulation, etc.)


src/runtime_src/tools/xclbinutil/RapidJsonUtilities.cxx
```diff
@@ -330,4 +362,161 @@ XclBinUtilities::read_cbor(std::istream& istr,
...
+  const rapidjson::Value::MemberIterator itemsItr = itrObject->value.FindMember("items");
+  if (itemsItr == itrObject->value.MemberEnd())
+    return;
+
+  const rapidjson::Value::MemberIterator typeItr = itemsItr->value.FindMember("type");
+  if (typeItr == itemsItr->value.MemberEnd()) 
+    return;
+
+  // -- array of objects --
+  if (std::string("object") == typeItr->value.GetString()) {
```
stsoe, line 421: early-exit ?
if (x != y)
  return;

But I see it doesn't fit the pattern of extension with value types so this probably shouldn't be changed.

src/runtime_src/tools/xclbinutil/RapidJsonUtilities.cxx
```diff
@@ -330,4 +362,161 @@ XclBinUtilities::read_cbor(std::istream& istr,
...
+  if(sd.Parse(schema.c_str()).HasParseError()) {
+    XUtil::TRACE(boost::str(boost::format("Schema:\n %s") % schema));
+    throw std::runtime_error("Error: The given JSON schema is not valid JSON.");
+  }
+
+  rapidjson::SchemaDocument schemaDoc(sd);        // Convert to a schema document
+  rapidjson::SchemaValidator validator(schemaDoc);
+
+  // Valid the given JSON file
+  if (!doc.Accept(validator)) {
```
stsoe, line 378: early-exit on success ?

src/runtime_src/tools/xclbinutil/RapidJsonUtilities.h
```diff
@@ -28,12 +28,15 @@
...
-  DType_BYTE_STRING
-} DType;
+enum class DType {
+  UNKNOWN,
+  INTEGER,
+  TEXT_STRING,
+  BYTE_STRING,
+  HEX_BYTE_STRING,
+  BYTE_FILE,
+  ENUMERATION,
```
stsoe, line 38: enum classes should use lower case enumerators.   The upper case ones still run then preprocessor risk of conflicting with brain-dead code that #defines say UNKNOWN to something odd.

stsoe (changes requested):
Minor comments

src/runtime_src/tools/xclbinutil/RapidJsonUtilities.cxx
rozumx: Currently this code only support an array of objects.   While not required for 2021.1, I am expected the need to support array of strings and integers.  When that should occur, we will have a set of conditions where early exits would not be possible.  

In other portions of the code, it is indeed written with early exits, but in this piece of code it isn't recommended.

src/runtime_src/tools/xclbinutil/RapidJsonUtilities.cxx
rozumx: Done

src/runtime_src/tools/xclbinutil/RapidJsonUtilities.h
rozumx: done

stsoe (approved):

PR #5008, opened 2021-03-29
Enable Native XRT tracing

jvillarre:
This pull request:

- Enables Native C/C++ API tracing (with the xrt.ini option native_xrt_trace=true)
- Restricts the user from enabling both Native API tracing and HAL level tracing
- Minimizes overhead by removing string manipulation
- Captures the timestamps displayed on the timeline trace as close as possible to the beginning and end of the observed function
- Adds a new set of unsorted events to the XDP database, which are stored in a vector and must be sorted before being dumped to the CSV file.  Currently, only used by the Native XRT events.


stsoe (approved):
Looks good.

PR #5018, opened 2021-03-30
CR-1093128: Fixing race condition in profiling

jvillarre:
This pull request fixes a race condition in the clock training thread used for continuous offload or continuous clock training and the resetting of the trace IP.  It does this by explicitly starting the thread only after we know that reset has been called.


hackwa (approved):
Clock training and race condition changes look great!

IshitaGhosh (approved):
Changes look good.

PR #5019, opened 2021-03-30
CR-1091068 NMI watchdog: BUG: soft lockup on xbutil query

larry9523:
Feature added
1) Don't release BO that holds PS kernel binary if it was not allocated
2) Added check to PS CU number. If it exceeds 128, error out.


mamin506 (approved):

larry9523:
retest this please

PR #5026, opened 2021-03-30
Disabling device continuous offload of trace and continuous dumping for hardware emulation.

jvillarre:
This disables continuous dumping of device files in hardware emulation, since we only read data from hardware emulation once there is no benefit to creating multiple files.


hackwa (approved):
disable continuous offload in hw emulation

PR #5031, opened 2021-03-31
CR-1091068 NMI watchdog: BUG: soft lockup on xbutil query (#5019)

larry9523:
Cherry-pick the fix to validate SCU numbers


larry9523:
retest this please

PR #5033, opened 2021-03-31
CR-1091338: Kernel reads/writes with no end event in trace

jvillarre:
This pull request approximates the end of device data transfer events in trace in two ways:
- When two starts of the same type on an AIM are detected with no end event in between, we assume a dropped end packet and approximate the end of the previous transaction at that point
- If there are any reads or writes that do not have end events at the end of execution, we assume a dropped end packet and approximate the end of the transaction at the end of the CU execution


IshitaGhosh (approved):
Changes look good

PR #5034, opened 2021-03-31
U30 SC related CRs

AShivangi:
- CR-1091088 xbmgmt examine reports SC not up to date. flash --scan --verbose disagrees
- CR-1092220 xbmgmt auto update errors out because of fixed SC
- CR-1091109 xbmgmt2 Bad magic number errors programming U30 with SDK shell

No _WARNING : SC image on the device is not up-to-date_ when SC is FIXED
```
$ /tools/xgs/bin/sudo Debug/opt/xilinx/xrt/bin/xbmgmt examine -d 40:00
1/1 [0000:40:00.0] : xilinx_u30_gen3x4_base_1
Flash properties
  Type                 : qspi_ps_x2_single
  Serial Number        : XFL1LBHB3RSO

  Platform             : xilinx_u30_gen3x4_base_1
  SC Version           : 6.3.5(FIXED)
  Platform ID          : 240E115C-4EE7-87DB-04D5-12C3F5522377
  Interface UUID       : 265FF36F-14F4-A973-3531-A7BF8EE32083

  Platform             : xilinx_u30_gen3x4_base_1
  SC Version           : 6.3.5
  Platform ID          : 0x240e115c4ee787db

```

No _ERROR: Failed to detect XMC, bad magic number: 0_ 
```
$ /tools/xgs/bin/sudo Debug/opt/xilinx/xrt/bin/xbmgmt examine -d 6d:00
---------------------------------------
1/1 [0000:6d:00.0] : xilinx_u30_GOLDEN
---------------------------------------
Flash properties
  Type                 : qspi_ps_x2_single
  Serial Number        :

  Platform             : xilinx_u30_GOLDEN
  SC Version           :
  Platform ID          : N/A

  Platform             : xilinx_u30_gen3x4_base_1
  SC Version           : 6.3.5
  Platform ID          : 0x240e115c4ee787db

WARNING  : Device is not up-to-date.
```

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
@@ -300,14 +300,12 @@ report_status(xrt_core::device_collection& deviceCollection, boost::property_tre
 }
 
 static void 
-isSameShellOrSC(DSAInfo& candidate, DSAInfo& current, bool *same_dsa, bool *same_bmc)
+isSameShellOrSC(const DSAInfo& candidate, const DSAInfo& current, bool& same_dsa, bool& same_bmc)
 {
-    if (!current.name.empty()) {
-        *same_dsa = ((candidate.name == current.name) &&
-            candidate.matchId(current));
-        *same_bmc = (current.bmcVerIsFixed() ||
-            (candidate.bmcVer == current.bmcVer));
-    }
+  if (!current.dsaname().empty()) {
+    same_dsa = ((candidate.dsaname() == current.dsaname()) && candidate.matchId(current));
+    same_bmc = (current.bmcVerIsFixed() || (candidate.bmc_ver() == current.bmc_ver()));
+  }
 }
 
 /* 
@@ -324,7 +322,7 @@ updateShellAndSC(unsigned int  boardIdx, DSAInfo& candidate, bool& reboot)
   bool same_dsa = false;
   bool same_bmc = false;
   DSAInfo current = flasher.getOnBoardDSA();
-  isSameShellOrSC(candidate, current, &same_dsa, &same_bmc);
+  isSameShellOrSC(candidate, current, same_dsa, same_bmc);
 
   // Always update Arista devices
   if (candidate.vendor_id == ARISTA_ID)
--- a/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
@@ -299,6 +299,17 @@ report_status(xrt_core::device_collection& deviceCollection, boost::property_tre
 
 }
 
+static void 
+isSameShellOrSC(DSAInfo& candidate, DSAInfo& current, bool *same_dsa, bool *same_bmc)
+{
+    if (!current.name.empty()) {
+        *same_dsa = ((candidate.name == current.name) &&
+            candidate.matchId(current));
+        *same_bmc = (current.bmcVerIsFixed() ||
+            (candidate.bmcVer == current.bmcVer));
+    }
+}
+
 /* 
  * Flash shell and sc firmware
  * Helper method for auto_flash
@@ -313,21 +324,19 @@ updateShellAndSC(unsigned int  boardIdx, DSAInfo& candidate, bool& reboot)
   bool same_dsa = false;
   bool same_bmc = false;
   DSAInfo current = flasher.getOnBoardDSA();
-  if (!current.name.empty()) {
-    same_dsa = (candidate.name == current.name &&
-      candidate.matchId(current));
+  isSameShellOrSC(candidate, current, &same_dsa, &same_bmc);
 
-    // Always update Arista devices
-    if (candidate.vendor_id == ARISTA_ID)
-        same_dsa = false;
+  // Always update Arista devices
+  if (candidate.vendor_id == ARISTA_ID)
+    same_dsa = false;
 
-    // getOnBoardDSA() returns an empty bmcVer in the case there is no SC,
-    // so do not update
-    if (current.bmcVer.empty())
-        same_bmc = true;
-    else
-        same_bmc = (candidate.bmcVer == current.bmcVer);
-  }
+  // getOnBoardDSA() returns an empty bmcVer in the case there is no SC,
+  // so do not update
+  if (current.bmcVer.empty())
+    same_bmc = true;
+  else
+    same_bmc = (candidate.bmcVer == current.bmcVer);
+  
   if (same_dsa && same_bmc) {
     std::cout << "update not needed" << std::endl;
     return 0;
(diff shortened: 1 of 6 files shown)


src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -299,6 +299,17 @@ report_status(xrt_core::device_collection& deviceCollection, boost::property_tre
 
 }
 
+static void 
+isSameShellOrSC(DSAInfo& candidate, DSAInfo& current, bool *same_dsa, bool *same_bmc)
```
rozumx, line 303: Please pass by reference not via pointers.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -313,21 +324,19 @@ updateShellAndSC(unsigned int  boardIdx, DSAInfo& candidate, bool& reboot)
...
-  }
+  isSameShellOrSC(candidate, current, &same_dsa, &same_bmc);
+
+  // Always update Arista devices
+  if (candidate.vendor_id == ARISTA_ID)
+    same_dsa = false;
+
+  // getOnBoardDSA() returns an empty bmcVer in the case there is no SC,
+  // so do not update
+  if (current.bmcVer.empty())
```
rozumx, line 335:
```suggestion
  same_dsa = current.bmcVer.empty() ? true : (candidate.bmcVer == current.bmcVer);
```

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
rozumx, line 303: Initialize return value to a know known value.

rozumx, line 303: Both `candidate `and `current `should be `const`.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -299,6 +299,17 @@ report_status(xrt_core::device_collection& deviceCollection, boost::property_tre
 
 }
 
+static void 
+isSameShellOrSC(DSAInfo& candidate, DSAInfo& current, bool *same_dsa, bool *same_bmc)
+{
+    if (!current.name.empty()) {
+        *same_dsa = ((candidate.name == current.name) &&
```
rozumx, line 306: We should never be reaching into a class to retrieve its variables.   

Please create a method to retrieve the `name `and `bmvVer `and compare the resulting values.   

Since we now have methods, please update all of the places where the code reaches into this class to observe these values.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -336,12 +345,20 @@ updateShellAndSC(unsigned int  boardIdx, DSAInfo& candidate, bool& reboot)
   if (!same_bmc) {
     std::cout << "Updating SC firmware on device[" << flasher.sGetDBDF() <<
       "]" << std::endl;
-    update_SC(boardIdx, candidate.file);
+    try {
+      update_SC(boardIdx, candidate.file);
+    } catch (const xrt_core::error& e) {
+      std::cout << "NOTE: Skipping SC flash. " << e.what() << std:: endl;
```
rozumx, line 351: It is always good to look at the error being caught if it is being ignored.  If it isn't what is expected then it should be re-thrown.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -336,12 +345,20 @@ updateShellAndSC(unsigned int  boardIdx, DSAInfo& candidate, bool& reboot)
...
+    }
   }
 
   if (!same_dsa) {
     std::cout << boost::format("[%s] : Updating base\n") % flasher.sGetDBDF();
-    update_shell(boardIdx, candidate.file, candidate.file);
+    try {
+      update_shell(boardIdx, candidate.file, candidate.file);
+    } catch (const xrt_core::error& e) {
+      std::cout << "NOTE: " << e.what() << std:: endl;
```
rozumx, line 360: It is always good to look at the error being caught if it is being ignored.  If it isn't what is expected then it should be re-thrown.

src/runtime_src/core/tools/xbmgmt2/flash/firmware_image.h
```diff
@@ -66,6 +66,7 @@ class DSAInfo
     bool matchId(const std::string& id) const;
     bool matchId(DSAInfo& dsa) const;
     bool matchIntId(std::string& id) const;
+    bool bmcVerIsFixed();
```
rozumx, line 69: This method should be `const`.

rozumx (changes requested):
Changes requested.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
AShivangi: That is what is being done here. We want to look at the error produced but we don't want to stop the auto-update. If one device fails, we want to look at it, but let the other devices continue to update

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
AShivangi: I'm not sure what you mean by return a known value. We are getting the return values in the 2 booleans, so the method returns void

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
AShivangi: Going forward, same_bmc will have to support more conditions. If I merge the if statement in this, it will be a very long logic in one line. We can revisit this when I have added the new logic.

rozumx: Understand.  There are trade offs between the code we implement today and adding support for tomorrow.  

For classes, libraries, etc. it makes sense to architect the code base to lead the developers to correctly integrate new functionality.  With regards to methods, these are small enough that if needed it can be refactored completely with new functionality as needed.   

In this case, it would be best to focus the code for ease-of-reading and let the next developer determine how best to add additional conditionals.   In this case the conditional "?" would be a better option.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -336,12 +343,20 @@ updateShellAndSC(unsigned int  boardIdx, DSAInfo& candidate, bool& reboot)
   if (!same_bmc) {
     std::cout << "Updating SC firmware on device[" << flasher.sGetDBDF() <<
       "]" << std::endl;
-    update_SC(boardIdx, candidate.file);
+    try {
+      update_SC(boardIdx, candidate.file);
+    } catch (const xrt_core::error& e) {
+      std::cout << "NOTE: Skipping SC flash. " << e.what() << std:: endl;
```
rozumx, line 349: Hmm...Just noticed this.  Do we wish to have "NOTE:", "Note:", or "Info:"?

rozumx (approved):
Some minor EoU changes that might wished to be considered.

PR #5039, opened 2021-04-01
Fixed issue for new KDS with 32bit zynq

saifuddin-xilinx:
-- For 32bit Zynq new KDS is crashing.
-- On new kds mode we are creating a kernel thread of xrt_cu_polling_thread inside zocl_create_cu(zdev) function under a write lock write_lock(&zdev->attr_rwlock). Which is causing the issue.  
-- Fixed this issue by moving zocl_create_cu(zdev) outside the write lock


saifuddin-xilinx:
Will crate a new one

PR #5043, opened 2021-04-02
Fixed issue for new KDS with 32bit zynq

saifuddin-xilinx:
-- For 32bit Zynq new KDS is crashing.
-- On new kds mode we are creating a kernel thread of xrt_cu_polling_thread inside zocl_create_cu(zdev) function under a write lock write_lock(&zdev->attr_rwlock). Which is causing the issue.
-- Fixed this issue by moving zocl_create_cu(zdev) outside the write lock

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
@@ -722,7 +722,6 @@ zocl_xclbin_read_axlf(struct drm_zocl_dev *zdev, struct drm_zocl_axlf *axlf_obj,
 		vfree(axlf);
 		return -EFAULT;
 	}
-
 	write_lock(&zdev->attr_rwlock);
 
 	/*
@@ -893,16 +892,6 @@ zocl_xclbin_read_axlf(struct drm_zocl_dev *zdev, struct drm_zocl_axlf *axlf_obj,
 		zdev->kernels = kernels;
 	}
 
-	if (kds_mode == 1) {
-		subdev_destroy_cu(zdev);
-		ret = zocl_create_cu(zdev);
-		if (ret)
-			goto out0;
-		ret = zocl_kds_update(zdev);
-		if (ret)
-			goto out0;
-	}
-
 	/* Populating AIE_METADATA sections */
 	size = zocl_read_sect(AIE_METADATA, &zdev->aie_data.data, axlf, xclbin);
 	if (size < 0) {
@@ -935,6 +924,18 @@ zocl_xclbin_read_axlf(struct drm_zocl_dev *zdev, struct drm_zocl_axlf *axlf_obj,
 		goto out0;
 	}
 
+	write_unlock(&zdev->attr_rwlock);
+
+	if (kds_mode == 1) {
+		subdev_destroy_cu(zdev);
+		ret = zocl_create_cu(zdev);
+		if (ret)
+			goto out0;
+		ret = zocl_kds_update(zdev);
+		if (ret)
+			goto out0;
+	}
+
 	zocl_clear_mem(zdev);
 	zocl_init_mem(zdev, zdev->topology);
 
@@ -948,7 +949,6 @@ zocl_xclbin_read_axlf(struct drm_zocl_dev *zdev, struct drm_zocl_axlf *axlf_obj,
 	zocl_xclbin_set_uuid(zdev, &axlf_head.m_header.uuid);
 
 out0:
-	write_unlock(&zdev->attr_rwlock);
 	vfree(aie_res);
 	vfree(axlf);
 	DRM_INFO("%s %pUb ret: %d", __func__, zocl_xclbin_get_uuid(zdev), ret);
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -935,6 +924,18 @@ zocl_xclbin_read_axlf(struct drm_zocl_dev *zdev, struct drm_zocl_axlf *axlf_obj,
...
+		ret = zocl_create_cu(zdev);
+		if (ret)
+			goto out0;
+		ret = zocl_kds_update(zdev);
+		if (ret)
+			goto out0;
+	}
+
 	zocl_clear_mem(zdev);
 	zocl_init_mem(zdev, zdev->topology);
```
mamin506, line 940: This two function need write lock protection as well.

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -935,6 +924,18 @@ zocl_xclbin_read_axlf(struct drm_zocl_dev *zdev, struct drm_zocl_axlf *axlf_obj,
...
+
+	if (kds_mode == 1) {
+		subdev_destroy_cu(zdev);
+		ret = zocl_create_cu(zdev);
+		if (ret)
+			goto out0;
+		ret = zocl_kds_update(zdev);
+		if (ret)
+			goto out0;
+	}
```
mamin506, line 937: Please leave a comment here to explain why this code block should be move out from the write lock region. Then others would not make same mistake and potentially can save a lot of time on debugging.

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -947,11 +948,15 @@ zocl_xclbin_read_axlf(struct drm_zocl_dev *zdev, struct drm_zocl_axlf *axlf_obj,
 	zdev->zdev_xclbin->zx_refcnt = 0;
 	zocl_xclbin_set_uuid(zdev, &axlf_head.m_header.uuid);
 
+	vfree(aie_res);
```
mamin506, line 951: I dislike this duplicated code. Maybe we can use below code flow.

...
write_lock();
// Do everything that should be protected by the write lock

out0:
write_unlock();
if (!ret && kds_mode = 1) {
    // create cu sub-device
}
...

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -722,7 +722,6 @@ zocl_xclbin_read_axlf(struct drm_zocl_dev *zdev, struct drm_zocl_axlf *axlf_obj,
 		vfree(axlf);
 		return -EFAULT;
 	}
-
 	write_lock(&zdev->attr_rwlock);
```
mamin506, line 725: Could you add a comment in before this lock to emphasize that "after this lock to unlock is an atomic context"?

mamin506 (changes requested):

chvamshi-xilinx (approved):
Looks good. Can you please add a clear comment about why you had to make this change.
Like we are calling kthread_run from zocl_create_cu which is causing a kernel panic etc..

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
saifuddin-xilinx: Added a comments. Please check

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
saifuddin-xilinx: Modified accordingly

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
saifuddin-xilinx: Added a comments

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -949,9 +940,22 @@ zocl_xclbin_read_axlf(struct drm_zocl_dev *zdev, struct drm_zocl_axlf *axlf_obj,
...
+			goto out1;
+		ret = zocl_kds_update(zdev);
+		if (ret)
+			goto out1;
+	}
+
+out1:
 	vfree(aie_res);
 	vfree(axlf);
-	DRM_INFO("%s %pUb ret: %d", __func__, zocl_xclbin_get_uuid(zdev), ret);
```
mamin506, line 954: why remove this DRM_INFO?

mamin506 (changes requested):

mamin506:
> Looks good. Can you please add a clear comment about why you had to make this change.
> Like we are calling kthread_run from zocl_create_cu which is causing a kernel panic etc..

I don't think we should comment this directly reason in the code. The root cause is that current context could fall into sleep in a kernel API (kthread_run() in this case). In current code, we have to pay attention when add more code, must be atomic context friendly, in this write lock region.

chvamshi-xilinx:
> > Looks good. Can you please add a clear comment about why you had to make this change.
> > Like we are calling kthread_run from zocl_create_cu which is causing a kernel panic etc..
> 
> I don't think we should comment this directly reason in the code. The root cause is that current context could fall into sleep in a kernel API (kthread_run() in this case). In current code, we have to pay attention when add more code, must be atomic context friendly, in this write lock region.

My point is same as "people has to pay attention when add more code". if we add a comment explicitly, people would think before adding a code. As this comes only for Zynq7000, We have high chances of breaking again.

xilinxgitops:
retest this please

mamin506:
> 
> My point is same as "people has to pay attention when add more code". if we add a comment explicitly, people would think before adding a code. As this comes only for Zynq7000, We have high chances of breaking again.

Agree 👍

mamin506 (approved):

PR #5045, opened 2021-04-02
AIE profile and trace runtime metrics

pgschuey:
Support for runtime-specified metrics for AIE profile and trace


src/runtime_src/core/edge/common/aie_parser.cpp
```diff
@@ -329,16 +397,16 @@ get_trace_gmio(const pt::ptree& aie_meta)
 
 namespace xrt_core { namespace edge { namespace aie {
 
-adf::driver_config
-get_driver_config(const xrt_core::device* device)
+std::vector<std::string>
```
larry9523, line 400: I think this probably revert what @jerryhsu-xilinx has done for common layer code. You need to talk to him about it. Looks to me, this will break the graph control path to use the common code between xrt and adf

larry9523 (changes requested):
Seems to break the graph control of xrt and adf common layer

src/runtime_src/core/edge/common/aie_parser.cpp
pgschuey: We had merged changes, and it looked confusing. I re-arranged the files so our changes look cleaner. Thanks for the heads up.

larry9523 (approved):
The AIE metadata parser part looks good to me.  @jerryhsu-xilinx should take a look at this part too.

jvillarre (approved):
Additions look good on the XDP side.

jerryhsu-xilinx (approved):
aie_parser side looks ok.

hackwa (approved):
Changes look good for now. In future it might a worth looking into configuring both modules of a Tile first rather than configure all core modules first and then all memory modules..

PR #5046, opened 2021-04-02
Profiling device trace improvements

jvillarre:
This pull request enhances device trace in the following ways:
- Ties off any outstanding reads, writes, and stall events when a compute unit ends to approximate end times.  This handles the case when device packets are dropped before we can process them in software.
- Fixes an issue where some of these approximated events had the end events before the start event


hackwa (approved):
stalls and read/write approximation looks good.

IshitaGhosh (approved):
changes look good

PR #5048, opened 2021-04-05
disable full bitstream load in kernel lower then 5.4

Himanshu-xilinx:
* full bitstream load only supported in kernel >=5.4
* dmabuf is not supported by fpga-mgr in kernel lower then 5.4


chvamshi-xilinx (approved):

chvamshi-xilinx:
@larry9523 is on vacation. Merging this as this is just a kernel check and breaking U30,

PR #5050, opened 2021-04-05
return code fix

sarab96:
return code fix

### DIFF
--- a/src/xma/src/xmaplugin/xmaplugin.cpp
+++ b/src/xma/src/xmaplugin/xmaplugin.cpp
@@ -1273,6 +1273,7 @@ int32_t xma_plg_work_item_return_code(XmaSession s_handle, XmaCUCmdObj* cmd_obj_
 
     std::vector<XmaCUCmdObj> cmd_vector(cmd_obj_array, cmd_obj_array+num_cu_objs);
     uint32_t num_errors = 0;
+    uint32_t tmp_index = 0;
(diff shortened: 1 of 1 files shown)


src/xma/src/xmaplugin/xmaplugin.cpp
```diff
@@ -1273,6 +1273,7 @@ int32_t xma_plg_work_item_return_code(XmaSession s_handle, XmaCUCmdObj* cmd_obj_
 
     std::vector<XmaCUCmdObj> cmd_vector(cmd_obj_array, cmd_obj_array+num_cu_objs);
     uint32_t num_errors = 0;
+    uint32_t tmp_index = 0;
```
stsoe, line 1276: Let's iterate cmd_obj_array directly rather than copying-by-value the objects into a std::vector.
You can use std::for_each with a lambda or just a good old for-loop

```
auto end = cmd_obj_array + num_cu_objs;
for (auto itr = cmd_obj_array; itr < end; ++itr) {
   auto& cmd = *itr;
   ...
}

```

Of course it may look nicer if you factor out the lambda into a

stsoe (changes requested):

src/xma/src/xmaplugin/xmaplugin.cpp
sarab96: Updated

stsoe (approved):

PR #5051, opened 2021-04-05
NSPR-9636 AWS V1 needs a single image in the flash device

xuhz:
1. add qspips support to xbflash.qspi
2. deliver identical 2 binaries -- xbflash/xbflash.qspi in a standalone pkg/tarball.
3. the pkg has no dependencies on anything in xrt pkg -- the binaries can be running with no xrt installation
4. the binaries will be compiled statically when possible (cmake >3.18)


xuhz:
extra changes:
1. leave the xbflash.qspi binary in xrt as-is
2. the new pkg only contains 'xbflash'

maxzhen (approved):

xilinxgitops:
retest this please

PR #5054, opened 2021-04-05
cr-1091997

jerryhsu-xilinx:
This CR is to use broadcast to enable AIE cores to minimize start latency between cores.


stsoe (approved):
Looks good to me.

PR #5057, opened 2021-04-06
Revert "cr-1091997"

chvamshi-xilinx:
Reverts Xilinx/XRT#5054

Looks like Build is failing with this change.

I don't see builds passing for the above PR.


chvamshi-xilinx:
Hi @jerryhsu-xilinx , I reverted this change as it is causing some build failures. Please resend your PR after fixing build issue,

PR #5060, opened 2021-04-06
Broadcast enable cores

jerryhsu-xilinx:
This CR is to use broadcast to enable AIE cores to minimize start latency between cores.


stsoe:
ok to test

PR #5071, opened 2021-04-08
use same logic to find the ip_index in both edge vs dc (xclIPName2Index API)

chvamshi-xilinx:
new KDS is not creating the kds_custat sysfs nodes, but xclIPName2Index is using this sysfs node to get the IP index.

We should use same logic between edge vs dc . We may-not require kds_custat sysfs node in new kds as we other sysfs nodes which are used for cu information.

This fixes one of the regression introduced by new KDS.

### DIFF
--- a/src/runtime_src/core/edge/user/shim.cpp
+++ b/src/runtime_src/core/edge/user/shim.cpp
@@ -938,7 +938,10 @@ xclIPName2Index(const char *name)
     return -EINVAL;
   }
   if (buf.empty())
+  {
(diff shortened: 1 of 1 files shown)


maxzhen:
I believe that @mamin506 is fixing this issue by always obtaining CU index from driver?

mamin506 (approved):
Thanks Vamshi to make this change.
As Max said, there is a XRT task to get CU index from driver.
Use same logic in both edge and dc can make my change easier. :)

Himanshu-xilinx (approved):
looks good

PR #5072, opened 2021-04-08
U30 xbutil2 reset

AShivangi:
- CR-1077374 xbutil --new reset errors out
- CR-1090281 xbutil2 - u30 - missing sensor: VCCINT_VCU_0V9
- Addtional notes:
  - Removed AIE reset related code as it is not functional at the moment 
  - Refactored the code to contain pragmas within shim.cpp
```
$ xbutil reset -d 43:00                      
Performing 'HOT Reset' on                            
  -[0000:43:00.1]                                    
Card level reset. This will reset all FPGAs on the card.
Successfully reset Device[0000:43:00.1] 
```


src/runtime_src/core/tools/xbutil2/SubCmdReset.cpp
```diff
@@ -45,7 +46,39 @@ pretty_print_action_list(xrt_core::device_collection& deviceCollection, xrt_core
 static void
 reset_device(std::shared_ptr<xrt_core::device> dev, xrt_core::query::reset_type reset)
 {
-  dev->reset(reset);
+  if (xrt_core::device_query<xrt_core::query::rom_vbnv>(dev).find("_u30_") != std::string::npos) {
+    /*
```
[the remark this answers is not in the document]
rozumx: For comments like this, it is better to use C++ comments (e.g., `//`) and not C comments (e.g., `/* */`).

The reasoning behind this thinking has to do with development.  There are times when developers wish to comment out a section of code and they use C comments to do so.  Unfortunately, nested C comments are supported only with a few compilers.

Note: It could be argued that the developer use a `#if 0   #endif` statement.

src/runtime_src/core/tools/xbutil2/SubCmdReset.cpp
```diff
@@ -45,7 +46,39 @@ pretty_print_action_list(xrt_core::device_collection& deviceCollection, xrt_core
...
+     * when FPGA is ready. so even if there is firewall trip now, we expect
+     * to be able to get S/N again
+     * Having SN info available also implies there is a working SC
+     */
+    std::string sn;
+    try {
+      sn = xrt_core::device_query<xrt_core::query::xmc_serial_num>(dev);
+    } catch (std::exception& ex) {
+      std::cerr << "ERROR:" << ex.what() << std::endl;
+      return;
```
rozumx, line 61: Weird set of method exits.

Sometimes there is an exit early and other times not.  

Also with regards to the catch, is there a reason why it is being caught here as opposed to the calling method.

src/runtime_src/core/tools/xbutil2/SubCmdReset.cpp
```diff
@@ -45,7 +46,39 @@ pretty_print_action_list(xrt_core::device_collection& deviceCollection, xrt_core
...
+    }
+    std::cout << "Card level reset. This will reset all FPGAs on the card." << std::endl;
+  }
+  //xocl reset is done through ioctl 
+  if(reset.get_key() == xrt_core::query::reset_key::user) {
+    try {
+      dev->user_reset(XCL_USER_RESET);
+    } catch(const xrt_core::error& e) {
+      std::cerr << "ERROR: " << e.what() << std::endl;
+    }
```
rozumx, line 77: The above pattern isn't followed.  Shouldn't there be a return here?

If not, then add a comment indicating that this is a "fall through" catch.

src/runtime_src/core/tools/xbutil2/SubCmdReset.cpp
```diff
@@ -80,14 +113,14 @@ SubCmdReset::execute(const SubCmdOptions& _options) const
   XBU::verbose("SubCommand: reset");
   // -- Retrieve and parse the subcommand options -----------------------------
   std::vector<std::string> devices = {"all"};
-  std::string resetType = "hot"; //default val
+  std::string resetType = "user"; //default val
```
rozumx, line 116: Comment here isn't needed. 

If one really wanted to comment that these values are being initialized to default values, then 1 comment prior to devices would be best.  Something like:
`// Initializing variables to their default values`

rozumx (changes requested):
Overall the changes look good.

I do have a few concerns regarding the reset_device() method regarding its flow.  Would like to know more why it was implemented the way it was.

src/runtime_src/core/common/ishim.h
```diff
@@ -391,6 +394,24 @@ struct shim : public DeviceType
...
   }
 
+  virtual void
+  user_reset(xclResetKind kind)
+  {
+    // NOTE: xclResetDevice is a deprecated public API and has been moved to deprecated/xrt.h
+    // However, we can still use it internally
+    // Until the API and the deprecation is removed, we have to use the pragmas
+
+    #ifdef __GNUC__
```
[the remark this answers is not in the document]
AShivangi: I couldn't think of anything else. I talked to Soren about this and he said we will have to use pragmas until we remove this API from deprecated/xrt.h. 
If you have a cleaner idea, I'm all ears

src/runtime_src/core/common/query_reset.h
```diff
@@ -36,7 +36,8 @@ enum class reset_key {
   ert = 3,
   ecc = 4,
   soft_kernel = 5,
-  aie = 6
+  aie = 6,
+  user = 7
```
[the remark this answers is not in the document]
AShivangi: I'm building on ubuntu 16.04 and I get an error if I don't add the comma

src/runtime_src/core/tools/xbutil2/SubCmdReset.cpp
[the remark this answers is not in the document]
AShivangi: Done

src/runtime_src/core/tools/xbutil2/SubCmdReset.cpp
AShivangi: user_reset throws an error, but here, we are catching and "cleanly" printing out the error. I'm missing a return in the catch which I have added.

src/runtime_src/core/tools/xbutil2/SubCmdReset.cpp
AShivangi: Done

src/runtime_src/core/tools/xbutil2/SubCmdReset.cpp
AShivangi: we do throw an error in the calling method, but we should catch that error and cleanly print it out. If we don't catch the error, we get a "header" for the error as well. Unless it is an unexpected error, we don't need the error header.

src/runtime_src/core/tools/xbutil2/SubCmdReset.cpp
```diff
@@ -145,7 +159,17 @@ SubCmdReset::execute(const SubCmdOptions& _options) const
 
   //perform reset actions
   for (const auto & dev : deviceCollection) {
-    reset_device(dev, type);
+    try {
+      reset_device(dev, type);
```
stsoe, line 163: `reset_device(dev.get(), type);`

stsoe (changes requested):
Few comments.  First and foremost no leading _ in function name.
Few others and one to disregard if you prefer.

src/runtime_src/core/tools/xbutil2/SubCmdReset.cpp
```diff
@@ -45,7 +46,22 @@ pretty_print_action_list(xrt_core::device_collection& deviceCollection, xrt_core
 static void
 reset_device(std::shared_ptr<xrt_core::device> dev, xrt_core::query::reset_type reset)
```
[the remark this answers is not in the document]
AShivangi: I'm calling `dev->user_reset(XCL_USER_RESET);` which requires me to pass in a shared dev ptr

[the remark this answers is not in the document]
stsoe: Why does this require a shared_ptr?   Yeah, you cant do 'const' but '`xrt_core::device* dev`' is then what you want.

rozumx (approved):

src/runtime_src/core/tools/xbutil2/SubCmdReset.cpp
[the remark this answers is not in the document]
AShivangi: Oh got it. Fixed

src/runtime_src/core/tools/xbutil2/SubCmdReset.cpp
AShivangi: Fixed

stsoe (approved):

PR #5076, opened 2021-04-08
for xma backward compatibility

sarab96:
for xma backward compatibility

### DIFF
--- a/src/xma/include/app/xmahw.h
+++ b/src/xma/include/app/xmahw.h
@@ -17,6 +17,8 @@
 #ifndef _XMA_HW_H_
 #define _XMA_HW_H_
 
+#include <assert.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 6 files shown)


src/xma/include/app/xmahw.h
```diff
@@ -43,6 +45,7 @@ typedef enum {
   XMA_CMD_STATE_PSK_CRASHED = 7, //PS kernel has crashed
   XMA_CMD_STATE_MAX = 8 // Always the last one
 } XmaCmdState;
+static_assert(sizeof(XmaCmdState) <= 4, "XmaCmdState size must be less than 5 bytes");
```
sonals, line 48: `static_assert(sizeof(XmaCmdState) <= sizeof(int32_t), "XmaCmdState size must be less than 4 bytes");`

sonals (changes requested):
Just update the assert

src/xma/include/app/xmahw.h
stsoe: Move the static_assert to a .cpp file.

sarab96: Done

sonals (approved):

stsoe (approved):

PR #5082, opened 2021-04-10
xclbinutil: Added new options to add/replace and add/merge sections

rozumx:
With the introduction of v++ packaging, the is a need to "add or replace" and "add or merge" a section.   

Unfortunately, the current options only allow a more atomic operation such as adding, removing, replacing, or appending a section.  It doesn't allow for "adding or replacing" or "adding or removing" in the same operation.  To accomplish those operations, requires several calls to xclbinutil to determine the the presents of a section to determine the appropriate options to use. 

This pull request, introduces the following two new options:
- --add-replace-section : If the section is not present it will be added, else the existing section will be replace with the new section.
- --add-merge-section : If the section is not present it will be added. else this new data will be merged with it (if the section supports it)

Work Done
---------
+ Created a new option --add-replace-section
+ Created a new option --add-merge-section
+ Added support to add and/or remove sections
+ Added support to add and/or merge sections
+ Added unit tests to validate correct operations of this new functionality 
+ Refactored some methods to use std::ostream classes instead of std::fstream. Doing so allows for writing to an std::ostreamstream.

### DIFF
--- a/src/runtime_src/tools/xclbinutil/XclBinClass.cxx
+++ b/src/runtime_src/tools/xclbinutil/XclBinClass.cxx
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2018-2020 Xilinx, Inc
+ * Copyright (C) 2018-2021 Xilinx, Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License"). You may
  * not use this file except in compliance with the License. A copy of the
--- a/src/runtime_src/tools/xclbinutil/XclBinClass.cxx
+++ b/src/runtime_src/tools/xclbinutil/XclBinClass.cxx
@@ -214,7 +214,7 @@ XclBin::addHeaderMirrorData(boost::property_tree::ptree& _pt_header) {
 
 
 void
-XclBin::writeXclBinBinaryHeader(std::fstream& _ostream, boost::property_tree::ptree& _mirroredData) {
+XclBin::writeXclBinBinaryHeader(std::ostream& _ostream, boost::property_tree::ptree& _mirroredData) {
   // Write the header (minus the section header array)
   XUtil::TRACE("Writing xclbin binary header");
   _ostream.write((char*)&m_xclBinHeader, sizeof(axlf) - sizeof(axlf_section_header));
@@ -229,7 +229,7 @@ XclBin::writeXclBinBinaryHeader(std::fstream& _ostream, boost::property_tree::pt
 
 
 void
-XclBin::writeXclBinBinarySections(std::fstream& _ostream, boost::property_tree::ptree& _mirroredData) {
+XclBin::writeXclBinBinarySections(std::ostream& _ostream, boost::property_tree::ptree& _mirroredData) {
   // Nothing to write
   if (m_sections.empty()) {
     return;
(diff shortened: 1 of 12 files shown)


src/runtime_src/tools/xclbinutil/XclBinClass.cxx
```diff
@@ -587,6 +587,23 @@ XclBin::addSection(Section* _pSection) {
...
 }
 
+void 
+XclBin::addReplaceSection(ParameterSectionData &_PSD)
+{
+  enum axlf_section_kind eKind;
+  if (Section::translateSectionKindStrToKind(_PSD.getSectionName(), eKind) == false) {
+    std::string errMsg = XUtil::format("ERROR: Section '%s' isn't a valid section name.", _PSD.getSectionName().c_str());
+    throw std::runtime_error(errMsg);
+  }
```
stsoe, line 597: This looks odd to me.  Why is translateSectionKindStrToKind returning an error code that has to be checked as opposed to itself throwing on error and return kind on success?

src/runtime_src/tools/xclbinutil/XclBinUtilMain.cxx
```diff
@@ -367,6 +369,11 @@ int main_(int argc, const char** argv) {
       inputFiles.push_back(psd.getFile());
     }
 
+    for (auto section : sectionsToAddReplace) {
```
stsoe, line 372: auto& (or const auto&) ?

src/runtime_src/tools/xclbinutil/XclBinUtilMain.cxx
```diff
@@ -478,6 +485,12 @@ int main_(int argc, const char** argv) {
   for (auto section : sectionsToRemove) 
     xclBin.removeSection(section);
 
+   // -- Add or Replace Sections --
+  for (auto section : sectionsToAddReplace) {
```
stsoe, line 489: ditto

src/runtime_src/tools/xclbinutil/XclBinUtilMain.cxx
```diff
@@ -478,6 +485,12 @@ int main_(int argc, const char** argv) {
...
+    ParameterSectionData psd(section);
+    xclBin.addReplaceSection( psd );
+  }
+
   // -- Replace Sections --
   for (auto section : sectionsToReplace) {
```
stsoe, line 495: ditto

stsoe (changes requested):

src/runtime_src/tools/xclbinutil/XclBinClass.cxx
rozumx: Agree, the existing code was odd. 

I have refactored the code base so that `translateSectionKindStrToKind()` now throws an exception and no longer returns a boolean value.   The code is indeed cleaner.

src/runtime_src/tools/xclbinutil/XclBinUtilMain.cxx
rozumx: Issue fixed with `const auto &` here and other parts of the code base that has a similar pattern.

src/runtime_src/tools/xclbinutil/XclBinUtilMain.cxx
rozumx: done

src/runtime_src/tools/xclbinutil/XclBinUtilMain.cxx
rozumx: fixed

stsoe (approved):

PR #5089, opened 2021-04-13
deprecate continuous_offload_interval_ms and rename trace dump interval flag

hackwa:


jvillarre (approved):
Renaming the ini file flags looks good.

PR #5093, opened 2021-04-14
xclbinutil : Added intelligent SmartNic section merging support

rozumx:
Work Done
---------
+ Created an intelligent boost property tree merge support for the SMARTNIC section.  This includes: array merges (with and without keys), validation of common schemas, data driven call back functionality.
+ Created several regression tests mimicking how Vitis would use this flow.


stsoe (approved):
Looks good.

PR #5097, opened 2021-04-14
CR-1092167 when the debug_mode is set to OFF, corrected not to set th…

venkatp-xilinx:
CR-1092167 when the debug_mode is set to OFF, corrected not to set th… ENVs like VITIS_WAVEFORM


xilinxgitops:
retest this please

PR #5098, opened 2021-04-14
Fixed CR-1093799 issue : ASTeR TC 19.03 - u30 - validate intermittently fails with "VCU send command failed"

saifuddin-xilinx:
Fixed CR-1093799 issue.
-- This issue is for multi device 
-- Issue is with encoder code.

### DIFF
--- a/tests/validate/xcl_vcu_test/src/plugin_dec.cpp
+++ b/tests/validate/xcl_vcu_test/src/plugin_dec.cpp
@@ -237,7 +237,7 @@ gstivas_xvcudec_close (XrtIvas_XVCUDec *dec)
   xvcudec_free_internal_buffers (dec);
 
   /* Close XRT context */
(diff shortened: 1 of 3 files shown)


mamin506:
Hi @saifuddin-xilinx, I see this error in dmesg "[ 6577.889887] xocl 0000:66:00.1:  ffff888859e5d0b0 kds_del_cu_context: CU(0) has never been reserved". This is because the host application is trying to close a wrong CU index.

Then I found "xclCloseContext (priv->xcl_handle, priv->xclbinId, 0);" in plugin_dec.cpp. It should not hardcode CU 0 in this place. Could you fix this issue as well?

saifuddin-xilinx:
Hi @mamin506, 
Thanks for the comments. I have updated the same.

chvamshi-xilinx (approved):

PR #5105, opened 2021-04-15
xbutil new command vs old command mapping related, link added, and simple filename changed

uday610:
xbutil new command vs old command mapping related, link added, and simple filename changed


maxzhen:
ok to test

PR #5108, opened 2021-04-15
updated iteration cnt for validate designs and provide env variable support

vishnuchebrolu:


tests/validate/bandwidth_test/src/host.cpp
```diff
@@ -124,8 +124,15 @@ int main(int argc, char** argv) {
     }
 
     double max_throughput = 0;
+    char* env_reps_cnt = getenv("XRT_VALIDATE_ITER_CNT");
+    int reps = 10000;
+    if (env_reps_cnt) {
+        reps = atoi(env_reps_cnt);
+        std::cout << "INFO : we found command line argument “XRT_VALIDATE_ITER_CNT”=" << reps
+                  << " and using the same for kernel iteration.\n";
+    }
```
stsoe, line 133: Should this env var be a command line argument instead?

tests/validate/slavebridge_test/src/host.cpp
```diff
@@ -125,8 +125,15 @@ int main(int argc, char** argv) {
...
 
     double max_throughput = 0;
+    char* env_reps_cnt = getenv("XRT_VALIDATE_ITER_CNT");
+    int reps = 10000;
+    if (env_reps_cnt) {
+        reps = atoi(env_reps_cnt);
+        std::cout << "INFO : we found command line argument “XRT_VALIDATE_ITER_CNT”=" << reps
+                  << " and using the same for kernel iteration.\n";
+    }
+
```
stsoe, line 135: Ditto

tests/validate/bandwidth_test/src/host.cpp
AShivangi: I had the same concern so we had a discussion via email. Here is what I got: "environmental variable is intended to be for internal test only and undocumented from a customer perspective". If you're interested in further details, I can add you to the email change.

AShivangi (approved):

tests/validate/bandwidth_test/src/host.cpp
stsoe: That's what hidden options are for.    Can we please change?

stsoe (changes requested):

tests/validate/bandwidth_test/src/host.cpp
vishnuchebrolu: Updated.

tests/validate/slavebridge_test/src/host.cpp
vishnuchebrolu: updated.

stsoe (approved):

PR #5111, opened 2021-04-16
CR-1091955 See "soft lockup " on RH78 management VM

larry9523:
When loading XCLBIN, there is a race condition for host xclmgmt driver to wait for IDLE (single packet) and then DONE (whole xclbin blob). In a corner case that IDLE and then DONE are set by device zocl driver before xclmgmt start to wait for IDLE. So host driver will never get IDLE signal.

The fix is, actually, for last packet, host driver does not need to wait for IDLE because no new data need to be transferred. Host can wait for done directly.


xdavidz (approved):
I assume you have tested on both u30 and versal.

PR #5116, opened 2021-04-19
CR-1089852: Moving xclIPName2Index to xrt.h from experimental/xrt-next.h

Himanshu-xilinx:


chvamshi-xilinx (approved):

maxzhen:
@Himanshu-xilinx , why are we doing this? We are still extending HAL layer API? I'd thought that we only promote xrt native APIs, no? @stsoe

stsoe:
I have the same question as @maxzhen

chvamshi-xilinx:
Hi @maxzhen & @stsoe , There was request from GStreamer team to move this API to xrt.h.  Me/Sonal/you decided to move this to xrt.h as this API is no-longer experimental in 2020.2. We didn't get a chance to do that in 2020.2, We are moving it now.  They are using HAL APIs in their plugins.

maxzhen (approved):

stsoe (approved):

PR #5124, opened 2021-04-20
CR-1095761: Fixing missing compute unit utilization information in OpenCL profile summary

jvillarre:
In order to properly print out compute unit utilization, the following two changes needed to be made:

1. Added statistics logging when approximating CU end events.  Previously we added the approximate end events but did not update the statistics so we would not print out information on those compute unit executions.
2. Force the reading and processing of trace to happen before the dumping of the summary file if possible.  Since the different plugins can be destroyed at the end of execution in non-deterministic order, this pull request adds an explicit broadcast call to force the reading of trace if it hadn't already been done at the end of execution

### DIFF
--- a/src/runtime_src/xdp/profile/database/database.h
+++ b/src/runtime_src/xdp/profile/database/database.h
@@ -40,7 +40,7 @@ namespace xdp {
   {
   public:
     // For messages sent to specific plugins
-    enum MessageType { READ_COUNTERS } ;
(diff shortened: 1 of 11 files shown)


dayeh-xilinx:
retest this please. Pipeline boardfarm test issue fixed.

IshitaGhosh (approved):
Changes look good.

PR #5126, opened 2021-04-20
last error header changes-1

sarab96:
last error header changes-1
Adding PID field as well

### DIFF
--- a/src/runtime_src/core/include/xclerr.h
+++ b/src/runtime_src/core/include/xclerr.h
@@ -39,6 +39,10 @@
 #ifndef XCLERR_H_
 #define XCLERR_H_
 
+#include "xrt_error_code.h"
+
+#define	XOCL_ERROR_CAPACITY	16
+
 /**
  * enum xclFirewallID - AXI Firewall IDs used to identify individual AXI Firewalls
  *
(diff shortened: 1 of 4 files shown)


larry9523 (commented):
Since there is no code use these structure, so it is not clear to me what is the plan. In zocl, there is similar data structure which are used by kernel driver only. So I am wondering if xclerr.h is the right place for them. Should they be put to a common header that shared by zocl and xocl/xclmgmt in kernel space only?

src/runtime_src/core/include/xclerr.h
```diff
@@ -88,4 +92,22 @@ struct xclErrorStatus {
...
+	xrtErrorTime	ts;		/* 64 bits; timestamp */
+	unsigned        pid;            /* 32 bits; pid associated with error, if available */
+};
+
+struct xocl_errors {
+	int		num_err;	/* number of errors recorded */
+	int		cap;		/* capacity of current error array */
+	struct xclErrorLast *errors;	/* error array pointer */
+};
+
```
stsoe, line 112: What is the intended end-use of these structs?  I am concerned that these structs are added to an exported public header file and that we will have application code dependencies on struct and their sizes.

Please provide the bigger picture here in terms of how these structs will be used and by whom.

stsoe (changes requested):

src/runtime_src/core/include/xclerr.h
sarab96: updated based on discussion

sarab96:
updated based on discussion

stsoe (approved):

src/runtime_src/core/include/xclerr_int.h
```diff
@@ -0,0 +1,70 @@
...
+struct xclErrorLast {
+	xrtErrorCode	zer_err_code;	/* 64 bits; XRT error code */
+	xrtErrorTime	zer_ts;		/* 64 bits; timestamp */
+	unsigned        pid;            /* 32 bits; pid associated with error, if available */
+};
+
+typedef struct xocl_errors {
+	int		num_err;	/* number of errors recorded */
+	int		cap;		/* capacity of current error array */
+	struct xclErrorLast *errors;	/* error array pointer */
```
larry9523, line 67: This struct is going to be interpreted from a sysfs node binary format, right? I assume it should be a contiguous chunk of  memory. Then struct xclErrorLast errors[1] seems better than a pointer, right?

larry9523 (commented):
see inline comment

larry9523 (approved):

src/runtime_src/core/include/xclerr_int.h
sarab96: correct. updated

sarab96:
updated
> see inline comment

maxzhen (approved):

PR #5127, opened 2021-04-20
create CU subdev for ap_ctrl_none cu

mamin506:
1. Create CU sub-device for ap_ctrl_none cu. This would allow user to use xclRegRead/xclRegWrite to access CU
2. Submit command to ap_ctrl_none CU through KDS is an invalid use case. xclExecBuf would return error.


maxzhen (approved):

dayeh-xilinx:
retest this please. Pipeline boardfarm test issue fixed.

PR #5133, opened 2021-04-21
Bug Fixes: Improved xbutil validate summary report and help

rozumx:
Work done
---------
+ EoU : Added summary of validation tests produced across multiple devices (CR-1092567)
+ EoU : Added quotes around report names in help (CR-1096035)
+ EoU:  Update default xbutil examine reports when using and not using the device option (CR-1096013)
            * If the user doesn't specify one, the default host report will be created.  
            * If the user specifies a device, then the platform and compute-units report will be produced.

### DIFF
--- a/src/runtime_src/core/tools/common/XBHelpMenus.cpp
+++ b/src/runtime_src/core/tools/common/XBHelpMenus.cpp
@@ -536,7 +536,7 @@ XBUtilities::create_suboption_list_string(const VectorPairStrings &_collection)
       pairs.first += "\'";     
     }
 
-    maxStringLength = std::max<unsigned int>(maxStringLength, pairs.first.length());
+    maxStringLength = std::max<unsigned int>(maxStringLength, static_cast<unsigned int>(pairs.first.length()));
   }
 
   const unsigned int indention = maxStringLength + 5;  // New line indention after the '-' character (5 extra spaces)
--- a/src/runtime_src/core/tools/common/XBHelpMenus.cpp
+++ b/src/runtime_src/core/tools/common/XBHelpMenus.cpp
@@ -536,7 +536,7 @@ XBUtilities::create_suboption_list_string(const VectorPairStrings &_collection)
       pairs.first += "\'";     
     }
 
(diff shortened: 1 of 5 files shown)


src/runtime_src/core/tools/common/XBHelpMenus.cpp
```diff
@@ -524,17 +524,29 @@ XBUtilities::create_suboption_list_string(const VectorPairStrings &_collection)
...
   unsigned int maxStringLength = 0;
-  for (const auto & pairs : _collection)
+  for (auto & pairs : workingCollection) {
+    // Determine if the keyName needs to have 'quotes', if so add them
+    if (pairs.first.find(' ') != std::string::npos ) {
+      pairs.first.insert(0, 1, '\'');  
+      pairs.first += "\'";     
+    }
+
     maxStringLength = std::max(maxStringLength, (unsigned int) pairs.first.length());
```
stsoe, line 539: `std::max<unsigned int>(maxStringLength, pairs.first.length())`

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -1306,6 +1307,154 @@ run_tests_on_devices( xrt_core::device_collection &deviceCollection,
   }
 }
 
+static
+std::string quote_name(const std::string name)
```
stsoe, line 1311: const std::string&

stsoe (commented):
Two comments.

src/runtime_src/core/tools/common/XBHelpMenus.cpp
rozumx: Addressed

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
rozumx: Fixed

AShivangi (approved):
Thank you for making these changes

PR #5135, opened 2021-04-22
CR-1082552: Change argument name size from 32 to 64

Himanshu-xilinx:


chvamshi-xilinx (approved):

chvamshi-xilinx:
We need to give clear error message when Argument name is morethan 64 characters.  You can send another PR for that.

mamin506 (approved):

larry9523:
While we don't commit the ioctl interface between userspace and kernel driver, we should always be cautious when changing it, especially the size of the argument structure. There are all different XRT users and not everybody update user library and kernel driver all together (on edge, some folks have their own xrt/zocl reciepe).

Also it would be nice we broadcast a flag day message when interfaces like this change. It is very hard to figure out what is going on when this kind of mismatch happens.

chvamshi-xilinx:
> While we don't commit the ioctl interface between userspace and kernel driver, we should always be cautious when changing it, especially the size of the argument structure. There are all different XRT users and not everybody update user library and kernel driver all together (on edge, some folks have their own xrt/zocl reciepe).
> 
> Also it would be nice we broadcast a flag day message when interfaces like this change. It is very hard to figure out what is going on when this kind of mismatch happens.

Sure Larry. will inform in xrt_general from next time. Anyone faced problem ?

PR #5136, opened 2021-04-22
Various u30 related CR fixes

AShivangi:
- CR-1096199 Crash programming the U30 from Beta -> GA_internal_bash
- CR-1096037 replace "invalided" with "invalid"
- CR-1096026 Add a space before [ in Validate device[0000:04:00.1]
- CR-1096025 Tracking CR: Don't advertise missing tests in the default validate
- CR-1096008 xbutil examine does not show state of card without deployment platform
- CR-1095169 Please enumerate the platform_id UUID in xbmgmt examine

### DIFF
--- a/src/runtime_src/core/tools/common/XBHelpMenus.cpp
+++ b/src/runtime_src/core/tools/common/XBHelpMenus.cpp
@@ -690,12 +690,16 @@ XBUtilities::produce_reports( xrt_core::device_collection _devices,
         
         //if factory mode
         std::string platform = "";
+        try {
           if (is_mfg) {
             platform = "xilinx_" + xrt_core::device_query<xrt_core::query::board_name>(device) + "_GOLDEN";
           }
           else {
             platform = xrt_core::device_query<xrt_core::query::rom_vbnv>(device);
           }
+        } catch(...) {
+          // proceed even if the platform name is not available
+        }
         std::string dev_desc = (boost::format("%d/%d [%s] : %s\n") % ++dev_idx % _devices.size() % ptDevice.get<std::string>("device_id") % platform).str();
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -1113,17 +1113,17 @@ static std::vector<TestCollection> testSuite = {
  * print basic information about a test
  */
 static void
-pretty_print_test_desc(const boost::property_tree::ptree& test, int test_idx,
-                       size_t testSuiteSize, std::ostream & _ostream, const std::string& bdf)
+pretty_print_test_desc(const boost::property_tree::ptree& test, int& test_idx,
```
rozumx, line 1116: How the variable `test_index` is being used here is unfortunately a kludge, please don't do this.  

In addition how this function is being used doesn't look right.  When you get a chance, please ping me so that we can go over this code.

rozumx (changes requested):

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
AShivangi: The refactoring requires a lot more work than anticipated. We will revisit this after finishing the U30 work

src/runtime_src/core/tools/common/XBHelpMenus.cpp
```diff
@@ -690,12 +690,16 @@ XBUtilities::produce_reports( xrt_core::device_collection _devices,
         
         //if factory mode
         std::string platform = "";
```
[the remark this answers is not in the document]
AShivangi: done

rozumx (approved):

PR #5144, opened 2021-04-23
Add support for vck190_pci platform

rbramand-xilinx:
> Added changes to support vck190 pci based platform
> Made changes to devices.h file as it is using legacy platforms flow
> Enabled flash feature which utilises ospi_versal subdev flash mechanism.
> Enabled programming aie only xclbin which utilises download axlf feature of ospi_versal subdev

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
@@ -24,6 +24,15 @@
 
 extern int kds_mode;
 
+static int zocl_load_aie_only_pdi(struct drm_zocl_dev *zdev, struct axlf *axlf,
+			char __user *xclbin, struct sched_client_ctx *client);
+
+static bool 
+is_aie_only(struct axlf *axlf)
+{
+        return (axlf->m_header.m_actionMask & AM_LOAD_AIE);
+}
+
 static int
 zocl_fpga_mgr_load(struct drm_zocl_dev *zdev, const char *data, int size, u32 flags)
 {
@@ -441,6 +450,7 @@ zocl_xclbin_load_pdi(struct drm_zocl_dev *zdev, void *data)
 	struct axlf *axlf_head = axlf;
 	char *xclbin = NULL;
 	char *section_buffer = NULL;
+	void *aie_res = 0;
 	size_t size_of_header;
 	size_t num_of_sections;
 	uint64_t size = 0;
@@ -453,25 +463,46 @@ zocl_xclbin_load_pdi(struct drm_zocl_dev *zdev, void *data)
 	}
 
 	mutex_lock(&zdev->zdev_xclbin_lock);
-	/* Check unique ID */
-	if (zocl_xclbin_same_uuid(zdev, &axlf_head->m_header.uuid)) {
-		DRM_INFO("%s The XCLBIN already loaded, uuid: %pUb",
-			 __func__, &axlf_head->m_header.uuid);
-		mutex_unlock(&zdev->zdev_xclbin_lock);
-		return ret;
-	}
 
-	write_lock(&zdev->attr_rwlock);
-
-	/* Get full axlf header */
 	size_of_header = sizeof(struct axlf_section_header);
 	num_of_sections = axlf_head->m_header.m_numSections-1;
 	xclbin = (char __user *)axlf;
-	ret =
-	    !ZOCL_ACCESS_OK(VERIFY_READ, xclbin, axlf_head->m_header.m_length);
-	if (ret) {
-		ret = -EFAULT;
+
+	ret = !ZOCL_ACCESS_OK(VERIFY_READ, xclbin, axlf_head->m_header.m_length);
+        if (ret) {
+                ret = -EFAULT;
+                goto out;
+        }
+
+	write_lock(&zdev->attr_rwlock);
+
+	zocl_read_sect(AIE_RESOURCES, &aie_res, axlf, xclbin);
+
+	/* Check unique ID */
+	if (zocl_xclbin_same_uuid(zdev, &axlf_head->m_header.uuid)) {
+		if (is_aie_only(axlf)) {
+                        ret = zocl_load_aie_only_pdi(zdev, axlf, xclbin, NULL);
+                        if (ret)
+                                DRM_WARN("read xclbin: fail to load AIE");
+                        else
+                                zocl_create_aie(zdev, axlf, aie_res);
+		} else {
+			DRM_INFO("%s The XCLBIN already loaded, uuid: %pUb",
+				 __func__, &axlf_head->m_header.uuid);
+		}
 		goto out;
+	}
+
+	if (is_aie_only(axlf)) {
+                ret = zocl_load_aie_only_pdi(zdev, axlf, xclbin, NULL);
+                if (ret)
+                        goto out;
+        }
+
+	size = zocl_read_sect(AIE_METADATA, &zdev->aie_data.data, axlf, xclbin);
+	if (size > 0 ) {
+		zdev->aie_data.size = size;
+		zocl_create_aie(zdev, axlf, aie_res);
 	}
 
 	size = zocl_offsetof_sect(BITSTREAM_PARTIAL_PDI, &section_buffer,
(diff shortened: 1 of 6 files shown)


src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -453,27 +463,48 @@ zocl_xclbin_load_pdi(struct drm_zocl_dev *zdev, void *data)
...
 	size_of_header = sizeof(struct axlf_section_header);
 	num_of_sections = axlf_head->m_header.m_numSections-1;
 	xclbin = (char __user *)axlf;
-	ret =
-	    !ZOCL_ACCESS_OK(VERIFY_READ, xclbin, axlf_head->m_header.m_length);
-	if (ret) {
-		ret = -EFAULT;
+
+	ret = !ZOCL_ACCESS_OK(VERIFY_READ, xclbin, axlf_head->m_header.m_length);
+        if (ret) {
```
xdavidz, line 472: indent?

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -453,27 +463,48 @@ zocl_xclbin_load_pdi(struct drm_zocl_dev *zdev, void *data)
...
+                        goto out;
+        }
+
+	size = zocl_read_sect(AIE_METADATA, &zdev->aie_data.data, axlf, xclbin);
+	if (size > 0 ) {
+		zdev->aie_data.size = size;
+		zocl_create_aie(zdev, axlf, aie_res);
+	}
+
 	size = zocl_offsetof_sect(BITSTREAM_PARTIAL_PDI, &section_buffer,
```
xdavidz, line 508: seems you can move this 
if (aie_only)
   read AIE, then load AIE into a function

Thus we can avoid dup code.

rbramand-xilinx: @xdavidz made changes as suggested

xdavidz (approved):

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
chvamshi-xilinx: Hi @rbramand-xilinx , i don't see this change yet. Can you please make this change to avoid dup code.

rbramand-xilinx: @chvamshi-xilinx made changes as requested

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -453,27 +463,40 @@ zocl_xclbin_load_pdi(struct drm_zocl_dev *zdev, void *data)
...
-	ret =
-	    !ZOCL_ACCESS_OK(VERIFY_READ, xclbin, axlf_head->m_header.m_length);
+
+	ret = !ZOCL_ACCESS_OK(VERIFY_READ, xclbin, axlf_head->m_header.m_length);
 	if (ret) {
 		ret = -EFAULT;
 		goto out;
 	}
 
+	write_lock(&zdev->attr_rwlock);
```
larry9523, line 477: any reason you need to move the acquiring write_lock down to here? 
And, by doing this, you will need to take care of line 474 (goto out). We will release write lock there.

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -453,27 +463,40 @@ zocl_xclbin_load_pdi(struct drm_zocl_dev *zdev, void *data)
...
+		ret = zocl_load_aie_only_pdi(zdev, axlf, xclbin, NULL);
+		if (ret) {
+			DRM_WARN("read xclbin: fail to load AIE");
+			goto out;
+		}
+		size = zocl_read_sect(AIE_METADATA, &zdev->aie_data.data, axlf, xclbin);
+		if (size > 0 ) {
+			zdev->aie_data.size = size;
+			zocl_create_aie(zdev, axlf, aie_res);
+		}
```
larry9523, line 490: Minor thing: we probably don't need to check if there is AIE_METADAT here. It will be checked inside ocl_create_aie so that we have a centralize place to do the sanity check.
However, more important, zocl_create_aie was added and tested for edge flow only so far. We may need to do the same thing for PCIe as well but I suggest we do it in a separate PR where we have a clear use case for zocl to request AIE partition to support control AIE from PS kernel and related support. This PR is for load/reload AIE PDI only, right?

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -24,6 +24,15 @@
 
 extern int kds_mode;
 
+static int zocl_load_aie_only_pdi(struct drm_zocl_dev *zdev, struct axlf *axlf,
+			char __user *xclbin, struct sched_client_ctx *client);
```
larry9523, line 28: We can move the function a head of zocl_xclbin_load_pdi so that we don't need to declare it here. This seems to be convention of this implementation.

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
```diff
@@ -1077,6 +1077,7 @@ static void xclmgmt_extended_probe(struct xclmgmt_dev *lro)
 
 	if (!(dev_info->flags & XOCL_DSAFLAG_DYNAMIC_IP) &&
 	    !(dev_info->flags & XOCL_DSAFLAG_SMARTN) &&
+	    !(dev_info->flags & XOCL_DSAFLAG_VERSAL_EDGE) &&
```
larry9523, line 1080: Can you please explain why do we need this flag?

Basically, I see you treat this new platform as regular VCK5000. But use this flag to make it special. What is the reason do we need to treat it specially? Is it because we don't have DMA IP where we can config the MSIX? If so why can't we support it here? Do we plan to have user function on this platform later? Do we need PF mail box between the two functions? 
I am trying to understand if the  XOCL_DSAFLAG_VERSAL_EDGE is just a temporary flag until we have MSI-X support. Or something will be fundamentally different between VCK5000 and this platform?

src/runtime_src/core/pcie/driver/linux/xocl/xocl_xclbin.c
```diff
@@ -53,7 +59,9 @@ static int versal_xclbin_download(xdev_handle_t xdev, void *args)
 
 	BUG_ON(!arg->xclbin);
 
-	if (xclbin->m_header.m_mode == XCLBIN_FLAT) {
+	/*aie only xclbin needs to be downloaded from host we should do it even
+	 for the falt shells */
```
larry9523, line 63: Please correct the comment c-style
/* 
 * 
 *
*/

src/runtime_src/core/pcie/driver/linux/xocl/xocl_xclbin.c
```diff
@@ -32,6 +32,12 @@ static int versal_xclbin_pre_download(xdev_handle_t xdev, void *args)
 	int ret = 0;
 
 	ret = xrt_xclbin_get_section(xclbin, PARTITION_METADATA, &metadata, &size);
+
+	/* aie only xclbin for edge_pcie platforms will not have partition metadata
+	   we should not error out by checking partition metadata */
+	if (xclbin->m_header.m_actionMask & AM_LOAD_AIE)
+		ret = 0;
```
larry9523, line 39: This is a bit confusing. Can we just move this check ahead and if Action Mask is set, just return. 
In AIE only XCLBIN, there will not be PARTITION_METADATA anyway. And we don't expect to tear down any URP sub driver either.

larry9523 (changes requested):
We probably need an off line sync up on the platform definition and its name. We introduce a new term VERSAL_EDGE. We need to make sure we are naming it correctly. @sonals

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
rbramand-xilinx: yes this flag is added due to no DMA IP where we configure MSIX, this is custom platform which customer has created they are going to enable DMA in next update of platform.
We are planning to use user functions also in this platform but no communication is required, as of now they are using userpf mainly to get aie errors back to host

larry9523 (approved):

chvamshi-xilinx (approved):

PR #5146, opened 2021-04-23
Fix error condition for mKernelFD

fabwu:
If the mKernelFD doesn't exist, no error message is shown because open() returns -1 instead of null. Fix this by checking for -1.


sonals:
ok to test

stsoe (changes requested):
@fabwu Thank you for fixing this.  Can you please sign your commit per https://github.com/Xilinx/XRT/blob/master/CONTRIBUTING.rst

stsoe (approved):

chvamshi-xilinx (approved):

PR #5147, opened 2021-04-23
CR-1094989: Fix continuous offload freeze in device trace offload

jvillarre:
This pull request addresses a condition where the circular buffer used in trace is overwritten and we have to stop offloading trace from the device during the execution.  Previously, there was a deadlock due to the stop_offload function not checking if the state was already STOPPED.
This also adds a warning to the user when the device offload stops due to circular buffer overflow.


hackwa (approved):
looks good!

PR #5148, opened 2021-04-23
Various Bug Fixes: Added global force option & temporarily addressed validation EoU output

rozumx:
Work Done
---------
+ Added a global force option
+ Removed "false" exception regarding using the "-f" option (CR 1096041)
+ Added support to overwrite existing files with the --force option (CR 1096038)
+ Added temporary fix for to indicate work is being done when validation is be performed when using the --output option (CR 1096051)

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2020 Xilinx, Inc
+ * Copyright (C) 2020-2021 Xilinx, Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License"). You may
  * not use this file except in compliance with the License. A copy of the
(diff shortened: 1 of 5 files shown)


src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -658,13 +654,16 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
...
     std::string empty = "";
     if(update.compare("all") == 0)
-      auto_flash(deviceCollection, force);
-    else if(update.compare("flash") == 0)
-      throw xrt_core::error("Platform only update is not supported");
-    else if(update.compare("sc") == 0)
-      throw xrt_core::error("SC only update is not supported");
-    else 
+      auto_flash(deviceCollection, XBU::getForce());
+    else {
```
AShivangi, line 658: why the split?

AShivangi (approved):
Looks good! Thanks for making these changes

Sample output:
```
$ xbutil validate -d all -r quick -o test.txt
Info: The output is being redirected to the file: 'test.txt'
Info: Validation started ... 
Info: Validation completed


$ xbutil validate -d all -r quick -o test.txt
ERROR: Output file already exists: 'test.txt': Invalid argument
...


$ xbutil validate -d all -r quick -o test.txt --force
Info: The output is being redirected to the file: 'test.txt'
Info: Validation started ... 
Info: Validation completed
```

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
rozumx: A ladder of `else if` statements causes obfuscation and it if preferred to not have them.  So I refactored the code to see what the intent of the code base was.  In the process, I discovered that "flash" and "sc" are not needed and will be removing this code in the next code base.

PR #5149, opened 2021-04-23
XRT-894 simplify QSFP sysfs interface

xdavidz:
make QSFP interface much more simpler. Will update test script and report here shortly.

All unit test passed, documented in https://jira.xilinx.com/browse/XRT-894.
Waiting for Alan to confirm if there is only one lower page0 as 128 bytes.
dynamic define n sysfs node will come after this PR.

Confluence page for how to use qsfp sysfs:
https://confluence.xilinx.com/display/~davidzha/QSFP+sysfs+layout

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
@@ -414,8 +414,7 @@ struct xmc_pkt {
 };
 
 /* QSFP related macros */
-#define CMC_OP_READ_QSFP_MAXLEN		128 /* In bytes */
-#define CMC_OP_WRITE_QSFP_MAXLEN	13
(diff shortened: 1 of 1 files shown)


maxzhen (approved):
This looks good. Just a few comments:
1. See if we can create sysfs node in a loop so that we can change the number of ports easier later?
2. Please confirm w/ Alan that, for diag node, there will ever be only one lower page and each page is 128 byte (never change).

houlz0507 (approved):

xdavidz:
> This looks good. Just a few comments:
> 
> 1. See if we can create sysfs node in a loop so that we can change the number of ports easier later?
> 2. Please confirm w/ Alan that, for diag node, there will ever be only one lower page and each page is 128 byte (never change).

All unit test passed, documented in https://jira.xilinx.com/browse/XRT-894.
Waiting for Alan to confirm if there is only one lower page0 as 128 bytes.
dynamic define n sysfs node will come after this PR.

PR #5150, opened 2021-04-23
CR-1096410: Adding the unique device ID to the names of the devices in the OpenCL Profile Summary

jvillarre:
This pull request changes the OpenCL profile summary to print out the device name as <shell>-<deviceID>.  This allows us to distinguish between multiple devices with the same shell.

### DIFF
--- a/src/runtime_src/xdp/profile/database/static_info_database.cpp
+++ b/src/runtime_src/xdp/profile/database/static_info_database.cpp
@@ -194,7 +194,7 @@ namespace xdp {
     std::vector<std::string> deviceNames ;
     for (auto device : deviceInfo)
     {
-      deviceNames.push_back((device.second)->deviceName) ;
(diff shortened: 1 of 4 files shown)


pgschuey (approved):
I approve as long as these changes work well with perf_analyze.

src/runtime_src/xdp/profile/database/static_info_database.cpp
```diff
@@ -194,7 +194,7 @@ namespace xdp {
     std::vector<std::string> deviceNames ;
     for (auto device : deviceInfo)
     {
-      deviceNames.push_back((device.second)->deviceName) ;
+      deviceNames.push_back((device.second)->getUniqueDeviceName()) ;
```
keryell, line 197: Since you are here...
```suggestion
      deviceNames.push_back(device.second->getUniqueDeviceName()) ;
```

PR #5158, opened 2021-04-26
Enable pybind11 based sanity test which runs at build time and resolve a bug in xclLogMsg binding

sonals:


src/CMake/nativeTests.cmake
```diff
@@ -12,11 +12,18 @@ add_test(NAME xbutil
   WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
 
 add_test(NAME xbutil2
-  COMMAND ${CMAKE_BINARY_DIR}/runtime_src/core/tools/xbutil2/xbutil2 --new scan
+  COMMAND ${CMAKE_BINARY_DIR}/runtime_src/core/tools/xbutil2/xbutil2 --new examine
```
rozumx, line 15: We should consider removing the nativeTests.cmake.

Instead, the individual unit tests would reside reside in the application's directory.   Look at xclbinutil's unit tests for an example, where:
1. The unit tests are in the "xclbinutil/unittests" directory.
2. They are added to the CMake testing harness via the applications CMakeLists.txt file.  The cmake function to add them is xrt_add_test().

src/python/pybind11/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ if (PythonLibs_FOUND)
     SET(PKGDIR "site-packages")
   endif(${LINUX_FLAVOR} MATCHES "^(Ubuntu|Debian)")
 
-  find_package(pybind11 2.6.0 PATHS "/usr/local/lib/python${PYTHONLIBS_VERSION_MAJOR}.${PYTHONLIBS_VERSION_MINOR}/${PKGDIR}/pybind11")
+  find_package(pybind11 2.6.0 REQUIRED PATHS "/usr/local/lib/python${PYTHONLIBS_VERSION_MAJOR}.${PYTHONLIBS_VERSION_MINOR}/${PKGDIR}/pybind11")
```
rozumx, line 17: Hmm...Are PATHS still required?

rozumx (approved):
Approved with a few comments to consider.

src/python/pybind11/CMakeLists.txt
sonals: **REQUIRED** essentially will fail XRT build if the minimum requested version is not found. pybiind11 is required (not optional) since our 22_verify.py test distributed with `xbutil validate` relies of pybind11 binding of XRT. Note the pybind11 package requirement is XRT compile time requirement and not a runtime requirement.

src/CMake/nativeTests.cmake
sonals: Certainly we should move to that model in a future revision.

src/python/pybind11/CMakeLists.txt
stsoe: What's the deal with 'local' as in '/usr/local'?    Is that always where pybind11 will be installed?
Is pip3 the proper way to install pybind11?  And will that always install under /usr/local ?
I think 18.04 and later support apt install python3-pybind11 ?

stsoe (commented):
Looks good.  Just one comment about /usr/local.

src/python/pybind11/CMakeLists.txt
sonals: pip3 is installing pybind11 to /usr/local. On all platforms we are getting >= version 2.6.0
apt install is providing an older version pybind11 (< version 2.6.0) which is not usable because of bugs related to stale format string pointer in memoryview from buffer constructor.

src/CMake/nativeTests.cmake
AShivangi, line 15: we don't need to specify '--new' with xbutil**2**

src/CMake/nativeTests.cmake
```diff
@@ -12,11 +12,18 @@ add_test(NAME xbutil
...
   WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
 
 add_test(NAME xbmgmt2
-  COMMAND ${CMAKE_BINARY_DIR}/runtime_src/core/tools/xbmgmt2/xbmgmt2 --new scan
+  COMMAND ${CMAKE_BINARY_DIR}/runtime_src/core/tools/xbmgmt2/xbmgmt2 --new --help
```
AShivangi, line 19: we don't need to specify '--new' with xbmgmt**2**

AShivangi (approved):

src/CMake/nativeTests.cmake
sonals: Good catch! Updated.

src/CMake/nativeTests.cmake
sonals: Good catch! Updated.

dayeh-xilinx:
retest this please. Pipeline problem with Docker image downloading should be fixed now.

PR #5160, opened 2021-04-26
xbutil/xbmgmt: Fixed previous line escape sequence

rozumx:
Unfortunately, not all terminals support the previous line escape sequence.  Updated the code base to instead perform the following action: "carriage return" followed by a "cursor up"

This PR fixes:
  CR-1096046 - xbutil2 validate and xbmgmt2 dump to a file print multiple status lines in console


AShivangi (approved):
Thanks for making these changes! Tested and approved

PR #5163, opened 2021-04-27
CR-1058379 Need ability to compare XRT version at compile-time

stsoe:
Add XRT_VERSION_CODE and macro XRT_VERSION(major, minor) to version.h.
Allows code to compile-time or run-time compare per CR request.

```
if (XRT_VERSION_CODE > XRT_VERSION(2, 5))
  ...;
else
  ...;
```

```
#if XRT_VERSION_CODE > XRT_VERSION(2, 5)
  ...;
#else
  ...;
#endif
```

Patch version (build number) comparison is not supported.


rozumx (approved):

As always, simple and elegant.

PR #5166, opened 2021-04-28
CR-1083145 XRT API to report interface UUID for 2RP shell

stsoe:
xrt::device param to retrieve interface uuid from device

```
xrt::device device = ...;
auto uuid = device.get_info<xrt::info::device::interface_uuid>();
std::cout << "interface uuid: " << uuid.to_string() << "\n";
```

See also `XRT/tests/xrt/query/main.cpp`


uday610 (approved):
Nice !! Thanks

PR #5167, opened 2021-04-28
CR-1096028: fix arguments count value for kernels with no arguments for AP_CTRL_HS/CHAIN

Himanshu-xilinx:
* fix arguments count value for kernels with no arguments for AP_CTRL_HS/CHAIN
* AP_CTRL_HS/CHAIN PL kernel have first 4 control registers so if kernel has no arguments or all streaming arguments it will "start_krnl_ecmd2xcmd" will produce wrong "isize" value and lead to crash.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -1169,6 +1169,7 @@ class kernel_impl
   {
     if (protocol == FAST_ADAPTER)
       amend_fa_args();
+    // For AP_CTRL_HS/CHAIN, first 4 are control register
     if (protocol == AP_CTRL_HS || protocol == AP_CTRL_CHAIN)
       regmap_size = std::max(regmap_size, 4LU);
   }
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx (approved):

chvamshi-xilinx:
https://github.com/Xilinx/XRT/blob/master/src/runtime_src/core/common/drv/kds_core.c#L1138

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1169,6 +1169,10 @@ class kernel_impl
   {
     if (protocol == FAST_ADAPTER)
       amend_fa_args();
+    // For AP_CTRL_HS/CHAIN, first 4 are control register
+    size_t regmap_size_default = 4;
+    if (protocol == AP_CTRL_HS || protocol == AP_CTRL_CHAIN)
+      regmap_size = std::max(regmap_size, regmap_size_default);
```
stsoe, line 1175: No need for variable. 

`regmap_size = std::max<size_t>(regmap_size, 4); `

stsoe:
I am merging this and will make adjustments later at next edit.

stsoe (approved):

PR #5172, opened 2021-04-28
CR-1097233: Fixing seg fault in continuous trace offload

jvillarre:
When translating PL hardware device events into host-aligned events, we were keeping track of the CU start events so we could use them in approximating end events if any hardware packets were dropped.  In continuous offload, however, the underlying event was dumped to disk and cleaned from memory, leaving a stray pointer to a deleted object.

This pull request changes the values we store in the device event creator to the event ID and device timestamp instead of the event object that wasn't owned by this class.


IshitaGhosh (approved):
Changes look good

PR #5173, opened 2021-04-28
CR-1095772: Fix seg fault on profiling in software emulation when multiple devices are present

jvillarre:
In software emulation, we collect some information to fill out our profile summary tables from the xocl structures, including parsing the memory topology axlf section.  When multiple software emulation devices exist, we were previously checking each one, which erroneously caused us to fetch a section on an uninitialized device.

This pull request adds a check to make sure the device is active before fetching the mem topology section to avoid a seg fault.


pgschuey (approved):
Looks good

PR #5176, opened 2021-04-29
Fix edge multi-processes regression

larry9523:
This is a regression that cu sub devices are destroyed/created even we have same xclbin already being loaded. This will lead to bad behavior in multi-processes where when process is actively using the cu sub devices and another process destroy/reset them.

The fix is resume only the behavior that we only destroy/create cu sub devices when we swap xclbin.


mamin506 (approved):
Thanks Larry.

saifuddin-xilinx (commented):
Changes are looks good for me.

saifuddin-xilinx (approved):

PR #5179, opened 2021-04-29
Print circular buffer warning only at the end and use user markers to indicate trace errors

hackwa:


dayeh-xilinx:
retest this please. XRT pipeline is fixed and ready to go.

jvillarre (approved):
Change looks good.

PR #5181, opened 2021-04-30
CR-1081043 : hbm grouping reporting is not showing terminus of hbm range

saifuddin-xilinx:
Modified group format for HBM, so that it will fit in 16bit buffer
-- Old Format was MBG[0,1,2,3]
-- New format is MBG[0:3]
Also for non-contagious banks i.e. MBG[4,2,3,0,6,8,9] becomes MBG[0,2:4,6,8:9] or MBG MBG[0:0,2:4,6:6,8:9]

### DIFF
--- a/src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx
+++ b/src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx
@@ -772,10 +772,11 @@ createMemoryBankGroupEntries( std::vector<WorkingConnection> & workingConnection
 
       // Add a tag value to indicate that this entry was the result of grouping memories
       std::string newTag = "MBG[";
-      for (unsigned int memIndex = startIndex; memIndex <= endIndex; ++memIndex) {
-        newTag += std::to_string(workingConnections[memIndex].memIndex);
-        newTag += (memIndex != endIndex) ? "," : "]";
-      }
+      newTag += std::to_string(workingConnections[startIndex].memIndex);
+      if (startIndex != endIndex)
+        newTag += ":" + std::to_string(workingConnections[endIndex].memIndex);
+
+      newTag += "]";
 
       // Record the new tag, honoring the size limitation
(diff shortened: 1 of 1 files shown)


src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx
```diff
@@ -772,10 +772,11 @@ createMemoryBankGroupEntries( std::vector<WorkingConnection> & workingConnection
 
       // Add a tag value to indicate that this entry was the result of grouping memories
       std::string newTag = "MBG[";
-      for (unsigned int memIndex = startIndex; memIndex <= endIndex; ++memIndex) {
-        newTag += std::to_string(workingConnections[memIndex].memIndex);
-        newTag += (memIndex != endIndex) ? "," : "]";
-      }
+      newTag += std::to_string(workingConnections[startIndex].memIndex);
+      if (startIndex != endIndex)
```
rozumx, line 776: The assumption with this change relies on the fact that the working connections on in numeric order and are continuous.   While this is the case here, it is not something that can be relied on.  Instead it would be better to: 
1. Collect all of the memIndex values.
2. Sort them in numeric order.
3. Iterate over the collection producing a more compress tag.

Examples:
- MBG[5,4,6]  becomes MBG[4:6]
- MBG[0,1,2,3,4,5,6] becomes MBG[0:6]
- MBG[0,3,2,6,5,8,7] becomes MBG[0:3,5:8]
- MBG[4,2,3,0,6,8,9] becomes MBG[0,2:4,6,8:9] or MBG MBG[0:0,2:4,6:6,8:9]

rozumx (changes requested):
Please update this fix to address all possible use-cases.

src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx
saifuddin-xilinx: Changed the log as per the review comments. Please review once again.

rozumx (approved):
Thank you make the changes.

It is much cleaner and easy to follow.

PR #5182, opened 2021-04-30
Fix clEnqueueReadBufferRect

Ralender:
in some cases the map function return an pointer to a posix_memalign instead of a shared mmap. this caused clEnqueueReadBufferRect to write garbage data to the host buffer.

I found this issues on our SYCL implementation.

### DIFF
--- a/src/runtime_src/xocl/api/clEnqueueReadBufferRect.cpp
+++ b/src/runtime_src/xocl/api/clEnqueueReadBufferRect.cpp
@@ -145,30 +145,31 @@ clEnqueueReadBufferRect(cl_command_queue     command_queue ,
     xocl::xocl(*event)->queue(true /*wait*/);
   }
 
-  // Now the event is running, this should be hard_event and handle asynchronously
   auto device = xocl::xocl(command_queue)->get_device();
-  auto xdevice = device->get_xdevice();
   auto boh = xocl::xocl(buffer)->get_buffer_object_or_error(device);
-  void* host_ptr = xdevice->map(boh);
-  
-  size_t yit,zit;
-  for(zit=0;zit<region[2];zit++){
-    for(yit=0;yit<region[1];yit++){
-      size_t buffer_row_origin_in_bytes =
-        buffer_origin_in_bytes+
-        zit*buffer_slice_pitch+
-        yit*buffer_row_pitch;
+  cl_int errc = 0;
+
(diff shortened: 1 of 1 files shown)


src/runtime_src/xocl/api/clEnqueueReadBufferRect.cpp
```diff
@@ -145,30 +145,31 @@ clEnqueueReadBufferRect(cl_command_queue     command_queue ,
...
-  for(zit=0;zit<region[2];zit++){
-    for(yit=0;yit<region[1];yit++){
-      size_t buffer_row_origin_in_bytes =
-        buffer_origin_in_bytes+
-        zit*buffer_slice_pitch+
-        yit*buffer_row_pitch;
+  cl_int errc = 0;
+
+  // This is memory map is bigger than we could make it when start offsets is not 0, 0, 0.
+  void *host_ptr = clEnqueueMapBuffer(command_queue, buffer, true, CL_MAP_READ,
```
keryell, line 153: Just curious: is it possible to mix and match high-level call to OpenCL API inside lower-level `xocl` code?

stsoe: @keryell What do you mean ?  In this case there is no mixing. If you are asking about xocl::xocl, then it's just a static_cast from opencl domain to implementation domain.  `xocl `originally stood for xilinx opencl, before someone hijacked it for the userpf driver.

stsoe: Btw.  I have no clue about the validity of the changes in this pull request.  I trust @Ralender has the understanding, I don't think anyone is using this API and I am sure it is not tested in any of our test suites.  The code is leftover from the original sdaccel days, so thank you for fixing it!

stsoe:
ok to test

keryell:
I am fine with this being merged, since it solves one of our SYCL issue. :-) The Intel https://github.com/intel/llvm implementation we are based on loves a lot these rectangle-related API...
My concern was just a design question about calling `clEnqueueMapBuffer` inside another OpenCL call and thus going again in the argument checking, compared to call a more lower-level API to shortcut this. But @Ralender suggested during our daily meeting it would require to factorize somehow the current XRT OpenCL implementation to avoid code duplication...
Thus, definitely a broader scope than this fix, so let's keep this as is for now.
Thanks.

PR #5186, opened 2021-04-30
AIE runtime trace and profile fixes

pgschuey:
* AIE trace config file is only written when proper resources are obtained
* Corrected first timer value in AIE profiling
* Various bug fixes and improved messaging in AIE trace and profile plugins
* Added optional trace flush before offloading from DDR
* Missing settings added to profile summary guidance params (CR-1097019, CR-1096222, CR-1077439)


src/runtime_src/xdp/profile/plugin/aie_trace/aie_trace_plugin.cpp
```diff
@@ -680,6 +710,68 @@ namespace xdp {
...
+  void AieTracePlugin::setFlushMetrics(uint64_t deviceId, void* handle)
+  {
+    auto drv = ZYNQ::shim::handleCheck(handle);
+    if (!drv)
+      return;
+    auto aieArray = drv->getAieArray();
+    if (!aieArray)
+      return;
+    auto Aie = aieArray->getDevInst();
+    auto aieDevice = std::make_shared<xaiefal::XAieDev>(Aie, false);
```
stsoe, line 722: Random comment. Don't use shared_ptr unless ownership is being shared.  Is this a case where std::make_unique should have been used?

pgschuey: Thanks for the feedback, Soren. The pointer actually is shared with our database.

stsoe: Hmm, I don't see it.  I'm curious where the sharing occurs?

hackwa: Hey @pgschuey  creating aieFAL class again might cause issues. We should store the shared pointer we created when we initialized trace and store it in database and reuse it.

hackwa: Another thought. When counters and trace both are used, we ideally should use same pointer in both aie trace and profile plugins.

pgschuey: Submitted fix today that improves on this issue. Instances of XAieDevInst and XAie_Device are included in the static database and used for AIE trace and/or profiling. Makes for simpler code, too.

jvillarre (approved):
Changes look good for now.  The new ini flag is for current development and testing and should not be documented and should be removed soon.

hackwa (approved):
changes look good.

PR #5188, opened 2021-05-03
CR-1078490 Double buffer created in no-dma platform when not using host_only flag

stsoe:
Adjust buffer flags to infer host_only when buffer is allocated in
host memory bank.  This change impacts NoDMA platforms only, where it
prevents unncessary double BO creation.

This PR reverts support for userptr for NoDMA platforms that was added
in #5187.  Also error out in general when application attempts to
construct host_only BOs with a userptr, which just like NoDMA case
would imply an extra memcpy managed in userspace XRT. XRT should not
hide extra memcpy in general, if host code really wants this, then do
it in the host code itself.

Finally, this PR adds MEM_HOST to the MEM_TYPE enumeration in
xclbin.h.  While the xclbin itself does not use this enumerator, it is
now used internally in XRT implementation when returning the memory
type of a bank based on the mem_data::m_tag comparison. Once supported
fully by upstream tools and platforms, XRT can avoid the string
comparison and rely on the enumerator value in mem_data::m_type.

### DIFF
--- a/src/runtime_src/core/common/device.cpp
+++ b/src/runtime_src/core/common/device.cpp
@@ -31,6 +31,7 @@
 #include <string>
 #include <iostream>
 #include <fstream>
+#include <functional>
 
 #ifdef _WIN32
 #pragma warning ( disable : 4996 )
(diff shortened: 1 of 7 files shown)


src/runtime_src/core/common/device.cpp
```diff
@@ -209,6 +210,23 @@ get_memidx_encoding(const uuid& xclbin_id) const
   return m_memidx_encoding;
 }
 
+device::memory_type
+device::
+get_memory_type(size_t memidx) const
+{
+  auto mem_topology = get_axlf_section<const ::mem_topology*>(ASK_GROUP_TOPOLOGY);
+  if (!mem_topology || mem_topology->m_count < static_cast<int32_t>(memidx))
```
rozumx, line 218: Minor item: Because of the static_cast and pointers, I missed the fact that we are basically doing `m_count < memidx` This oversight was due to obfuscation and I later went back to this line after reviewing line 221.

This code needs to be here, but i don't know what we can do to assist the next developer in seeing the intent. 

Again, just a minor issue I encountered while reviewing the code.

rozumx (approved):
Overall this look really good.

It was really enjoyable to review :)

PR #5189, opened 2021-05-03
xbutil/xbmgmt: Various bug fixes and EoU enhancments

rozumx:
Work Done
---------
+ Improved the try/catch functionality:
  - Added helper methods to print out a catch message (both runtime_error and system_error)
  - Added support in the entry level portion of the code to not create a debug stack if an std::errc::operation_canceled exception is caught.  Doing so allows for the code to exit the codebase gracefully.
+ Enhanced all command and sub-command help menus to also report the global options.
+ Fixed various spelling errors.
+ Removed format support for human readable 'text' output to files.
+ Enhanced all tools to always produce human readable 'text' output to the console, regardless if the "--output" is used.
+ Cleaned up various error messages
+ Removed dead code
+ Updated copyright notices
+ Renamed XBU::is_esc_enabled() to XBU::is_escape_codes_disabled() for the logic for this function is negative.
+ Replaced the wrap_paragraph() and wrap_paragraphs() function with a brand new wrap_paragraph() function that address bugs related to formating.  In addition, the new function is easier to read.
+ Exposed ProgressBar::formatTime() as a static method.
+ Updated Process to clear the progress bar.
+ Updated Process to throw an exception if the test times out.
+ Updated XBU::can_proceed() with a force option.
+ Command: Validate
  - Cleanup the running output.  The "running" status completion bar is removed after a test completes.
  - Cleaned up formated output to be consistant (e.g., The placement of the colon ':' character)
  - Updated the warning and error printed keys to be Warning(s) and Error(s)
  - Changed the default format type to "json"
  - Refactored how the JSON output file is openned and written to.  Output file write confirmation is echoed to the display.
+ Command: xbutil Examine
  - Changed the default format type to "json"
  - Updated code flow to throw an operation canceled exception for clean exits with an error
  - Updated the DRC checks to validate the format and output file existance cleanly
  - Refactored how the JSON output file is openned and written to.   Output file write confirmation is echoed to the display.
  - Added force support
+ Command: xbmgmt Examine
  - Added force support
  - Refactored how the JSON output file is openned and written to.   Output file write confirmation is echoed to the display.
+ Command: xbutil/xbmgmt Advanced
  - Added support for sub-options to print out the global help data
+ Class XBHelpMenus
  - Added support to report global options
  - Fixed various bugs related to wrapping of paragraphs
  - Updated console output to be write to a console stream instead of directly to std::cout.
  - Updated methods to use a output stream for formatted output.
+ Reports
  - Updated all the human readable reports to be produced from an given property tree, instead of an internally created one.  This tree is obtain and given to the individual reports via the Report class.
  - Cleaned up the formatting of the reports to be more colon ':' column align
  - For writeReport() addressed some missing default values
  - For writeReport() updated const for many of the variables
+ Fixed issue where an exception is thrown when checking the firewall status of a golden card.
+ Addressed various root privileges notificaiton issues.
+ Command reset now reports available devices if one isn't given.

CRs
-----
CR-1096635 - xbutil validate should still report progress to the console with use the --output option
CR-1096027 - verbose does not show in xbutil validate -h
CR-1097111 - /opt/xilinx/xrt/bin/xbutil validate output the json file has non JSON format information
CR-1089728 - Feature request: Force reset device using xbutil
CR-1090669 - xbmgmt2 more information than needed for no sudo

### DIFF
--- a/src/runtime_src/core/tools/common/ProgressBar.cpp
+++ b/src/runtime_src/core/tools/common/ProgressBar.cpp
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2020 Xilinx, Inc
+ * Copyright (C) 2020-2021 Xilinx, Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License"). You may
  * not use this file except in compliance with the License. A copy of the
@@ -55,8 +55,8 @@ static boost::format fmtBatchPF("\n[%s]: %s < %s >\n");
 
 // ------ S T A T I C   F U N C T I O N S -------------------------------------
 
-static std::string
-format_time(std::chrono::duration<double> duration) 
+std::string
+ProgressBar::formatTime(std::chrono::duration<double> duration) 
 {
   auto hours = std::chrono::duration_cast<std::chrono::hours>(duration);
   auto minutes = std::chrono::duration_cast<std::chrono::minutes>(duration);
@@ -88,7 +88,7 @@ ProgressBar::ProgressBar(const std::string &_opNname, unsigned int _maxNumIterat
     , m_lastUpdated(std::chrono::high_resolution_clock::now()) 
 {
   if (!m_isBatch) 
-    m_ostr << fmtUpdate % "" % /*Percent*/ 0 % m_opName % format_time(m_elapsedTime) << std::endl;
+    m_ostr << fmtUpdate % "" % /*Percent*/ 0 % m_opName % formatTime(m_elapsedTime) << std::endl;
   else 
     m_ostr << m_opName << ": ";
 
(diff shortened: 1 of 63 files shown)


src/runtime_src/core/tools/common/ProgressBar.cpp
```diff
@@ -55,8 +55,8 @@ static boost::format fmtBatchPF("\n[%s]: %s < %s >\n");
 
 // ------ S T A T I C   F U N C T I O N S -------------------------------------
 
-static std::string
-format_time(std::chrono::duration<double> duration) 
+std::string
+ProgressBar::formatTime(std::chrono::duration<double> duration)
```
AShivangi, line 59: since this is no longer a static fucntion, should it go under a "member function" comment?

rozumx: It is still a static function for the header file was updated as well to be a static value.

Ultimately, it needs to go into the XBUtility, but didn't go to that in this release (it was getting fairly large).  In my next PR, it will be moved.

AShivangi (approved):
These look good! Thank you for making these changes

PR #5196, opened 2021-05-03
CR 1096198, 1090494: Add deprecation message for xrt_profile; Add TRACE_BUFFER_FULL guidance rule

IshitaGhosh:
CR 1096198 : Bring back "xrt_profile" configuration in 2021.1 release and add deprecation message for next release
CR 1090494 : Add TRACE_BUFFER_FULL guidance rule for both TS2MM and FIFO. Set trace buffer full status in device offload plugin and also in hal and opencl plugin


jvillarre (approved):
Changes look good.

PR #5197, opened 2021-05-04
xclbinutil : Added DRC check to limit softkernel instances to 128.

rozumx:
Work Done
------------
+ Updated the DRC check to validated that the number of softkernel instances count doesn't exceed 128.
+ Added regression tests to validate DRC check

CRs
----
CR-1090185 - xclbinutil verify ps_kernel symbol name

### DIFF
--- a/src/runtime_src/tools/xclbinutil/SectionSoftKernel.cxx
+++ b/src/runtime_src/tools/xclbinutil/SectionSoftKernel.cxx
@@ -230,7 +230,7 @@ SectionSoftKernel::copyBufferUpdateMetadata(const char* _pOrigDataSection,
     XUtil::TRACE(XUtil::format("  mpo_symbol_name (0x%lx): '%s'", softKernelHdr.mpo_symbol_name, sValue.c_str()).c_str());
 
     // DRC Check checking to see if the maximum symbol name length has been violated
-    static const unsigned int MAX_SYMBOL_NAME_LENGTH = 19;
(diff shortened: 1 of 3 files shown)


AShivangi (approved):

src/runtime_src/tools/xclbinutil/SectionSoftKernel.cxx
```diff
@@ -244,6 +244,14 @@ SectionSoftKernel::copyBufferUpdateMetadata(const char* _pOrigDataSection,
     uint32_t value = ptSK.get<uint32_t>("m_num_instances", defaultValue);
     softKernelHdr.m_num_instances = value;
     XUtil::TRACE(XUtil::format("  m_num_instances: %d", softKernelHdr.m_num_instances).c_str());
+
+    // DRC Check checking the maximum number of instances
+    static const unsigned int MAX_NUM_INSTANCES = 128;
```
stsoe, line 249: `constexpr unsigned int max_num_instances = 128`; would be better.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconst-constexpr

stsoe (commented):
One comment

src/runtime_src/tools/xclbinutil/SectionSoftKernel.cxx
rozumx: Just learned something new today.

PR #5204, opened 2021-05-04
Add deprecation for the QDMA streaming APIs

sonals:


houlz0507 (approved):

haeseunglee (approved):
Approved the pull request

uday610 (approved):

stsoe (changes requested):
@sonal.  The point of deprecation is to not only move to deprecated/xrt.h as is done, but also remove all comments.   Let's make it as hard as possible to use these deprecated APIs.

sonals:
Removed the documentation of deprecated streaming APIs

sonals:
> 
> 
> @sonal. The point of deprecation is to not only move to deprecated/xrt.h as is done, but also remove all comments. Let's make it as hard as possible to use these deprecated APIs.

Updated PR

syed-ahmed:
Thanks for this! In our [course](https://www.seas.upenn.edu/~ese532/), we struggled on understanding why the stream APIs used qdma and what was meant by h2k not available in the embedded platforms in the vitas docs. Could you elaborate a little on the slave bridge comment? Streaming from host to device is now just done through enqueMigrateMemObjects?

PR #5205, opened 2021-05-05
More clang-tidy fixes and disables

stsoe:
Disabled:
// used commonly with __func__, etc
+-hicpp-no-array-decay,

// we do lots of bit manip
+-hicpp-signed-bitwise,

// not sure about these ones
+-modernize-use-trailing-return-type,
+-cppcoreguidelines-pro-bounds-pointer-arithmetic,

// ert_packet is C-style union and used commonly
+-cppcoreguidelines-pro-type-union-access,

// used commonly with __func__, etc
+-cppcoreguidelines-pro-bounds-array-to-pointer-decay,

// we do lots of reinterpreting (e.g. xclbin structs)
+-cppcoreguidelines-pro-type-reinterpret-cast,

// data indexing in ert command objects is too common
+-cppcoreguidelines-pro-bounds-constant-array-index

Fixed xrt_ip.cpp, xrt_bo.cpp, xrt_kernel.cpp to build cleanly.
Very tiduous, no errors detected, but good warnings in general.

### DIFF
--- a/src/.clang-tidy
+++ b/src/.clang-tidy
@@ -4,7 +4,26 @@
 # clang-tidy "-checks=cert-*,bugprone-*,clang-analyzer-*,concurrency-*,cppcoreguidelines-*,hicpp-*,modernize-*" --dump-config
 # clang-tidy looks for .clang-tidy file walking up the source directory structure to identify which checks to run
 # One can customize checks in specific source directories by placing custom .clang-tidy in those source directories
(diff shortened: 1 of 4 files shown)


stsoe:
@sonals Let's make sure we document when/if we disable lint checks.

sonals (approved):
Wonderful! Once we enable daily build of clang-tidy we can track any regressions.

PR #5207, opened 2021-05-05
CR-1096474 Added the Clear ERROR when users provide the DEBUG_MODE op…

venkatp-xilinx:
CR-1096474 Added the Clear ERROR when users provide the DEBUG_MODE option as GDB and halting the design to run.

### DIFF
--- a/src/runtime_src/core/pcie/emulation/common_em/config.h
+++ b/src/runtime_src/core/pcie/emulation/common_em/config.h
@@ -143,7 +143,8 @@ namespace xclemulation{
   enum DEBUG_MODE {
     OFF,
     BATCH,
-    GUI
+    GUI,
+    GDB
   };
   
   enum ERTMODE {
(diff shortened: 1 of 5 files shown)


src/runtime_src/core/pcie/emulation/common_em/config.h
```diff
@@ -143,7 +143,8 @@ namespace xclemulation{
   enum DEBUG_MODE {
```
stsoe, line 143: These enums (DEBUG_MODE and ERTMODE) are very dangerous.    You should consider changing them to enum classes with lower case enumerators that are scoped and don't run the risk of conflicting with #defines.  
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-class
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-caps

venkatp-xilinx: Hi Soren, can we handle this cleanup for next time. We are in SRCB, did not want to do these changes at this point of time. I will clean up it in HEAD once XRT 2021.1 branches out.

stsoe (approved):
Thanks for changing the enums!

PR #5222, opened 2021-05-07
CR-1096826 remap bar doesn't exist in u2 platform

rajkumar-xilinx:
Since remap bar doesn't exist in u2 platform, do not return any error
code during subdev p2p construct() phase of xbutil reset operation.

With this, validate test is passed after xbutil reset operation as well.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/p2p.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/p2p.c
@@ -1060,7 +1060,10 @@ static int p2p_remap_resource(struct platform_device *pdev, int bar_idx,
 
 	if (!p2p->remapper) {
 		p2p_err(p2p, "remap does not exist");
-		return -EINVAL;
+		/* remap bar doesn't exist on u2 platforms and remap_range set to entire
+		 * p2p bar len in p2p_mem_init(). So, check remap_range val before returning err code.
+		 */
+		return (p2p->remap_range ? 0 : -EINVAL);
 	}
 
 	mutex_lock(&p2p->p2p_lock);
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/p2p.c
```diff
@@ -1060,7 +1060,10 @@ static int p2p_remap_resource(struct platform_device *pdev, int bar_idx,
 
 	if (!p2p->remapper) {
 		p2p_err(p2p, "remap does not exist");
-		return -EINVAL;
+		/* remap bar doesn't exist on u2 platforms and remap_range set to entire
+		 * p2p bar len in p2p_mem_init(). So, check remap_range val before returning err code.
+		 */
+		return (p2p->remap_range ? 0 : -EINVAL);
```
houlz0507, line 1066: I think this brings a real case that ERT is implemented on P2P bar and P2P bar does not support remapper. 
So I would suggest to change 1062 to p2p_info(), and return -ENODEV unconditionally.

rajkumar-xilinx: If we just return -ENODEV unconditionally then in __xocl_subdev_construct() function at https://github.com/Xilinx/XRT/blob/master/src/runtime_src/core/pcie/driver/linux/xocl/xocl_subdev.c#L420, driver is simply goto error as remap_resource returned -ENODEV and it skips reading resources of the p2p subdev.
This sequence is called during "xbutil reset", and when xbutil validate called, it is failing on u2.

So, to fix this issue, remap_resouce shall return 0 on u2 platforms, it is like skipping this function on u2 altogether.

Please suggest any other way to fix this problem on u2 case.

houlz0507 (approved):

PR #5224, opened 2021-05-07
trace fixes

hackwa:
1. Print circular buffer warning as soon as we see overrun (CR-1097504)
2. Update overrun warning for new flag (CR-1098578)
3. Use locks to access trace offload status (CR-1098581)


jvillarre (approved):
Additions look good.

PR #5226, opened 2021-05-08
Bug fix:  xbmgmt program could timeout early when programming the u30 SC.

rozumx:
Issue
------
The u30 SC flash programming is done via a python script which in turn calls another helper function.  Depending on the load on the file server could cause delays in the overall programming time.   In some cases, this programming time exceeds the default 60 second time for all operations.

Resolution
-----------
The code was updated to support different timeouts depending on the operation.  For the SC flash, it was increased to 120 seconds.

CR(s)
----
CR-1098830 - [Ubuntu 18.04.4] xbmgmt2 --base --shell xilinx_u30_gen3x4_base_1 --device all shows failure when SC firmware from 6.3.5 to 6.3.7

### DIFF
--- a/src/runtime_src/core/tools/common/Process.cpp
+++ b/src/runtime_src/core/tools/common/Process.cpp
@@ -56,8 +56,6 @@
 #include <iostream>
 #include <thread>
 // ------ S T A T I C   V A R I A B L E S -------------------------------------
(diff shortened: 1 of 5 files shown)


rozumx:
Wrong pull request to make.

PR #5228, opened 2021-05-08
Bug Fix: Addressed flash timeout and SC firmware hang issues

rozumx:
Issues
------
The u30 SC flash programming is done via a python script which in turn calls another helper function. Depending on the load on the file server could cause delays in the overall programming time. In some cases, this programming time exceeds the default 60 second time for all operations.  Note: Currently, it takes 59 seconds to program the flash device.

In addition, the python flashing script could hang on some systems do multiple open device handles.   

Resolutions
-----------
- The code was updated to support different timeouts depending on the operation. For the SC flash, it was increased to 120 seconds
- After the device information has been gathered, all of the device handles are release prior to programming the flash images.

CR(s)
-----
CR-1098821 - [Ubuntu 18.04.4] xbmgmt2 --base --shell xilinx_u30_gen3x4_base_1 --device all shows failure when SC firmware from 6.3.5 to 6.3.7
CR-1098830 - [Ubuntu 18.04.4] xbmgmt2 --base --shell xilinx_u30_gen3x4_base_1 --device all shows failure when SC firmware from 6.3.5 to 6.3.7

### DIFF
--- a/src/runtime_src/core/tools/common/Process.cpp
+++ b/src/runtime_src/core/tools/common/Process.cpp
@@ -56,8 +56,6 @@
 #include <iostream>
 #include <thread>
 // ------ S T A T I C   V A R I A B L E S -------------------------------------
-// bandwidth testcase takes up-to a min to run
-static const int max_test_duration = 60;
 
 // ------ F U N C T I O N S ---------------------------------------------------
 #ifdef NO_BOOST_PROCESS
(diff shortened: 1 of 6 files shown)


AShivangi (approved):
Thanks for making the changes. This looks good :)

xuhz (approved):

PR #5231, opened 2021-05-10
CR 1086988 - AIE Trace data mover IP is not writing raw events data received from AIE to DDR Memory

pgschuey:
* made adjustments to trace flush and turned it on by default
* added some debug messaging

### DIFF
--- a/src/runtime_src/xdp/profile/plugin/aie_trace/aie_trace_plugin.cpp
+++ b/src/runtime_src/xdp/profile/plugin/aie_trace/aie_trace_plugin.cpp
@@ -722,7 +722,7 @@ namespace xdp {
     // New start & end events for trace control and counters
     //coreTraceStartEvent = XAIE_EVENT_TIMER_SYNC_CORE;
     coreTraceStartEvent = XAIE_EVENT_TRUE_CORE;
-	  coreTraceEndEvent   = XAIE_EVENT_TIMER_VALUE_REACHED_CORE;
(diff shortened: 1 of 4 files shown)


src/runtime_src/xdp/profile/plugin/aie_trace/aie_trace_plugin.cpp
```diff
@@ -720,8 +720,9 @@ namespace xdp {
     }
 
     // New start & end events for trace control and counters
-    coreTraceStartEvent = XAIE_EVENT_TIMER_SYNC_CORE;
-	  coreTraceEndEvent   = XAIE_EVENT_TIMER_VALUE_REACHED_CORE;
+    //coreTraceStartEvent = XAIE_EVENT_TIMER_SYNC_CORE;
+    coreTraceStartEvent = XAIE_EVENT_TRUE_CORE;
```
hackwa, line 724: The reason why timer sync doesn't work is because we call reset *before* starting trace in fal. Since we use timer reset event as trace start, we should call reset *after* trace start. We should do it regardless of whether we use XAIE_EVENT_TIMER_SYNC_CORE or not.
Also need to verify the effect of using XAIE_EVENT_TRUE_CORE with Abraham

pgschuey: I re-arranged timer reset to be after core start

IshitaGhosh (approved):
Changes look good.

PR #5241, opened 2021-05-12
Reinitialize trace datamover when circular buffer is used

hackwa:
CR-1098948- Datamover reset isn't working properly


jvillarre (approved):
Changes look good.

PR #5243, opened 2021-05-12
CR-1099496: Fixing seg fault in continuous offload

jvillarre:
While processing device trace, we store start events so we can match them with end events that come later.  In the continuous offload use case, these events can be dumped to a file and erased from the database before the end events are detected, leading to a segmentation fault when processing the next batch of device events.

This pull request changes the information stored from the device event itself to a tuple of necessary information consisting of: the type, the event id, the host timestamp, and the device timestamp.


hackwa (approved):
looks good. we  don't store events in multiple places now

IshitaGhosh (approved):
Changes look good

PR #5244, opened 2021-05-13
CR-1083341: removed SDAccel from public docs 

uday610:
removed SDAccel from public docs


uday610:
https://jira.xilinx.com/browse/CR-1083341

sonals:
ok to test

PR #5251, opened 2021-05-14
update trace warning messages and do one final trace read for circular buffer

hackwa:
CR-1099502 - Add warning for large number of files
CR-1099485 - Add warning when we process > 50mb of trace
CR-1098345- Do one extra trace read to completely flush trace buffer
Other changes -
Print some warnings only once
Fix race condition when continuous trace is stopped
Fix spaces


jvillarre (approved):
Additions look good

PR #5252, opened 2021-05-14
use device interface locks and nullptr checks to avoid contention between aie and pl trace

hackwa:


pgschuey (approved):
Changes should fix the seg faults

jvillarre (approved):
Changes look good

PR #5253, opened 2021-05-14
more AIE runtime fixes

pgschuey:
* Catching when compile-time AIE trace metrics are specified
* Set group event masks for AIE profiling
* Default for runtime trace metrics is functions
* Active cycles now uses end event 28


jvillarre (approved):
Change looks good

PR #5261, opened 2021-05-17
Update changelog and add a note about restricted usage of Unmanaged D…

sonals:
…MA APIs


jvillarre (approved):
Profiling additions to changelog look good.

uday610 (approved):

PR #5263, opened 2021-05-17
make aie fal work with counter thresholds

hackwa:
fal function to set counter threshold is a dummy and doesn't do anything. without that, trace will not work on es1 devices and trace flush functionality won't work. For now we directly configure threshold using driver


pgschuey (approved):
changes look great

jvillarre (approved):
Change looks good

PR #5274, opened 2021-05-18
Trap improper use of null xrt::bo in OpenCL interoperability

stsoe:
OpenCL memory objects (cl_mem) are not associated with xrt::bo buffers
until they are 1-1 with a memory bank,  which sometimes is only after
the memory object has been assigned to a kernel argument.

Using xrt::opencl::get_xrt_bo(cl_device, cl_mem) now throws if the
memory object is not yet associated with an xrt::bo object.

### DIFF
--- a/src/runtime_src/xocl/api/xlnx/cl2xrt.cpp
+++ b/src/runtime_src/xocl/api/xlnx/cl2xrt.cpp
@@ -28,16 +28,19 @@ namespace xrt { namespace opencl {
 xrt::device
 get_xrt_device(cl_device_id device)
 {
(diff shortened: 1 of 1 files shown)


keryell:
Is this behavior documented somewhere?
We have been beaten by this in the past...

stsoe:
> Is this behavior documented somewhere?
> We have been bitten by this in the past...

@keryell yes, but you really really have to know what to look for:
https://www.xilinx.com/html_docs/xilinx2020_2/vitis_doc/devhostapp.html#osv1555367508419

PR #5281, opened 2021-05-19
 CR-1093195 => zocl lpddr memory enhancement support

saifuddin-xilinx:
CR Number : https://jira.xilinx.com/browse/CR-1093195
The changes are as follows:
Memory Initialization:
1.	Currently CMA memory is initialized based on tag value substr of “DDR” and base address is “0x0”
       a.	We need to find better alternative to initialize CMA memory from zocl 
2.	Created a link list based on similar memory tag.

Memory Allocation:
1.	If memory is not available from one memory region, XRT will try to allocation from other memory region which is having the similar tag. 
2.	NON-CACHEABLE memory can also be allocated from CMA region if tag value is same. 
3.	CACHEABLE memory will be allocating only from the CMA memory region. 

**Clarifications:**
1. If user asks for Cacheable, We are allocating in non-cacheable area if Cacheable memory is not available. (for backward compatibility)

2. If user asks for Non-Cacheable, We are allocating Non-cacheable only, but We are allocating from CMA area which user can use for Cacheable memory. We added a warning/info saying that allocating non-cacheable memory from CMA.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
@@ -184,7 +184,9 @@ zocl_create_range_mem(struct drm_device *dev, size_t size, struct zocl_mem *mem)
 		if (mem->zm_type == ZOCL_MEM_TYPE_CMA) {
 			struct drm_zocl_bo *cma_bo = zocl_create_cma_mem(dev, size);
 			if (!IS_ERR(cma_bo)) {
-				/* Get the memory in CMA memory region */
+				/* Get the memory in CMA memory region as no memory 
+				 * avilable for non-cachable region.
+				 */
 				mutex_unlock(&zdev->mm_lock);
 				kfree(bo->mm_node);
 				drm_gem_object_release(&bo->gem_base);
@@ -192,6 +194,8 @@ zocl_create_range_mem(struct drm_device *dev, size_t size, struct zocl_mem *mem)
 				cma_bo->flags |= ZOCL_BO_FLAGS_CMA;
 				return cma_bo;
 			}
+			DRM_WARN("Memory allocated from CMA region as"
+				       " non-chanhable memory exhausted\n");
 		}
 		else {
 			err = drm_mm_insert_node_generic(mem->zm_mm,
@@ -386,8 +390,11 @@ zocl_create_bo_ioctl(struct drm_device *dev, void *data, struct drm_file *filp)
 		if (bank < zdev->num_mem && zdev->mem[bank].zm_used) {
 			if (zdev->mem[bank].zm_type == ZOCL_MEM_TYPE_CMA)
 				args->flags |= ZOCL_BO_FLAGS_CMA;
-		} else
+		} else {
+			DRM_WARN("Allocating BO from CMA as bank[%d] not available\n", 
+					bank);
 			args->flags |= ZOCL_BO_FLAGS_CMA;
+		}
 	}
 
 	if (!(args->flags & ZOCL_BO_FLAGS_CACHEABLE)) {
--- a/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
@@ -15,6 +15,7 @@
  */
 
 #include <linux/pagemap.h>
+#include <linux/of_address.h>
 #include <linux/iommu.h>
 #include <asm/io.h>
 #include "xrt_drv.h"
@@ -172,8 +173,8 @@ zocl_create_range_mem(struct drm_device *dev, size_t size, struct zocl_mem *mem)
 	bo->mm_node = kzalloc(sizeof(struct drm_mm_node),
 			GFP_KERNEL);
 	if (IS_ERR(bo->mm_node)) {
-		kfree(bo);
 		drm_gem_object_release(&bo->gem_base);
+		kfree(bo);
 		return ERR_PTR(-ENOMEM);
 	}
 
@@ -186,8 +187,8 @@ zocl_create_range_mem(struct drm_device *dev, size_t size, struct zocl_mem *mem)
 				/* Get the memory in CMA memory region */
 				mutex_unlock(&zdev->mm_lock);
 				kfree(bo->mm_node);
-				kfree(bo);
 				drm_gem_object_release(&bo->gem_base);
+				kfree(bo);
 				cma_bo->flags |= ZOCL_BO_FLAGS_CMA;
 				return cma_bo;
 			}
@@ -213,8 +214,8 @@ zocl_create_range_mem(struct drm_device *dev, size_t size, struct zocl_mem *mem)
 				(long)size);
 		mutex_unlock(&zdev->mm_lock);
 		kfree(bo->mm_node);
-		kfree(bo);
 		drm_gem_object_release(&bo->gem_base);
+		kfree(bo);
 		return ERR_PTR(-ENOMEM);
 	}
 
@@ -981,6 +982,38 @@ void zocl_update_mem_stat(struct drm_zocl_dev *zdev, u64 size, int count,
 	write_unlock(&zdev->attr_rwlock);
 }
 
+/* This function retern True if given region are reserved
+ * on device tree. Else return False
+ */
+static int check_for_reserved_memory(uint64_t start_addr, size_t size)
+{
+	struct device_node *mem_np;
+	struct device_node *np_it;
+	struct resource res_mem;
+	int err;
+
+	mem_np = of_find_node_by_name(of_root, "reserved-memory");
+	if(!mem_np)
+		return -EINVAL;
+
+	/* Traverse through all the child nodes */
+	for (np_it = NULL; (np_it = of_get_next_child(mem_np, np_it)) != NULL;) {
+		if (strcmp(np_it->name, "buffer") == 0) {
+			err = of_address_to_resource(np_it, 0, &res_mem);
+			if (!err) {
+				/* Check the given address and size fall
+				 * in this reserved memory region
+				 */
+				if (start_addr >= res_mem.start &&
+						size <= resource_size(&res_mem))
+					return true;
+			}
+		}
+	}
+
+	return false;
+}
+
 /*
  * Initialize the memory structure in zocl driver based on the memory
  * topology extracted from xclbin.
@@ -1023,7 +1056,8 @@ void zocl_init_mem(struct drm_zocl_dev *zdev, struct mem_topology *mtopo)
 		memp->zm_used = 1;
 		INIT_LIST_HEAD(&memp->zm_mm_list);
 
-		if (strstr(md->m_tag, "DDR") && (memp->zm_base_addr == 0x0)) {
+
+		if (!check_for_reserved_memory(memp->zm_base_addr, memp->zm_size)) {
 			memp->zm_type = ZOCL_MEM_TYPE_CMA;
 			continue;
 		}
--- a/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
@@ -1044,7 +1044,7 @@ void zocl_init_mem(struct drm_zocl_dev *zdev, struct mem_topology *mtopo)
 			if ((i == j) || !mtopo->m_mem_data[j].m_used)
 				continue;
 
-			if (strstr(mtopo->m_mem_data[i].m_tag, mtopo->m_mem_data[j].m_tag) && 
+			if (strcmp(mtopo->m_mem_data[i].m_tag, mtopo->m_mem_data[j].m_tag) && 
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -967,17 +1021,36 @@ void zocl_init_mem(struct drm_zocl_dev *zdev, struct mem_topology *mtopo)
...
+	for (i = 0; i < zdev->num_mem; i++) {
+		memp = &zdev->mem[i];
+		if (!memp->zm_used)
+			continue;
+
+		for (j = 0; j < zdev->num_mem; j++) {
+			if ((i == j) || !mtopo->m_mem_data[j].m_used)
+				continue;
+
+			if (strstr(mtopo->m_mem_data[i].m_tag, mtopo->m_mem_data[j].m_tag) &&
```
chvamshi-xilinx, line 1047: strstr just looks for substrings. We need full comparision. Use strcmp

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -135,12 +135,114 @@ void zocl_free_userptr_bo(struct drm_gem_object *gem_obj)
...
+				break;
+			}
+		}
+
+		/* No memory left to this memory manager.
+		 * Try to allocate from similer memory manger link list
+		 */
+		mem = list_entry(mem->zm_mm_list.next, typeof(*mem), zm_mm_list);
+
+	} while (&mem->zm_mm_list != &head_mem->zm_mm_list);
```
chvamshi-xilinx, line 209: When does this condition becomes true? You are not adding its own entry in the list. You may need to add a null check?

chvamshi-xilinx (changes requested):
Overall, changes are looking good. Couple of clarifications.
Please verify following cases in VCK190 as it is the only one which has all three types of memories.

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
saifuddin-xilinx: I am using linux list API. Which by default create a circular link list. Hence we need this check.

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -967,17 +1021,36 @@ void zocl_init_mem(struct drm_zocl_dev *zdev, struct mem_topology *mtopo)
 		/* In mem_topology, size is in KB */
 		memp->zm_size = md->m_size * 1024;
 		memp->zm_used = 1;
+		INIT_LIST_HEAD(&memp->zm_mm_list);
 
-		if (!strstr(md->m_tag, "MIG") && !strstr(md->m_tag, "LPDDR")) {
+		if (strstr(md->m_tag, "DDR") && (memp->zm_base_addr == 0x0)) {
```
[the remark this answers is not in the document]
saifuddin-xilinx: Couldn't find any API which provide the CMA range in Linux

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -135,12 +135,114 @@ void zocl_free_userptr_bo(struct drm_gem_object *gem_obj)
...
+
+	do { 
+		if (mem->zm_type == ZOCL_MEM_TYPE_CMA) {
+			struct drm_zocl_bo *cma_bo = zocl_create_cma_mem(dev, size);
+			if (!IS_ERR(cma_bo)) {
+				/* Get the memory in CMA memory region */
+				mutex_unlock(&zdev->mm_lock);
+				kfree(bo->mm_node);
+				kfree(bo);
+				drm_gem_object_release(&bo->gem_base);
```
houlz0507, line 190: bo already freed?

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -135,12 +135,114 @@ void zocl_free_userptr_bo(struct drm_gem_object *gem_obj)
...
+			if (!err) {
+				/* Got memory from this Range memory manager */
+				break;
+			}
+		}
+
+		/* No memory left to this memory manager.
+		 * Try to allocate from similer memory manger link list
+		 */
+		mem = list_entry(mem->zm_mm_list.next, typeof(*mem), zm_mm_list);
```
houlz0507, line 207: Why do not use list_next_entry? what if the mem is last entry?

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -135,12 +135,114 @@ void zocl_free_userptr_bo(struct drm_gem_object *gem_obj)
...
+
+	} while (&mem->zm_mm_list != &head_mem->zm_mm_list);
+	
+	if (err) {
+		DRM_ERROR("Fail to allocate BO: size %ld\n",
+				(long)size);
+		mutex_unlock(&zdev->mm_lock);
+		kfree(bo->mm_node);
+		kfree(bo);
+		drm_gem_object_release(&bo->gem_base);
```
houlz0507, line 217: bo is freed?

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -967,17 +1021,36 @@ void zocl_init_mem(struct drm_zocl_dev *zdev, struct mem_topology *mtopo)
...
+		if (strstr(md->m_tag, "DDR") && (memp->zm_base_addr == 0x0)) {
 			memp->zm_type = ZOCL_MEM_TYPE_CMA;
 			continue;
 		}
 
 		memp->zm_mm = vzalloc(sizeof(struct drm_mm));
-		memp->zm_type = ZOCL_MEM_TYPE_PL_LP_DDR;
-
+		memp->zm_type = ZOCL_MEM_TYPE_RANGE_ALLOC;
+
```
larry9523, line 1033: extra space in blank line

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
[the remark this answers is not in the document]
saifuddin-xilinx: Updated the logic

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
saifuddin-xilinx: Updated.

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
saifuddin-xilinx: Fixed. Thanks

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
saifuddin-xilinx: Fixed. Thanks

src/runtime_src/core/edge/drm/zocl/zocl_bo.c
```diff
@@ -135,12 +136,114 @@ void zocl_free_userptr_bo(struct drm_gem_object *gem_obj)
...
+	}
+
+	mutex_unlock(&zdev->mm_lock);
+
+	/*
+	 * Set up a kernel mapping for direct BO access.
+	 * We don't have to fail BO allocation if we can
+	 * not establish the kernel mapping. We just can not
+	 * access BO directly from kernel.
+	 */
```
keryell, line 229: Oh, some big comment! This is always appreciated...
Go on! :-)

saifuddin-xilinx: Thanks

chvamshi-xilinx (approved):
Thank you. Looks good now. 
Please add a Warning message whenever You are defaulting to CMA by providing a reason
Also, Print a warning message when App asked for non-cacheable but you are allocating cacheable.

maxzhen:
> 
> 
> Thank you. Looks good now.
> Please add a Warning message whenever You are defaulting to CMA by providing a reason
> Also, Print a warning message when App asked for non-cacheable but you are allocating cacheable.

I think that it's OK to allocate non-cacheable if user ask for cacheable, which may or may not cause performance drop. But it should never be OK to allocate cacheable if user ask for non-cacheable. User will not flush cache and will result in error. I'm not sure if a warning msg is good enough?

chvamshi-xilinx:
> > Thank you. Looks good now.
> > Please add a Warning message whenever You are defaulting to CMA by providing a reason
> > Also, Print a warning message when App asked for non-cacheable but you are allocating cacheable.
> 
> I think that it's OK to allocate non-cacheable if user ask for cacheable, which may or may not cause performance drop. But it should never be OK to allocate cacheable if user ask for non-cacheable. User will not flush cache and will result in error. I'm not sure if a warning msg is good enough?

Hi @maxzhen , I should have clarified bit better.  

- If user asks for Cacheable, We are allocating in non-cacheable area if Cacheable memory is not available. (for backward compatibility)

- If user asks for Non-Cacheable, We are allocating Non-cacheable only, but We are allocating from CMA area which user can use for Cacheable memory. We can give a warning/info saying that allocating non-cacheable memory from CMA.

CMA memory is the only cacheable memory in edge.

maxzhen:
> 
> 
> > > Thank you. Looks good now.
> > > Please add a Warning message whenever You are defaulting to CMA by providing a reason
> > > Also, Print a warning message when App asked for non-cacheable but you are allocating cacheable.
> > 
> > 
> > I think that it's OK to allocate non-cacheable if user ask for cacheable, which may or may not cause performance drop. But it should never be OK to allocate cacheable if user ask for non-cacheable. User will not flush cache and will result in error. I'm not sure if a warning msg is good enough?
> 
> Hi @maxzhen , I should have clarified bit better.
> 
>     * If user asks for Cacheable, We are allocating in non-cacheable area if Cacheable memory is not available. (for backward compatibility)
> 
>     * If user asks for Non-Cacheable, We are allocating Non-cacheable only, but We are allocating from CMA area which user can use for Cacheable memory. We can give a warning/info saying that allocating non-cacheable memory from CMA.
> 
> 
> CMA memory is the only cacheable memory in edge.
> 
> Updated the comment above aswell.

OK, this is clearer. I'm not sure if we need to issue this warning msg if what we're doing (allocating non-cacheable memory) is what user is asking for. We warn user because the memory is from CMA area? Can we allocate from some other places? I thought that we need to allocate memory that is accessible by CU, so the connectivity decides everything. Since connectivity is created by user, we will be warning user something he already knew about?

chvamshi-xilinx:
> > > > Thank you. Looks good now.
> > > > Please add a Warning message whenever You are defaulting to CMA by providing a reason
> > > > Also, Print a warning message when App asked for non-cacheable but you are allocating cacheable.
> > > 
> > > 
> > > I think that it's OK to allocate non-cacheable if user ask for cacheable, which may or may not cause performance drop. But it should never be OK to allocate cacheable if user ask for non-cacheable. User will not flush cache and will result in error. I'm not sure if a warning msg is good enough?
> > 
> > 
> > Hi @maxzhen , I should have clarified bit better.
> > ```
> > * If user asks for Cacheable, We are allocating in non-cacheable area if Cacheable memory is not available. (for backward compatibility)
> > 
> > * If user asks for Non-Cacheable, We are allocating Non-cacheable only, but We are allocating from CMA area which user can use for Cacheable memory. We can give a warning/info saying that allocating non-cacheable memory from CMA.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > CMA memory is the only cacheable memory in edge.
> > Updated the comment above aswell.
> 
> OK, this is clearer. I'm not sure if we need to issue this warning msg if what we're doing (allocating non-cacheable memory) is what user is asking for. We warn user because the memory is from CMA area? Can we allocate from some other places? I thought that we need to allocate memory that is accessible by CU, so the connectivity decides everything. Since connectivity is created by user, we will be warning user something he already knew about?

XRT still looks the connectivity and allocates only in those banks where CU is connected. But, there is use-case from DPU that, they will connect CU's to both DDR and LPDDR. DDR is cacheable but LPDDR memory is non-cacheable. We just need to add warning/info in dmesg to warn them because of CMA.

maxzhen:
@chvamshi-xilinx , thanks for your clarification. looks fine then.

houlz0507 (approved):

keryell:
> @chvamshi-xilinx , thanks for your clarification. looks fine then.

But this clarification will be lost because it is in a review and not in the comments at the first time... :-(
So I suggest to add the comments in this PR.

chvamshi-xilinx:
> > @chvamshi-xilinx , thanks for your clarification. looks fine then.
> 
> But this clarification will be lost because it is in a review and not in the comments frm the beginning... :-(
> So I suggest to add the comments in this PR.

@saifuddin-xilinx , Can you please add a clear comment in the code.

larry9523 (approved):

PR #5287, opened 2021-05-20
CR-1081475: Issues with device trace on NoDMA platform

jvillarre:
Reading trace from the device may throw an exception in rare cases.  This pull request catches those exceptions and continues on to reading counters and other devices in those rare cases.  Additionally, this pull request removes useless code that could cause seg faults in rare instances when the xrt_xocl::device has been destroyed or invalidated before counter offload happens at the end of execution.


hackwa (approved):
looks good

IshitaGhosh (approved):
Changes look good.

PR #5289, opened 2021-05-21
Three fixes for AIE event trace & profiling

pgschuey:
1. For now, 'all' is the same as 'functions_all_stalls'. Resource manager has limited support for combo events nor were they defined in our plugin.
2. To support flush for the data mover with burst size of 128 (formerly 16), we needed to increase the timer count to flush out more dummy packets.
3. Group DMA mask was changed to 0xf000 (all 4 stalled locks acquired)


src/runtime_src/xdp/profile/plugin/aie_trace/aie_trace_plugin.cpp
```diff
@@ -781,8 +786,10 @@ namespace xdp {
     //        We need 64 packets, so we need 7 * 64 = 448 words.
     //        Existing counters generates 1.5 words for every perf counter 2 triggers.
     //        Thus, 300 (299 exactly,, but 300 would have no harm) perf 2 counter events.
-    uint32_t timerTrigValueLow  = 0x129A92C0;
-    uint32_t timerTrigValueHigh = 0x0;
+    //uint32_t timerTrigValueLow  = 0x129A92C0;
+    //uint32_t timerTrigValueLow  = 0x4A6A4B00;
+    uint32_t timerTrigValueLow  = 0x29A92C00;
```
hackwa, line 791: @pgschuey was this tested on a big design like beamformer? Just want to make sure we don't run into overruns.

jvillarre (approved):
Change looks good.

PR #5292, opened 2021-05-24
kernel execution models are revised

uday610:
kernel execution models are revised.

Doc updated
Added note regarding XRT new API 
Added note regarding user-managed kernel


stsoe:
ok to test

PR #5294, opened 2021-05-24
CR-1100856 u50 NoDMA PLRAM read using sub-buffers causes panic

stsoe:
Fix xrt::bo sub buffer sync to handle the case where parent buffer is
a nodma buffer.

General unrelated cleanup, remove dead code.

Prevent kernel panic in case user land calls sync ioctl on nodma buffer.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_bo.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_bo.c
@@ -748,14 +748,15 @@ int xocl_sync_bo_ioctl(struct drm_device *dev,
 
 	xobj = to_xocl_bo(gem_obj);
 	BO_ENTER("xobj %p", xobj);
-	sgt = xobj->sgt;
-	sg = sgt->sgl;
 
 	if (!xocl_bo_sync_able(xobj->flags)) {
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_bo.c
```diff
@@ -748,15 +748,16 @@ int xocl_sync_bo_ioctl(struct drm_device *dev,
 
 	xobj = to_xocl_bo(gem_obj);
 	BO_ENTER("xobj %p", xobj);
-	sgt = xobj->sgt;
-	sg = sgt->sgl;
 
-	if (!xocl_bo_sync_able(xobj->flags)) {
+	if (!xocl_bo_sync_able(xobj->flags) || !xobj->sgt) {
```
maxzhen, line 752: Wondering if we don't have to change this line? Basically, checking !xocl_bo_sync_able(xobj->flags) might be good enough... It makes sense to first do the !xocl_bo_sync_able(xobj->flags) check before start doing the real job.

maxzhen (changes requested):

maxzhen (approved):

stsoe:
retest this please - no idea why it failed, but there were a few force pushes that may have caused this?

xilinxgitops:
retest this please

PR #5296, opened 2021-05-24
Fix the host code for test 003_bringup0 

haeseunglee:
Add the missing data migration to the host code.


maxzhen:
ok to test

stsoe:
ok to test - why is this not testing?

PR #5298, opened 2021-05-25
CR-1100413: Fixing the get cu index API for kds_mode=0

chvamshi-xilinx:
This change contains the fix to get the CU index in old kds mode. Taking decision based on the kds_mode sysfs node.

### DIFF
--- a/src/runtime_src/core/edge/user/shim.cpp
+++ b/src/runtime_src/core/edge/user/shim.cpp
@@ -926,20 +926,12 @@ int
 shim::
 xclIPName2Index(const char *name)
 {
-  uint32_t kds_mode;
-  try {
-    kds_mode = xrt_core::device_query<xrt_core::query::kds_mode>(mCoreDevice);
-  } catch (...){
-    kds_mode = 0;
-  }
-
-  //if kds_mode is enabled, then use kds_cu_stat to get the ip index. Otherwise
-  //use old mechanism to get the cu index
+  //if kds_mode is enabled, use kds_cu_stat to get the ip index. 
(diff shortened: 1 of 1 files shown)


mamin506 (approved):

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -926,18 +926,30 @@ int
...
+      for (auto& stat : xrt_core::device_query<xrt_core::query::kds_cu_stat>(mCoreDevice))
+        if (stat.name == name)
+          return stat.index;
+
+      xclLog(XRT_ERROR, "%s not found", name);
+      return -ENOENT;
+    }
+    catch (const xrt_core::query::no_such_key&) {
+    }
+  }
```
stsoe, line 952: Can be simplified I think:

```
try {
  if (xrt_core::device_query<xrt_core::query::kds_mode>(mCoreDevice)) {
    for (const auto& stat : ...)
      ...
 
   return -ENOENT;
}
catch (const xrt_core::query::no_such_key&) {
}
```

stsoe (changes requested):

src/runtime_src/core/edge/user/shim.cpp
chvamshi-xilinx: Sure @stsoe , will make that change

chvamshi-xilinx: @stsoe , Updated the code. Please review.

rbramand-xilinx (approved):

PR #5309, opened 2021-05-26
CR-1099159: Fixing VAI issue where EXEC_WRITE shouldnt skip any words for edge

chvamshi-xilinx:
This change contains a fix where VAI is using EXEC_WRITE and XRT is skipping first six bytes from that command packet
We are not skipping any words for edge In old KDS, but in new KDS, we are skipping six words which is causing their tests to fail.

TODO in next release:
We should definitely remove this difference between edge/DC in next release. 
We should also identify from where this 6 words skipping request came from for DC.

### DIFF
--- a/src/runtime_src/core/common/drv/include/kds_command.h
+++ b/src/runtime_src/core/common/drv/include/kds_command.h
@@ -111,7 +111,8 @@ void cfg_ecmd2xcmd(struct ert_configure_cmd *ecmd,
 void start_krnl_ecmd2xcmd(struct ert_start_kernel_cmd *ecmd,
 			  struct kds_command *xcmd);
 void exec_write_ecmd2xcmd(struct ert_start_kernel_cmd *ecmd,
-			  struct kds_command *xcmd);
+			  struct kds_command *xcmd,
+			  u32 skip);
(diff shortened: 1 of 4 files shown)


maxzhen:
Skipping a hard-coded "6" words sounds a very bad design. We should remove this from our code and expose the bad test/use case and fix it there. Considering that this is too late for 2021.1, we could do this for 2021.2. But please file a CR for this now.

chvamshi-xilinx:
> Skipping a hard-coded "6" words sounds a very bad design. We should remove this from our code and expose the bad test/use case and fix it there. Considering that this is too late for 2021.1, we could do this for 2021.2. But please file a CR for this now.

Completely Agree Max. I will file a CR. I too don't know why we have chosen 6 words skip :-(

chvamshi-xilinx:
> > Skipping a hard-coded "6" words sounds a very bad design. We should remove this from our code and expose the bad test/use case and fix it there. Considering that this is too late for 2021.1, we could do this for 2021.2. But please file a CR for this now.
> 
> Completely Agree Max. I will file a CR. I too don't know why we have chosen 6 words skip :-(

Created CR-1101219 for 2021.2

PR #5311, opened 2021-05-26
Minor fixes in xbutil/xbmgmt

AShivangi:
- CR-1088371 "xbutil --new examine" command hangs when a card is not installed property
- Fix xbutil examine to enforce single device
- DRC check misaligned in xbmgmt examine
- CR-1101231 XRT - xbutil2 examine -d [device bdf] reports 'no report generator found for report compute-units'
- CR-1100874 ASTeR TC19.15 - xbutil2 validate fails when p2p is enabled

### DIFF
--- a/src/runtime_src/core/tools/common/XBHelpMenus.cpp
+++ b/src/runtime_src/core/tools/common/XBHelpMenus.cpp
@@ -720,8 +720,10 @@ XBUtilities::produce_reports( xrt_core::device_collection devices,
       consoleStream << dev_desc;
       consoleStream << std::string(dev_desc.length(), '-') << std::endl;
 
+      auto is_ready = xrt_core::device_query<xrt_core::query::is_ready>(device);
+
       for (auto &report : reportsToProcess) {
-        if (report->isDeviceRequired() == false)
+        if (report->isDeviceRequired() == false || !is_ready)
           continue;
 
         boost::property_tree::ptree ptReport;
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/tools/xbutil2/SubCmdExamine.cpp
```diff
@@ -230,7 +230,7 @@ SubCmdExamine::execute(const SubCmdOptions& _options) const
     }
 
     // enforce 1 device specification if multiple reports are requested
-    if(deviceCollection.size() > 1 && (reportsToProcess.size() > 1 && reportNames.front().compare("host") != 0)) {
+    if(deviceCollection.size() > 1 && (reportsToProcess.size() > 1 || reportNames.front().compare("host") != 0)) {
```
rozumx, line 233: The error message and the logic doesn't match.
I expect to see only:
```
if(deviceCollection.size() > 1) {
...
}
```

rozumx (changes requested):
Please refactor the business rule logic of the multiple device error message.

rozumx (approved):
I understand that additional work is need to correctly address this issue.  Being that we have run out of development time, I have approved this change.  Please do correct it, as soon as, 2021.1 goes out the door.

src/runtime_src/core/tools/common/XBHelpMenus.cpp
```diff
@@ -720,8 +720,10 @@ XBUtilities::produce_reports( xrt_core::device_collection devices,
       consoleStream << dev_desc;
       consoleStream << std::string(dev_desc.length(), '-') << std::endl;
 
+      auto is_ready = xrt_core::device_query<xrt_core::query::is_ready>(device);
+
       for (auto &report : reportsToProcess) {
-        if (report->isDeviceRequired() == false)
+        if (report->isDeviceRequired() == false || !is_ready)
```
sonals, line 726: `if (!report->isDeviceRequired() || !is_ready)`

PR #5313, opened 2021-05-26
XGQ hw emu support queue roll over

larry9523:
1. Get rid off a workaroud (bit 15) to detect new entry in polling mode
2. Add support for queue roll over so that we can support continious incoming/completing commands


maxzhen (approved):

chienwei-lan (approved):

maxzhen (changes requested):
Just talked to @chienwei-lan , we should change the definition of tail pointer value to be the "next free slot", instead of the "latest updated slot". This matches the HW init value.

chienwei-lan (approved):

maxzhen (approved):

PR #5316, opened 2021-05-26
write correct amount of data in aie trace

hackwa:
We were writing 4 times the amount of data we got because the counter was incremented by byte count instead of wordcount. This change affects most of xrt trace and needs to be tested well before merge


jvillarre (approved):
Change looks good to convert number of bytes into number of 32-bit words that are dumped.

PR #5284, opened 2021-05-20
 zocl dma enable 64 bit addressing

saifuddin-xilinx:
CR: 
CR-1098008 : Upstream zocl driver patch to enable 64 bit addressing

This PR has the following changes:
1. Enable 64 bit addressing on zocl dma
2. Changes are under the Macro

PR #5324, opened 2021-05-27
 Reverting the DMA BIT MASK settings in edge as it is causing DFx platform to fail (2021.1)

chvamshi-xilinx:
This change is reverting #5284 as this is causing DFx base tests also to fail.

@saifuddin-xilinx ,
Please fix with right fix and verify DFx platform also


chvamshi-xilinx:
CLosing this as we are cherry-picking this change .

PR #5325, opened 2021-05-27
xbutil xbmgmt reset related changes

AShivangi:
- CR-1101288 xbmgmt reset -r hot -d BDF doesnot work
- CR-1101289 request to use -t for --type instead of -r for reset


rozumx (approved):
Changes look good.

PR #5331, opened 2021-05-27
CR-1079617: Add Guidance for AIE counters usage (active vs.. stalls)

pgschuey:
* made distinction in reporting of AIE core and memory module counter resources
* turned off write to group error enable register since it's blocked


jvillarre (approved):
Additional guidance looks good

hackwa (approved):

PR #5332, opened 2021-05-28
CR-1100943 buffer allocation fix in case of no_dma

venkatp-xilinx:
CR-1100943 buffer allocation fix in case of no_dma


stsoe (changes requested):
Thanks, just one minor comment.

PR #5334, opened 2021-05-28
CR-1093969 support older shell's hello kernels in iops test

rajkumar-xilinx:
U2 older shell (xilinx_u2_gen3x4_xdma_gc_base_1) has hello kernel's
cu name defined as hello_cu0. So, support this case also in iops test


rajkumar-xilinx:
dupe of https://github.com/Xilinx/XRT/pull/5335

PR #5338, opened 2021-05-28
Fix C-API error return code

stsoe:
The C-API implementation in most cases catches C++ exception, where
payload is POSIX error code

Set thread local errno to positive POSIX error code from exception
https://man7.org/linux/man-pages/man3/errno.3.html
https://en.cppreference.com/w/cpp/error/errc
Return -1 documented error return value to indicate error


larry9523 (approved):
Looks good. Thanks for fixing this!

PR #5339, opened 2021-05-28
CR-1100341 Copy Buffer Examples failing - Host application did not complete - pass/end string not found. Exiting as a failure

mamin506:
1. Move zocl_dma_channel_instance() to zocl_bo.c, so that it is accessible for both sched_exec.c and zocl_kds.c.
2. Add copybo command support.

Test with hw_emu test case from CR, the test case can pass now.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_kds.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_kds.c
@@ -320,8 +320,8 @@ int zocl_command_ioctl(struct drm_zocl_dev *zdev, void *data,
 
 	zocl_bo = to_zocl_bo(gem_obj);
 	if (!zocl_bo_execbuf(zocl_bo)) {
-		ret = -EINVAL;
-		goto out;
+		DRM_ERROR("Command buffer is not exec buf\n");
+		return -EINVAL;
 	}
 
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/edge/drm/zocl/zocl_kds.c
```diff
@@ -337,35 +385,54 @@ int zocl_command_ioctl(struct drm_zocl_dev *zdev, void *data,
...
+		DRM_ERROR("Unsupport command\n");
+		ret = -EINVAL;
+		goto out1;
+	}
 
 	/* Now, we could forget execbuf */
 	ret = kds_add_command(&zdev->kds, xcmd);
+	return ret;
 
+out1:
```
xdavidz, line 429: literally, I don't "like" out1, out2, out etc. We'd better have a reasonable name and the next developer can follow what you mean and add his "out2" to a reasonable name. 

Given the scheduler is tied, we might fix this later.

xdavidz (approved):

maxzhen (approved):

src/runtime_src/core/edge/drm/zocl/zocl_kds.c
mamin506: This is a good suggestion, let me enhance this later

chvamshi-xilinx (approved):

PR #5340, opened 2021-05-28
write correct amount of aie trace

hackwa:


jvillarre (approved):
Change from master looks good

PR #5355, opened 2021-05-31
CR-1101237 Old to New Map include path for 2RP platforms

uday610:
Added " xbmgmt program --shell " for https://jira.xilinx.com/browse/CR-1101237


stsoe:
ok to test

PR #5361, opened 2021-06-01
sleep for 100ms to allow trace flush

hackwa:


pgschuey (approved):
needed to complete flush

jvillarre (approved):
Change to wait a bit longer for flush to complete looks good.

PR #5370, opened 2021-06-03
CR-1102097 - Fix segmentation fault in aie_only _xclbin load

rbramand-xilinx:
> Added check to return empty vector when mem_topology section is not there, this happens with aie_only_xclbin


chvamshi-xilinx:
Please cherry-pick this change to master

PR #5374, opened 2021-06-03
CR-1102097 fixed coverity warning

sadasivanshaiju:


xuhz (approved):

stsoe (changes requested):
Hi @sadasivanshaiju Something is wrong with this PR :-)
The CR mentioned in the description has already been fixed and the change in device.cpp has already been merged.  I don't think the PR description should contain the CR number and it should not contain device.cpp.  Please update accordingly.

stsoe:
Close this PR.  It is replaced by #5376

PR #5377, opened 2021-06-04
CentOS8 cmake3 no longer exists

stsoe:
Starting with latest CentOS8 updates the default 'cmake' is 3.18.2 and
cmake3 as a link to cmake no longer exists.

This PR changes changes build.sh to check major version of CMake
before complaining that cmake3 is not installed.

xrtdeps.sh is changed for CentOS8 to no longer attempt yum install of
cmake3.  Also removed package dependency on opencv which no longer
exists.


stsoe:
incomplete, redo.

PR #5378, opened 2021-06-04
Fix for CR 1100233 : For 2021.1, disable PL device trace when AIE trace is enabled

IshitaGhosh:
HALDeviceOffloadPlugin creates and uses its won xclDeviceHandles to allocate and manage BOs. This helps in avoiding crashes due to deletion of BOs by XRT when user provided device handles are used. However, AIETracePlugin still uses user provided device handles. This causes failure in exporting BO when both PL device trace and AIE trace are enabled together. In such a case for GMIO trace offload flow, HALDeviceOffloadPlugin owned device handle is used to allocate trace buffer but user provided device handle is used in xclExportBO call. Hence it fails. Ideally allocation and export BO should use the same owned device handle. But that fix will be quite involved and risky at this stage of release. So, as a workaround for 2021.1, PL device trace will be disabled when AIE trace is enabled.


dayeh-xilinx:
retest this please. u50 board corrupted. switched to a different host.

jvillarre (approved):
Changes to disable the loading of the plugin in certain situation looks good.

PR #5379, opened 2021-06-04
Fix build failure for CentOS 8.* OS versions

rbramand-xilinx:
> cmake3 is no longer available in CentOs 8.* versions
> Latest cmake has a dependency on libarchive package which is added in dependency list
> Repos AppStream and PowerTools name is changed for CentOs versions > 8.1 as appstream and powertools. Added a fix for this change
> Removed opencv package for CentOs 8.* versions as it is not available


dayeh-xilinx:
retest this please. u50 board corrupted. switched to a different host.

stsoe (approved):

PR #5382, opened 2021-06-04
Publish xgq.h

larry9523:
Publish xgq.h so that modules out of XRT source tree can use XGQ to talk to XRT.


larry9523:
retest this please

chienwei-lan (approved):

PR #5383, opened 2021-06-04
CR-1102321 system reboot with latest 2021.1 xrt (build from source 20…

xdavidz:
We should not put a mutex inside a interrupt context. Centos can check this issue better. 🔢

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mailbox_versal.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/mailbox_versal.c
@@ -52,7 +52,6 @@ struct mailbox_versal {
 	irqreturn_t             (*mbv_irq_handler)(void *arg);
 	void 			*mbv_irq_arg;
 
-	struct mutex		mbv_lock;
 };
 
 static inline void mailbox_versal_reg_wr(struct mailbox_versal *mbv,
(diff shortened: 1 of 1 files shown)


larry9523 (approved):
What was the mutex try to protect before? If we were protecting something (does not seem to be the case here), we may think of spink lock.

xdavidz:
> What was the mutex try to protect before? If we were protecting something (does not seem to be the case here), we may think of spink lock.

I think we original thought these APIs (get/set) could be called from multiple routines. That's why we want to protect the mbr_status. But current situation is that only ert_user.c will call it inside irq_handle.

I will add a spinlock later.

mamin506 (changes requested):
If the solution is to remove mbv_lock, this change haven't delete all of them. Like, "mutex_init(&mbv->mbv_lock);" in mailbox_versal_probe and "struct mutex mbv_lock;" in struct mailbox_versal.

Could we replace the lock by spinlock?

xdavidz:
> If the solution is to remove mbv_lock, this change haven't delete all of them. Like, "mutex_init(&mbv->mbv_lock);" in mailbox_versal_probe and "struct mutex mbv_lock;" in struct mailbox_versal.
> 
> Could we replace the lock by spinlock?

Switched to spinlock. The original design was not for irq handle, we should enforce and protect it by using spinlock.

mamin506 (approved):

xdavidz:
We don't need lock as of now, remove the lock for cleaner version.

chienwei-lan (approved):

PR #5390, opened 2021-06-07
CR-1101062 separate option for old KDS

sarab96:
Use separate option to run with old KDS.
New KDS does not support API xclExecBufWithWaitList
Default is to use new KDS 2.0

### DIFF
--- a/src/runtime_src/core/common/config_reader.h
+++ b/src/runtime_src/core/common/config_reader.h
@@ -436,6 +436,16 @@ get_xma_cpu_mode()
   return value;
 }
 
+/**
+ * Use XMA with old KDS; Default for XMA is to assume KDS2.0
+ */
+inline bool
+get_xma_kds_old()
+{
+  static bool value = detail::get_bool_value("Runtime.xma_kds_old",false);
+  return value;
+}
+
 inline bool
 get_enable_flat()
 {
--- a/src/xma/include/lib/xmaapi.h
+++ b/src/xma/include/lib/xmaapi.h
@@ -42,6 +42,7 @@ typedef struct XmaSingleton
 {
     XmaHwCfg          hwcfg;
     bool              xma_initialized;
+    bool              kds_old;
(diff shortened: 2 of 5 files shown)


src/runtime_src/core/common/config_reader.h
```diff
@@ -436,6 +436,16 @@ get_xma_cpu_mode()
...
 
+/**
+ * Use XMA with old KDS; Default for XMA is to assume KDS2.0
+ */
+inline bool
+get_xma_kds_old()
+{
+  static bool value = detail::get_bool_value("Runtime.xma_kds_old",false);
+  return value;
+}
```
stsoe, line 447: Who sets this config option?   Someone who knows old KDS is running?
I don't think that is safe.   Can you do a programmatic check for new KDS?

stsoe (changes requested):

src/runtime_src/core/common/config_reader.h
sarab96: Right now upper layers do not have anyway to check for new KDS.
Only driver level code has that info.
In any case to use old KDS special install instruction is required "modprobe xocl kds_mode=0".
So anyone who want to use old KDS can easily add this config ini setting as well.
And default for all will work functionally even if mistake of ini setting.
So I think approach is fine for now and this can be modified when old KDS is removed completely

stsoe: Upper level code can check for KDS mode, it's done in a few places.  But at any rate, I agree with your comment, the false negative makes the test safe.  I am okay leaving it like you have it.

stsoe (approved):

stsoe (commented):
I  am not comprehending the latest changes.  The code lacks comments.

stsoe (changes requested):
Per last comments

sarab96:
updated as per comments

stsoe (approved):

PR #5392, opened 2021-06-08
Added the M2M enablement under param control. Will enable it once the…

venkatp-xilinx:
Added the M2M enablement under param control. Will enable it once the M2M kernel change is addressed

### DIFF
--- a/src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
+++ b/src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
@@ -2491,19 +2491,10 @@ uint32_t HwEmShim::getAddressSpace (uint32_t topology)
       systemUtil::makeSystemCall(sdxTraceKernelFile, systemUtil::systemOperation::REMOVE, "", boost::lexical_cast<std::string>(__LINE__));
     }
 
-    std::string lf = "";
-    if (getenv("ENABLE_HAL_HW_EMU_DEBUG")) {
-      lf = std::string(pPath) + "/hal_log.txt";
-    }
-    else {
-      lf = "";
-    }
-
-    //if ( logfileName && (logfileName[0] != '\0'))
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -2777,42 +2768,46 @@ int HwEmShim::xclCopyBO(unsigned int dst_boHandle, unsigned int src_boHandle, si
     return -1;
   }
    
-  // Disabling the m2m for timebeing as it is not working as expected. So still data is getting transferred thru DMA.
-  // Will enable this logic unless we have a clarity from the m2m hw_emu kernel. Please do not remove this code 
-  /*if ( deviceQuery(key_type::m2m) && getM2MAddress() != 0 ) {
+  // Enabling the copy buffer thru the M2M under the env control. Once the issues regarding this
+  // is resolved from the sim_m2m kernel, we will remove this env control.
+  if (getenv("ENABLE_M2M_KERNEL")) {
```
chvamshi-xilinx, line 2773: Why don't you use one more xrt.ini variable to decide. Its better to stop using the environment variables wherever possible.

venkatp-xilinx: This is very temporary and will stay for a week or so, hence went with this short cut. INI option is generally used for enabling or disabling the feature and that to for a long time. This is added for convivence for time being to facilitate the Hemants team to test and fix the issue from model side.

venkatp-xilinx: as suggested, replaced the ENV with the ini param

chvamshi-xilinx: Thanks @venkatp-xilinx

PR #5394, opened 2021-06-09
Update mailbox example

stsoe:
Add simpler (no_mailbox) test case that use the same kernels (xclbin),
but doesn't use the mailbox and restart feature.

Update the mailbox example (use_mailbox) to conservatively wait for
the pipeline to finish before updating mailbox and writing mailbox to
hw.

### DIFF
--- a/tests/xrt/mailbox/CMakeLists.txt
+++ b/tests/xrt/mailbox/CMakeLists.txt
@@ -1,11 +1,14 @@
 set(TESTNAME "mailbox")
 
-add_executable(${TESTNAME} main.cpp)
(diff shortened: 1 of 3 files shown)


keryell:
It could be nice to remove the use of most of the `unsigned` and the `static_cast`.

stsoe:
> It could be nice to remove the use of most of the `unsigned` and the `static_cast`.

The code is explicit about intent.  data_size is by choice size_t to represent a size entity, kernel arguments are int*, something got to give.  Even if there was overflow (narrowing cast), the example would still be correct.

PR #5400, opened 2021-06-09
Initial implementation of generic profiling summary to replace the OpenCL specific summary

jvillarre:
This pull request disables the writer for the OpenCL summary and replaces it with a generic summary file produced whenever any profiling plugin is loaded.  Now, every execution that enables any profiling aspect will generate a "summary.csv" file with the applicable information in tables.  Each plugin is now responsible for registering with the database the information that is available for the summary file.

This initial submission also adds two new tables to the profile summary for XRT Native API calls and HAL level calls when information exists.


IshitaGhosh (approved):
Changes look good.

hackwa (approved):
initial implementation looks great

PR #5405, opened 2021-06-14
Change profiling summary generation so the proper tables are output in software emulation

jvillarre:


IshitaGhosh (approved):
Change looks good.

PR #5407, opened 2021-06-15
Fixed the issue when interrupt fails and fallback to polling mode for some of the CUs. 

saifuddin-xilinx:
This fixed the following CR.
CR-1102913

This is the fix when interrupt fails and fallback to polling mode for some of the CUs. 
Performance is degrading because of that. 
Observed this on interrupt mode, on polling mode there is no issue.


xilinxgitops:
retest this please

chvamshi-xilinx (approved):

PR #5410, opened 2021-06-15
CR-1102977 Saw message "mailbox error detected, error=0xdec0de1c" whe…

houlz0507:
…n start up VM


xuhz (commented):
PCI_CFG_SPACE_EXP_SIZE in pci_regs.h is what you want
it is standard. no need to define a xocl exp cfg size

houlz0507:
> PCI_CFG_SPACE_EXP_SIZE in pci_regs.h is what you want
> it is standard. no need to define a xocl exp cfg size

It is not defined with earlier kernel. Maybe it is not hurt to define one in header to make sure there is not compiling issue.

xuhz (approved):

PR #5411, opened 2021-06-16
CR-1103270 ERT CQ slot size needs a cap

mamin506:
We should not support > 4KB ERT CQ slot size.


maxzhen (changes requested):
please also add a check to reject CU cmd that exceeds the CQ slot size.

maxzhen (approved):

dayeh-xilinx:
retest this please. I break the boardfarm host by accident. Test again.

PR #5412, opened 2021-06-16
CR-1103258 XRT workaround for CR-1100270

mamin506:
For U30, the xclbin is constructed in a special way. Some CU metadata is missing. This change is to make this kind of xclbin can go through the flow. The missing metadata would be hardcode for U30.

We will file another CR to back this change out.


maxzhen (changes requested):
this change is for https://jira.xilinx.com/browse/CR-1103258, so please mention it in this PR.

maxzhen (approved):

mamin506:
Please merge #5411 before this one. Otherwise, U30 VCU test case would hang.

PR #5414, opened 2021-06-16
Profiling summary improvements

jvillarre:
This pull request adds the following capabilities to the profile summary:

- No longer print out table headers on empty tables
- Remove restriction that compute unit utilization only applies to OpenCL applications.  Now, if device data is available in HW and HW emulation the compute unit utilization table will be output.
- Added four new tables for information collected when Native XRT APIs are traced:
- Host reads from global memory (via sync)
- Host writes to global memory (via sync)
- Top memory reads from global memory (via sync)
- Top memory writes to global memory (via sync)


IshitaGhosh (approved):
Changes look good.

PR #5418, opened 2021-06-18
Move hwemu XGQ driver to use common XGQ APIs

larry9523:
This PR added the following feature
1. Abstracted a common XGQ APIs into xgq.h which can be used by different platform specific code. Each platform can make use of the common XGQ API without caring about the Queue implementation detail. (Co-contributed by @maxzhen and @chienwei-lan )
2. Enhanced hwemu XGQ driver to use common XGQ APIs
3. Redefine the XGQ completion entry to move sqheader field to queue ring buffer header. And reorder completion entry so that the new entry state flag is put at a common place as the same as submission queue entry.

@stsoe could you please help to check if this PR (especially the new xgq.h) breaks Clang-tidy?


src/runtime_src/core/include/xgq.h
```diff
@@ -215,4 +232,284 @@ struct xrt_cmd_exit_ert {
...
+		xgq_mem_write32(io_hdl, tgt, src[i]);
+}
+
+static inline void xgq_copy_from_ring(uint64_t io_hdl, void *buf, uint64_t src, size_t len)
+{
+	size_t i;
+	uint32_t *tgt = (uint32_t *)buf;
+
+	for (i = 0; i < len / 4; i++, src += 4)
+		tgt[i] = xgq_mem_read32(io_hdl, src);
```
stsoe, line 317: This will not compile since xgq_mem_read32 is not visible by a compilation unit that includes just this header file.
The function appears to be defined in xgq_hwemu_plat.h but that is a indirect dependency that shouldn't be there.

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/xgq_hwemu_plat.h
```diff
@@ -0,0 +1,71 @@
...
+ *  License for more details.  You should have received a copy of the
+ *  GNU General Public License along with this program; if not, write
+ *  to the Free Software Foundation, Inc., 59 Temple Place, Suite 330,
+ *  Boston, MA 02111-1307 USA
+ *
+ */
+
+#ifndef __XGQ_HWEMU_PLAT_H__
+#define __XGQ_HWEMU_PLAT_H__
+
```
stsoe, line 40: This header file does not compile, it needs at least <cstdint> and "xgq.h".    But xgq.h it self also doesn't compile because it needs xgq_mem_read32 which is defined in this  header file.

stsoe (changes requested):

maxzhen (approved):

src/runtime_src/core/include/xgq.h
```diff
@@ -215,4 +232,284 @@ struct xrt_cmd_exit_ert {
...
+}
+
+static inline bool xgq_can_produce(struct xgq *xgq)
+{
+	struct xgq_ring *ring = xgq->xq_is_server ? &xgq->xq_cq : &xgq->xq_sq;
+
+	if (!xgq_ring_full(ring))
+		return true;
+	xgq_ring_read_consumed(xgq->io_hdl, ring);
+	return xgq_ring_full(ring);
```
chienwei-lan, line 375: also need to be return !xgq_ring_full(ring);

larry9523: done

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/xgq_hwemu_plat.h
larry9523: xgq_hwemu_plat.h compiles after including <cstdint>

src/runtime_src/core/include/xgq.h
larry9523: The idea here is to abstract a generic XGQ implementation which can be shared by all platforms using XGQ. For platform specific operations like xgq_mem_read32, they need to be defined in a platform specific header file that is included before this generic header.

Any suggestion we can achieve this while xgq.h can be compiled by itself? Or can we make some exception to skip this check for xgq.h?

chienwei-lan (approved):

src/runtime_src/core/include/xgq.h
maxzhen: can we do #ifndef xgq_mem_read32 then define it as a empty function inside xgq.h? Also put some comment here to say that this is just for this check. it won't work when you really run the code for sure.

larry9523: Hmm, this won't work since inside the xgq_xxx_platform.h, functions like xgq_mem_read32 is inline function instead of macro. And we'd better keep them as inline functions.

larry9523: Made the changes as discussed.
1) split xgq.h to xgq.h and xgq_impl.h
2) add XGQ_IMPL macro and include xgq_impl.h in xgq_hwemu_plat.h
3) add virtual platform specific definition in xgq_impl.h if XGQ_IMPL is not defined.

@maxzhen  and @stsoe  please take a look again. Thanks!

maxzhen (approved):

stsoe (changes requested):

src/runtime_src/core/include/xgq_impl.h
```diff
@@ -0,0 +1,347 @@
...
+ *
+ * Note: This file (xgq_impl.h) should only be included by xgq_XXX_plat.h.
+ */
+
+#ifndef XGQ_IMPL
+#define ____cacheline_aligned_in_smp
+static inline void xgq_mem_write32(uint64_t hdl, uint64_t addr, uint32_t val) {}
+static inline void xgq_reg_write32(uint64_t hdl, uint64_t addr, uint32_t val) {}
+static inline uint32_t xgq_mem_read32(uint64_t hdl, uint64_t addr) {}
+static inline uint32_t xgq_reg_read32(uint64_t hdl, uint64_t addr) {}
```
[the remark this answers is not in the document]
larry9523: return 0xFFFFFFFF indicating something is not right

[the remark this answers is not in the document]
larry9523: Wondering why
g++ -std=c++14 -I$XILINX_XRT/include -o /tmp/foo.o core/include/xgq_impl.h
did not catch it?

src/runtime_src/core/include/CMakeLists.txt
```diff
@@ -15,7 +15,8 @@ set(XRT_HEADER_SRC
   xclhal2_mem.h
   xrt_mem.h
   xrt_error_code.h
-  xgq.h)
+  xgq_cmd.h
+  xgq_impl.h)
```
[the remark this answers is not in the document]
larry9523: Yeah, there is some entities which is out of XRT repo may need to talk to XRT through XGQ.

stsoe (approved):

PR #5430, opened 2021-06-22
Add IPU_MEM_HBM_ECC sub-device

chienwei-lan:
To reenable HBM ECC, XRT will create the sub-device for the newly introduced IP type: IPU_MEM_HBM_ECC


rozumx (approved):
Looks good.

PR #5455, opened 2021-06-30
CR-1104200: xbutil warning note

uday610:
added notes about missing xbutil/xbmgmt commands such as "host_mem", though there is no new command exist, XRT still shows warnings (which is a false warning). Added note that the command will be implemented in later release


AShivangi (approved):
Thanks for updating the doc :)

maxzhen:
ok to test

PR #5459, opened 2021-07-01
CR-1103864 Checkout compatible OpenCL libraries using correct tag

pranjalv-xilinx:
Get the compatible version of OpenCL library while setting up Windows build environment for XRT. Current checking out version 2020.3.13


stsoe (approved):
Thank you for fixing this.

I wonder if this is kicking the can down the road and if it really is CL_EXT_SUFFIX__VERSION_2_0 that should be changed?  Regardless, it not a good idea to track main branch of OCL so what you did is correct.  Thanks!

pranjalv-xilinx:
> Thank you for fixing this.
> 
> I wonder if this is kicking the can down the road and if it really is CL_EXT_SUFFIX__VERSION_2_0 that should be changed? Regardless, it not a good idea to track main branch of OCL so what you did is correct. Thanks!

I agree with you. This is more like basic process change to not track main branch.  I believe we should take changing or upgrading library version as a separate "task" and managed accordingly.

PR #5473, opened 2021-07-07
Initialized the buffer to 0 before the buffer is synched with real da…

venkatp-xilinx:
Initialized the buffer to 0 before the buffer is synched with real data for sw_emu.

### DIFF
--- a/src/runtime_src/core/pcie/emulation/cpu_em/generic_pcie_hal2/shim.cxx
+++ b/src/runtime_src/core/pcie/emulation/cpu_em/generic_pcie_hal2/shim.cxx
@@ -1428,7 +1428,7 @@ int CpuemShim::xclCopyBO(unsigned int dst_boHandle, unsigned int src_boHandle, s
       return -1;
   }
   else {
-    std::cerr << "ERROR: Copy buffer from source to destination faliled" << std::endl;
+    std::cerr << "ERROR: Copy buffer from source to destination failed" << std::endl;
     return -1;
   }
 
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/pcie/emulation/cpu_em/generic_pcie_hal2/shim.cxx
```diff
@@ -1484,6 +1484,7 @@ void *CpuemShim::xclMapBO(unsigned int boHandle, bool write)
     if (mLogStream.is_open()) mLogStream << "posix_memalign failed" << std::endl;
     pBuf=nullptr;
   }
+  memset(pBuf, 0, bo->size);
```
stsoe, line 1487: pBuf can be nullptr per line 1485, this will crash the program

stsoe (changes requested):

src/runtime_src/core/pcie/emulation/cpu_em/generic_pcie_hal2/shim.cxx
venkatp-xilinx: Thanks Soren for the feedback, incorporate it and committed the change.

stsoe: Are you sure bo->buf (set on line 1491) will not be garbage when you return without setting it?

venkatp-xilinx (commented):
Please approve this based on the comment added

stsoe (approved):

chvamshi-xilinx (approved):

PR #5474, opened 2021-07-07
remove bool usage from xgq implementation

maxzhen:
Let's avoid using 'bool' in xgq implementation as it's portability across multiple OSes/platforms is poor.


larry9523 (approved):

sonals:
#include <stdbool.h>should work with C

maxzhen:
include stdbool.h causes some issues when compile with lx6 compiler. @larry9523 has details. Integer 1 and 0 works better on all OSes with all compilers on all platforms...

PR #5479, opened 2021-07-08
VITIS-2614 Add C++ APIs for AIE multi-process

stsoe:
Extend xrt_aie.h and xrt_graph.h with C++ access mode support which in
turn facilitates support for multi-process.


larry9523 (approved):
Thanks @stsoe !

PR #5482, opened 2021-07-09
CR-1104824 xbutil: domain in bdf is non-zero value in VMs

rajkumar-xilinx:
pcie domain is non-zero value in VMs, so consider domain parameter as well in bdf checks.


AShivangi (commented):
You might have to make changes in device_*.cpp files as well

chvamshi-xilinx:
@rajkumar-xilinx , you have to make changes to edge device* files as well.

src/runtime_src/core/common/query_requests.h
```diff
@@ -387,7 +387,7 @@ struct pcie_express_lane_width_max : request
 
 struct pcie_bdf : request
 {
-  using result_type = std::tuple<uint16_t,uint16_t,uint16_t>;
+  using result_type = std::tuple<uint16_t, uint16_t, uint16_t, uint16_t>;
```
rozumx, line 390: Unfortunately, the "tuple" value is getting out-of-control for it is resulting in "magic" hard coded entry values.  Unfortunately, moving away from using this may not be possible so what minimally needs to be done done is to:
1. Document what each entry is (including a sample value).
2. Replace the hard coded index values (e.g., 0, 1, 2, etc.) with a static "string" value representing the index value.   

For example: BDF_DOMAIN, BDF_BUS, etc. where BDF_DOMAIN would have a value of 0, BDF_BUS 1, etc.

Note: Please **_do not define_** these values with define statements (e.g., #define BDF_DOMAIN 0)

rozumx (changes requested):
See comments.

src/runtime_src/core/common/query_requests.h
rajkumar-xilinx: Hi @rozumx Shall I add enum for this indexing purpose in query_request.h file, like below

enum class bdf_index
{
 BDF_DOMAIN,
 BDF_BUS,
 BDF_DEV,
 BDF_FUNC,
 BDF_MAX
};

src/runtime_src/core/common/query_requests.h
```diff
@@ -32,6 +32,15 @@ namespace xrt_core {
 
 namespace query {
 
+enum bdf_index
```
AShivangi, line 35: please define an enum class and make all enumerations lower case:
```
enum class bdf {
domain,
bus,
dev,
...
}
```
and calling it would look something like:
`std::get<bdf::bus>(value)`

AShivangi (changes requested):

AShivangi (approved):
Looks good!

src/runtime_src/core/common/query_requests.h
```diff
@@ -32,6 +32,14 @@ namespace xrt_core {
 
 namespace query {
 
+enum bdf
```
rozumx, line 35: Being that these values are used to get in an "index" value,  it is best to:
1. Describe that this enumeration represents.
2. Initialize each value instead of relying on the default behavior.
3. Add a 1 line comment at the initializing of the enumeration value indicating purpose and an example value

Again, the purpose is readability and helping the next developer support this code base.

src/runtime_src/core/pcie/windows/mgmt.cpp
```diff
@@ -210,7 +210,7 @@ struct mgmt
   }
 
   void
-  get_bdf_info(uint16_t bdf[3])
+  get_bdf_info(uint16_t bdf[4])
```
rozumx, line 213: Please refactor out the magic values.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -50,7 +50,7 @@ struct bdf
   get(const xrt_core::device* device, key_type)
   {
     auto pdev = get_pcidev(device);
-    return std::make_tuple(pdev->bus,pdev->dev,pdev->func);
+    return std::make_tuple(pdev->domain, pdev->bus, pdev->dev, pdev->func);
```
rozumx, line 53: Please update copyright date at the beginning of this file.

src/runtime_src/core/pcie/emulation/common_em/query.h
```diff
@@ -69,7 +69,7 @@ struct device_info
     case key_type::kds_numcdmas:
       return static_cast<xrt_core::query::kds_numcdmas::result_type>(info.mNumCDMA);
     case key_type::pcie_bdf:
-      return xrt_core::query::pcie_bdf::result_type{0,device->get_device_id(),0};
+      return xrt_core::query::pcie_bdf::result_type{0, 0, device->get_device_id(), 0};
```
rozumx, line 72: Please update the copyright date at the start of this file.

rozumx (commented):
Once the enumeration values are initialized and the comments added, this PR can be approved.

src/runtime_src/core/pcie/windows/device_windows.cpp
```diff
@@ -520,7 +521,7 @@ struct bdf
     }
 
     auto& bdf = (*it).second;
-    return std::make_tuple(bdf.bus,bdf.device,bdf.function);
+    return std::make_tuple(bdf.domain, bdf.bus, bdf.device, bdf.function);
```
stsoe, line 524: Either convert to a struct under query::pcie_bdf or leave as tuple as done here, but do not mix tuple and the enum.

src/runtime_src/core/common/query_requests.h
```diff
@@ -32,6 +32,26 @@ namespace xrt_core {
...
+ * Retrieve the individual elements from bdf tuple using below enum bdf members.
+ */
+enum bdf
+{
+  domain = 0,
+  bus = 1,
+  dev = 2,
+  func = 3,
+  max = 4,
+};
```
stsoe, line 53: Plain enums are not allowed under any circumstances in C++.   I see the enum values used in some indexing 
 in string conversion in XBUtilities.cpp, but it doesn't accomplish anything when make_tuple is still used.

While I think a tuple of 4 entries for 'dbdf' is pretty darn obvious with a comment under struct query::pcie_bdf, I am okay if the tuple is replaced with an embedded struct under struct query::pcie_bdf, but the enum is not needed at all and must be removed.

src/runtime_src/core/pcie/windows/mgmt.cpp
```diff
@@ -210,7 +211,7 @@ struct mgmt
   }
 
   void
-  get_bdf_info(uint16_t bdf[3])
+  get_bdf_info(uint16_t bdf[xrt_core::query::bdf::max])
```
stsoe, line 214: Probably better to use 
```
xrt_core::query::pcie_bdf::dbdf
get_bdf_info() 
{
 ...
};
```
where dbdf is a struct.     Either use the tuple of 4 entries or switch everything to embedded struct.

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -458,15 +462,18 @@ str2index(const std::string& str, bool _inUserDomain)
...
     int idx(boost::lexical_cast<int>(str));
     auto device = _inUserDomain ? xrt_core::get_userpf_device(idx) : xrt_core::get_mgmtpf_device(idx);
 
     auto bdf = xrt_core::device_query<xrt_core::query::pcie_bdf>(device);
     // if the bdf is zero, we are dealing with an edge device
-    if(std::get<0>(bdf) == 0 && std::get<1>(bdf) == 0 && std::get<2>(bdf) == 0) {
+    if(std::get<xrt_core::query::bdf::domain>(bdf) == 0 &&
+       std::get<xrt_core::query::bdf::bus>(bdf) == 0 &&
+       std::get<xrt_core::query::bdf::dev>(bdf) == 0 &&
+       std::get<xrt_core::query::bdf::func>(bdf) == 0) {
```
stsoe, line 476: Ditto

stsoe (changes requested):
Thanks for suggesting enum, however, if we decide to drop the not-so-magic (IMO) tuple, then please do it with an embedded struct and propagate all necessary changes throughout XRT.

src/runtime_src/core/pcie/windows/device_windows.cpp
rajkumar-xilinx: In this line, there is no enum used, the variable "bdf" is array of uint16 received from windows/mgmt.cpp.
Using this array, a tuple is made.

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -435,14 +437,16 @@ bdf2index(const std::string& bdfstr, bool _inUserDomain)
...
       if (func != std::numeric_limits<uint16_t>::max())
-        return func == std::get<2>(bdf);
+        return func == std::get<xrt_core::query::bdf::func>(bdf);
       return true;
     };
 
-    if (bus == std::get<0>(bdf) && dev == std::get<1>(bdf) && cmp_func(func))
+    if (domain == std::get<xrt_core::query::bdf::domain>(bdf)
+        && bus == std::get<xrt_core::query::bdf::bus>(bdf)
+        && dev == std::get<xrt_core::query::bdf::dev>(bdf) && cmp_func(func))
```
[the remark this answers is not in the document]
rajkumar-xilinx: this line is for removing hard coded values 0, 1, 2 by using enum. 
If enum itself is not correct then the entire patch needs to be modified again.

src/runtime_src/core/pcie/windows/device_windows.cpp
stsoe: I understand, my point is that the enum is useless and doesn't provide any safety at all.  Either convert to embedded struct pcie_bdf { struct dbdf { ... }; ... }; 
or stay with the tuple as before adding the enum.   At least get rid of the enum under all circumstances.

src/runtime_src/core/tools/common/XBUtilities.cpp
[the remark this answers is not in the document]
stsoe: Yeah, the enum is not great. Please remove it, stay with tuple and magic 0,1,2,3, or convert to embedded struct with 4 members (which really is exactly what a tuple is).

rajkumar-xilinx:
Hi @stsoe @rozumx ,
Adding a embedded struct in place of tuple requires changes in linux/edge as well, and I am not confident in adding this change. So, I have removed enum code completely and moved back to index way.

Please take care of adding suggested "embedded struct" in pcie_bdf change in your next PRs or code clean up patches. Thank you.

src/runtime_src/core/pcie/windows/mgmt.cpp
rajkumar-xilinx: I removed enum and using indexes only.

src/runtime_src/core/common/query_requests.h
rajkumar-xilinx: Removed enum code.

src/runtime_src/core/pcie/windows/device_windows.cpp
rajkumar-xilinx: removed enum and stay wit tuple for now. Please clean up this part in next PRs.

stsoe:
Thank you @rajkumar-xilinx  I agree with you ponding on creating an embedded struct, which would have ripple effects as you note.

stsoe (approved):

rozumx (approved):
Approved.  Thank you.

PR #5485, opened 2021-07-12
Fix for incomplete AIE trace packet : Ignore the last words if they do not make a complete 4-word packet

IshitaGhosh:


hackwa (approved):
don't write partial aie trace data to files

pgschuey (approved):

stsoe (approved):
It's magic :-) Not obvious why wordCount has to be a multiple of 4?
What is a word?

IshitaGhosh:
Hi Soren, a complete AIE trace packet has 4 words. Thanks!

keryell:
> Hi Soren, a complete AIE trace packet has 4 words. Thanks!

I guess what @stsoe means is that we should improve the code quality of XRT by not committing magical code without comments or even better use instead clearer abstraction easy to understand without comments.
You could argue that it is trivial code, but if it was trivial there would have been no bug at the first place. ;-)
Some guidance:
- if your N+2 manager does not understand the code, add some comments; :-)
- if @stsoe finds some magic in the code, double the amount of comments. ;-)

PR #5490, opened 2021-07-14
added metric sets to AIE profiling

pgschuey:


src/runtime_src/core/edge/common/aie_parser.cpp
```diff
@@ -446,6 +483,19 @@ get_tiles(const xrt_core::device* device, const std::string& graph_name)
   return ::get_tiles(aie_meta, graph_name);
 }
 
+std::vector<tile_type>
+get_event_tiles(const xrt_core::device* device, const std::string& graph_name,
+                e_tile_type type)
+{
+  auto data = device->get_axlf_section(AIE_METADATA);
+  if (!data.first || !data.second)
+    return std::vector<tile_type>();
```
stsoe, line 492: `return {};`

src/runtime_src/core/edge/common/aie_parser.h
```diff
@@ -28,6 +28,13 @@ class device;
 
 namespace edge { namespace aie {
 
+enum e_tile_type {
```
stsoe, line 31: Please use class enum and lower case enumerators, which are scoped.  Let's never again write a plain enum unless we are writing C code :-)

src/runtime_src/xdp/profile/plugin/aie_trace/aie_trace_plugin.cpp
```diff
@@ -795,17 +814,22 @@ namespace xdp {
     // Create AIE Trace Offloader
     AIETraceDataLogger* aieTraceLogger = new AIETraceDataLogger(deviceId);
 
-    std::string flowType = (isPLIO) ? "PLIO" : "GMIO";
-    std::string msg = "Total " + std::to_string(aieTraceBufSize) + " size is used for AIE trace buffer for " + std::to_string(numAIETraceOutput) + " " + flowType + " streams.";
-    xrt_core::message::send(xrt_core::message::severity_level::debug, "XRT", msg);
+    if (xrt_core::config::get_verbosity() >= (uint32_t)xrt_core::message::severity_level::debug) {
```
stsoe, line 817: static_cast

src/runtime_src/core/edge/common/aie_parser.cpp
```diff
@@ -208,6 +209,42 @@ get_tiles(const pt::ptree& aie_meta, const std::string& graph_name)
...
+  std::vector<tile_type> tiles;
+
+  // Not supported yet
+  if (type == xrt_core::edge::aie::AIE_TILE_SHIM)
+    return tiles;
+
+  const char* key_cols[xrt_core::edge::aie::AIE_TILE_NUM_TYPES] =
+      {"core_columns", "dma_columns", "N/A"};
+  const char* key_rows[xrt_core::edge::aie::AIE_TILE_NUM_TYPES] =
+      {"core_rows", "dma_rows", "N/A"};
```
stsoe, line 225: Regarding the enum see later.

`const std::array<const char*,  static_cast<int>(xrt_core::edge::aie::e_tile_type::tile_num_types)> key_cols = { ... };`

Not much advantage, but clang_tidy would object to the old-style arrays.

src/runtime_src/core/edge/common/aie_parser.cpp
```diff
@@ -208,6 +209,42 @@ get_tiles(const pt::ptree& aie_meta, const std::string& graph_name)
...
+
+  for (auto& graph : aie_meta.get_child("aie_metadata.EventGraphs")) {
+    if (graph.second.get<std::string>("name") != graph_name)
+      continue;
+
+    int count = 0;
+      for (auto& node : graph.second.get_child( key_cols[type] )) {
```
stsoe, line 232: When type becomes an enum class (see later), you will have to` static_cast` the type to integral.  While that looks cumbersome, it is in fact better coding style since your intend is more clearly stated.  For example as is right now, I could be let to think that using `type` as array index is a bug, but if you clearly static_cast it to integral, then it is expressed clearly what the intend is.

src/runtime_src/core/edge/common/aie_parser.cpp
```diff
@@ -208,6 +209,42 @@ get_tiles(const pt::ptree& aie_meta, const std::string& graph_name)
   return tiles;
 }
 
+std::vector<tile_type>
+get_event_tiles(const pt::ptree& aie_meta, const std::string& graph_name,
+                e_tile_type type)
+{
+  std::vector<tile_type> tiles;
```
stsoe, line 216: Move closer to where used, around like 226

src/runtime_src/core/edge/common/aie_parser.cpp
```diff
@@ -208,6 +209,42 @@ get_tiles(const pt::ptree& aie_meta, const std::string& graph_name)
...
+
+  // Not supported yet
+  if (type == xrt_core::edge::aie::AIE_TILE_SHIM)
+    return tiles;
```
stsoe, line 220: `return {};`

stsoe (changes requested):

src/runtime_src/core/edge/common/aie_parser.h
```diff
@@ -28,6 +28,12 @@ class device;
 
 namespace edge { namespace aie {
 
+enum class e_module_type {
```
stsoe, line 31: I have to ask what's the 'e_' in 'e_module_type'?    

The very nice thing about enum classes is that enumerators must be and are scoped by the class name, so you don't need long names anymore, maybe just 'core', 'dma', 'shim'.

hackwa (approved):
looks good

IshitaGhosh (approved):
Looks good.

src/runtime_src/core/edge/common/aie_parser.cpp
```diff
@@ -208,6 +209,40 @@ get_tiles(const pt::ptree& aie_meta, const std::string& graph_name)
...
+        tiles.push_back(tile_type());
+        auto& t = tiles.at(count++);
+        t.col = std::stoul(node.second.data());
+      }
+
+      int num_tiles = count;
+      count = 0;
+      for (auto& node : graph.second.get_child(row_name))
+        tiles.at(count++).row = std::stoul(node.second.data());
+      throw_if_error(count < num_tiles,"rows < num_tiles");
```
keryell, line 240: Look at this make me think that `throw_if_error` has not the right name.
`throw_if_true`? `throw_runtime_error_if_true`?

src/runtime_src/core/edge/common/aie_parser.cpp
keryell: Curious why this was not accepted?

src/runtime_src/core/edge/common/aie_parser.cpp
```diff
@@ -446,6 +481,19 @@ get_tiles(const xrt_core::device* device, const std::string& graph_name)
...
+get_event_tiles(const xrt_core::device* device, const std::string& graph_name,
+                module_type type)
+{
+  auto data = device->get_axlf_section(AIE_METADATA);
+  if (!data.first || !data.second)
+    return std::vector<tile_type>();
+
+  pt::ptree aie_meta;
+  read_aie_metadata(data.first, data.second, aie_meta);
+  return ::get_event_tiles(aie_meta, graph_name, type);
```
keryell, line 494: Are these top level functions could conflict with user-defined functions?

stsoe: @keryell I think impossible since ::get_event_tiles is in TU with internal linkage, and user-defined function is definitely not.

src/runtime_src/core/edge/common/aie_parser.cpp
stsoe: Argh...   I'd expect throw_if_error(...) to check the return value of a function (or condition) and flag an error if the value is different from 0.  Alas I think throw_if_error is the right name, you pass it a value that represents the presence of an error.

jvillarre (approved):

src/runtime_src/core/edge/common/aie_parser.cpp
pgschuey: This file is maintained by another development team, and my return statement is consistent with all other functions.

PR #5497, opened 2021-07-16
Vitis-1144 Software resettable kernels

mamin506:
1. KDS could support abort command from user space. The payload is the bo handle of execbuf command ('target command' in below) to abort
2. Pass down the sw_reset bit to CU subdevice while load xclbin.
3. KDS try submit a abort command to each CU until a CU find target command to abort. If target command is not found by any CU, user would get ERT_CMD_STATE_ERROR from the abort command.
4. If CU foundtarget command, and it isn't submitted to CU hardware, abort it from the internal running queue.
5. If CU found target command, and it is submitted to CU hardware, try reset CU and wait about 100 us then check if CU reset:
     a) If CU reset is done, abort target command and return ERT_CMD_STATE_COMPLETE to user.
     6) If CU reset is not done, abort target command and ask for reset the whole board. User would get ERT_CMD_STATE_TIMEOUT.

Since HLS bug, I have no positive test case, that sw reset works. I mimic different hang situations by modified driver to test different code path. I have done some simple test:  
    
1. One CU xclbin on U50
2. Multiple CUs xclbin on U200

About performance, test with xbutil validate on U50 and U200, the IOPS test case shows similar result w/ and wo/ this change.

### DIFF
--- a/src/runtime_src/core/common/drv/xrt_cu.c
+++ b/src/runtime_src/core/common/drv/xrt_cu.c
@@ -259,6 +259,100 @@ static inline void process_pq(struct xrt_cu *xcu)
 		xcu->max_running = xcu->num_rq;
 }
 
+static inline void
+try_abort_cmd(struct xrt_cu *xcu, struct kds_command *abort_cmd)
+{
+	struct kds_command *xcmd;
+	struct kds_command *tmp;
+	u32 handle;
+
+	/* Never call this function on the performance critical path */
+	handle = *(u32 *)abort_cmd->info;
+	list_for_each_entry_safe(xcmd, tmp, &xcu->rq, list) {
+		if (xcmd->exec_bo_handle != handle)
+			continue;
+
+		/* Found the xcmd to abort! */
+		abort_cmd->status = KDS_COMPLETED;
+
+		xcu_info(xcu, "Abort command(%d) on running queue", handle);
+		xcmd->status = KDS_ABORT;
+		move_to_queue(xcmd, &xcu->cq, &xcu->num_cq);
+		--xcu->num_rq;
+		return;
+	}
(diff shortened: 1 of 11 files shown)


src/runtime_src/core/common/drv/xrt_cu.c
```diff
@@ -259,6 +259,100 @@ static inline void process_pq(struct xrt_cu *xcu)
...
+
+	list_for_each_entry_safe(xcmd, tmp, &xcu->sq, list) {
+		if (xcmd->exec_bo_handle != handle)
+			continue;
+
+		xcu_info(xcu, "Abort command(%d) on submitted queue", handle);
+		/* Found the xcmd to abort! */
+		if (!xcu->info.sw_reset) {
+			xcu_warn(xcu, "No sw resset. Device goto bad state");
+			abort_cmd->status = KDS_TIMEOUT;
```
maxzhen, line 293: do we want to wait for a few seconds before claim that the cmd has timed out?

maxzhen (changes requested):
when we call xrt_cu_abort() for a client, should we also reset cu, if a cmd is currently running on the cu?

src/runtime_src/core/common/drv/xrt_cu.c
mamin506: I assume user would only send abort command down when he thought CU is hang or something is wrong. This is why I don't delay in here. I could wait a few seconds at this place.

maxzhen: Oh, then you need to return KDS_ABORTED, not KDS_TIMEOUT.

maxzhen (approved):

PR #5504, opened 2021-07-20
CR-1104824 xbutil: domain in bdf is non-zero value in VMs (#5482)

rajkumar-xilinx:


chvamshi-xilinx:
Merging this as this is needed for customer.

PR #5507, opened 2021-07-20
CR-1101219 Implementation between Edge/DC shouldn't be different for EXEC_WRITE opcode

mamin506:
ERT_EXEC_WRITE command required skip the first 6 words in the payload for DC platforms but there is no need to skip any words for EDGE platforms.
This pull request add a new execbuf opcode, which unify this behavior and NO need to skip any words on DC and edge platforms.

After this change, anyone that use ERT_EXEC_WRITE on DC platform would get exec buf error with dmesg said "xocl_command_ioctl: ERT_EXEC_WRITE is obsoleted, use ERT_START_KEY_VAL".
On edge platform, user would expect to see a warning with same content.

On DC, this change helps the ERT performance since we decreased the write to the ERT CQ.

Shell: xilinx_u200_gen3x16_xdma_base_1
Before (skip 6 words):
 xcl_iops_test.exe /opt/xilinx/firmware/u200/gen3x16-xdma/base/test/ -d 0000:05:00.1
Overall Commands:  100000 **IOPS: 357096** (verify)
TEST PASSED

After (no skip):
xcl_iops_test.exe /opt/xilinx/firmware/u200/gen3x16-xdma/base/test/ -d 0000:05:00.1
Overall Commands:  100000 **IOPS: 457794** (verify)
TEST PASSED

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_kds.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_kds.c
@@ -414,13 +414,15 @@ int zocl_command_ioctl(struct drm_zocl_dev *zdev, void *data,
 		start_krnl_ecmd2xcmd(to_start_krnl_pkg(ecmd), xcmd);
 		break;
 	case ERT_EXEC_WRITE:
-		/* third argument in following function is representing number of
-		 * words to skip when configuring CU. This should be consistent
-		 * for both edge/DC, but due to performance and siome use cases,
-		 * this has veen decided that , DC flows skip 6 words whereas
-		 * edge flows doesnt skip any words.
-		 */
-		exec_write_ecmd2xcmd(to_start_krnl_pkg(ecmd), xcmd, 0);
+		DRM_WARN("ERT_EXEC_WRITE is obsoleted, use ERT_START_KEY_VAL\n");
+#if KERNEL_VERSION(5, 4, 0) > LINUX_VERSION_CODE
+		__attribute__ ((fallthrough));
+#else
+		__attribute__ ((__fallthrough__));
+#endif
+		/* pass through */
+	case ERT_START_KEY_VAL:
+		start_krnl_kv_ecmd2xcmd(to_start_krnl_pkg(ecmd), xcmd);
 		break;
 	case ERT_START_FA:
 		start_fa_ecmd2xcmd(to_start_krnl_pkg(ecmd), xcmd);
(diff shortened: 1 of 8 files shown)


mamin506:
@stsoe Please let me know if you want me to update src/runtime_src/xrt/xrt++ as well.

maxzhen (approved):

src/runtime_src/core/edge/drm/zocl/zocl_kds.c
```diff
@@ -414,13 +414,15 @@ int zocl_command_ioctl(struct drm_zocl_dev *zdev, void *data,
...
 		break;
 	case ERT_EXEC_WRITE:
-		/* third argument in following function is representing number of
-		 * words to skip when configuring CU. This should be consistent
-		 * for both edge/DC, but due to performance and siome use cases,
-		 * this has veen decided that , DC flows skip 6 words whereas
-		 * edge flows doesnt skip any words.
-		 */
-		exec_write_ecmd2xcmd(to_start_krnl_pkg(ecmd), xcmd, 0);
+		DRM_WARN("ERT_EXEC_WRITE is obsoleted, use ERT_START_KEY_VAL\n");
```
stsoe, line 417: I am a little concerned about this breaking existing applications.  Do you think maybe we can shift the payload 6 words up (at performance impact of course, with the warning you have), then call start_krnl_kv_ecmd2xcmd?   The good news is that API level usage will automatically use ERT_START_KEY_VAL, it is only if someone out there is explicitly using ERT_EXEC_WRITE on an explicitly allocated exec buffer.

mamin506: On EDGE, we educated everyone that explicitly using ERT_EXEC_WRITE to not add that 6 words.
So, for zocl_kds.c, ERT_EXEC_WRITE is equal to ERT_START_KEY_VAL.

mamin506: For DC platforms, do you suggest to move the payload 6 words up?

stsoe: Sorry, I didn't see that I commented on the zocl version.  Yeah, I was suggesting moving xocl 6 words up.

mamin506: @stsoe update xrt++ and test with modified tests/unit_test/experimental/exec_write test case.
For xocl, if user still use ERT_EXEC_WRITE command, driver would convert it into ERT_START_KEY_VAL without considering performance impact.

I would add this change to 2021.2 release CHANGELOG after it is merged.

chvamshi-xilinx (approved):

chienwei-lan (approved):

stsoe (approved):

PR #5512, opened 2021-07-21
CR-1105543 kernel name change

sarab96:
xclbin::get_cus is expecting kernel name is specific format which is different from the name in ip_layout
removed redundant code
kernel channels not in use in XMA


dayeh-xilinx:
retest this please.  build failed due to an infra issue.

PR #5513, opened 2021-07-21
CR-1102740 mailbox delays writing to hw when kernel is in restart mode

stsoe:
Update example per CR notes.  For the stream datamover kernel provide
two scalar adder inputs and update each adders in lock step by
incrementing one and decrementing the other.  The streaming auto
restarted datamover kernel now picks up both scalar adders prior to
asserting asserting AP_START for next iteration.

The example is verified to work in HW and HW_EMU for
xilinx_u200_xdma_201830_2 XSA.


dayeh-xilinx:
retest this please.  build failed due to an infra issue.

PR #5520, opened 2021-07-22
high cpu usage while mailbox is idle

maxzhen:
mailbox polling thread is polling way too frequently due to that we passed in ms instead of us to the usleep_range(). this will result in high cpu usage (~3%) even when mailbox is idle. After the fix, it drops to almost 0.
this PR also fixes a couple of driver compilation warning.


stsoe (approved):
Thanks Max.

xdavidz (approved):

houlz0507 (approved):

rajkumar-xilinx (approved):

keryell:
Good! I noticed the problem recently...

PR #5522, opened 2021-07-22
VITIS-1586 xbtestv6: XRT alternative to DUMP for Performance and Reliability 2021.1

AShivangi:
Added queries related to:
- platform
- pcie-info
- xrt-info
- memory
- xclbin uuid

Refactored:
- power_rails and power_consumption -> electrical
- fans -> mechanical
- thermals -> thermal


src/runtime_src/core/common/info_platform.cpp
```diff
@@ -0,0 +1,237 @@
...
+    static_region.add("jtag_idcode", "N/A");
+  }
+
+  try {
+    static_region.add("fpga_name", xrt_core::device_query<xrt_core::query::rom_fpga_name>(device));
+  }
+  catch (const xrt_core::query::no_such_key&) {
+    // Ignoring if not available: Edge Case 
+    static_region.add("fpga_name", "N/A");
+  }
```
stsoe, line 130: This is the same pattern used.  I think this could be namespace { .. helper function .. }, that would be one way of a little bit of refactoring.

Same helper function can be used below also, I get very dizzy looking at all these try/catch instances.

src/runtime_src/core/common/info_platform.cpp
```diff
@@ -0,0 +1,237 @@
...
+  pt_platform.put_child("controller", controller);
+
+  std::vector<char> raw; 
+  try { 
+    raw = xrt_core::device_query<xrt_core::query::clock_freq_topology_raw>(device);
+  }
+  catch (const xrt_core::query::no_such_key&) {
+    // Ignoring if not available: Edge Case 
+  }
+  if(!raw.empty()) {
```
stsoe, line 193: Missing space

src/runtime_src/core/common/info_platform.cpp
```diff
@@ -0,0 +1,237 @@
...
+    } catch(...) {}
+    ptree.add("cpu_affinity", xrt_core::device_query<xrt_core::query::cpu_affinity>(device));
+    ptree.add("max_shared_host_mem_aperture_bytes", xrt_core::device_query<xrt_core::query::max_shared_host_mem_aperture_bytes>(device));
+    ptree.add("shared_host_mem_size_bytes", xrt_core::device_query<xrt_core::query::shared_host_mem>(device));
+  } catch(...) {}
+  
+  return ptree;
+}
+
+}} // platform, xrt
```
stsoe, line 237: missing new line

src/runtime_src/core/common/info_platform.h
```diff
@@ -0,0 +1,41 @@
...
+#define COMMON_INFO_PLATFORM_H
+
+// Local - Include Files
+#include "device.h"
+
+// ------ S T A T I C   F U N C T I O N S -------------------------------------
+namespace {
+
+} // unnamed namespace
+
```
stsoe, line 27: Remove

src/runtime_src/core/common/utils.cpp
```diff
@@ -217,4 +219,110 @@ issue_id()
...
+
+std::string 
+parse_clock_id(const std::string& id)
+{
+  auto clock_str = clock_map.find(id);
+  return clock_str != clock_map.end() ? clock_str->second : "N/A";
+}
+
+std::string
+string_to_UUID(std::string str)
```
stsoe, line 306: const std::string&
Is it not possible to reuse the code in xrt::uuid?

src/runtime_src/core/common/utils.h
```diff
@@ -97,6 +106,27 @@ XRT_CORE_COMMON_EXPORT
 uint64_t
 issue_id();
 
+XRT_CORE_COMMON_EXPORT
+int
+parse_p2p_config(const std::vector<std::string> config, std::string &msg);
```
stsoe, line 111: Really, std::vector by value?  And should return msg, not error code.

src/runtime_src/core/common/utils.h
```diff
@@ -97,6 +106,27 @@ XRT_CORE_COMMON_EXPORT
...
+parse_clock_id(const std::string& id);
+
+/**
+ * string_to_UUID(): convert a string to hyphen formatted UUID
+ * 
+ * Returns: 00000000-0000-0000-0000-000000000000 formatted uuid
+ */
+XRT_CORE_COMMON_EXPORT
+std::string
+string_to_UUID(std::string str);
```
stsoe, line 128: By value?

src/runtime_src/core/include/xrt/xrt_device.h
```diff
@@ -69,14 +69,22 @@ namespace info {
...
+ * @var memory
+ *  Memory information present on the device (std::string)
+ * @var platform
+ *  Platforms flashed on the device (std::string)
+ * @var pcie_info
+ *  Pcie information of the device (std::string)
+ * @var host
+ *  Host information (std::string)
+ * @var dynamic_regions
+ *  Information about xclbin on the device (std::string)
```
stsoe, line 87: I believe these std::strings  are in fact json formatted, right?  If so, maybe state that in the doxygen comments.

stsoe (changes requested):
Nice refactoring and sharing of code between xbutil and core/common.
I like the APIs entry points to all this refactored code (the header files) being very simple.

But let's raise the bar in terms of code structure.  Use very short functions preferable 10 lines or less, at most 20, so in other words, re-factor a lot of this code into smaller functions.  The point is that it is very difficult to understand what a long function does, where as very short concise functions are generally self-explanatory.

There is a lot of try/catch that makes the code difficult to follow.  I think some of those can be factored into maybe a templated (or not templated) helper function leaving the calling code clean and easy to read.

A lot of comments, but they are simple and very similar, so after a few you get the pattern.

src/runtime_src/core/common/info_platform.cpp
```diff
@@ -0,0 +1,237 @@
...
+  static_region.add("vbnv", xrt_core::device_query<xrt_core::query::rom_vbnv>(device));
+
+  std::vector<std::string> interface_uuids;
+  // the vectors are being populated by empty strings which need to be removed
+  try {
+    interface_uuids = xrt_core::device_query<xrt_core::query::interface_uuids>(device);
+    interface_uuids.erase(
+      std::remove_if(interface_uuids.begin(), interface_uuids.end(),	
+                  [](const std::string& s) { return s.empty(); }), interface_uuids.end());
+  } catch (const xrt_core::query::no_such_key&) {}
```
keryell, line 113: When does this happen?

src/runtime_src/core/common/utils.cpp
```diff
@@ -217,4 +219,110 @@ issue_id()
...
+  {0xbd80, "Inspur"},
+  {0x12eb, "Amazon"},
+  {0x2b79, "Google"}
+};
+
+std::string 
+parse_oem_id(const std::string& oemid)
+{
+  unsigned int oem_id_val = 0;
+  std::stringstream ss;
```
keryell, line 279: Cf comment on `std::stoull`

src/runtime_src/core/common/utils.cpp
```diff
@@ -217,4 +219,110 @@ issue_id()
   return id++;
 }
 
+int
+parse_p2p_config(const std::vector<std::string> config, std::string &msg)
```
keryell, line 223:
```suggestion
parse_p2p_config(const std::vector<std::string>& config, std::string& msg)
```
Perhaps you could also return a `pair` of `int` and `std::string` instead?

src/runtime_src/core/common/utils.cpp
keryell: The problem here is that the `str` is mutable and move to uppercase.
I have the feeling the logic is overly complicated.
Why not just starting by upper-casing the `str` and then skip the token upper-case?

src/runtime_src/core/common/utils.cpp
keryell: Add some comments on the declaration of the function too?

src/runtime_src/core/common/utils.cpp
```diff
@@ -217,4 +219,110 @@ issue_id()
...
+  {0x10da, "Xilinx"},
+  {0x02a2, "Dell"},
+  {0x12a1, "IBM"},
+  {0xb85c, "HP"},
+  {0x2a7c, "Super Micro"},
+  {0x4a66, "Lenovo"},
+  {0xbd80, "Inspur"},
+  {0x12eb, "Amazon"},
+  {0x2b79, "Google"}
+};
```
keryell, line 273: There is no existing library with a full list?

rozumx:
Removed my name as a reviewer.   Lots of great suggestions and recommendations from both Soren and Ronan.  It is evident that there will be more changes and updates to this PR, which I will be more than happy to review.

dayeh-xilinx:
retest this please

dayeh-xilinx:
retest this please

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -1003,7 +1004,13 @@ p2pTest(const std::shared_ptr<xrt_core::device>& _dev, boost::property_tree::ptr
 
   std::string msg;
   XBU::xclbin_lock xclbin_lock(_dev);
-  XBU::check_p2p_config(_dev.get(), msg);
+  std::vector<std::string> config;
+  try {
+    config = xrt_core::device_query<xrt_core::query::p2p_config>(_dev);
+  }
+  catch (const xrt_core::query::no_such_key&) {  }
```
stsoe, line 1011: As best I can tell based on pipeline logs is that xbutil validate fails on u50 because it calls device_query   on line 1009, the key exists, but the sysfs node cannot be opened.  The latter results in std::runtime_error, which is not caught and causes xbutil to exit with the exception message:
`[xbutil] ERROR: Failed to open /sys/bus/pci/devices/0000:3b:00.1/p2p.u.11534336/config for reading: No such file or directory`

I am not liking that the sysfs getters throws generic std::runtime_error, I think we should have a specific exception for sysfs error, maybe` xrt_core::query::sysfs_error`, similar to `xrt_core::query::no_such_key`.  If we derive both off of a `xrt_core::query::error `exception then most code that catches `...`, or `std::runtime_error` can be changed to catch xrt_core::query::error instead.

stsoe: @AShivangi Please check #5531 for exception hierarchy for query errors.  If you catch xrt_core::query::exception& then you catch errors related to query requests (no_such_key, sysfs_error, a few more), but you will not catch unrelated "real" errors.

src/runtime_src/core/common/info_memory.cpp
```diff
@@ -0,0 +1,260 @@
...
+    ptMem.put("type", str);
+    ptMem.put("tag", map->m_mem_data[i].m_tag);
+    ptMem.put("enabled", map->m_mem_data[i].m_used ? true : false);
+    ptMem.put("base_address", boost::format("0x%x") % map->m_mem_data[i].m_base_address);
+    ptMem.put("range_bytes", boost::format("0x%x") % (map->m_mem_data[i].m_size << 10));
+    if (!temp_buf.empty() && temp[i] != invalid_sensor_value)
+      ptMem.put("extended_info.temperature_C", temp[i]);
+    ptMem.put("extended_info.usage.allocated_bytes", memoryUsage);
+    ptMem.put("extended_info.usage.buffer_objects_count", boCount);
+    ptMem_array.push_back(std::make_pair("",ptMem));
```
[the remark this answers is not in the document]
stsoe: emplace not supported for boost::property_tree

PR #5523, opened 2021-07-22
KDS2 CU stats

mamin506:
1. add per CU sysfs node for statistics
2. add per cu usage counter

### DIFF
--- a/src/runtime_src/core/common/drv/include/kds_stat.h
+++ b/src/runtime_src/core/common/drv/include/kds_stat.h
@@ -38,10 +38,6 @@ struct cu_stats {
 	u64		  usage[MAX_CUS];
 };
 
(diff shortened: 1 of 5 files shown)


maxzhen (commented):
do you really need per-cpu data structure here? update is only done by cu thread, read is very infrequently, right?

mamin506:
> do you really need per-cpu data structure here? update is only done by cu thread, read is very infrequently, right?

The usage variable is a u64 type, I afraid that only read the lower 32bits but the higher 32bits haven't been updated. For in flight commands, I don't have this concerns.

mamin506:
@maxzhen you are right. There is no need to use per-cpu variable on a 64 bits system in this case.

PR #4342, opened 2020-10-29
CR-1073067: Constructing CU index Vs base address map using IP_LAYOUT in hw_emu

vboggara-xilinx:
Issue: xclRegRead/Write() using CU index Vs base address map, which is constructed by EMBEDDED_METADATA in xclbin, which is wrong.
Fix: Constructing CU index Vs base address map using IP_LAYOUT in xclbin instead of using EMBEDDED_METADATA.
Test: Ran some test cases manually and also ran canary, results are clean.
Risk: Low.

PR #5525, opened 2021-07-23
Fix CU indexing in HW emulation

stsoe:
Make sure HW emulation is using the same indexing order as user space
XRT.  Without this fix, the index order used by xclRegRW was using
IP_LAYOUT order as opposed to interrupt index ordering used by rest of
XRT.

Fix is to use already computed compute unit indeces from core device.

Eliminate the use of legacy boost::lexical_cast.

See also #4342


chvamshi-xilinx (approved):

stsoe:
retest this please

stsoe:
retest this please

PR #5528, opened 2021-07-23
Default scalar adders to some non-zero value

stsoe:
This means that valid run of incr reflects non default values thus
guarantees incr is indeed updated via mailbox.


stsoe:
retest this please

PR #5531, opened 2021-07-25
Add query request exception hierarchy for more granularity

stsoe:
Provides granularity for calling code to catch errors specific to
query request which are often acceptable errors because some
devices or OSes may not support all types of query requests.

Other non query exceptions signal a different kind of error which
should maybe not be caught.

The addition of the query request exception hierarchy does not
break existing code that catches std::exception (or all errors)
because ultimately the base query exception is-a std::exception

### DIFF
--- a/src/runtime_src/core/common/error.h
+++ b/src/runtime_src/core/common/error.h
@@ -99,7 +99,7 @@ public:
   {}
 
   explicit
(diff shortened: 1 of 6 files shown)


AShivangi (approved):
Thanks for introducing this additional layer of exception hadling. This will be helpful :)

PR #5532, opened 2021-07-26
Add support for qspi_versal flash type

rbramand-xilinx:
> Added qspi flash support for versal devices, it uses ospi versal concept
> User can give flash_type "qspi_versal" in xbmgmt flash command

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/devices.h
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/devices.h
@@ -77,7 +77,7 @@ enum {
 #define	FLASH_TYPE_QSPIPS_X2_SINGLE	"qspi_ps_x2_single"
 #define	FLASH_TYPE_QSPIPS_X4_SINGLE	"qspi_ps_x4_single"
 #define	FLASH_TYPE_OSPI_VERSAL	"ospi_versal"
-#define	FLASH_TYPE_OSPI_VERSAL	"qspi_versal"
+#define	FLASH_TYPE_QSPI_VERSAL	"qspi_versal"
 
 #define XOCL_SUBDEV_MAX_RES		32
 #define XOCL_SUBDEV_RES_NAME_LEN	64
--- a/src/runtime_src/core/pcie/driver/linux/xocl/devices.h
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/devices.h
@@ -77,6 +77,7 @@ enum {
 #define	FLASH_TYPE_QSPIPS_X2_SINGLE	"qspi_ps_x2_single"
 #define	FLASH_TYPE_QSPIPS_X4_SINGLE	"qspi_ps_x4_single"
 #define	FLASH_TYPE_OSPI_VERSAL	"ospi_versal"
+#define	FLASH_TYPE_OSPI_VERSAL	"qspi_versal"
 
 #define XOCL_SUBDEV_MAX_RES		32
 #define XOCL_SUBDEV_RES_NAME_LEN	64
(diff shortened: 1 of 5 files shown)


chvamshi-xilinx:
@rbramand-xilinx , You need to make changes to new tool (xbmgmt2) as well.

rbramand-xilinx:
> @rbramand-xilinx , You need to make changes to new tool (xbmgmt2) as well.

@chvamshi-xilinx, I have added changes to new tools as well. Please review

chvamshi-xilinx (approved):

maxzhen (changes requested):
Do we still need to support OSPI versal type? I see that we removed it from driver but kept in tools, so wondering why?

chvamshi-xilinx:
> Do we still need to support OSPI versal type? I see that we removed it from driver but kept in tools, so wondering why?

vck5000 uses OSPI type. vck190 uses QSPI. Underlying code is same for both types.

maxzhen:
> 
> 
> > Do we still need to support OSPI versal type? I see that we removed it from driver but kept in tools, so wondering why?
> 
> vck5000 uses OSPI type. vck190 uses QSPI. Underlying code is same for both types.

but we have removed OSPI versal type from driver? is this ok?

chvamshi-xilinx:
> > > Do we still need to support OSPI versal type? I see that we removed it from driver but kept in tools, so wondering why?
> > 
> > 
> > vck5000 uses OSPI type. vck190 uses QSPI. Underlying code is same for both types.
> 
> but we have removed OSPI versal type from driver? is this ok?

No Max. Its not removed. OSPI is replaced with QSPI only for VCK190 device structure.

AShivangi:
Oh I was confused about the same. In that case, can we add a comment in flasher.cpp explaining this

chvamshi-xilinx:
> Oh I was confused about the same. In that case, can we add a comment in flasher.cpp explaining this

Sure Shivangi. Are you suggesting to add a comment something like "Underlying code to support qspi & ospi types is same.  Two different CLI's are provided for better user experience. VCK190 has qspi flash and VCK5000 has ospi flash"
or anything more?

AShivangi:
> > Oh I was confused about the same. In that case, can we add a comment in flasher.cpp explaining this
> 
> Sure Shivangi. Are you suggesting to add a comment something like "Underlying code to support qspi & ospi types is same. Two different CLI's are provided for better user experience. VCK190 has qspi flash and VCK5000 has ospi flash"
> or anything more?

This looks pretty good! Thanks

maxzhen:
> 
> 
> > > > Do we still need to support OSPI versal type? I see that we removed it from driver but kept in tools, so wondering why?
> > > 
> > > 
> > > vck5000 uses OSPI type. vck190 uses QSPI. Underlying code is same for both types.
> > 
> > 
> > but we have removed OSPI versal type from driver? is this ok?
> 
> No Max. Its not removed. OSPI is replaced with QSPI only for VCK190 device structure.

OK, I see. Thanks for clarifying.

maxzhen (approved):

AShivangi:
Approved. Please add sample output for the new xbmgmt flash before merging.

AShivangi (approved):

maxzhen:
@rbramand-xilinx please still provide what @AShivangi is asking for. I missed @AShivangi 's ask before merge.

rbramand-xilinx:
Hi all,

The following is the sample output that is displayed when flashing using new tool xbmgmt2 :

**xbmgmt program -d 0000:18:00.0 --base --image /lib/firmware/xilinx/fedcba987654321fedcba987654321fe/BOOT_FLASH.BIN --flash-type qspi_versal**   
                                                                  
XRT build version: 2.12.0
Build hash: dd2bbb635ea2e7a251d594a7592f9d6ed5174231
Build date: 2021-07-28 04:08:53
Git branch: master
PID: 97762
UID: 0
[Tue Jul 27 22:41:50 2021 GMT]
HOST: xhdamitkasat-lab.xilinx.com
[XRT] WARNING: Overriding flash mode is not recommended.
You may damage your device with this option.
INFO: ***PDI has 133696048 bytes


-> Also if we don't explicitly specify flash-type also it will pick up the type as qspi_versal as a sysfs node flash_type is already being generated in driver.

PR #5536, opened 2021-07-27
Error get last for xcol driver errors

sarab96:
Error get last for xocl driver
query & xocl_error class for pcie x86 side

### DIFF
--- a/src/runtime_src/core/common/api/xrt_error.cpp
+++ b/src/runtime_src/core/common/api/xrt_error.cpp
@@ -24,6 +24,7 @@
 #include "native_profile.h"
 #include "device_int.h"
 #include "error_int.h"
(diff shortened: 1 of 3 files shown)


larry9523 (commented):
I am not very familiar with this part of code. So forgive me if it is a stupid question. I am wondering why do we need a special query xocl_error? Isn't it be handled automatically among xocl, xclmgmt, zocl? cc @stsoe

stsoe (changes requested):
This pull request needs a lot more details, please.
It's void of comments as to the purpose of added classes and what the changes do.
Please redo.  And please add @stsoe for any user space core/common changes.

sarab96:
Will submit new PR

PR #5541, opened 2021-07-28
CR-1105038: Generate stream tables in profile summary when ASMs are configured without trace

jvillarre:
This pull request fixes an issue where AIMs and ASMs had their slot IDs only tracked if they were configured with trace, leading to the summary file not finding counter values for ASMs that only had counters in mixed designs.  Profiling now stores the slot IDs of all AIMs and ASMs and separately tracks the slot IDs of AIMs and ASMs that have trace enabled.  This allows the summary generation to print tables correctly and the trace generation is unaffected.

This pull request also does some slight refactoring in the setup of the database information.


pgschuey (approved):

IshitaGhosh (approved):
The changes look good.

PR #5542, opened 2021-07-28
Fix access of kernel xml attributes for restart and sw reset

stsoe:
The xml meta data is using camelCase so check for countedAutoRestart
and swReset.

The xrt.ini hooks are still in place so that xclbins without the
meta-data can continue to be used.


dayeh-xilinx:
retest this please

PR #5547, opened 2021-07-29
CR-1106238: Domain is not mandatory in xbutil/xbmgmt if it is 0000

chvamshi-xilinx:
found a bug while verifying one of the feature. Looks like this bug got recently introduced where xbutil/xbmgmt is always expecting domain value. This is not the case earlier.
Removing that hard check when domain is "0000"

### DIFF
--- a/src/runtime_src/core/tools/common/XBUtilities.cpp
+++ b/src/runtime_src/core/tools/common/XBUtilities.cpp
@@ -425,7 +425,10 @@ bdf2index(const std::string& bdfstr, bool _inUserDomain)
     dev = static_cast<uint16_t>(std::stoi(std::string(tokens[0]), nullptr, radix));
   }
   bus = static_cast<uint16_t>(std::stoi(std::string(tokens[1]), nullptr, radix));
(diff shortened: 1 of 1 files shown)


AShivangi (approved):
Thanks for fixing this. There is a CR related to this which I have reassigned to you

chvamshi-xilinx:
> Thanks for fixing this. There is a CR related to this which I have reassigned to you

Good that this is already being tracked with a CR. I found in one my testing.  I will mark this CR fixed after merging. We may need to backport this to 2021.1 also.

chvamshi-xilinx:
> > Thanks for fixing this. There is a CR related to this which I have reassigned to you
> 
> Good that this is already being tracked with a CR. I found in one my testing. I will mark this CR fixed after merging. We may need to backport this to 2021.1 also.
Looks like it is DSV_ESC CR. I have to mark it ASAP :-)

chvamshi-xilinx:
Updated description with CR,

rajkumar-xilinx (approved):

PR #5548, opened 2021-07-29
ERT USER refactoring

chienwei-lan:


mamin506 (approved):
A big change. :) Overall this looks good to me.

maxzhen (commented):
1. cfg_gpio.c is for MB ERT only. the file name is too generic. please consider a better name.
2. command_queue.c is for ERT only. please also consider a better name, say, ert_cq.c.
3. do we think we can move versal mailbox cq to it's own cq subdev?

PR #5549, opened 2021-07-30
CR-1106238: Domain is not mandatory if it is 0000 (#5547)

chvamshi-xilinx:
found a bug while verifying one of the feature. Looks like this bug got recently introduced where xbutil/xbmgmt is always expecting domain value. This is not the case earlier.
Removing that hard check when domain is "0000"


chvamshi-xilinx:
Critical for customers.

PR #5551, opened 2021-07-30
added new validate_zcu design which validate zynqmp platforms

vishnuchebrolu:


vishnuchebrolu:
Edge Validate is not required in XRT git.

PR #5554, opened 2021-07-30
Coverity Fix: Remove dead code in profiling

jvillarre:
Comparing if an unsigned int is less than zero will always be false, so this check is being removed


IshitaGhosh (approved):
Change looks good.

PR #5555, opened 2021-07-30
CR-1100587: Align device profiling events on Edge 

jvillarre:
For xclbins that do not have a clock_freq_topology section, we were assuming a device frequency of 300 MHz.  On Edge devices, this was not always accurate leading to a misalignment between host profiling events and device profiling events.

This pull request adds an additional check on the embedded_metadata section and uses it to find device clock speed.


IshitaGhosh (approved):
Changes look good

PR #5556, opened 2021-07-30
XRT-910 Shim changes to parse the xclbin in usercode to get kernel information

arpitxilinx:
This PR -
1. Adds the parsing of xclbin to get the kernel information and passing the information to windows kernel driver
2. It fixes the download of xclbin on Windows platform.


src/runtime_src/core/pcie/windows/shim.cpp
```diff
@@ -550,49 +552,125 @@ struct shim
...
+    //axlf_obj.kds_cfg.slot_size = mCoreDevice->get_ert_slots().second;
+    auto xml_hdr = xrt_core::xclbin::get_axlf_section(top, EMBEDDED_METADATA);
+    if (!xml_hdr)
+        throw std::runtime_error("No xml metadata in xclbin");
+    auto xml_size = xml_hdr->m_sectionSize;
+    auto xml_data = reinterpret_cast<const char*>(reinterpret_cast<const char*>(top) + xml_hdr->m_sectionOffset);
+    axlf_obj.kds_cfg.slot_size = (uint32_t)m_core_device->get_ert_slots(xml_data, xml_size).second;
+
+    /* calculate the complete length of buffer to be sent as output buffer. */
+    len = off + sizeof(struct xocl_kds);
```
stsoe, line 657: Is this copied verbatim from linux/shim.cpp?   If so, we should re-factor.  On the other hand if the drivers evolve independently maybe there is an advantage in keeping the copied code?

stsoe (approved):

src/runtime_src/core/pcie/windows/shim.cpp
arpitxilinx: Yes, it is for most part verbatim, with some changes specific to windows.

rajkumar-xilinx (approved):

PR #5558, opened 2021-07-30
Fix for CR-1105261 : Fix addition of invalid memory entry in memoryInfo map in XDP Static Info DB

IshitaGhosh:
When design xclbin does not have any IP_LAYOUT, XDP Summary Writer hits a crash while writing summary.csv due to accessing an invalid memory info entry. This crash is fixed by stopping addition of invalid entry in memoryInfo map in XDP Static Info DB


jvillarre (approved):
Additional check looks good and prevents a creation of a map entry when there shouldn't be one.

PR #5562, opened 2021-08-02
Using same driver for hw & hw_Emu on edge. Removing unused functions also

chvamshi-xilinx:
This PR is cleaning up driver codebase for hw and hw_emu for edge. Code is getting diverged if we use differrent files.
Using same code reduces the maintaining burden.


dayeh-xilinx:
retest this please

chvamshi-xilinx:
retest this please

PR #5563, opened 2021-08-02
add binary sysfs node to show ERT CQ for easy of debug

chienwei-lan:


dayeh-xilinx:
retest this please

src/runtime_src/core/pcie/driver/linux/xocl/subdev/command_queue.c
```diff
@@ -545,6 +595,12 @@ static int command_queue_probe(struct platform_device *pdev)
 		}
 	}
 
+	err = sysfs_create_group(&pdev->dev.kobj, &cmd_queue_attr_group);
+	if (err) {
+		xocl_err(&pdev->dev, "create ert_cq sysfs attrs failed: %d", err);
```
mamin506, line 600: later on, we might have more and more sysfs node for other purpose. Should we print "Create command queue sysfs node group failed" here?

mamin506 (changes requested):
Thanks Chien-wei. This is a good to have for debug hang issue.
Just a minor change.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/command_queue.c
mamin506: ert_cq will be the sub-dev name. Not just the argument.

mamin506 (approved):

PR #5566, opened 2021-08-03
Add sanity checks for valid access to AIM map, NOC list, GMIOlist in XDP Static DB

IshitaGhosh:


jvillarre (approved):
Additional checks look good

PR #5567, opened 2021-08-04
CR-1101266 xbmgmt dump help does not make it clear it needs a .ini ou…

sadasivanshaiju:
…tput extension


AShivangi (approved):
Looks good! Thanks

PR #5571, opened 2021-08-04
Add changes for parsing pci bar mappings in xsabin(firmware)

rbramand-xilinx:
> Added changes for parsing pci bar mappings using fdt parser
> Subdev resources with 64 bit offset is converted to 32 bit based on parsed bar mappings as pci expects offset from bar not complete address
> Experiments done on vck190_pcie flat shell by running pl vadd test case and things are working as expected

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/xocl_fdt.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/xocl_fdt.c
@@ -910,7 +910,8 @@ static bool get_userpf_info(void *fdt, int node, u32 pf)
 	offset = fdt_parent_offset(fdt, node);
 	val = fdt_get_name(fdt, offset, NULL);
 
-	if (!val || strncmp(val, NODE_ENDPOINTS, strlen(NODE_PROPERTIES)))
+	if (!val || (strncmp(val, NODE_ENDPOINTS, strlen(NODE_PROPERTIES))
+			&& strncmp(val, NODE_BARS, strlen(NODE_BARS))))
 		return true;
 
 	do {
@@ -1467,8 +1468,7 @@ static int xocl_fdt_get_pci_addr(xdev_handle_t xdev_hdl)
         int offset;
         const u32* bar;
         const u32* pfn;
-        const u64* base_addr;
-        const u64* range;
+        const u64* io_off;
         u32 pf, bar_idx;
         int ret = 0, count = 0;
 
@@ -1483,7 +1483,7 @@ static int xocl_fdt_get_pci_addr(xdev_handle_t xdev_hdl)
                 return -EINVAL;
         }
 
-        offset = fdt_path_offset(core->fdt_blob, "/pcie/bars");
+        offset = fdt_path_offset(core->fdt_blob, "/" NODE_PCIE "/" NODE_BARS);
         if (offset < 0) {
                 xocl_xdev_dbg(xdev_hdl, "pcie bars nodes are not present in firmware data");
                 return -EINVAL;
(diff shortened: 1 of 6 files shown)


chvamshi-xilinx (approved):

src/runtime_src/core/pcie/driver/linux/xocl/xocl_subdev.c
```diff
@@ -462,6 +463,18 @@ static int __xocl_subdev_construct(xdev_handle_t xdev_hdl,
...
+			 * to pass only offset.
+			 * Get the offset by comparing with bar mappings of CPM.
+			 */
+			if(core->bars) {
+				bar_start = core->bars[bar_idx].base_addr;
+				if ((bar_start >> 32) == (res[i].start >> 32))
+					res[i].start ^= bar_start;
+				if ((bar_start >> 32) == (res[i].end >> 32))
+					res[i].end ^= bar_start;
+			}
```
houlz0507, line 477: I am confused. 
(1) Is low 32-bit of bar_start always 0? Should it be res[i].start -= bar_start; 
(2) Why do we have to check if ((bar_start >> 32) == (res[i].start >> 32))? I think as long as BAR metadata shows up, we should do adjust the endpoint offset?

src/runtime_src/core/pcie/driver/linux/xocl/xocl_fdt.c
```diff
@@ -1461,6 +1461,87 @@ static void xocl_pack_subdev(xdev_handle_t xdev_hdl, struct xocl_subdev *subdev)
...
+                ret = -ENOMEM;
+                goto done;
+        }
+        memset(core->bars, 0, sizeof(struct pci_bars) * NUM_PCI_BARS);
+
+        for (offset = fdt_first_subnode(core->fdt_blob, offset);
+                offset >= 0;
+                offset = fdt_next_subnode(core->fdt_blob, offset)) {
+
+                pfn = fdt_getprop(core->fdt_blob, offset,"physical_function", NULL);
```
houlz0507, line 1503: please define a macro for "physical_function" in xocl_fdt.h. Actually, could we consider to change the xclbinutil tool to generate unified keywords which is using by endpoints. E.g.

#define PROP_PF_NUM "pcie_physical_function"
#define PROP_BAR_IDX "pcie_bar_mapping"
#define PROP_IO_OFFSET "reg"

houlz0507 (commented):
Please add code to get_userpf_info() in xocl_fdt.c to filter out the PF0 metadata from the userpf fdt blob

src/runtime_src/core/pcie/driver/linux/xocl/xocl_subdev.c
chvamshi-xilinx: As discussed offline, we will simplify this code for better understanding.

src/runtime_src/core/pcie/driver/linux/xocl/xocl_fdt.c
chvamshi-xilinx: Discussing with @rozumx .

src/runtime_src/core/pcie/driver/linux/xocl/xocl_subdev.c
rbramand-xilinx: @houlz0507 simplified the code and also added code to filter out PF0 info from userpf fdt blob, please review

houlz0507 (approved):

maxzhen (approved):

PR #5574, opened 2021-08-04
Error get last for xocl driver

sarab96:
Adding error get last support for xocl driver side.
errors to be captured in sysfs node.
Format of sysfs node to be changed to binary for both xocl & zocl

### DIFF
--- a/src/runtime_src/core/common/api/xrt_error.cpp
+++ b/src/runtime_src/core/common/api/xrt_error.cpp
@@ -208,10 +208,10 @@ public:
   {
     //Code for new format; Binary array of error structs in sysfs
     try {
-      std::vector<char> buf = xrt_core::device_query<xrt_core::query::xocl_errors>(device);
+      auto buf = xrt_core::device_query<xrt_core::query::xocl_errors>(device);
       if (buf.empty())
         return;
-      const xcl_errors *errors_buf = reinterpret_cast<xcl_errors *>(buf.data());
+      auto errors_buf = reinterpret_cast<xcl_errors *>(buf.data());
       if (errors_buf->num_err <= 0)
         return;
       if (errors_buf->num_err > XCL_ERROR_CAPACITY)
--- a/src/runtime_src/core/common/api/xrt_error.cpp
+++ b/src/runtime_src/core/common/api/xrt_error.cpp
@@ -24,6 +24,7 @@
 #include "native_profile.h"
 #include "device_int.h"
 #include "error_int.h"
+#include "core/include/xclerr_int.h"
 #include "core/common/error.h"
 #include "core/common/system.h"
 #include "core/common/device.h"
@@ -205,6 +206,27 @@ class error_impl
 public:
   error_impl(const xrt_core::device* device, xrtErrorClass ecl)
   {
+    //Code for new format; Binary array of error structs in sysfs
+    try {
+      std::vector<char> buf = xrt_core::device_query<xrt_core::query::xocl_errors>(device);
+      if (buf.empty())
+        return;
+      const xcl_errors *errors_buf = reinterpret_cast<xcl_errors *>(buf.data());
+      if (errors_buf->num_err <= 0)
+        return;
+
+      for (int i = 0; i < errors_buf->num_err; i++) {
+        if (XRT_ERROR_CLASS(errors_buf->errors[i].err_code) != ecl)
+          continue;
+        m_errcode = XRT_ERROR_NUM(errors_buf->errors[i].err_code);
+        m_timestamp = errors_buf->errors[i].ts;
+      }
+      return;
+    } catch (const xrt_core::query::no_such_key&) {
+      // Ignoring for now. Check below for edge if not available
+      // query table of zocl doesn't have xocl_errors key
+    }
+    //Below code will be removed after zocl changes for new format
     auto errors = xrt_core::device_query<xrt_core::query::error>(device);
     for (auto& line : errors) {
       auto ect = xrt_core::query::error::to_value(line);
--- a/src/runtime_src/core/common/api/xrt_error.cpp
+++ b/src/runtime_src/core/common/api/xrt_error.cpp
@@ -214,11 +214,13 @@ public:
       const xcl_errors *errors_buf = reinterpret_cast<xcl_errors *>(buf.data());
       if (errors_buf->num_err <= 0)
         return;
+      if (errors_buf->num_err > XCL_ERROR_CAPACITY)
+        return;
 
       for (int i = 0; i < errors_buf->num_err; i++) {
         if (XRT_ERROR_CLASS(errors_buf->errors[i].err_code) != ecl)
           continue;
-        m_errcode = XRT_ERROR_NUM(errors_buf->errors[i].err_code);
+        m_errcode = errors_buf->errors[i].err_code;
         m_timestamp = errors_buf->errors[i].ts;
       }
       return;
--- a/src/runtime_src/core/common/api/xrt_error.cpp
+++ b/src/runtime_src/core/common/api/xrt_error.cpp
@@ -211,18 +211,8 @@ public:
       auto buf = xrt_core::device_query<xrt_core::query::xocl_errors>(device);
       if (buf.empty())
         return;
-      auto errors_buf = reinterpret_cast<xcl_errors *>(buf.data());
-      if (errors_buf->num_err <= 0)
-        return;
-      if (errors_buf->num_err > XCL_ERROR_CAPACITY)
-        return;
-
-      for (int i = 0; i < errors_buf->num_err; i++) {
-        if (XRT_ERROR_CLASS(errors_buf->errors[i].err_code) != ecl)
-          continue;
-        m_errcode = errors_buf->errors[i].err_code;
-        m_timestamp = errors_buf->errors[i].ts;
-      }
+      auto ect = xrt_core::query::xocl_errors::to_value(buf, ecl);
+      std::tie(m_errcode, m_timestamp) = ect;
       return;
     } catch (const xrt_core::query::no_such_key&) {
       // Ignoring for now. Check below for edge if not available
--- a/src/runtime_src/core/common/query_requests.cpp
+++ b/src/runtime_src/core/common/query_requests.cpp
@@ -4,6 +4,7 @@
  */
 #define XRT_CORE_COMMON_SOURCE // in same dll as core_common
 #include "query_requests.h"
+#include "core/include/xclerr_int.h"
 #include <map>
 #include <string>
 
@@ -120,6 +121,49 @@ parse(const std::string& clock)
 
   auto clock_str = clock_map.find(clock);
   return clock_str != clock_map.end() ? clock_str->second : "N/A";
+}
+
+std::pair<uint64_t, uint64_t>
+xrt_core::query::xocl_errors::
+to_value(const std::vector<char>& buf, xrtErrorClass ecl)
+{
+  uint64_t errcode = 0;
+  uint64_t ts = 0;
+  if (buf.empty())
+    return {0, 0};
+  auto errors_buf = reinterpret_cast<const xcl_errors *>(buf.data());
+  if (errors_buf->num_err <= 0)
+    return {0, 0};
+  if (errors_buf->num_err > XCL_ERROR_CAPACITY)
+    throw xrt_core::system_error(EINVAL, "Invalid data in sysfs");
+
+  for (int i = errors_buf->num_err; i >= 0; i--) {
+    if (XRT_ERROR_CLASS(errors_buf->errors[i].err_code) != ecl)
+      continue;
+    errcode = errors_buf->errors[i].err_code;
+    ts = errors_buf->errors[i].ts;
+    break;
+  }
+  return {errcode, ts};
+}
+
+std::vector<xclErrorLast>
+xrt_core::query::xocl_errors::
+to_errors(const std::vector<char>& buf)
+{
+  if (buf.empty())
+    return {};
+  auto errors_buf = reinterpret_cast<const xcl_errors *>(buf.data());
+  if (errors_buf->num_err <= 0)
+    return {};
+  if (errors_buf->num_err > XCL_ERROR_CAPACITY)
+    throw xrt_core::system_error(EINVAL, "Invalid data in sysfs");
+
+  std::vector<xclErrorLast> errors;
+  for (int i = 0; i < errors_buf->num_err; i++)
+    errors.emplace_back(errors_buf->errors[i]);
+
+  return errors;
 }
   
 }} // query, xrt_core
--- a/src/runtime_src/core/common/query_requests.cpp
+++ b/src/runtime_src/core/common/query_requests.cpp
@@ -127,24 +127,21 @@ std::pair<uint64_t, uint64_t>
 xrt_core::query::xocl_errors::
 to_value(const std::vector<char>& buf, xrtErrorClass ecl)
 {
-  uint64_t errcode = 0;
-  uint64_t ts = 0;
   if (buf.empty())
     return {0, 0};
+
   auto errors_buf = reinterpret_cast<const xcl_errors *>(buf.data());
   if (errors_buf->num_err <= 0)
     return {0, 0};
+
   if (errors_buf->num_err > XCL_ERROR_CAPACITY)
     throw xrt_core::system_error(EINVAL, "Invalid data in sysfs");
 
-  for (int i = errors_buf->num_err; i >= 0; i--) {
-    if (XRT_ERROR_CLASS(errors_buf->errors[i].err_code) != ecl)
-      continue;
-    errcode = errors_buf->errors[i].err_code;
-    ts = errors_buf->errors[i].ts;
-    break;
+  for (int i = errors_buf->num_err-1; i >= 0; i--) {
+    if (XRT_ERROR_CLASS(errors_buf->errors[i].err_code) == ecl)
+      return {errors_buf->errors[i].err_code, errors_buf->errors[i].ts};
   }
-  return {errcode, ts};
+  return {0, 0};
 }
 
 std::vector<xclErrorLast>
--- a/src/runtime_src/core/common/query_requests.h
+++ b/src/runtime_src/core/common/query_requests.h
@@ -1085,46 +1085,14 @@ struct xocl_errors : request
   get(const device*) const = 0;
 
   // Parse sysfs line and from class get error code and timestamp
+  XRT_CORE_COMMON_EXPORT
   static std::pair<uint64_t, uint64_t>
-  to_value(const std::vector<char>& buf, xrtErrorClass ecl)
-  {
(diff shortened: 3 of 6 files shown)


src/runtime_src/core/common/api/xrt_error.cpp
```diff
@@ -205,6 +206,27 @@ class error_impl
 public:
   error_impl(const xrt_core::device* device, xrtErrorClass ecl)
   {
+    //Code for new format; Binary array of error structs in sysfs
+    try {
+      std::vector<char> buf = xrt_core::device_query<xrt_core::query::xocl_errors>(device);
+      if (buf.empty())
+        return;
+      const xcl_errors *errors_buf = reinterpret_cast<xcl_errors *>(buf.data());
```
stsoe, line 214: Don't use repeated types.   Should be 
`auto error_buf = reinterpret_cast<const xcl_errors*>(buf.data());`

And coding convention is `type* var` but you avoid this convention by using auto.

src/runtime_src/core/common/api/xrt_error.cpp
```diff
@@ -205,6 +206,27 @@ class error_impl
 public:
   error_impl(const xrt_core::device* device, xrtErrorClass ecl)
   {
+    //Code for new format; Binary array of error structs in sysfs
+    try {
+      std::vector<char> buf = xrt_core::device_query<xrt_core::query::xocl_errors>(device);
```
stsoe, line 211: Please use auto, the type is explicit in xrt_core::query::xocl_errors.

src/runtime_src/core/common/api/xrt_error.cpp
```diff
@@ -205,6 +206,27 @@ class error_impl
...
+      for (int i = 0; i < errors_buf->num_err; i++) {
+        if (XRT_ERROR_CLASS(errors_buf->errors[i].err_code) != ecl)
+          continue;
+        m_errcode = XRT_ERROR_NUM(errors_buf->errors[i].err_code);
+        m_timestamp = errors_buf->errors[i].ts;
+      }
+      return;
+    } catch (const xrt_core::query::no_such_key&) {
+      // Ignoring for now. Check below for edge if not available
+      // query table of zocl doesn't have xocl_errors key
```
stsoe, line 227: This is a parsing routine specific to the query request.  I don't think this code should do reinterpret_cast of the query request return value as it is details that are specific to the xrt_core::query::xocl_errors query request.

Use the `struct xrt_core::query::xocl_errors` in query_requests.h to define your parsing of what the query request returns.

The code in this file should preferably not interpret raw data.  You should call the query request as you do, but then call a parsing routine in `struct xrt_core::query::xocl_errors` that interprets the raw data and validates it also.

There is nothing critical in this code, so I think a parsing routine in the struct could return a std::vector of pairs which would then leave the original code almost intact.

The new code is not checking for multiple errors for specified error class, it is just using the last. Maybe this is fine, but it is a change.

src/runtime_src/core/common/api/xrt_error.cpp
```diff
@@ -205,6 +206,27 @@ class error_impl
...
+    //Code for new format; Binary array of error structs in sysfs
+    try {
+      std::vector<char> buf = xrt_core::device_query<xrt_core::query::xocl_errors>(device);
+      if (buf.empty())
+        return;
+      const xcl_errors *errors_buf = reinterpret_cast<xcl_errors *>(buf.data());
+      if (errors_buf->num_err <= 0)
+        return;
+
+      for (int i = 0; i < errors_buf->num_err; i++) {
```
stsoe, line 218: What if num_err is bogus and exceeds the fixed size of `errors`?  There has to be some validation here.

stsoe (changes requested):
Looks simple enough, but I think we can simplify even more by moving some parsing and interpreting code to its proper place.

You can check other parsing routines in query_requests.{h,cpp}.

src/runtime_src/core/common/api/xrt_error.cpp
sarab96: updated

src/runtime_src/core/common/api/xrt_error.cpp
sarab96: updated

src/runtime_src/core/common/api/xrt_error.cpp
sarab96: check added

src/runtime_src/core/common/api/xrt_error.cpp
sarab96: Done.
Yes change in behavior as latest entry will be used which will be at the end of the array.
Multiple entries to be allowed

src/runtime_src/core/common/api/xrt_error.cpp
```diff
@@ -205,6 +206,19 @@ class error_impl
 public:
   error_impl(const xrt_core::device* device, xrtErrorClass ecl)
   {
+    //Code for new format; Binary array of error structs in sysfs
+    try {
+      auto buf = xrt_core::device_query<xrt_core::query::xocl_errors>(device);
```
stsoe, line 211: Much cleaner and leaner now.  Thanks!

xocl_errors is not a great name for a query unless this is temporary until error has moved to new format on zocl also.
At the very least, can we keep the key of the query_request the same for both xocl_errors and error?

src/runtime_src/core/common/query_requests.h
```diff
@@ -81,6 +82,7 @@ enum class key_type
   kds_cu_stat,
   kds_scu_stat,
   ps_kernel,
+  xocl_errors,
```
stsoe, line 85: Do we need this key?  Can't we just use the "error" key?

src/runtime_src/core/common/query_requests.h
```diff
@@ -1073,6 +1075,58 @@ struct error : request
   }
 };
 
+// Retrieve asynchronous xocl errors from xocl driver
+struct xocl_errors : request
+{
+  using result_type = std::vector<char>;
+  static const key_type key = key_type::xocl_errors;
```
stsoe, line 1082: I mean could this be key_type::error ?  I think the idea is that both zocl and xocl will respond to the same query request and return the same format, if so then when that happens we can simply remove existing `struct error` and rename this struct to error?

src/runtime_src/core/common/query_requests.h
```diff
@@ -17,6 +17,7 @@
 #ifndef xrt_core_common_query_requests_h
 #define xrt_core_common_query_requests_h
 
+#include "core/include/xclerr_int.h"
```
stsoe, line 20: Let's not include this one here.  You should move the parsing routine implementation to `query_requests.cpp` where this header is then included.  I do see that you need a forward declaration of `struct xclErrorLast` for your parsing function declarations.

src/runtime_src/core/common/query_requests.h
```diff
@@ -1073,6 +1075,58 @@ struct error : request
...
+{
+  using result_type = std::vector<char>;
+  static const key_type key = key_type::xocl_errors;
+
+  virtual boost::any
+  get(const device*) const = 0;
+
+  // Parse sysfs line and from class get error code and timestamp
+  static std::pair<uint64_t, uint64_t>
+  to_value(const std::vector<char>& buf, xrtErrorClass ecl)
```
stsoe, line 1089: Move implementation to query_requests.cpp

src/runtime_src/core/common/query_requests.h
```diff
@@ -1073,6 +1075,58 @@ struct error : request
...
+        continue;
+      errcode = errors_buf->errors[i].err_code;
+      ts = errors_buf->errors[i].ts;
+    }
+    return std::make_pair(errcode, ts);
+  }
+
+  // Parse sysfs raw data and get list of errors
+  static std::vector<xclErrorLast>
+  to_errors(const std::vector<char>& buf)
```
stsoe, line 1112: Move implementation to query_requests.cpp, but is the function even used???

src/runtime_src/core/common/query_requests.h
```diff
@@ -1073,6 +1075,58 @@ struct error : request
...
+      return std::make_pair(errcode, ts);
+    if (errors_buf->num_err > XCL_ERROR_CAPACITY)
+      return std::make_pair(errcode, ts);
+
+    for (int i = 0; i < errors_buf->num_err; i++) {
+      if (XRT_ERROR_CLASS(errors_buf->errors[i].err_code) != ecl)
+        continue;
+      errcode = errors_buf->errors[i].err_code;
+      ts = errors_buf->errors[i].ts;
+    }
```
stsoe, line 1106: What is the point of overwriting errcode and ts for every match.  You might as well return the first match, right?
The code would be greatly simpler by just returning
`{errors_buf->errors[i].err_code, errors_buf->errors[i].ts}`
You can get rid if 1091 and 1092, and for all the early-returns, just return `{0,0}`.

src/runtime_src/core/common/query_requests.h
```diff
@@ -1073,6 +1075,58 @@ struct error : request
...
+  {
+    uint64_t errcode = 0;
+    uint64_t ts = 0;
+    if (buf.empty())
+      return std::make_pair(errcode, ts);
+    auto errors_buf = reinterpret_cast<const xcl_errors *>(buf.data());
+    if (errors_buf->num_err <= 0)
+      return std::make_pair(errcode, ts);
+    if (errors_buf->num_err > XCL_ERROR_CAPACITY)
+      return std::make_pair(errcode, ts);
```
stsoe, line 1099: Is this not an error condition?  If so, please throw.

src/runtime_src/core/common/query_requests.h
```diff
@@ -1073,6 +1075,58 @@ struct error : request
...
+  // Parse sysfs line and from class get error code and timestamp
+  static std::pair<uint64_t, uint64_t>
+  to_value(const std::vector<char>& buf, xrtErrorClass ecl)
+  {
+    uint64_t errcode = 0;
+    uint64_t ts = 0;
+    if (buf.empty())
+      return std::make_pair(errcode, ts);
+    auto errors_buf = reinterpret_cast<const xcl_errors *>(buf.data());
+    if (errors_buf->num_err <= 0)
```
stsoe, line 1096: Why would num_err ever be less than 0?  Should num_err have been an unsigned int?

src/runtime_src/core/common/query_requests.h
```diff
@@ -1073,6 +1075,58 @@ struct error : request
...
+    if (buf.empty())
+      return errors;
+    auto errors_buf = reinterpret_cast<const xcl_errors *>(buf.data());
+    if (errors_buf->num_err <= 0)
+      return errors;
+    if (errors_buf->num_err > XCL_ERROR_CAPACITY)
+      return errors;
+
+    for (int i = 0; i < errors_buf->num_err; i++) {
+      errors.emplace_back(errors_buf->errors[i]);
```
stsoe, line 1124: No curly brace around single statement, but add empty line after statement.

Missing std::move:
`errors.emplace_back(std::move(errors_buf->errors[i]));`

src/runtime_src/core/common/query_requests.h
```diff
@@ -1073,6 +1075,58 @@ struct error : request
...
+    return std::make_pair(errcode, ts);
+  }
+
+  // Parse sysfs raw data and get list of errors
+  static std::vector<xclErrorLast>
+  to_errors(const std::vector<char>& buf)
+  {
+    std::vector<xclErrorLast> errors;
+    if (buf.empty())
+      return errors;
```
stsoe, line 1116: return {};

src/runtime_src/core/common/query_requests.h
```diff
@@ -1073,6 +1075,58 @@ struct error : request
...
+      ts = errors_buf->errors[i].ts;
+    }
+    return std::make_pair(errcode, ts);
+  }
+
+  // Parse sysfs raw data and get list of errors
+  static std::vector<xclErrorLast>
+  to_errors(const std::vector<char>& buf)
+  {
+    std::vector<xclErrorLast> errors;
```
stsoe, line 1114: move to line 1122 closer to where needed.

src/runtime_src/core/common/query_requests.h
```diff
@@ -1073,6 +1075,58 @@ struct error : request
...
+    if (buf.empty())
+      return errors;
+    auto errors_buf = reinterpret_cast<const xcl_errors *>(buf.data());
+    if (errors_buf->num_err <= 0)
+      return errors;
```
stsoe, line 1119: return {};

src/runtime_src/core/common/query_requests.h
```diff
@@ -1073,6 +1075,58 @@ struct error : request
...
+    if (errors_buf->num_err > XCL_ERROR_CAPACITY)
+      return errors;
```
stsoe, line 1121: return {};

stsoe (changes requested):
Much better organization. thanks.
Still few style issues and questions.

src/runtime_src/core/common/api/xrt_error.cpp
sarab96: Yes this is temporary. Will change to "errors" later as it will contain multiple errors.

src/runtime_src/core/common/query_requests.h
sarab96: Difficult. It has enum xrtErrorClass as well.

src/runtime_src/core/common/query_requests.h
sarab96: Yes it will be changed to same name for zocl & xocl. maybe "errors"

src/runtime_src/core/common/query_requests.h
sarab96: Done

src/runtime_src/core/common/query_requests.h
sarab96: It is defined as integer. I think better to check for <= 0 as sysfs may get garbage sometimes :)

src/runtime_src/core/common/query_requests.h
sarab96: Updated

src/runtime_src/core/common/query_requests.h
sarab96: Updated

src/runtime_src/core/common/query_requests.h
sarab96: Done

src/runtime_src/core/common/query_requests.h
sarab96: Done

src/runtime_src/core/common/query_requests.h
sarab96: Done

src/runtime_src/core/common/query_requests.h
sarab96: Done

src/runtime_src/core/common/query_requests.h
sarab96: Done

src/runtime_src/core/common/query_requests.h
sarab96: Updated. Not using move to keep raw sysfs data in original state

src/runtime_src/core/common/query_requests.h
sarab96: Yes it will be become one common key "errors" in future

src/runtime_src/core/common/query_requests.h
```diff
@@ -1073,6 +1075,26 @@ struct error : request
...
+
+  // Parse sysfs line and from class get error code and timestamp
+  XRT_CORE_COMMON_EXPORT
+  static std::pair<uint64_t, uint64_t>
+  to_value(const std::vector<char>& buf, xrtErrorClass ecl);
+
+  // Parse sysfs raw data and get list of errors
+  XRT_CORE_COMMON_EXPORT
+  static std::vector<xclErrorLast>
+  to_errors(const std::vector<char>& buf);
```
stsoe, line 1095: I don't see this function used?

src/runtime_src/core/common/query_requests.cpp
```diff
@@ -121,5 +122,48 @@ parse(const std::string& clock)
...
+
+std::pair<uint64_t, uint64_t>
+xrt_core::query::xocl_errors::
+to_value(const std::vector<char>& buf, xrtErrorClass ecl)
+{
+  uint64_t errcode = 0;
+  uint64_t ts = 0;
+  if (buf.empty())
+    return {0, 0};
+  auto errors_buf = reinterpret_cast<const xcl_errors *>(buf.data());
```
stsoe, line 134: new line before 134

src/runtime_src/core/common/query_requests.cpp
```diff
@@ -121,5 +122,48 @@ parse(const std::string& clock)
...
+  if (errors_buf->num_err <= 0)
+    return {0, 0};
+  if (errors_buf->num_err > XCL_ERROR_CAPACITY)
```
stsoe, line 137: new line before 137

src/runtime_src/core/common/query_requests.cpp
```diff
@@ -121,5 +122,48 @@ parse(const std::string& clock)
...
+    throw xrt_core::system_error(EINVAL, "Invalid data in sysfs");
+
+  for (int i = errors_buf->num_err; i >= 0; i--) {
+    if (XRT_ERROR_CLASS(errors_buf->errors[i].err_code) != ecl)
+      continue;
+    errcode = errors_buf->errors[i].err_code;
+    ts = errors_buf->errors[i].ts;
+    break;
```
stsoe, line 145: Maybe just
```
if (XRT_ERROR_CLASS(errors_buf->errors[i].err_code) != ecl)
      return {errors_buf->errors[i].err_code, errors_buf->errors[i].ts};
```

And remove line 130, 131

stsoe (changes requested):
Minor tweaks.  Looks good.

src/runtime_src/core/common/query_requests.cpp
sarab96: updated

src/runtime_src/core/common/query_requests.cpp
sarab96: updated

src/runtime_src/core/common/query_requests.cpp
sarab96: updated

src/runtime_src/core/common/query_requests.h
sarab96: For future use. Like in xbutil to dump all errors

stsoe (approved):

maxzhen (approved):

larry9523 (approved):

PR #5575, opened 2021-08-04
Added the XILINX_HLS paths to LD_LIBRARY_PATH before launching the de…

venkatp-xilinx:
Added the XILINX_HLS paths to LD_LIBRARY_PATH before launching the device process of SW_EMU


chvamshi-xilinx:
It is better to have loader for genericpcimodel instead of adding all the LD_LIBRARY_PATHs in driver before launching the device process.
Merging this change , but lets make sure you create a loader.

PR #5576, opened 2021-08-04
Fixing memory allocator issue in edge

chvamshi-xilinx:
Fixing an issue while checking whether memory is reserved or not in device tree.
We should return false if reserved-memory node is not available


chvamshi-xilinx:
Looks like space issue on machine.

chvamshi-xilinx:
retest this please

saifuddin-xilinx (approved):

PR #5585, opened 2021-08-06
Using existing property names for the new schema (bar mappings) in partition metadata for consistency

chvamshi-xilinx:
This fix contain following changes.

- We already have pcie_physical_function property which represents pf . Renaming "physical_function" with "pcie_physical_function".
- We already have  pcie_base_address_register which represents bar index. Renaming "bar" with "pcie_base_address_register"
- We have a reg property which represents offset and range. Using same here as well.


chvamshi-xilinx:
Changes are in https://github.com/Xilinx/XRT/pull/5595

PR #5590, opened 2021-08-06
CR-1107213 cmd_queue_versal_isr: not in submitted queue 2

chienwei-lan:


src/runtime_src/core/pcie/driver/linux/xocl/xocl_drv.h
```diff
@@ -1877,8 +1877,7 @@ struct xocl_intc_funcs {
 /* Only used in ERT sub-device polling mode */
 #define xocl_intc_ert_read32(xdev, off) \
 	(INTC_CB(xdev, csr_read32) ? \
-	 INTC_OPS(xdev)->csr_read32(INTC_DEV(xdev), off) : \
-	 -ENODEV)
+	 INTC_OPS(xdev)->csr_read32(INTC_DEV(xdev), off) : \ 0)
```
xdavidz, line 1880: not sure I understand this correctly, so if not ERT, we just return 0 instead of -ENODEV. Is this a workaround or logically correct?

chienwei-lan: We have to look the context

		mask = xocl_intc_ert_read32(xdev, (section_idx<<2));
		if (!mask)
			continue;
			
if INTC subdev does not present, it should return 0 instead of -ENODEV aka 0xFFFFFFED

xdavidz (approved):

PR #5591, opened 2021-08-08
Enable pybind11 doc generation and 23_bandwidth.py code cleanup

sonals:


src/python/pybind11/src/pyxrt.cpp
```diff
@@ -33,263 +33,264 @@ namespace py = pybind11;
...
+
+                                 i++;
+                             }
+
+                             r.start();
+                             return r;
+                         })
+        .def("group_id", &xrt::kernel::group_id, "Get the memory bank group id of an kernel argument")
+        .def("write_register", &xrt::kernel::write_register)
+        .def("read_register", &xrt::kernel::read_register);
```
stsoe, line 209: We are considering deprecating read/write_register from managed kernel APIs (xrt::kernel).   It would be better if we didn't wrap these APIs in python.  read/write_register is supported through custom IPs (xrt::ip).

build/build.sh
```diff
@@ -268,6 +268,8 @@ if [[ $opt == 1 ]]; then
 
   if [[ $docs == 1 ]]; then
       echo "make xrt_docs"
+      # Source the XRT environment so we can find modules like pyxrt
+      source ./opt/xilinx/xrt/setup.sh
```
stsoe, line 272: Does build.sh -docs really need to depend on having run a full build first?
Before this change, all you had to do was clone xrt; build.sh -docs,  where latter does not build XRT only docs, which makes it super fast to build/clean docs.

stsoe (changes requested):
The more python wrapping code I see, the less I like python.  This does not look scalable.
Few requests for changes.

src/python/pybind11/src/pyxrt.cpp
sonals: Makes sense, dropped it from the pybind11 doc.

build/build.sh
sonals: sphinx-build imports the python module (pyxrt in our case) and introspects the class/methods to generate the documentation. It looks like the following:

```


class pyxrt.bo

    Represents a buffer object

    address(self: pyxrt.bo) → int

        Return the device physical address of the buffer object

    class flags

        Buffer object creation flags

        Members:

            normal

            cacheable

            device_only

            host_only

            p2p

            svm

        name

    map(self: pyxrt.bo) → memoryview

        Create a byte accessible mmemory view of the buffer object

    read(self: pyxrt.bo, arg0: int, arg1: int) → numpy.ndarray[numpy.int8]

        Read from the buffer object requested number of bytes starting from specified offset

    size(self: pyxrt.bo) → int

        Return the size of the buffer object

    sync(self: pyxrt.bo, arg0: pyxrt.xclBOSyncDirection, arg1: int, arg2: int) → None

        Synchronize (DMA or cache flush/invalidation) the buffer in the requested direction

    write(self: pyxrt.bo, arg0: buffer, arg1: int) → None

        Write the provided data into the buffer object starting at specified offset


```

src/python/pybind11/src/pyxrt.cpp
```diff
@@ -33,263 +33,262 @@ namespace py = pybind11;
...
+        .value("ERT_CMD_STATE_QUEUED", ert_cmd_state::ERT_CMD_STATE_QUEUED)
+        .value("ERT_CMD_STATE_COMPLETED", ert_cmd_state::ERT_CMD_STATE_COMPLETED)
+        .value("ERT_CMD_STATE_ERROR", ert_cmd_state::ERT_CMD_STATE_ERROR)
+        .value("ERT_CMD_STATE_ABORT", ert_cmd_state::ERT_CMD_STATE_ABORT)
+        .value("ERT_CMD_STATE_SUBMITTED", ert_cmd_state::ERT_CMD_STATE_SUBMITTED)
+        .value("ERT_CMD_STATE_TIMEOUT", ert_cmd_state::ERT_CMD_STATE_TIMEOUT)
+        .value("ERT_CMD_STATE_NORESPONSE", ert_cmd_state::ERT_CMD_STATE_NORESPONSE)
+        .value("ERT_CMD_STATE_SKERROR", ert_cmd_state::ERT_CMD_STATE_SKERROR)
+        .value("ERT_CMD_STATE_SKCRASHED", ert_cmd_state::ERT_CMD_STATE_SKCRASHED)
+        .value("ERT_CMD_STATE_MAX", ert_cmd_state::ERT_CMD_STATE_MAX);
```
stsoe, line 63: I don't think we should expose ERT_CMD_STATE_MAX, it should have no use in host code.  In general, I don't really think we should have an enumerator exposed that can change value from release to release; maybe not so bad python but in a compiled application relying on ERT_CMD_STATE_MAX would not be good.  We should probably get rid of this enumerator.

build/build.sh
stsoe: Ok, guess there is no way around this then.

AShivangi (approved):

uday610 (approved):

xilinxgitops:
retest this please

src/python/pybind11/src/pyxrt.cpp
sonals: We will need the enums as that is the status of a few kernel methods. I will create another PR to remove the "ERT_CMD_STATE_MAX".

stsoe: @sonal.  I was not suggesting to remove all the enumerators, only the ERT_CMD_STATE_MAX which shouldn't be there in the first place in ert.h.

PR #5592, opened 2021-08-08
Vitis-1444 API for aborting hanging command

stsoe:
Add xrt::run::abort() to abort a run object.

Add example demonstrating use of abort() API.


stsoe:
retest this please - why the hold up?

PR #5598, opened 2021-08-09
Kill 23_bandwidth.py's numpy dependency

sonals:


chvamshi-xilinx:
Thanks @sonals  for making these changes. Things are working well on my local now.

PR #5601, opened 2021-08-09
2294: Generate a data transfer table in profile summary if memory monitors are present

jvillarre:
Check if monitors were added on memory resources, and create a table that shows the aggregate information on the monitored banks in the profile summary if profiling is enabled.


IshitaGhosh (approved):
looks good

PR #5606, opened 2021-08-10
CR-1096354 ASTeR xbutil(2) - host_mem command is not supported

AShivangi:
### Host-mem implementation
Tested on u250 2RP with multiple sizes = [4M, 128M, 512M, 1024M, 1G]
```
$ sudo xbutil configure --host-mem enable -d 0000:17:00.1 --size 256M

Host-mem enabled successfully
$ sudo xbutil configure --host-mem disable -d 0000:17:00.1 

Host-mem disabled successfully
```

The following STRs give errors with a custom error message:
```
# multiple devices
$ sudo xbutil configure --host-mem enable -d 0000:17:00.1 0000:18:00.1 --size 256M

# size missing unit
$ sudo xbutil configure --host-mem enable -d 0000:17:00.1 --size 256

# invalid size
$ sudo xbutil configure --host-mem enable -d 0000:17:00.1 --size aM

# missing size option for enable
$ sudo xbutil configure --host-mem enable -d 0000:17:00.1 

# invalid action
$ sudo xbutil configure --host-mem dis -d 0000:17:00.1

# missing sudo
$ xbutil configure --host-mem disable -d 0000:17:00.1
```


src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
```diff
@@ -0,0 +1,169 @@
+/**
+ * Copyright (C) 20211 Licensed under the Apache License, Version
```
rozumx, line 2: One too many 1's :^)

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+  try {
+    po::store(po::command_line_parser(_options).options(m_optionsDescription).positional(m_positionalOptions).run(), vm);
+    po::notify(vm); // Can throw
+  }
+  catch (po::error& e) {
+    std::cerr << "ERROR: " << e.what() << "\n\n";
+
+    printHelp();
+    return;
```
rozumx, line 97: This is an error.  Since it is one, we wish to indicate so with the return command.
```suggestion
    throw xrt_core::error(std::errc::operation_canceled);
```

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  // Exit if neither action or device specified
+  if(m_help || (m_action.empty() || m_devices.empty())) {
```
rozumx, line 101: Please don't combine these messages together.   Please create a unique check for each one so that the correct message can be produced prior to printing the help.

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
```diff
@@ -0,0 +1,169 @@
...
+    printHelp();
+    return;
```
rozumx, line 103:
```suggestion
    throw xrt_core::error(std::errc::operation_canceled);
```

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  bool enable = false;
+  try {
+    uint64_t size = 0;
```
rozumx, line 108: This code should be in an helper utility method. Lines 107-128.

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
```diff
@@ -0,0 +1,169 @@
...
+    if(!m_size.empty()) {
+      uint64_t bytes = 0;
+      auto units = m_size.substr(m_size.length()-1);
+      boost::to_upper(units);
+      if(units.compare("M") == 0)
+        bytes = 1024*1024;
+      else if(units.compare("G") ==0)
+        bytes = 1024*1024*1024;
+      else
+        throw xrt_core::error(std::errc::invalid_argument, "Please specify a valid size unit (M|G), eg: `256M`");
```
rozumx, line 118: A lot of assumptions are being made here.  What about k and T and what about the units, e.g., bytes.

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+  bool enable = false;
+  try {
+    uint64_t size = 0;
+    if(!m_size.empty()) {
+      uint64_t bytes = 0;
+      auto units = m_size.substr(m_size.length()-1);
+      boost::to_upper(units);
+      if(units.compare("M") == 0)
+        bytes = 1024*1024;
```
rozumx, line 114: It would be better to shift then multiple by 1024.

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
```diff
@@ -0,0 +1,169 @@
...
+#include "tools/common/XBUtilities.h"
+
+// 3rd Party Library - Include Files
+#include <boost/program_options.hpp>
+#include <boost/algorithm/string.hpp>
+namespace po = boost::program_options;
+
+namespace {
+bool
+string2action(std::string str)
```
rozumx, line 28: This algorithm while it does work seems to be an overkill, why didn't you use boost::iequals().

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
```diff
@@ -0,0 +1,169 @@
...
+        size = std::stoll(m_size.substr(0, m_size.length()-1));
+      } 
+      catch (const std::exception& ex) {
+        //out of range, invalid argument ex
+        throw xrt_core::error(std::errc::invalid_argument, "Value supplied to --size option is invalid");
+      }
+      size = size*bytes;
+    }
+
+    enable = string2action(m_action);
```
rozumx, line 130: Being that this is a boolean variable, it would have been simpler to use the following code:
```
if ((boost::iequals(m_action, "ENABLE") != 0) && (boost::iequals(m_action, "DISABLE") != 0) {
   std::cerr << boost::format("ERROR: Invalid action value: '%s'\n) % m_action;
   printHelp();
 }

bool enable =  boost::iequals(m_action, "ENABLE");

```

src/runtime_src/core/tools/xbutil2/OO_HostMem.h
```diff
@@ -0,0 +1,37 @@
...
+
+ public:
+  OO_HostMem( const std::string &_longName, bool _isHidden = false);
+
+ private:
+  std::vector<std::string> m_devices;
+  std::string m_action;
+  std::string m_size;
+  bool m_help;
+
```
rozumx, line 34: Please remove blank line.

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
```diff
@@ -0,0 +1,169 @@
...
+    : OptionOptions(_longName, _isHidden, "Controls host-mem functionality")
+    , m_devices({})
+    , m_action("")
+    , m_size("")
+    , m_help(false)
+{
+  m_optionsDescription.add_options()
+    ("device,d", boost::program_options::value<decltype(m_devices)>(&m_devices)->multitoken(), "The Bus:Device.Function (e.g., 0000:d8:00.0) device of interest")
+    ("action", boost::program_options::value<decltype(m_action)>(&m_action)->required(), "Action to perform: ENABLE or DISABLE")
+    ("size,s", boost::program_options::value<decltype(m_size)>(&m_size), "Size of host memory to be enabled (e.g. 256M, 1G)")
```
rozumx, line 62:
```suggestion
    ("size,s", boost::program_options::value<decltype(m_size)>(&m_size), "Size of host memory (bytes) to be enabled (e.g. 256M, 1G)")
```

src/runtime_src/core/tools/xbutil2/SubCmdConfigure.cpp
```diff
@@ -0,0 +1,134 @@
...
+    allow_unregistered().           // Allow for unregistered options
+    run();                          // Parse the options
+
+  po::variables_map vm;
+
+  try {
+    po::store(parsedCommonTop, vm);  // Can throw
+    po::notify(vm);                  // Can throw (but really isn't used)
+
+    // Multual DRC
```
rozumx, line 97: Should be refactored into a helper function for it is used in multiple locations.

src/runtime_src/core/tools/xbutil2/SubCmdConfigure.cpp
```diff
@@ -0,0 +1,134 @@
...
+    for (unsigned int index1 = 0; index1 < subOptionOptions.size(); ++index1) {
+      for (unsigned int index2 = index1 + 1; index2 < subOptionOptions.size(); ++index2) {
+        conflictingOptions(vm, subOptionOptions[index1]->longName(), subOptionOptions[index2]->longName());
+      }
+    }
+  } catch (const std::exception & e) {
+    std::cerr << "ERROR: " << e.what() << std::endl;
+    printHelp(commonOptions, hiddenOptions, subOptionOptions);
+    return;
```
rozumx, line 106:
```suggestion
    throw xrt_core::error(std::errc::operation_canceled);
```

src/runtime_src/core/tools/xbutil2/SubCmdConfigure.cpp
```diff
@@ -0,0 +1,134 @@
...
+    if (vm.count(subOO->longName().c_str()) != 0) {
+      optionOption = subOO;
+      break;
+    }
+  }
+
+  // No suboption print help
+  if (!optionOption) {
+    printHelp(commonOptions, hiddenOptions, subOptionOptions);
+    return;
```
rozumx, line 121:
```suggestion
    throw xrt_core::error(std::errc::operation_canceled);
```

rozumx (changes requested):
Changes requested.

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
AShivangi: fixed

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
AShivangi: fixed

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
AShivangi: fixed

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
AShivangi: fixed

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
AShivangi: fixed

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
AShivangi: fixed

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
AShivangi: Isn't shift a "C" thing? IMO, multiplication is much more readable

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
AShivangi: The user can only specify size in M or G. The minimum is 4M and max is 1G. 
I talked to Uday about the interface and he said that he hasn't gotten any complaints from the users. So I have ported the "--size" interface as it is from the old tools

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
AShivangi: Hmm. I am a little confused...
Is it better to have 2-3 logics in the same if? In one of the previous comments, you asked me to separate each logic out.

src/runtime_src/core/tools/xbutil2/OO_HostMem.h
AShivangi: fixed

src/runtime_src/core/tools/xbutil2/SubCmdConfigure.cpp
AShivangi: Just the following 3 lines?

src/runtime_src/core/tools/xbutil2/SubCmdConfigure.cpp
AShivangi: fixed

src/runtime_src/core/tools/xbutil2/SubCmdConfigure.cpp
AShivangi: fixed

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
```diff
@@ -0,0 +1,179 @@
...
+#include "tools/common/XBUtilities.h"
+
+// 3rd Party Library - Include Files
+#include <boost/program_options.hpp>
+#include <boost/algorithm/string.hpp>
+namespace po = boost::program_options;
+
+namespace {
+bool
+string2action(const std::string &str)
```
rozumx, line 28: Please remove this method.  It doesn't help the readability of the code base.

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
rozumx: The first half of sample code  validates if the value is correct.  If not an error occurs and lets the user know what has happen.

After that we know that the value is either "ENABLE" or "DISABLE" and we use the iequals to determine it.   

The flow here is:
1. Validate the values.
2. Determine what the correct value is.

Note: This is preferred for a boolean value.

src/runtime_src/core/tools/xbutil2/SubCmdConfigure.cpp
rozumx: Yes, the following three lines.  While the algorithm is short, not all folks will know how to create it so creating a helper function would be best.

src/runtime_src/core/tools/xbutil2/SubCmdConfigure.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+  // No suboption print help
+  if (!optionOption) {
+    printHelp(commonOptions, hiddenOptions, subOptionOptions);
+    throw xrt_core::error(std::errc::operation_canceled);
+  }
+
+  // 2) Process the top level options
+  std::vector<std::string> topOptions = po::collect_unrecognized(parsedCommonTop.options, po::include_positional);
+  if (help) {
```
rozumx, line 126: The curly brackets (e.g., '{' and '}') should not be used for 1 line operations.  Yeah, I know that I have done so in the past, but Soren has convinced me that we should do so.

src/runtime_src/core/tools/xbutil2/SubCmdConfigure.cpp
AShivangi: I tried that, but looks like I'll have to change a lot of structure, adding the helper in SubCmdConfigure doesn't work. I would have to add it to SubCmd.h which didn't seem right to me.

xilinxgitops:
retest this please

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+// 3rd Party Library - Include Files
+#include <boost/program_options.hpp>
+#include <boost/algorithm/string.hpp>
+namespace po = boost::program_options;
+
+namespace {
+
+uint64_t 
+string2size(const std::string &str)
```
rozumx, line 29: Lots going on here:
1. This should be a helper method to convert a string to a value.  Something like:  uint64_t stringToBytes(const std::string &str)

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
```diff
@@ -0,0 +1,176 @@
...
+    if ((boost::iequals(m_action, "ENABLE") != 0) && (boost::iequals(m_action, "DISABLE") != 0)) {
+      std::cerr << boost::format("ERROR: Invalid action value: '%s'\n") % m_action;
+      printHelp();
+      throw xrt_core::error(std::errc::operation_canceled);
+    }
+    enable =  boost::iequals(m_action, "ENABLE");
+
+    // Exit if ENABLE action is specified but size is not
+    if(enable && size == 0)
+      throw xrt_core::error(std::errc::invalid_argument, "Please specify the memory size for enablement");
```
rozumx, line 140:
```suggestion
      throw xrt_core::error(std::errc::invalid_argument, "Please specify a non-zero memory size.");
```

rozumx (changes requested):

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -691,3 +691,99 @@ XBUtilities::str_to_reset_obj(const std::string& str)
...
+    auto units = str.substr(str.length()-1);
+    boost::to_upper(units);
+    if(units.compare("B") == 0)
+      bytes = 1;
+    else if(units.compare("k") == 0)
+      bytes = 1024;
+    else if(units.compare("M") == 0)
+      bytes = 1024*1024;
+    else if(units.compare("G") == 0)
+      bytes = 1024*1024*1024;
```
rozumx, line 775: And finally:
`bytes = 1 < 30;`

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -691,3 +691,99 @@ XBUtilities::str_to_reset_obj(const std::string& str)
...
+XBUtilities::stringToBytes(const std::string &str)
+{
+  uint64_t size = 0;
+  if(!str.empty()) {
+    uint64_t bytes = 0;
+    auto units = str.substr(str.length()-1);
+    boost::to_upper(units);
+    if(units.compare("B") == 0)
+      bytes = 1;
+    else if(units.compare("k") == 0)
```
rozumx, line 770: Hmm...did you test to see if this value is triggered.   In other words, I do believe 'k' should be capitalized.

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -692,6 +692,102 @@ XBUtilities::str_to_reset_obj(const std::string& str)
...
+  return clock_str != clock_map.end() ? clock_str->second : "N/A";
+}
+
+uint64_t 
+XBUtilities::stringToBytes(const std::string &str)
+{
+  uint64_t size = 0;
+  if(!str.empty()) {
+    uint64_t bytes = 0;
+    auto units = str.substr(str.length()-1);
```
rozumx, line 766: This code depends on the need to always have a unit.  It should be updated to support a "default" value of bytes.

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -692,6 +692,102 @@ XBUtilities::str_to_reset_obj(const std::string& str)
...
+      bytes = 1024;
+    else if(units.compare("M") == 0)
+      bytes = 1024*1024;
+    else if(units.compare("G") == 0)
+      bytes = 1024*1024*1024;
+    else
+      throw xrt_core::error(std::errc::invalid_argument);
+      
+    try {
+      size = std::stoll(str.substr(0, str.length()-1));
```
rozumx, line 780: Hard coded assumption.   Which is dangerous.   

What should occur is:
1. The original string should be examined to determine the unit.
2. The unit should be removed from the string (e.g., create a new string with just the value) (e.g., `"35 K" -> "35 "`)
3. The new string value should have any leading and training white space removed (e.g., `"35 " -> "35"`)
4. Then the string values should be determined.

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -692,6 +692,102 @@ XBUtilities::str_to_reset_obj(const std::string& str)
...
+    } 
+    catch (const std::exception&) {
+      //out of range, invalid argument ex
+      throw xrt_core::error(std::errc::invalid_argument);
+    }
+    size = size*bytes;
```
rozumx, line 786:
```suggestion
    size *= unit_bytes;
```

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -692,6 +692,102 @@ XBUtilities::str_to_reset_obj(const std::string& str)
...
+{
+  auto clock_str = clock_map.find(id);
+  return clock_str != clock_map.end() ? clock_str->second : "N/A";
+}
+
+uint64_t 
+XBUtilities::stringToBytes(const std::string &str)
+{
+  uint64_t size = 0;
+  if(!str.empty()) {
```
rozumx, line 764: Hmm...why are we not throwing an exception?

rozumx (changes requested):
Changes requested.

dayeh-xilinx:
retest this please

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -692,6 +692,111 @@ XBUtilities::str_to_reset_obj(const std::string& str)
...
+}
+
+uint64_t 
+XBUtilities::string_to_bytes(const std::string &str)
+{
+  uint64_t size = 0;
+  if(str.empty())
+    throw xrt_core::error(std::errc::invalid_argument);
+
+  uint64_t unit_bytes = 0;
```
rozumx, line 769: Please move this 'down" in this method to just before it is needed.

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -692,6 +692,111 @@ XBUtilities::str_to_reset_obj(const std::string& str)
...
+  std::string units;
+  try {
+    for(const auto& x : str) {
+      if(std::isdigit(x))
```
rozumx, line 773: Why is this code here?  It doesn't do anything :(

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -692,6 +692,111 @@ XBUtilities::str_to_reset_obj(const std::string& str)
...
+        size = size*10 + (x - '0');
+      else if(std::isalpha(x))
+        units += x;
```
rozumx, line 776: While this detects if there is more than one unit value, it makes the error rather difficult to understand later.

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -692,6 +692,111 @@ XBUtilities::str_to_reset_obj(const std::string& str)
...
+    }
+    size = std::stoll(str.substr(0, str.length()-1));
```
rozumx, line 778: Doesn't this fail by given the wrong result if there are no units.  For example: "1234" would be evaluated as "123".

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -692,6 +692,111 @@ XBUtilities::str_to_reset_obj(const std::string& str)
...
+{
+  auto clock_str = clock_map.find(id);
+  if (clock_str != clock_map.end())
+    return clock_str->second;
+  
+  throw xrt_core::error(std::errc::invalid_argument);
+}
+
+uint64_t 
+XBUtilities::string_to_bytes(const std::string &str)
```
rozumx, line 763: This helper function should support and detect issues for the following strings values:
1.  "123", "123B", "123K", "123M", and "123G"
2.  " 123 K "  << Space before, between value and unit, and space at the end.

The later isn't as nasty as it looks.  The algorithm could be as follows:
1. Remove leading and trailing spaces (boost::algorithm::trim();)
2. Examine the last character to see if it is a unit.  If so record and remove it from the string.
3. Trim the string again to remove and spaces between the value and unit.
4. Decode the string value.

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
```diff
@@ -0,0 +1,158 @@
...
+    size = XBUtilities::string_to_bytes(m_size);
+  } 
+  catch(const xrt_core::error&) {
+    std::cerr << "Value supplied to --size option is invalid" << std::endl;
+    throw xrt_core::error(std::errc::operation_canceled);
+  }
+
+  bool enable = false;
+  try {
+
```
rozumx, line 111: Lots empty lines....is there a reason for this?

rozumx (changes requested):
Changes requested.

rozumx (approved):
Approved

dayeh-xilinx:
retest this please. Build failed for my pipeline bug. retest again.

PR #5612, opened 2021-08-11
Support setting PS kernel global buffer argument

stsoe:
Create arg setter for PS kernel and override setting of global buffer
argument to include the size of the BO.


dayeh-xilinx:
retest this please. boardfarm test failed. try again.

stsoe:
Not needed this release

PR #5618, opened 2021-08-12
For Windows IO mgr need to pass complete buffer as it cannot read any

pranjalv-xilinx:
random user virtual addresses directly.  So avoid doing copy from  user virtual address and instead copy all data to single buffer and send it to through IO manager.

### DIFF
--- a/src/runtime_src/core/pcie/windows/shim.cpp
+++ b/src/runtime_src/core/pcie/windows/shim.cpp
@@ -561,7 +561,7 @@ done:
     auto kernels = xrt_core::xclbin::get_kernels(top);
     /* Calculate size of kernels */
     for (auto& kernel : kernels) {
-        ksize += sizeof(kernel_info) + sizeof(argument_info) * kernel.args.size();
+        ksize += sizeof(kernel_info) + sizeof(argument_info) * kernel.args.size() -1;
     }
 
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/pcie/windows/shim.cpp
```diff
@@ -553,18 +553,22 @@ struct shim
...
 
     auto top = reinterpret_cast<const axlf*>(ImageBuffer);
 
-    XOCL_READ_AXLF_ARGS axlf_obj = {0};
-
     auto kernels = xrt_core::xclbin::get_kernels(top);
     /* Calculate size of kernels */
     for (auto& kernel : kernels) {
-        axlf_obj.ksize += sizeof(kernel_info) + sizeof(argument_info) * kernel.args.size();
+        ksize += sizeof(kernel_info) + sizeof(argument_info) * kernel.args.size();
```
arpitxilinx, line 564: should we do -1 for the sake of completion?

pranjalv-xilinx: Yes.. fixed

arpitxilinx (approved):

dayeh-xilinx:
retest this please

dkotipal-xilinx (approved):

PR #5633, opened 2021-08-13
use AIE FAL to report resource utilization

hackwa:
use aie fal to reserve profile counters
use aie fal to check trace resource availability
print utilization
make "conflicts" as default aie profile metric
use aie fal function to set threshold instead of aie driver

### DIFF
--- a/src/runtime_src/xdp/profile/plugin/aie_profile/aie_plugin.cpp
+++ b/src/runtime_src/xdp/profile/plugin/aie_profile/aie_plugin.cpp
@@ -265,7 +265,8 @@ namespace xdp {
       std::string modType = isCore ? "core" : "memory";
       msg << "Only " << numFreeCtr << " out of " << numTotalEvents
           << " metrics were available for aie "
-          << modType <<  " module profiling due to resource constraints."
+          << modType <<  " module profiling due to resource constraints. "
+          << "AIE profiling uses performance counters which could be taken up by aie trace, ecc etc."
           << std::endl;
 
       msg << "Available metrics : ";
--- a/src/runtime_src/xdp/profile/plugin/aie_profile/aie_plugin.cpp
+++ b/src/runtime_src/xdp/profile/plugin/aie_profile/aie_plugin.cpp
@@ -26,7 +26,6 @@
 #include "core/common/config_reader.h"
 #include "core/include/experimental/xrt-next.h"
 #include "core/edge/user/shim.h"
-#include "core/edge/common/aie_parser.h"
 #include "xdp/profile/database/database.h"
 
 #include <boost/algorithm/string.hpp>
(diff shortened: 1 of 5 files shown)


dayeh-xilinx:
retest this please

jvillarre (approved):
Changes look good.  This pull request additionally changes the default metric set to "conflicts."

src/runtime_src/xdp/profile/plugin/aie_profile/aie_plugin.cpp
```diff
@@ -146,6 +145,36 @@ namespace xdp {
       {"dma_stalls_mm2s",       {XAIE_EVENT_DMA_MM2S_0_STALLED_LOCK_ACQUIRE_MEM,
                                  XAIE_EVENT_DMA_MM2S_1_STALLED_LOCK_ACQUIRE_MEM}}
     };
+
+    // String event values for guidance and output
+    mCoreEventStrings = {
```
pgschuey, line 150: It would be better if we instead provided a function that converts an XAie_Events enum to a string. That way, it would automatically pick up any adjustments to metric sets.

src/runtime_src/xdp/profile/plugin/aie_profile/aie_plugin.cpp
```diff
@@ -173,6 +202,91 @@ namespace xdp {
...
+    auto& EventStrings =    isCore ? mCoreEventStrings[metricSet] : mMemoryEventStrings[metricSet];
+
+    auto numTotalEvents = requestedEvents.size();
+    if (numFreeCtr < numTotalEvents) {
+      std::stringstream msg;
+      std::string modType = isCore ? "core" : "memory";
+      msg << "Only " << numFreeCtr << " out of " << numTotalEvents
+          << " metrics were available for aie "
+          << modType <<  " module profiling due to resource constraints. "
+          << "AIE profiling uses performance counters which could be taken up by aie trace, ecc etc."
```
pgschuey, line 269: Please change "taken up by aie trace, ecc etc." to "already used by AIE trace, ECC, etc." or alike.

src/runtime_src/xdp/profile/plugin/aie_profile/aie_plugin.cpp
hackwa: good idea. but it'd need to contain all enums which is a large amount. We can add it in future.

pgschuey (changes requested):
Changes look good. Please address my two comments.

pgschuey (approved):

src/runtime_src/xdp/profile/plugin/aie_profile/aie_plugin.cpp
hackwa: filed a CR for that

PR #5634, opened 2021-08-15
CR-1105866 : Unnecessarily alarming message when skipping SC flash

rozumx:
Due to business and implementation logic being intertwined a large amount of refactoring was needed to support this request.

Work Done
---------
+ Updated the flow to be more friendly with Fixed SC images.
+ Removed support for multiple devices.
+ Refactored code to have separate functions to program the SC and base images.
+ Simplified status logic

Sample Output
```
$ sudo xbmgmt program -d 0000:65:00.0 -b --force
Device : [0000:65:00.0]

Current Configuration
  Platform             : xilinx_u30_gen3x4_base_1
  SC Version           : 6.3.6(FIXED)
  Platform ID          : 0x26291aa7c111172e


Incoming Configuration
  Deployment File      : partition.xsabin
  Deployment Directory : /lib/firmware/xilinx/26291aa7c111172ecc8675d18da15593
  Size                 : 64,095,544 bytes
  Timestamp            : Wed May  5 03:45:06 2021

  Platform             : xilinx_u30_gen3x4_base_1
  SC Version           : 6.3.6
  Platform UUID        : 26291AA7-C111-172E-CC86-75D18DA15593
Are you sure you wish to proceed? [Y/n]: Y (Force override)

INFO: Fixed Satellite Controller.
INFO: Forcing flashing of the base (e.g., shell) image (Force flag is set).
[0000:65:00.0] : Updating base (e.g., shell) flash image...
[PASSED] : Flash erased < 1m 42s >
[PASSED] : Flash programmed < 1m 8s >
[PASSED] : Flash verified < 25s >
Report                                                                                                                                                                          
  [0000:65:00.0] : Satellite Controller (SC) is either up-to-date, fixed, or not installed. No actions taken.                                                                   
  [0000:65:00.0] : Successfully flashed the base (e.g., shell) image                                                                                                            
                                                                                                                                                                                
Device flashed successfully.                                                                                                                                                    
```


src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -278,203 +289,238 @@ pretty_print_platform_info(const boost::property_tree::ptree& _ptDevice)
...
+        need_warm_reboot = true;
+      } else
+        report_stream << boost::format("  [%s] : Satellite Controller (SC) is either up-to-date, fixed, or not installed. No actions taken.\n") % getBDF(p.first);
+
+      // 2) Flash shell image
+      if (update_shell(p.first, p.second) == true)  {
+        report_stream << boost::format("  [%s] : Successfully flashed the base (e.g., shell) image\n") % getBDF(p.first);
+        needreboot = true;
+      } else
+        report_stream << boost::format("  [%s] : Base (e.g., shell) image is up-to-date.  No actions taken.\n") % getBDF(p.first);
```
AShivangi, line 507: Should SC flash be independent of shell flash, i.e., have their own try/catches. I think we should continue to try and flash the shell even if the SC flash fails (eg: when an SC is not operational)

AShivangi (commented):
Looks good

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
rozumx: Hmm...is this a requirement that Max or Sonal has mentioned.   

My logic behind this code is simple:
_If the SC flash failed then we shouldn't attempt to program the shell._

For if we do, then we could brick the board even further.

Note: Failure to program the board is NOT the same as skipping programing the SC (e.g., fixed or versions the same).

Thoughts?

AShivangi: As far as I know, it is not a requirement. But as per the old flashing code, we would give a warning:
`WARNING: Failed to update SC firmware on card [...]`
and continue with flashing the shell.

@maxzhen, what is the desired behavior here?

maxzhen: In the early days, SC flash fails left and right and board works just fine without SC. Hence we don't really care if SC flashed properly back then. I think we should stop updating the board if SC can't be flash now since if we continue, the board will be in a unsupported state.
I'm merging this PR now since @AShivangi has already approved the change (but forgot to update the status)

PR #5636, opened 2021-08-16
CR-1107957: Print profiling summary table for data transfers to memory resources

jvillarre:
This pull request reintroduces code to set when xclbins have memory monitors in the XDP data structures that was erroneously removed when developing the feature and merging with the master branch.


IshitaGhosh (approved):
Changes look good

PR #5641, opened 2021-08-17
CR-1107824: Align xocl_debug log with profiling data

jvillarre:
This pull request changes the internal OpenCL debug file generation (enabled with xocl_debug=true) to use the zero point inside time_ns() instead of its own local zero point.  Removing this local zero time aligns the timestamps in the generated debug file with any timestamps generated by profiling.

### DIFF
--- a/src/runtime_src/xocl/core/debug.cpp
+++ b/src/runtime_src/xocl/core/debug.cpp
@@ -40,8 +40,15 @@
 
 namespace {
 
+static unsigned long get_time_zero()
+{
+  // Originally: return xocl::time_ns();
+  return 0 ;
+}
+
 static bool s_debug_on = false;
 static std::string s_debug_log;
+static unsigned long s_zero = get_time_zero();
 
 // Event information
 namespace event {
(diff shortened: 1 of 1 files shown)


src/runtime_src/xocl/core/debug.cpp
```diff
@@ -42,7 +42,6 @@ namespace {
 
 static bool s_debug_on = false;
 static std::string s_debug_log;
-static unsigned long s_zero = xocl::time_ns();
```
stsoe, line 45: Let's just do 
`s_zero = get_time_zero(); `

and get_time_zero() returns 0 by default but can be changed to original behavior under ini key if for some reason the old behavior is needed.

src/runtime_src/xocl/core/debug.cpp
```diff
@@ -68,10 +67,10 @@ struct info
...
         m_times[CL_RUNNING] = m_times[CL_COMPLETE];
     ostr << id << " " << m_command_type << " "
-         << (m_times[CL_QUEUED]-s_zero) << " "
-         << (m_times[CL_SUBMITTED]-s_zero) << " "
-         << (m_times[CL_RUNNING]-s_zero) << " "
-         << (m_times[CL_COMPLETE]-s_zero);
+         << m_times[CL_QUEUED] << " "
+         << m_times[CL_SUBMITTED] << " "
+         << m_times[CL_RUNNING] << " "
+         << m_times[CL_COMPLETE];
```
stsoe, line 73: No change per above.

src/runtime_src/xocl/core/debug.cpp
```diff
@@ -187,9 +186,6 @@ init()
 
   ::event::init();
 
-  // reset time zero
-  s_zero = xocl::time_ns();
```
stsoe, line 191: get_time_zero()

stsoe (changes requested):
Let's keep the existing code for easy resurrection in case it is needed.

src/runtime_src/xocl/core/debug.cpp
jvillarre: Added get_time_zero function as requested.

src/runtime_src/xocl/core/debug.cpp
jvillarre: Reverted change

src/runtime_src/xocl/core/debug.cpp
jvillarre: Replaced with call to get_time_zero

dayeh-xilinx:
retest this please.

stsoe (approved):

PR #5642, opened 2021-08-17
CR-1107266: sw_emu - tcp socket creation in thread for new edge flow

vboggara-xilinx:
Implementation for tcp socket creation in thread for new edge flow

### DIFF
--- a/src/runtime_src/core/edge/sw_em/generic_pcie_hal2/shim.cxx
+++ b/src/runtime_src/core/edge/sw_em/generic_pcie_hal2/shim.cxx
@@ -865,6 +865,13 @@ namespace xclcpuemhal2 {
     return aieFlag;
   }
 
+  void CpuemShim::socketConnection(bool isTCPSocket) {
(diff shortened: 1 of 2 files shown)


dayeh-xilinx:
retest this please.

src/runtime_src/core/edge/sw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -865,6 +865,13 @@ namespace xclcpuemhal2 {
     return aieFlag;
   }
 
+  void CpuemShim::socketConnection(bool isTCPSocket) {
+    if (mLogStream.is_open()) mLogStream << __func__ << "TCP connection started" << std::endl;
+    if (!sock)
+      sock = new unix_socket(isTCPSocket);
```
chvamshi-xilinx, line 871: The names are going to be confusing. Please rename class with socket instead of unix_socket in next PR. That way it will be easier to know type of socket it is using for connection.

PR #5643, opened 2021-08-17
CR-1102209 : xbutil2 validate should use non-zero return code if validation failed.

rozumx:
Work Done
---------
+ Updated run_test_suite_device() with a test failure return value
+ Updated run_tests_on_device() with a test summary failure return value
+ Update execute() to throw an operation_canceled exception if any validation test fails.  This will result in a non-zero return value from xbutil.


dayeh-xilinx:
retest this please.

AShivangi (approved):

PR #5644, opened 2021-08-18
work-around for all Sprite failures. Considering SC is fixed is is_sc…

chvamshi-xilinx:
This change contains a  work-around for all Sprite failures. Considering SC is fixed if is_sc_fixed query throws an exception.

Verified on U200 and U.2 Platform.


rozumx (approved):
Thank you for isolating the issue and making this change.

PR #5647, opened 2021-08-18
CR-1108292: Resolve references in profiling trace files to missing strings in the mapping section

jvillarre:
Occasionally, events will be generated in trace files that refer to strings that do not exist in the string mapping section of the file. 
 This was due to multithreaded access and addition to the database.  This pull request adds multithreaded protection for the string table in the dynamic database.


hackwa (approved):
looks good!

PR #5646, opened 2021-08-18
CR-1107361 u55N xbutil examine inconsistent output

AShivangi:
Output:
```
$ sudo xbmgmt2 examine -d b3:00

------------------------------------------
1/1 [0000:b3:00.0] : xilinx_u55n_recovery
------------------------------------------
Flash properties
  Type                 : spi
  Serial Number        : N/A

Device properties
  Type                 : u55n
  Name                 : N/A
  Config Mode          : 2
  Max Power            : N/A

Flashable partitions running on FPGA
  Platform             : xilinx_u55n_recovery
  SC Version           : INACTIVE
  Platform ID          : 0x0

Flashable partitions installed in system
  <none found>

WARNING  : No shell is installed on the system.

```

PR #5653, opened 2021-08-19
CR-1101247 xbutil2 validate platform ID is 0x0 for U50 and U250

AShivangi:
Work Done:
- Use ReportPlatform to get basic information about the device
- Since we only operate on a single device, remove the validation summary
- Small fix related to: https://github.com/Xilinx/XRT/pull/5646

Sample output:

```
$ xbutil validate -d 65:00 -r quick
Starting validation for 1 devices

Validate Device           : [0000:65:00.1]
    Platform              : xilinx_u200_xdma_201830_2
    SC Version            : 4.2.0
    Platform ID           : 0x5d1211e8
....


$ xbutil validate -d 17:00 -r quick                                         
Starting validation for 1 devices                                                                                                

Validate Device           : [0000:17:00.1]
    Platform              : xilinx_u30_gen3x4_base_1
    SC Version            : 6.3.6                   
    Platform ID           : 937ED708-67CF-3350-BC06-304053F4293C
....
```

### DIFF
--- a/src/runtime_src/core/tools/common/XBHelpMenus.cpp
+++ b/src/runtime_src/core/tools/common/XBHelpMenus.cpp
@@ -700,7 +700,7 @@ XBUtilities::produce_reports( xrt_core::device_collection devices,
       bool is_mfg = false;
       try {
         is_mfg = xrt_core::device_query<xrt_core::query::is_mfg>(device);
-      } catch (...) {}
+      } catch (const xrt_core::query::exception&) {}
 
       //if factory mode
       std::string platform = "<not defined>";
@@ -711,7 +711,7 @@ XBUtilities::produce_reports( xrt_core::device_collection devices,
         else {
           platform = xrt_core::device_query<xrt_core::query::rom_vbnv>(device);
         }
-      } catch(...) {
+      } catch(const xrt_core::query::exception&) {
         // proceed even if the platform name is not available
       }
       std::string dev_desc = (boost::format("%d/%d [%s] : %s\n") % ++dev_idx % devices.size() % ptDevice.get<std::string>("device_id") % platform).str();
@@ -721,7 +721,11 @@ XBUtilities::produce_reports( xrt_core::device_collection devices,
       consoleStream << std::string(dev_desc.length(), '-') << std::endl;
 
       auto is_ready = xrt_core::device_query<xrt_core::query::is_ready>(device);
-      auto is_recovery = xrt_core::device_query<xrt_core::query::is_recovery>(device);
+      bool is_recovery = false;
+      try {
+        is_recovery = xrt_core::device_query<xrt_core::query::is_recovery>(device);
+      }
+      catch(const xrt_core::query::exception&) { }
 
       for (auto &report : reportsToProcess) {
         if (report->isDeviceRequired() == false)
--- a/src/runtime_src/core/tools/common/XBHelpMenus.cpp
+++ b/src/runtime_src/core/tools/common/XBHelpMenus.cpp
@@ -700,10 +700,13 @@ XBUtilities::produce_reports( xrt_core::device_collection devices,
       bool is_mfg = false;
       try {
         is_mfg = xrt_core::device_query<xrt_core::query::is_mfg>(device);
-      } catch (const xrt_core::query::exception&) {}
+      } 
+      catch (const xrt_core::query::exception&) {
+        is_mfg = false;
+      }
 
       //if factory mode
-      std::string platform = "<not defined>";
+      std::string platform;
       try {
         if (is_mfg) {
           platform = "xilinx_" + xrt_core::device_query<xrt_core::query::board_name>(device) + "_GOLDEN";
@@ -711,8 +714,10 @@ XBUtilities::produce_reports( xrt_core::device_collection devices,
         else {
           platform = xrt_core::device_query<xrt_core::query::rom_vbnv>(device);
         }
-      } catch(const xrt_core::query::exception&) {
+      } 
+      catch(const xrt_core::query::exception&) {
         // proceed even if the platform name is not available
+        platform = "<not defined>";
       }
       std::string dev_desc = (boost::format("%d/%d [%s] : %s\n") % ++dev_idx % devices.size() % ptDevice.get<std::string>("device_id") % platform).str();
       consoleStream << std::endl;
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -1349,9 +1349,21 @@ get_platform_info(const std::shared_ptr<xrt_core::device>& device,
 {
   auto bdf = xrt_core::device_query<xrt_core::query::pcie_bdf>(device);
   ptTree.put("device_id", xrt_core::query::pcie_bdf::to_string(bdf));
-  ptTree.put("platform", xrt_core::device_query<xrt_core::query::rom_vbnv>(device));
-  ptTree.put("sc_version", xrt_core::device_query<xrt_core::query::xmc_sc_version>(device));
-  ptTree.put("platform_id", (boost::format("0x%x") % xrt_core::device_query<xrt_core::query::rom_time_since_epoch>(device)));
+
+  boost::property_tree::ptree platform_report;
+  boost::property_tree::ptree empty_ptree;
```
rozumx, line 1354: It would be great if this value was a "const".  My guess is that line 1358 will fail if it is....don't 100% know.

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -1349,9 +1349,21 @@ get_platform_info(const std::shared_ptr<xrt_core::device>& device,
...
+  auto report = std::make_shared<ReportPlatforms>();
+  report->getPropertyTreeInternal(device.get(), platform_report);
+  
+  const boost::property_tree::ptree& platforms = platform_report.get_child("platforms", empty_ptree);
+  for(auto& kp : platforms) {
+    const boost::property_tree::ptree& pt_platform = kp.second;
+    const boost::property_tree::ptree& pt_static_region = pt_platform.get_child("static_region", empty_ptree);
+    ptTree.put("platform", pt_static_region.get<std::string>("vbnv"));
```
rozumx, line 1362: If the child "static_region" doesn't exist, then the empty_ptree will be returned (which is good).

Unfortunately, "vbnv", "interface_uuid", and "controller.satellite_controller.version" will not be found and exceptions will be raised, which isn't so nice :(.

Wouldn't it be better if there was a` pt_static_region.empty() `check before retrieving this data?

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -1349,9 +1349,21 @@ get_platform_info(const std::shared_ptr<xrt_core::device>& device,
...
-  ptTree.put("sc_version", xrt_core::device_query<xrt_core::query::xmc_sc_version>(device));
-  ptTree.put("platform_id", (boost::format("0x%x") % xrt_core::device_query<xrt_core::query::rom_time_since_epoch>(device)));
+
+  boost::property_tree::ptree platform_report;
+  boost::property_tree::ptree empty_ptree;
+  auto report = std::make_shared<ReportPlatforms>();
+  report->getPropertyTreeInternal(device.get(), platform_report);
+  
+  const boost::property_tree::ptree& platforms = platform_report.get_child("platforms", empty_ptree);
+  for(auto& kp : platforms) {
```
rozumx, line 1359: Is it possible for there to be more than 1 platform in the collection.   

If it is then the data below will overwrite the previous data...which isn't good.

If it isn't, then it would be nice to add a check right before the for loop to see if the number of elements is 0 or 1.  If not then and internal exception should be raised.

src/runtime_src/core/tools/common/XBHelpMenus.cpp
```diff
@@ -700,7 +700,7 @@ XBUtilities::produce_reports( xrt_core::device_collection devices,
       bool is_mfg = false;
       try {
         is_mfg = xrt_core::device_query<xrt_core::query::is_mfg>(device);
-      } catch (...) {}
+      } catch (const xrt_core::query::exception&) {}
```
rozumx, line 703: As constructed this code works for it relies on the default value.  
That said, have a blank catch body is makes it more difficult to understand what the intention is.   For this use-case (and similar ones),  it would be best to "repeat" the intention.  In this case, it would be:

```
} catch (const xrt_core::query::exception&) {
  is_mfg = false;
}
```

src/runtime_src/core/tools/common/XBHelpMenus.cpp
```diff
@@ -711,7 +711,7 @@ XBUtilities::produce_reports( xrt_core::device_collection devices,
         else {
           platform = xrt_core::device_query<xrt_core::query::rom_vbnv>(device);
         }
-      } catch(...) {
+      } catch(const xrt_core::query::exception&) {
```
rozumx, line 714: Following a similar use case as mentioned above, have the catch body set the platform name to "<not defined">.

If this is performed, then set the initial value of platform to an empty string.

src/runtime_src/core/tools/common/XBHelpMenus.cpp
```diff
@@ -721,7 +721,11 @@ XBUtilities::produce_reports( xrt_core::device_collection devices,
       consoleStream << std::string(dev_desc.length(), '-') << std::endl;
 
       auto is_ready = xrt_core::device_query<xrt_core::query::is_ready>(device);
-      auto is_recovery = xrt_core::device_query<xrt_core::query::is_recovery>(device);
+      bool is_recovery = false;
+      try {
+        is_recovery = xrt_core::device_query<xrt_core::query::is_recovery>(device);
+      }
+      catch(const xrt_core::query::exception&) { }
```
rozumx, line 728: In the catch body, set is_recovery to false.

rozumx (changes requested):
Some changes requested.

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
AShivangi: I can't declare `platform_report` as const :(

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
AShivangi: It isn't possible to have more than one platform. (At least not yet)

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
AShivangi: Yes, I missed those. Instead of checking for empty, the validation should continue. So I added `N/A` if any of the information is not found

rozumx (approved):
Approved

PR #5654, opened 2021-08-19
Fixes for CR-1108351, CR-1108354, CR-1108359, and CR-1108361

jvillarre:
CR-1108351: Renamed the "Host API" section for each of the host side writers to reflect which plugin generated it.  Native API Host Trace, Low Overhead OpenCL Host Trace, HAL Host Trace, and OpenCL Host trace are the new section names so they can be distinguished on the final visualization.

CR-1108354: Adjusted Native API Host Trace so that data transfers appeared at the same level as OpenCL and Low Overhead OpenCL sections.

CR-1108359: Changed the section names of Low Overhead OpenCL and OpenCL trace so they can be distinguished in the final visualization.

CR-1108361: Low Overhead OpenCL and OpenCL profiling plugins were collecting the same types of events, leading to all the OpenCL API calls going in one section when both plugins were loaded.  This pull request adds a boolean to the OpenCL API event type in profiling to distinguish between the two, and the writers now only collect the events for the proper section.

### DIFF
--- a/src/runtime_src/xdp/profile/database/events/opencl_api_calls.cpp
+++ b/src/runtime_src/xdp/profile/database/events/opencl_api_calls.cpp
@@ -21,9 +21,10 @@
 namespace xdp {
 
   OpenCLAPICall::OpenCLAPICall(uint64_t s_id, double ts, uint64_t /*f_id*/,
-                               uint64_t name, uint64_t q)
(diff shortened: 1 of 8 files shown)


IshitaGhosh (approved):
Changes look good

PR #5655, opened 2021-08-19
Fixing Iops test issue for U.2

chvamshi-xilinx:
Currently IOPS test case is using one global data structure in multiple threads with an assumption that all kernels would have same name/index etc.
This assumption is valid if there are no errors while extracting the index from kernel.
We are updating the global data structure when there are no errors.

### DIFF
--- a/tests/validate/xcl_iops_test/src/xcl_api_iops.cpp
+++ b/tests/validate/xcl_iops_test/src/xcl_api_iops.cpp
@@ -43,7 +43,7 @@ struct krnl_info {
 
 bool verbose = false;
 barrier barrier;
-thread_local struct krnl_info krnl = {"hello", false};
+struct krnl_info krnl = {"hello", false};
 
 static void usage(char *prog)
 {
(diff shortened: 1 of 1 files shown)


tests/validate/xcl_iops_test/src/xcl_api_iops.cpp
```diff
@@ -197,13 +197,18 @@ void runTestThread(arg_t &arg)
...
+        krnl.cu_idx = cu_idx;
+    else {
         // hello:hello_cu0 is U2 shell special
         cu_name = krnl.name + ":" + krnl.name + "_cu0";
-        krnl.cu_idx = xclIPName2Index(handle, cu_name.c_str());
-        if (krnl.cu_idx < 0)
+        cu_idx = xclIPName2Index(handle, cu_name.c_str());
+        if (cu_idx < 0)
             throw std::runtime_error(cu_name + " not found");
+	krnl.cu_idx = cu_idx;
```
mamin506, line 211: Since krnl.cu_idx would get updated success path.
Below code would be neater.

int cu_idx = xclIPName2Index();
if (cu_idx < 0) {
...
if (something_wrong)
    throw;
}
// looks good
krnl.cu_idx = cu_idx;
...

mamin506 (changes requested):
Test with U200 and U50, looks good.

tests/validate/xcl_iops_test/src/xcl_api_iops.cpp
chvamshi-xilinx: Thanks MIn for the suggestion. Made suggested changes.

mamin506 (approved):

PR #5656, opened 2021-08-19
CR 1108275 : Show AXI Memory Monitor section in Device Trace only for trace enabled floating AIMs 

IshitaGhosh:


jvillarre (approved):
Changes look good

PR #5659, opened 2021-08-20
Promote p2p to a first class command

AShivangi:
```
$ xbutil configure -h

COMMAND: configure

DESCRIPTION: Device and host configuration.

USAGE: xbutil configure [-h] --[ host-mem | p2p ] [commandArgs]

OPTIONS:
  -h, --help         - Help to use this sub-command
  --host-mem         - Controls host-mem functionality
  --p2p              - Controls P2P functionality

GLOBAL OPTIONS:
```

### DIFF
--- a/src/runtime_src/core/tools/xbutil2/SubCmdAdvanced.cpp
+++ b/src/runtime_src/core/tools/xbutil2/SubCmdAdvanced.cpp
@@ -20,7 +20,6 @@
 #include "OO_MemRead.h"
 #include "OO_AieRegRead.h"
 #include "OO_MemWrite.h"
-#include "OO_P2P.h"
 #include "XBReport.h"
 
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/tools/xbutil2/SubCmdAdvanced.cpp
```diff
@@ -20,7 +20,6 @@
 #include "OO_MemRead.h"
 #include "OO_AieRegRead.h"
 #include "OO_MemWrite.h"
-#include "OO_P2P.h"
```
rozumx, line 23: I do like it when we remove code (even if it is being moved)

rozumx (approved):
Looks good.  Thanks for move this functionality.

PR #5663, opened 2021-08-23
CR-1108702 xocl should skip checking interface uuid for xclbins if th…

xuhz:
…ey are lite
To solve u30 v1 compatibility issue,  don't check interface uuid if the xclbin is lite (strip bitstream and ps kernel) in xocl
1. there is behavior change to other alveo cards
2. xclmgmt changes will be in later (not backport to u30 branch, see CR-1108703)


houlz0507 (approved):

xdavidz (approved):
Would it be better to move those dup code into a MACRO like:
#define LITE_XCLBIN(!(BITSTREAM || SOFTKERNEL))?

xuhz:
> Would it be better to move those dup code into a MACRO like:
> #define LITE_XCLBIN(!(BITSTREAM || SOFTKERNEL))?

It would be better to use same functions to get and check all headers in the 2 files, but these 2 files all have its own function defined. so I have to use the same function the original code is using.

PR #5672, opened 2021-08-24
CR-1081361 and CR-1090664

AShivangi:
- CR-1081361 Please provide a valid range for no dma host mem size
- CR-1090664 xbmgmt2 help doesn't show how to update SC

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
@@ -616,9 +616,9 @@ find_flash_image_paths(const std::vector<std::string> image_list)
 
 SubCmdProgram::SubCmdProgram(bool _isHidden, bool _isDepricated, bool _isPreliminary)
     : SubCmd("program", 
-             "Update device and/or Satallite Controler (SC) firmware image for a given device")
+             "Update device and/or Satellite Controller (SC) firmware image for a given device")
 {
-  const std::string longDescription = "Updates the flash image for the device and/or the Satallite Controller (SC) firmware image for a given device.";
+  const std::string longDescription = "Updates the flash image for the device and/or the Satellite Controller (SC) firmware image for a given device.";
   setLongDescription(longDescription);
   setExampleSyntax("");
   setIsHidden(_isHidden);
@@ -664,7 +664,7 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
                                                                          "  SC    - Satellite controller"*/)
     ("user,u", boost::program_options::value<decltype(xclbin)>(&xclbin), "The xclbin to be loaded.  Valid values:\n"
                                                                       "  Name (and path) of the xclbin.")
-    ("revert-to-golden", boost::program_options::bool_switch(&revertToGolden), "Resets the FPGA PROM back to the factory image. Note: The Satellite Control (MSP432) will not be reverted for a golden image does not exist.")
+    ("revert-to-golden", boost::program_options::bool_switch(&revertToGolden), "Resets the FPGA PROM back to the factory image. Note: The Satellite Controller (MSP432) will not be reverted for a golden image does not exist.")
     ("help,h", boost::program_options::bool_switch(&help), "Help to use this sub-command")
   ;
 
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/tools/xbmgmt2/ReportPlatform.cpp
```diff
@@ -301,22 +302,26 @@ ReportPlatform::writeReport( const xrt_core::device* /*_pDevice*/,
                              const std::vector<std::string>& /*_elementsFilter*/,
                              std::ostream & _output) const
 {
+  auto pretty_print =[](std::string str) {
```
rozumx, line 305: Please change the name.....too generic.  Especially since `pretty_print` usually implies not changing the data being returned only the format.

Some suggestions:
```
string_or_NA
if_empty_NA
```

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -616,9 +616,9 @@ find_flash_image_paths(const std::vector<std::string> image_list)
 
 SubCmdProgram::SubCmdProgram(bool _isHidden, bool _isDepricated, bool _isPreliminary)
     : SubCmd("program", 
-             "Update device and/or Satallite Controler (SC) firmware image for a given device")
+             "Update images for a given device")
 {
-  const std::string longDescription = "Updates the flash image for the device and/or the Satallite Controller (SC) firmware image for a given device.";
+  const std::string longDescription = "Updates the images for a given device.";
```
rozumx, line 621:
```suggestion
  const std::string longDescription = "Updates the image(s) for a given device.";
```

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -658,13 +658,13 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
...
+    ("base,b", boost::program_options::value<decltype(update)>(&update)->implicit_value("all"), "Update the persistent images and/or the Satellite controller (SC) firmware image."/*  Value values:\n"
                                                                          "  ALL   - All images will be updated"
                                                                          "  FLASH - Flash image\n"
                                                                          "  SC    - Satellite controller"*/)
     ("user,u", boost::program_options::value<decltype(xclbin)>(&xclbin), "The xclbin to be loaded.  Valid values:\n"
                                                                       "  Name (and path) of the xclbin.")
-    ("revert-to-golden", boost::program_options::bool_switch(&revertToGolden), "Resets the FPGA PROM back to the factory image. Note: The Satellite Control (MSP432) will not be reverted for a golden image does not exist.")
+    ("revert-to-golden", boost::program_options::bool_switch(&revertToGolden), "Resets the FPGA PROM back to the factory image. Note: The Satellite Controller (MSP432) will not be reverted for a golden image does not exist.")
```
rozumx, line 667: F.Y.I. - It is my understanding that newer Satellite Controllers may not involved a MSP432 chip.

```suggestion
    ("revert-to-golden", boost::program_options::bool_switch(&revertToGolden), "Resets the FPGA PROM back to the factory image. Note: The Satellite Controller will not be reverted for a golden image does not exist.")
```

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
```diff
@@ -117,7 +117,7 @@ OO_HostMem::execute(const SubCmdOptions& _options) const
 
     // Exit if ENABLE action is specified but size is not
     if(enable && size == 0)
-      throw xrt_core::error(std::errc::invalid_argument, "Please specify a non-zero memory size.");
+      throw xrt_core::error(std::errc::invalid_argument, "Please specify a non-zero memory size between 4M and 1G.");
```
rozumx, line 120: Hmm..are there other check regarding the size of the memory.

In other words, what happens if one sets it to 1K, will there be an error produced?

rozumx (changes requested):

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
AShivangi: Yes, but that level of error handling is done by the driver. I had a chat with Chien-Wei about this earlier.

PR #5697, opened 2021-08-27
Update success message for p2p

AShivangi:
- Update return codes
- Print success message when p2p is enabled or disabled
```
$ sudo xbutil configure --p2p enable -d 65:00
Please WARM reboot the machine to enable P2P now.
$ sudo xbutil configure --p2p disable -d 65:00
Please WARM reboot the machine to disable P2P now.
```

### DIFF
--- a/src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
+++ b/src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
@@ -62,6 +62,19 @@ string2action(std::string str)
     throw xrt_core::generic_error(EINVAL, "Invalid p2p action '" + str + "'");
 }
 
+std::string
+print_success_msg(action_type action)
+{
+  if(action == action_type::enable)
+    return "Please WARM reboot the machine to enable P2P now.";
+  if(action == action_type::disable)
+    return "Please WARM reboot the machine to disable P2P now.";
+  if(action == action_type::validate)
+    return "P2P validated successfully.";
+  
+  throw xrt_core::error(std::errc::operation_canceled);
+}
+
 static auto
 p2p_config(xrt_core::device* device)
 {
@@ -284,7 +297,7 @@ OO_P2P::execute(const SubCmdOptions& _options) const
   // Honor help option first
   if (std::find(_options.begin(), _options.end(), "--help") != _options.end()) {
     printHelp();
-    return;
+    throw xrt_core::error(std::errc::operation_canceled);
   }
 
   // Parse sub-command ...
@@ -303,12 +316,13 @@ OO_P2P::execute(const SubCmdOptions& _options) const
 
   if (m_help) {
     printHelp();
-    return;
+    throw xrt_core::error(std::errc::operation_canceled);
   }
 
   // Validate the correct action value is used
+  action_type action;
   try {
-     string2action(m_action);
+    action = string2action(m_action);
   } catch (const xrt_core::generic_error &e) {
     std::cerr << boost::format("ERROR: %s\n") % e.what();
     throw xrt_core::error(std::errc::operation_canceled);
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
```diff
@@ -284,7 +297,7 @@ OO_P2P::execute(const SubCmdOptions& _options) const
   // Honor help option first
   if (std::find(_options.begin(), _options.end(), "--help") != _options.end()) {
     printHelp();
-    return;
+    throw xrt_core::error(std::errc::operation_canceled);
```
rozumx, line 300: There is no error here.

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
```diff
@@ -303,12 +316,13 @@ OO_P2P::execute(const SubCmdOptions& _options) const
 
   if (m_help) {
     printHelp();
-    return;
+    throw xrt_core::error(std::errc::operation_canceled);
```
rozumx, line 319: There is no error here.

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
```diff
@@ -62,6 +62,19 @@ string2action(std::string str)
     throw xrt_core::generic_error(EINVAL, "Invalid p2p action '" + str + "'");
 }
 
+std::string
+print_success_msg(action_type action)
```
rozumx, line 66: This is not a case where we should be encapsulating print messages.

In addition, these types of methods should NOT be throwing an exception.

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
```diff
@@ -303,12 +316,13 @@ OO_P2P::execute(const SubCmdOptions& _options) const
...
     printHelp();
-    return;
+    throw xrt_core::error(std::errc::operation_canceled);
   }
 
   // Validate the correct action value is used
+  action_type action;
   try {
-     string2action(m_action);
+    action = string2action(m_action);
```
rozumx, line 325: After this operation, we know that the action value is good.  No need to check it again later in the code.

src/runtime_src/core/tools/xbutil2/OO_P2P.cpp
```diff
@@ -349,6 +363,12 @@ OO_P2P::execute(const SubCmdOptions& _options) const
...
 
-  for (auto& device : deviceCollection)
-    p2p(device.get(), string2action(m_action), false);
+  try {
+    p2p(deviceCollection[0].get(), action, XBU::getForce());
+  } catch (const xrt_core::system_error& ex) {
+    std::cerr << "ERROR: " << ex.what() << std::endl;
+    throw xrt_core::error(std::errc::operation_canceled); 
+  }
+  std::cout << print_success_msg(action) << std::endl;
```
rozumx, line 372: Please refactor to something cleaner.  Something like:

```
// Help the user know what their next steps are
switch (action) {
   case action_type::enable:
    std::cout <<  "Please WARM reboot the machine to enable P2P now.\n";
    break;
   
  case action_type::disable:
    std::cout << "Please WARM reboot the machine to disable P2P now.\n";
    break;

  case action_type::validate:
    std::cout << "P2P validated successfully.\n";
}
```
Note: Since we are using modern c++ compilers, if in the future the action_type enumeration should get another value, the compilers will detect that not all switch values are covered and the compilation will stop.

rozumx (changes requested):
Changes requested.

rozumx (approved):
Approved

PR #5698, opened 2021-08-27
use dma tiles for memory profiling

hackwa:


dayeh-xilinx:
retest this please. test was cancelled to give way to urgent u30 build.

PR #5702, opened 2021-08-27
Introduce SchemaVersion enum in get_info() template

AShivangi:
- Share `SchemaVersion` enum between `xrt_device.h` and `Report.h`
- Updated documentation for github.io
- Fixed test cases


rozumx (changes requested):
Changes requested.  When they have been made I will re-review this submission.

PR #5703, opened 2021-08-27
allow xgq to attach and continue from where it's left

maxzhen:


xdavidz (approved):
Tested on XGQ host driver via VMR(Versal Management on R5).

PR #5705, opened 2021-08-28
CR-1109011: load xclbin errors in sysfs

sarab96:
Add support for load xclbin errors into sysfs node


src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_ioctl.c
```diff
@@ -354,10 +355,18 @@ static int xocl_preserve_mem(struct xocl_drm *drm_p, struct mem_topology *new_to
 
 static bool xocl_xclbin_in_use(struct xocl_dev *xdev)
```
maxzhen, line 356: xocl_xclbin_in_use() is a simple function that performs the check, the side effect of inserting an error into some global list is not expected. please move them to outside of this function.

maxzhen (changes requested):

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_ioctl.c
sarab96: updated

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_ioctl.c
```diff
@@ -377,20 +378,36 @@ xocl_read_axlf_helper(struct xocl_drm *drm_p, struct drm_xocl_axlf *axlf_ptr)
...
 
+
+	err_last.pid = 0;
+	err_last.ts = 0; //TODO timestamp
 	if (!xocl_is_unified(xdev)) {
 		userpf_err(xdev, "XOCL: not unified Shell\n");
+		err_last.err_code = XRT_ERROR_CODE_BUILD(XRT_ERROR_NUM_XCLBIN_INCOMPATIBLE, 
+			XRT_ERROR_DRIVER_XOCL, XRT_ERROR_SEVERITY_CRITICAL, 
+			XRT_ERROR_MODULE_XCLBIN, XRT_ERROR_CLASS_SYSTEM);
+		xocl_insert_error_record(&xdev->core, &err_last);
```
maxzhen, line 391: I think this is for a not supported shell, not for incompatible xclbin. @houlz0507 can point you to a more suitable location where xclbin compatibility is checked.

houlz0507: all the xclbin checks on xocl side are in xocl_read_axlf_helper().

larry9523 (commented):
So we are using the error handling frame work which was built to handle asynchronous error to handle synchronous error too? 

If so, we may want to update our error reporting conf page as well so that we have a clear document about what synchronous error are reported through xrtErrorGetLast

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_ioctl.c
sarab96: updated

sarab96:
Yes I will ask for that doc/info to be updated

> So we are using the error handling frame work which was built to handle asynchronous error to handle synchronous error too?
> 
> If so, we may want to update our error reporting conf page as well so that we have a clear document about what synchronous error are reported through xrtErrorGetLast

maxzhen:
> 
> 
> Yes I will ask for that doc/info to be updated
> 
> > So we are using the error handling frame work which was built to handle asynchronous error to handle synchronous error too?
> > If so, we may want to update our error reporting conf page as well so that we have a clear document about what synchronous error are reported through xrtErrorGetLast

I'm not sure if this is desired. We have return code, do we need a 2nd way to get these errors? If this is not in Larry's original design, we need to discuss more. I've marked this PR as do-not-merge until we have a clear design first.

sarab96:
Will add other async erros

PR #5711, opened 2021-08-30
CR-1101816 The transfer of gm2aie/aie2gm will longer when I allocate

larry9523:
extra unused space with GM

XRT GMIO implementation was calling XAie_DmaSetAddrLen to config BD, which requires passing in a virtual address. To obtain a virtual address from a DMA buf, we need call mmap/munmap in the critical path.
AIE driver has a new API XAie_DmaSetAddrOffsetLen where we only need to pass in offset of the DMA buf.

This PR is to make use of the new API to avoid mmap/munmap in the critical path.

@jerryhsu-xilinx please take a look at the code changes, especially the impact to AIE SIM.


dayeh-xilinx:
retest this please. Failed on u250 test. Try again.

larry9523:
@jerryhsu-xilinx has reviewed the change and another test passes on GMIO tutorial perf_gmio design

PR #5716, opened 2021-08-31
Addressed windows debug build compiler warnings

rozumx:
The Windows Debug Build of XBMain.c produced a C4702 compiler warning that indicated code was unreachable.  Further examination of the code suggests that that issue is associated with the MS compiler and not the actual code. 

Never the less, the code was refactored to remove the `break` statement (the code that was being complained about) from the `for` loop.

### DIFF
--- a/src/runtime_src/core/tools/common/XBMain.cpp
+++ b/src/runtime_src/core/tools/common/XBMain.cpp
@@ -155,10 +155,11 @@ void  main_(int argc, char** argv,
     sDevice.clear();
     boost::property_tree::ptree available_devices = XBU::get_available_devices(isUserDomain);
 
+    // DRC: Are there any devices
     if (available_devices.empty()) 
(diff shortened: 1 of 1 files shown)


AShivangi (approved):

src/runtime_src/core/tools/common/XBMain.cpp
```diff
@@ -171,11 +172,10 @@ void  main_(int argc, char** argv,
...
     }
 
-    // We have a match, get first entry
-    for (auto &kd : available_devices) {
+    // We have only 1 item in the array, get it
+    for (const auto &kd : available_devices) 
       sDevice = kd.second.get<std::string>("bdf"); // Exit after the first item
-      break;
-    }
+
```
AShivangi, line 178: instead of for loop can we not do:
`available_devices[0].second.get<std::string>("bdf");`?

PR #5717, opened 2021-09-01
Fix for soft kernel IOCTL cu_idx check

jeffli-xilinx:
CR-1109454

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_sk.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_sk.c
@@ -97,6 +97,11 @@ zocl_sk_create_ioctl(struct drm_device *dev, void *data, struct drm_file *filp)
 	struct drm_zocl_bo *bo;
 	uint32_t cu_idx = args->cu_idx;
 
+	if (cu_idx >= MAX_SOFT_KERNEL) {
+		DRM_ERROR("Fail to create soft kernel: CU index %d > %d.\n",
+			  cu_idx,MAX_SOFT_KERNEL);
+		return -EINVAL;
+	}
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/edge/drm/zocl/zocl_sk.c
```diff
@@ -107,6 +107,13 @@ zocl_sk_create_ioctl(struct drm_device *dev, void *data, struct drm_file *filp)
 
 	mutex_lock(&sk->sk_lock);
 
+	if (cu_idx >= 128) {
```
larry9523, line 110: 1. Use MAX_SOFT_KERNEL instead of 128
2. Maybe we should move this check to the first validation in this function so that we don't need to worry about BO refcount and mutex protect sk_cu

larry9523 (changes requested):
Thanks for fixing this. Two minor nits.

src/runtime_src/core/edge/drm/zocl/zocl_sk.c
```diff
@@ -97,6 +97,11 @@ zocl_sk_create_ioctl(struct drm_device *dev, void *data, struct drm_file *filp)
 	struct drm_zocl_bo *bo;
 	uint32_t cu_idx = args->cu_idx;
 
+	if (cu_idx >= MAX_SOFT_KERNEL) {
+		DRM_ERROR("Fail to create soft kernel: CU index %d > %d.\n",
+			  cu_idx,MAX_SOFT_KERNEL);
```
larry9523, line 102: Need a ',' between cu_idx and MAX_SOFT_KERNEL

larry9523 (approved):
Looks good. Just a c-style nit.

PR #5727, opened 2021-09-03
CR-1109858 and CR-1090667

AShivangi:
- CR-1109858 xbmgmt examine - command not working in 2021.2
- CR-1090667 xbmgmt2 program --base displays no information before proceed


rozumx (approved):
Approved

PR #5730, opened 2021-09-05
CR-1108493 : Fix for multiple zocl load/unload

saifuddin-xilinx:
CR-1108493 : dfx-manager fails to load overlay after ~32 iterations . Linux OS hangs

**Issue:**
Unloading and loading zocl driver repeatedly over 45 times causes the Linux operating system to hang . 

**Root Cause:**
To get the pcap driver we are searching from device tree using "of_find_node_by_name" Kernel function. 
The params we are passing as "of_root" which is causing the issue. As per petaLinux kernel none of the module/driver used this param. Instead used NULL as a param to traverse all the node in the device tree. 

**Solution:**
Update the param to "NULL" in of_find_node_by_name function resolved the issue.  

**Test:**
```
for i in {1..100} 
do
  echo "Welcome $i times";
  rmmod zocl;
  insmod zocl.ko; 
done
```

Tested this solution on following boards:
zcu102
zcu102_dfx
vck190

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_bo.c
@@ -1032,11 +1032,14 @@ static bool check_for_reserved_memory(uint64_t start_addr, size_t size)
 			 * in this reserved memory region
 			 */
 			if (start_addr >= res_mem.start &&
-					size <= resource_size(&res_mem))
+					size <= resource_size(&res_mem)) {
+				of_node_put(mem_np);
 				return true;
+			}
 		}
 	}
 
+	of_node_put(mem_np);
 	return false;
 }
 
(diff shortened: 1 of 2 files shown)


xilinxgitops:
retest this please

chvamshi-xilinx (approved):
Looks good to me. Please verify DFx platforms and check whether you are getting the right node or not.

larry9523 (changes requested):
The fix looks good. But there is a missing part. After we found a match, we need to call of_node_put to decrease the reference count of the node we just got.

P.S. the other way to fix it is to call of_nod_get(of_root) before calling of_find_node_by_name() and keep using of_root instead of NULL. But the "NULL" solution seems cleaner.

larry9523 (approved):

PR #5732, opened 2021-09-07
CR-1110133 fixing xbutil examine issue for edge

chvamshi-xilinx:
xbutil examine is failing to get an instance as corresponding query is not defined in edge driver. Fixed in user space code.
xbutil examine -d 0 is crashing in zocl while holding the xclbin. uuid_equal is failing when we pass null uuid. Fixed zocl.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
@@ -1023,7 +1023,7 @@ zocl_xclbin_hold(struct drm_zocl_dev *zdev, const xuid_t *id)
 
 	// check whether uuid is null or not
 	if (uuid_is_null(id)) {
-		DRM_WARN("NULL uuid to hold.");
+		DRM_WARN("NULL uuid to hold\n");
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -1021,7 +1021,11 @@ zocl_xclbin_hold(struct drm_zocl_dev *zdev, const xuid_t *id)
 {
 	xuid_t *xclbin_id = (xuid_t *)zocl_xclbin_get_uuid(zdev);
 
-	WARN_ON(uuid_is_null(id));
+	// check whether uuid is null or not
```
mamin506, line 1024: code is clear. No need to comment at all.

mamin506 (approved):

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
chvamshi-xilinx: Thanks Min. Will remove the comment in next PR.

PR #5734, opened 2021-09-07
reverting uuid null initialization 

chvamshi-xilinx:


mamin506 (changes requested):
1. What if zx_uuid is NULL? In linux, uuid_is_null(zx_uuid) is equal to "memcmp(zx_uuid, &uuid_null, sizeof(uuid_t)) == 0;"

2. If zx_uuid is not NULL but uuid_is_null(zx_uuid) return true, then zdev->zdev_xclbin->zx_uuid would be updated but the old zx_uuid is not being free (memory leak right?).

3. Could you describe what we are trying to fix by this PR in the descriptions?

chvamshi-xilinx:
> 1. In linux, uuid_is_null(zx_uuid) is equal to "memcmp(zx_uuid, &uuid_null, sizeof(uuid_t)) == 0;" What if zx_uuid is NULL? No null pointer issue in your test?
> 2. If zx_uuid is not NULL but uuid_is_null(zx_uuid) returns true, then zdev->zdev_xclbin->zx_uuid would be updated but the old zx_uuid is not being free (memory leak right?).
> 3. Could you describe what we are trying to fix by this PR in the descriptions?

Hi @mamin506 , we are always initializing zx_uuid to "uuid_null" it is never going to be NULL.
zx_uuid is initilized in uuid_null, so wherever we were doing NULL checks earlier, we are doing is_null checks now. This is one missing null check where i have added.

chvamshi-xilinx:
@mamin506 , Reverted the other change also and merging as previous PR depends on this changes
We can continue discussing here..

xilinxgitops:
retest this please

xdavidz:
@chvamshi-xilinx what is the major issue of using NULL instead of uuid_null? The logic was if uuid_null, then we can re-download the xclbin. I guess we should also free existing cached xclbin as well, there might be some existing memory leaks?

Basically, either way is ok, the uuid apis also have uuid_clear and uuid_compare, a typical utility should use those APIs rather than NULL. For our special case, since it is a pointer, we can use NULL.

For *uuid cases, we should use uuid_is_null to check if uuid has been cleared. Otherwise, we create confusion for other developers what is the purpose of using NULL instead of uuid_is_null.

mamin506:
> @mamin506 , Reverted the other change also and merging as previous PR depends on this changes
> We can continue discussing here..

Could you ref the previous PR so that we are on the same page?

chvamshi-xilinx:
@xdavidz , Actually xbutil is opening a context without actually loading a xclbin. Kernel is crashing in that case as uuid_eqaul is comparing with NULL object. We can return an error when Opening a context. I just thought inillizing with uuid would make things simple :-) but looks like there are other places also where it is checking NULL and vfree cannot be called when we initialized with uuid_null. Thatswhy reverted the change.
@mamin506 , i am talking about https://github.com/Xilinx/XRT/pull/5732

xdavidz:
I'd prefer a sophisticate thought of handling coding issues and using APIs, if uuid_is_null is preferred, we should enforce using uuid_is_null everywhere and add comments for specific handling. 
But in this case, it seems that we are getting rid of uuid_is_null and switching to NULL, which confused me what is the original propose. For example, if I want to add some code in this area, should I use NULL or uuid_is_null?

I was not involved in #5732, but I will take a closer look.

chvamshi-xilinx:
> I'd prefer a sophisticate thought of handling coding issues and using APIs, if uuid_is_null is preferred, we should enforce using uuid_is_null everywhere and add comments for specific handling.
> But in this case, it seems that we are getting rid of uuid_is_null and switching to NULL, which confused me what is the original propose. For example, if I want to add some code in this area, should I use NULL or uuid_is_null?
> 
> I was not involved in #5732, but I will take a closer look.

Agress. FIled a new PR-https://github.com/Xilinx/XRT/pull/5749 which fixes these issues. Added you as reviewer.

PR #5735, opened 2021-09-07
configure md0 and md1 for floating pt execeptions

hackwa:


src/runtime_src/xdp/profile/plugin/aie_profile/aie_plugin.cpp
```diff
@@ -505,6 +505,22 @@ namespace xdp {
           //else if (startEvents.at(i) == XAIE_EVENT_GROUP_ERRORS_MEM)
           //  XAie_EventGroupControl(aieDevInst, loc, mod, startEvents.at(i), GROUP_ERROR_MASK);
 
+          // Enable floating point exception events
```
sonals, line 508: AIEProfilingPlugin::setMetrics() is about 250 lines long. Please break up this function into logical pieces.

sonals (changes requested):

hackwa:
This change is no longer required. I'll close this PR for now. @sonals will make your suggested change in a new PR.

PR #5743, opened 2021-09-08
CR-1109983: Considering default domain as 0000 in xrt c++ apis

chvamshi-xilinx:
This issue fixes the XRT Native APIs regression. User can create xrt_device object with only bus , device and function if domain is 0000.


stsoe (approved):
Thank you!

PR #5745, opened 2021-09-08
CR-1110328: HAL APIs reported the same in profile summary and trace

jvillarre:
Changing the name of the HAL functions logged to match the original names.  The names being logged appeared on the trace but were not found when looking for summary information due to discrepancies.  This pull request changes the HAL profiling plugin to use the original names and match what is looked for in the summary writer.


IshitaGhosh (approved):
Change looks good.

PR #5749, opened 2021-09-09
CR-1110133: fixing multiple null checks around uuid usage

chvamshi-xilinx:
This change fixes the multiple uuid null checks.
1) xclbin uuid sysfs node comes empty when no xclbin is loaded. We were getting some uninitialized value earlier.
2) lock/unlock bistreams should fail whenever lock device called without loading xclbin
3) uuid constructor with string should initialize uuid with clear state.
4) don't open the context when xclbin is null

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_sysfs.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_sysfs.c
@@ -72,7 +72,7 @@ static ssize_t xclbinid_show(struct device *dev,
 	struct drm_zocl_dev *zdev = dev_get_drvdata(dev);
 	ssize_t size;
 
-	if (!zdev)
+	if (!zdev || !zdev->zdev_xclbin || !zdev->zdev_xclbin->zx_uuid)
 		return 0;
 
 	size = sprintf(buf, "%pUb\n", zdev->zdev_xclbin->zx_uuid);
--- a/src/runtime_src/core/include/xrt/xrt_uuid.h
+++ b/src/runtime_src/core/include/xrt/xrt_uuid.h
@@ -71,12 +71,17 @@ public:
    * uuid() - Construct uuid from a string representaition
    *
    * @param uuid_str
-   *  A string formatted as a uuid string 
+   *  A string formatted as a uuid string
    *
    * A uuid string is 36 bytes with '-' at 8, 13, 18, and 23
    */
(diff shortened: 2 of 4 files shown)


src/runtime_src/core/include/xrt/xrt_uuid.h
```diff
@@ -71,12 +71,17 @@ class uuid
...
    *
    * A uuid string is 36 bytes with '-' at 8, 13, 18, and 23
    */
   explicit uuid(const std::string& uuid_str)
   {
+    if(uuid_str.empty())
+    {
+      uuid_clear(m_uuid);
+      return;
+    }
```
stsoe, line 84: Per convention space between if and brace, and brace on same line as condition.

stsoe (approved):
Minor style comment

dayeh-xilinx:
retest this please. build failed for infra error.

src/runtime_src/core/edge/drm/zocl/zocl_sysfs.c
```diff
@@ -72,7 +72,7 @@ static ssize_t xclbinid_show(struct device *dev,
 	struct drm_zocl_dev *zdev = dev_get_drvdata(dev);
 	ssize_t size;
 
-	if (!zdev)
+	if (!zdev || !zdev->zdev_xclbin || !zdev->zdev_xclbin->zx_uuid)
 		return 0;
```
mamin506, line 76: If we print uuid_null, we don't really need any user space code change. Right?

Like this,
if (!zdev || !zdev->zdev_xclbin || !zdev->zdev_xclbin->zx_uuid) {
    size = sprintf(buf, "%pUb\n", &uuid_null);
} else {
    size = sprintf(buf, "%pUb\n", zdev->zdev_xclbin->zx_uuid);
}

mamin506 (approved):
The driver changes look good to me. If you don't want to make user space code change, please use my suggested code in zocl_sysfs.c.

src/runtime_src/core/edge/drm/zocl/zocl_sysfs.c
chvamshi-xilinx: Thats a good suggestion, but we still need to make user space changes to check whether uuid is null or not to call xclOpenConext. xrt_uuid constructor change is also required as it gets used in multiple places.

xdavidz (approved):

PR #5751, opened 2021-09-09
CR-1108384 OpenCL message for ap_ctrl_none requires improvement

stsoe:
Simplify message to say that AP_CTRL_NONE is only supported by xrt::ip.


dayeh-xilinx:
retest this please. build failed for infra error.

PR #5752, opened 2021-09-09
xlcbin_parser: fix compilation error on GCC 11.2.0

mclayton1000:
Without the missing headers the following error arises with GCC 11.2.0:

XRT/src/runtime_src/core/common/xclbin_parser.h:32:43: error:  
numeric_limits’ is not a member of ‘std’
   32 |   static constexpr size_t no_index { std::numeric_limits<size_t>::max() };
      |                                           ^~~~~~~~~~~~~~

### DIFF
--- a/src/runtime_src/core/common/xclbin_parser.h
+++ b/src/runtime_src/core/common/xclbin_parser.h
@@ -21,6 +21,8 @@
 #include "xclbin.h"
 #include <string>
 #include <vector>
+#include <stdexcept>
+#include <limits>
 
 namespace xrt_core { namespace xclbin {
 


stsoe:
ok to test

stsoe:
retest this please - infra

keryell:
The CI could run https://github.com/include-what-you-use/include-what-you-use on a regular basis.

src/runtime_src/core/common/xclbin_parser.h
```diff
@@ -21,6 +21,8 @@
 #include "xclbin.h"
 #include <string>
 #include <vector>
+#include <stdexcept>
```
keryell, line 24: Please sort the `#include`

keryell (changes requested):

src/runtime_src/core/common/xclbin_parser.h
mclayton1000: Hi @keryell ,

you should I re-submit the patch with alphabetically sorted "include" ?

keryell: If you do not mind, you can contribute even more into code cleaning. \o/

mclayton1000: Ok, no problem. Should I invalidate this commit before? How?

mclayton1000:
Patch resubmitted with suggestions from @keryell

src/runtime_src/core/common/xclbin_parser.h
keryell: > Patch resubmitted with suggestions from @keryell

I am unsure about what you meanwith *patch* here.
With `git` you just add a new commit with the new content and you push on your published branch.

mclayton1000: A commit which contains a patch...

PR #5753, opened 2021-09-09
CR-1109237 Permission denied: "C:\Xilinx" on Linux

stsoe:
Scanning for installed XSAs is iterating a list of possible firmware
directories, which includes C:\Xilinx even on Linux.  When current
directory is only accessible by current user, any sudo command will
fail to locate C:\Xilinx because it is interpreted as a relative path
on Linux.

Catch exception.

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/flash/firmware_image.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/flash/firmware_image.cpp
@@ -14,19 +14,9 @@
  * under the License.
  */
 
-
-#include <iostream>
-#include <fstream>
-#include <algorithm>
-#include <climits>
-#include <iomanip>
-#include <memory>
-#include <regex>
-#include <sstream>
-#include <stdint.h>
-#include <cstring>
-#include <vector>
-#include <locale>
+#include "firmware_image.h"
+#include "core/common/utils.h"
+#include "core/include/xclbin.h"
 
 // 3rd Party Library - Include Files
 #include <boost/format.hpp>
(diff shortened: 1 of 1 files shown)


AShivangi (commented):
If Linux fails to locate "C:\\Xilinx" it should skip right?
`if (!boost::filesystem::exists(root) || !boost::filesystem::is_directory(root))` should be able to handle that case

stsoe:
> If Linux fails to locate "C:\Xilinx" it should skip right?
> `if (!boost::filesystem::exists(root) || !boost::filesystem::is_directory(root))` should be able to handle that case
No, because of the CWD permissions Linux can't even check the current directory for the existence of the relative dir.

AShivangi (approved):
Oh I see. I'm glad you were able to figure out the root cause of this issue!

AShivangi (approved):

rozumx (approved):
**_Approve with Reservations_**

Where we stand in the release warrants a low risk fix for this issue, which is what this pull request reflects.

After an off-line discussion with Soren, we do believe the following code refactoring code changes should be made post 2021.2:

1. The directory locations should be determine via the shim layer and not in firmware_image.cpp.  A possible function name could be: `get_firmware_dirs()`.
2. We need to remove the #ifdef directive.  The code should NOT have OS compiler directives.
3. `exist()` and `is_directory()` functions should be wrapped in a try/catch block.  To catch future issues like this.

keryell:
> We need to remove the #ifdef directive. The code should NOT have OS compiler directives.

Does it mean looking blindly for "C:\\Xilinx" on Linux?
It would be interesting to build an attack with some virused firmware in a local directory made with `mkdir "C:\\Xilinx"` and wait for some admin to just execute a Xilinx command to get `sudo` access...

PR #5757, opened 2021-09-09
CR-1110454 ASTeR TC19.15 - validate can intermittently crash (BUG: unable to handle kernel paging request / command_queue_poll+0x15e/0x250 [xocl])

mamin506:
Somehow, the configure command some time needs more than 3 seconds to complete. But we should consume "kds->comp" after abort, otherwise the next time wait_for_completion_timeout() will return immediately.


dayeh-xilinx:
retest this please. Build failed because I was testing the u280 host.

chienwei-lan (approved):

PR #5758, opened 2021-09-10
xlcbin_parser: fix compilation error on GCC 11.2.0

mclayton1000:
Without the missing headers the following error arises with GCC 11.2.0:

XRT/src/runtime_src/core/common/xclbin_parser.h:32:43: error:
numeric_limits’ is not a member of ‘std’
32 | static constexpr size_t no_index { std::numeric_limits<size_t>::max() };
| ^~~~~~~~~~~~~~


keryell (approved):
Thanks for working towards a better and more modern world.

stsoe:
ok to test

PR #5759, opened 2021-09-10
Add epoch to package versions for debian and rpm

stsoe:
This is per suggestion by Canonical to differentiate Xilinx versioning
from Canonical's.  For consistency, RPMs will also add epoch.


dayeh-xilinx:
retest this please. Just to double check if it is an infra error.

stsoe:
Hi @dayeh-xilinx .  This pull request will not work.   We found that adding epoch to the version breaks other packages that depend on a maximum version of xrt, e.g., cmc which as (xrt < 3), but 1:2.x.x. is actually greater than 3 so epoch breaks that requirement.  The PR will be closed once I have confirmed there is no way around this.

stsoe:
When there are dependencies on versions there is little that can be done but re-building those packages to change the versions of the dependencies (i.e. <<1:3 instead of <<3).  It seems epoch cannot really be used.

PR #5763, opened 2021-09-10
CR-1110676: Disable trace when only profiling counters enabled in xrt.ini 

jvillarre:
The memory monitor table in profile summary exposed a problem where trace was being enabled on designs with TS2MM if only counters were enabled.  This change makes sure that trace is not initialized in cases where the only device information we need from profiling is counter values.


IshitaGhosh (commented):
Change looks good

IshitaGhosh (approved):

PR #5767, opened 2021-09-11
xclbinutil: Added alternative way to add PS Kernels

rozumx:
Work Done
----------
+ Created a new option '--add-ps-kerenl' to add a kernel on a given line (not need to have a separate JSON file.
+ Create a new method XclBinClass::addPsKernel() that will mananage the insertion of the PS Kernel shared libary and metadata into the xclbin image.
+ Updated regression that validates functionality and round trip expected data.
+ Updated boost::program_options::value to use decltype instead of duplicating the variable's type value
+ Change the readSubPayload stream for fstream to istream.  Doing so allows for the use of istringstream.  Note: Both fstream and istringstream inherit from istream.


AShivangi (approved):

jeffli-xilinx (approved):

rozumx:
retest this please

rozumx:
Binary check is a false positive for the file in question is being removed from the repository.
Merging code into master.

PR #5768, opened 2021-09-11
Updated the code to be solely relying on the boost process library.

rozumx:


dayeh-xilinx:
retest this please. Failed with Cento78, amazon luniux and centos edge. But builds on other OSs are fine.

PR #5773, opened 2021-09-13
CR-1110906 don't use ECC stalls in heat map

hackwa:


pgschuey (approved):

jvillarre (approved):
Change to mask looks good

PR #5747, opened 2021-09-08
Reserve host memory regardless address translator presence

chienwei-lan:

PR #5779, opened 2021-09-13
Show reserved HOST MEM in legacy xbutil query

chienwei-lan:


AShivangi (changes requested):
Please change it in https://github.com/Xilinx/XRT/blob/929578c23c8f22c14f95134b68952209788a3f91/src/runtime_src/core/pcie/linux/device_linux.cpp#L595 as well

stsoe (approved):

chienwei-lan:
> Please change it in
> 
> https://github.com/Xilinx/XRT/blob/929578c23c8f22c14f95134b68952209788a3f91/src/runtime_src/core/pcie/linux/device_linux.cpp#L595
> 
> as well

After this PR https://github.com/Xilinx/XRT/pull/5747, users now can reserve the host memory regardless if the device has address translator IP or not and the value is shown as Shared Host Memory as usual, which behavior does not change.

And the effective host memory size is shown and introduced as a new field as Enable Host Memory, which depends on the capability of the address translator of the device, not always same size as user reserved

`Pcie Info
Shared Host Memory     : 1 GB
Max Shared Host Memory : 16 GB
Enabled Host Memory    : 1 GB`

maxzhen (approved):

AShivangi (approved):

rozumx (approved):

PR #5781, opened 2021-09-13
CR-1110055

larry9523:
Improve check for mapping host BO to prevent arg->size overflow. In previous check, paddr + size can overflow and falls into the valid address range.


dayeh-xilinx:
retest this please. Failed with infra error.

mamin506 (approved):

PR #5783, opened 2021-09-14
CR-1110239:  Added A DRC that AP_CTRL_NONE shouldnt have interrupts enabled

chvamshi-xilinx:
This change fixes the segmentation fault causing for AP_CNTRL_NONE kernels.
We shouldn't create threads for ap_ctrl_none. We are not calling xrt_cu_init also for the ap_ctrl_none kernels.
Need to check how this test is passing in DC cases.

### DIFF
--- a/src/runtime_src/core/common/drv/kds_core.c
+++ b/src/runtime_src/core/common/drv/kds_core.c
@@ -1198,7 +1198,6 @@ int kds_cfg_update(struct kds_sched *kds)
 				continue;
 
 			xcu = cu_mgmt->xcus[i];
-
 			ret = xrt_cu_cfg_update(xcu, kds->cu_intr);
 			if (!ret)
 				cu_mgmt->cu_intr[i] = kds->cu_intr;
--- a/src/runtime_src/core/common/drv/kds_core.c
+++ b/src/runtime_src/core/common/drv/kds_core.c
@@ -1198,11 +1198,6 @@ int kds_cfg_update(struct kds_sched *kds)
 				continue;
 
 			xcu = cu_mgmt->xcus[i];
-			/* Do not start polling/interrupt thread
-			 * for ap_ctrl_none kernels
-			 */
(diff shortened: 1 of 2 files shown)


mamin506 (changes requested):
This change doesn't make sense to me.

For AP_CTRL_NONE CU, it should not have interrupt capability. In this case, xrt_cu_cfg_update() should return -ENOSYS.

See xrt_cu.c,

int xrt_cu_cfg_update(struct xrt_cu *xcu, int intr)
{
        int (* cu_thread)(void *data);
        int err = 0;

        /* Check if CU support interrupt in hardware */
        if (!xcu->info.intr_enable)
                return -ENOSYS;
                
        ....
}

Could you check if the AP_CTRL_NONE cu has intr_enable set to true? It should be false.
If this is the case, could you check why and fix?

chvamshi-xilinx:
> This change doesn't make sense to me.
> 
> For AP_CTRL_NONE CU, it should not have interrupt capability. In this case, xrt_cu_cfg_update() should return -ENOSYS.
> 
> See xrt_cu.c,
> 
> int xrt_cu_cfg_update(struct xrt_cu _xcu, int intr) { int (_ cu_thread)(void *data);
> int err = 0;
> 
> ```
>     /* Check if CU support interrupt in hardware */
>     if (!xcu->info.intr_enable)
>             return -ENOSYS;
>             
>     ....
> ```
> 
> }
> 
> Could you check if the AP_CTRL_NONE cu has intr_enable set to true? It should be false.
> If this is the case, could you check why and fix?

That is a good point @mamin506 . I will move CR to filer as this is for RTL kernel. They might have packaged it incorrectly.
I am keeping this PR still open to print DRC instead of crashing the kernel.

dayeh-xilinx:
retest this please. Don't know why this gets build 4 times. Try again.

chvamshi-xilinx:
@mamin506 , Can you please check the DRC.

mamin506 (approved):

PR #5786, opened 2021-09-14
CR-1110289 XRT install ends with "ERROR: Unknown command: 'config'"

maxzhen:


rozumx (approved):

PR #5795, opened 2021-09-15
 CR-1109015: xbutil to report x86 all Async errors

sarab96:
xbutil to report x86 all Async errors in time series order.
Report all errors from all processes

### DIFF
--- a/src/runtime_src/core/tools/common/ReportAsyncError.cpp
+++ b/src/runtime_src/core/tools/common/ReportAsyncError.cpp
@@ -22,6 +22,8 @@
 #include "core/common/api/error_int.h"
 #include "core/include/experimental/xrt_error.h"
 #include "core/include/xrt_error_code.h"
+#include "core/common/query_requests.h"
+
 #include <boost/algorithm/string.hpp>
 
 const static long unsigned NanoSecondsPerSecond = 1000000000;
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/tools/common/ReportAsyncError.cpp
```diff
@@ -22,15 +22,51 @@
...
-  boost::property_tree::ptree pt;
   boost::property_tree::ptree error_array;
+
+  //Code for new format; Binary array of error structs in sysfs
+  try {
+    auto buf = xrt_core::device_query<xrt_core::query::xocl_errors>(device);
+    if (buf.empty())
+      return error_array;
+    auto all_errors = xrt_core::query::xocl_errors::to_errors(buf);
+    for (auto error: all_errors) {
```
stsoe, line 42: `const auto&`  maybe

stsoe (changes requested):
@sarab96 One minor tweak otherwise good.

src/runtime_src/core/tools/common/ReportAsyncError.cpp
```diff
@@ -57,8 +93,6 @@ populate_async_error(const xrt_core::device * device)
     // ignore, likely that async error not supported
   }
 
-  pt.add_child("errors", error_array);
```
rozumx, line 60: The JSON schema has changed rather drastically, (e.g., removal of the errors array, asynchronous_errors becoming and array (from originally being a dictionary)).  

It "might" be O.K. for there most likely might not be any customers using this schema.  Never the less, it should be discussed at the XRT PRT.

rozumx (changes requested):
Please update the comments with:

1. What is being changed.
2. And example of the old and new JSON schemas
3. What the new human report looks like.

src/runtime_src/core/tools/common/ReportAsyncError.cpp
rozumx: Chatted with Sarabjeet off-line.  This schema was introduced in 2021.2, so schema changes don't require XRT PRT approval.

src/runtime_src/core/tools/common/ReportAsyncError.cpp
sarab96: done

sarab96:
No change for edge side reports.
For x86 host side xbutil will report the errors in same format/schema as before.
But for x86 host side xbutil will report errors from all processes.

> Please update the comments with:
> 
> 1. What is being changed.
> 2. And example of the old and new JSON schemas
> 3. What the new human report looks like.

stsoe (approved):

AShivangi (approved):

rozumx (approved):

PR #5796, opened 2021-09-15
CR-1110672: Clarify profiling summary memory resource table entry

jvillarre:
Updating the tooltip on the new Memory Resource table to clarify what directions READ and WRITE is in this context refer to.


pgschuey (approved):
looks good

PR #5797, opened 2021-09-15
CR-1110810: Correct latency type printed in profile summary table

jvillarre:
Change average latency to be printed as a double as expected by downstream tools.


pgschuey (approved):
change to double looks good

PR #5803, opened 2021-09-16
Updated cmake build flow for ps kernel support

jeffli-xilinx:
Added libdw, libelf, and libffi to cmake
Updated debian build for the above packages
Updated xrtdeps.sh for the above packages


stsoe (approved):
Looks good to me.  Do we want this to make it into 21.2 release?

rozumx (approved):

PR #5812, opened 2021-09-17
CR-1111520 Do not end xdp profiling when aie event profiling api is called

hackwa:


jvillarre (approved):
Removal of code looks good to enable both profiling APIs from host and xrt.ini profiling switches.

PR #5815, opened 2021-09-18
Fix compilation of xrt tests on windows

stsoe:


stsoe:
retest this please - bad sprite test which no one wants to fix

PR #5816, opened 2021-09-18
Ensure schema compatible xrt::device::get_info return values

stsoe:
In order to ensure that json formatted return values of
xrt::device::get_info conforms to the same schema in future
versions of XRT, the call is capturing the version of XRT used
when the host application is built.

The compiled version is passed to implementation of get_info() which
can then ensure schema compatibility as necessary.

* No user visible API change, just call templated public get_info(param)
  as normal.  This templated function in turn calls new overloaded
  private get_info().

* Add overloaded private get_info(param, abi) where the abi argument
  is really the version of XRT used when the host application was
  built.

* Pass the ABI (the application version of XRT) to json_str(…) and use
  in the future when / if new schema.

* Leave existing private get_info(param) untouched so that 2021.1 host
  application that reference this symbol can continue to find it in
  new versions of XRT.  Remove after a few years.


src/runtime_src/core/common/api/xrt_device.cpp
```diff
@@ -20,8 +20,8 @@
 #define XCL_DRIVER_DLL_EXPORT  // exporting xrt_device.h
 #define XRT_CORE_COMMON_SOURCE // in same dll as core_common
 
-#include "core/include/experimental/xrt_device.h"
-#include "core/include/experimental/xrt_aie.h"
+#include "core/include/xrt/xrt_device.h"
+#include "core/include/xrt/xrt_aie.h"
```
keryell, line 24: Do we expect the `experimental` files to be renamed or removed if they are already in `xrt`?

src/runtime_src/core/common/api/xrt_device.cpp
```diff
@@ -289,28 +291,39 @@ get_info(info::device param) const
...
   }
 
   throw std::runtime_error("internal error: unreachable");
 }
 
+// Deprecated but left for support of old existing binaries in the
+// field that reference this symbol. Unused since xrt-2.12.x
+boost::any
+device::
+get_info(info::device param) const
```
keryell, line 320: So, you deprecate this.
But you could also keep this by default and not advertising the one with the ABI, if the application does not care about the ABI besides just using the one it does understand, which means the default one defines in the headers when compiled.

keryell (commented):
It looks like you got the infrastructure.
Given this, you could use the ABI-explicit API just for the CI unit tests testing the various ABI but not advertising it for normal users.

src/runtime_src/core/common/api/xrt_device.cpp
stsoe: At some point yes.   Ideally I would like to remove them immediately when promoting the APIs from experimental to mature, but that would create a hassle for DSV in their sprite tests.   It's probably one of those things that we will never get around to?

src/runtime_src/core/common/api/xrt_device.cpp
stsoe: I don't advertise the ABI one.   The only public API is templated xrt::device::get_info(), which now calls implementation get_info(..., abi).     The ABI argument is an implementation detail.   The old get_info() implementation would be referenced in binaries compiled against xrt-2.11.x, so in order for these binaries to continue to run (without recompilation), the old get_info symbol is preserved.

dayeh-xilinx:
retest this please. Fixed a missing folder during branching.

PR #5819, opened 2021-09-20
CR-1111672 start aie profile ctrs after group config

hackwa:
Instead of starting counters before group, we now start them after group config


jvillarre (approved):
Moving code to after group events looks good

PR #5822, opened 2021-09-21
CR-1111608: Fixing dfx issues on VCK190 DFx platform

chvamshi-xilinx:
CR Number
CR-1111608 :On vck190 dfx platform, depending on the AIE xclbin packaging order, using multiple AIE xclbin files causes kernel panic

Issues
- Previously loaded xclbin related stuff is not getting cleaned up when we are downloading new xclbin. 
- We are accessing old IPs after loading new xclbin which is causing a crash
- Graph reset is not being called

Solution
Cleanup previously loaded xclbin stuff.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
@@ -843,14 +843,12 @@ zocl_xclbin_read_axlf(struct drm_zocl_dev *zdev, struct drm_zocl_axlf *axlf_obj,
 			 /*
 			  * cleanup previously loaded xclbin related data
 			  * before loading new bitstream/pdi
-			  * */
+			  */
 			if (kds_mode == 1 && (zocl_xclbin_get_uuid(zdev) != NULL)) {
 				subdev_destroy_cu(zdev);
 				if (zdev->aie) {
 					zocl_aie_reset(zdev);
-					mutex_lock(&zdev->aie_lock);
-					zdev->aie->aie_reset = false;
-					mutex_unlock(&zdev->aie_lock);
+					zocl_destroy_aie(zdev);
 				}
 			}
 			/*
--- a/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
@@ -840,6 +840,19 @@ zocl_xclbin_read_axlf(struct drm_zocl_dev *zdev, struct drm_zocl_axlf *axlf_obj,
 			DRM_INFO("disable partial bitstream download, "
 			    "axlf flags is %d", axlf_obj->za_flags);
 		} else {
+			 /*
+			  * cleanup previously loaded xclbin related data
+			  * before loading new bitstream/pdi
+			  * */
+			if (kds_mode == 1 && (zocl_xclbin_get_uuid(zdev) != NULL)) {
+				subdev_destroy_cu(zdev);
+				if (zdev->aie) {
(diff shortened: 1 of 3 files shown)


xilinxgitops:
retest this please

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -840,6 +840,19 @@ zocl_xclbin_read_axlf(struct drm_zocl_dev *zdev, struct drm_zocl_axlf *axlf_obj,
...
+			 /*
+			  * cleanup previously loaded xclbin related data
+			  * before loading new bitstream/pdi
+			  * */
+			if (kds_mode == 1 && (zocl_xclbin_get_uuid(zdev) != NULL)) {
+				subdev_destroy_cu(zdev);
+				if (zdev->aie) {
+					zocl_aie_reset(zdev);
+					mutex_lock(&zdev->aie_lock);
+					zdev->aie->aie_reset = false;
```
larry9523, line 852: This is a bit weird because we just reset AIE but saying aie_reset is false. I think we are safe to move aie out of reset state only after the PDI is successfully loaded.

There might be a corner case if we don't set the aie_reset correctly, e.g. somehow the following load process fails but we record aie is not reset. Then if user load a non-dfx xclbin, and then access AIE array, this will cause SError.

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -840,6 +840,19 @@ zocl_xclbin_read_axlf(struct drm_zocl_dev *zdev, struct drm_zocl_axlf *axlf_obj,
 			DRM_INFO("disable partial bitstream download, "
 			    "axlf flags is %d", axlf_obj->za_flags);
 		} else {
+			 /*
+			  * cleanup previously loaded xclbin related data
+			  * before loading new bitstream/pdi
+			  * */
```
larry9523, line 846: */

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
chvamshi-xilinx: This is because of following check in zocl_create_aie. It fails to create aie if aie_reset is true
https://github.com/Xilinx/XRT/blob/master/src/runtime_src/core/edge/drm/zocl/zocl_aie.c#L242
I think, best is to call zocl_destory_aie instead of setting this variable.
https://github.com/Xilinx/XRT/blob/master/src/runtime_src/core/edge/drm/zocl/zocl_aie.c#L319

larry9523: I understand why you are setting it. My point is we need a mechanism to protect AIE array from accessing before a PDI is loaded, using this flag or something else. 
Setting this flag to false before loading PDI will compromise this protection.

chvamshi-xilinx: Let me set it after successful pdi load. Hope that should be fine.

larry9523: Can't we check the presence AIE_METADAT section for that purpose?

larry9523: Then we are safe to set aie_reset to true after loading PDI, right?
But I am totally fine with your proposal to destroy aie. And seems to me, this destroy aie operation should be done before loading pdi.

mamin506 (approved):

xilinxgitops:
retest this please

PR #5825, opened 2021-09-21
Update xrtdeps

rbramand-xilinx:
> Updated xrtdeps to install packages properly on all supported os
> In case of centos kernel-devel and kernel-header packages of $(uname -r) version are not available always, so adding package kernel-devel-$(uname -r) will cause xrtdeps to fail intermittently. But package kernel-devel and kernel-headers are always available, so including those.

### DIFF
--- a/src/runtime_src/tools/scripts/xrtdeps.sh
+++ b/src/runtime_src/tools/scripts/xrtdeps.sh
@@ -118,23 +118,14 @@ rh_package_list()
 
     if [ $MAJOR == 8 ]; then
 
-        RH_LIST+=(systemd-devel)
+        RH_LIST+=(\
+          systemd-devel \
+          libarchive \
+          )
 
-        if [ $FLAVOR == "centos" ]; then
-            #fix cmake issue in centos 8.*
+        if [ $FLAVOR == "rhel" ]; then
             RH_LIST+=(\
-            libarchive \
-            )
-        else
-            RH_LIST+=(\
-            opencv \
-            )
-        fi
-
-        if [ $docker == 0 ]; then
-            RH_LIST+=(\
-             kernel-devel-$(uname -r) \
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx (approved):

stsoe (approved):

maxzhen (commented):
When the specific version of the kernel-devel and kernel-header packages are not available, shouldn't we fail the installation and ask user to update the kernel version so that the matching packages can be installed? It does not make much sense to install the kernel-devel and kernel-header packages for a different version of kernel than the one that is currently running on the host, unless we do not intent to run the built XRT package on the same system where we build it?

chvamshi-xilinx:
> When the specific version of the kernel-devel and kernel-header packages are not available, shouldn't we fail the installation and ask user to update the kernel version so that the matching packages can be installed? It does not make much sense to install the kernel-devel and kernel-header packages for a different version of kernel than the one that is currently running on the host, unless we do not intent to run the built XRT package on the same system where we build it?

Hi @maxzhen , we are not going to get kernel-header-$(uname -r)/kernel-devl-$(uname -r) for most of the Centos versions. They always keep kernel-header and kernel-devl packages without $(uname -r).
We are currently exiting from installing all other packages when these kernel* packages are not found. Its better to install whatever is available.  It is good if these packages match with the kernel version which is running, if not, they have to install specific kernel-headers/devels from some repo. There is no standard repo for Centos, We mention this in the documentation also.

maxzhen (approved):

chvamshi-xilinx:
@stsoe , this is required in 2021.2 aswell as People would run it issues on Centos 8.3 when try to install dependencies using xrtdeps.sh. 
This PR contain a change to support RHEL 8.3 aswell (libarchive required for RHEL 8.3)

dayeh-xilinx:
retest this please

PR #5828, opened 2021-09-21
Hide doxygen for xrt::xclbin axlf specific APIs

stsoe:
We are not documenting xclbin binary structure, so we should also
not document how users can access the axlf and specific sections.
The APIs are there for use if you know what you are doing.


uday610:
Thanks

uday610 (approved):

dayeh-xilinx:
retest this please

PR #5830, opened 2021-09-22
CR-1110308 xmc: xbmgmt load-config passes true to xmc driver to enable clock scaling feature

rajkumar-xilinx:
Runtime clock scaling feature is one of the device configs and corresponding driver expects enable/disable string to enable/disable this feature through xbmgmt commands. But, new xbmgmt dumps device configs in true/false onto .ini file.
So, just aligning with this new tool, modified xmc driver to accept true/false instead of enable/disable.
xbmgmt legacy commands also works well with true/false.

Legacy (old) commands: to enable/disable Runtime clock scaling feature.
xbmgmt --legacy config --device --card 62:00.0 --runtime_clk_scale true
xbmgmt --legacy config --device --card 62:00.0 --runtime_clk_scale false
xbmgmt --legacy config --device --card 62:00.0 --runtime_clk_scale enable
xbmgmt --legacy config --device --card 62:00.0 --runtime_clk_scale disable

New commands: modify .ini file to enable/disable Runtime clock scaling feature.
scaling_enabled=true
scaling_enabled=false

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
@@ -2272,8 +2272,7 @@ static ssize_t scaling_enabled_store(struct device *dev,
 {
 	struct xocl_xmc *xmc = platform_get_drvdata(to_platform_device(dev));
 
-	if ((strncmp(buf, "enable", strlen("enable")) == 0) ||
-	   (strncmp(buf, "true", strlen("true")) == 0))
+	if (strncmp(buf, "true", strlen("true")) == 0)
 		runtime_clk_scale_enable(xmc);
 	else
 		runtime_clk_scale_disable(xmc);
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xmc.c
@@ -1,7 +1,7 @@
 /*
  * A GEM style device manager for PCIe based OpenCL accelerators.
  *
- * Copyright (C) 2016-2020 Xilinx, Inc. All rights reserved.
+ * Copyright (C) 2016-2021 Xilinx, Inc. All rights reserved.
  *
  * Authors: chienwei@xilinx.com
  *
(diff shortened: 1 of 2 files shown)


maxzhen:
Why not change the tool to pass "enable" instead?

rajkumar-xilinx:
> Why not change the tool to pass "enable" instead? or if we believe "true" / "false" would be more appropriate, why do we need to still support "enable"? There is no backward compatibility requirement b/w driver and tools.

Yes, we can remove "enable" option now. We were using this option only in the sysfs path, but we can start using "true/false" instead of "enable". I will update PR. Thanks.

maxzhen:
> 
> 
> > Why not change the tool to pass "enable" instead? or if we believe "true" / "false" would be more appropriate, why do we need to still support "enable"? There is no backward compatibility requirement b/w driver and tools.
> 
> Yes, we can remove "enable" option now. We were using this option only in the sysfs path, but we can start using "true/false" instead of "enable". I will update PR. Thanks.

I'm assuming removing support for "enable" does not impact legacy xbmgmt, right?

chvamshi-xilinx:
> > Why not change the tool to pass "enable" instead? or if we believe "true" / "false" would be more appropriate, why do we need to still support "enable"? There is no backward compatibility requirement b/w driver and tools.
> 
> Yes, we can remove "enable" option now. We were using this option only in the sysfs path, but we can start using "true/false" instead of "enable". I will update PR. Thanks.

I hope xbmgmt dump is printing scaling_support=true/false in ini file.

stsoe (changes requested):
Hi @rajkumar-xilinx  Could you add a few lines the description of this PR with the highlights of what was changed and why?

Is the PR also for 2021.2 or just for 2022.1?

rajkumar-xilinx:
> > > Why not change the tool to pass "enable" instead? or if we believe "true" / "false" would be more appropriate, why do we need to still support "enable"? There is no backward compatibility requirement b/w driver and tools.
> > 
> > 
> > Yes, we can remove "enable" option now. We were using this option only in the sysfs path, but we can start using "true/false" instead of "enable". I will update PR. Thanks.
> 
> I'm assuming removing support for "enable" does not impact legacy xbmgmt, right?

No impact on legacy commands as well. Verified this case also.

maxzhen (approved):

rajkumar-xilinx:
> Hi @rajkumar-xilinx Could you add a few lines the description of this PR with the highlights of what was changed and why?
> 
> Is the PR also for 2021.2 or just for 2022.1?

I have updated description and it is for 2021.2

stsoe (approved):
Thanks @rajkumar-xilinx  Unfortunately we are no longer auto merging to 2021.2 so please cherry pick for 2021.2 and create another pull request for that branch.

rajkumar-xilinx:
2021.2 branch PR https://github.com/Xilinx/XRT/pull/5834

rajkumar-xilinx:
> Thanks @rajkumar-xilinx Unfortunately we are no longer auto merging to 2021.2 so please cherry pick for 2021.2 and create another pull request for that branch.

2021.2 branch PR #5834

chvamshi-xilinx (approved):

PR #5831, opened 2021-09-22
refactor mailbox test code -- avoid duplicate code introduced by vitis-3558 Mailbox positive/negative test framework support (#5811)

xuhz:
avoid duplicate code introduced by  
vitis-3558 Mailbox positive/negative test framework support (#5811) 

so it is also for 2021.2


stsoe (changes requested):
@xuhz  Please add a proper description as to what was done in this pull request, basically the highlights of this PR.

Another question:  Is this for 2021.2 also, or only for 2022.1?

maxzhen (approved):

stsoe (approved):

PR #5832, opened 2021-09-22
CR-1111463: xbutil examine is slow when device is not ready

chvamshi-xilinx:
CR-1111463  : xbutil examine is very slow if card is not in ready state

Xbutil examine is trying to get the sensor information in xclOpen. We shouldn't try to get any information which requires mailbox communication when device is not ready. Each request takes some time to timeout.

### DIFF
--- a/src/runtime_src/core/pcie/linux/shim.cpp
+++ b/src/runtime_src/core/pcie/linux/shim.cpp
@@ -999,6 +999,15 @@ void shim::xclSysfsGetDeviceInfo(xclDeviceInfo2 *info)
     info->mNumClocks = numClocks(info->mName);
 
     mDev->sysfs_get<unsigned short>("mb_scheduler", "kds_numcdmas", errmsg, info->mNumCDMA, static_cast<unsigned short>(-1));
+    
+    mDev->sysfs_get("", "link_width", errmsg, info->mPCIeLinkWidth, static_cast<unsigned short>(-1));
+    mDev->sysfs_get("", "link_speed", errmsg, info->mPCIeLinkSpeed, static_cast<unsigned short>(-1));
+    mDev->sysfs_get("", "link_speed_max", errmsg, info->mPCIeLinkSpeedMax, static_cast<unsigned short>(-1));
+    mDev->sysfs_get("", "link_width_max", errmsg, info->mPCIeLinkWidthMax, static_cast<unsigned short>(-1));
+    
+    //dont try to get any information which needs mailbox communication when device is not ready.
+    if(!mDev->is_mgmt() && !mDev->is_ready)
+        return;
 
     //get sensors
     unsigned int m12VPex, m12VAux, mPexCurr, mAuxCurr, mDimmTemp_0, mDimmTemp_1, mDimmTemp_2,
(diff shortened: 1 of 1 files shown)


maxzhen:
@chvamshi-xilinx I'm confused... are you saying that reading these sysfs nodes ("link_width",  "link_speed", "link_speed_max", "link_width_max") will trigger mailbox communication?

chvamshi-xilinx:
> @chvamshi-xilinx I'm confused... are you saying that reading these sysfs nodes ("link_width", "link_speed", "link_speed_max", "link_width_max") will trigger mailbox communication?

No Max. Getting sensors value triggers a mailbox communication which gets timedout if device is not ready. Please check this condition. If device is not ready for userpf , then am returning without reading sensor information. getting link* sysfs nodes is bottom of this function earlier, thatswhy i just moved it before return sothat xbutil gets these values even if device is not ready.
 
    //dont try to get any information which needs mailbox communication when device is not ready.
    if(!mDev->is_mgmt() && !mDev->is_ready)
        return;

maxzhen:
@chvamshi-xilinx thanks for explaining this. Now I get it.

maxzhen (approved):

AShivangi:
I have 2 questions:
1. Is "link_width", "link_speed", "link_speed_max", "link_width_max" needed when a device is not ready? 
Why don't we add
```
if(!mDev->is_mgmt() && !mDev->is_ready)
        return;
```
  before any kind of sysfs call?

2. What time improvement did you notice?

chvamshi-xilinx:
> I have 2 questions:
> 
> 1. Is "link_width", "link_speed", "link_speed_max", "link_width_max" needed when a device is not ready?
>    Why don't we add
> 
> ```
> if(!mDev->is_mgmt() && !mDev->is_ready)
>         return;
> ```
> 
> before any kind of sysfs call?
> 
> 1. What time improvement did you notice?

1) I have only removed fields which require mailbox communication under "ready" check so that there wont be any difference in output earlier and now.

2) My machine has 3 cards out of which 2 are not ready. It used to take  5 minutes + to 2 seconds.

AShivangi:
That's a great improvement! Thanks for providing that info.

AShivangi (approved):

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -999,6 +999,15 @@ void shim::xclSysfsGetDeviceInfo(xclDeviceInfo2 *info)
...
 
     mDev->sysfs_get<unsigned short>("mb_scheduler", "kds_numcdmas", errmsg, info->mNumCDMA, static_cast<unsigned short>(-1));
+    
+    mDev->sysfs_get("", "link_width", errmsg, info->mPCIeLinkWidth, static_cast<unsigned short>(-1));
+    mDev->sysfs_get("", "link_speed", errmsg, info->mPCIeLinkSpeed, static_cast<unsigned short>(-1));
+    mDev->sysfs_get("", "link_speed_max", errmsg, info->mPCIeLinkSpeedMax, static_cast<unsigned short>(-1));
+    mDev->sysfs_get("", "link_width_max", errmsg, info->mPCIeLinkWidthMax, static_cast<unsigned short>(-1));
+    
+    //dont try to get any information which needs mailbox communication when device is not ready.
+    if(!mDev->is_mgmt() && !mDev->is_ready)
```
rozumx, line 1009: From a flow perspective, this code looks good.   

From a initialization perspective, there are variables in the info structure that are NOT initialized, which will result in unexpected behavior.  It would be better if these values were initialized.

Normally, it would be great to initialize all of these values at the beginning of this method, but looking at the flow suggests doing this blindly will not work for methods calling this method does set some values.

Being that we are nearly the end-of-the-release, it would be better to do the following:
1.  Move line 1009 and invert the logic to 1017.  The idea here is to read the sensor only when the board is in a good state.  
2. Remove the return on line 1010.
2. Initialize the variables on line 1013 to zero.  Simply adding an `= -1` should work, assuming that indeed that is what the default variable should be.

Doing this will allow for more consistent initialization of this structure.

rozumx (changes requested):
Code refactoring requested to initialize the xclDeviceInfo2 structure.

src/runtime_src/core/pcie/linux/shim.cpp
rozumx: Thank you Vamshi for review the code with me and showing:

1. The device structure is initialized to zeros prior to calling these function.
2. The `sysfs_get_sensor()` function returns back a zero if errors occur.
3. The reports are never called if the device is not ready.

For these reasons, the issues I had earlier have been resolved.

rozumx (approved):
Approved.  Thank you Vamshi for taking the time to explain how these changes will not impact existing reporting functionality.

chvamshi-xilinx:
Thanks @rozumx for the closure in a call :-)

PR #5837, opened 2021-09-23
Update release version after separation from 2021.2

stsoe:
XRT release version is 202210
XRT version is 2.13.x


build/build.sh
```diff
@@ -276,8 +276,8 @@ if [[ $opt == 1 ]]; then
   if [[ $driver == 1 ]]; then
     unset CC
     unset CXX
-    echo "make -C usr/src/xrt-2.12.0/driver/xocl"
-    make -C usr/src/xrt-2.12.0/driver/xocl
+    echo "make -C usr/src/xrt-2.13.0/driver/xocl"
```
rozumx, line 279: Can we make updates like this to be more "data driven"?

rozumx (approved):
Approved, but I can't help to think we can refactor the code to be more data driven.

Unfortunately, we don't have time at this time in the release, but we might wish to consider this at the next branch.

PR #5840, opened 2021-09-23
CR-1112068 Sometimes configure command timeout

mamin506:
Instead of schedule a worker to handle interrupt, directly call the interrupt handle function intc_polling() in the interrupt routine.

To avoid interrupt storm, intc_polling() would only poll max_try times. This is the reason we schedule a worker in before.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/intc.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/intc.c
@@ -13,7 +13,6 @@
  * GNU General Public License for more details.
  */
 
-#include <linux/workqueue.h>
 #include "../xocl_drv.h"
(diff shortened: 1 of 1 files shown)


maxzhen (approved):

stsoe (changes requested):
The CR describes the what, but the commit message should explain just briefly the how.  

It doesn't have to be more than a few lines, but since the code isn't really documented to begin with, it is difficult to understand how the changes fixed the CR.

mamin506:
> The CR describes the what, but the commit message should explain just briefly the how.
> 
> It doesn't have to be more than a few lines, but since the code isn't really documented to begin with, it is difficult to understand how the changes fixed the CR.

Done. 👍

stsoe (approved):

PR #5843, opened 2021-09-24
CR-1111463: xbutil examine is slow when device is not ready: Merge pull request #5832 from chvamshi-xilinx/xbutil_slow

chvamshi-xilinx:
CR-1111463: xbutil examine is slow when device is not ready
(cherry picked from commit a75bcd2b8e872e9ace794c22d30964bfdb6bb749)

**Root cause:**
Xbutil examine is trying to get the sensor information in xclOpen. We shouldn't try to get any information which requires mailbox communication when device is not ready. Each request takes some time to timeout.

**Fix:**
Don't read sensors information if device is not in ready state

**Reviewers:**
Max/Shivangi/Stephen


anupcs-xlnx (approved):
This was already on the list of CRs targeting 2021.2. Approved.

PR #5850, opened 2021-09-24
Hal-level profiling callback mechanism refactoring

jvillarre:
This pull request simplifies the callback mechanism between HAL/Shim functions and the XDP library to use the same mechanism as used in native profiling (auto template function wrapper).  The new interface is simplified in the following manner:
- The callback class hierarchy on both the XRT side and the XDP side has been simplified to a generic callback class and a buffer transfer callback class
- Both the edge and pcie callback mechanism has been refactored. 
- *_CB macros have been removed and replaced with profiling_wrapper() function calls in the shims which forward the internal implementations as a lambda function to the wrapper
- The profiling_wrapper functions will not create the logger instances if HAL level tracing is disabled so the overhead will be minimal.
- In order to support the compiler detecting the return type of the lambda functions, there needed to be some explicit casting of nullptrs added to some shim function implementations.

Also changed:
- Fixing a bug in native_profile.h where the wrong type of constructors were specified for deletion.


stsoe (approved):
No detailed review.  But looks good, nice to see how much simpler this is, look at all the code removed :-)

IshitaGhosh (approved):
This refactoring looks good.

chvamshi-xilinx (approved):
Looks much simple and scalable. Thanks

PR #5851, opened 2021-09-24
ert polling mode cleanup

mamin506:
Even in ERT interrupt mode, the command queue subdev will read CSR registers to get the completed command. This is redundant. Since when interrupt occurs, intc subdev will read CSR register and let command queue know which slot has interrupted.

I aware of that struct command_queue has defined polling_mode but it is not used. This change will update polling mode flag when configure command queue. Only read CRS register when polling mode is true.

In my test, this change helps to improve performance on u200 201820_2 card, which is using legacy ERT subsystem.
Before:
$/opt/xilinx/xrt/test/xcl_iops_test.exe -k /opt/xilinx/dsa/xilinx_u200_xdma_201830_2/test/verify.xclbin -d 0000:05:00.1
Overall Commands:  100000 **IOPS: 99897 (hello)**
TEST PASSED

After:
$/opt/xilinx/xrt/test/xcl_iops_test.exe -k /opt/xilinx/dsa/xilinx_u200_xdma_201830_2/test/verify.xclbin -d 0000:05:00.1
Overall Commands:  100000 **IOPS: 104482 (hello)**
TEST PASSED

On u50 nodma shell with ERT ssv3. The iops test result has no change.

### DIFF
--- a/src/runtime_src/core/common/drv/include/xrt_ert.h
+++ b/src/runtime_src/core/common/drv/include/xrt_ert.h
@@ -55,7 +55,7 @@ struct xrt_ert_queue_funcs {
 
 	int (*submit)(struct xrt_ert_command *ecmd, void *queue_handle);
 
-	int  (*queue_config)(uint32_t slot_size, void *ert_handle, void *queue_handle);
+	int  (*queue_config)(uint32_t slot_size, bool polling, void *ert_handle, void *queue_handle);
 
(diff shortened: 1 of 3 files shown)


maxzhen (approved):

dayeh-xilinx:
retest this please. Failed with one pcie_hw_emu test. Try again.

chienwei-lan (approved):

PR #5853, opened 2021-09-25
CR 1111369 - xbutil2 examine command should be ok without "-d" option

rozumx:
Work Done
----------
+ Creat a new CMake variable to enable / disable implied -d option.
+ For non-native builds, enabled (e.g. defined ENABLE_DEFAULT_ONE_DEVICE_OPTION) the CMake variable
+ Updated XBMain to enable / disable the implied -d option for all commands except for "examine".  Note: The "examine" command produces different reports depending on if the --device option was used.

### DIFF
--- a/src/runtime_src/core/tools/common/XBMain.cpp
+++ b/src/runtime_src/core/tools/common/XBMain.cpp
@@ -149,6 +149,17 @@ void  main_(int argc, char** argv,
   if (bHelp == true) 
     opts.push_back("--help");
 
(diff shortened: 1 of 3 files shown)


chvamshi-xilinx (approved):
Thanks Stephen. Looks good.

AShivangi (approved):

chvamshi-xilinx:
Hi @rozumx , Based on your PR Description, you wanted to enable this new flag only for non-native. But it is enabled in both native and non-native in the code.

chvamshi-xilinx (approved):
Thank you. Looks good.

PR #5864, opened 2021-09-27
CR-1111986 ASTeR TC 22.07 - u55c/n + X3 - validate does not run the host memory test after it has been enabled

chienwei-lan:
XRT now looks for the value of reserved host memory to decide either we should run the host memory test or not.

Worth notice that any platform can reserve host memory even without the address translator IP.

If users reserve the host memory with the device w/o address translator IP, host memory test will fail due to unable to allocate buffer over HOST[0]

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
@@ -789,7 +789,7 @@ int xocl_cleanup_mem_nolock(struct xocl_drm *drm_p)
 				continue;
 
 			if (IS_HOST_MEM(topology->m_mem_data[i].m_tag))
-				xocl_addr_translator_disable_remap(drm_p->xdev);
+				xocl_addr_translator_clean(drm_p->xdev);
 
 			xocl_info(drm_p->ddev->dev, "Taking down DDR : %d", i);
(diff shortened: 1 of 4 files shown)


maxzhen (approved):

AShivangi (approved):

dayeh-xilinx:
retest this please. Ran into an infra error.

rozumx (approved):

chienwei-lan:
Turns out it's because the pipeline doesn't have libboost,

#ifdef NO_BOOST_PROCESS
  is_done = true;
  bool exitCode = (os_stdout.str().find("PASS") != std::string::npos) ? true : false;
  run_test.get()->finish(exitCode, final_description);
  // Workaround: Clear the default progress bar output so as to print the Error: before printing [FAILED]
  // Remove this once busybar is implemented
  std::cout << EscapeCodes::cursor().prev_line() << EscapeCodes::cursor().clear_line();
  t.join();

  return exitCode ? 0 : 1;
}
#else 
  int exitCode = runningProcess.exit_code();
  run_test.finish(exitCode == 0 /*Success or failure*/, final_description);

  // Erase the "Pass Fail" message
  if (erasePassFailMessage && (XBUtilities::is_escape_codes_disabled() == 0)) 
    std::cout << EscapeCodes::cursor().prev_line() << EscapeCodes::cursor().clear_line();

  return exitCode;
}
#endif

PR #5870, opened 2021-09-28
updated mapping table for xbutil dmatest and its new method

uday610:
updated mapping table for xbutil dmatest and its new method


AShivangi (approved):
Thank you

PR #5872, opened 2021-09-29
CR-1111986 ASTeR TC 22.07 - u55c/n + X3 - validate does not run the host memory test after it has been enabled 

chienwei-lan:
(cherry picked from commit e47885ea0e55779ef84b0f7e6e1c3fe36d83464d)

### DIFF
--- a/src/runtime_src/core/tools/common/Process.cpp
+++ b/src/runtime_src/core/tools/common/Process.cpp
@@ -163,14 +163,20 @@ XBUtilities::runScript( const std::string & env,
   }
 
   is_done = true;
-  bool exitCode = (os_stdout.str().find("PASS") != std::string::npos) ? true : false;
-  run_test.get()->finish(exitCode, final_description);
+  bool passed = (os_stdout.str().find("PASS") != std::string::npos) ? true : false;
+  bool skipped = (os_stdout.str().find("EOPNOTSUPP") != std::string::npos) ? true : false;
+  run_test.get()->finish(passed, final_description);
   // Workaround: Clear the default progress bar output so as to print the Error: before printing [FAILED]
   // Remove this once busybar is implemented
   std::cout << EscapeCodes::cursor().prev_line() << EscapeCodes::cursor().clear_line();
   t.join();
 
-  return exitCode ? 0 : 1;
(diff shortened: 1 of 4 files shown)


maxzhen (approved):

anupcs-xlnx (approved):
Approved for 2021.2 backport

AShivangi (approved):

src/runtime_src/core/tools/common/Process.cpp
```diff
@@ -163,14 +163,20 @@ XBUtilities::runScript( const std::string & env,
   }
 
   is_done = true;
-  bool exitCode = (os_stdout.str().find("PASS") != std::string::npos) ? true : false;
-  run_test.get()->finish(exitCode, final_description);
+  bool passed = (os_stdout.str().find("PASS") != std::string::npos) ? true : false;
+  bool skipped = (os_stdout.str().find("EOPNOTSUPP") != std::string::npos) ? true : false;
```
rozumx, line 167: Did you look at the code below (part of the #else section) to see if this SKIPPING data is needed?

Yeah, depending on the OS Ubuntu or CentOS, the boost library may or may not be the version we need, hence this compiler directive.

rozumx (changes requested):

src/runtime_src/core/tools/common/Process.cpp
AShivangi: Yes this is needed as the code is failing on Amazon linux which doesn't have boost::process

chienwei-lan: We can capture the exact exit_code in the below section and don't have to handcraft the return code by parsing the output string thanks to <boost/process>, this is how it works in my ubuntu 18.04 machine.

The behavior of runTestCase() should be the same across all the environments, should not be failed on one OS and skipped on another even though parsing the output string to decide the return code is not ideal.

I think since we can get the exit_code in the else section, so we don't have to handcraft the return there.

maxzhen:
Had offline discussion w/ @rozumx. @chienwei-lan , please file a CR to track the concern for now.

PR #5873, opened 2021-09-29
Add xrt::bo constructor from xclBufferHandle

stsoe:
Allow applications to bridge into XRT APIs from old XCL APIs.

Specifically requested to support an IVAS framework where buffers are
allocated using old style APIs but kernel APIs are used for execution.

By allowing construction of xrt::bo from xclBufferHandle, the xrt::bo
can be used as an xrt::kernel argument.

The C++ and C API addition have been tested with XRT/tests/xrt/22_verify/xcl.cpp.


stsoe:
retest this please - embedded team is looking into intermittent failure

PR #5878, opened 2021-09-30
Add xrt::bo constructor from xclBufferHandle

stsoe:
Allow applications to bridge into XRT APIs from old XCL APIs.

Specifically requested to support an IVAS framwork where buffers are
allocated using old style APIs but kernel APIs are used for execution.

By allowing construction of xrt::bo from xclBufferHandle, the xrt::bo
can be used as an xrt::kernel argument.

The C and C++ API addition have been tested with XRT/tests/xrt/22_verify/xcl.cpp.

(cherry picked from commit 98ec1bdd8e0d87073c2f1b976e9b2450e86fb812)

### DIFF
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -184,6 +184,14 @@ public:
     size = prop.size;
   }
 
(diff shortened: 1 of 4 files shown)


anupcs-xlnx (approved):
Approved for 2021.2 as per Slack discussions

PR #5880, opened 2021-09-30
CR-1111450 “xclExecBuf” cannot be executed correctly in 2021.2 XRT

mamin506:
1. PS ERT (zocl) is not sync with host code change. We have to skip first 6 data in EXEC_WRITE command
2. PS ERT is required command slot size to be power of 2. This is unexpected in before. Since we update PS ERT is not easy, we will round up the slot size to next power of 2 in KDS.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_kds.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_kds.c
@@ -414,7 +414,6 @@ int zocl_command_ioctl(struct drm_zocl_dev *zdev, void *data,
 		start_krnl_ecmd2xcmd(to_start_krnl_pkg(ecmd), xcmd);
 		break;
 	case ERT_EXEC_WRITE:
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/edge/drm/zocl/zocl_kds.c
```diff
@@ -414,7 +414,6 @@ int zocl_command_ioctl(struct drm_zocl_dev *zdev, void *data,
 		start_krnl_ecmd2xcmd(to_start_krnl_pkg(ecmd), xcmd);
 		break;
 	case ERT_EXEC_WRITE:
-		DRM_WARN("ERT_EXEC_WRITE is obsoleted, use ERT_START_KEY_VAL\n");
```
maxzhen, line 417: can you do WARN_ONCE(), so that you don't have to remove it completely?

maxzhen (approved):

src/runtime_src/core/edge/drm/zocl/zocl_kds.c
mamin506: done.

PR #5885, opened 2021-10-01
azure: parse xclbin and check if it is valid to sent to mgmt

rajkumar-xilinx:
Parse xclbin in the azure mpd plugin before sending it to mgmt and check if it is valid
xclbin to send.
If the xclbin has BITSTREAM section then it is either DCP xclbin or bit.xclbin (full bitstream), both
these xclbins are not valid to program.

So, adding a check in azure mpd plugin to verify the same.

https://dev.azure.com/AzureXilinx/XRT%20Collaboration/_workitems/edit/84/

### DIFF
--- a/src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.cpp
+++ b/src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.cpp
@@ -222,6 +222,22 @@ static size_t WriteCallback(void *contents, size_t size, size_t nmemb, void *use
     return size * nmemb;
 }
 
+bool AzureDev::xclbin_section_hdr_present(const struct axlf *xclbin, enum axlf_section_kind kind)
+{
+    unsigned int i = 0;
+    unsigned int sections = XCLBIN_MAX_NUM_SECTION;
+
+    if (xclbin->m_header.m_numSections < XCLBIN_MAX_NUM_SECTION)
+        sections = xclbin->m_header.m_numSections;
+
+    for (i = 0; i < sections; i++) {
+        if (xclbin->m_sections[i].m_sectionKind == kind)
+            return true;
+    }
+
+    return false;
+}
+
 int AzureDev::azureLoadXclBin(const xclBin *buffer)
 {
     char *xclbininmemory = reinterpret_cast<char*> (const_cast<xclBin*> (buffer));
@@ -233,6 +249,13 @@ int AzureDev::azureLoadXclBin(const xclBin *buffer)
     if (fpgaSerialNumber.empty())
         return -E_EMPTY_SN;
     std::cout << "LoadXclBin FPGA serial No: " << fpgaSerialNumber << std::endl;
+
+    // check if the xclbin is valid
+    if (xclbin_section_hdr_present(buffer, BITSTREAM)) {
+        std::cout << "xclbin is not valid" << std::endl;
+        return -EINVAL;
+    }
+
     int index = 0;
     std::string imageSHA;
     std::vector<std::string> chunks;
--- a/src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.h
+++ b/src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.h
@@ -134,6 +134,7 @@ private:
     void get_fpga_serialNo(std::string &fpgaSerialNo);
     void msleep(long msecs);
     int goingTimeout();
+    bool xclbin_section_hdr_present(const struct axlf *xclbin, enum axlf_section_kind kind);
 };
 
 
--- a/src/runtime_src/core/tools/xbutil2/SubCmdProgram.cpp
+++ b/src/runtime_src/core/tools/xbutil2/SubCmdProgram.cpp
@@ -156,7 +156,7 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
       auto hdl = dev->get_device_handle();
       auto bdf = xrt_core::query::pcie_bdf::to_string(xrt_core::device_query<xrt_core::query::pcie_bdf>(dev));
       if (auto err = xclLoadXclBin(hdl,reinterpret_cast<const axlf*>(raw.data())))
-        throw xrt_core::error(err, "Could not program device" + bdf);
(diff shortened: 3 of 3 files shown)


AShivangi (changes requested):

AShivangi:
FYI, Stephen is on leave for 2 weeks

src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.cpp
```diff
@@ -222,6 +222,22 @@ static size_t WriteCallback(void *contents, size_t size, size_t nmemb, void *use
...
+
+    if (xclbin->m_header.m_numSections < XCLBIN_MAX_NUM_SECTION)
+        sections = xclbin->m_header.m_numSections;
+
+    for (i = 0; i < sections; i++) {
+        if (xclbin->m_sections[i].m_sectionKind == kind)
+            return true;
+    }
+
+    return false;
```
stsoe, line 238: How about 
`return xclbin::get_axlf_section(xclbin, kind) != nullptr;`

stsoe (changes requested):

src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.cpp
```diff
@@ -233,6 +249,13 @@ int AzureDev::azureLoadXclBin(const xclBin *buffer)
     if (fpgaSerialNumber.empty())
         return -E_EMPTY_SN;
     std::cout << "LoadXclBin FPGA serial No: " << fpgaSerialNumber << std::endl;
+
+    // check if the xclbin is valid
+    if (xclbin_section_hdr_present(buffer, BITSTREAM)) {
```
chvamshi-xilinx, line 254: We dont need new function. Please use existing function to check the existence of section

rajkumar-xilinx:
Hi @stsoe 

xrt_logmsg() is somehow not printing it's last argument properly. So, replaced xrt_logmsg() with message::send()
    For example:

    code1: xrt_logmsg(XRT_ERROR, "1000: %d", 1000);
    output: 1000: 100

    code2: xrt_logmsg(XRT_ERROR, "1000: %d, 1001: %d", 1000, 1001);
    output: 1000: 1000, 1001: 100

chvamshi-xilinx (approved):

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1337,7 +1337,11 @@ int shim::xclLoadXclBin(const xclBin *buffer)
         xrt_logmsg(XRT_ERROR, "CU was deadlocked? Hardware is not stable");
         xrt_logmsg(XRT_ERROR, "Please reset device with 'xbutil reset'");
       }
-      xrt_logmsg(XRT_ERROR, "See dmesg log for details. err=%d", ret);
+      else if (ret == -EINVAL) {
+        xrt_logmsg(XRT_ERROR, "Trying to load/program invalid xclbin");
+      }
+      xrt_core::message::send(xrt_core::message::severity_level::error, "XRT",
+         std::string("See dmesg log for details. err = " + std::to_string(ret)));
```
stsoe, line 1344: Please fix xrt_logmsg, in particular fix the bug in xrt_core::message::send in message.h.  It smells like an off-by-one some where.  Working around the bug in one place is not a good solution given that many places use xrt_logmsg.  Get to the root of the problem, please.

Turns out the bug is in max number of bytes passed to snprintf; it is indeed off by one.  
```
    std::vector<char> buf(sz+1);
    snprintf(buf.data(), sz, format, args ...);
    send(l, tag, buf.data());
```
Since snprintf adds the null terminator to the buffer, the buffer is allocated with room for that, but the call to snprintf doesn't reflect the actual size of the buffer, and hence is chopped off one byte.  So, should be:
```
    std::vector<char> buf(sz+1);
    snprintf(buf.data(), sz+1, format, args ...);
    send(l, tag, buf.data());
```

stsoe (changes requested):

src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.cpp
```diff
@@ -233,6 +233,13 @@ int AzureDev::azureLoadXclBin(const xclBin *buffer)
     if (fpgaSerialNumber.empty())
         return -E_EMPTY_SN;
     std::cout << "LoadXclBin FPGA serial No: " << fpgaSerialNumber << std::endl;
+
+    // check if the xclbin is valid
+    if (xclbin::get_axlf_section(buffer, BITSTREAM) != nullptr) {
+        std::cout << "xclbin is invalid, please provide azure xclbin" << std::endl;
+        return -EINVAL;
```
xuhz, line 240: I would suggest we define azure specific errno, like E_INVALID_XCLBIN.
EINVAL is also used in xocl driver, sometimes when we see this errno, we don't know it fails in xocl or plugin.

stsoe: What is an azure specific error code?  Any errors propagated to end user should be exception (if C++) or POSIX error code if (C).   If the error ends up in a syslog some place, the error code shouldn't matter, should it?

xuhz: the errno in the log should clearly tell what is wrong. The practice in azure plugin is, any failure in the plugin returns a specific defined error code, instead of the POSIX errno that are also used in xocl driver, so that, we can identify which part of the plugin fails by examining the log

stsoe: Got it.  I see the error codes defined in azure_rest_err.

AShivangi (approved):

src/runtime_src/core/pcie/linux/shim.cpp
rajkumar-xilinx: Thanks @stsoe , I completely agree with you. As a starting point, I have just replaced in once shim.cpp because I don't want to disturb already working code/setup. 

I have added fix in message.h file now.

PR #5887, opened 2021-10-01
CR-1112403: Reenable loading of HAL-level profile plugins

jvillarre:
Add hooks in profiling to load all of the HAL level plugins, regardless of the xrt_trace ini setting.  This restores the functionality from before the refactoring.


IshitaGhosh (approved):
Changes look good.

PR #5889, opened 2021-10-01
CR-1112545 use start, done, idle count instead of stalls

hackwa:


jvillarre (approved):
Change to mask looks good to do the appropriate functionality.

PR #5891, opened 2021-10-01
CR-1112537: Fix issue where trace was missing KERNEL_WRITE approximate ends

jvillarre:
In adding approximate data ends for kernel reads and kernel writes, the code was incorrectly using slot ID instead of trace ID, meaning that we were missing the kernel writes that were not tied down.


hackwa (approved):
looks good

PR #5894, opened 2021-10-03
CR-1112584 : Fix issue while single process tries to use multiple xclbin files

saifuddin-xilinx:
**Issue :** 
CR-1112584 : on vck190 dfx platform, kernel panic happens at the second AIE xclbin download when a single process tries to use multiple xclbin files
[https://jira.xilinx.com/browse/CR-1112584](url)

**Root Cause :** 
While loading the second xclbin, we are not resetting the AIE and AIE Daemon.
Without reset the AIE if we run another test it will not work. Hence either user application has to initiate a AIE Reset or XRT has to do that.

**Solution :** 
1. User application has to call deviceHandle.reset_array(); after a test to use same device for the xecond xclbin
2. XRT has to stop aie daemon before load second xclbin. 
3. XRT has to reset AIE device if user didn't do that. 

**Test :** 
With the above changes the tests mentioned in the CR are working fine;


xilinxgitops:
retest this please

PR #5898, opened 2021-10-04
CR-1112584: Fixing vck190 DFx issue where multiple xclbins loaded from same App (#5895)

chvamshi-xilinx:
(cherry picked from commit d0daf249f469db01650531cb9c285aea97c4bcab)

Issue:

1. Whenever same application is loading multiple xclbins, We have asked Application to call resetAieArray as XRT cannot reset AIE as it doesn't know whether current xclbin load would be successful or not. When Application is calling Aierest, then xclbin download failing by throwing an error about AIE reset multiple times.
2. AIED is trying to get some commands from zocl to fetch the AIE state which panics the kernel

Fix:

1. ZOCL is checking whether AIE is in reset state or not before calling AIE reset.
2. Added an INI option to completely disable AIED as changes in zocl would be risky at this time.

Verification:

1. Same app multiple xclbin downloads (AIE, PL, AIE+PL)
2. Different app multiple xclbin downloads (AIE, PL, AIE+PL)

Doc Update:

1. No need to document this option is VCK190_DFx is EA in this release

Reviewers: Soren/Saif
Approved by: Sumit

### DIFF
--- a/src/runtime_src/core/common/config_reader.h
+++ b/src/runtime_src/core/common/config_reader.h
@@ -560,6 +560,13 @@ get_enable_pr()
 }
 
 inline bool
+get_enable_aied()
+{
+  static bool value = detail::get_bool_value("Runtime.enable_aied",true);
+  return value;
+}
+
+inline bool
 get_multiprocess()
 {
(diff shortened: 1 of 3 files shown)


anupcs-xlnx (approved):
Approved

PR #5900, opened 2021-10-04
Profiling code cleanup

jvillarre:
This cleans up the OpenCL/profiling callback layer in the following ways:

- Removes unused code that calculated dependencies during callbacks
- Removes an unused const std::string& parameter in profiling event callbacks
- Removed dead code in the XDP side that was used in the old architecture
- Removed dead code for loading the profiling plugin (replaced with module_loader)


stsoe (approved):

hackwa (approved):

pgschuey (approved):
looks good

IshitaGhosh (approved):
Changes look good.

PR #5909, opened 2021-10-07
CR-1112934 : Reading the exported fd to local buffer using the linux …

venkatp-xilinx:
CR-1112934 : Reading the exported fd to local buffer using the linux 'read' API call instead of using the device to host API


dayeh-xilinx:
retest this please. Build killed by Jenkins.

PR #5911, opened 2021-10-07
Shutdown mgmt pf before updating SC (#5910)

AShivangi-zz:
_**Issue:**_ 
CR-1112544 ASTeR 17.01 - u55c/n - flashing can fail from the factory SC FW to the latest (7.1.14) if an xclbin is loaded

**_Issue solved:_**
New xbmgmt doesn't issue a hot reset which can brick the board in some scenarios. Eg: when 2 boards are flashed simultaneously 

**_Temporary solution:_**
Added linux specific code to reset mgmt pf. A full-blown OS-agnostic solution would require more time and architectural changes which we can't afford given we are nearing the release. (We have a CR tracking the bigger effort)

**_Tests:_**
1. Check dmesg for reset
```
...
[  +0.279017] xocl 0000:b3:00.1:  ffff96eb6b5700b0 xocl_hot_reset: resetting device...
...
[  +0.000012] xclmgmt 0000:b3:00.0: xclmgmt_reset_pci: Reset PCI
...
```
2. Remove userpf and make sure that sc flash gives an error
```
...
Report

Device flashing encountered errors:
ERROR: Only proceed with SC update if all user applications for the target card(s) are stopped.: Invalid argument
```

* shutdown mgmt pf before updating SC
* hot reset before updating SC

(cherry picked from commit 21f7ee18d10f5f6b1319383d6643d645d7d654b2)

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
@@ -31,6 +31,10 @@ namespace XBU = XBUtilities;
 #include "core/common/message.h"
 #include "core/common/utils.h"
 #include "flash/flasher.h"
+// Remove linux specific code
+#ifdef __GNUC__
+#include "core/pcie/linux/scan.cpp"
+#endif
 
 // 3rd Party Library - Include Files
 #include <boost/format.hpp>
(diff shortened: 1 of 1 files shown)


dayeh-xilinx:
retest this please. Could be a problem of the EDGE tests.

PR #5914, opened 2021-10-08
CR-1113125 Crash during xbutil validate -d on U.2

chienwei-lan:
Issue:
CR-1113125 Crash during xbutil validate -d on U.2

Issue solved:
ERT may try to wait a struct which is freed

Solution:
Rearrange the logic to fix the race condition issue


maxzhen:
I believe that this is the final solution, not a temporary one.

PR #5916, opened 2021-10-10
add XGQ user command type

mamin506:
What & why,
This PR is trying to define XGQ commands that will be used in XGQ based ERT (the ERT4 firmware).
The XGQ based ERT command would be used in between KDS and ERT firmware. It is not exposed to user (It could be but we are not planning to do it now).
This PR includes some code clean up, like prefix change and define command header structure. 

How,
1. Rename xgq_cmd.h to xgq_cmd_common.h, which defines the common part of XGQ spec.
2. Add xgq_cmd_user.h to define user command type struct of XGQ spec. It includes xgq_cmd_common.h. In this way, we are able to separately develop XGQ command set for VMR, ERT, etc..
3. Define submission and completion queue header struct and used in all of the command struct
4. Add static assertion for the size of some key data structures.

### DIFF
--- a/src/runtime_src/core/include/xgq_cmd_common.h
+++ b/src/runtime_src/core/include/xgq_cmd_common.h
@@ -62,7 +62,7 @@
 
   // Create an artifitial assertion via a bad divide by zero assertion.
   #define XGQ_STATIC_ASSERT(e,m) \
-    enum { XGQ_ASSERT_CONCAT(assert_line_, __LINE__) = 1/(int)(!!(e)) }
+    enum { XGQ_ASSERT_CONCAT(xgq_assert_line_, __LINE__) = 1/(int)(!!(e)) }
 #endif
 
 #define XGQ_SUB_Q1_SLOT_SIZE	512	// NOLINT


stsoe:
> I see we have some macro names prefixed by "XGQ_" and some by "XRT_". Can we have a unified prefix? Maybe we should replace "xrt_" with "xgq_" since these commands are used only with XGQ? So that we can have file names as "xgq_cmd_user.h", "xgq_cmd_mgmt.h" and "xgq_cmd_common.h"?

Yeah, I agree xgq_ prefix is better given that these are for use with XGQ.  If we can do as you suggest it would be great.

xdavidz:
> > I see we have some macro names prefixed by "XGQ_" and some by "XRT_". Can we have a unified prefix? Maybe we should replace "xrt_" with "xgq_" since these commands are used only with XGQ? So that we can have file names as "xgq_cmd_user.h", "xgq_cmd_mgmt.h" and "xgq_cmd_common.h"?
> 
> Maybe not "xgq_cmd_user.h", but "xgq_cmd_ert.h" since they are only meant for ERT to consume through XGQ. We may have XGQ connecting to other sub devices on user pf...

xgq_cmd_ert.h and xgq_cmd_vmr.h, @maxzhen would those 2 look ok?

xilinxgitops:
retest this please

PR #5921, opened 2021-10-12
Clarification of multiprocess buffer transfer requires proper IPC mechanism 

uday610:
Buffer export and import API gets the FD associated with the buffer, but that FD cannot be passed to another buffer as a raw data, rather a proper IPC mechanism is needed for FD translation. This fact in clarified in 2 places
   1) API documentation 
   2) ``xrt::bo`` constructor ``xrt::bo::export_bo`` header


uday610:
@stsoe , not necessarily needed to cherry-pick to 21.2... However, doing it no harm though as 21.2 is not officially released yet. But if it needs approval and all to merge it to 21.2, then let it be.

PR #5928, opened 2021-10-13
linking from one .rst file to another .rst file needed change to fix the link in html pages

uday610:
## Problem solved by the commit

There are many links in the HTML pages are not working with the sphinx version 1.8.5 

## How problem was solved, alternative solutions (if any) and why they were rejected

Mainly the changes are the formatting as below

Whenever linking to another RST file :ref: is changed to :doc:  and also .rst extension is removed.


xilinxgitops:
retest this please

PR #5962, opened 2021-10-29
Enabled the usages of AIE XGRAPH Calls even for PCIe targets. Ideally…

venkatp-xilinx:
Enabled the usages of AIE XGRAPH Calls even for PCIe targets. Ideally these are used only for the sw_emu target, for hw and hw_emu targets, it issues a runtime error.


stsoe:
I would like to request that this PR be broken into enabling aie graph calls across the supported platforms with one copy of aie_graph.cpp stubs used in all shims.  Then a second pull request that implements the stubs for sw_emu on x86.   Let's keep pull requests as simple as possible and decouple this one from sw_emu.  Thanks!

PR #5964, opened 2021-11-01
Enabled the usage of AIE XGRAPH Calls even for PCIe targets.

venkatp-xilinx:
New enhancement done as part of the VITIS-3745 Running PS APP as x86 process
Enable AIE shim level calls by default on x86 in preparation for supporting x86 software emulation of graph host code.

Add stub AIE xcl stub implementation that is compiled into all shim core libraries
Enable x86 definition of AIE graph level calls.   In this pull request the stubs will throw a runtime exception.  A subsequent pull request will implement the stubs for x86 software emulation.
 
No risks as the enabled code should not be called at all.

New framework enabled, did not add any tests.

### DIFF
--- a/src/runtime_src/core/common/CMakeLists.txt
+++ b/src/runtime_src/core/common/CMakeLists.txt
@@ -38,10 +38,6 @@ file(GLOB XRT_CORE_COMMON_LIB_FILES
   "api/*.cpp"
   )
 
-if (${XRT_NATIVE_BUILD} STREQUAL "yes")
-  add_definitions (-DXRT_ENABLE_AIE -DXRT_AIE_BUILD)
-endif()
-
 if (DEFINED XRT_AIE_BUILD)
   file(GLOB XRT_CORE_COMMON_AIE_LIB_FILES
     "api/aie/*.cpp"
--- a/src/runtime_src/core/common/CMakeLists.txt
+++ b/src/runtime_src/core/common/CMakeLists.txt
@@ -38,12 +38,6 @@ file(GLOB XRT_CORE_COMMON_LIB_FILES
   "api/*.cpp"
   )
 
-if (${XRT_NATIVE_BUILD} STREQUAL "yes")
-  set(XRT_AIE_BUILD "yes")
-  set(XRT_ENABLE_AIE "yes")
-  add_definitions (-DXRT_ENABLE_AIE -DXRT_AIE_BUILD)
-endif()
-
 if (DEFINED XRT_AIE_BUILD)
   file(GLOB XRT_CORE_COMMON_AIE_LIB_FILES
     "api/aie/*.cpp"
--- a/src/runtime_src/core/common/CMakeLists.txt
+++ b/src/runtime_src/core/common/CMakeLists.txt
@@ -38,10 +38,16 @@ file(GLOB XRT_CORE_COMMON_LIB_FILES
   "api/*.cpp"
   )
 
+if (${XRT_NATIVE_BUILD} STREQUAL "yes")
+  set(XRT_AIE_BUILD "yes")
+  set(XRT_ENABLE_AIE "yes")
+  add_definitions (-DXRT_ENABLE_AIE -DXRT_AIE_BUILD)
+endif()
+
 if (DEFINED XRT_AIE_BUILD)
   file(GLOB XRT_CORE_COMMON_AIE_LIB_FILES
     "api/aie/*.cpp"
-  )
+    )
   list(APPEND XRT_CORE_COMMON_LIB_FILES ${XRT_CORE_COMMON_AIE_LIB_FILES})
 endif()
 
--- a/src/runtime_src/core/common/CMakeLists.txt
+++ b/src/runtime_src/core/common/CMakeLists.txt
@@ -38,10 +38,14 @@ file(GLOB XRT_CORE_COMMON_LIB_FILES
   "api/*.cpp"
   )
 
+if (${XRT_NATIVE_BUILD} STREQUAL "yes")
+  add_definitions (-DXRT_ENABLE_AIE -DXRT_AIE_BUILD)
+endif()
+
 if (DEFINED XRT_AIE_BUILD)
   file(GLOB XRT_CORE_COMMON_AIE_LIB_FILES
     "api/aie/*.cpp"
-    )
+  )
   list(APPEND XRT_CORE_COMMON_LIB_FILES ${XRT_CORE_COMMON_AIE_LIB_FILES})
 endif()
 
--- a/src/runtime_src/core/include/CMakeLists.txt
+++ b/src/runtime_src/core/include/CMakeLists.txt
@@ -13,6 +13,7 @@ set(XRT_HEADER_SRC
   xrt.h
   xcl_app_debug.h
   xcl_macros.h
+  xcl_graph.h
   xcl_app_debug.h
   xclperf.h
   xclhal2_mem.h
--- a/src/runtime_src/core/include/CMakeLists.txt
+++ b/src/runtime_src/core/include/CMakeLists.txt
@@ -13,7 +13,6 @@ set(XRT_HEADER_SRC
   xrt.h
   xcl_app_debug.h
   xcl_macros.h
-  xcl_graph.h
   xcl_app_debug.h
   xclperf.h
   xclhal2_mem.h
--- a/src/runtime_src/core/pcie/common/aie_stubs.cpp
+++ b/src/runtime_src/core/pcie/common/aie_stubs.cpp
@@ -1,6 +1,3 @@
-#ifndef __AIE_STUBS_H__
-#define __AIE_STUBS_H__
-
 #include <sstream>
 
 #include "core/include/experimental/xrt_aie.h"
@@ -143,7 +140,5 @@ xclLoadXclBinMeta(xclDeviceHandle /*handle*/, const xclBin* /*buffer*/)
 {
   throw std::runtime_error(function_not_implemented(__func__));
 }
-
-#endif
 
 //////////// END OF XCL XRT AIE APIS///////////////////////////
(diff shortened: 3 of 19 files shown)


chvamshi-xilinx:
Please add Description of this PR.

venkatp-xilinx:
> Please add Description of this PR.

Add the description as requested

chvamshi-xilinx (approved):
Looks much cleaner now.

stsoe (changes requested):
Hi Prasad.  This looks very simple now.  Thanks for changing.

I have one request regarding the aie_stubs.h file you added under core/common.  I apologize for not seeing this sooner.  

I am realizing only now, that the stubs are for the pcie platforms, alas please move aie_stubs.h to core/common/pcie/aie_stubs.cpp and change the pcie/{linux,windows,emulation???}/CMakeLists.txt files to compile aie_stubs.cpp just like it is compiling other files from pcie/common.  

You'd also have to change pcie/common/CMakeLists.txt to exclude aie_stubs.cpp like some other files are.   The hack in pcie/common/CMakeLists.txt files could be simplified by simply explicitly listing the files that should go into XRT_CORE_PCIE_COMMON_FILES, but feel free to continue the current hack.

The reason I prefer this compilation of the stubs is that it is even less intrusive on the code, as no shims have to explicitly include the stubs and the stubs remain defined in a .cpp file with no side effects at all (like implicit includes).

Please declare execpStr in aie_stubs.cpp as static.

venkatp-xilinx:
> Hi Prasad. This looks very simple now. Thanks for changing.
> 
> I have one request regarding the aie_stubs.h file you added under core/common. I apologize for not seeing this sooner.
> 
> I am realizing only now, that the stubs are for the pcie platforms, alas please move aie_stubs.h to core/common/pcie/aie_stubs.cpp and change the pcie/{linux,windows,emulation???}/CMakeLists.txt files to compile aie_stubs.cpp just like it is compiling other files from pcie/common.
> 
> You'd also have to change pcie/common/CMakeLists.txt to exclude aie_stubs.cpp like some other files are. The hack in pcie/common/CMakeLists.txt files could be simplified by simply explicitly listing the files that should go into XRT_CORE_PCIE_COMMON_FILES, but feel free to continue the current hack.
> 
> The reason I prefer this compilation of the stubs is that it is even less intrusive on the code, as no shims have to explicitly include the stubs and the stubs remain defined in a .cpp file with no side effects at all (like implicit includes).
> 
> Please declare execpStr in aie_stubs.cpp as static.

Implemented all the changes suggested. Please review and approve it

stsoe (approved):

stsoe:
Hi @venkatp-xilinx  Please run some basic examples and debug why they are not working.
Part of the problem is that you cannot just simply enable XRT_ENABLE_AIE without examining what code actually ends up getting built.  I believe XRT_ENABLE_AIE and maybe XRT_AIE_BUILD result in different code path, it is definitely more than just enabling graph xcl functions.

src/runtime_src/core/common/CMakeLists.txt
```diff
@@ -38,10 +38,14 @@ file(GLOB XRT_CORE_COMMON_LIB_FILES
   "api/*.cpp"
   )
 
+if (${XRT_NATIVE_BUILD} STREQUAL "yes")
+  add_definitions (-DXRT_ENABLE_AIE -DXRT_AIE_BUILD)
+endif()
+
```
stsoe, line 44: I think this is the reason for the runtime failures observed in the pipeline.  You should run and debug a few examples to understand the implications on enabling these switches on x86.   For example load_xclbin takes a very different path.

stsoe (changes requested):

src/runtime_src/core/common/CMakeLists.txt
venkatp-xilinx: Enabling the definitions under core/common, I am under impression that it even for the core/pcie even. If thats not the case, we need to explicitly enable it inside core/pcie even. And you suggested to use src/CMakeLists.txt to add these definitions, wondering by then we have the XRT_NATIVE_BUILD is set appropriately.

venkatp-xilinx: resulting CMake errors when used the add_compile_definitions, I dont know if this is supported with the CMake version I have.

src/runtime_src/core/common/CMakeLists.txt
```diff
@@ -41,7 +41,7 @@ file(GLOB XRT_CORE_COMMON_LIB_FILES
 if (DEFINED XRT_AIE_BUILD)
   file(GLOB XRT_CORE_COMMON_AIE_LIB_FILES
     "api/aie/*.cpp"
-    )
+  )
```
stsoe, line 44: Aren't api/aie/*.cpp supposed to be built?

stsoe (commented):
@venkatp-xilinx question about api/aie/*.cpp

src/runtime_src/core/common/CMakeLists.txt
venkatp-xilinx: Its just a space change, not removed it Soren

venkatp-xilinx: we have added the define XRT_AIE_BUILD in src/CMakeLists.txt when XRT_NATIVE_BUILD is set to yes. So it ideally builds even for the x86

stsoe: No it is not.   You didn't set XRT_AIE_BUILD for CMake, you set it for g++.

venkatp-xilinx: Thanks for correcting me Soren, I did not realize its only for GCC, Let me correct as you said and verify the changes before submit.

stsoe: No problem.   Yes, please validate a few examples before pushing the changes again.  You are close now :-)

src/runtime_src/core/include/CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ set(XRT_HEADER_SRC
   xrt.h
   xcl_app_debug.h
   xcl_macros.h
+  xcl_graph.h
```
stsoe, line 16: No, this header should not be exported.  We don't want end applications to use xcl APIs.

stsoe (changes requested):

src/runtime_src/core/include/CMakeLists.txt
venkatp-xilinx: Removed it from the include list

stsoe:
@venkatp-xilinx The windows build failed because xrt_graph.cpp has two references to uuid_t which should be changed to xuid_t in order for windows to compile.   The change would match the declaration in xrt_graph.h which is using xuid_t.

xilinxgitops:
retest this please

stsoe (changes requested):

stsoe (changes requested):
One link error and some warnings on windows.  See comment in aie_stubs.cpp for link error.

```
Please make sure all these warnings are eliminated:
03:52:28   xrt_graph.obj : warning LNK4217: locally defined symbol ??0bo@xrt@@QEAA@PEAX@Z (public: __cdecl xrt::bo::bo(void *)) imported in function "int __cdecl xrtSyncBOAIENB(void *,void *,char const *,enum xclBOSyncDirection,unsigned __int64,unsigned __int64)" (?xrtSyncBOAIENB@@YAHPEAX0PEBDW4xclBOSyncDirection@@_K3@Z) [R:\builds\xrt\aie_pcie_enablement\10018\build\WRelease\runtime_src\core\common\xrt_coreutil.vcxproj]
03:52:28   xrt_graph.obj : warning LNK4217: locally defined symbol xrtDeviceOpen imported in function xrtAIEDeviceOpen [R:\builds\xrt\aie_pcie_enablement\10018\build\WRelease\runtime_src\core\common\xrt_coreutil.vcxproj]
03:52:28   xrt_graph.obj : warning LNK4217: locally defined symbol ?send_exception_message@xrt_core@@YAXPEBD0@Z (void __cdecl xrt_core::send_exception_message(char const *,char const *)) imported in function __catch$?xrtAIEReadProfiling@@YA_KPEAXH@Z$0 [R:\builds\xrt\aie_pcie_enablement\10018\build\WRelease\runtime_src\core\common\xrt_coreutil.vcxproj]
03:52:28   xrt_graph.obj : warning LNK4217: locally defined symbol ?send@message@xrt_core@@YAXW4severity_level@12@PEBD1@Z (void __cdecl xrt_core::message::send(enum xrt_core::message::severity_level,char const *,char const *)) imported in function __catch$?xrtAIEReadProfiling@@YA_KPEAXH@Z$1 [R:\builds\xrt\aie_pcie_enablement\10018\build\WRelease\runtime_src\core\common\xrt_coreutil.vcxproj]
03:52:28 
```

src/runtime_src/core/pcie/common/aie_stubs.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+static std::string function_not_implemented(const char* funcName) {
+  std::stringstream excepStreamMsg;
+  excepStreamMsg << funcName << " is not supported for hw and hw_emu PCIe targets";
+  std::string exceptionStr(excepStreamMsg.str());
+  return exceptionStr;
+}
+
+void*
+xclGraphOpen(xclDeviceHandle /*handle*/, const uuid_t /*xclbin_uuid*/, const char* /*graph*/, xrt::graph::access_mode /*am*/)
```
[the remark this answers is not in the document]
venkatp-xilinx: Thanks Soren. I don't have windows setup locally to build PCIe XRT. Now I have the setup to run the windows build. Corrected this and ran the XRT build in the local windows machine, I could see 0 Errors and 0 Warnings now. So I assume it will be resolved now.

stsoe (approved):

PR #5966, opened 2021-11-01
pkgapu.sh: fix mkimage path

houlz0507:
Incorrect env variable dependency.

### DIFF
--- a/src/runtime_src/tools/scripts/pkgapu.sh
+++ b/src/runtime_src/tools/scripts/pkgapu.sh
@@ -31,8 +31,8 @@ usage()
         echo "          -output                         output path"
 	echo "This script requires tools: mkimage, xclbinutil, bootgen, rpmbuild, dpkg-deb. "
 	echo "There is mkimage in petalinux build, e.g."
-	echo "/proj/petalinux/2021.2/petalinux-v2021.2_daily_latest/tool/petalinux-v2021.2-final/"
-	echo "components/yocto/buildtools/sysroots/x86_64-petalinux-linux/usr/bin/mkimage"
+	echo "/proj/petalinux/2021.2/petalinux-v2021.2_daily_latest/tool/petalinux-v2021.2-final/components/yocto/buildtools/sysroots/x86_64-petalinux-linux/usr/bin/mkimage"
+	echo ""
 }
 
 usage_and_exit()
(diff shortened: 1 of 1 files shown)


src/runtime_src/tools/scripts/pkgapu.sh
```diff
@@ -27,10 +27,12 @@ usage()
 	echo "  options:"
 	echo "          -help                           Print this usage"
 	echo "          -images                         Versal images path"
-	echo "          -petalinux                      Petalinux path"
 	echo "          -clean                          Remove build files"
         echo "          -output                         output path"
-	echo ""
+	echo "This script requires tools: mkimage, xclbinutil, bootgen, rpmbuild, dpkg-deb. "
```
rozumx, line 32: You might wish to check to see if the XILINX_XRT and XILINX_VITIS environment variables are set OR do a check to see if these applications are found with the `which` command.

rozumx (approved):
Approved with comments.

src/runtime_src/tools/scripts/pkgapu.sh
houlz0507: Added which command to check tools.

src/runtime_src/tools/scripts/pkgapu.sh
```diff
@@ -154,8 +156,8 @@ if [[ ! -d $IMAGES_DIR ]]; then
 fi
 IMAGES_DIR=`realpath $IMAGES_DIR`
 
-if [ -f $SETTINGS_FILE ]; then
-	source $SETTINGS_FILE
+if [[ ! (`which mkimage` && `which bootgen` && `which xclbinutil`) ]]; then
+	error "Please source Xilinx VITIS and Petalinux tools to make sure mkimage, bootgen and xclbinutil is accessible."
```
rbramand-xilinx, line 160: @houlz0507 sourcing petalinux will not set mkimage path in PATH variable, so mkimage will not be picked automatically. That is the reason why I have hardcoded mkimage path earlier.

so users of this script can either install mkimage using **yum install -y uboot-tools** in centos/rhel machine to get mkimage binary or add petalinux mkimage path to PATH variable to autopick mkimage.

should we add setting of PATH variable in error message or it is okay?

rbramand-xilinx (approved):

src/runtime_src/tools/scripts/pkgapu.sh
houlz0507: The mkimage path is shown in error message
Usage:  [options]
  options:
          -help                           Print this usage
          -images                         Versal images path
          -clean                          Remove build files
          -output                         output path
This script requires tools: mkimage, xclbinutil, bootgen, rpmbuild, dpkg-deb. 
There is mkimage in petalinux build, e.g.
/proj/petalinux/2021.2/petalinux-v2021.2_daily_latest/tool/petalinux-v2021.2-final/components/yocto/buildtools/sysroots/x86_64-petalinux-linux/usr/bin/mkimage

rbramand-xilinx: looks good, I will update the BAI with this change. Thanks.

stsoe (approved):

PR #5968, opened 2021-11-02
Vitis-3290 AWS F1 NoDMA Small Shell - XRT Support

xuhz:
@xdavidz please review the changes in icap.c
since m2m is created during xclbin is loaded, the mem topology and group metadata have to be cached after the m2m is created. Please review whether the sequence change will bring any issues.
@houlz0507 
there is change in interface-uuid check since is this case, xclbin has partition metadata, while the shell doesn't. for aws nodma xclbin, there is no interface-uuid in the xclbin partition metadata

1. add xrt support to nodma small shell on aws f1
2. add device id and/or shell change xrt support on aws f1 
1. xclbin for nodma shell is built with raptor flow which has partition metadata, nodma shell on aws is built by aws, there is no metadata at all. xrt needs to handle this case where the m2m and slave bridge IPs are in xclbin metadata
2. CR-1113151 device id change support in aws f1
It is a known issue that we don't support device id change for xdma shell. we loaded a default agfi to change the device id before we open mailbox. once mailbox is opened, we don't allow device id change then
It is a more desired requirement to support device id change in xrt when we have nodma shell support since nodma shell will have a different device id than xdma shell
1. m2m and slavebridge subdevices metadata are in xclbin, and they will be created during xclbin download
2. on aws f1, device id change needs a pcie removal and rescan during xclbin load, which will unload driver, this will make the xclbin load fail.
the fix is in aws plugin, if removal and rescan is required during xclbin load, it returns a EAGAIN  to xrt shim, and at the same time do the removal&rescan in a separate thread. The shim, when seeing the EAGAIN, doesn't return failure to user, it will reload the xclbin later, in which case, the device id has been changed already.  
with the fix, xrt doesn't depend on a default agfi again.
xrt will not return EAGAIN for other alveo cards during xclbin load, so the change doesn't have any impact or behavior change to non-aws case
xdma and nodma test on aws f1
alveo u50 nodma test

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
@@ -2463,14 +2463,6 @@ static int __icap_download_bitstream_user(struct platform_device *pdev,
 
 	xocl_subdev_destroy_by_level(xdev, XOCL_SUBDEV_LEVEL_URP);
 
-	/* TODO: link this comment to specific function in xocl_ioctl.c */
-	/* has to create mem topology even with failure case
-	 * please refer the comment in xocl_ioctl.c
-	 * without creating mem topo, memory corruption could happen
-	 */
-	icap_cache_bitstream_axlf_section(pdev, xclbin, MEM_TOPOLOGY);
-	icap_cache_bitstream_axlf_section(pdev, xclbin, ASK_GROUP_TOPOLOGY);
-
 	err = __icap_peer_xclbin_download(icap, xclbin, force_download);
 
 	/* TODO: Remove this after new KDS replace the legacy one */
@@ -2501,14 +2493,22 @@ static int __icap_download_bitstream_user(struct platform_device *pdev,
 
 	icap_create_subdev_debugip(pdev);
 
-	icap_cache_max_host_mem_aperture(icap);
-
 	/* Initialize Group Topology and Group Connectivity */
 	icap_cache_bitstream_axlf_section(pdev, xclbin, ASK_GROUP_CONNECTIVITY);
 
 	icap_probe_urpdev_all(pdev, xclbin);
 	xocl_subdev_create_by_level(xdev, XOCL_SUBDEV_LEVEL_URP);
 done:
+	/* TODO: link this comment to specific function in xocl_ioctl.c */
+	/* has to create mem topology even with failure case
+	 * please refer the comment in xocl_ioctl.c
+	 * without creating mem topo, memory corruption could happen
+	 */
+	icap_cache_bitstream_axlf_section(pdev, xclbin, MEM_TOPOLOGY);
+	icap_cache_bitstream_axlf_section(pdev, xclbin, ASK_GROUP_TOPOLOGY);
+
+	icap_cache_max_host_mem_aperture(icap);
+
 	if (err) {
 		uuid_copy(&icap->icap_bitstream_uuid, &uuid_null);
 	} else {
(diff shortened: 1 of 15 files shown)


src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1464,7 +1464,25 @@ int shim::xclLoadAxlf(const axlf *buffer)
...
     int ret = mDev->ioctl(mUserHandle, DRM_IOCTL_XOCL_READ_AXLF, &axlf_obj);
-    if(ret)
+    if (ret && errno == EAGAIN) {
+        //special case for aws
+        //if EAGAIN is seen, that means a pcie removal&rescan is ongoing, let's just
+        //wait and reload 2nd time -- this time the there will be no device id
+        //change, hence no pcie removal&rescan, anymore
+        int dev_hotplug_done = 0;
+        std::string err;
+        dev_fini();
```
stsoe, line 1474: Without really being able to say why, I don't think dev_ini() and  dev_fini() are safe outside constructor and destructor.  I have never seen these called arbitrarily as is done here.

src/runtime_src/core/pcie/tools/cloud-daemon/aws/aws_dev.cpp
```diff
@@ -451,15 +454,61 @@ int awsReadP2pBarAddr(size_t index, const xcl_mailbox_p2p_bar_addr *addr, int *r
     return ret;
 }
 
+#ifndef INTERNAL_TESTING_FOR_AWS
+static void awsPciRescan(int index)
```
stsoe, line 458: If this is C++ code, then shouldn't it be refactored and not use gotos.

stsoe (changes requested):

src/runtime_src/core/pcie/linux/shim.cpp
xuhz: there is a hotplug happening here, so we have to close the opened device otherwise the unloading of the driver will hang.
same logic in this file also applies to reset, p2p config
I will add more comments here

stsoe (approved):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
```diff
@@ -2982,19 +2977,17 @@ static int icap_cache_bitstream_axlf_section(struct platform_device *pdev,
...
-			if (!xocl_m2m_host_bank(xdev, &hbase, &hsz)) {
-				mem_topo->m_mem_data[i].m_used = 1;
-				mem_topo->m_mem_data[i].m_base_address = hbase;
-				mem_topo->m_mem_data[i].m_size = (hsz >> 10);
-			}
+			xocl_m2m_host_bank(xdev,
+				&(mem_topo->m_mem_data[i].m_base_address),
+				&(mem_topo->m_mem_data[i].m_size),
+				&(mem_topo->m_mem_data[i].m_used));
 		}
```
xdavidz, line 2991: I don't recall I added this piece of code, please have @houlz0507 review this login if it is correct.

xuhz: this part need to change since the host bank info may come from shell (u50 nodma case) or the xclbin itself (there is no shell metadata at all)
this pr moves the 	icap_cache_bitstream_axlf_section(pdev, xclbin, ASK_GROUP_TOPOLOGY) to after the m2m subdevice is created.
please think about it whether it has issue.

xdavidz: This sequence should be fine during normal driver attach. But I recall the icap won't be teared down (remove/probe) during reset, instead the icap has offline/online callbacks. Please double check if "reset" also works as expected.

xuhz: this change is only related to xclbin download. I don't think it will impact reset.
xclbin needs to be reloaded anyway after reset

houlz0507 (approved):

dayeh-xilinx:
retest this please. Fixed an issue with the new boardfarm test script.

maxzhen (approved):

PR #5973, opened 2021-11-06
Leverage xrt::xclbin abstraction in OpenCL

stsoe:
This PR continues #5969, with elimination of the majority of OpenCL
xclbin metadata layer.  In particular all old symbol references now
are entirely based on xrt::xclbin.

Risks are same as for #5969.  Local testing on cached OpenCL HW tests
has been done, but does not cover all cases.

### DIFF
--- a/src/runtime_src/xdp/profile/plugin/device_offload/opencl/opencl_device_offload_plugin.cpp
+++ b/src/runtime_src/xdp/profile/plugin/device_offload/opencl/opencl_device_offload_plugin.cpp
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2020 Xilinx, Inc
+ * Copyright (C) 2020-2021 Xilinx, Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License"). You may
  * not use this file except in compliance with the License. A copy of the
@@ -21,6 +21,7 @@
 // Includes from xrt_coreutil
 #include "core/common/system.h"
 #include "core/common/message.h"
+#include "core/common/xclbin_parser.h"
 
 // Includes from xilinxopencl
 #include "xocl/core/platform.h"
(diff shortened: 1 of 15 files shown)


src/runtime_src/xdp/profile/plugin/device_offload/opencl/opencl_device_offload_plugin.cpp
```diff
@@ -338,11 +338,15 @@ namespace xdp {
 
         // Construct the argument list of each port
         std::string arguments = "" ;
-        for (auto arg : matchingCU->get_symbol()->arguments)
+        for (const auto& arg : matchingCU->get_args())
         {
-          if ((arg.address_qualifier != 1 && arg.address_qualifier != 4) ||
-              arg.atype != xocl::xclbin::symbol::arg::argtype::indexed)
-            continue ;
```
stsoe, line 345: @hackwa This is repeated few times below but does not look right.  I check with @jvillarre and he wasn't sure.
Skipping non-indexed args is fine. But skipping any args where address_qualifier is different from '1' (global) and '4' (stream) doesn't make sense to me?  I think the intend is to skip non-memory banked arguments, which are actually '1' (global) and '2' (constant) ?

The changes in this PR preserves the old behavior, but please review and confirm.

stsoe:
retest this please - cmake failed, infra

src/runtime_src/xdp/profile/plugin/device_offload/opencl/opencl_device_offload_plugin.cpp
hackwa: hey Soren this looks good to me.

jvillarre (approved):
Changes to profiling code look good.

PR #5977, opened 2021-11-09
CR-1100237: Reducing memory usage when processing large amounts of trace

jvillarre:
The profiling library took excessive amounts of memory when processing large trace buffers offloaded at the end of application execution.

This pull request now does two things:
1.  Removes the usage of the traceResults structure and combines FIFO offload and TS2MM offload processing into a single code path.
2. Forces dumping to disk when memory used to store device events exceeds a threshold.

This change has been tested on the large design that originally exposed the problem as well as a small set of test cases.  Full testing on larger suites will be needed after merging.

### DIFF
--- a/src/runtime_src/xdp/profile/database/database.h
+++ b/src/runtime_src/xdp/profile/database/database.h
@@ -40,7 +40,7 @@ namespace xdp {
   {
   public:
     // For messages sent to specific plugins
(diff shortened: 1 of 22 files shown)


hackwa (approved):

IshitaGhosh (approved):
Changes look good

PR #5979, opened 2021-11-10
CR-1114813: Removing loading XDP plugins or enabling XDP functionality with deprecated switches

jvillarre:
Since 2021.1, the xrt.ini switches "profile," "timeline_trace," and "xrt_profile" have issued deprecation messages when they were used to enable XDP functionality.  This pull request disables loading XDP plugins based on these switches and removes the deprecation messages.

The "profile" switch still has one aspect that needs to be called out as deprecated and removed in a future release.  "profile=true" will still add the CL_QUEUE_PROFILING_ENABLE flag to all command_queue objects created in OpenCL.


stsoe (approved):

hackwa (approved):
looks good! Maybe we can give a heads up to DSV if the testing is using the old flags.

IshitaGhosh (approved):
Change looks good.

dayeh-xilinx:
retest this please. u50 test failure may be caused by new boardfarm script.

PR #5980, opened 2021-11-10
Enabled the usage of AIE XGRAPH Calls even for PCIe targets. (#5964)

stsoe:
Enable the usage of AIE XGRAPH Calls even for PCIe targets.

New enhancement done as part of the VITIS-3745 Running PS APP as x86
process Enable AIE shim level calls by default on x86 in preparation
for supporting x86 software emulation of graph host code.

Add stub AIE xcl stub implementation that is compiled into all shim
core libraries.  In this pull request the stubs will throw a runtime
exception. A subsequent pull request will implement the stubs for x86
software emulation.

No risks as the enabled code should not be called at all.

New framework enabled, did not add any tests.

Cherry picked from commit 48928c8 with modifications.

Co-authored-by: Aleti Venkata Prasad <venkatp@xilinx.com>

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -49,6 +49,15 @@ if (${XRT_NATIVE_BUILD} STREQUAL "no")
   set (CMAKE_INSTALL_PREFIX "/usr")
 endif()
 
+# Enable AIE even on x86.  This is POC, once
+# complete we will remove the need for these
+# macros and defines.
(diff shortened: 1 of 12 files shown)


chvamshi-xilinx (approved):

larry9523 (approved):

jvillarre (approved):
Additional guard on XDP side looks good.

PR #5981, opened 2021-11-11
Remove fpga-mgr dependency for xclmgmt loading

rbramand-xilinx:
> We currently rely on CONFIG_FPGA configuration to check whether fpga-mgr module support is present or not
> But in some Kernel even though this config is enabled fpga-mgr module is not available which is causing xclmgmt driver loading to fail.
> Commented out adding fpga-mgr dependency for xclmgmt as it is not used anywhere at present.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/Makefile
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/Makefile
@@ -105,7 +105,10 @@ ifeq ($(SYMBOL),1)
 	ccflags-y += -g
 endif
 
-ccflags-$(CONFIG_FPGA) += -DENABLE_FPGA_MGR_SUPPORT
+# We can only rely on CONFIG_FPGA to check whether system has fpga-mgr module support or not.
+# But it is not true in some Kernels where this config is enabled but fpga-mgr module is not available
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx (approved):

stsoe (commented):
> We currently rely on CONFIG_FPGA configuration to check whether fpga-mgr module support is present or not
> But in some Kernel even though this config is enabled fpga-mgr module is not available which is causing
> xclmgmt driver loading to fail.

Isn't this an error in the target system kernel .config file.  Why are we working around a system problem?

> Commented out adding fpga-mgr dependency for xclmgmt as it is not used anywhere at present.

It's used conditionally in subdev/fmgr.c so this comment isn't accurate.  Please be more specific.

maxzhen:
IIRC, fpga manager code is added by @sonals. It is not really used today. If this is preventing our driver from loading, maybe we should remove fpga manager code completely. I'm adding @sonals and @houlz0507 as reviewer for their input.

maxzhen (approved):

houlz0507 (approved):

anupcs-xlnx (approved):

chvamshi-xilinx:
@stsoe , Can you please approve based on email conversations,

PR #5991, opened 2021-11-12
Separate old kds related code to sched_exec file

saifuddin-xilinx:
This is a kds cleanup PR. No logical changes are done. Just restructured the code so that old kds related code is in a same place.
The changes are as follows:
- Move all old kds related code to sched_exec file 
- Create a separate function for some of the old kds flat code and move it to sched_exec file 
- From other files, only function calls are invoked based on kds mode. 

Testing :
- New kds are working fine. 
- Old kds are not working top of the master XRT


mamin506 (approved):
Look good to me.

chvamshi-xilinx (approved):

PR #5992, opened 2021-11-12
Coverity fix: Adding a missing break in switch statement

jvillarre:
Coverity detected a switch statement in recently modified code that did not have a break, resulting in fall through behavior that could result in empty device trace files being generated.

The missing break was introduced in pull request 5977 on November 11 and was discovered via Coverity scan.


hackwa (approved):

IshitaGhosh (approved):
Looks good

PR #5994, opened 2021-11-14
Add domain to kernel cu index

stsoe:
In preparation for xrt::kernel support for pskernel, the compute unit
index has to differentiate between pl and ps kernels.  Rather than
adhoc add 128 (maxcu) to offset the ps kernel index, this PR
introduces the concept of a CU domain.

The CU domain is owned by xocl kernel module.  The domain is different
for a PL kernel CU (cu) and a PS kernel CU (scu), and it is encoded in
the index returned by the driver through sysfs.

In userspace, the encoded index is used when opening a context (shared
or exclusive) on the CU.  The index within a domain (the domain index)
is used in the cumask of the command packet sent to KDS.  The domain
index is obtained by stripping off the domain from the encoded index.

With this change in place, xrt::kernel will be remain more or less
unchanged in the code path that processes and opens compute units
associated with the kernel.

While transitioning to using encoded cu indices, the core device class
adds a new data member which is a mapping from CU name to encoded
index.  Following PRs will make use of this mapping when constructing
xrt::kernel objects and opening contexts, but in this PR the mapping
is unused.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -47,6 +47,7 @@ usage()
     echo "[-noctest]                  Skip unit tests"
     echo "[-with-static-boost <boost> Build binaries using static linking of boost from specified boost install"
     echo "[-clangtidy]                Run clang-tidy as part of build"
+    echo "[-pskernel]                 Enable building of POC ps kernel"
     echo "[-docs]                     Enable documentation generation with sphinx"
     echo "[-j <n>]                    Compile parallel (default: system cores)"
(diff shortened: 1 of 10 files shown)


stsoe:
retest this please - there are no logs for failures?

jeffli-xilinx (approved):
Looks good to me!

mamin506 (approved):
LGTM

PR #5999, opened 2021-11-15
CR-1114812 and CR-1114810: Profiling flag streamlining

jvillarre:
Deprecating confusing profiling options and replacing them with clearer options more aligned to the new profiling architecture's model.

Newly deprecated switches now issue warnings when used and the changelog has been updated to reflect the switch changes.

### DIFF
--- a/src/runtime_src/core/common/config_reader.h
+++ b/src/runtime_src/core/common/config_reader.h
@@ -306,9 +306,9 @@ get_opencl_device_counter()
 }
 
 inline bool
-get_device_counter()
+get_device_counters()
 {
-  static bool value = detail::get_bool_value("Debug.device_counter", false);
+  static bool value = detail::get_bool_value("Debug.device_counters", false);
   return value;
 }
 
--- a/src/runtime_src/core/common/config_reader.h
+++ b/src/runtime_src/core/common/config_reader.h
@@ -150,6 +150,16 @@ get_data_transfer_trace()
 }
 
 inline std::string
+get_data_transfer_trace_dep_message()
+{
+  static bool emitted = false ;
+  if (!emitted) {
+    return "The xrt.ini flag \"data_transfer_trace\" is deprecated and will be removed in a future release.  Please use the equivalent flag \"device_trace.\"" ;
+  }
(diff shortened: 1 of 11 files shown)


src/runtime_src/core/common/config_reader.h
```diff
@@ -156,6 +149,24 @@ get_data_transfer_trace()
   return value;
 }
 
+inline std::string
+get_data_transfer_trace_dep_message()
+{
+  static bool emitted = false ;
+  if (!emitted) {
```
stsoe, line 156: This is not thread safe, not sure if it matters, probably doesn't?

stsoe (approved):

src/runtime_src/core/common/config_reader.h
jvillarre: This function is added specifically for a problem we have right now where the OpenCL level device offload and the hardware level device offload are both being loaded and the deprecation message can appear twice.  This happens in a single thread at different times, so thread safety isn't necessary.  Additionally, we are working on delivering a change in 2022.1 where device offload is only loaded at the hardware level once we support device level information on the hardware emulation side without going through the OpenCL level, at which point this function can be removed and the deprecation message moved into the appropriate loaders.

src/runtime_src/core/common/config_reader.h
```diff
@@ -303,6 +305,13 @@ get_opencl_device_counter()
   return value;
 }
 
+inline bool
+get_device_counter()
+{
+  static bool value = detail::get_bool_value("Debug.device_counter", false);
```
hackwa, line 311: @jvillarre @pgschuey would "device_counters" be more accurate then "device_counter"?

dayeh-xilinx:
retest this please. Fixed boardfarm test issue.

dayeh-xilinx:
retest this please. Failed to download metadata for repo 'epel'

pgschuey (approved):
Changes look good

hackwa (approved):

IshitaGhosh (approved):
Change looks good.

PR #6001, opened 2021-11-16
Xbutil os

IshitaGhosh:
Removes OS specific code in ReportDebugIPStatus

Tested xbutil2 examine -r debug-ip-status on PCIe devices


src/runtime_src/core/common/debug_ip.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+#define XRT_CORE_COMMON_SOURCE
+
+#include "debug_ip.h"
+
+#include "core/include/xcl_perfmon_parameters.h"
+
+namespace xrt_core { namespace debug_ip {
+
+std::vector<uint64_t> getAIMCounterResult(const xrt_core::device* device, debug_ip_data* dbgIpData)
```
stsoe, line 25: One minor request.  For consistency prefer` lower_case `functions and variables in all core/common/*.
Also, please keep everything to the left, e.g.
```
return_type
function_name(params...)
{
}
```
And lastly, please add comments.

src/runtime_src/core/common/debug_ip.h
```diff
@@ -0,0 +1,49 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef XRT_CORE_DEBUG_IP_H
+#define XRT_CORE_DEBUG_IP_H
+
+#include "core/common/device.h"
+
+#include "core/include/xrt.h"
```
stsoe, line 22: Not needed.

src/runtime_src/core/common/debug_ip.h
```diff
@@ -0,0 +1,49 @@
...
+#include "core/include/xclbin.h"
+
+#include<vector>
+
+namespace xrt_core {  namespace debug_ip {
+
+  XRT_CORE_COMMON_EXPORT
+  std::vector<uint64_t> getAIMCounterResult(const xrt_core::device* , debug_ip_data*);
```
stsoe, line 30: Same comments as in .cpp file.   lower_case and left centered.

src/runtime_src/core/common/debug_ip.h
```diff
@@ -0,0 +1,49 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+#ifndef XRT_CORE_DEBUG_IP_H
+#define XRT_CORE_DEBUG_IP_H
+
+#include "core/common/device.h"
```
stsoe, line 20: I think you should forward declare and not include device.h and xclbin.h when it really isn't needed.
```
struct debug_ip_data;

namespace xrt_core {

class device;

namespace debug_ip {

...

}} // namespace debug_ip, xrt_core
```

src/runtime_src/core/common/debug_ip.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include "core/include/xrt.h"
+#include "core/include/xclbin.h"
+
+#include<vector>
```
stsoe, line 25: and
`#include <cstdint>`

src/runtime_src/core/common/query_requests.h
```diff
@@ -18,6 +18,7 @@
 #define xrt_core_common_query_requests_h
 
 #include "core/include/xclerr_int.h"
+#include "core/include/xclbin.h"
```
stsoe, line 21: Definitely forward declare content of this file to avoid inclusion.

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -46,6 +46,8 @@ extern "C" {
 #endif
 #endif
 
+#include "core/common/debug_ip.h"
```
stsoe, line 49: move before line 23.

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -442,6 +444,84 @@ struct aie_reg_read
...
 };
 
+struct aim_counter
+{
+  using result_type = query::aim_counter::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type key, const boost::any& arg1)
+  {
+    const auto dbgIpData = boost::any_cast<query::aim_counter::debug_ip_data_type>(arg1);
```
stsoe, line 454: avoid camelCase

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -533,6 +613,17 @@ struct function3_get : QueryRequestType
...
 };
 
+template <typename QueryRequestType, typename Getter>
+struct function4_get : virtual QueryRequestType
+{
+  boost::any
+  get(const xrt_core::device* device, const boost::any& arg1) const
+  {
+    auto k = QueryRequestType::key;
+    return Getter::get(device, k, arg1);
```
stsoe, line 623: Just to point out what may look  inconsistent but in fact is not:  template types and always CamelCaseTypes by convention :-)

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +280,315 @@ struct mac_addr_list
...
 };
 
+struct aim_counter
+{
+  using result_type = query::aim_counter::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type key, const boost::any& arg1)
+  {
+    const auto dbgIpData = boost::any_cast<query::aim_counter::debug_ip_data_type>(arg1);
```
stsoe, line 290: No camelCase for variables.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +280,315 @@ struct mac_addr_list
   }
 };
 
+struct aim_counter
```
stsoe, line 283: You may want a comment about this struct.  I know this is not followed consistently, but comments are always good.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +280,315 @@ struct mac_addr_list
...
+    std::string namePath = pdev->get_sysfs_path(aimName.c_str(), "name");
+
+    std::size_t pos = namePath.find_last_of('/');
+    std::string path = namePath.substr(0, pos+1);
+    path += "counters";
+
+    result_type retvalBuf(9, 0);
+
+    std::ifstream ifs(path.c_str());
+    if(!ifs) {
```
stsoe, line 306: Space between if/while and expression please.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +280,315 @@ struct mac_addr_list
...
+
+    std::string namePath = pdev->get_sysfs_path(aimName.c_str(), "name");
+
+    std::size_t pos = namePath.find_last_of('/');
+    std::string path = namePath.substr(0, pos+1);
+    path += "counters";
+
+    result_type retvalBuf(9, 0);
+
+    std::ifstream ifs(path.c_str());
```
stsoe, line 305: ifstream takes std::string, so don't cast to c-string.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +280,315 @@ struct mac_addr_list
...
+    if(!ifs) {
+      std::cout << "\nINFO: Incomplete AIM counter data in " << path << std::endl;
+      return retvalBuf;
+    }
+
+    const size_t sz = 256;
```
stsoe, line 311: Make sure value is done at compile time (it may be for 'const' but is not guaranteed)
`constexpr size_t sz = 256`

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +280,315 @@ struct mac_addr_list
...
+    char buffer[sz];
```
stsoe, line 312: `std::array<char, sz> buffer = {0};`

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +280,315 @@ struct mac_addr_list
...
+    std::memset(buffer, 0, sz);
+    ifs.getline(buffer, sz);
+
+    result_type valBuf;
+
+    while(!ifs.eof()) {
+      valBuf.push_back(strtoull((const char*)(&buffer), NULL, 10));
```
stsoe, line 319: nullptr.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +280,315 @@ struct mac_addr_list
...
+      std::memset(buffer, 0, sz);
+      ifs.getline(buffer, sz);
+    }
+
+    ifs.close();
```
stsoe, line 324: remove, it is not needed.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +280,315 @@ struct mac_addr_list
...
+
+    if(valBuf.size() < 13) {
```
stsoe, line 326: 13 is magic.  Needs a symbolic constant

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +280,315 @@ struct mac_addr_list
...
+
+    retvalBuf[0] = valBuf[0];
+    retvalBuf[1] = valBuf[1];
+    retvalBuf[2] = valBuf[4];
+    retvalBuf[3] = valBuf[5];
+    retvalBuf[4] = valBuf[8];
+    retvalBuf[5] = valBuf[9];
+    retvalBuf[6] = valBuf[10];
+    retvalBuf[7] = valBuf[11];
+    retvalBuf[8] = valBuf[12];
```
stsoe, line 341: Argh,  took me a while to notice the jump from 5->8; I was suggesting std::copy, which you could still do and split into two.  Either that or at least a comment.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +280,315 @@ struct mac_addr_list
...
+
+    std::string namePath = pdev->get_sysfs_path(amName.c_str(), "name");
+
+    std::size_t pos = namePath.find_last_of('/');
+    std::string path = namePath.substr(0, pos+1);
+    path += "counters";
+
+    result_type valBuf(10, 0);
+
+    std::ifstream ifs(path.c_str());
```
stsoe, line 370: c_str() not needed.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +280,315 @@ struct mac_addr_list
...
+      idx++;
+      std::memset(buffer, 0, sz);
+      ifs.getline(buffer, sz);
+    }
+
+    if(idx < 10) {
+      std::cout << "\nINFO: Incomplete AM counter data in " << path << std::endl;
+    }
+
+    ifs.close();
```
stsoe, line 393: Remove

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +280,315 @@ struct mac_addr_list
...
+    while(!ifs.eof()) {
+      valBuf[idx] = strtoull((const char*)(&buffer), NULL, 10);
+      idx++;
+      std::memset(buffer, 0, sz);
+      ifs.getline(buffer, sz);
+    }
+
+    if(idx < 10) {
+      std::cout << "\nINFO: Incomplete AM counter data in " << path << std::endl;
+    }
```
stsoe, line 391: No braces for single expression

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +280,315 @@ struct mac_addr_list
...
+
+    result_type valBuf(10, 0);
+
+    std::ifstream ifs(path.c_str());
+    if(!ifs) {
+      std::cout << "\nINFO: Incomplete AM counter data in " << path << std::endl;
+      return valBuf;
+    }
+
+    const size_t sz = 256;
```
stsoe, line 376: constexpr

stsoe (changes requested):
Looks good, overall a good generic change.  
A few style comments.   I stopped entering the same comments after a while,  
Please address and request re-review.

stsoe:
One more comment.  You should generally use `git rebase master` rather than merging master to your feature branch.  That way your changes are always kept on top of current master.  End result will be the same, but the commit history will look a bit simpler for your PR.

xilinxgitops:
retest this please

xilinxgitops:
retest this please

dayeh-xilinx:
retest this please. Fixed boardfarm test issue.

jvillarre (approved):
Register accesses look good after latest changes.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +282,324 @@ struct mac_addr_list
...
+
+  static result_type
+  get(const xrt_core::device* device, key_type key, const boost::any& arg1)
+  {
+    const auto dbg_ip_data = boost::any_cast<query::asm_counter::debug_ip_data_type>(arg1);
+    auto pdev = get_pcidev(device);
+
+    // read counter values
+    std::string asm_name("axistream_mon_");
+    asm_name = asm_name + std::to_string(dbg_ip_data->m_base_address);
```
rozumx, line 421:
```suggestion
    asm_name += std::to_string(dbg_ip_data->m_base_address);
```

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +282,324 @@ struct mac_addr_list
...
+  using result_type = query::lapc_status::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type key, const boost::any& arg1)
+  {
+    const auto dbg_ip_data = boost::any_cast<query::lapc_status::debug_ip_data_type>(arg1);
+    auto pdev = get_pcidev(device);
+
+    std::string lapc_name("lapc_");
+    lapc_name = lapc_name + std::to_string(dbg_ip_data->m_base_address);
```
rozumx, line 471:
```suggestion
    lapc_name += std::to_string(dbg_ip_data->m_base_address);
```

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +282,324 @@ struct mac_addr_list
...
+  using result_type = query::spc_status::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type key, const boost::any& arg1)
+  {
+    const auto dbg_ip_data = boost::any_cast<query::spc_status::debug_ip_data_type>(arg1);
+    auto pdev = get_pcidev(device);
+
+    std::string spc_name("spc_");
+    spc_name = spc_name + std::to_string(dbg_ip_data->m_base_address);
```
rozumx, line 522:
```suggestion
    spc_name += std::to_string(dbg_ip_data->m_base_address);
```

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +282,324 @@ struct mac_addr_list
...
+  get(const xrt_core::device* device, key_type key, const boost::any& arg1)
+  {
+    const auto dbg_ip_data = boost::any_cast<query::accel_deadlock_status::debug_ip_data_type>(arg1);
+    auto pdev = get_pcidev(device);
+
+    result_type val_buf = 0;
+
+    // read counter values
+    std::string mon_name("accel_deadlock_");
+    mon_name = mon_name + std::to_string(dbg_ip_data->m_base_address);
```
rozumx, line 575:
```suggestion
    mon_name += std::to_string(dbg_ip_data->m_base_address);
```

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +282,324 @@ struct mac_addr_list
...
+    ifs.getline(buffer.data(), sz);
+
+    if (!ifs.eof()) {
+      val_buf = std::stoi((const char*)(&buffer), nullptr, 10);
+    } else {
+      std::cout << "\nINFO: Incomplete Accelerator Deadlock detector status in " << path << std::endl;
+    }
+
+    return val_buf;
+
```
rozumx, line 599: Extra slot should be removed.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -379,127 +318,17 @@ DebugIpStatusCollector::readAIMCounter(debug_ip_data* dbgIpInfo)
...
-    aimResults.ReadTranx[index]       = valBuf[5];
-    aimResults.OutStandCnts[index]    = valBuf[8];
-    aimResults.LastWriteAddr[index]   = valBuf[9];
-    aimResults.LastWriteData[index]   = valBuf[10];
-    aimResults.LastReadAddr[index]    = valBuf[11];
-    aimResults.LastReadData[index]    = valBuf[12];
-
-    ifs.close();
+  std::vector<uint64_t> valBuf = xrt_core::device_query<xrt_core::query::aim_counter>(device, dbgIpInfo);
+  aimResults.WriteBytes[index]    = valBuf[0];
```
rozumx, line 322: It would be good to have an alternate way of representing the valBuf index instead of using an magic number.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -1802,7 +1265,7 @@ ReportDebugIpStatus::getPropertyTree20202( const xrt_core::device * _pDevice,
   pt.put("description","Status of Debug IPs present in xclbin loaded on device");
   auto handle = _pDevice->get_device_handle();
 
-  DebugIpStatusCollector collector(handle, true);
+  DebugIpStatusCollector collector(handle, _pDevice);
```
rozumx, line 1268: Thank you

rozumx (changes requested):

src/runtime_src/core/common/debug_ip.cpp
```diff
@@ -0,0 +1,304 @@
...
+                    &sample_interval, sizeof(uint32_t));
+
+  // If applicable, read the upper 32-bits of the 64-bit debug counters
+  if (dbg_ip_data->m_properties & XAIM_64BIT_PROPERTY_MASK) {
+    for (int c = 0 ; c < XAIM_DEBUG_SAMPLE_COUNTERS_PER_SLOT ; ++c) {
+      device->xread(XCL_ADDR_SPACE_DEVICE_PERFMON,
+                        dbg_ip_data->m_base_address + aim_upper_offsets[c],
+                        &curr_data[c], sizeof(uint32_t));
+    }
+    ret_val[0] = (((uint64_t)(curr_data[0])) << 32 );
```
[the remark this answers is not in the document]
stsoe: static_cast<uint64_t>(...)

src/runtime_src/core/common/debug_ip.cpp
```diff
@@ -0,0 +1,304 @@
...
+    ret_val[8] = (((uint64_t)(curr_data[8])) << 32 );
+  }
+
+  for (int c=0; c < XAIM_DEBUG_SAMPLE_COUNTERS_PER_SLOT; c++) {
+    device->xread(XCL_ADDR_SPACE_DEVICE_PERFMON,
+                      dbg_ip_data->m_base_address + aim_offsets[c],
+                      &curr_data[c], sizeof(uint32_t));
+  }
+
+  ret_val[0] |= curr_data[0];
```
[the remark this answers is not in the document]
stsoe: static_cast<uint64_t>(...)

src/runtime_src/core/common/query_requests.h
```diff
@@ -29,6 +29,8 @@
 #include <boost/any.hpp>
 #include <boost/format.hpp>
 
+struct debug_ip_data;
```
AShivangi, line 32: This doesn't look right. We have custom struct return types for many of these queries which are not declared as global variables

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +282,324 @@ struct mac_addr_list
...
+      return val_buf;
+    }
+
+    constexpr size_t sz = 256;
+    std::array<char, sz> buffer = {0};
+    ifs.getline(buffer.data(), sz);
+
+    size_t idx = 0;
+    while(!ifs.eof()) {
+      val_buf[idx] = std::stoi((const char*)(&buffer), nullptr, 10);
```
AShivangi, line 493: reinterpret_cast<const char*>(...)

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +282,324 @@ struct mac_addr_list
...
```
AShivangi, line 544: reinterpret_cast<const char*>(...)

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -278,6 +282,324 @@ struct mac_addr_list
...
+    if (!ifs) {
+      return val_buf;
+    }
+
+    constexpr size_t sz = 256;
+    std::array<char, sz> buffer = {0};
+    ifs.getline(buffer.data(), sz);
+
+    if (!ifs.eof()) {
+      val_buf = std::stoi((const char*)(&buffer), nullptr, 10);
```
AShivangi, line 593: reinterpret_cast<const char*>(...)

AShivangi (changes requested):

src/runtime_src/core/common/query_requests.h
IshitaGhosh: This is forward declaration. I added it to avoid including the file which declares "debug_ip_data" datatype.

src/runtime_src/core/pcie/linux/device_linux.cpp
IshitaGhosh: I could not follow which one are you referring as extra slot.

91palash91:
retest this please

xilinxgitops:
retest this please

dayeh-xilinx:
retest this please. temporarily disable U250 hardware test  for 2021.2 and 2022.1

src/runtime_src/core/pcie/windows/alveo/device_windows.cpp
```diff
@@ -1187,6 +1197,86 @@ struct mailbox
...
+  {
+    const auto dbg_ip_data = boost::any_cast<query::aim_counter::debug_ip_data_type>(arg1);
+
+    return xrt_core::debug_ip::get_aim_counter_result(device, dbg_ip_data);
+  }
+};
+
+struct am_counter
+{
+  using result_type = query::am_counter::result_type;
```
keryell, line 1216: Already defined in the headers?

src/runtime_src/core/pcie/windows/alveo/device_windows.cpp
```diff
@@ -1187,6 +1197,86 @@ struct mailbox
...
+    return xrt_core::debug_ip::get_spc_status(device, dbg_ip_data);
+  }
+};
+
+struct accel_deadlock_status
+{
+  using result_type = query::accel_deadlock_status::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type key, const boost::any& arg1)
```
keryell, line 1271: Here is it static and not virtual member function?
Why?
I guess there is a huge lack of comments at the top of the files to explain the design and what is going on.

src/runtime_src/core/pcie/windows/alveo/device_windows.cpp
```diff
@@ -1275,6 +1365,19 @@ struct function2_getter : QueryRequestType
   }
 };
 
+
+template <typename QueryRequestType, typename Getter>
+struct function4_get : virtual QueryRequestType
```
keryell, line 1370: Why `virtual` here? Do you have a diamond problem to solve?
https://en.cppreference.com/w/cpp/language/derived_class

keryell (changes requested):
Quite a commitment in this PR!
There are a few points that could be solved by configuring some tooling and applying some coding rules.

src/runtime_src/core/common/query_requests.h
stsoe: Comment as forward declaration.   This is a pretty common header, which arguably could be split into multiple ones that are more domain specific, but until so, the includes should be at a minimum and types should be forward declared when that is sufficient.

src/runtime_src/core/common/debug_ip.cpp
```diff
@@ -0,0 +1,288 @@
...
+  uint32_t curr_data[XAM_DEBUG_SAMPLE_COUNTERS_PER_SLOT] = {0};
+
+  uint32_t sample_interval;
+  // Read sample interval register to latch the sampled metric counters
+  device->xread(XCL_ADDR_SPACE_DEVICE_PERFMON,
+                  dbg_ip_data->m_base_address + XAM_SAMPLE_OFFSET,
+                  &sample_interval, sizeof(uint32_t));
+
+  auto dbg_ip_version = std::make_pair(dbg_ip_data->m_major, dbg_ip_data->m_minor);
+  auto ref_version   = std::make_pair((uint8_t)1, (uint8_t)1);
```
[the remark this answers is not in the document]
stsoe: We are at the mercy of the OSes we have to support.  It is quite possible that C++17 is supported now by all default compilers installed on the system, but no-one has looked at that AFAIK.

src/runtime_src/core/common/query_requests.h
```diff
@@ -803,6 +813,15 @@ struct ip_layout_raw : request
   get(const device*) const = 0;
 };
 
+struct debug_ip_layout_raw : request
+{
+  using result_type = std::vector<char>;
+  static const key_type key = key_type::debug_ip_layout_raw;
```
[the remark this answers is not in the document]
stsoe: This is wholesale change that should be made to this file.  I agree and it is my issue to do this for whole file.

src/runtime_src/core/common/query_requests.h
```diff
@@ -2707,6 +2726,66 @@ struct heartbeat_stall : request
   get(const device*) const = 0;
 };
 
+struct aim_counter : request
+{
+  using result_type = std::vector<uint64_t>;
+  using debug_ip_data_type = debug_ip_data*;
+  static const key_type key = key_type::aim_counter;
```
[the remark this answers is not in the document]
stsoe: Agreed, but I will do this on a rainy cleanup day for this file in its entirety.

src/runtime_src/core/common/query_requests.h
```diff
@@ -2707,6 +2726,66 @@ struct heartbeat_stall : request
...
+
+  virtual boost::any
+  get(const xrt_core::device* device, const boost::any& arg1) const = 0;
+};
+
+struct am_counter : request
+{
+  using result_type = std::vector<uint64_t>;
+  using debug_ip_data_type = debug_ip_data*;
+  static const key_type key = key_type::am_counter;
```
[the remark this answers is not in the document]
stsoe: Good suggestion.  Thanks!  Leave this for me.

src/runtime_src/core/common/query_requests.h
```diff
@@ -2707,6 +2726,66 @@ struct heartbeat_stall : request
...
+
+struct accel_deadlock_status : request
+{
+  using result_type = uint32_t;
+  using debug_ip_data_type = debug_ip_data*;
+  static const key_type key = key_type::accel_deadlock_status;
+
+  virtual boost::any
+  get(const xrt_core::device* device, const boost::any& arg1) const = 0;
+};
```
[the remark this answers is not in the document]
stsoe: As generic as this is not quite possible unless you create a few domain specific base classes and it would obscure the usage IMO.  E.g. I know I need some device data from sysfs and I want and I know this is a query request.  I would like to understand what query request returns, how it formats the data, what argument it takes etc; all very specific to the request itself and easier to read within a self contained struct.   I agree with the key though.

I definitely see that a comment regarding what 'arg1' is is missing.

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -442,6 +444,84 @@ struct aie_reg_read
   }
 };
 
+struct aim_counter
+{
+  using result_type = query::aim_counter::result_type;
```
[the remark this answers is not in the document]
stsoe: That's why it is used here in the implementation struct.  Note, that this struct has nothing to do with the one in header file other than being the struct in which the data collection for the query::aim_counter is implemented.

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -442,6 +444,84 @@ struct aie_reg_read
...
+
+  static result_type
+  get(const xrt_core::device* device, key_type key, const boost::any& arg1)
```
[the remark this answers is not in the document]
stsoe: You are and rightly so.  This is the pattern for implementing the query request getters you see in query_request.h.  You will have to follow a trail of template instantiation to get this this get() function.  In this case the get function has custom implementation, in other cases it will translate into a sysfs call.

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -533,6 +613,17 @@ struct function3_get : QueryRequestType
   }
 };
 
+template <typename QueryRequestType, typename Getter>
+struct function4_get : virtual QueryRequestType
+{
+  boost::any
+  get(const xrt_core::device* device, const boost::any& arg1) const
```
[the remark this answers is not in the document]
stsoe: Agreed, this is the virtual dispatch from the override in query::request.   I will add this to my to-do list and add override everywhere, it is not just here.

src/runtime_src/core/pcie/windows/alveo/device_windows.cpp
stsoe: Lack of comments prevent me from answering this.  I have no clue why virtual derivation was added, but I originally did this and I am fully aware of what it means, but for the heck of it I cannot see why it is there now, and github logs do not tell me why I added it.  Could have been some interim implementation that was since changed?  Again, my item on my to-do list.

src/runtime_src/core/common/debug_ip.cpp
```diff
@@ -0,0 +1,288 @@
...
+    }
+
+    ret_val[XAM_IOCTL_EXECUTION_COUNT_INDEX] = (static_cast<uint64_t>(curr_data[XAM_ACCEL_EXECUTION_COUNT_INDEX])) << 32;
+    ret_val[XAM_IOCTL_EXECUTION_CYCLES_INDEX] = (static_cast<uint64_t>(curr_data[XAM_ACCEL_EXECUTION_CYCLES_INDEX])) << 32;
+    ret_val[XAM_IOCTL_STALL_INT_INDEX] = (static_cast<uint64_t>(curr_data[XAM_ACCEL_STALL_INT_INDEX])) << 32;
+    ret_val[XAM_IOCTL_STALL_STR_INDEX] = (static_cast<uint64_t>(curr_data[XAM_ACCEL_STALL_STR_INDEX])) << 32;
+    ret_val[XAM_IOCTL_STALL_EXT_INDEX] = (static_cast<uint64_t>(curr_data[XAM_ACCEL_STALL_EXT_INDEX])) << 32;
+    ret_val[XAM_IOCTL_MIN_EXECUTION_CYCLES_INDEX] = (static_cast<uint64_t>(curr_data[XAM_ACCEL_MIN_EXECUTION_CYCLES_INDEX])) << 32;
+    ret_val[XAM_IOCTL_MAX_EXECUTION_CYCLES_INDEX] = (static_cast<uint64_t>(curr_data[XAM_ACCEL_MAX_EXECUTION_CYCLES_INDEX])) << 32;
+    ret_val[XAM_IOCTL_START_COUNT_INDEX] = (static_cast<uint64_t>(curr_data[XAM_ACCEL_TOTAL_CU_START_INDEX])) << 32;
```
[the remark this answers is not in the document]
IshitaGhosh: The indices used on the LHS are not consecutive. So, I could not put the assignment in the loop above. I could not follow your suggestion of using a function here.

src/runtime_src/core/common/debug_ip.cpp
```diff
@@ -0,0 +1,288 @@
...
+    device->xread(XCL_ADDR_SPACE_DEVICE_PERFMON, dbg_ip_data->m_base_address+am_offsets[c], &curr_data[c], sizeof(uint32_t));
+  }
+  ret_val[XAM_IOCTL_EXECUTION_COUNT_INDEX] |= curr_data[XAM_ACCEL_EXECUTION_COUNT_INDEX];
+  ret_val[XAM_IOCTL_EXECUTION_CYCLES_INDEX] |= curr_data[XAM_ACCEL_EXECUTION_CYCLES_INDEX];
+  ret_val[XAM_IOCTL_STALL_INT_INDEX] |= curr_data[XAM_ACCEL_STALL_INT_INDEX];
+  ret_val[XAM_IOCTL_STALL_STR_INDEX] |= curr_data[XAM_ACCEL_STALL_STR_INDEX];
+  ret_val[XAM_IOCTL_STALL_EXT_INDEX] |= curr_data[XAM_ACCEL_STALL_EXT_INDEX];
+  ret_val[XAM_IOCTL_MIN_EXECUTION_CYCLES_INDEX] |= curr_data[XAM_ACCEL_MIN_EXECUTION_CYCLES_INDEX];
+  ret_val[XAM_IOCTL_MAX_EXECUTION_CYCLES_INDEX] |= curr_data[XAM_ACCEL_MAX_EXECUTION_CYCLES_INDEX];
+  ret_val[XAM_IOCTL_START_COUNT_INDEX] |= curr_data[XAM_ACCEL_TOTAL_CU_START_INDEX];
```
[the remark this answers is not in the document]
IshitaGhosh: The indices used in the LHS expression are not consecutive. So, I could put these in the loop above.

src/runtime_src/core/common/debug_ip.cpp
```diff
@@ -0,0 +1,288 @@
...
+                    dbg_ip_data->m_base_address + XAIM_SAMPLE_OFFSET,
+                    &sample_interval, sizeof(uint32_t));
+
+  // If applicable, read the upper 32-bits of the 64-bit debug counters
+  if (dbg_ip_data->m_properties & XAIM_64BIT_PROPERTY_MASK) {
+    for (int c = 0 ; c < XAIM_DEBUG_SAMPLE_COUNTERS_PER_SLOT ; ++c) {
+      device->xread(XCL_ADDR_SPACE_DEVICE_PERFMON,
+                        dbg_ip_data->m_base_address + aim_upper_offsets[c],
+                        &curr_data[c], sizeof(uint32_t));
+      ret_val[c] = (((uint64_t)(curr_data[c])) << 32 );
```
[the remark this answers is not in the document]
IshitaGhosh: I think use of cast and shift here makes the intent clear. Hiding those in a function will hide how we are reading 64bit counter register, if present, in 32 bit chunks. As the current function handles a specific monitor type, we can keep this detail here.

xilinxgitops:
retest this please

xilinxgitops:
retest this please

xilinxgitops:
retest this please

xilinxgitops:
retest this please

src/runtime_src/core/common/debug_ip.cpp
```diff
@@ -0,0 +1,285 @@
...
+    XAIM_SAMPLE_LAST_WRITE_DATA_UPPER_OFFSET,
+    XAIM_SAMPLE_LAST_READ_ADDRESS_UPPER_OFFSET,
+    XAIM_SAMPLE_LAST_READ_DATA_UPPER_OFFSET
+  };
+
+  std::vector<uint64_t> ret_val(XAIM_DEBUG_SAMPLE_COUNTERS_PER_SLOT);
+
+  uint32_t curr_data[XAIM_DEBUG_SAMPLE_COUNTERS_PER_SLOT];
+
+  uint32_t sample_interval;
```
AShivangi, line 60: initialize

src/runtime_src/core/common/debug_ip.cpp
```diff
@@ -0,0 +1,285 @@
...
+    XAM_ACCEL_MAX_EXECUTION_CYCLES_UPPER_OFFSET,
+    XAM_ACCEL_TOTAL_CU_START_UPPER_OFFSET
+  };
+
+  std::vector<uint64_t> ret_val(XAM_TOTAL_DEBUG_COUNTERS_PER_SLOT);
+
+  // Read all metric counters
+  uint32_t curr_data[XAM_DEBUG_SAMPLE_COUNTERS_PER_SLOT] = {0};
+
+  uint32_t sample_interval;
```
AShivangi, line 120: initialize

src/runtime_src/core/common/debug_ip.cpp
```diff
@@ -0,0 +1,285 @@
...
+  // Read sample interval register to latch the sampled metric counters
+  device->xread(XCL_ADDR_SPACE_DEVICE_PERFMON,
+                dbg_ip_data->m_base_address + XAM_SAMPLE_OFFSET,
+                &sample_interval, sizeof(uint32_t));
+
+  auto dbg_ip_version = std::make_pair(dbg_ip_data->m_major, dbg_ip_data->m_minor);
+  auto ref_version    = std::make_pair((uint8_t)1, (uint8_t)1);
```
AShivangi, line 127: `static_cast<uint8_t>(...)`

src/runtime_src/core/common/debug_ip.cpp
```diff
@@ -0,0 +1,285 @@
...
+    XASM_NUM_TRANX_OFFSET,
+    XASM_DATA_BYTES_OFFSET,
+    XASM_BUSY_CYCLES_OFFSET,
+    XASM_STALL_CYCLES_OFFSET,
+    XASM_STARVE_CYCLES_OFFSET
+  };
+
+  std::vector<uint64_t> ret_val(XASM_DEBUG_SAMPLE_COUNTERS_PER_SLOT);
+
+  uint32_t sample_interval ;
```
AShivangi, line 200: init

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -59,13 +63,57 @@ get_render_value(const std::string dir)
...
 inline pdev
 get_pcidev(const xrt_core::device* device)
 {
   return pcidev::get_dev(device->get_device_id(), device->is_userpf());
 }
 
+
+std::vector<uint64_t> 
+get_counter_status_from_sysfs(std::string mon_name_address, 
+                              std::string sysfs_file_name,
```
stsoe, line 75: const std::string&

AShivangi (approved):
The changes look good to me with minor changes requested. Please address/ resolve the existing comments

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -59,13 +63,57 @@ get_render_value(const std::string dir)
...
+                              size_t size, 
+                              const xrt_core::device* device)
+{
+  auto pdev = get_pcidev(device);
+
+  /* Get full path to "name" sysfs file. 
+   * Then use that path to form full path to "counters"/"status" sysfs file 
+   * which contains counter/status data for the monitor
+   */
+  std::string name_path = pdev->get_sysfs_path(mon_name_address.c_str(), "name");
```
stsoe, line 85: Remove c_str()

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -59,13 +63,57 @@ get_render_value(const std::string dir)
...
+  std::ifstream ifs(path);
+  if (!ifs) {
+    std::cout << "\nINFO: Incomplete counter data in " << path << std::endl;
+    return val_buf;
+  }
+
+  std::string buffer;
+  ifs >> buffer;
+
+  size_t idx = 0;
```
stsoe, line 102: Add a comment as to what is being parsed here, I mean what format are you expecting?

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -59,13 +63,57 @@ get_render_value(const std::string dir)
...
+  while (!ifs.eof()) {
+    val_buf[idx] = strtoull(buffer.c_str(), nullptr, 10);
+    idx++;
+    buffer = "";
+    ifs >> buffer;
+  }
+
+  if (idx < size)
```
stsoe, line 110: what if idx >= size ?   val_buf is sized per argument size, but what is the correlation between argument and content of sysfs node.   This is not clean I think.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -59,13 +63,57 @@ get_render_value(const std::string dir)
...
 
-
 inline pdev
 get_pcidev(const xrt_core::device* device)
 {
   return pcidev::get_dev(device->get_device_id(), device->is_userpf());
 }
 
+
+std::vector<uint64_t>
```
stsoe, line 73: static ?

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -59,13 +63,57 @@ get_render_value(const std::string dir)
...
+  auto pdev = get_pcidev(device);
+
+  /* Get full path to "name" sysfs file. 
+   * Then use that path to form full path to "counters"/"status" sysfs file 
+   * which contains counter/status data for the monitor
+   */
+  std::string name_path = pdev->get_sysfs_path(mon_name_address.c_str(), "name");
+
+  std::size_t pos = name_path.find_last_of('/');
+  std::string path = name_path.substr(0, pos+1);
```
stsoe, line 88: what if pos is std::string::npos?

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -59,13 +63,57 @@ get_render_value(const std::string dir)
...
+  path += sysfs_file_name;
+
+  std::vector<uint64_t> val_buf(size);
+
+  std::ifstream ifs(path);
+  if (!ifs) {
+    std::cout << "\nINFO: Incomplete counter data in " << path << std::endl;
```
stsoe, line 95: Nah, let's not do std::cout here.   Use the messaging system if you have to send a message.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -59,13 +63,57 @@ get_render_value(const std::string dir)
...
+  size_t idx = 0;
+  while (!ifs.eof()) {
+    val_buf[idx] = strtoull(buffer.c_str(), nullptr, 10);
+    idx++;
+    buffer = "";
+    ifs >> buffer;
+  }
+
+  if (idx < size)
+    std::cout << "\nINFO: Incomplete counter data in " << path << std::endl;
```
stsoe, line 111: No std::cout

stsoe (changes requested):
Refactoring helped a lot.  Thanks!
A few comments in the new factored out function.

rozumx (approved):
Approved with minor issue.  Some of the copyright dates are in the future.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -59,13 +64,66 @@ get_render_value(const std::string dir)
...
+    val_buf[idx] = strtoull(buffer.c_str(), nullptr, 10);
+    idx++;
+    buffer = "";
+    ifs >> buffer;
+  }
+
+  if (idx < size) {
+    std::string msg = "Incomplete counter data in " + path + ". Using default values in results.";
+    xrt_core::message::send(xrt_core::message::severity_level::info, "XRT", msg);
+  }
```
keryell, line 121: OK, but if you want to read integers, why not reading them at the first place?
Think Python on non speed critical code. :-)
```suggestion
try {
  // Enable exception on reading error
  ifs.exceptions(std::ifstream::failbit);
  for (std::size_t idx = 0; idx < size; ++idx)
    ifs >> val_buf[idx];
  } catch (const std::ios_base::failure& fail) {
  std::string msg = "Incomplete data for counter " + idx " in " + path + " because " + fail.what()
    + "\nUsing 0 in results starting at position " + idx + ".";
  xrt_core::message::send(xrt_core::message::severity_level::info, "XRT", msg);
}
```
I let you adapt the code to compile, perhaps adding some `to_string` and so on.

src/runtime_src/core/pcie/linux/device_linux.cpp
keryell: @rozumx Perhaps that you are parsing a full `std::vector` of data while you are using only the first element?

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -140,9 +127,9 @@ private :
 
 
 DebugIpStatusCollector::DebugIpStatusCollector(xclDeviceHandle h,
-					       bool jsonFormat,
-					       std::ostream& _output)
+                                               const xrt_core::device* d)
     : handle(h)
+    , device(d)
     , infoMessage("")
```
keryell, line 133:
```suggestion
    , infoMessage{}
```

keryell (changes requested):
Great!
Almost there.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -59,13 +64,66 @@ get_render_value(const std::string dir)
...
+
+  return val_buf;
```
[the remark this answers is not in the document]
IshitaGhosh: "val_buf" is constructed with "size" number of uint64_t with default value 0. When idx < size, the rest of the elements keep their default value. So, I think it is quite obvious that the default values are used.

src/runtime_src/core/pcie/linux/device_linux.cpp
IshitaGhosh: The current code works fine. Do you see any blocking issue with the current code?

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -17,14 +17,19 @@
...
+#include "core/include/xcl_app_debug.h"
 
 #include "common/utils.h"
 #include "xrt.h"
 #include "scan.h"
 #include <string>
 #include <iostream>
+#include <fstream>
 #include <map>
+#include <array>
```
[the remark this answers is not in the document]
IshitaGhosh: This is not a blocking issue. So, I think it will be better if I address this in a new PR.

src/runtime_src/core/pcie/linux/device_linux.cpp
IshitaGhosh: @keryell , Stephen had added the comment before "get_counter_status_from_sysfs" method was introduced. So, he is definitely not referring to using std::vector for a single element.

src/runtime_src/core/include/xcl_app_debug.h
```diff
@@ -43,14 +43,68 @@ extern "C" {
...
+#define XAM_IOCTL_MIN_EXECUTION_CYCLES_INDEX  9
 
 /************************ ASM Debug Counters ********************************/
 #define XASM_DEBUG_SAMPLE_COUNTERS_PER_SLOT    5
 
+#define XASM_NUM_TRANX_INDEX      0
+#define XASM_DATA_BYTES_INDEX     1
+#define XASM_BUSY_CYCLES_INDEX    2
+#define XASM_STALL_CYCLES_INDEX   3
+#define XASM_STARVE_CYCLES_INDEX  4
```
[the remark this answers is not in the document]
IshitaGhosh: Actually, I have thought of a better solution. We can introduce new struct type for each monitor result and use it in device query methods and the get_*_counter_result methods. Like 
typedef struct {
  unsigned long long int WriteBytes;
  unsigned long long int WriteTranx;
  unsigned long long int ReadBytes;
  unsigned long long int ReadTranx;
  unsigned long long int OutStandCnts;
  unsigned long long int LastWriteAddr;
  unsigned long long int LastWriteData;
  unsigned long long int LastReadAddr;
  unsigned long long int LastReadData;
} xclAIMCounterResult;
Then we can refer the counter values with the member fields of the struct, instead of # defined constants or enums. I have already discussed this with @jvillarre and we have planned it for a new PR. The current PR is already too big and mainly addresses OS specific handling for accessing counter values for different monitor types.

src/runtime_src/core/pcie/linux/device_linux.cpp
keryell: Sure, sure...
Code quality? Beauty? Empathy with the reader? :-)

src/runtime_src/core/include/xcl_app_debug.h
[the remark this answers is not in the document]
keryell: Oh I see. Then you can "map" the `struct` on the driver data or even copy into it and just access each field.
This is clever!

src/runtime_src/core/pcie/linux/device_linux.cpp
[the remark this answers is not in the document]
keryell: Yes, this is quite obvious for *you* who wrote the code. :-)
I just spent 1 hour on this review today, of course not only on these lines.
Feel free to review my code too in exchange.

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
stsoe: Why not just leave it out?

src/runtime_src/core/pcie/linux/device_linux.cpp
stsoe: Ronan's suggestion for reading the buffer is great.  Much clearer to me what the code is actually trying to do.
I agree this about code quality, beautiful code, and readability.

src/runtime_src/core/pcie/linux/device_linux.cpp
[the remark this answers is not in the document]
stsoe: Please sort now while you are at it.   This is often overlooked, and I frequently forget, but it is a good suggestion.

src/runtime_src/core/common/debug_ip.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+  std::vector<uint64_t> ret_val(XAM_TOTAL_DEBUG_COUNTERS_PER_SLOT, 0);
+
+  // Read all metric counters
+  uint32_t curr_data[XAM_DEBUG_SAMPLE_COUNTERS_PER_SLOT] = {0};
+
+  uint32_t sample_interval;
+  // Read sample interval register to latch the sampled metric counters
+  device->xread(XCL_ADDR_SPACE_DEVICE_PERFMON,
+                  dbg_ip_data->m_base_address + XAM_SAMPLE_OFFSET,
```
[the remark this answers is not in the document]
IshitaGhosh: I have updated the indentation in all the places.

src/runtime_src/core/pcie/windows/alveo/device_windows.cpp
IshitaGhosh: This is the same issue as addressed by Soren in the following comment.

dayeh-xilinx:
retest this please. Disabled PCIE emulation test on Ubuntu.

xilinxgitops:
retest this please

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -424,6 +745,17 @@ struct function0_get : virtual QueryRequestType
   }
 };
 
+template <typename QueryRequestType, typename Getter>
+struct function4_get : virtual QueryRequestType
+{
+  boost::any
+  get(const xrt_core::device* device, const boost::any& arg1) const
```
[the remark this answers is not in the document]
IshitaGhosh: As per @stsoe  comment, he is going to look into these suggestions applicable for all of the device query methods and related data structures.

src/runtime_src/core/pcie/windows/alveo/device_windows.cpp
IshitaGhosh: As per @stsoe  comment, he is going to look into these suggestions applicable for all of the device query methods and related data structures.

src/runtime_src/core/common/debug_ip.cpp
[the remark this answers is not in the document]
IshitaGhosh: Thanks for the suggestion. I have added the suggested lambda function for upper and lower bytes.

src/runtime_src/core/common/query_requests.h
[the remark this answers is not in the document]
IshitaGhosh: I have renamed "arg1" as "dbg_ip_data" to indicate that debug_ip_data object is expected to be passed

src/runtime_src/core/common/debug_ip.cpp
```diff
@@ -0,0 +1,304 @@
...
+  bool has_dataflow = (dbg_ip_version > ref_version) ? true : false;
+
+  // If applicable, read the upper 32-bits of the 64-bit debug counters
+  if (dbg_ip_data->m_properties & XAM_64BIT_PROPERTY_MASK) {
+    for (int c = 0 ; c < XAM_DEBUG_SAMPLE_COUNTERS_PER_SLOT ; ++c) {
+      device->xread(XCL_ADDR_SPACE_DEVICE_PERFMON,
+            dbg_ip_data->m_base_address + am_upper_offsets[c],
+            &curr_data[c], sizeof(uint32_t));
+    }
+    ret_val[0] = ((uint64_t)(curr_data[0])) << 32;
```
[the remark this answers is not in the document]
IshitaGhosh: I have added # defined symbolic constants for now.

src/runtime_src/core/pcie/linux/device_linux.cpp
IshitaGhosh: @rozumx Could you please explain this comment? I could not follow what are you referring as extra slot.

src/runtime_src/core/pcie/linux/device_linux.cpp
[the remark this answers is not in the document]
IshitaGhosh: I have updated this block of code as per your suggestion about using try-catch for exception. However, I modified the message to 
std::string msg = "Incomplete counter data read from " + path + " due to " + fail.what()
                         + ".\n Using 0 as default value in results.";
I think mentioning "idx" in the message won't be very helpful to the user as user does not have any knowledge about the 
std::vector used in implementation.

src/runtime_src/core/pcie/linux/device_linux.cpp
IshitaGhosh: I have updated this block of code as per your suggestion about using try-catch for exception. However, I modified the message to
std::string msg = "Incomplete counter data read from " + path + " due to " + fail.what()
+ ".\n Using 0 as default value in results.";
I think mentioning "idx" in the message won't be very helpful to the user as user does not have any knowledge about the
std::vector used in implementation.

src/runtime_src/core/pcie/linux/device_linux.cpp
IshitaGhosh: I have added a comment

stsoe (approved):

src/runtime_src/core/common/debug_ip.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+  uint32_t sample_interval = 0;
+  // Read sample interval register to latch the sampled metric counters
+  device->xread(XCL_ADDR_SPACE_DEVICE_PERFMON,
+                dbg_ip_data->m_base_address + XAM_SAMPLE_OFFSET,
+                &sample_interval, sizeof(uint32_t));
+
+  auto dbg_ip_version = std::make_pair(dbg_ip_data->m_major, dbg_ip_data->m_minor);
+
+  std::pair<uint8_t, uint8_t> ref_version { 1, 1};
```
keryell, line 128:
```suggestion
  std::pair<uint8_t, uint8_t> ref_version { 1, 1 };
```

keryell (approved):
Great contribution and code cleaning.
Thank you for making Xilinx code safer and more beautiful! :-)

PR #6003, opened 2021-11-16
Base headers for support of kernels with multiple interrupts

sarab96:
New feature support for multiple interrupts requires callbacks to be supported in
user managed kernels. Adding header changes for callback functions.

xrt::ip to use a thread (cb_thread) to run callback functions.
User space can manage these callbacks for "user managed kernels".
One callback per interrupt bit to be supported

Following PRs will add more header changes and more functionality for the feature


sarab96:
Closing after discussion. Will submit new PR

PR #6004, opened 2021-11-18
In the similar lines of the Versal and ZynqMP, moving the device proc…

venkatp-xilinx:
In the similar lines of the Versal and ZynqMP, moving the device process execution to x86 even for zynq7 platforms. Post this submit we have Vitis perforce checkins

Unified flow for all the edge platforms

Enhancement to improve the kernel execution time

There are no alternate solutions

No risk, this is enhancment request

This is being edge platform, created the RPMs for aarch32 (zynq7 architecture) and test the designs manually

Nope


xilinxgitops:
retest this please

PR #6006, opened 2021-11-18
xball : Fixed issue on how xbmgmt and xbutil are called.

rozumx:
In 2021.1, there is a bug where xbmgmt and xbutil expect the --device option to be at the end of the command line and not in the front.  This change updates xball to add it to the end of the command line.

CR 1115358 -U30 XRT release : xball doesn't honor xbmgmt / xbutil --device option placement


AShivangi (approved):

anupcs-xlnx (approved):
Approved but could you change the summary of the CR itself to reflect the actual issue? Thanks.

PR #6011, opened 2021-11-21
VITIS-2372 : header update for bo type usage stats

sarab96:
For rugged XRT, header change to store bo usage based on type of bo.

Array of bo stats to be tracked in driver. Stats will be available in sysfs and with xbutil

Following PRs will make use of this array to update usage info

Will be documented ad part of xbutil usage info on completion

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
@@ -885,6 +885,10 @@ int xocl_cleanup_mem(struct xocl_drm *drm_p)
 	int ret;
 	mutex_lock(&drm_p->mm_lock);
 	ret = xocl_cleanup_mem_nolock(drm_p);
+	if (drm_p->bo_usage_stat) {
+		vfree(drm_p->bo_usage_stat);
+		drm_p->bo_usage_stat = NULL;
+	}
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
```diff
@@ -885,6 +885,10 @@ int xocl_cleanup_mem(struct xocl_drm *drm_p)
 	int ret;
 	mutex_lock(&drm_p->mm_lock);
 	ret = xocl_cleanup_mem_nolock(drm_p);
+	if (drm_p->bo_usage_stat) {
+		vfree(drm_p->bo_usage_stat);
+		drm_p->bo_usage_stat = NULL;
+	}
```
maxzhen, line 891: why these cleanup are not done inside xocl_cleanup_mem_nolock() while others are done there?

maxzhen (changes requested):

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
sarab96: Updated. Was not clear if that will work for xclbin with no DDR

maxzhen (approved):

PR #6013, opened 2021-11-22
Disable ABI compile time check when ccache is enabled

stsoe:
Ever since PR #5816, ccache has been close to useless because
abi.h includes CMake generated version.h which changes from
build to build.

This PR disables inclusion of version.h when ccache is enabled, it
assumes official pipeline master builds are not using ccache.


stsoe:
retest this please - this PR cannot cause u250 board failure

PR #6014, opened 2021-11-22
update XGQ common code change

mamin506:
Update XGQ common and impl header file for performance friendly changes.

No document impact.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
@@ -1271,8 +1271,7 @@ static int xgq_probe(struct platform_device *pdev)
 		goto attach_failed;
 	}
 
-	flags &= ~XGQ_SERVER;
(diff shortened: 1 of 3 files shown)


stsoe (approved):

maxzhen (approved):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -1271,8 +1271,7 @@ static int xgq_probe(struct platform_device *pdev)
 		goto attach_failed;
 	}
 
-	flags &= ~XGQ_SERVER;
```
xdavidz, line 1274: why change it to magic number 0, it is unclear this is not a server.

xdavidz: also, if the hdl can be NULL, can we just pass NULL instead of 0?

xdavidz (approved):

PR #6016, opened 2021-11-23
Fix PS kernel instance name to match up with IP_LAYOUT

stsoe:
Modify instance name to match up with xclbin convention where
instance names are numbered scu_0..n as in kernel_name:scu_[0..n]
for kernel_name. It is important that the instance name matches up with the
expectations of core XRT, which is based off of the instance names in
the IP_LAYOUT.

The instance names are created in user space XRT from the data
obtained from kernel driver, which doesn't use the IP_LAYOUT section
for PS kernels and in fact is not concerned with instance name beyond
the kernel name itself.

This PR uses a using a knm -> idx map, which allows the driver to
assign internal indeces in any order it likes and sysfs node does not
have to list all kernel instances in order of kernel names.

There is no real risk with this PR as only new #6010 needs the
instance names matching with IP_LAYOUT, this PR just enables support
for multiple different PS kernels each with any number of instances.

### DIFF
--- a/src/runtime_src/core/pcie/linux/device_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/device_linux.cpp
@@ -168,7 +168,8 @@ struct kds_scu_stat
     if (!errmsg.empty())
       throw xrt_core::query::sysfs_error(errmsg);
 
-    result_type cuStats;
+    result_type cu_stats;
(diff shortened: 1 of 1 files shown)


rozumx (approved):

dayeh-xilinx:
retest this please. temporarily disable U250 hardware test  for 2021.2 and 2022.1

PR #6035, opened 2021-12-01
disable 'xbutil reset' if the reset mailbox opcode is disabled

xuhz:
In some cases, like aws vt1, the reset mailbox opcode is disabled, so that xclmgmt will not do reset.
but at userland of xocl side, something still happens, like, killing apps.
This will cause confusion to the user
in vt1 instance, user run 'xbutil reset' in guest vm, the xrm and ffmpeg get killed, but card doesn't get reset
check the mailbox disable config info in 'xbutil reset', and return if the reset opcode is disabled
very low
in above case, 'xbutil reset' return immediately 
na

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_ioctl.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_ioctl.c
@@ -23,6 +23,7 @@
 #endif
 #include "version.h"
 #include "common.h"
+#include "mailbox_proto.h"
 
 extern int kds_mode;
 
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/tools/xbutil2/SubCmdReset.cpp
```diff
@@ -20,6 +20,7 @@
 #include "SubCmdReset.h"
 #include "core/common/query_requests.h"
 #include "tools/common/XBUtilities.h"
+#include "core/pcie/driver/linux/include/mailbox_proto.h"
```
AShivangi, line 23: this is common code for pcie and edge devices. We can't have pcie specific include files.

AShivangi (changes requested):

src/runtime_src/core/tools/xbutil2/SubCmdReset.cpp
xuhz: got it.
move the changes into xocl driver

AShivangi (approved):
Thanks for moving it

maxzhen (approved):

rozumx (approved):

PR #6038, opened 2021-12-02
CR-1112292  Not giving clear message when host mem is not a power of 2

dbenusov-xilinx:
Neither the malformed command, nor the unsupported memory size convey the power of 2 restriction

```
[dpeascoe@xcowtsalveo1 /scratch]$ sudo xbutil configure --host-mem enable -d 82:00.1 -s 6M
ERROR: Not enough host mem. Please check grub settings.: Cannot allocate memory
```

Add check to verify input size is a power of two

```
root@xsjpranjal01:~# xbutil configure --host-mem enable -d 02:00.1 6M
ERROR: too many positional options have been specified on the command line


DESCRIPTION: Controls host-mem functionality

USAGE: xbutil configure --host-mem [-h] [-d arg] [-s arg] action
  <action>        - Action to perform: ENABLE or DISABLE

OPTIONS:
  -d, --device       - The Bus:Device.Function (e.g., 0000:d8:00.0) device of interest
  -s, --size         - Size of host memory (bytes) to be enabled (e.g. 256M, 1G)
  -h, --help         - Help to use this sub-command

GLOBAL OPTIONS:
root@xsjpranjal01:~# xbutil configure --host-mem enable -d 02:00.1 -s 6M

ERROR: Please specify a non-zero memory size between 4M and 1G as a power of 2.: Invalid argument
```

### DIFF
--- a/src/runtime_src/core/tools/common/XBUtilities.cpp
+++ b/src/runtime_src/core/tools/common/XBUtilities.cpp
@@ -811,13 +811,6 @@ XBUtilities::string_to_bytes(std::string str)
   return size;
 }
 
-bool 
-XBUtilities::
-isPowerOf2(const uint64_t x)
-{
-  return !(x==0 || (x & (x - 1)));
-}
-
 std::string
 XBUtilities::
 get_xrt_pretty_version()
--- a/src/runtime_src/core/tools/common/XBUtilities.cpp
+++ b/src/runtime_src/core/tools/common/XBUtilities.cpp
@@ -811,6 +811,13 @@ XBUtilities::string_to_bytes(std::string str)
   return size;
 }
 
+bool 
+XBUtilities::
+isPowerOf2(const uint64_t x)
+{
+  return !(x & (x - 1));
+}
+
 std::string
 XBUtilities::
 get_xrt_pretty_version()
--- a/src/runtime_src/core/tools/common/XBUtilities.cpp
+++ b/src/runtime_src/core/tools/common/XBUtilities.cpp
@@ -815,7 +815,7 @@ bool
 XBUtilities::
 isPowerOf2(const uint64_t x)
 {
-  return !(x & (x - 1));
+  return !(x==0 || (x & (x - 1)));
 }
 
 std::string
--- a/src/runtime_src/core/tools/common/XBUtilities.h
+++ b/src/runtime_src/core/tools/common/XBUtilities.h
@@ -152,8 +152,16 @@ namespace XBUtilities {
   uint64_t 
   string_to_bytes(std::string str);
 
-  bool
-  isPowerOf2(const uint64_t x);
+  inline bool 
+  is_power_of_2(const uint64_t x)
+  {
+    /*
+    * Verify that the given value is greater than zero
+    * and that only one bit is set in the given value
+    * if only one bit is set this implies the given value is a power of 2
+    */
+    return (x != 0) && ((x & (x - 1)) == 0);
+  }
 
  /*
   * xclbin locking
--- a/src/runtime_src/core/tools/common/XBUtilities.h
+++ b/src/runtime_src/core/tools/common/XBUtilities.h
@@ -152,6 +152,9 @@ namespace XBUtilities {
   uint64_t 
   string_to_bytes(std::string str);
 
+  bool
+  isPowerOf2(const uint64_t x);
+
  /*
   * xclbin locking
   */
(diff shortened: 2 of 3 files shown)


src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
```diff
@@ -116,8 +116,9 @@ OO_HostMem::execute(const SubCmdOptions& _options) const
     }
     enable =  boost::iequals(m_action, "ENABLE");
 
-    // Exit if ENABLE action is specified but size is not
-    if(enable && size == 0)
+    // Exit if ENABLE action is specified but size is not 
+    // or is a power of 2
+    if(enable && (size == 0 || (size & (size - 1))))
```
rozumx, line 121: I like the simple solution you did with regards to determine if a value is a power of two.   

To carry it further and have it available for others to use, please do create a new helper function, something like:
`bool isPowerOf2(const uint64_t x);`

and place it in the XBUtilities.cpp (and of course use it here :^) ).

rozumx (changes requested):

AShivangi (approved):

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
dbenusov-xilinx: Good idea. Just finished the change in the latest commit.

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -811,6 +811,13 @@ XBUtilities::string_to_bytes(std::string str)
   return size;
 }
 
+bool 
+XBUtilities::
+isPowerOf2(const uint64_t x)
+{
+  return !(x & (x - 1));
```
rozumx, line 818: There is a bug here.  If x is zero it returns back an incorrect result for zero isn't a power of two.
```suggestion
  return !(x ==0) && !(x & (x - 1));
```

rozumx (changes requested):

src/runtime_src/core/tools/xbutil2/OO_HostMem.cpp
```diff
@@ -116,8 +116,9 @@ OO_HostMem::execute(const SubCmdOptions& _options) const
     }
     enable =  boost::iequals(m_action, "ENABLE");
 
-    // Exit if ENABLE action is specified but size is not
-    if(enable && size == 0)
+    // Exit if ENABLE action is specified and 
+    // size is not a power of 2
+    if(enable && !XBUtilities::isPowerOf2(size))
```
rozumx, line 121: I understand why you may have removed the "size==0" condition, since zero is not a power of two, but doing so does make the next developer work harder in understanding what you are testing for.   In addition, you are relying on a size effect of isPowerOf2() to determine a zero value.

This will come back an haunt the next developer if the power of 2 restriction is removed.

For this line only, would you please bring back the size comparison.
```suggestion
    if(enable && ((0 == size) || !XBUtilities::isPowerOf2(size)))
```

rozumx (changes requested):

src/runtime_src/core/tools/common/XBUtilities.h
```diff
@@ -152,6 +152,9 @@ namespace XBUtilities {
   uint64_t 
   string_to_bytes(std::string str);
 
+  bool
+  isPowerOf2(const uint64_t x);
```
stsoe, line 156: I can't help notice some inconsistency in the naming convention used.
I know that tools (xbutil/xbmgmt) tend to use camelCase  whereas core XRT is lower_case_under_score style.  Here though, I see under utility functions being lower_case ones, so maybe keep is_power_of_2()?

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -811,6 +811,13 @@ XBUtilities::string_to_bytes(std::string str)
   return size;
 }
 
+bool 
+XBUtilities::
+isPowerOf2(const uint64_t x)
+{
+  return !(x==0 || (x & (x - 1)));
```
stsoe, line 818: Candidate for being inline as it does nothing really.  Also either stick with equality check for both numbers or none at all for both, either is fine.

stsoe (changes requested):

src/runtime_src/core/tools/common/XBUtilities.h
dbenusov-xilinx: Fixed!

src/runtime_src/core/tools/common/XBUtilities.cpp
dbenusov-xilinx: I added the equality check for both. Are we comfortable with placing the definition of the function into the Utilities header file?

rozumx (approved):

dbenusov-xilinx:
retest this please.

dayeh-xilinx:
retest this please.

stsoe (approved):

PR #6042, opened 2021-12-03
CR-1112943: Fixing issue in profiling when both memory AIMs and trace-enabled AIMs are included

jvillarre:
As device trace is processed, we keep track of the last transaction on each AIM we see so we can approximate events that might have been dropped in the hardware.  We allocate vectors to store this last transaction data based on the number of AIMs that exist in the loaded xclbin.  The function that was being called, however, was returning the number of AIMs that had trace enabled, and not the total number of AIMs, so xclbins created with a mix of AIMs configured for trace and only counters would sometimes lead to an out of bounds access to this vector, causing a seg fault in some other part of memory when it was attempted to be reclaimed.

This pull request changes the function that returns the number of AIMs in an xclbin to the total number instead of just the number with trace, as expected.


IshitaGhosh (approved):
Change looks good.

PR #6043, opened 2021-12-04
Bug fixing to xrt enqueue operations

stsoe:
This PR reworks certain parts of the experimental XRT enqueue implementation.

While experimenting with enqueue operations a few bugs were identified

Fix callable async check which was using wrong type_traits check.

The run_impl command implementation now makes sure enqueued runs are
managed so that event (if any) can be notified upon completion.

Change xrt::event_queue::task signature to take event_impl as argument
and make task holder same for both async and sync task.

Replace bind with lambda when creating task and move event notification
to lambda body.  Event notification is different for async and sync tasks.

Trick synchronous lambda to handle void and return-value tasks by
using at_exit scope guard trick.

Changed xrt::run_impl::start to take default nullptr event parameter (shared_ptr).  
This add maybe a tiny overhead, but should not be noticeable.

Provide complete complex event graph example using the enqueue APIs.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_enqueue.cpp
+++ b/src/runtime_src/core/common/api/xrt_enqueue.cpp
@@ -77,6 +77,7 @@ public:
       return;
     m_chain.push_back(ev);
     ++ev->m_wait_count;
+    XRT_DEBUGF("event_impl(%d) chains ev(%d) ev_wc(%d)\n", m_uid, ev->m_uid, ev->m_wait_count);
   }
 
   // Try to submit this event for execution.  This function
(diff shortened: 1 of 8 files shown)


stsoe:
@keryell I welcome your feedback on xrt_enqueue.h, it is experimental;  I will probably simplify this more and comment better.  I used some of your feedback from when we chatted last week, e.g. lambda vs bind.  I am not finding this event based execution model all that useful, it's flexible yes, but comes with quite a bit of overhead managing event notification, especially in terms of detecting when a command has completed execution.

I will likely move enqueuing model to some sequential in-order queue with fences of some sort.  The execution then doesn't have to use event notification for other than the fence (maybe), which itself uses synchronous wait() operations all all earlier enqueued operations.

PR #6049, opened 2021-12-07
CR-1113703 and CR-1113150

AShivangi:
- CR-1113150 xbutil validate support in aws f1
- CR-1113703 xbutil examine -r platform on aws f1 not working
#### Bug / issue (if any) fixed, and how
2 issues are fixed by this PR:
- AWS f1 doesn't have xmc sub-driver which was causing query_requests to falsely fail in xbutil. _try/catch were added around query_requests which called sysfs nodes within the xmc driver_
- AWS f1 shell doesn't have a defined path to validate package. _A new option "--path" was added to xbutil validate which takes in the path to the test package. This overrides auto validate package search_

_NOTE: the test package content needs to be standard, i.e., it should have platform.json, .exes and corresponding .xclbins_
- xbutil examine -r all was tested against xbutil --legacy query output
- xbutil validate -p [path to a new dir]
- xbutil validate -p [invalid path]

### DIFF
--- a/src/runtime_src/core/common/info_memory.cpp
+++ b/src/runtime_src/core/common/info_memory.cpp
@@ -86,7 +86,7 @@ struct memory_info_collector
 
   const std::vector<char> mem_topo_raw;    // xclbin raw mem topology
   const std::vector<char> grp_topo_raw;    // xclbin raw grp topology
-  const std::vector<char> mem_temp_raw;    // xclbin temperator raw data
+  std::vector<char> mem_temp_raw;    // xclbin temperator raw data
 
   const mem_topology* mem_topo = nullptr;  // xclbin mem topology from device
   const mem_topology* grp_topo = nullptr;  // xclbin group topology from device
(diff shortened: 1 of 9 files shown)


stsoe (approved):

src/runtime_src/core/common/info_memory.cpp
```diff
@@ -334,12 +334,17 @@ struct memory_info_collector
...
     , grp_topo_raw(xrt_core::device_query<xq::group_topology>(device))
-    , mem_temp_raw(xrt_core::device_query<xq::temp_by_mem_topology>(device))
     , mem_topo(mem_topo_raw.empty() ? nullptr : reinterpret_cast<const mem_topology*>(mem_topo_raw.data()))
     , grp_topo(grp_topo_raw.empty() ? nullptr : reinterpret_cast<const mem_topology*>(grp_topo_raw.data()))
     , mem_stat(xrt_core::device_query<xq::memstat_raw>(device))
-    , mem_temp(mem_temp_raw.empty() ? nullptr : reinterpret_cast<const uint32_t*>(mem_temp_raw.data()))
   {
+    try {
+      mem_temp_raw = xrt_core::device_query<xq::temp_by_mem_topology>(dev);
+      mem_temp = mem_temp_raw.empty() ? nullptr : reinterpret_cast<const uint32_t*>(mem_temp_raw.data());
```
rozumx, line 343: Being that mem_temp is being used below, shouldn't it be initialized to zero prior to executing the first operation in this try block.   This will assure that the mem_temp will be set to a known value.

src/runtime_src/core/common/query_requests.cpp
```diff
@@ -110,17 +110,11 @@ parse(const xrt_core::query::oem_id::result_type& value)
 
 std::string
 xrt_core::query::clock_freq_topology_raw::
-parse(const std::string& clock)
+parse(uint8_t clock)
```
rozumx, line 113: If you are going to refactor the code, why not use with an enumeration type?

src/runtime_src/core/common/query_requests.cpp
```diff
@@ -110,17 +110,11 @@ parse(const xrt_core::query::oem_id::result_type& value)
...
-  {
-   {"DATA_CLK", "Data"},
-   {"KERNEL_CLK", "Kernel"},
-   {"SYSTEM_CLK", "System"},
-  };
-
-  auto clock_str = clock_map.find(clock);
-  return clock_str != clock_map.end() ? clock_str->second : "N/A";
+    // from CLOCK_TYPE enum in xclbin.h
+    std::vector<std::string> clock_names = { "Unused", "Data", "Kernel", "System"};
```
rozumx, line 116:
```suggestion
    static std::vector<std::string> clock_names = { "Unused", "Data", "Kernel", "System"};
```

rozumx (changes requested):
There seems to be changes associated with the clocks that don't match the description.  By chance, is there another CR that is apart of this PR.

rozumx (approved):

dayeh-xilinx:
retest this please.

dayeh-xilinx:
retest this please.

dayeh-xilinx:
retest this please.

PR #6057, opened 2021-12-09
Fix for CR 1116896 : If compile trace option is set, run time metric can be empty

IshitaGhosh:
AIE designs not generating AIE trace output files in XRT flow when compile trace option is set
Also, there was a breakage in generation of shared objects for xdp_aie_trace_plugin and xdp_aie_profile_plugin due to a bad conditional check in xdp/CMakeLists.txt. This PR addresses both the issues and now the plugins are built and AIE trace output files are generated when compile trace option is used.

CR 1116896

unit tests


jvillarre (approved):
Addition looks good to catch compile-time metrics.

PR #6058, opened 2021-12-09
CR-1117245 P2P write from NoDMA to XDMA cause server to crash instead…

houlz0507:
… of reporting error

vm page fault panic
the imported BO is P2P, thus it should use the same way with P2P BO for insert_page.
check imported BO in vm page fault
low
verified with environment mentioned in the CR


maxzhen:
Does this also work for non-P2P BO imported?

houlz0507:
> Does this also work for non-P2P BO imported?

I looked into linux source code. insert_mixed calls into insert_page if it is ordinary memory. So _mixed is a unified call.

maxzhen:
> > Does this also work for non-P2P BO imported?
> 
> I looked into linux source code. insert_mixed calls into insert_page if it is ordinary memory. So _mixed is a unified call.

OK. Please add a comment to clarify so that we don't need to look into the kernel again next time :-)

maxzhen (approved):

dayeh-xilinx:
retest this please.

PR #6059, opened 2021-12-09
Removing obsolete firmware build files

maxzhen:
CMC/SC is not shipped in their own standalone package. We can remove them from XRT's build and packaging flow.
The mgmt.bin is no longer used any more, we can remove the code and its BSP file from XRT repo, too.
Risk is low since these are not used any more.
Modified the CMake file and did a clean full build.
There should be no doc change.


rozumx (approved):
A little be more then commenting out the CMakelLists.txt to prevent downloading the CMC and BMC images.  That said, this is the right thing to do.  Thank you for doing this 👍 .

stsoe (approved):

stsoe:
Looks good.  binary check failed because it triggered on binary file being removed.

PR #6062, opened 2021-12-10
Vitis-3928 IPC facility using XRT BO APIs for export/import

stsoe:
This PR adds limited support for contructing an xrt::bo object from a
buffer exported by another process.

Vitis-3928, requested support.

The implementation is supported on x86 Linux only where both the
system used to build XRT and the system running XRT must have pidfd
support. Per documentation Linux > 5.6 should support this but it is
dependent on a system call that is not implemented by libc and as such
sys/syscall.h must have support as well.  Please check pidfd_getfd(2)
for details, and note also the ptrace permission requirement.

The process id and buffer export handle must be known / communicated
to the importing process.

xrt::bo constructor added to take process id along with buffer export
handle. Implementation code for importing the exported buffer is
conditional on pidfd support or errors out otherwise.

No risks, code is new and currently unused.

Sample test case added for illustration.

@uday610 do we need documentation above and beyond what is documented
in API header file?

### DIFF
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -185,6 +185,14 @@ public:
     size = prop.size;
   }
 
+  bo_impl(xclDeviceHandle dhdl, pid_t pid, xclBufferExportHandle ehdl)
+    : device(xrt_core::get_userpf_device(dhdl)), handle(device->import_bo(pid, ehdl)), free_bo(true)
+  {
+    xclBOProperties prop{};
+    device->get_bo_properties(handle, &prop);
+    size = prop.size;
+  }
+
   bo_impl(xclDeviceHandle dhdl, xcl_buffer_handle xhdl)
     : device(xrt_core::get_userpf_device(dhdl)), handle(xhdl.bhdl), free_bo(false)
   {
@@ -465,14 +473,39 @@ public:
 // class buffer_imported - Buffer imported from another device
 //
 // The exported buffer handle is an opaque type from a call
-// to export_buffer() on a buffer to be exported.
+// to export_buffer() on a buffer to be exported.  The exported
+// buffer can be imported within same process or from another
+// process (linux pidfd support required)
 class buffer_import : public bo_impl
 {
   void* hbuf;
 
 public:
+  // buffer_import() - Import the buffer 
+  //
+  // @device:  device to import to
+  // @ehdl:    export handle obtained by calling export_buffer
   buffer_import(xclDeviceHandle dhdl, xclBufferExportHandle ehdl)
     : bo_impl(dhdl, ehdl)
+  {
+    try {
+      hbuf = device->map_bo(handle, true);
+    }
+    catch (const std::exception&) {
+      hbuf = nullptr;
+    }
+  }
+
+  // buffer_import() - Import the buffer from another process 
+  //
+  // @device:  device to import to
+  // @pid:     process id of exporting process
+  // @ehdl:    export handle obtained from exporting process
+  //
+  // This consrructor works on linux only and require pidfd support in
+  // linux kernel.
+  buffer_import(xclDeviceHandle dhdl, pid_t pid, xclBufferExportHandle ehdl)
+    : bo_impl(dhdl, pid, ehdl)
   {
     try {
       hbuf = device->map_bo(handle, true);
@@ -826,6 +859,12 @@ alloc_import(xclDeviceHandle dhdl, xclBufferExportHandle ehdl)
 }
 
 static std::shared_ptr<xrt::bo_impl>
+alloc_import_from_pid(xclDeviceHandle dhdl, pid_t pid, xclBufferExportHandle ehdl)
+{
+  return std::make_shared<xrt::buffer_import>(dhdl, pid, ehdl);
+}
+
+static std::shared_ptr<xrt::bo_impl>
 alloc_sub(const std::shared_ptr<xrt::bo_impl>& parent, size_t size, size_t offset)
 {
   return std::make_shared<xrt::buffer_sub>(parent, size, offset);
@@ -948,6 +987,12 @@ bo::
 bo(xclDeviceHandle dhdl, xclBufferExportHandle ehdl)
   : handle(xdp::native::profiling_wrapper("xrt::bo::bo",
 	   alloc_import, dhdl, ehdl))
+{}
+
+bo::
+bo(xclDeviceHandle dhdl, pid_t pid, xclBufferExportHandle ehdl)
+  : handle(xdp::native::profiling_wrapper("xrt::bo::bo",
+            alloc_import_from_pid, dhdl, pid , ehdl))
 {}
 
 bo::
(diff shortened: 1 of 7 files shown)


uday610 (approved):
Thanks a lot!!
I dont think any additional documentation is needed because we dont have any specific documentation about inter-process communication. Perhaps we should add interprocess-communication documentation in future when we will have capability of sending event FD as well (https://jira.xilinx.com/browse/VITIS-1115) and more documentation about it

src/runtime_src/core/common/api/xrt_bo.cpp
```diff
@@ -185,6 +185,14 @@ class bo_impl
     size = prop.size;
   }
 
+  bo_impl(xclDeviceHandle dhdl, pid_t pid, xclBufferExportHandle ehdl)
+    : device(xrt_core::get_userpf_device(dhdl)), handle(device->import_bo(pid, ehdl)), free_bo(true)
+  {
+    xclBOProperties prop{};
```
rozumx, line 191: Learn something new today, didn't know we could do this.  

I always used the following syntax `xclBOProperties prop = {};`

src/runtime_src/core/include/windows/types.h
```diff
@@ -18,5 +18,6 @@
 #define core_include_windows_types_h_
 
 typedef size_t ssize_t;
+typedef int pid_t;
```
rozumx, line 21: Going forward, I thought we were to use `using` instead of `typedef`

tests/xrt/import_bo/import.cpp
```diff
@@ -0,0 +1,186 @@
...
+// % import.exe -k verify.xclbin
+//
+// # userspace allocates host buffer (fails in xrt-2.12.x)
+// % import.exe -k verify.xclbin --ubuf
+
+// Golden pattern from kernel
+static constexpr char gold[] = "Hello World\n";
+
+// A pipe to communicate exported bo handle to child
+static int talk[2];
```
rozumx, line 44: Magic number.  What is so special about 2 :^)

tests/xrt/import_bo/import.cpp
```diff
@@ -0,0 +1,186 @@
...
+            << "  -d <bdf | device_index>\n"
+            << ""
+            << "  [--ubuf]: BO host backing should be created in user space (default kernel space)\n";
+}
+
+
+static void
+child(const std::string& device_id, pid_t pid)
+{
+  close(talk[1]);
```
rozumx, line 60: Yeah, we need to define:
1. The size of the talk array.
2. What element 0 and element 1 represents.

tests/xrt/import_bo/import.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+  std::cout << "child pid: " << pid << '\n';
+  std::cout << "child fd: " << fd << '\n';
+  
+  xrt::device device{device_id};
+  xrt::bo bo{device, pid, fd};
+
+  try {
+    auto bo_data = bo.map<char*>();
+    int count = 5;
```
rozumx, line 73: A comment on why 5 is a good number to use here.

tests/xrt/import_bo/import.cpp
```diff
@@ -0,0 +1,186 @@
...
+    while (!std::equal(std::begin(gold), std::end(gold), bo_data) && --count)
+      bo.sync(XCL_BO_SYNC_BO_FROM_DEVICE);
+
+    std::cout << "child reads: " << bo_data << '\n';
+    if (!count)
+      std::cout << "child times out\n";
+    else
```
rozumx, line 80: Why the else?  Looking at the code, it will be executing two `bo.write("child");` operations, is this desired?

tests/xrt/import_bo/import.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+static void
+parent(const std::string& device_id, const std::string& xclbin_fnm, xrt::bo::flags flags)
+{
+  close(talk[0]);
+  
+  xrt::device device{device_id};
+  auto uuid = device.load_xclbin(xclbin_fnm);
+  xrt::kernel hello(device, uuid, "hello");
+  xrt::bo bo(device, 1024, flags, hello.group_id(0));
```
rozumx, line 97: Magic number of 1K

tests/xrt/import_bo/import.cpp
```diff
@@ -0,0 +1,186 @@
...
+  auto bo_data = bo.map<char*>();
+
+  // clear device data
+  std::fill(bo_data, bo_data + 1024, 0);
```
rozumx, line 101: Please, define this buffer as a variable.  doing so will remove the need to update this value in multiple locations if it should happen to change in the future.

tests/xrt/import_bo/import.cpp
```diff
@@ -0,0 +1,186 @@
...
+  std::vector<std::string> args(argv+1,argv+argc);
+  std::string xclbin_fnm;
+  std::string device_id = "0";
+  xrt::bo::flags flags = xrt::bo::flags::cacheable;
+  
+  std::string cur;
+  for (auto& arg : args) {
+    if (arg == "-h") {
+      usage();
+      return 1;
```
rozumx, line 136: Why should help produce an error (e.g., a return value of 1)?  
I would think that the operation was successful, one requested a help menu and one was generated.

rozumx (changes requested):

src/runtime_src/core/include/windows/types.h
stsoe: Nope. this is C code.

tests/xrt/import_bo/import.cpp
stsoe: It's not magic, it is a linux pipe, man pipe https://man7.org/linux/man-pages/man2/pipe.2.html

tests/xrt/import_bo/import.cpp
stsoe: https://man7.org/linux/man-pages/man2/pipe.2.html

tests/xrt/import_bo/import.cpp
stsoe: yep indeed

tests/xrt/import_bo/import.cpp
stsoe: Points to a missing sleep, thanks.

tests/xrt/import_bo/import.cpp
stsoe: Nah, not really,  Allocate a buffer of 1K, this is not important, this what all our tests are doing.  I am not going to overkill this other than adding a comment.

tests/xrt/import_bo/import.cpp
stsoe: No.   bo_data is mapped buffer object data, your comment makes no sense to me.

tests/xrt/import_bo/import.cpp
stsoe: Doesn't really matter. IMO.

dayeh-xilinx:
retest this please.

rozumx (approved):

PR #6075, opened 2021-12-15
User managed interrupt: decrement on each poll

sarab96:
On poll num of events was not decremented and thus event check was incorrect

ucu_event is decremented on each poll execution and thus will always have correct value on each poll call

Tested locally


sarab96:
retest this please. 
This pipeline issue is fixed

dayeh-xilinx:
retest this please.

sarab96:
Closing after discussing with Min. Right keeping that only read will clear num of event counter. So poll + read will be required together

PR #6076, opened 2021-12-15
User managed interrupt: add check on number of interrupt

sarab96:
None. Warning added for number of interrupts. Usual behavior expects only one interrupt

New check added

None

### DIFF
--- a/src/runtime_src/core/pcie/linux/device_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/device_linux.cpp
@@ -1012,9 +1012,12 @@ void
 device_linux::
 wait_ip_interrupt(xclInterruptNotifyHandle handle)
 {
-  int pending = 0;
+  int num_interrupts = 0;
   if (::read(handle, &pending, sizeof(pending)) == -1)
     throw error(errno, "wait_ip_interrupt failed POSIX read");
+
+  if (num_interrupts > 1)
+    xrt_core::message::send(xrt_core::message::severity_level::warning, "XRT", "More than one interrupt was received");
 }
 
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -1012,9 +1012,12 @@ void
 device_linux::
 wait_ip_interrupt(xclInterruptNotifyHandle handle)
 {
-  int pending = 0;
+  int num_interrupts = 0;
   if (::read(handle, &pending, sizeof(pending)) == -1)
     throw error(errno, "wait_ip_interrupt failed POSIX read");
+
+  if (num_interrupts > 1)
+    xrt_core::message::send(xrt_core::message::severity_level::warning, "XRT", "More than one interrupt was received");
```
stsoe, line 1020: I don't see what this PR is doing.  Don't just add unused code.

stsoe (changes requested):

src/runtime_src/core/pcie/linux/device_linux.cpp
sarab96: Updated the code

stsoe (changes requested):
I understand the change, but I don't understand the bigger picture?
What does num_interrupts represent and why is it a warning if the number is > 1?
It is important that you document the changes in code (comments) if it is not obvious.

sarab96:
Will discuss with Soren and resubmit later. Right now this generates too many warnings with the way xrt::interrupt::wait() is implemented. As wait() automatically reenables interrupts.

PR #6085, opened 2021-12-21
Fix compile error in XlcBin::setKeyValue by converting keyvalue to reference

vkomenda:
The PR fixes compilation with GCC 11.2.

On Debian testing, compilation with GCC 11.2 fails with
```
/home/vk/src/x/XRT.fail/src/runtime_src/tools/xclbinutil/XclBinClass.cxx: In member function ‘void XclBin::setKeyValue(const string&)’:
/home/vk/src/x/XRT.fail/src/runtime_src/tools/xclbinutil/XclBinClass.cxx:1588:21: error: loop variable ‘keyvalue’ creates a copy from type ‘const boost::property_tree::basic_ptree<std::__cxx11::basic_string<char>, std::__cxx11::basic_string<char> >’ [-Werror=range-loop-construct]
 1588 |     for (const auto keyvalue : keyValues) {
      |                     ^~~~~~~~
/home/vk/src/x/XRT.fail/src/runtime_src/tools/xclbinutil/XclBinClass.cxx:1588:21: note: use reference type to prevent copying
 1588 |     for (const auto keyvalue : keyValues) {
      |                     ^~~~~~~~
      |                     &
```

A reference was added as per the GCC hint.

Possible incompatibility with earlier versions of GCC.

Only GCC 11.2 was tested. The PR should be tested against earlier versions of GCC to confirm it doesn't break those.

None.

### DIFF
--- a/src/runtime_src/tools/xclbinutil/XclBinClass.cxx
+++ b/src/runtime_src/tools/xclbinutil/XclBinClass.cxx
@@ -1585,7 +1585,7 @@ XclBin::setKeyValue(const std::string& _keyValue)
 
     // Now create a new tree to add back into the section
     boost::property_tree::ptree ptKeyValuesNew;
-    for (const auto keyvalue : keyValues) {
+    for (const auto &keyvalue : keyValues) {
       ptKeyValuesNew.push_back(std::make_pair("", keyvalue));
     }
 


rozumx (approved):

rozumx (approved):

stsoe:
ok to test

keryell:
I am just hitting this bug on xsjsycl41. :-(
It would be nice to add some GitHub action to compile with various versions of the compiler, independently from the internal CI.
Perhaps you can adapt https://github.com/triSYCL/triSYCL/tree/master/.github/workflows if you are looking for some inspiration.

stsoe:
> Perhaps you can adapt https://github.com/triSYCL/triSYCL/tree/master/.github/workflows if you are looking for some inspiration.

Much too slow without self-hosted runners.

PR #6091, opened 2021-12-22
CR-1116870: 1.Emu driver uses xclbin_parser logic 2. Canary is neat

udaybhas-xilinx:
Description::
1.Emu driver uses xclbin_parser logic 

CR's:: 1. https://jira.xilinx.com/browse/CR-1116870
          2. https://jira.xilinx.com/browse/CR-1116752

Problem solved: Emu driver uses xclbin parser logic instead of parsing the xml file

Risk: Low

Tests: 1. Canary is neat.

### DIFF
--- a/src/runtime_src/core/common/xclbin_parser.cpp
+++ b/src/runtime_src/core/common/xclbin_parser.cpp
@@ -540,6 +540,54 @@ get_cus(const ip_layout* ip_layout, const std::string& kname)
   return ips;
 }
 
+std::map<std::string, std::vector<instance_object>>
+get_kernel_instance_map( const char* xml_data, size_t xml_size) {
+ 
+  std::map<std::string, std::vector<instance_object>> kernelInstanceMap;
+  std::vector<instance_object> instances;  
+
+  pt::ptree xml_project;
+  std::stringstream xml_stream;
+  xml_stream.write(xml_data, xml_size);
+  pt::read_xml(xml_stream, xml_project);
+  std::string kernelName;
+
+  for (auto& xml_kernel : xml_project.get_child("project.platform.device.core")) 
+  {
+    if (xml_kernel.first != "kernel")
+    continue;
+    kernelName = xml_kernel.second.get<std::string>("<xmlattr>.name");
+
+    std::vector<instance_object> instances;
+    for (auto& xml_inst : xml_kernel.second) 
+    {
+      if (xml_inst.first != "instance")
+        continue;
+      std::string instanceName = xml_inst.second.get<std::string>("<xmlattr>.name");
+
+      for (auto& xml_remap : xml_inst.second)
+      {
+        if (xml_remap.first != "addrRemap")
+          continue;
+
+        uint64_t base = convert(xml_remap.second.get<std::string>("<xmlattr>.base"));
+        size_t range = convert(xml_remap.second.get<std::string>("<xmlattr>.range"));
+        std::string port = xml_remap.second.get<std::string>("<xmlattr>.port","");
+        instances.emplace_back(instance_object{
+        instanceName
+        ,base
+        ,range
+        ,port
+        });
+
+      }
+    
+    }
+    kernelInstanceMap[kernelName] = instances;
+  }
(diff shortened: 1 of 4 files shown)


stsoe (changes requested):
I don't think changes / addition to xclbin_parser.cpp to parse xml data are necessary.
Please leverage xrt::xclbin APIs, both public (include/experimental/xrt_xclbin.h) and internal ones (core/common/api/xclbin_int.h).  The xrt::xclbin objects capture all meta data associated with an xclbin.

Secondly, none of the code in this PR is documented, if any new code is added, please make sure to add good comments for new functions and explain what the code is doing and why.

I suggest this PR is closed, and reworked per above suggestions.

src/runtime_src/core/common/xclbin_parser.cpp
```diff
@@ -540,6 +540,54 @@ get_cus(const ip_layout* ip_layout, const std::string& kname)
   return ips;
 }
 
+std::map<std::string, std::vector<instance_object>>
+get_kernel_instance_map( const char* xml_data, size_t xml_size) {
```
stsoe, line 544: This is xrt::xclbin::get_ips() and variants.  The xrt::xclbin::ip has member functions for accessing most of the data you need,  if some data is missing, please let @stsoe know.

src/runtime_src/core/common/xclbin_parser.cpp
```diff
@@ -983,6 +1031,103 @@ get_vbnv(const axlf* top)
   return {vbnv, strnlen(vbnv, vbnv_length)};
 }
 
+std::string
+get_device_name(const char* xmlFile,size_t xmlFileSize)
```
stsoe, line 1035: The device name per XML is not currently extracted by xrt::xclbin, but please use same approach as used for project name (see common/api/xclbin_int.h, xrt_xclbin.cpp).

src/runtime_src/core/common/xclbin_parser.cpp
```diff
@@ -983,6 +1031,103 @@ get_vbnv(const axlf* top)
...
+ 
+    }
+  }
+    
+  return fpgaDevice;
+	
+}
+
+bool
+check_if_multiple_platform_exists(const char* xml_data, size_t xml_size)
```
stsoe, line 1067: This is always false, no where in XRT do we support multiple platforms per xml

src/runtime_src/core/common/xclbin_parser.cpp
```diff
@@ -983,6 +1031,103 @@ get_vbnv(const axlf* top)
...
+      //Give error and return from here
+      return false;
+    }
+  }
+  return true;
+}
+
+bool
+check_if_multiple_core_exists(const char* xml_data, size_t xml_size)
+{
```
stsoe, line 1090: Also false, the xml will never have multiple cores, that is not supported nor will it ever be supported via XML core element.

src/runtime_src/core/common/xclbin_parser.cpp
```diff
@@ -983,6 +1031,103 @@ get_vbnv(const axlf* top)
...
+      //Give error and return from here
+      return false;
+    }
+  }
+  return true;
+}
+
+std::map<std::string,uint64_t>
+get_address_range_map(const char* xml_data, size_t xml_size)
+{
```
stsoe, line 1111: This is a kernel property which is accessible via core/common/api/xclbin_int.h:get_properties(kernel).

src/runtime_src/core/common/xclbin_parser.h
```diff
@@ -78,6 +78,15 @@ struct kernel_object
...
 };
 
+//instance object
+struct instance_object
+{
+  std::string name;
+  uint64_t base;
+  size_t range;
+  std::string port;
+};
```
stsoe, line 88: Should not be needed.

stsoe (changes requested):

udaybhas-xilinx:
Creating a new pull request with changes made..

PR #6072, opened 2021-12-14
Add drc to skip if xclbin is not found

AShivangi:
Missed a drc check to make sure that the xclbin path is valid

Sample output:
```
$ ./xbutil2_static validate -d -p test-xclbins/ -r "Bandwidth kernel" --verbose
Verbose: Enabling Verbosity
Verbose: Sub command: --path
Starting validation for 1 devices

Validate Device           : [0000:85:00.0]
    Platform              : xilinx_aws-vu9p-f1_dynamic-shell
    SC Version            : N/A
    Platform ID           : 0xabcd
-------------------------------------------------------------------------------
Test 1 [0000:85:00.0]     : Bandwidth kernel 
    Description           : Run 'bandwidth kernel' and check the throughput
    Test Status           : [SKIPPED]
-------------------------------------------------------------------------------
Validation completed
```

PR #6094, opened 2021-12-22
Add drc to skip if xclbin is not found (#6072)

AShivangi:
* add drc to skip if xclbin is not found

* remove redundant drc check

(cherry picked from commit c440ece357256b30ec05cfb7ecdbab8de1059953)


chvamshi-xilinx:
retest this please

PR #6095, opened 2021-12-22
CR-1095534 ASTeR xbmgmt(2) program --sc_firmware command is not available

dbenusov-xilinx:
[https://jira.xilinx.com/browse/CR-1095534](url)
xbmgmt(1) command is:

xbmgmt flash -sc_firmware --path /opt/xilinx/firmware/sc-fw/u50/sc-fw-u50-5.2.6*.txt --card 0000:63:00.0

The xbmgmt(2) equivalent command does not exist yet.

Added the shell and sc options into the --base sub option.

The SC version number is updated when only the shell is flashed. The SC itself is not updated but that may confuse some people.

Manual testing. There needs to be an additional test with the .txt file for SC programming only.

Shell only
```
root@xsjpranjal01:~# xbmgmt program -d 02:00 --base shell --image="/lib/firmware/xilinx/10ee-500c-000e-000000005e278820.xsabin"
Bitstream guard installed on flash @0x1002000
Extracting bitstream from MCS data:
.................................................
Extracted 51305636 bytes from bitstream @0x1002000
Writing bitstream to flash 0:
.................................................
Bitstream guard removed from flash
```
SC only
```
root@xsjpranjal01:~# xbmgmt program -d 02:00 --base sc --image="/lib/firmware/xilinx/10ee-500c-000e-000000005e278820.xsabin"
Stopping user function...
INFO     : found 5 sections
[PASSED] : SC successfully updated < 27s >
INFO     : Loading new firmware on SC
```

Need to update documentation to reflect the new options for --base.


src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -840,54 +789,44 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
...
+      auto & workingDevice = deviceCollection[0];
+      XBU::verbose("Sub command: --base");
+      XBUtilities::sudo_or_throw("Root privileges are required to update the devices flash image");
+      if (update.compare("all") == 0) {
+        if (validated_images.empty())
+          auto_flash(workingDevice);
+        else if(boost::filesystem::extension(validated_images[0]).compare(".xsabin") == 0)
+          auto_flash(workingDevice, flashType, validated_images[0]);
+        else
+          throw xrt_core::error("Please provide a valid xsabin file or specify a specific base to flash");
```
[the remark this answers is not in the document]
dbenusov-xilinx: That is why the warning is there. If you pass a non xsabin file with --base all it gives you a warning

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -840,54 +789,44 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
...
+    const auto validated_images = find_flash_image_paths(image);
+    if (validated_images.size() <= 2) {
+      // Get the device
+      auto & workingDevice = deviceCollection[0];
+      XBU::verbose("Sub command: --base");
+      XBUtilities::sudo_or_throw("Root privileges are required to update the devices flash image");
+      if (update.compare("all") == 0) {
+        if (validated_images.empty())
+          auto_flash(workingDevice);
+        else if(boost::filesystem::extension(validated_images[0]).compare(".xsabin") == 0)
```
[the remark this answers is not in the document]
dbenusov-xilinx: The first means that no image after the find_image_flash_paths call was valid (it doesn't exist or was not given in the first place). So we assume auto flash with no image specified. It would probably be better to write image.empty() instead. Second means we have a file that exists, but, the extension may not be xsabin. The find_flash_image_path function returns a file if it exists or matches a shell.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -840,54 +789,44 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
...
+          if (!validated_images.empty())
+            update_shell(workingDevice.get()->get_device_id(), validated_images, flashType);
+          else
+            throw xrt_core::error("Usage: xbmgmt program --device='0000:00:00.0' --base shell --image='/path/to/flash_image OR shell_name'");
+      }
+      else
+        throw xrt_core::error("Please specify a valid value");
     }
+    else
+      throw xrt_core::error("Please specify either 1 or 2 flash images");
```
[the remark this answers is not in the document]
dbenusov-xilinx: I can format out the middle 2 if statements into a separate function, but, I do not think that would help with readability. I am open to any changes though

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -840,54 +789,44 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
...
-
-      if (update.compare("sc") == 0)
-        throw xrt_core::error("SC only update is not supported");
-       
-      throw xrt_core::error("Please specify a valid value");
+    // Get a list of images known exist
+    const auto validated_images = find_flash_image_paths(image);
+    if (validated_images.size() <= 2) {
+      // Get the device
+      auto & workingDevice = deviceCollection[0];
```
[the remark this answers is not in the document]
dbenusov-xilinx: Makes sense. There is a drc check at the beginning of the processing, so that is covered. But something like this would probably be out of scope for this PR. If we want to put it in I am not opposed though.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -115,21 +88,21 @@ update_shell(unsigned int index, const std::string& flashType,
   if(!flasher.isValid())
     throw xrt_core::error(boost::str(boost::format("%d is an invalid index") % index));
 
-  if (primary.empty())
+  if (image_paths.empty())
     throw xrt_core::error("Base not specified");
 
-  auto pri = std::make_unique<firmwareImage>(primary.c_str(), MCS_FIRMWARE_PRIMARY);
+  auto pri = std::make_unique<firmwareImage>(image_paths[0].c_str(), MCS_FIRMWARE_PRIMARY);
```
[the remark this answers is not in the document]
dbenusov-xilinx: I used [0] instead of front to match the other reference in the function using [1] on [https://github.com/Xilinx/XRT/pull/6095/files#diff-5b1fd3c2284500459e3cda9231fc218d31e9798c321e9ea84e9ecc79e054cef0R100]()

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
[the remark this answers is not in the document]
AShivangi: Oh, so the first case is where we do the auto-update, i.e., find the file on disk. and 2nd is where the user specifies the `--image` option. Right?

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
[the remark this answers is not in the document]
AShivangi: Oh, I see what you mean. Yeah, that does make sense I guess. Instead of using specify 2 times, why don't we change it to `or specify the type of base you would like to flash`

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -115,21 +88,21 @@ update_shell(unsigned int index, const std::string& flashType,
...
   if (pri->fail())
-    throw xrt_core::error(boost::str(boost::format("Failed to read %s") % primary));
+    throw xrt_core::error(boost::str(boost::format("Failed to read %s") % image_paths[0]));
```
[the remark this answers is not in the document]
dbenusov-xilinx: Same as the above comment id rather have the references into image_paths be consistent with both references indexing into the vector.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -68,45 +68,18 @@ namespace po = boost::program_options;
...
-  std::cout << boost::format("%-8s : %s \n") % "INFO" % "Base flash image has been programmed successfully.";
-}
 
-/*
- * Update shell on the board for manual flash
+/**
+ * @brief Update shell on the board for manual flash
+ * 
+ * @param[in] index The index of the board to be flashed
+ * @param[in] image_paths The images to plash onto the board
```
keryell, line 76: plash?

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -68,45 +68,18 @@ namespace po = boost::program_options;
...
+ * @param[in] flash_type How the board should be flashed. Typically left as blank
```
keryell, line 77: Could you explain what `type`? Because funnily in the code here it is a string, so almost as typed as `void*`... :-)

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -68,45 +68,18 @@ namespace po = boost::program_options;
...
  */
 static void 
-update_shell(unsigned int index, const std::string& flashType,
-  const std::string& primary, const std::string& secondary)
+update_shell(unsigned int index, const std::vector<std::string>& image_paths, const std::string& flash_type="")
```
keryell, line 80:
```suggestion
update_shell(unsigned int index, const std::vector<std::string>& image_paths, const std::string& flash_type = "")
```
Check also the line is shorter than 80 characters.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
keryell: We can have several images? What is the semantics? Loading all of them on the board?

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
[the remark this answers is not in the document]
keryell: Both styles look good.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -467,7 +440,9 @@ update_shell(unsigned int boardIdx, DSAInfo& candidate)
   // Program the shell
   boost::format programFmt("[%s] : %s...\n");
   std::cout << programFmt % flasher.sGetDBDF() % "Updating base (e.g., shell) flash image";
-  update_shell(boardIdx, candidate.file, candidate.file);
+  std::vector<std::string> validated_image;
+  validated_image.push_back(candidate.file);
```
keryell, line 444: Do not obfuscate the code
```suggestion
  std::vector<std::string> validated_image { candidate.file };
```

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -659,44 +605,47 @@ program_plp(const xrt_core::device* dev, const std::string& partition)
...
 }
 
+/**
+ * @brief Create a list of images that are known to exist based on given names and paths
+ * 
+ * @param[in] image_list The list of images paths and names to be validated
+ * @return std::vector<std::string> A list of known and valid images for flashing
+ */
 static std::vector<std::string>
 find_flash_image_paths(const std::vector<std::string> image_list)
```
keryell, line 615: While you are here
```suggestion
find_flash_image_paths(const std::vector<std::string>& image_list)
```

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -435,7 +408,7 @@ update_sc(unsigned int boardIdx, DSAInfo& candidate)
  * Helper method for auto_flash
  */
 static bool  
-update_shell(unsigned int boardIdx, DSAInfo& candidate)
+update_shell(unsigned int boardIdx, DSAInfo& candidate, std::string flash_type)
```
keryell, line 411:
```suggestion
update_shell(unsigned int boardIdx, DSAInfo& candidate, std::string& flash_type)
```

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -115,21 +88,21 @@ update_shell(unsigned int index, const std::string& flashType,
   if(!flasher.isValid())
```
keryell, line 88: While you are here :-)
```suggestion
  if (!flasher.isValid())
```

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -659,44 +605,47 @@ program_plp(const xrt_core::device* dev, const std::string& partition)
...
-    
-    for(auto const& shell : installedShells) {
-      if(shell_name.compare(shell.name) == 0) {
-        multiple_shells++;
-        path = shell.file;
-      }
-    }
+  for(const auto& img : image_list) {
+    // Check if the passed in image is absolute path
+    if(boost::filesystem::exists(boost::filesystem::path(img)))
```
keryell, line 622: Perhaps this works:
```suggestion
    if (boost::filesystem::exists(img))
```

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -659,44 +605,47 @@ program_plp(const xrt_core::device* dev, const std::string& partition)
...
-    
-    //if multiple shells with the same vbnv are installed on the system, we don't want to 
-    //blindly update the device. in this case, the user needs to specify the complete path
-    if(multiple_shells > 1) 
-      throw xrt_core::error("Specified base matched mutiple installed bases. Please specify the full path.");
+      /*
+       * Iterate over installed shells
+       * Checks the vbnv against the vnbv passed in by the user
+       */
+      std::string img_path = "";
```
keryell, line 631: Do not repeat what the constructor does
```suggestion
      std::string img_path;
```

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -659,44 +605,47 @@ program_plp(const xrt_core::device* dev, const std::string& partition)
...
+      for(auto const& shell : installedShells) {
+        if(img.compare(shell.name) == 0) {
```
keryell, line 633:
```suggestion
        if (img.compare(shell.name) == 0) {
```

keryell: What about replacing the `if` chain with `&&`?

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -659,44 +605,47 @@ program_plp(const xrt_core::device* dev, const std::string& partition)
...
+          if(img_path.empty())
+            img_path = shell.file;
+          /*
+           * If multiple shells with the same vbnv are installed on the system, we don't want to
+           * blindly update the device. in this case, the user needs to specify the complete path
```
keryell, line 638:
```suggestion
           * blindly update the device. In this case, the user needs to specify the complete path
```

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -659,44 +605,47 @@ program_plp(const xrt_core::device* dev, const std::string& partition)
...
+           */
+          else
+            throw xrt_core::error("Specified base matched mutiple installed bases. Please specify the full path.");
+        }
+      }
 
-    return path;
-  };
+      if(img_path.empty())
```
keryell, line 645:
```suggestion
      if (img_path.empty()) 
```
What about configuring you editor or IDE?

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -751,10 +700,10 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
                                                                       "  Name (and path) of the partition.")
 
     // TODO: Auto update the 'base' values
-    ("base,b", boost::program_options::value<decltype(update)>(&update)->implicit_value("all"), "Update the persistent images and/or the Satellite controller (SC) firmware image."/*  Value values:\n"
-                                                                         "  ALL   - All images will be updated"
-                                                                         "  FLASH - Flash image\n"
-                                                                         "  SC    - Satellite controller"*/)
+    ("base,b", boost::program_options::value<decltype(update)>(&update)->implicit_value("all"), "Update the persistent images and/or the Satellite controller (SC) firmware image.  Value values:\n"
```
keryell, line 703: Line too long.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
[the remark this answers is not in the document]
keryell: I guess if there are some questions about the logic here, are not comments missing in the code?

keryell, line 803:
```suggestion
        else if (boost::filesystem::extension(validated_images[0]).compare(".xsabin") == 0)
```

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -840,54 +789,44 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
...
+            throw xrt_core::error("Usage: xbmgmt program --device='0000:00:00.0' --base sc --image='/path/to/flash_image");
+      }
+      else if (update.compare("shell") == 0) {
+          if (!validated_images.empty())
+            update_shell(workingDevice.get()->get_device_id(), validated_images, flashType);
+          else
+            throw xrt_core::error("Usage: xbmgmt program --device='0000:00:00.0' --base shell --image='/path/to/flash_image OR shell_name'");
+      }
+      else
+        throw xrt_core::error("Please specify a valid value");
```
keryell, line 821: For what? Clarify the error messages as much as possible!

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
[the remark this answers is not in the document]
keryell: @AShivangi That is a very good point. For example the LLVM coding style favors early exit to avoid this.
For example here you could have just very early:
```c++
if (validated_images.size() > 2) { 
       throw xrt_core::error("Please specify only 1 or 2 flash images"); 
```
and then you do not need the `else`. :-)
You can generalize the idea everywhere with `throw`/`break`/`return`...

keryell (changes requested):
It looks like you are solving a problem I was recently banging my head on! :-)
Thanks.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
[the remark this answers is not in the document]
AShivangi: No that's fine. Let's break it up and do this work in the next PR as this PR is already making a huge change. Please do make a note of it so that you don't forget it :)

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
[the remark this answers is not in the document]
AShivangi: Yes, some comments will be helpful. In fact, the confusing part for me was the function name `auto_flash`. Auto flash suggests that we look up the xsabin for the user and flash it. Thus the `else if` with passing n the image doesn't qualify as an auto_flash. Ideally, the function should be renamed to `flash_shell_and_SC` and have a comment in L801 indicating that we are doing an auto flash

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
dbenusov-xilinx: I updated the comment to explain this defines which format is used to flash the board.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
dbenusov-xilinx: Is 80 lines the max length or 120? There are a lot of statements that go over 80

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
dbenusov-xilinx: For this one it must be left separate. We can only throw an error if multiple shells match the current image, not if the current image does not match the shell. IT would be more code than it is now. Ill add an extra comment to explain why.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
dbenusov-xilinx: There was once a board that took two images when being flashed as far as I was  told. So it is in fact a vector of images to flash. It was much cleaner than having multiple image variables.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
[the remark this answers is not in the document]
dbenusov-xilinx: I was able to get remove all but the `all` condition statement. And removed the images validation statement to cause an earlier exit.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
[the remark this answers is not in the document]
dbenusov-xilinx: Code is much simpler and has comments now!

keryell (approved):
Thanks for the cleaning!

dayeh-xilinx:
retest this please

dbenusov-xilinx:
retest this please

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -746,29 +697,36 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
...
-                                                                         "  SC    - Satellite controller"*/)
-    ("user,u", boost::program_options::value<decltype(xclbin)>(&xclbin), "The xclbin to be loaded.  Valid values:\n"
-                                                                      "  Name (and path) of the xclbin.")
-    ("image", boost::program_options::value<decltype(image)>(&image)->multitoken(), "Specifies an image to use used to update the persistent device.  Value values:\n"
-                                                                    "  Name (and path) to the mcs image on disk\n"
-                                                                    "  Name (and path) to the xsabin image on disk")
-    ("revert-to-golden", boost::program_options::bool_switch(&revertToGolden), "Resets the FPGA PROM back to the factory image. Note: The Satellite Controller will not be reverted for a golden image does not exist.")
-    ("help,h", boost::program_options::bool_switch(&help), "Help to use this sub-command")
+    ("device,d", boost::program_options::value<decltype(device)>(&device)->multitoken(),
```
rozumx, line 700: Ow!  This new format change is difficult to read, why the changes?

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -840,54 +799,54 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
...
-      update_shell(workingDevice->get_device_id(), flashType, image_paths.front(), (image_paths.size() == 2 ? image_paths[1]: ""));
-    }
-    
-    if (image.size() > 2)
-      throw xrt_core::error("Please specify either 1 or 2 flash images");
-    return;
-  }
+  // Only two images options are supported
+  if (image.size() > 2)
+    throw xrt_core::error("Please specify either 1 or 2 flash images");
```
rozumx, line 804: Lets update this message as well.  Something like,
```suggestion
    throw xrt_core::error("Multiple flash images provided.  Please specify either 1 or 2 flash images.");
```

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -840,54 +799,54 @@ SubCmdProgram::execute(const SubCmdOptions& _options) const
...
+     * Fail early here to reduce additional conditions below
+     * Technically validated_images will never be empty as: if image is not empty but has a bad
+     * path or bad shell name find_flash_image_paths exits early. This statement can be removed
+     * or left here as a precaution.
+     */
+    if (validated_images.empty())
+      throw xrt_core::error("Please provide a valid xsabin file or specify the type of base to flash");
+
+    // Get the device
+    auto & workingDevice = deviceCollection[0];
```
rozumx, line 820: What was the thinking behind moving where this variable was declared.  

I ask for it was original declared after line 801.  Doing so allowed the developer to quickly see that the deviceCollection[0] assignment was valid for it was original checked as being not empty.

rozumx (changes requested):

dayeh-xilinx:
retest this please.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -68,68 +68,42 @@ namespace po = boost::program_options;
...
+ * 
+ * @param[in] index The index of the board to be flashed
+ * @param[in] image_paths The images to flash onto the board
+ * @param[in] flash_type The format in which the board will be flashed. Leave 
+ *                       blank to use the boards default flashing mode
  */
 static void 
-update_shell(unsigned int index, const std::string& flashType,
-  const std::string& primary, const std::string& secondary)
+update_shell(unsigned int index, const std::vector<std::string>& image_paths, const std::string& flash_type = "")
```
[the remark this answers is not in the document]
dbenusov-xilinx: No problem with implementing the map but the issue stems from the input of the command line arguments passed in. It would still result in the 0 index being primary and the 1 index being the secondary objects in the map. But I will implement the map it is a good idea.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -488,8 +463,8 @@ auto_flash(std::shared_ptr<xrt_core::device> & workingDevice, const std::string&
   // Collect all indexes of boards need updating
   std::vector<std::pair<unsigned int , DSAInfo>> boardsToUpdate;
 
-  std::string image_path;
-  if(image.empty()) {
+  std::string image_path = image;
```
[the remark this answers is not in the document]
dbenusov-xilinx: The DSAInfo object below should use the passed in image,unless the given image is empty. If the image is empty the block below populates image_path.

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
dbenusov-xilinx: @keryell thought the lines were too long. Most of the splits occur on commas or new lines. Plus the entire string is visible on a half screen which is convenient

src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -652,51 +598,56 @@ program_plp(const xrt_core::device* dev, const std::string& partition)
...
-
-    return path;
-  };
-
   for(const auto& img : image_list) {
-    // check if the passed in image is absolute path
-    if(boost::filesystem::exists(boost::filesystem::path(img))) {
+    // Check if the passed in image is absolute path
+    if (boost::filesystem::exists(img)){
+      if (boost::filesystem::extension(img).compare(".xsabin") != 0)
```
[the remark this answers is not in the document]
dbenusov-xilinx: My pleasure!

rozumx (approved):

dayeh-xilinx:
retest this please.

AShivangi (approved):

PR #6106, opened 2022-01-04
CR-1116820 Buffer export handle is not closed

stsoe:
When exporting a buffer object, the export handle (file descriptor
linux) is not closed automatically so the underlying drm buffer is
not freed even when the exporting buffer object is deleted.

CR-1116820 testing of p2p buffer export and import.    Iterating export/import 
while properly releasing allocated buffer objects causes out-of-memory because 
the underlying export fd are not closed.

In this PR, the export handle is managed by the exporting buffer
object implementation and lifetime is tied to that of the buffer
object.  When the buffer object is destructed, the export handle (if
any) is closed.  The changes in this PR apply to both native XRT APIs
and OpenCL.

There are two semantic changes in this PR that should be noted:

First, the export handle of a buffer object is now stored with the
exporting buffer object.  If a buffer object is exported multiple
times, the same export handle (fd) is returned.

Second, the export handle lifetime is tied to the exporting buffer
which cannot be closed until after the export handle is no longer
needed.

The changes were tested per the associated CR.

Documentation probably should reflect the lifetime requirement
mentioned above.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020-2021, Xilinx Inc - All rights reserved
+ * Copyright (C) 2020-2022, Xilinx Inc - All rights reserved
  * Xilinx Runtime (XRT) Experimental APIs
  *
  * Licensed under the Apache License, Version 2.0 (the "License"). You may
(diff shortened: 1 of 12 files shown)


uday610:
@stsoe , if I understood correctly then this PR won't require the host code to do a explicit `close(fd);` .. So before releasing the memory the driver will close it (if that is already opened).. Is that correct understanding? Thanks

stsoe:
> @stsoe , if I understood correctly then this PR won't require the host code to do a explicit `close(fd);` .. So before releasing the memory the driver will close it (if that is already opened).. Is that correct understanding? Thanks

Correct.  No requirement for host code other than being aware of the two changes I mentioned in the description; I don't see these being an issue.

uday610 (approved):

stsoe:
retest this please

dayeh-xilinx:
retest this please

PR #6111, opened 2022-01-05
Fix boost/any.hpp transitive dep

bryanloz-xilinx:
XRT Native APIs are using boost/any.hpp in their public interface. Yet, the CMake infrastructure does not inform consumers that they must use the boost include directory in the project include path.

### DIFF
--- a/src/runtime_src/core/pcie/linux/CMakeLists.txt
+++ b/src/runtime_src/core/pcie/linux/CMakeLists.txt
@@ -55,6 +55,8 @@ target_link_libraries(xrt_core
   rt
   dl
   uuid
+  INTERFACE
+  Boost::boost
   )
 
 # Targets linking with xrt_core_static must also link with additional
@@ -71,6 +73,7 @@ target_link_libraries(xrt_core_static
   dl
   rt
   pthread
+  Boost::boost
   )
 
 install(TARGETS xrt_core
--- a/src/CMake/config/xrt.fp.in
+++ b/src/CMake/config/xrt.fp.in
@@ -45,3 +45,8 @@ set(@PROJECT_NAME@_SWEMU_LIBRARIES @PROJECT_NAME@::xrt_swemu)
 set(@PROJECT_NAME@_HWEMU_LIBRARIES @PROJECT_NAME@::xrt_hwemu)
 
 set(@PROJECT_NAME@_FOUND True)
+
+message(STATUS "Found @PROJECT_NAME@: ${@PROJECT_NAME@_CMAKE_DIR} (found version \"${@PROJECT_NAME@_VERSION}\")")
+
+# XRT is requiring Boost in its interface
+find_package(Boost REQUIRED)
--- a/src/runtime_src/core/common/CMakeLists.txt
+++ b/src/runtime_src/core/common/CMakeLists.txt
@@ -72,7 +72,9 @@ set_target_properties(xrt_coreutil PROPERTIES
 target_link_libraries(xrt_coreutil
   PRIVATE
   ${Boost_FILESYSTEM_LIBRARY}
-  ${Boost_SYSTEM_LIBRARY})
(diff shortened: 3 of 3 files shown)


bryanloz-xilinx:
Ongoing discussion with @stsoe , please don't merge immediately

src/runtime_src/core/pcie/linux/CMakeLists.txt
```diff
@@ -55,6 +55,8 @@ target_link_libraries(xrt_core
   rt
   dl
   uuid
+  INTERFACE
+  Boost::boost
```
stsoe, line 59: Why this change for xrt_core?   I follow the xrt_coreutil because that library is required for xrt::device which compile time depends on boost.  But xrt_core library is not linked directly (usually) to host application and even if it was, it doesn't have client requirements beyond the run-time dependencies that are installed automatically with the rpm/deb.

src/runtime_src/core/pcie/linux/CMakeLists.txt
```diff
@@ -71,6 +73,7 @@ target_link_libraries(xrt_core_static
   dl
   rt
   pthread
+  Boost::boost
```
stsoe, line 76: ditto.

stsoe (commented):
Thanks Bryan for taking a stab at this.  Please check my question,

src/runtime_src/core/pcie/linux/CMakeLists.txt
bryanloz-xilinx: We are linking both of them
```
$ ldd rt-engine/librt-engine.so 
        linux-vdso.so.1 (0x00007ffe1fdfb000)
        libxrt_core.so.2 => /opt/xilinx/xrt/lib/libxrt_core.so.2 (0x00007f8f901da000)
        libxrt_coreutil.so.2 => /opt/xilinx/xrt/lib/libxrt_coreutil.so.2 (0x00007f8f8feab000)
```

I guess I didn't know which library I should apply this to, so I chose both.

bryanloz-xilinx: Also, what about hardware emulation? In that case I'm also not sure which library to apply this to.

stsoe, line 59: Let's remove from libxrt_core.

stsoe: hw_emu is same as xrt_core, it should not need the dependency.

src/runtime_src/core/pcie/linux/CMakeLists.txt
stsoe: I hear you about std::any, and for sure we'd prefer that, unfortunately C++17 is not yet universally supported on all the platforms we need to build against.  But maybe we should consider some conditionally magic, at least that would resolve it for your use-case and we can then punt on the INTERFACE changes?

rozumx:
> Ongoing discussion with @stsoe , please don't merge immediately

Bryan, I have added the DO NOT MERGE tag.  When you are ready to have these changes merged, please remove it.

src/runtime_src/core/pcie/linux/CMakeLists.txt
rozumx: Hmm...If there are C++ 17 issues, shouldn't the XRT-Pipeline catch them for it is my understanding that each supported platform is built.

rozumx (approved):

dayeh-xilinx:
retest this please.

stsoe:
@bryanloz-xilinx I will be closing this PR and adding another one that default switches to c++17 and std::any.

bryanloz-xilinx:
> @bryanloz-xilinx I will be closing this PR and adding another one that default switches to c++17 and std::any.

Sounds good @stsoe

PR #6112, opened 2022-01-06
CR-1116870: Emu driver uses xrt_xclbin.h and xclbin_int.h API's

udaybhas-xilinx:
Emu driver uses xrt_xclbin.h and xclbin_int.h API's for parsing xclbin

https://jira.xilinx.com/browse/CR-1116870
https://jira.xilinx.com/browse/CR-1116752

Low  Risk

Tests: 1. Canary is neat.

### DIFF
--- a/src/runtime_src/core/common/drv/kds_core.c
+++ b/src/runtime_src/core/common/drv/kds_core.c
@@ -157,6 +157,7 @@ static int kds_polling_thread(void *data)
 	int loop_cnt = 0;
 	int cu_idx;
 
+	printk("kds polling thread is running\n");
 	while (!kds->polling_stop) {
 		busy_cnt = 0;
 		for (cu_idx = 0; cu_idx < num_cus; cu_idx++) {
@@ -182,6 +183,7 @@ static int kds_polling_thread(void *data)
 
 		wait_event_interruptible(kds->wait_queue, kds_wake_up_poll(kds));
 	}
+	printk("kds polling thread is stopped\n");
 
 	return 0;
 }
--- a/src/runtime_src/core/pcie/driver/linux/include/xocl_xgq.h
+++ b/src/runtime_src/core/pcie/driver/linux/include/xocl_xgq.h
@@ -20,9 +20,12 @@
 struct xocl_xgq_info {
 	int			 xi_id;
 	u64			 xi_addr;
+	void __iomem		*xi_sq_prod;
 	void __iomem		*xi_sq_prod_int;
+	void __iomem		*xi_cq_prod;
 };
 
+ssize_t xocl_xgq_dump_info(void *xgq_handle, char *buf, int count);
 int xocl_xgq_set_command(void *xgq_handle, int id, u32 *cmd, size_t sz);
 void xocl_xgq_notify(void *xgq_handle);
 int xocl_xgq_get_response(void *xgq_handle, int id);
--- a/src/runtime_src/core/pcie/driver/linux/xocl/devices.h
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/devices.h
@@ -279,6 +279,7 @@ enum {
 #define	XOCL_XGQ_VMR		"xgq_vmr"
 #define	XOCL_HWMON_SDM		"hwmon_sdm"
 #define XOCL_ERT_CTRL           "ert_ctrl"
+#define XOCL_ERT_CTRL_VERSAL    "ert_ctrl.versal"
 
 #define XOCL_DEVNAME(str)	str SUBDEV_SUFFIX
 
@@ -2916,6 +2917,7 @@ struct xocl_subdev_map {
 #define	XOCL_BOARD_VERSAL_USER_RAPTOR2					\
 	(struct xocl_board_private){					\
 		.flags = XOCL_DSAFLAG_DYNAMIC_IP |			\
+			XOCL_DSAFLAG_MB_SCHE_OFF |			\
 			XOCL_DSAFLAG_VERSAL,				\
 		.subdev_info = RES_USER_VERSAL_VSEC,			\
 		.subdev_num = ARRAY_SIZE(RES_USER_VERSAL_VSEC),		\
@@ -2924,6 +2926,7 @@ struct xocl_subdev_map {
 #define	XOCL_BOARD_VERSAL_USER_RAPTOR2_ES3				\
 	(struct xocl_board_private){					\
 		.flags = XOCL_DSAFLAG_DYNAMIC_IP |			\
+			XOCL_DSAFLAG_MB_SCHE_OFF |			\
 			XOCL_DSAFLAG_VERSAL_ES3 |			\
 			XOCL_DSAFLAG_VERSAL,				\
 		.subdev_info = RES_USER_VERSAL_VSEC,			\
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_ctrl.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/ert_ctrl.c
@@ -19,6 +19,7 @@
 #include "xgq_cmd_ert.h"
 #include "../xgq_xocl_plat.h"
 #include "xocl_xgq.h"
+#include "xrt_drv.h"
 
 #define	EC_ERR(ec, fmt, arg...)	\
 	xocl_err(&(ec)->ec_pdev->dev, fmt "\n", ##arg)
@@ -47,7 +48,20 @@
 
 #define CTRL_XGQ_SLOT_SIZE          512	
 
+/* XGQ IP offsets */
+#define XGQ_SQ_REG		0x0
+#define XGQ_CQ_REG		0x100
+
 static uint16_t	g_ctrl_xgq_cid;
+struct xocl_drv_private ert_ctrl_xgq_drv_priv;
+
+struct ert_ctrl_xgq_cu {
+	int			 ecxc_id;
+	resource_size_t		 ecxc_xgq_reg;
+	resource_size_t		 ecxc_xgq_range;
+
+	void __iomem		*ecxc_xgq_base;
+};
 
 struct ert_ctrl {
 	struct kds_ert		 ec_ert;
@@ -60,10 +74,11 @@ struct ert_ctrl {
 	struct xgq		 ec_ctrl_xgq;
 	struct mutex		 ec_xgq_lock;
 
+	struct ert_ctrl_xgq_cu	*ec_xgq_ips;
+	size_t			 ec_num_xgq_ips;
 	/* ERT XGQ instances for CU */
 	void			**ec_exgq;
 	uint32_t		 ec_exgq_capacity;
-
 
 	uint64_t		timestamp;
 	uint32_t		cq_read_single;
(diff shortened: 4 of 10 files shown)


stsoe (changes requested):
I have zero confidence in the implementation of the hal layer function in hw_em.  I don't think this PR should merged.  I think team needs to redesign and rewrite all the hw_em hal code code to conform to modern C++  coding standards and idioms.  I see some errors in this PR, but it is pointless to mention them in the context of all the poor code in the current implementation.

venkatp-xilinx:
> I have zero confidence in the implementation of the hal layer function in hw_em. I don't think this PR should merged. I think team needs to redesign and rewrite all the hw_em hal code code to conform to modern C++ coding standards and idioms. I see some errors in this PR, but it is pointless to mention them in the context of all the poor code in the current implementation.

Hi Soren, thanks for your valuable feedback, we would need to consider the revamp of hw_emu code base as a separate task which would take decent amount of time and verification effort. We will plan it accordingly shortly. With the existing logic and understanding we have, we have revamped the code to an extent possible by making use of the common APIs from the xrt core layer to fix the CRs reported. Considering the scenario, you can help us to fix the errors in the modified code.

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -577,44 +573,33 @@ namespace xclhwemhal2 {
...
-
-      if (++count > 1)
+      // iterate devices
+      for (auto& xml_device : xml_project.get_child("project.platform"))
       {
-        //Give error and return from here
+        if (xml_device.first != "device")
+          continue;
+
+        fpgaDevice = xml_device.second.get<std::string>("<xmlattr>.fpgaDevice");
```
stsoe, line 595: I think fpgaDevice is used to determine if this is versal or not?  If so, that should be abstracted as opposed to doing string compares.   If the XSA name could be used instead (still needs abstractions for versal test), then you could get that from xrt::xclbin if this class object was available here.

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -625,102 +610,73 @@ namespace xclhwemhal2 {
...
-      {
-        //Give error and return from here
-      }
-    }
 
-    std::vector<std::string> kernels;
-
-    // iterate kernels
-    for (auto& xml_kernel : xml_project.get_child("project.platform.device.core"))
+    auto xclbin_object = new xrt::xclbin(args.m_top);
```
stsoe, line 614: This is a bug.  You can't construct on heap and not delete the heap memory when done.  This should not be on the heap to begin with.   Also, this is where some restructuring is called for, because creating the xclbin object shouldn't really be necessary given that xrt_core::device ends up containing / constructing this object also.   That's for a complete redesign, but for now consider moving the xrt::xclbin construction earlier so you can use it for xsa name and if xsa name isn't sufficient and you indeed need the fpgaDevice attribute value, then get that from the xrt::xclbin object through some internal (common/api/xclbin_int.h) method.

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -625,102 +610,73 @@ namespace xclhwemhal2 {
...
+    auto kernels = xclbin_object->get_kernels();
+    std::map<uint64_t, KernelArg> kernelArgInfo;
+    std::string instance_name = "";
+    uint64_t base_address = 0;
+    for (auto kernel : kernels)
```
stsoe, line 619: `const auto&`
or if you don't need `kernels` outside the for loop, then
```
for (const auto& kernel : xclbin_object.get_kernels()) {
 ...
}
```

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -625,102 +610,73 @@ namespace xclhwemhal2 {
...
-      if (mLogStream.is_open())
-         mLogStream << __func__ << " Filling kernel " << kernelName << " info from xclbin.xml" << std::endl;
-
-      for (auto& xml_kernel_info : xml_kernel.second)
+      // get properties of each kernel object
+      auto props = xrt_core::xclbin_int::get_properties(kernel);
+      //get CU's of each kernel object
+      auto cus = kernel.get_cus();
+      //iterate over CU's to get arguments
+      for (auto cu :cus)
```
stsoe, line 626:
```
for (const autu& cu : kernel.get_cus()) {
 ...
}
```

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -625,102 +610,73 @@ namespace xclhwemhal2 {
...
       {
-        std::map<uint64_t, KernelArg> kernelArgInfo;
-        if (xml_kernel_info.first == "arg")
+        auto args = cu.get_args();
```
stsoe, line 628: Move variables to where they are used.  `args` is not used until line 644 and can even be combined with the for-loop there.

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -625,102 +610,73 @@ namespace xclhwemhal2 {
...
-          KernelArg kArg;
-          kArg.name = kernelName + ":" + name;
-          kArg.size = size;
-          kernelArgInfo[offset] = kArg;
-
-          if (mLogStream.is_open())
-            mLogStream << __func__ << " Filling kernel Args name: " << name << " id: " << id << " port: " << port << " info from xclbin.xml" << std::endl;
+          mCuBaseAddress = xclemulation::config::getInstance()->getCuBaseAddrForce();
+        } 
+        else if(mVersalPlatform)
```
stsoe, line 636: Conditional code for platform type doesn't look right, it is definitely not scalable; this should be considered for redesign.

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -625,102 +610,73 @@ namespace xclhwemhal2 {
...
-          std::string mode = xml_kernel_info.second.get<std::string>("<xmlattr>.mode");
-          if (mode == "slave") {
-            address_range = convert(xml_kernel_info.second.get<std::string>("<xmlattr>.range"));
-            if (mLogStream.is_open())
-              mLogStream << __func__ << " Getting the Address Range of mode : " << mode << " info from xclbin.xml" << std::endl;
-          }
+        //fetch instance name
+        instance_name = cu.get_name();
+        //iterate over arguments and populate kernelArg structure
+        for (auto arg : args)
```
stsoe, line 644:
```
for (const auto& arg : cu.get_args()) {
  ...
}
```

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -625,102 +610,73 @@ namespace xclhwemhal2 {
...
-            mKernelOffsetArgsInfoMap[base] = kernelArgInfo;
-            if (xclemulation::config::getInstance()->isMemLogsEnabled())
-            {
-              std::ofstream* controlStream = new std::ofstream;
-              controlStream->open(instanceName + "_control.mem");
-              mOffsetInstanceStreamMap[base] = controlStream;
-            }
-            break;
-          }
+          std::ofstream* controlStream = new std::ofstream;
```
stsoe, line 661: Why is this on the heap?  And it is not deleted, I think.  Don't construct objects on the heap, in fact you should never ever have raw new and delete in code you write.

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -625,102 +610,73 @@ namespace xclhwemhal2 {
...
+          //trim instance_name to get actual instance name
+          size_t position = instance_name.find(":");   
+          std:: string trimmed_instance_name = instance_name.substr(position+1);
```
stsoe, line 664: Why is there a space between `std::` and `string`?   You should use `auto` here and in many other places.

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -625,102 +610,73 @@ namespace xclhwemhal2 {
...
+          controlStream->open(trimmed_instance_name + "_control.mem");
+          mOffsetInstanceStreamMap[base_address] = controlStream;
```
stsoe, line 666: Alright, I see you manage the stream object here, but still it shouldn't be a floating heap object because ownership is not obvious.   If need be, then construct as shared pointer to express shared ownership, or unique_ptr to express exclusive ownership, or non heap object and std::move it on line 666.

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -625,102 +610,73 @@ namespace xclhwemhal2 {
...
         }
-
-        if (address_range != 0 && !kernelName.empty() && !instanceName.empty() ) {
-          std::string kernelInstanceStr = kernelName + ":" + instanceName;
-          mCURangeMap[kernelInstanceStr] = address_range;
+        mKernelOffsetArgsInfoMap[base_address] = kernelArgInfo;
```
stsoe, line 668: This incurs a copy of the complete std::map.    You have two options
(1) construct the map in the mKernelOffsetArgsInfoMap[...] location to begin with and just take a reference to use as an alias when you populate the map.
(2) std::move the map on line 668 to avoid the copy.

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -625,102 +610,73 @@ namespace xclhwemhal2 {
...
+        if (props.address_range != 0 && !props.name.empty() && !instance_name.empty() )
+        {
+          mCURangeMap[instance_name] = props.address_range;
         }
```
stsoe, line 672: Bracing is inconsistent throughout this file.

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -625,102 +610,73 @@ namespace xclhwemhal2 {
...
       }
+
     }
+    std::string xclBinName = xrt_core::xclbin::get_project_name(args.m_top);
```
stsoe, line 676: I think I saw this earlier also?   The functions are much too long to really comprehend anything outside of coding style.

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -1087,7 +1043,7 @@ namespace xclhwemhal2 {
           simMode = launcherArgs.c_str();
 
         if (!file_exists(sim_file))
-          sim_file = "simulate.sh";
+          sim_file = "simulate.sh";
```
stsoe, line 1046: Why this space change?

stsoe (changes requested):
Some coding style comments and bugs.   In general the code (not the changes in this PR) is incomprehensible because is it so poorly designed and functions are much too long.

chvamshi-xilinx:
I Agree with Soren. You can plan these changes in the new code base. I know it could be difficult at this time, but better to start a parallel driver with modern C++ implementation. You can switch to the new driver in next release.

dayeh-xilinx:
retest this please.

PR #6114, opened 2022-01-06
Issue #6027 Update Vitis default paths in scripts

### LINKED ISSUE #6027: Strange XILINX_VITIS default paths
While trying to solve how to compile ERT as an outsider, I ran into some strange configurations found in the source tree:
- around https://github.com/Xilinx/XRT/blob/master/build/build.sh#L227 :
  ```bash
  # we pick microblaze toolchain from Vitis install
  if [[ -z ${XILINX_VITIS:+x} ]]; then
    export XILINX_VITIS=/proj/xbuilds/2019.2_released/installs/lin64/Vitis/2019.2
  fi
  ```
- around https://github.com/Xilinx/XRT/blob/master/build/run.sh#L10 :
  ```bash
  # Set to location of your preferred Vitis install
  vitis=/proj/xbuilds/2021.1_released/installs/lin64/Vitis/2021.1
  ```
  it is *very unlikely* it is always the preferred Vitis install of the reader. :-)

While this sounds like a good idea to peak some default, at the end it might lead to very subtle errors and rapid code churning, so it might be preferable to just remove this random default picks.

Probably https://github.com/Xilinx/XRT/blob/master/src/runtime_src/doc/toc/test.rst would require some update on the versions and tools too.

stsoe:
Update default VITIS paths in build scripts

Issues #6027 raises concerns about magically defaulting Vitis paths in build.sh.

Defaulting the paths will not change, but they are updated to more recent versions and warning is issued if path cannot be found. Removing the paths and relying on internal users managing the setting of XILINX_VITIS is not feasible.  The scripts are primarily for convenience of internal users.

Risk is minimal resulting from path update in build.sh and preservation of XILINX_XRT in run.sh.

This PR closes #6027.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -225,8 +225,14 @@ if [[ ! -z ${XRT_BOOST_INSTALL:+x} ]]; then
 fi
 
 # we pick microblaze toolchain from Vitis install
-if [[ -z ${XILINX_VITIS:+x} ]]; then
-    export XILINX_VITIS=/proj/xbuilds/2019.2_released/installs/lin64/Vitis/2019.2
+if [[ -z ${XILINX_VITIS:+x} ]] || [[ ! -d ${XILINX_VITIS} ]]; then
+    export XILINX_VITIS=/proj/xbuilds/2021.2_released/installs/lin64/Vitis/2021.2
+    if [[ ! -d ${XILINX_VITIS} ]]; then
+        echo "****************************************************************"
+        echo "* XILINX_VITIS is undefined or not accessible                  *"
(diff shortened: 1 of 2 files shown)


rozumx (approved):
Approved

That said, I really don't like it that we have this the default XILINX_VITIS value hard coded for the very reason why this PR is needed.

The next time we have a chance to chat, lets discuss how might (if possible) we can address this issue.

keryell:
This looks like an improvement by removing 1 of the hard-coded version of Vitis.

PR #6120, opened 2022-01-07
Update build_edge.sh script to pick bsps from correct path

rbramand-xilinx:
> Petalinux bsps path are changed because of which pipeline breaks
> Update build_edge.sh to pick bsps from correct path


maxzhen (approved):

maxzhen:
Merging it now since pipeline is broken without this fix.

PR #6122, opened 2022-01-07
Extending HAL/Shim level tracing and device level offload of trace data to Hardware Emulation

jvillarre:
Previously, device-level trace information in hardware emulation was only available to applications written to interface with the OpenCL layer, and xrt_trace was unavailable for hardware emulation completely.  This pull request removes that restriction by rearchitecting the profiling device offload plugins to remove the dependency on the OpenCL abstraction layer and create two distinct plugins for device offload: one for hardware runs and one for hardware emulation runs (software emulation is still not supported).  These plugins are built with and communicate with the appropriate HAL/Shim implementation.

The user controls have not changed, and device_trace or device_counters in the xrt.ini file will now enable the loading of the appropriate plugin based on which flow is being run (emulation or hardware).

Running XRT level applications in hardware emulation lacked any way to get device-level profiling or trace data, regardless of the controls the user specified in the xrt.ini file.  Only OpenCL level applications in hardware emulation provided device level data.

The OpenCLDeviceOffloadPlugin functionality was split into the current "information-gathering only" plugin and the remaining functionality was lowered into a new plugin loaded when hardware emulation is called.  The callbacks to load this plugin are in located the hardware emulation shim for PCIe and also located in the Zocl shim protected by compile time guards.  The hardware flow already had a plugin solution for device offload of profiling/trace data, so that is unchanged.  Since XRT already has a mechanism to prevent multiple shims from being loaded, these plugins are guaranteed to be mutually exclusive.

Hardware emulation designs targeting PCIe and Edge platforms have been tested with the new plugins


chvamshi-xilinx (approved):
Looks Good.

IshitaGhosh (approved):

dayeh-xilinx:
retest this please.

PR #6124, opened 2022-01-10
Add temporary changes in xclCloseExportHandle function of edge platforms to make IVTs pass

rbramand-xilinx:
> xclCloseExportHandle function is added recently to close export buffer handle automatically when buffer is freed.
> But in edge platforms closing of fd is failing with Bad file descriptor error
> Made temporary changes so that IVTs will pass as common Images for versal are quite old and needs to be updated.
> I will revert changes once common Images for versal are updated
> I will also debug why the issue is coming for edge platforms when closing export buffer file descriptor

### DIFF
--- a/src/runtime_src/core/edge/user/shim.cpp
+++ b/src/runtime_src/core/edge/user/shim.cpp
@@ -1766,7 +1766,8 @@ xclImportBO(xclDeviceHandle handle, int fd, unsigned flags)
 int
 xclCloseExportHandle(int fd)
 {
-  return close(fd) ? -errno : 0;
+  //return close(fd) ? -errno : 0;
+  return 0;
 }
 
 static int


stsoe:
Most likely the host code is not properly releasing buffers so that the closing of the fd happens after the device has been closed.  @rbramand-xilinx please keep @stsoe in the loop as what you find when debugging this.   Can you please file a CR so we don't forget to debug?

stsoe (approved):

chvamshi-xilinx:
> Most likely the host code is not properly releasing buffers so that the closing of the fd happens after the device has been closed. @rbramand-xilinx please keep @stsoe in the loop as what you find when debugging this. Can you please file a CR so we don't forget to debug?

Hi Soren, I didnt find any issue in test case. Issue could be in the driver. Will keep you posted.
filed https://jira.xilinx.com/browse/CR-1118948 to track the issue

chvamshi-xilinx (approved):

stsoe:
Merging without waiting for pipeline.

PR #6128, opened 2022-01-11
Fix aie status file writing

hackwa:
Once aie status thread is finished, we do not have access to aie status data.
Fix race condition when starting thread.


pgschuey (approved):
Code looks good, but I'm unclear why we're not using the device ID. For devices where we're requesting AIE status for, will they always start at index 0?

hackwa:
@pgschuey yes device IDs start from 0 and provide a clean way to track devices in database

hackwa:
This is covered by PR 6134

PR #6135, opened 2022-01-11
Adding default value of 3 to MINOR if null from command: MINOR=cat /etc/centos-release | awk -F. '{ print $2 }'

kbhardwa-xilinx:
We are trying to install Centos Stream 8 docker image and while running the xrtdeps.sh script in the container, the MINOR variable value is getting NULL as the /etc/centos-release is CentOS Stream release 8 without a dot.

Adding a default value of 3 if the MINOR value from the command gets null.  This is so that following will executed
```
if [ $MINOR -gt "2" ]; then
  yum config-manager --set-enabled powertools
  yum config-manager --set-enabled appstream
```

None, change only has effect when MINOR is not defined.

### DIFF
--- a/src/runtime_src/tools/scripts/xrtdeps.sh
+++ b/src/runtime_src/tools/scripts/xrtdeps.sh
@@ -462,6 +462,9 @@ prep_centos8()
     echo "Enabling PowerTools and AppStream repo for CentOS8 ..."
     #minor version of CentOs
     MINOR=`cat /etc/centos-release | awk -F. '{ print $2 }'`
+    if [ -z "$MINOR" ]; then 
+        MINOR=3
+    fi
     if [ $MINOR -gt "2" ]; then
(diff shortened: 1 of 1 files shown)


kbhardwa-xilinx:
We are trying to install Centos Stream 8 docker image and while running the xrt-deps.sh script in the container, the MINOR variable value is getting NULL as the /etc/centos-release is CentOS Stream release 8 without a dot.

MINOR=`cat /etc/centos-release | awk -F. '{ print $2 }'`
    if [ $MINOR -gt "2" ]; then
        yum config-manager --set-enabled powertools
        yum config-manager --set-enabled appstream
        
   Adding a default value of **3** if the MINOR value from the command gets null

stsoe (approved):

dayeh-xilinx:
retest this please.

PR #6136, opened 2022-01-11
XRT support for TS2MM for each SLR

IshitaGhosh:
If a design has CUs across different SLRs but only one Ts2MM and buffer is used for offloading trace via memory, then we get very poor timing. Using different Ts2MM and buffer for every SLR used, improves timing. This PR adds the XRT support for multiple TS2MM and processing of separate trace buffers for each memory.


dayeh-xilinx:
retest this please.

jvillarre (approved):

PR #6140, opened 2022-01-12
CR-1118948 xrtSyncBOAIE failed to close export handle

stsoe:
Add try/catch around close of BO export handle to protect
against code having manually closed the handle already.

Changes made in #6106 to tie the lifetime of an export handle to the
exporting BO breaks AIE sync because the export handle is closed
explicitly.

We could have removed the explicit close from the code, but it's
preferable to be resilient against application or implementation code
explicitly closing the export handle, and a simple try/catch was
chosen.

There should be no risks related to this change.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -214,8 +214,15 @@ public:
   virtual
   ~bo_impl()
   {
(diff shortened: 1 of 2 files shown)


chvamshi-xilinx (approved):

dayeh-xilinx:
retest this please.

PR #6142, opened 2022-01-13
fix XGQ attached failed error

mamin506:
This PR is to fix the XGQ attached issue. See #6138 

Recently, we update the Microblaze compiler from 2019.2 to 2021.2 Vitis (#6114 ). But the new XGQ ERT binary doesn't work well. It would not initial the xh_major and xh_minor correctly. We suspect it is related to how 'uint16_t' is handled or optimized by the compiler. But very hard to prove or narrow it down by a small application. 

Use 'uint32_t' xh_version instead and add macros to get major/minor version number.
In this change, if XGQ attach error occurs, xclbin download will fail instead of silently switch to KDS mode.

We still doesn't fully understand why the new compiler generates not working binary.

Test on U50 ssv3 and ssv2 shells.

### DIFF
--- a/src/runtime_src/core/include/xgq_impl.h
+++ b/src/runtime_src/core/include/xgq_impl.h
@@ -136,14 +136,16 @@ static inline uint32_t xgq_double_read32(uint64_t io_hdl, uint64_t addr, int is_
 #define XGQ_MAJOR			1
 #define XGQ_MINOR			0
 #define XGQ_MIN_NUM_SLOTS		2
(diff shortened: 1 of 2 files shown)


maxzhen (approved):

maxzhen:
> Too strange. We should get to the bottom of this somehow. Is there anyway to look at assembler code to see if store or fetch is wrong? Can we check with MicroBlaze team?

We wrote a small main() function and checked the asm code, looks sane. We did not check the ERT code since it has a lot of code so better be checked by an expert. @chienwei-lan will send the question to MB team later.

PR #6156, opened 2022-01-18
Remove boost::any dependency from xrt_device.h

stsoe:
Default building of XRT to use std=c++17 and use std::any in place of
boost::any in the xrt::device exposed get_info() API.  This removes
the need to install boost-dev when using native XRT APIs.

Support backwards compatibility with host code compiled with earlier
versions of XRT.  If host application was compiled against earlier
version of XRT it references boost::any get_info symbols.  These
symbols are retained and old binaries will continue to work with new
XRT.  Also new binaries that are built with -std=c++14 will use
boost::any and require that boost-dev is installed.

The boost::any version of get_info() is hereby deprecated and will
be removed in the future per deprecation policy.

Installing XRT on a bare system doesn't install boost-dev package, 
hence using the native APIs (xrt::device) didn't work out of the box.
This problem was introduced in xrt-2.11.x where xrt::device::get_info
was added.

Switching to -std=c++17 implies we no longer support some older
compilers,.  In particular Ubuntu-16.04 can no longer be used to
build XRT without specifying `./build.sh -cppstd 14 ...`

Backwards compatibility has been checked with tests/xrt/query/  where an
earlier built binary has been verified to work with latest xrt.

@uday610 we should document that c++17 capable compiler is required 
when building XRT.  For now user space code can still build using c++14, but 
will  need to manually install boost-dev.

This PR resolves #6111 which can be closed after this PR is merged.


jvillarre (approved):

src/CMake/nativeWin.cmake
```diff
@@ -33,7 +33,17 @@ find_package(Boost
 set(XRT_BOOST_VERSION ${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION})
```
rozumx, line 33: Line 2 copyright date needs to be updated.

src/CMakeLists.txt
```diff
@@ -19,7 +19,9 @@ include_directories(
   ${CMAKE_BINARY_DIR}/gen
```
rozumx, line 19: Line 2 copyright date needs to be updated.

src/runtime_src/core/common/api/xrt_device.cpp
```diff
@@ -79,21 +79,21 @@ namespace query {
 // that the query request result type matches the xrt::info::device
```
rozumx, line 79: Line 2 copyright date needs to be updated.

src/runtime_src/core/include/xrt/CMakeLists.txt
```diff
@@ -14,7 +14,8 @@ install (FILES ${XRT_XRT_HEADER_SRC} DESTINATION ${XRT_INSTALL_INCLUDE_DIR}/xrt
```
rozumx, line 14: Line 2 copyright date needs to be updated.

src/runtime_src/core/include/xrt/detail/any.h
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2021 Xilinx, Inc. All rights reserved.
```
rozumx, line 2: Copyright date needs to be updated.

src/runtime_src/core/include/xrt/xrt_device.h
```diff
@@ -24,16 +24,16 @@
```
rozumx, line 24: Line 2 copyright date needs to be updated.

tests/validate/CMakeLists.txt
```diff
@@ -3,7 +3,9 @@
 #
```
rozumx, line 3: Line 2 copyright date needs to be updated.

tests/xrt/CMakeLists.txt
```diff
@@ -3,8 +3,8 @@
 #
```
rozumx, line 3: Line 2 copyright date needs to be updated.

rozumx (approved):
Approved.

uday610 (approved):

dayeh-xilinx:
retest this please. Cancelled from queue to give way to PR#6159

chvamshi-xilinx (approved):

PR #6157, opened 2022-01-19
Move lock for register initialisation to speed up register access

ctlange:
Concurrent access to registers and registers access speed up (lower latency).

Massive parallel access of registers. Latency test of registers.

### DIFF
--- a/src/runtime_src/core/pcie/linux/shim.cpp
+++ b/src/runtime_src/core/pcie/linux/shim.cpp
@@ -2201,8 +2201,6 @@ int shim::xclGetDebugProfileDeviceInfo(xclDebugProfileDeviceInfo* info)
 
 int shim::xclRegRW(bool rd, uint32_t ipIndex, uint32_t offset, uint32_t *datap)
 {
-  std::lock_guard<std::mutex> lk(mCuMapLock);
-
   if (ipIndex >= mCuMaps.size()) {
     xrt_logmsg(XRT_ERROR, "%s: invalid CU index: %d", __func__, ipIndex);
     return -EINVAL;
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -2211,20 +2209,23 @@ int shim::xclRegRW(bool rd, uint32_t ipIndex, uint32_t offset, uint32_t *datap)
   auto& cumap = mCuMaps[ipIndex];  // {base, size}
```
stsoe, line 2209: I don't think std::map::operator[] is thread safe against insertion by other threads, so multiple threads can hit this index operator at different times while one thread is busy changing the map.

The only optimization I see we can do in this function is to move 2232-2235 into the initialization block.

stsoe (changes requested):

src/runtime_src/core/pcie/linux/shim.cpp
ctlange: `std::vector<std::pair<uint32_t*, uint32_t>> mCuMaps;`
I think it has not to be thread safe, because mCuMaps is initalised in the constructor with 128 pairs:
`mCuMaps{128, {nullptr, 0}}`
 the vector itself is never resized only the pair values are accessed.

stsoe: Alright, yes it is a vector, but it is not thread safe for same reason.  When one thread is reading the value of mCUMaps[idx] and checking cumap.first, another thread could be in the process of populating the same pair with values.

PR #6162, opened 2022-01-20
CR-1113695 Added support for P2P buffer in XclCopyBO

subeeshm-xilinx:
P2P support for Destination P2P and source Non-P2P is enabled in the xclCopyBO

This is a legacy issue which was caught as part of the CR-1113695

There is no risk

We have run the Canary suite and also verified various combination testcases of P2P manually

No impact on Documentation


chvamshi-xilinx:
ok to test

chvamshi-xilinx (approved):

PR #6173, opened 2022-01-24
add XGQ header size to payload; add error handle in xgq_notify

mamin506:
When Ctrl ERT failed to submit commands, KDS doesn't get error code.

During debug CR-1119857

If submit error happen, xgq_nofity() will add error code to XGQ response.


maxzhen (approved):

dayeh-xilinx:
retest this please. Failed with centos public repo issue.

chienwei-lan (approved):

PR #6178, opened 2022-01-25
CR-1119829: Fixing issue where power profiling file names were misformed

jvillarre:
When power profiling was turned on, the files created had malformed names on CentOS machines.  The intended behavior was for the file name to be in the format of power_profile_<shell>-<unique device id>.csv but instead they were appearing without the <unique device id>.csv portion.

The bug was in how the file name was constructed, and the fix was to force the uint64_t for the unique device id to be translated to a string before being used to construct the file name.


dayeh-xilinx:
retest this please. Jenkins server reboot.

dayeh-xilinx:

pgschuey (approved):

PR #6179, opened 2022-01-25
Report sensor data using sensor API

stsoe:
Fix code to not go via xrt::device::get_info(), which is an end-user API, not a tool API.

Report function is called with a mgmtpf device, which was incorrectly converted to
userpf device because the device id was reopened as an xrt::device object.

Core XRT (implies tools) should use xrt_core::device exclusively.   The core device is opened
explicitly as a mgmtpf device by xbmgmt thus all query calls operate on the mgmtpf device 
expected.   Also, it was wasteful to first convert ptree to string only to read the string back as a ptree, now the code stays with ptree.

Can't think of any

Manual xbmgmt test before and after

### DIFF
--- a/src/runtime_src/core/tools/common/ReportMechanical.cpp
+++ b/src/runtime_src/core/tools/common/ReportMechanical.cpp
@@ -20,9 +20,6 @@
 
 #include "core/common/sensor.h"
 
-// 3rd Party Library - Include Files
-#include <boost/property_tree/json_parser.hpp>
-
 void
 ReportMechanical::getPropertyTreeInternal( const xrt_core::device * _pDevice,
                                            boost::property_tree::ptree &_pt) const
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/tools/common/ReportMechanical.cpp
```diff
@@ -18,35 +18,31 @@
 // Local - Include Files
 #include "ReportMechanical.h"
 
+#include "core/common/sensor.h"
+
```
AShivangi, line 22: I don't think we need json_parser.hpp anymore

AShivangi (commented):
Thanks for making this change! Should we make this change for all the reports?

src/runtime_src/core/tools/common/ReportMechanical.cpp
stsoe: I removed it.  Thanks!

dayeh-xilinx:
retest this please. Jenkins server reboot.

dayeh-xilinx:

chvamshi-xilinx (approved):
This looks simpler . Are you making this change in all reports?

PR #6182, opened 2022-01-26
CR-1118968 Disable PDI reload on warm reboot

xdavidz:
Fix the warm reboot hung issue by only setting pmc registers during hot reset.
For the case where system is already hung, we continue to set the registers on the host, then issue a second bus reset. 

CR-1118968 Disable PDI reload on warm reboot

There are multiple cases:
1) warm reboot, since we don't see the pmc register, no issue.
2) hot reset, we config vmr then set pmc register, no issue.
3) vmr hung, we config vmr and will fail(xgq service stopped), continue to set pmc register, no issue.
4) switch to B boot, we config vmr, then set pmc regiseter, as long as vmr is working, no issue.
   if vmr is not working, this is the same case as No.3. After system is back from 3, we repeat this step.

should re-test hot reset 1000+ times and see how reliable of this new solution.

tested with latest shell


dayeh-xilinx:

houlz0507 (approved):

maxzhen (approved):

PR #6184, opened 2022-01-27
CR-1116717 ASTER TC 18.01 - xbutil dmatest: "ERROR: DMA failed: No such file or directory" when running test for 256MB

chienwei-lan:
The issue is calling xclSyncBo with non-exist BO handle. In this case, we have only one BO and two DMA threads, adjust_e will be b+2 which is not true and the number should be b+1 since we only have one BO, choose the smaller number between BO list size and DMA threads

#4971

### DIFF
--- a/src/runtime_src/core/pcie/common/dmatest.h
+++ b/src/runtime_src/core/pcie/common/dmatest.h
@@ -85,8 +85,8 @@ namespace xcldev {
                 return future0.get();
             }
 
-            auto len = ((e - b) < count) ? 1 : (e - b)/count;
-            const auto ajust_e = b + len * count;
+            unsigned len = ((e - b) < count) ? 1 : (e - b)/count;
+            const auto ajust_e = b + len * std::min(count, len);
             std::vector<std::future<int>> threads;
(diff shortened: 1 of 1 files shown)


xuhz (approved):

chienwei-lan:
retest this please

src/runtime_src/core/pcie/common/dmatest.h
```diff
@@ -86,7 +86,7 @@ namespace xcldev {
             }
 
             auto len = ((e - b) < count) ? 1 : (e - b)/count;
-            const auto ajust_e = b + len * count;
+            const auto ajust_e = b + len * std::min(count, (unsigned)len);
```
stsoe, line 89:
```suggestion
            const auto ajust_e = b + len * std::min<unsigned int>(count, len);
```

stsoe (changes requested):

stsoe (approved):

dayeh-xilinx:

sabarinaxil:
retest this please

xilinxgitops:
retest this please

PR #6185, opened 2022-01-27
Use xrt::bo in xdp::HalDevice

IshitaGhosh:
Improves implementation of xdp::HalDevice by using reference counted xrt::bo s in place of C APIs. Also, xrt::bo s works well in both PCIe and Edge flows. So, use of xcl*BO APIs for Edge has been removed.


stsoe (approved):
Much simplified. Nice!

jvillarre (approved):

PR #6187, opened 2022-01-27
VITIS-3288 Cleanup XRT tests to use CMake only build

eshwar-xilinx:

  Support is added to test cases present in XRT/tests/xrt XRT/tests/unit_test to use CMake to build host code and xclbin files.
  Each test case can be build independently. By default, xclbin is not generated.

  Support is added to the CMakeLists.txt files to build xclbin using kernel src
 xrt_create_xo() and xrt_create_xclbin() macro is included in utils.cmake for generating xclbin. Sample usage is as follows :
```
xrt_create_xo(
  "path/to/kernel.cl"                         #path to kernel source file
  "--hls.pre_tcl path/to/tcl/file -k krnl"    #any extra compile options required
  "kernel"                                    #name of xo file to be generated
  ) 

xrt_create_xclbin(         
  "kernel"                                    #output file name
  "--config path/to/vitis_link.cfg"           #extra_link_options required for generating xclbin
  )
```
  By default, xclbin generation will not occur. `-xclbin` option should be included when running build.sh for xclbin creation. e.g.: `./build.sh -em sw_emu -xclbin`
  Dependency of xclbin on .xo files was added.

  All the variables, such as `CMAKE_CXX_STANDARD` , `XRT libraries`, find_package(v++) are set in utils.cmake. This `utils.cmake` file is included in every CMakeLists.txt file so that each test case can be build independently.

  One of the file, `xcl.cpp`, in the test case `tests/xrt/22_verify` uses `std::memset` which is a member of `cstring`. Hence, `cstring` header file is added to `22_verify/xcl.cpp`
  Renamed `22_verify/hello.cl`, `102_multiproc_verify/hello.cl` to `22_verify/kernel.cl`, `102_multiproc_verify/kernel.cl`

  No Risk

  Manually tested by building the testcases for `sw_emu` and `hw_emu`

  No


chvamshi-xilinx (changes requested):
Please look at the feedback in my prev comment.

chvamshi-xilinx (changes requested):
As discussed offline, Please make it simpler by creating all variables in common utility

tests/xrt/CMakeLists.txt
```diff
@@ -2,15 +2,23 @@
...
 set(XRT_CORE_LIBRARY xrt_core)
-set(MODE hw)
 
-set(XRT_DIR ${XILINX_XRT}/share/cmake/XRT)
+if(NOT MODE)
+  set(MODE sw_emu)
+endif()
+
+if(NOT PLATFORM)
+  set(PLATFORM /proj/xbuilds/2022.1_daily_latest/internal_platforms/xilinx_u200_gen3x16_xdma_1_202110_1/xilinx_u200_gen3x16_xdma_1_202110_1.xpfm)
```
chvamshi-xilinx, line 18: i think this shouldn't be required.

tests/xrt/CMakeLists.txt
```diff
@@ -2,15 +2,23 @@
...
+endif()
+
+set(XRT_DIR XILINX_XRT/share/cmake/XRT)
```
chvamshi-xilinx, line 21: what is this? why r we pointing share/cmake/XRT?

chvamshi-xilinx (approved):
Looks very good now. just couple of questions..

stsoe (changes requested):
I can't deal with camelCase especially when used where the convention is under_score.
I suggest renaming all internal macros to 
xrt_create_xo(), xrt_create_xclbin(), xrt_create_emconfig()
Please replace XOs with XOS.

However, please consider a single function to build XO and XCLBIN such that all test cases CMake files only provide the kernel source files, the xclbin name, and options when different from default.   This would simply each test dir makefile.

How will the xclbin get built?  If by default, then please provide an option to not build xclbin.

dayeh-xilinx:

dayeh-xilinx:

dayeh-xilinx:

tests/xrt/utils.cmake
```diff
@@ -0,0 +1,144 @@
...
+# OUT             Name of the xclbin to be generated 
+# KERNEL_SRC      List of Kernel source files required for generating xclbin
+macro(xrt_create_xclbin)
+  
+  # cmake_parse_arguments general structure is :
+  # cmake_parse_arguments(SUFFIX "${options}" "${oneValueArg}" "${multipleValueArg}" ..)
+  # Here, OUT and LINK_OPTIONS can have only one arg
+  # KERNEL_SRC can have multiple arguments. This is useful in testcases where there are multiple src files.
+  set(one_value_arg OUT LINK_OPTIONS)
+  cmake_parse_arguments(XCLBIN "" "${one_value_arg}" KERNEL_SRC "${ARGN}")
```
chvamshi-xilinx, line 106: This looks pretty complex. Instead of this i would suggest you to go back and have create_xo for each kernel and create_xclbin for xclbin creation.

eshwar-xilinx:
Hi @stsoe 
XCLBIN creation is made not mandatory.  However, a option is added to build.sh script for its generation. camelCase naming convention is replaced by xrt_create_xclbin() macros. Implementation of xrt_create_xclbin() is present in utils.cmake. Please suggest changes, if any, on its complexity.

tests/xrt/utils.cmake
chvamshi-xilinx: cc: @stsoe , Please suggest. It is better to have two functions instead of this complex parsing in one function to figure out which one for compilation and which one is for linking etc.
create_xo(kernel_src, extra_comp_args)
create_xclbin(kernel_xos, extra_link_args)

chvamshi-xilinx (approved):
Looks good. Please change the copyright in all files.

xilinxgitops:
retest this please.

kbhardwa-xilinx:
retest this please

salindac:
retest this please

dayeh-xilinx:
retest this please.

tests/CMake/utils.cmake
```diff
@@ -0,0 +1,123 @@
...
+
+if(NOT VITIS_COMPILER)
+  find_program(VITIS_COMPILER 
+  NAMES v++
+  HINTS XILINX_VITIS}
+  )
+  if(VITIS_COMPILER)
+    message("Vitis Compiler is found at ${VITIS_COMPILER}")
+  else()
+   message(FATAL_ERROR "Vitis Compiler v++ is not found, Please source vitis and try again")
```
stsoe, line 20: No finding Vitis should not be a fatal error.   It should at most be a warning saying that xclbin cannot be built without a Vitis installation available

stsoe (changes requested):
Thank you.  I think the xclbin creation is a little bit awkward, but I don't have a any better suggestion other than not making xclbin creation part of CMake.  I am okay with what is done here.

One change requested in utils.mk.

tests/CMake/utils.cmake
eshwar-xilinx: Hi @stsoe ,
A warning might lead to confusion . So the message (if v++ not found ) is removed all together.

stsoe (approved):
Looks good.  Thank you for these changes.

xilinxgitops:
retest this please.

kbhardwa-xilinx:
retest this please

PR #6191, opened 2022-01-27
VITIS 3208: xbmgmt tool bringup with SDR support

rajkumar-xilinx:
Present model:
 All the sensors values are being read from XMC subdev driver. Each sensor is
 having a sysfs node under this device and the same sensor is  being registered
 with HWMON driver. So, single sensor is having two sysfs nodes at different
 locations. xbutil & xbmgmt tool accessing the sensor values using XMC driver.

Proposed SDR data driven model (Enable only on VMR supported platforms):
 All the sensors are being registered with HWMON driver from hwmon_sdm driver. There is
 NO duplication of creating sensor sysfs nodes again in this hwmon_sdm subdev driver.

Solution: Adding changes in xrt common & xrt tools level to access these sensors directly
 from HWMON sysfs directory. There is no hard coding of sensor names in this approach.
 It is scalable and data driven model.


rajkumar-xilinx:
Validated the change on U50 setup due to versal setup issues. All the result comparison between old model and data driven model are captured at https://confluence.xilinx.com/display/XSW/XRT+Support+ASDM+Sensor+Data+Model

src/runtime_src/core/common/sensor.cpp
```diff
@@ -211,15 +215,96 @@ read_electrical(const xrt_core::device * device)
...
+        pt.put("average", xrt_core::utils::format_base10_shiftdown3(tmp.average));
+        pt.put("highest", xrt_core::utils::format_base10_shiftdown3(tmp.highest));
+        pt.put("current.is_present", true);
+        pt.put("voltage.is_present", false);
+        pt.put("power.is_present", false);
+        sensor_array.push_back(std::make_pair("", pt));
+      }
+      for(size_t i = 0; i < voltage.size(); i++)
+      {
+        auto tmp = voltage[i];
```
stsoe, line 256: ditto

src/runtime_src/core/common/sensor.cpp
```diff
@@ -211,15 +215,96 @@ read_electrical(const xrt_core::device * device)
...
+        sensor_array.push_back(std::make_pair("", pt));
+      }
+      for(size_t i = 0; i < power.size(); i++)
+      {
+        auto tmp = power[i];
+        pt.put("label", tmp.label);
+        pt.put("instantaneous", xrt_core::utils::format_base10_shiftdown6(tmp.input));
+        pt.put("max", xrt_core::utils::format_base10_shiftdown6(tmp.max));
+        pt.put("average", xrt_core::utils::format_base10_shiftdown6(tmp.average));
+        pt.put("highest", xrt_core::utils::format_base10_shiftdown6(tmp.highest));
```
stsoe, line 274: ditto

src/runtime_src/core/common/sensor.cpp
```diff
@@ -211,15 +215,96 @@ read_electrical(const xrt_core::device * device)
...
+    }
+  }
+  catch(const xrt_core::query::no_such_key&) {
+    is_data_driven = false;
+  }
+  catch (const std::exception& ex) {
+    sensor_array.push_back(std::make_pair("", pt.put("error_msg", ex.what())));
+  }
+
+  if (!is_data_driven)
```
AShivangi, line 295: I believe we should have a 3rd function `get_thermals()` which calls `read_legacy_current()` and `read_new_current()`
in get_thermals() -> 
if the sensors are data driven
  `read_new_current()`
else
  `read_legacu_current()`

src/runtime_src/core/common/sensor.cpp
```diff
@@ -211,15 +215,96 @@ read_electrical(const xrt_core::device * device)
...
+    return read_legacy_electrical(device);
```
stsoe, line 296: How about just return this from the first catch?

src/runtime_src/core/common/sensor.cpp
```diff
@@ -236,16 +321,97 @@ read_thermals(const xrt_core::device * device)
...
+ptree_type
+read_thermals(const xrt_core::device * device)
+{
+  ptree_type root;
+  ptree_type thermal_array;
+  ptree_type pt;
+  bool is_data_driven = true;
+
+  try {
+    auto output  = xrt_core::device_query<xq::sdm_sensor_info>(device, (int)THERMAL_REQ);
```
stsoe, line 336: Don't use int there, use the emum class type.

src/runtime_src/core/common/sensor.cpp
```diff
@@ -236,16 +321,97 @@ read_thermals(const xrt_core::device * device)
...
+  }
+  catch(const xrt_core::query::no_such_key&) {
+    is_data_driven = false;
+  }
+  catch (const std::exception& ex) {
+    thermal_array.push_back(std::make_pair("", pt.put("error_msg", ex.what())));
+  }
+
+  if (!is_data_driven)
+    thermal_array = read_legacy_thermals(device);
```
stsoe, line 361: Do this in the first catch

src/runtime_src/core/common/sensor.cpp
```diff
@@ -236,16 +321,97 @@ read_thermals(const xrt_core::device * device)
...
 read_mechanical(const xrt_core::device * device)
 {
   ptree_type root;
   ptree_type fan_array;
-  fan_array.push_back(std::make_pair("", populate_fan(device, "fpga_fan_1", "FPGA Fan 1")));
+  ptree_type pt;
+  bool is_data_driven = true;
+
+  try {
+    auto output  = xrt_core::device_query<xq::sdm_sensor_info>(device, (int)MECHANICAL_REQ);
```
stsoe, line 389: enum class type

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -128,6 +129,120 @@ struct bdf
...
 };
 
+struct sdm_sensor_info
+{
+  using result_type = query::sdm_sensor_info::result_type;
+  using data_type = query::sdm_sensor_info::data_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type, const boost::any& reqType)
+  {
```
AShivangi, line 139: please add some comments for future developer :)

src/runtime_src/core/common/sensor.cpp
```diff
@@ -236,16 +321,97 @@ read_thermals(const xrt_core::device * device)
...
+  }
+  catch(const xrt_core::query::no_such_key&) {
+    is_data_driven = false;
+  }
+  catch (const std::exception& ex) {
+    fan_array.push_back(std::make_pair("", pt.put("error_msg", ex.what())));
+  }
+
+  if (!is_data_driven)
+    fan_array = read_legacy_mechanical(device);
```
stsoe, line 414: do in first catch

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -128,6 +129,120 @@ struct bdf
...
+    std::string parent_path = pdev->get_sysfs_path("", "hwmon");
+    result_type output;
+    std::string errmsg;
+    std::string path;
+    bool found = false;
+
+    boost::filesystem::path render_dirs(parent_path);
+    if (boost::filesystem::is_directory(render_dirs)) {
+      boost::filesystem::recursive_directory_iterator end_iter;
+      for(boost::filesystem::recursive_directory_iterator iter(render_dirs); iter != end_iter; ++iter) {
```
stsoe, line 151: space missing between for and '('

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -128,6 +129,120 @@ struct bdf
...
+        std::string name;
```
stsoe, line 152: move closer to where used, .e.g. right before line 156

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -128,6 +129,120 @@ struct bdf
...
+  using result_type = query::sdm_sensor_info::result_type;
+  using data_type = query::sdm_sensor_info::data_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type, const boost::any& reqType)
+  {
+    const auto req_type = boost::any_cast<query::sdm_sensor_info::req_type>(reqType);
+    auto pdev = get_pcidev(device);
+    std::string parent_path = pdev->get_sysfs_path("", "hwmon");
+    result_type output;
```
stsoe, line 143: not used

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -128,6 +129,120 @@ struct bdf
...
+              break;
+            }
+          }
+        }
+      }
+    }
+
+    if (!found) {
+      return output;
+    }
```
stsoe, line 170: No braces around single statement conditional expressions.   Also this is first place `output` is used, so just return result_type{};

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -128,6 +129,120 @@ struct bdf
...
+    const auto req_type = boost::any_cast<query::sdm_sensor_info::req_type>(reqType);
+    auto pdev = get_pcidev(device);
+    std::string parent_path = pdev->get_sysfs_path("", "hwmon");
+    result_type output;
+    std::string errmsg;
+    std::string path;
+    bool found = false;
+
+    boost::filesystem::path render_dirs(parent_path);
+    if (boost::filesystem::is_directory(render_dirs)) {
```
AShivangi, line 149: Please use early exits. For eg:
```
if (!boost::filesystem::is_directory(render_dirs)) {
  return output
}
for(...) {
    if(!errmsg.empty())
        return;
```

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -128,6 +129,120 @@ struct bdf
...
+          }
+        }
+      }
+    }
+
+    if (!found) {
+      return output;
+    }
+
+    int input;
```
AShivangi, line 172:
```suggestion
    int input = 0;
```

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -128,6 +129,120 @@ struct bdf
   }
 };
 
+struct sdm_sensor_info
+{
+  using result_type = query::sdm_sensor_info::result_type;
+  using data_type = query::sdm_sensor_info::data_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type, const boost::any& reqType)
```
stsoe, line 138: Refactor this function please. No functions can be more that 10+ lines.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -128,6 +129,120 @@ struct bdf
...
+          if (errmsg.empty()) {
+            if (boost::algorithm::contains(name, "xilinx_vck5000_gen4x8_xdma_base_1"))
+            {
+              found = true;
+              break;
+            }
+          }
+        }
+      }
+    }
```
stsoe, line 166: Must too complex and convoluted.  Reduce the levels by refactoring.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -128,6 +129,120 @@ struct bdf
...
+
+    int input;
+    std::string label;
+    data_type data;
+
+    std::string start[5] = {"curr", "in", "power", "temp", "fan"};
+    std::string end[5] = {"label", "input", "max", "average", "highest"};
+    std::string underscore = "_";
+    std::string slash = "/";
```
AShivangi, line 179: const

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -128,6 +129,120 @@ struct bdf
...
+      case POWER_REQ:
+      case THERMAL_REQ: //thermal means temp readings
+      case MECHANICAL_REQ: //mechanical means fan readings
+        target_sysfs.push_back(req_type);
+        break;
+      case ELECTRICAL_REQ: //electrical combines curr + volt + power readings
+        target_sysfs.push_back(0);
+        target_sysfs.push_back(1);
+        target_sysfs.push_back(2);
+        break;
```
stsoe, line 195: Use the proper enum class enumerators here.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -128,6 +129,120 @@ struct bdf
...
+    }
+
+    for (size_t i = 0; i < target_sysfs.size(); i++) {
+      sid = target_sysfs[i];
+      bool next_id = false;
+      int start_id = 1;
+      int end_id = 0;
+      std::string type = start[sid];
+
+      if (!strcmp(type.c_str(), "in"))
```
AShivangi, line 205: `compare`

src/runtime_src/core/tools/common/ReportElectrical.cpp
```diff
@@ -57,22 +57,38 @@ ReportElectrical::writeReport( const xrt_core::device* /*_pDevice*/,
...
-    bool volts_is_present = pt_sensor.get<bool>("voltage.is_present");
-    std::string volts = pt_sensor.get<std::string>("voltage.volts");
-    bool amps_is_present = pt_sensor.get<bool>("current.is_present");
-    std::string amps = pt_sensor.get<std::string>("current.amps");
+    if(pt_sensor.get<bool>("data_driven", false)) {
+      std::string name = pt_sensor.get<std::string>("label");
+      auto val = pt_sensor.get<std::string>("instantaneous");
+      bool volts_is_present = pt_sensor.get<bool>("voltage.is_present", false);
+      bool amps_is_present = pt_sensor.get<bool>("current.is_present", false);
+      bool power_is_present = pt_sensor.get<bool>("power.is_present", false);
```
stsoe, line 68: auto, don't repeat the type.

stsoe (changes requested):

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -128,6 +129,120 @@ struct bdf
...
+	    }
+          }
+          end_id++;
+        } //while(end_id)
+        if (strcmp((data.label).c_str(), "N/A"))
+          output.push_back(data);
+        start_id++;
+      } //while (!next_id)
+   } //for()
+    return output;
```
AShivangi, line 242: I believe this function should be refactored into smaller functions

AShivangi (changes requested):

rozumx (changes requested):
I agree will all of Soren's and Shivangi's comments.  

When those code changes have been completed, please request me to re-review them.

src/runtime_src/core/common/sensor.cpp
```diff
@@ -134,12 +134,16 @@ populate_fan(const xrt_core::device * device, const std::string& loc_id, const s
...
 ptree_type
-read_electrical(const xrt_core::device * device)
+read_legacy_electrical(const xrt_core::device * device)
 {
   ptree_type root;
   ptree_type sensor_array;
+  ptree_type pt;
+
+  sensor_array.push_back(std::make_pair("", pt.put("data_driven", false)));
+  sensor_array.push_back(std::make_pair("", pt.put("is_present", "true")));
```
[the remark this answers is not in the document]
rajkumar-xilinx: Added check in ReportElectrical file if the data is present or not.

dayeh-xilinx:
retest this please. re-enabled PR branch.

src/runtime_src/core/common/sensor.cpp
rajkumar-xilinx: I have added new functions for data driver model as "read_data_driven_current" etc. as per the suggestion.

src/runtime_src/core/common/sensor.cpp
rajkumar-xilinx: it may miss case where ew sdr model returns empty output then code try to access legacy

src/runtime_src/core/common/sensor.cpp
rajkumar-xilinx: empty output case might be missed.

src/runtime_src/core/common/sensor.cpp
rajkumar-xilinx: empty output case might be missed.

src/runtime_src/core/common/query_requests.h
```diff
@@ -820,6 +831,24 @@ struct debug_ip_layout_raw : request
   get(const device*) const = 0;
 };
 
+struct sdm_sensor_info : request
+{
+  struct data {
+    std::string label;
+    int input;
```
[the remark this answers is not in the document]
rajkumar-xilinx: As per sysfs it is input, but for readability I am using instantaneous while storing this input value in property tree

src/runtime_src/core/common/sensor.cpp
```diff
@@ -211,15 +215,96 @@ read_electrical(const xrt_core::device * device)
...
+      {
+        auto tmp = current[i];
+        pt.put("label", tmp.label);
+        pt.put("instantaneous", xrt_core::utils::format_base10_shiftdown3(tmp.input));
+        pt.put("max", xrt_core::utils::format_base10_shiftdown3(tmp.max));
+        pt.put("average", xrt_core::utils::format_base10_shiftdown3(tmp.average));
+        pt.put("highest", xrt_core::utils::format_base10_shiftdown3(tmp.highest));
+        pt.put("current.is_present", true);
+        pt.put("voltage.is_present", false);
+        pt.put("power.is_present", false);
```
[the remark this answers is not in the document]
rajkumar-xilinx: These are used in ReportElectrical to distinguish between voltage, power, current readings so that value extensions like "Amps", "Watts" etc added there. Please check this file once.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -73,6 +74,97 @@ get_pcidev(const xrt_core::device* device)
...
+                const query::sdr_req_type req_type,
+                const std::string path)
+{
+    auto pdev = get_pcidev(device);
+    query::sdm_sensor_info::result_type output;
+    query::sdm_sensor_info::data_type data;
+    std::vector<int> target_sysfs;
+    //All device sensors are stored in hwmon sysfs nodes with starts & ends as follows.
+    std::vector<std::string> start {"curr", "in", "power", "temp", "fan"};
+    std::vector<std::string> end {"label", "input", "max", "average", "highest"};
```
keryell, line 88: You do not need so many dynamic allocations. Here everything is constant and local.
```suggestion
    std::array start {"curr", "in", "power", "temp", "fan"};
    std::array end {"label", "input", "max", "average", "highest"};
```
guessing we are up to C++17 now.

keryell (changes requested):
Interesting!

dayeh-xilinx:
retest this please. Infra error. gitenterprise.xilinx.com went down for maintainance.

src/runtime_src/core/common/query_requests.h
```diff
@@ -820,6 +836,24 @@ struct debug_ip_layout_raw : request
...
 };
 
+struct sdm_sensor_info : request
+{
+  struct sensor_data {
+    std::string label;
+    uint32_t input;
+    uint32_t max;
+    uint32_t average;
+    uint32_t highest;
```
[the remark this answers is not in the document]
stsoe: >  are you using Doxygen in the project?
This is internal header.  Comments should be added but not for doxygen.  We use doxygen on exported public headers.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -73,6 +74,97 @@ get_pcidev(const xrt_core::device* device)
...
+            std::string label;
+	    pdev->sysfs_get("", path + tmp + end[end_id], errmsg, label);
+	    if (errmsg.empty()) {
+                  data.label = label;
+	    } else {
+		  //go and read next sysfs node
+                  next_id = true;
+		  end_id = max_end_types;
+	    }
+          } else {
```
stsoe, line 148: Use early-exit or refactor, I cannot comprehend these many levels.
Please also use the bracing style we embrace (pun intended):

```
if (cond1)
  stmt1;

if (cond2) {
  stmt1;
  stmt2;
}
```

And use proper indentation following what the file is using.

src/runtime_src/core/pcie/linux/device_linux.cpp
stsoe: Please add comments.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -128,6 +220,57 @@ struct bdf
...
+  using data_type = query::sdm_sensor_info::data_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type, const boost::any& reqType)
+  {
+    const query::sdr_req_type req_type = boost::any_cast<query::sdm_sensor_info::req_type>(reqType);
+    auto pdev = get_pcidev(device);
+    const std::string target_dir = "hwmon";
+    const std::string target_file = "name";
+    const std::string target_platform = "xilinx_vck5000_gen4x8_xdma_base_1";
```
stsoe, line 235: We should not have  platform names encoded here.  Can you find some other way?

stsoe, line 235: We should not have  platform names encoded here.  Can you find some other way?  Like can't this check (whatever the reason is for the conditional code path) be made data driven?  You see the code is not documented to the point where it is clear what is being done.    Why is there a check on targeted platform, it's a mystery!!!

stsoe (changes requested):
More requested changes.

salindac:
retest please

src/runtime_src/core/pcie/linux/device_linux.cpp
rajkumar-xilinx: I fixed indentation. I didn't get other suggestion related to bracing style. Copying the if else code below. 
Do I need to remove brances for if condition since it has only 1 line of code? 

if (errmsg.empty()) {
   data.label = label;
} else {
  //go and read next sysfs node
  next_id = true;
  end_id = max_end_types;
}

rajkumar-xilinx:
retest this please

dayeh-xilinx:
retest this please. Fixed the centos8 repo issue.

chvamshi-xilinx (approved):
looks good. Just couple of minor comments.

rajkumar-xilinx:
Hi all, could you please review and let me know, thank you.

src/runtime_src/core/common/query_requests.h
```diff
@@ -820,6 +821,48 @@ struct debug_ip_layout_raw : request
...
+   * struct sensor_data: used to store sensor information and
+   * each sensor contains following information.
+   *  label    : name
+   *  input    : instantaneous value
+   *  max      : maximum value
+   *  average  : average value
+   *  highest  : highest value (used for temperature sensors)
+   */
+  struct sensor_data {
+    std::string label;
```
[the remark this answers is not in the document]
keryell: Are you thinking to this `std::string`? This is C++, so it is initialized just right here already:
https://en.cppreference.com/w/cpp/string/basic_string/basic_string
> 1) Default constructor. Constructs empty string (zero size and unspecified capacity). If no allocator is supplied, allocator is obtained from a default-constructed instance.

src/runtime_src/core/pcie/linux/device_linux.cpp
stsoe: It's fine, but `else` should line up with `if`.

rajkumar-xilinx: ok, so the code will look like below. Shall I change to below style? 

```
          if (errmsg.empty()) {
            data.label = label;
          }
          else {
            //go and read next sysfs node
            next_id = true;
            end_id = max_end_types;
          }
```

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -73,6 +74,97 @@ get_pcidev(const xrt_core::device* device)
...
+          }
+        } else {
+          std::string errmsg;
+          uint32_t input = 0;
+          pdev->sysfs_get<uint32_t>("", path + tmp + end[end_id], errmsg, input, EINVAL);
+          if (errmsg.empty()) {
+            if (end_id == 1) data.input = input;
+            else if (end_id == 2) data.max = input;
+            else if (end_id == 3) data.average = input;
+            else data.highest = input;
```
stsoe, line 156: New line please, else debugging becomes nightmarish.
```
if (...)
  blah;
else if (...)
  blah;
else if (...)
  blah;
else
  blah;
```

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -73,6 +74,97 @@ get_pcidev(const xrt_core::device* device)
   return pcidev::get_dev(device->get_device_id(), device->is_userpf());
 }
 
+static query::sdm_sensor_info::result_type
+get_sdm_sensors(const xrt_core::device* device,
+                const query::sdm_sensor_info::sdr_req_type req_type,
+                const std::string path)
```
stsoe, line 80: You forget the references on the last two arguments, don't pass-by-value unless you need the copied value.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -73,6 +74,97 @@ get_pcidev(const xrt_core::device* device)
...
+          std::string label;
+          pdev->sysfs_get("", path + tmp + end[end_id], errmsg, label);
+          if (errmsg.empty()) {
+            data.label = label;
+          } else {
+            //go and read next sysfs node
+            next_id = true;
+            end_id = max_end_types;
+          }
+        } else {
```
stsoe, line 148: Align `else` with `if`

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -73,6 +74,97 @@ get_pcidev(const xrt_core::device* device)
   return pcidev::get_dev(device->get_device_id(), device->is_userpf());
 }
 
+static query::sdm_sensor_info::result_type
+get_sdm_sensors(const xrt_core::device* device,
```
stsoe, line 78: Move this function inside `struct sdm_sensor_info `you have below.
Refactor the function to make it simpler and easier to comprehend; keep the refactored (sub-functions) within the struct.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -73,6 +74,97 @@ get_pcidev(const xrt_core::device* device)
...
+    case query::sdm_sensor_info::sdr_req_type::electrical:
+      //electrical combines curr + volt + power readings
+      target_sysfs.push_back(0);
+      target_sysfs.push_back(1);
+      target_sysfs.push_back(2);
+      break;
+  }
+
+  //Iterate over hwmon sysfs nodes based on sensor request type.
+  for (const auto& sid : target_sysfs) {
```
stsoe, line 108: As an example of refactoring, maybe have a function that processes a single target_sysfs.

stsoe (changes requested):

src/runtime_src/core/common/query_requests.h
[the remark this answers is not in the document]
keryell: Ah, yes. :-)
Perhaps just default `uint32_t input {};`

AShivangi:
@rajkumar-xilinx, what does `xbutil examine -r electrical` output look like? I ask as we have a table view where we associate 2 values, i.e., voltage and current, with one sensor

rajkumar-xilinx:
> @rajkumar-xilinx, what does `xbutil examine -r electrical` output look like? I ask as we have a table view where we associate 2 values, i.e., voltage and current, with one sensor

image

rajkumar-xilinx:
Hi all, could you please review and let me know, thank you.

stsoe (approved):
Thanks for addressing so many requests for changes.  I still think some refactoring in device_linux could be done, but it isn't a show stopper.

dayeh-xilinx:

dayeh-xilinx:

dayeh-xilinx:

rajkumar-xilinx:
Hi @AShivangi @rozumx could you please review, thank you.

AShivangi:
Hi @rajkumar-xilinx, 

I have a few concerns:
1. The json has been changed drastically, can you please provide the json output as well so we can make sure that we haven't changed any of the existing node names. Otherwise, we will be breaking backward compatibility
2. The text output should look as close to the current output as possible for readability purposes:
  a. Power shouldn't be another column. It should be updated in the power fields which are defined at the top of the electrical report
  b. The current and voltage of a sensor should be grouped together. We will have to search the boost ptree, which is fine
  c. The names should be more readable, i.e., we should have a _label_to_name_ map that converts "12V AUX" to "12 Volts Auxillary". If a label to name doesn't exist, we can simply print the label

Please let me know if you have any questions. 
Thanks!

chvamshi-xilinx:
> Hi @rajkumar-xilinx,
> 
> I have a few concerns:
> 
> 1. The json has been changed drastically, can you please provide the json output as well so we can make sure that we haven't changed any of the existing node names. Otherwise, we will be breaking backward compatibility
> 2. The text output should look as close to the current output as possible for readability purposes:
>    a. Power shouldn't be another column. It should be updated in the power fields which are defined at the top of the electrical report
>    b. The current and voltage of a sensor should be grouped together. We will have to search the boost ptree, which is fine
>    c. The names should be more readable, i.e., we should have a _label_to_name_ map that converts "12V AUX" to "12 Volts Auxillary". If a label to name doesn't exist, we can simply print the label
> 
> Please let me know if you have any questions. Thanks!

Hi @AShivangi , 
Regarding the JSON format, This change is only for Versal Devices. This is going to be data driven. There is no backward compatibility issues as this is the first time we are supporting Versal Devices.
Whenever we support ultrascale platforms with same mechanism, we need to add checks in shim to make sure Json is backward compatible. 

For point 2. We are just printing instantaneous values of Voltage/Current etc. We cannot group them together if we need to print Max and Average etc in future. Following snippet shows output of sensors tool which seperated out Voltage/Current etc



If there are other suggestions, we can consider them in the next PR, Please let me know if you have more questions. We can have a meeting to close those..

chvamshi-xilinx:
> > Hi @rajkumar-xilinx,
> > I have a few concerns:
> > 
> > 1. The json has been changed drastically, can you please provide the json output as well so we can make sure that we haven't changed any of the existing node names. Otherwise, we will be breaking backward compatibility
> > 2. The text output should look as close to the current output as possible for readability purposes:
> >    a. Power shouldn't be another column. It should be updated in the power fields which are defined at the top of the electrical report
> >    b. The current and voltage of a sensor should be grouped together. We will have to search the boost ptree, which is fine
> >    c. The names should be more readable, i.e., we should have a _label_to_name_ map that converts "12V AUX" to "12 Volts Auxillary". If a label to name doesn't exist, we can simply print the label
> > 
> > Please let me know if you have any questions. Thanks!
> 
> Hi @AShivangi , Regarding the JSON format, This change is only for Versal Devices. This is going to be data driven. There is no backward compatibility issues as this is the first time we are supporting Versal Devices. Whenever we support ultrascale platforms with same mechanism, we need to add checks in shim to make sure Json is backward compatible.
> 
> For point 2. We are just printing instantaneous values of Voltage/Current etc. We cannot group them together if we need to print Max and Average etc in future. Following snippet shows output of sensors tool which seperated out Voltage/Current etc
> 
> MicrosoftTeams-image
> 
> If there are other suggestions, we can consider them in the next PR, Please let me know if you have more questions. We can have a meeting to close those..

Hi @AShivangi , Thanks for joining a short sync up on this. After thinking for some more time. It makes more sense to use same node names in new mechanism also. We cannot merge current/voltage sensors etc, but we can definitely make sure that node names are same for new and old flow.

src/runtime_src/core/common/sensor.cpp
```diff
@@ -64,14 +66,16 @@ populate_sensor(const xrt_core::device * device, const std::string& loc_id, cons
     pt.put("current.error_msg", ex.what());
   }
   pt.put("current.amps", xrt_core::utils::format_base10_shiftdown3(current));
-  pt.put("current.is_present", current != 0 ? "true" : "false");
-  
+  pt.put("current.is_present", current != 0 ? true : false);
```
chvamshi-xilinx, line 69: As per my knowledge, this true and false are kept as string format for json output. Better not to change these to boolean. (all other places too.)

chvamshi-xilinx (approved):

rajkumar-xilinx:
**Attached JSON files for normal legacy model**
[ele_ta.txt](https://github.com/Xilinx/XRT/files/8035064/ele_ta.txt)
[mechanical_ta.txt](https://github.com/Xilinx/XRT/files/8035065/mechanical_ta.txt)
[thermal_ta.txt](https://github.com/Xilinx/XRT/files/8035066/thermal_ta.txt)

**Attached JSON files for data driven changes.**
[out_u50_dd2.txt](https://github.com/Xilinx/XRT/files/8035093/out_u50_dd2.txt)

AShivangi:
@chvamshi-xilinx, why can't we merge the sensors? I think the logic is something like:
Taking example of 12V AUX
1. populate voltage sensor info as 
```
{
  "id": "12V AUX",
  "description": "12 volts auxiliary",
  "current":
    {
        "is_present": "false"
    },
    "voltage":
    {
        "is_present": "true",
        "volts": "11.982",
         "max": "11.982",
         "average": "11.982"
    },
},
```
2. when populating current, search the boost::ptree to see if current label contains the volage label:
```
if (current_id.find(ptree.get("id")) != std::string::npos)
  get the sensor with voltage info
  put(current.is_present("true"))
  populate with the current info
else 
  create a new node
```

This should populate the node as:
```
{
  "id": "12V AUX",
  "description": "12 volts auxiliary",
  "current":
    {
        "amps": "1.228",
         "max": "1.228",
          "average": "1.228",
          "is_present": "true"
    },
    "voltage":
    {
        "is_present": "true",
        "volts": "11.982",
         "max": "11.982",
         "average": "11.982"
    },
},
```

Please let me know if there is a reason we are not doing this. 

@rajkumar-xilinx, additional few comments on this submission:
1. Power is a calculated value and is only present per device, not per sensor. Please remove "power" and add all the power info to the following nodes:
```
"power_consumption_watts": "15.559396",
 "power_consumption_max_watts": "NA",
 "power_consumption_warning": "NA"
``` 
2. When we are collecting current we have the following in the json
```
{
                        "description": "VCC INT Current",
                        "current":
                        {
                            "amps": "5.500",
                            "max": "5.500",
                            "average": "5.500",
                            "is_present": "true"
                        },
                        "voltage":
                        {
                            "is_present": "false"
                        },
```
but when we are collecting voltage we show 
```
{
                        "description": "12V PEX",
                        "current":
                        {
                            "amps": "0.000",
                            "max": "0.000",
                            "average": "0.000",
                            "is_present": "false"
                        },
                        "voltage":
                        {
                            "is_present": "true",
                            "volts": "11.982",
                            "max": "11.982",
                            "average": "11.982"
                        },
```
if the current is not present, we shouldn't populate "amps", "max", "average" like we do for voltage in the first case. 

3. For backward compatibility, we should have "id" as well as "description" nodes. It is ok to have the same value for both these nodes. However, we can't simply remove a node without changing the schema version. 

Thanks for the work so far!

chvamshi-xilinx:
> @chvamshi-xilinx, why can't we merge the sensors? I think the logic is something like: Taking example of 12V AUX
> 
> 1. populate voltage sensor info as
> 
> ```
> {
>   "id": "12V AUX",
>   "description": "12 volts auxiliary",
>   "current":
>     {
>         "is_present": "false"
>     },
>     "voltage":
>     {
>         "is_present": "true",
>         "volts": "11.982",
>          "max": "11.982",
>          "average": "11.982"
>     },
> },
> ```
> 
> 2. when populating current, search the boost::ptree to see if current label contains the volage label:
> 
> ```
> if (current_id.find(ptree.get("id")) != std::string::npos)
>   get the sensor with voltage info
>   put(current.is_present("true"))
>   populate with the current info
> else 
>   create a new node
> ```
> 
> This should populate the node as:
> 
> ```
> {
>   "id": "12V AUX",
>   "description": "12 volts auxiliary",
>   "current":
>     {
>         "amps": "1.228",
>          "max": "1.228",
>           "average": "1.228",
>           "is_present": "true"
>     },
>     "voltage":
>     {
>         "is_present": "true",
>         "volts": "11.982",
>          "max": "11.982",
>          "average": "11.982"
>     },
> },
> ```
> 
> Please let me know if there is a reason we are not doing this.
> 
> @rajkumar-xilinx, additional few comments on this submission:
> 
> 1. Power is a calculated value and is only present per device, not per sensor. Please remove "power" and add all the power info to the following nodes:
> 
> ```
> "power_consumption_watts": "15.559396",
>  "power_consumption_max_watts": "NA",
>  "power_consumption_warning": "NA"
> ```
> 
> 2. When we are collecting current we have the following in the json
> 
> ```
> {
>                         "description": "VCC INT Current",
>                         "current":
>                         {
>                             "amps": "5.500",
>                             "max": "5.500",
>                             "average": "5.500",
>                             "is_present": "true"
>                         },
>                         "voltage":
>                         {
>                             "is_present": "false"
>                         },
> ```
> 
> but when we are collecting voltage we show
> 
> ```
> {
>                         "description": "12V PEX",
>                         "current":
>                         {
>                             "amps": "0.000",
>                             "max": "0.000",
>                             "average": "0.000",
>                             "is_present": "false"
>                         },
>                         "voltage":
>                         {
>                             "is_present": "true",
>                             "volts": "11.982",
>                             "max": "11.982",
>                             "average": "11.982"
>                         },
> ```
> 
> if the current is not present, we shouldn't populate "amps", "max", "average" like we do for voltage in the first case.
> 
> 3. For backward compatibility, we should have "id" as well as "description" nodes. It is ok to have the same value for both these nodes. However, we can't simply remove a node without changing the schema version.
> 
> Thanks for the work so far!

Hi @AShivangi , Thanks for your review.
We can group them very easily, but the problem is, sensor names are not same for current/voltage etc. Anyway, we will try to merge if there is a common sensor between current/voltage. for example: "12V PEX" and "12V PEX Current" comes as two different nodes. 
We can fix other three points as well.

AShivangi:
@chvamshi-xilinx, 
Thanks, I remember talking to David about the sensor names. It is the SC team's responsibility to name the sensors correctly and make sure that the voltage/current label pair matches. We can reiterate this to them if you'd like. 

@rajkumar-xilinx, after making these changes, please update the PR desc with the new text and json output. 

Thanks again!

kbhardwa-xilinx:
retest this please

rajkumar-xilinx:
Hi @AShivangi 

Please review the output and let me know.
electrical_dd_14feb
[electrical_dd_14feb.txt](https://github.com/Xilinx/XRT/files/8060623/electrical_dd_14feb.txt)
[mechanical_dd_14feb.txt](https://github.com/Xilinx/XRT/files/8060632/mechanical_dd_14feb.txt)
[thermal_dd_14feb.txt](https://github.com/Xilinx/XRT/files/8060633/thermal_dd_14feb.txt)

rajkumar-xilinx:
retest this please

kbhardwa-xilinx:
retest this please

AShivangi:
Hi @rajkumar-xilinx, 
I still see a couple of issues:
```
 {
                        "id": "5V5 SYS",
                        "description": "5V5 SYS",
                        "current":
                        {
                            "amps": "0.000",
                            "max": "0.000",
                            "average": "0.000",
                            "is_present": "false"
                        },
                        "voltage":
                        {
                            "is_present": "true",
                            "volts": "5.001",
                            "max": "5.001",
                            "average": "5.001"
                        },
                        "power":
                        {
                            "is_present": "false"
                        }
                    },
```

1. Please address point 2 from my prev comment, which was, if current is false, we shouldn't populate amps, max or avg. It should look like:
```
...
                   "description": "5V5 SYS",
                        "current":
                        {
                            "is_present": "false"
                        },
                        "voltage":
...
```
2. Why do we have power?
```
                      "power":
                        {
                            "is_present": "false"
                        }
```
This node should be completely removed

3. When populating voltage/current pair, please use the voltage label instead of the current label. This is because current label has the string "current" in it, which doesn't make sense as a sensor name. For eg: `12V PEX Current` should be replaced by `12V PEX`. This should be easy to fix as you can switch to populating voltage before current

4. In mechanical and thermal json, please replace `id` node with `location_id` for consistency purposes


**Question about thermal rep:**
How are you getting a label, if the sensor is not present?
```
...
               {
                    "id": "DIMM1 TEMP",
                    "description": "DIMM1 TEMP",
                    "temp_C": "0",
                    "highest": "0.000",
                    "is_present": "false"
                },
...
```

rajkumar-xilinx:
> Question about thermal rep:
> How are you getting a label, if the sensor is not present?

Hi @AShivangi ,

I added a check like below, so marking is_present only true if the received value non-zero. So, the received sensor's value is zero. Please let me know if we don't need a check here.

pt.put("is_present", temp_C != 0 ? "true" : "false");

rajkumar-xilinx:
Hi @AShivangi 

Attached json outputs again, please review

[mechanical_16feb.txt](https://github.com/Xilinx/XRT/files/8077585/mechanical_16feb.txt)
[mechanical_dd_16feb.txt](https://github.com/Xilinx/XRT/files/8077586/mechanical_dd_16feb.txt)

[thermal_16feb.txt](https://github.com/Xilinx/XRT/files/8077587/thermal_16feb.txt)
[thermal_dd_16feb.txt](https://github.com/Xilinx/XRT/files/8077588/thermal_dd_16feb.txt)

[electrical_16feb.txt](https://github.com/Xilinx/XRT/files/8077589/electrical_16feb.txt)
[electrical_dd_16feb.txt](https://github.com/Xilinx/XRT/files/8077584/electrical_dd_16feb.txt)

rajkumar-xilinx:
image

image

AShivangi:
Hi @rajkumar-xilinx, 
I agree. I think we should print 0 temp with the label. However in your latest json I still see, is_present as false. If you removed that check, then all the sensors which are data driven should have "true". 

Mechanical report: "id" should be "location_id"

Electrical report:
```
{
                        "id": "12V AUX",
                        "description": "12V AUX",
                        "voltage":
                        {
                            "volts": "0.000",
                            "max": "0.000",
                            "average": "0.000",
                            "is_present": "true"
                        },
                        "current":
                        {
                            "is_present": "true",
                            "amps": "0.000",
                            "max": "0.000",
                            "average": "0.000"
                        }
                    },
```
How are these sensor values present? This doesn't look right. I would assume that a sensor has some current and voltage...

rajkumar-xilinx:
Hi @AShivangi 
Attaching updated results, please review

[thermal_19feb.txt](https://github.com/Xilinx/XRT/files/8101444/thermal_19feb.txt)
[thermal_dd_19feb.txt](https://github.com/Xilinx/XRT/files/8101445/thermal_dd_19feb.txt)

[electrical_19feb.txt](https://github.com/Xilinx/XRT/files/8101446/electrical_19feb.txt)
[electrical_dd_19feb.txt](https://github.com/Xilinx/XRT/files/8101447/electrical_dd_19feb.txt)

[mechanical_19feb.txt](https://github.com/Xilinx/XRT/files/8101448/mechanical_19feb.txt)
[mechanical_dd_19feb.txt](https://github.com/Xilinx/XRT/files/8101443/mechanical_dd_19feb.txt)

rajkumar-xilinx:
Tried these changes on VCK5000, below results are from VCK5000-es3 setup
image

Temp & Fan data needs driver changes, filed CR https://jira.xilinx.com/browse/CR-1123156

AShivangi (approved):
Looks good. Thanks for making these changes

rozumx (approved):
Hello Raj,

With how the code is currently implemented it is sufficient to be merged into the code base.   That said, it is a regression with regards to the "description".  Initializing the description with the ID should only be done as the last resort.

I do understand that there has been some conversations already regarding this, but as present isn't a solution we should go with.   Good news is that I have some ideas on how to best address this short coming and will send it to you via CRs :^).

Thank you again for your effort regarding this code.

PR #6192, opened 2022-01-27
XRT-938 Add image hash and package manifest to apu pdi

houlz0507:
APU firmware does not have version information
Pack petalinux kernel image and rootfs hash and package manifest to apu pdi. XRT driver and VMR will be able to read and show those information.

### DIFF
--- a/src/runtime_src/tools/scripts/pkgapu.sh
+++ b/src/runtime_src/tools/scripts/pkgapu.sh
@@ -109,6 +109,10 @@ SYSTEM_DTB_ADDR="0x40000"
 KERNEL_ADDR="0x20100000"
 ROOTFS_ADDR="0x21000000"
 
+# this address needs to be in sync with VMR
+METADATA_ADDR="0x7FBD0000"
+METADATA_BUFFER_LEN=131072
+
 clean=0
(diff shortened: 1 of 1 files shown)


xdavidz (approved):

src/runtime_src/tools/scripts/pkgapu.sh
```diff
@@ -109,6 +109,10 @@ SYSTEM_DTB_ADDR="0x40000"
 KERNEL_ADDR="0x20100000"
 ROOTFS_ADDR="0x21000000"
 
+# this address needs to be in sync with VMR
```
rozumx, line 112: Could you add a comment to where one can find it in the VMR.   

My concern is that if someone changes it in the VMR how would this script know.   Is it possible for this date to be data driving by a JSON or some other file.

rozumx (approved):

src/runtime_src/tools/scripts/pkgapu.sh
houlz0507: This is temporary solution. We will change it to data driven after xsabin section metadata is supported.

PR #6197, opened 2022-01-28
CR-1120170 Fix for seg fault when connectivity section is missing in sw emu xclbin

ramyam-xilinx:
Fix for seg fault when connectivity section is missing in sw emu xclbin

There are few cases where connectivity section is missing because all kernel args are scalars.  
This lead to a crash because there is no connectivity.

Since 2021.2 we have rtd getting generated for all software emulation xclbins, and we can 
now remove the xclbin_swemu.cpp code path that kicks in when connectivity section is missing.

Removes sw emulation support for legacy xclbins that do not include runtime data.
Supposedly there should be none of these by now and if there is, the xclbins should be regenerated.

Canary run and the results are good.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_xclbin.cpp
+++ b/src/runtime_src/core/common/api/xrt_xclbin.cpp
@@ -754,21 +754,8 @@ class xclbin_full : public xclbin_impl
 
     m_uuid = uuid(m_top->m_header.uuid);
 
-    const ::ip_layout* ip_layout = nullptr;
     for (auto kind : kinds) {
       auto hdr = xrt_core::xclbin::get_axlf_section(m_top, kind);
-
-      // software emulation xclbin does not have all sections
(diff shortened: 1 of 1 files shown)


stsoe (approved):

stsoe:
Genuine error, please fix
```
 19:27:51 R:\builds\xrt\master\10680\src\runtime_src\core\common\api\xrt_xclbin.cpp(73): error C2220: warning treated as error - no 'object' file generated [R:\builds\xrt\master\10680\build\WRelease\runtime_src\core\common\xrt_coreutil.vcxproj]
19:27:51 R:\builds\xrt\master\10680\src\runtime_src\core\common\api\xrt_xclbin.cpp(73): warning C4505: '`anonymous-namespace'::is_sw_emulation': unreferenced local function has been removed 
```

salindac:
retest this please. Infra error, xtc data dir got removed

PR #6201, opened 2022-01-29
CR-1119283: Adding a generic control switch for enabling host trace

jvillarre:
Applications can interact with XRT by calling APIs at the OpenCL level, the Native XRT level, and the HAL level.  Individual controls exist to turn on the generation of trace for each of these levels, but can lead to unexpected results (such as turning on opencl_trace=true on an application that uses Native XRT APIs generates no trace).

This pull request adds a generic "host_trace" switch to the xrt.ini file that will enable either opencl_trace, native_xrt_trace, or xrt_trace based on the first function called by the application, thereby providing the topmost layer of host trace regardless of how the application was written.  The other ini switches remain available for fine grained control.


src/runtime_src/core/common/config_reader.h
```diff
@@ -289,6 +289,29 @@ get_profile_api()
...
+
+  // host_trace is a generic switch that should enable one layer of 
+  //  trace (either OpenCL level, Native XRT API level, or HAL level).
+  //  Whichever function is called from the host code first should enable
+  //  that level of trace.  Therefore, an OpenCL application will only
+  //  enable OpenCL trace, a Native XRT API application will only enable
+  //  Native XRT trace, and a HAL level application will only enable HAL trace.
+  bool result = get_host_trace() && !called;
+  called = true ;
+  return result ;
```
stsoe, line 312: This is not thread safe.  `called` can be written by one thread on line 311, while some other thread computes 310 and a third read executes 312.    That sort of logic does not belong here in this header I won't even claim to understand what it does and why, the comment didn't help me.

stsoe (changes requested):

src/runtime_src/core/common/config_reader.h
jvillarre: The function was renamed and moved out of the config_reader.h file into the utils.cpp file, multithreaded protection was added via lock_guard, and the comments were redone to clarify what host_trace=true is meant to enable and what the load_host_trace() functionality and usage is.

stsoe (approved):

pgschuey (approved):
changes look good

chvamshi-xilinx (approved):

IshitaGhosh (approved):

keryell:
I am just compiling on `xsjsycl41` and got
```
 5/15 Test  #5: python_binding ....................***Failed    0.05 sec
Traceback (most recent call last):
  File "/var/tmp/rkeryell/SYCL/XRT/src/../tests/python/200_binding/200_main.py", line 16, in <module>
    import xrt_binding as xx
  File "/var/tmp/rkeryell/SYCL/XRT/build/Debug/opt/xilinx/xrt/python/xrt_binding.py", line 26, in <module>
    from xclbin_binding import *
  File "/var/tmp/rkeryell/SYCL/XRT/build/Debug/opt/xilinx/xrt/python/xclbin_binding.py", line 24, in <module>
    libc = ctypes.CDLL(os.environ['XILINX_XRT'] + "/lib/libxrt_core.so")
  File "/usr/lib/python3.9/ctypes/__init__.py", line 374, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: ./opt/xilinx/xrt/lib/libxrt_core.so: undefined symbol: _ZN8xrt_core5utils15load_host_traceEv
```
which is `xrt_core::utils::load_host_trace()`.
Is this just a coincidence?

keryell:
Thanks to @jvillarre who helped me finding the problem! \o/
I had in my `LD_LIBRARY_PATH` a leftover of a Vitis+XRT install `/opt/xilinx/xrt/lib:/opt/xilinx/Vitis/2021.2/lib/lnx64.o:/llvm/build/lib` which were picked instead of the `rpath`-related just built local library:
```
 ldd ./Release/runtime_src/core/pcie/linux/libxrt_core.so
	linux-vdso.so.1 (0x00007fff2bdb1000)
	libxrt_coreutil.so.2 => /opt/xilinx/xrt/lib/libxrt_coreutil.so.2 (0x00007f7a269e2000)
```

PR #6203, opened 2022-01-29
xclbinutil : Fix PSKernel hang issue

rozumx:
xclbinutil would 'hang" the objdump returned more than 4K of data.

**_Underlying issue_**
If the calling OS process return more data then the buffer associated with std::out has removed for, the calling process would become block, which in turns hangs the application. 

This issue is not a regression.  Instead it was discovered during stress testing.

This issue was addressed by making a boost  asyncronous process call instead of a syncronous process call.

No risks.  Code is now more stable

Manual testing against the AIE profiling PS Kernel.  

N/A

### DIFF
--- a/src/runtime_src/tools/xclbinutil/ElfUtilities.cxx
+++ b/src/runtime_src/tools/xclbinutil/ElfUtilities.cxx
@@ -32,7 +32,7 @@
 namespace XUtil = XclBinUtilities;
 
 std::vector<std::string> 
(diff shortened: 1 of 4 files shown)


salindac:
retest this please. Infra error, xtc data dir got removed

dayeh-xilinx:
retest this please. Fixed the centos8 repo issue.

stsoe (approved):
LGTM

PR #6207, opened 2022-01-31
Fix for 1120119 : Seg fault in AIE design with aie_trace enabled

IshitaGhosh:
Seg fault in AIE design was due to access to deleted AieTracePlugin static object from Edge SHIM destructor to finish AIE device flush
This is fixed by introducing a status flag in AieTracePlugin and checking its validity before any callback.


jvillarre (approved):
Added guards look good to protect against the situation when the static destructors are called in the order where the shim destructor and callbacks happen after the plugin is destroyed.

pgschuey (approved):
Changes looks good. Great detective work.

dayeh-xilinx:
retest this please.

dayeh-xilinx:

dayeh-xilinx:

PR #6210, opened 2022-02-01
CR-1116870 Emu driver uses xclbin_parser, xrt_xclbin.h and xclbin_int.h API's for parsing xclbin

udaybhas-xilinx:
Emu driver uses xclbin_parser, xrt_xclbin.h and xclbin_int.h API's for parsing xclbin

https://jira.xilinx.com/browse/CR-1116870
https://jira.xilinx.com/browse/CR-1116752

Low Risk

Tests: 1. Canary is neat.

### DIFF
--- a/src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
+++ b/src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
@@ -232,11 +232,6 @@ namespace xclhwemhal2 {
     }
   }
 
-  static size_t convert(const std::string& str)
-  {
-    return str.empty() ? 0 : std::stoul(str,0,0);
-  }
-
   static void sigHandler(int sn, siginfo_t *si, void *sc)
   {
     switch(sn) {
@@ -304,12 +299,12 @@ namespace xclhwemhal2 {
     ssize_t pdiSize = 0;
     ssize_t emuDataSize = 0;
 
-    char* zipFile = nullptr;
-    char* xmlFile = nullptr;
-    char* debugFile = nullptr;
-    char* memTopology = nullptr;
-    char* pdi = nullptr;
-    char* emuData = nullptr;
+    std::unique_ptr<char[]> zipFile;
+    std::unique_ptr<char[]> xmlFile;
+    std::unique_ptr<char[]> debugFile;
+    std::unique_ptr<char[]> memTopology;
+    std::unique_ptr<char[]>  pdi;
+    std::unique_ptr<char[]> emuData;
 
     if (std::memcmp(bitstreambin, "xclbin2", 7)) {
       PRINTENDFUNC;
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -579,44 +533,14 @@ namespace xclhwemhal2 {
...
-
-      fpgaDevice = xml_device.second.get<std::string>("<xmlattr>.fpgaDevice");
-
-      if (++count > 1)
-      {
-        //Give error and return from here
-      }
-    }
+    auto projectName = xrt_core::xclbin::get_project_name(args.m_top);
+    auto xclbin_object = std::make_unique<xrt::xclbin>(args.m_top);
```
stsoe, line 539: Don't new this object.  Just keep in an xrt::xclbin stack object.

stsoe (changes requested):
Thanks!  A step in the right direction.

src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -321,87 +316,49 @@ namespace xclhwemhal2 {
...
-        pdi = nullptr;
-      }
-
-      if (emuData) {
-        delete[] emuData;
-        emuData = nullptr;
-      }
-      return -1;
+      emuData = std::make_unique<char[]>(emuDataSize);
+      memcpy(emuData.get(), bitstreambin + sec->m_sectionOffset, emuDataSize);
```
keryell, line 345: While looking at this assembly code, I have the feeling that `std::vector<char>` everywhere would be simpler. But probably an über-refactoring for next century.

stsoe: @keryell I was thinking the same, but didn't bother to comment on it.  Maybe with the exception of where you need a custom deleter, I don't think std::unique_ptr<[]> should be used.

udaybhas-xilinx: @stsoe So, **instead of** "unique_ptr" should i use "new" then ... Please confirm !!

stsoe: @udaybhas-xilinx You should never ever use 'new and delete'.
What we are saying is that `std::vector<char> vec(size)` is as good as, or better than, `std::unique_ptr<char[]>`

chvamshi-xilinx (approved):

dayeh-xilinx:

dayeh-xilinx:

dayeh-xilinx:

PR #6200, opened 2022-01-28
Built-in PS Kernel Build flow

chvamshi-xilinx:
#### Problem solved by the commit
This PR provides a mechanism to build PS Kernel xclbin and package it to RPM/DEB. This is part of the pkgapu.sh script. These PS Kernels are not final Kernels. I will add a hello world PS-Kernel once mechanism to release PS Kernel xclbin is clear.
1) Added Jason provided built-in PS Kernel sources
2) Added CMake files to build these PS Kernel
3) Updated CMake files to release these PS Kernels libraries to /usr/lib/ps_kernels_lib
4) Updated pkgapu.sh build one xclbin from these PS Kernel so's and added it to RPM/DEB

#### Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered
NA

#### How problem was solved, alternative solutions (if any) and why they were rejected
NA

#### Risks (if any) associated the changes in the commit
No

#### What has been tested and how, request additional testing if necessary
Verified the APU Package build using latest 2022.1 Petalinux

PR #6211, opened 2022-02-01
updating built-in PS-kernel with sample PS kernel

chvamshi-xilinx:

- We are building built-in PS Kernel xclbin as part of the APU Package creation. 
- We were using aie_trace PS Kernel as an example , This kernel is still under development by Jason.
- It is going to be confusing for Jason to have two copies in source tree 
- Removing aie trace PS kernel and adding a simple Hello-world PS Kernel.

https://github.com/Xilinx/XRT/pull/6200 Added aie_trace PS Kernel in this PR.

Used very simple kernel as this PR is to explain how APU packages are built

No

Created APU Package and verified .


rozumx (approved):
Approved

stsoe (approved):
Thanks for explaining.

PR #6212, opened 2022-02-01
hwmon_sdm: update & store sensor values when required

rajkumar-xilinx:
hwmon_sdm driver has to update the required sensors by issuing xgq requests to VMR/VMC firmware when required.

Added a simple timer in hwmon_sdm driver for issuing xgq requests and storing sensors in cache if the sensors are being accessed by user.

So, sensors in cache will be updated only when user is trying to access them. Until then, no xgq commands issued.

Configured HWMON_SDM_DEFAULT_EXPIRE_SECS  to 1 second by default. So, if the sensors are accessed within 1 second of time, then cached values are sent to user. If the sensors are accessed after 1 second, then new xgq command issued, and received sensors are stored in cache, and shared with the user.

With this commit, we will see updated/latest sensor values when accessed through xbmgmt tool or hwmon sysfs directory.

Tested it on VCK5000 device.


rajkumar-xilinx:
Validated the changes on vck5000 VMR setup

$cat /sys/bus/pci/devices/0000\:d9\:00.0/hwmon/hwmon3/temp1_label
FAN RPM

$cat /sys/bus/pci/devices/0000\:d9\:00.0/hwmon/hwmon3/temp1_ins
233

//Reading the sensor again after 60 seconds
$ cat /sys/bus/pci/devices/0000\:d9\:00.0/hwmon/hwmon3/temp1_ins
194

houlz0507 (approved):

maxzhen (approved):
This code is on xclmgmt side, right? Do we have code in xocl to fetch sensor from xclmgmt? If we do, will this cache timer also add to the total cache timer on xocl side?

rajkumar-xilinx:
> This code is on xclmgmt side, right? Do we have code in xocl to fetch sensor from xclmgmt? If we do, will this cache timer also add to the total cache timer on xocl side?

Hi @maxzhen yes, it is for mgmt side. I am adding changes in hwmon_sdm driver w.r.t xocl part now and the same timer will be used there as well.

maxzhen:
> > This code is on xclmgmt side, right? Do we have code in xocl to fetch sensor from xclmgmt? If we do, will this cache timer also add to the total cache timer on xocl side?
> 
> Hi @maxzhen yes, it is for mgmt side. I am adding changes in hwmon_sdm driver w.r.t xocl part now and the same timer will be used there as well.

So, if we do that, then the cache on xocl side may expire b/w 1 seconds and 2 seconds since we'll go through two timers? That does not sound ideal...

src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
```diff
@@ -480,50 +521,71 @@ static int hwmon_sdm_probe(struct platform_device *pdev)
...
 		xocl_err(&pdev->dev, "request is failed with err: %d", ret);
+		sdm->sensor_data_avail[repo_id] = false;
+	}
+
+	return ret;
+}
+
+static void hwmon_sdm_get_sensors_list(struct platform_device *pdev, bool create_sysfs)
+{
+	(void) hwmon_sdm_update_sensors_by_type(pdev, SDR_TYPE_BDINFO, false);
```
chvamshi-xilinx, line 568: We need to pass create_sysfs instead of false?

rajkumar-xilinx: we are not creating any sysfs entries for bdinfo. So, passing false for BDINFO request.

chvamshi-xilinx: Thanks Raj for clarifying. This means, we need to capture bdinfo in some other way sothat xbutil can read.

chvamshi-xilinx (approved):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
rajkumar-xilinx: yes, plan is to create a sysfs entries under hwmon_sdm device, so that xbutil can read these values.

dayeh-xilinx:

PR #6213, opened 2022-02-01
Remove "XRT upload build artifacts" step from minipipeline

eshwar-xilinx:

`XRT build artifacts` step is responsible for uploading the `XRT rpm/deb packages` (generated during building XRT) to github. 
Normally, it takes around 10 min for uploading all the build artifacts to github but it can go on for as long as 20 min based on network connectivity and github actions. 
The `verification` job itself has a timeout of 40 min. Hence, PR run will be timed-out, showing as "run has failed" eventhough it is cancelled in midway by github actions. The commit will resolve this timeout issue.

This issue was discovered through manual observation

Removing the XRT build artifacts step resolves the issue. It further speeds up the minipipeline runs. 
Alternative solution is to increase timeout for verification job. This is rejected because allocating 10-20 min of PR build for uploading `1.2GB` folder to github is not a good choice.

NA

NA

NA

### DIFF
--- a/.github/workflows/pipeline.yml
+++ b/.github/workflows/pipeline.yml
@@ -87,13 +87,13 @@ jobs:
 
       # XRT build aritfacts step uploads the build/Release folder to github action in zip format
       # This step uses upload-artifacts action which is maintained by github
-    - name: XRT build artifacts
+#    - name: XRT build artifacts
       # if always() keyword is set so that even if previous step fails or workflow is cancelled, this step is set to run always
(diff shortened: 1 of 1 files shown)


eshwar-xilinx:
ok to test

chvamshi-xilinx:
add to whitelist

chvamshi-xilinx (approved):

chvamshi-xilinx (approved):

PR #6214, opened 2022-02-02
CR-1120136 Kernel limit error

stsoe:
Extend kernel and instance name to 64 bytes each in driver.

Customer bug encountered problem supposedly when using clSetKernelArg
in older 2021.1 release.  Debugged the problem on XRT main branch and
saw that core XRT could not find a requested compute unit because the
instance name was greater than 32 bytes.

The instance names are returned via sysfs from driver and shows
truncated names. Expanded the arrays holding kernel and instance name
to 64 bytes each.  Per xclbin a kernel:instance name can be 64 bytes,
but there is no special limit on the kernel and instance name and
either can be 64 bytes minus few special chars.

Not seeing any risks.

Old HW regression tests

### DIFF
--- a/src/runtime_src/core/common/drv/include/xrt_cu.h
+++ b/src/runtime_src/core/common/drv/include/xrt_cu.h
@@ -227,8 +227,8 @@ struct xrt_cu_info {
 	bool			 sw_reset;
 	struct xrt_cu_arg	*args;
 	u32			 num_args;
-	char			 iname[32];
-	char			 kname[32];
+	char			 iname[64];
+	char			 kname[64];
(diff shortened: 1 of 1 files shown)


mamin506 (approved):

salindac:
retest please - infra error '/proj/rdi/testcases/xtc/REL/2022.1/tc/open/scout/xrt/pipeline/' does not exist

stsoe:
retest this please - emulation test failures, but changes in this PR are not used in emulation

dayeh-xilinx:

dayeh-xilinx:

PR #6218, opened 2022-02-02
CR-1118707 Python based host code design support for emulationflow

udaybhas-xilinx:
Problem solved by the commit
Python based host code design support for emulationflow

#### Bug / issue (if any) fixed, 
https://jira.xilinx.com/browse/CR-1118707

Made changes to load the proper library based on “XCL_EMULATION_MODE” 

Low Risk

Verified sw_emu and hw_emu examples  for u250 paltform

NA

### DIFF
--- a/src/python/ert_binding.py
+++ b/src/python/ert_binding.py
@@ -22,13 +22,10 @@ import ctypes
 import os
 
 if os.environ.get('XCL_EMULATION_MODE') == 'hw_emu':
-    print("## in ert hw emu case ## ")
     libc = ctypes.CDLL(os.environ['XILINX_XRT'] + "/lib/libxrt_hwemu.so")
 elif os.environ.get('XCL_EMULATION_MODE') == 'sw_emu':
-    print("## in ert sw emu case ## ")
     libc = ctypes.CDLL(os.environ['XILINX_XRT'] + "/lib/libxrt_swemu.so") 
 else:
-    print("## in ert else case ## ")
     libc = ctypes.CDLL(os.environ['XILINX_XRT'] + "/lib/libxrt_core.so")
 
 ##  START OF ENUMS  ##
--- a/src/python/ert_binding.py
+++ b/src/python/ert_binding.py
@@ -21,7 +21,15 @@
 import ctypes
 import os
 
-libc = ctypes.CDLL(os.environ['XILINX_XRT'] + "/lib/libxrt_core.so")
+if os.environ.get('XCL_EMULATION_MODE') == 'hw_emu':
+    print("## in ert hw emu case ## ")
+    libc = ctypes.CDLL(os.environ['XILINX_XRT'] + "/lib/libxrt_hwemu.so")
+elif os.environ.get('XCL_EMULATION_MODE') == 'sw_emu':
+    print("## in ert sw emu case ## ")
+    libc = ctypes.CDLL(os.environ['XILINX_XRT'] + "/lib/libxrt_swemu.so") 
+else:
+    print("## in ert else case ## ")
+    libc = ctypes.CDLL(os.environ['XILINX_XRT'] + "/lib/libxrt_core.so")
 
 ##  START OF ENUMS  ##
 
--- a/src/python/xclbin_binding.py
+++ b/src/python/xclbin_binding.py
@@ -22,13 +22,10 @@ import os
 import ctypes
 
 if os.environ.get('XCL_EMULATION_MODE') == 'hw_emu':
-    print("## in xclbin hw emu case ## ")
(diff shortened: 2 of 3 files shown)


chvamshi-xilinx (commented):
Please update the print messages. These messages looks like debug messages..
Is sw_emu crash issue fixed? If no, are there more changes required?

chvamshi-xilinx:
ok to test

udaybhas-xilinx:
> Looks good. Please update the print messages. These messages looks like debug messages.. Is sw_emu crash issue fixed? If no, are there more changes required?

No crash is observed in sw_emu...

chvamshi-xilinx:
> > Looks good. Please update the print messages. These messages looks like debug messages.. Is sw_emu crash issue fixed? If no, are there more changes required?
> 
> No crash is observed in sw_emu...

Good to know.

chvamshi-xilinx (approved):

stsoe (commented):
I thought we were moving exclusively to pybind11?
Are we sure these ctypes bindings are what we want?

AShivangi:
Good point Soren. I see python test cases including xrt_binding.py, etc. as well as pyxrt. That doesn't make sense. I think we need to refactor the test cases to not rely on *_binding.py and get rid of those files all together

dayeh-xilinx:

dayeh-xilinx:

dayeh-xilinx:

AShivangi (approved):

stsoe (changes requested):
Hi @udaybhas-xilinx.  Please respond to my comment about pybind11 and why you are using ctypes.

venkatp-xilinx:
> Hi @udaybhas-xilinx. Please respond to my comment about pybind11 and why you are using ctypes.

@stsoe, some legacy test cases are using ctypes still, just to address those legacy designs, we need to do these modifications. I could see the test case in the CR is using the xrt_bindings and pybind11 together. We tried commenting the ctypes usage in the test case, yet resulted the same issue. However with these changes we could see test is passing.

stsoe (changes requested):
I don't think we should change the ctypes bindings. I updated the CR.

stsoe:
> @stsoe, some legacy test cases are using ctypes still, just to address those legacy designs, we need to do these modifications. I could see the test case in the CR is using the xrt_bindings and pybind11 together. We tried commenting the ctypes usage in the test case, yet resulted the same issue. However with these changes we could see test is passing.

Let not cater to legacy tests that should be updated.  It's not a good idea to add obsolete code to the system.

stsoe:
> @stsoe, some legacy test cases are using ctypes still, just to address those legacy designs, we need to do these modifications. I could see the test case in the CR is using the xrt_bindings and pybind11 together. We tried commenting the ctypes usage in the test case, yet resulted the same issue. However with these changes we could see test is passing.

Let not cater to legacy tests that should be updated.  It's not a good idea to add obsolete code to the system.

stsoe:
Hi @AShivangi Do you think you can help understand why these the changes proposed in this PR would be necessary if the application is using pyxrt pybind11?   I am not understanding why xrt_bindings. needs to be modified to check an XCL_EMULATION_MODE when xrt_coreutil does this and when pyxrt is linked with this library.

venkatp-xilinx:
> > Hi @AShivangi Do you think you can help understand why these the changes proposed in this PR would be necessary if the application is using pyxrt pybind11? I am not understanding why xrt_bindings. needs to be modified to check an XCL_EMULATION_MODE when xrt_coreutil does this and when pyxrt is linked with this library.
> 
> Hi @stsoe, I believe that the python test cases are still dependent on xclbin_binding.py and ert_binding.py. For eg: in `tests/python/utils_binding.py` we have `wrap_get_axlf_section(..)` which comes from xclbin_binding.py. I agree with you that we shouldn't be supporting the ctypes anymore, but adding xclbin and ert bindings through pybind is a much bigger task.

@stsoe, I can see a label do-not-merge added to this PR. However with the latest comments from @AShivangi, seems its necessary to change the xclbin_bindings and ert bindings even though users use just the pybind11. Are we going to merge this change?

AShivangi:
@stsoe, I hadn't looked at python tests for a while, so wasn't aware of what is not being used anymore. Thanks for looking into it. I agree with you that we should improve the python test cases to not use the bindings anymore. This leads me to question the ctypes bindings altogether. Should we still expose them to the user?
Also, who will be fixing the tests? I do not have the bandwidth to do it for this release.

chvamshi-xilinx:
Closing as emulations tests working with pyxrt(native xrt apis)

PR #6224, opened 2022-02-03
Update kernel default address range to 64k

stsoe:
Change kernel_properties::address_range to default 64k.

Normally address range is extracted from XML meta data in xclbin and
returned to core XRT in a kernel_property struct.  For some RTL
kernels, the XML section may not be present, but kernel properties are
still created used by xrt::ip to determine the address range.

The address range should be be meta data in xclbin one way or another,
and defaulting to 64k should not be done.  However, we still have to
figure out where address range can be maintained for kernels without
XML metadata.

Affects only kernels without XML metadata and when used with xrt::ip.  Without this 
change xrt::ip cannot even be used with these kernels.


dayeh-xilinx:

xilinxgitops:
retest this please

PR #6227, opened 2022-02-03
CR-1120827 xbutil validate failure on discovery platform

mamin506:
Fix CR-1120827.

This bug only happen on a shell + verify.xclbin that has a very small memory bank. In this CR, it is 128KB.
The xcl_iops_test will launch 2 threads to run test concurrently. Each thread will try to allocate 128 output buffers.
The thread is still okay to run with less than 128 buffers, but the performance maybe lower.

The small memory bank device doesn't have enough memory to satisfy both threads. 
If one thread cannot allocate any buffers, then the test will crash. This is the bug.

1. If one thread cannot get event 1 command, the test case should return EOPNOTSUPP instead of crash.
2. Synchronize threads to fairly allocate buffers.

For other device, all of the threads can allocate 128 output buffers. No impact on those devices.

Test with Versal discovery shell.

No.

### DIFF
--- a/tests/validate/xcl_iops_test/src/xcl_api_iops.cpp
+++ b/tests/validate/xcl_iops_test/src/xcl_api_iops.cpp
@@ -43,7 +43,8 @@ struct krnl_info {
 };
 
 bool verbose = false;
-barrier barrier;
+barrier bo_barrier;
+barrier exec_barrier;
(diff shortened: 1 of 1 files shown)


maxzhen (approved):

dayeh-xilinx:

xilinxgitops:
retest this please

PR #6228, opened 2022-02-04
Add xrt::bo APIs for getting memory bank and flags

stsoe:
Per request provide access to the flags and memory group used when the
buffer object was constructed.

Accessors were already available in implementation, just expose these
in class xrt::bo.


dayeh-xilinx:

xilinxgitops:
retest this please

PR #6235, opened 2022-02-04
Pkg apu bootargs

houlz0507:
petalinux boot args should be removed from device tree
Use uboot script to override bootargs


chvamshi-xilinx (approved):

xilinxgitops:
retest this please

PR #6237, opened 2022-02-07
CR-1119807 - Fix writing data to fd from xclExportBO api

chvamshi-xilinx:
Fixing an issue in exprotBO handle for LPDDR memory. We were allocating only cma memory earlier. We are now allocating non-cma memory also for which we need to call different gem call to export buffer.

writing to fd using mmap exported from xclExportBO is not working in edge platforms

callback for gem_prime_mmap is set to drm_gem_prime_mmap which handles both cma and non cma memory

none

tested with dpu example and things are working as expected


chvamshi-xilinx:
Closing as more changes required..

PR #6238, opened 2022-02-07
CR-1120533 xgq: add application ID member to the XGQ cmd structure

rajkumar-xilinx:
XGQ new protocol introduced a new field called API_ID which
tells the VMR the actual request client is looking for.

Each API_ID value has fixed and dedicated API associated in VMR/VMC firmware.

By default, XGQ request for complete sensor data for any repo type request.


rajkumar-xilinx:
I will add code owners once @xdavidz approves this PR.

maxzhen (approved):

src/runtime_src/core/include/xgq_cmd_vmr.h
```diff
@@ -102,7 +112,8 @@ struct xgq_cmd_log_payload {
 	uint64_t address;
 	uint32_t size;
 	uint32_t offset;
-	uint32_t pid:16;
+	uint32_t aid:8;
+	uint32_t pid:8;
```
xdavidz, line 116: We cannot directly change this without considering other commands. Currently the log_page and sensor_data are sharing exact same command payload.

Given the sensor_data has it's own spec, we will have to have a new payload for sensor_data itself.

So, again the change process would be:
1) design the spec first: https://confluence.xilinx.com/pages/viewpage.action?pageId=302345261#XGQ1.0CommandandQueueSpecification(WIP)
2) change the code on host
3) sync the code on VMR

Let's discuss this with a meeting.

8bits seems good range to me. but what the pid will be used for?

xdavidz (changes requested):

src/runtime_src/core/include/xgq_cmd_vmr.h
rajkumar-xilinx: pid value will be one of the values of "enum xgq_cmd_sensor_page_id" defined in vmr.h file.
It specifies sensor request type (*GET_SIZE, *BDINFO etc.)

chvamshi-xilinx:
Hi @rajkumar-xilinx /@xdavidz . This PR has "do not merge" tag. Can we remove this tag and merge?

PR #6241, opened 2022-02-07
CR-1120790 ASTeR TS04 +07 - u50, u250 - xbutil validate fails during verify test while ERT is disabled

mamin506:
1. Fix CR-1120790, when ert=false, xclExecbuf failed
2. Add UUID into config CU command
3. Fix some coverity issues

1. When ert=false, xclExecbuf go check slot size, which is 0 if XGQ is used. Then it failed.
2. Need UUID in config CU command to let device be able to identify XCLBIN of the CU.

If ert=false, then don't config XGQ ERT and use KDS mode.

No see any risks.

xbutil validate on u50 and versal discovery shell.

No


maxzhen (approved):

jeffli-xilinx (approved):
Change for uuid looks good to me.

stsoe (approved):

PR #6244, opened 2022-02-08
Fix CR-1119807 - handle drm callbacks for non cma memory in edge platforms

rbramand-xilinx:
In edge platforms drm callbacks handle only cma memory, for non cma memory these callbacks are not working properly. This PR fixes some of those callbacks.

While handling [CR-1119807](https://jira.xilinx.com/browse/CR-1119807) this bug was discovered.

For CMA buffer we are appending extra flag to indicate memory is allocated in CMA region, checking that flag to determine between cma vs non-cma

Minimal testing is done, at present we don't have testcases to verify all the possible combinations

Tested locally with dpu application that uses mmap on exportBO and the test is passing with both cma and non-cma memory. Also used XRT pipeline to verify.

None

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_drv.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_drv.c
@@ -814,6 +814,56 @@ static int zocl_iommu_init(struct drm_zocl_dev *zdev,
 	return 0;
 }
 
+static int zocl_gem_prime_mmap(struct drm_gem_object *obj,
+		struct vm_area_struct *vma)
+{
+	struct drm_zocl_bo *zocl_obj;
+	zocl_obj = to_zocl_bo(obj);
+
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 12, 0)
+	return drm_gem_prime_mmap(obj, vma);
+#else
+	if (zocl_obj->flags & ZOCL_BO_FLAGS_CMA)
+		return drm_gem_cma_prime_mmap(obj, vma);
+	else
+		return drm_gem_prime_mmap(obj, vma);
+#endif
+}
+
+#if LINUX_VERSION_CODE < KERNEL_VERSION(5, 11, 0)
+void *zocl_gem_prime_vmap(struct drm_gem_object *obj)
+{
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/edge/drm/zocl/zocl_drv.c
```diff
@@ -814,6 +814,56 @@ static int zocl_iommu_init(struct drm_zocl_dev *zdev,
 	return 0;
 }
 
+static int zocl_gem_prime_mmap(struct drm_gem_object *obj,
+		struct vm_area_struct *vma)
+{
+	struct drm_zocl_bo *zocl_obj;
```
saifuddin-xilinx, line 820: Make it oneline like other place.
i.e. struct drm_zocl_bo *zocl_obj = to_zocl_bo(obj);

src/runtime_src/core/edge/drm/zocl/zocl_drv.c
```diff
@@ -814,6 +814,56 @@ static int zocl_iommu_init(struct drm_zocl_dev *zdev,
...
+
+	if (zocl_obj->flags & ZOCL_BO_FLAGS_CMA)
+		return drm_gem_cma_prime_vmap(obj);
+	else
+		return zocl_obj->vmapping;
+}
+
+void zocl_gem_prime_vunmap(struct drm_gem_object *obj, void *vaddr)
+{
+
```
saifuddin-xilinx, line 846: Add comments and add a return statement.

src/runtime_src/core/edge/drm/zocl/zocl_drv.c
```diff
@@ -814,6 +814,56 @@ static int zocl_iommu_init(struct drm_zocl_dev *zdev,
...
+		return drm_gem_cma_vmap(obj, map);
+	else {
+		dma_buf_map_set_vaddr(map, zocl_obj->vmapping);
+		return 0;
+	}
+}
+
+void zocl_gem_prime_vunmap(struct drm_gem_object *obj, struct dma_buf_map *map)
+{
+
```
saifuddin-xilinx, line 863: Add comments and add a return statement.

rbramand-xilinx:
closing this PR as we need to discuss more on how to handle non cma memory call backs

PR #6247, opened 2022-02-08
1119614 : Improve message for xbutil examine -r debug-ip-status when loaded bitstream has no debug ips

IshitaGhosh:
After overhaul Debug IP Status report generation, the following old message for absence/empty debug_ip_layout in the loaded bitstream has been removed 
INFO: Failed to find any Debug IP Layout section in the bitstream loaded on device. Ensure that a valid bitstream with debug IPs (AIM, LAPC) is successfully downloaded.

However, the message in the new flow can be improved.
Here is a sample message with the updates in this PR :
```
xbutil examine -d 0000:3b:00.1 -r debug-ip-status 

1/1 [0000:3b:00.1] : xilinx_u200_gen3x16_xdma_base_1

Debug IP Status 
  Number of IPs found: 0
```


AShivangi (changes requested):
Can we remove the "INFO" message and match it to the rest of the report styles? For eg:
```
$ xbutil examine -d 17:00 -r mechanical

1/1 [0000:17:00.1] : xilinx_u30_gen3x4_base_2
Mechanical
  Fans
    Not present
```

IshitaGhosh:
@AShivangi , I can remove the Info message. However, one question, the DebugIpStatusCollector has some other INFO messages written to std::cout. Shall I remove them too? @rozumx , you can add your feedback too.

AShivangi:
> @AShivangi , I can remove the Info message. However, one question, the DebugIpStatusCollector has some other INFO messages written to std::cout. Shall I remove them too? @rozumx , you can add your feedback too.

@IshitaGhosh, yes please remove any std::cout. All the outputs should be redirected to `output`

pgschuey (changes requested):
Changes look really good. One simple request: can we change the order in if statements where the constant is the second value, e.g., line 150 should be:
if (map.size() == 0) {
I see this issue on lines 150, 502, 793, and 1275

IshitaGhosh:
@AShivangi , with removal of the INFO messages written to std::cout, the output is now 
xbutil examine -d 0000:3b:00.1 -r debug-ip-status

1/1 [0000:3b:00.1] : xilinx_u200_gen3x16_xdma_base_1

Debug IP Status 
  Number of IPs found :: 0

Although it does not state the reason for "0" IPs like before, I think it is still fine.

AShivangi:
Yes, that is fine. Is there a reason you have double colons (::)?
Also, can you indent the information by 2 spaces?
```
...
Debug IP Status
  Number of IPs found: 0
```

IshitaGhosh:
@AShivangi The second line is already indented by two spaces. Somehow it does not show in the comment here.

AShivangi (approved):

pgschuey (approved):

jvillarre (approved):

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -495,12 +493,14 @@ DebugIpStatusCollector::readAccelDeadlockDetector(debug_ip_data* dbgIpInfo)
 }
 
 
-void 
+bool
```
rozumx, line 496: Why the introduction of a boolean return type.

I ask for when I look at the code, it "seems" that this value is used only do indicate that the `total_num_debug_ips` value is non-zero.  Instead of doing "double duty" here, it would be better if this count is extracted from the returned property tree, like you are doing else where in the code:
`uint64_t num_debug_ips = _dbgIpStatus_pt.get<uint64_t>("total_num_debug_ips");
`

src/runtime_src/core/tools/common/ReportDebugIpStatus.cpp
```diff
@@ -1262,9 +1267,10 @@ ReportDebugIpStatus::getPropertyTree20202( const xrt_core::device * _pDevice,
   auto handle = _pDevice->get_device_handle();
 
   DebugIpStatusCollector collector(handle, _pDevice);
-  collector.populateOverview(pt);
-  collector.collect();
-  collector.populateAllResults(pt);
+  if (true == collector.populateOverview(pt)) {
```
rozumx, line 1270:
```suggestion
  collector.populateOverview(pt)
  if (pt.get<uint64_t>("total_num_debug_ips",0) != 0) {
```

xilinxgitops:
retest this please.

PR #6250, opened 2022-02-08
Fixed the  issue for AIE seg fault

saifuddin-xilinx:
Canon facing AIE application issue as follows:
Segfault occurs when program finishes | Program runs and behaves as expected but segfaults on exit.

The issue is the order of the destructor. Right now we need to go through full AIE to fic this issue. But this is a workaround to avoid segmentation fault issues.


saifuddin-xilinx:
This is just a workaround for this specific test case. We will come up with a proper fix.

PR #6251, opened 2022-02-09
remove apu package metadata

houlz0507:
RPU crash while loading APU image


houlz0507:
It may be a RPU lscript issue. I will confirm with david first

PR #6253, opened 2022-02-09
CR-1120952: Fixing Crash where all banks are either unused or streaming

chvamshi-xilinx:
- Whenever all the banks are either unused or streaming, XRT user space is crashing with vector::_M_range_check error. 
- memory bank encoding size should be always equal to number of memories in xclbin.

Creating a vector of number of banks size

No

Verified CR and couple of other cases

No


stsoe (approved):
Thanks!

PR #6160, opened 2022-01-20
XRT support Multi-slot solution for edge platform

saifuddin-xilinx:
Story:
VITIS-4756 : Architecting XRT to support Multi-slot solution

Feature Specifications:
https://confluence.xilinx.com/display/SPEC/Feature+Specifications+for+Multi+Slot+Support

Problem Statement:
We want to be able to support multiple slots for Versal/zynqMP devices. Where more than one slot will be active concurrently.

Solution:
We have modified ZOCL driver to support the multiple slot. 
In the zocl Following modules are changed:
1. Memory Manager
2. XCLBIN Load 
3. CU scheduling 
4. KDS 
5. Client Context 

Test : 
As multi slot platform is not available we have tested this solution with single slot platform only. 
Test covered till now : 
- Versal Hw
- Versal hw_emul
- PCIe platform

PR #6255, opened 2022-02-10
Hot FIX: fix zocl crash when probe on versal device

mamin506:
On versal device, when zocl probe, kernel will crash.

#6160 introduce this issue. The of_property_read_u64() returns 0 on success. The bug code use taint pr_num to update num_pr_slot, which is used as the loop boundary. The loop will allocate memory and the system will crash when out of memory.

Correct the condition.

No seeing any risk.

On versal discovery shell.

No.


maxzhen (approved):

xilinxgitops:
retest this please.

saifuddin-xilinx (approved):

PR #6258, opened 2022-02-10
VITIS-3208 hwmon_sdm: add support xocl/userpf hwmon_sdm driver

rajkumar-xilinx:
Added changes to support & enable hwmon_sdm driver in userpf.
Disable XMC on vck5000-es3 (device_id 0x5049)
Added changes to create sysfs entries under hwmon directory for all eligible sensors.
Added new mailbox request to receive sensors from unprivileged access.
Device sensors are registered with hwmon driver and are visible under hwmon sysfs directory.


rajkumar-xilinx:
dmesg log:
[Feb11 06:12] xocl 0000:d8:00.1: mailbox.u.9437184 ffff9f108906e810 _mailbox_request: sending request: 16 via HW
[  +0.075898] xclmgmt 0000:d8:00.0: mailbox.m.9437184 ffff9f108d9d3410 process_request: received request from peer: 16, passed on
[  +0.000018] xclmgmt 0000:d8:00.0: xclmgmt_read_subdev_req: req kind 8
[  +0.000018] xclmgmt 0000:d8:00.0: mailbox.m.9437184 ffff9f108d9d3410 mailbox_post_response: posting response for: 16 via HW

**temp sensors sysfs nodes**: registered with hwmon driver
bash-4.2# ls /sys/bus/pci/devices/0000\:d8\:00.1/hwmon/hwmon5/
device  subsystem    temp1_label  temp2_label  temp3_label  temp4_label  temp5_label  temp6_label  temp7_label
power   temp1_input  temp2_input  temp3_input  temp4_input  temp5_input  temp6_input  temp7_input  uevent

`bash-4.2# cat hwmon5/temp*
69
Fan RPM
0
QSFP_TEMP1
0
QSFP_TEMP0
0
Sysmon Temp
0
Board Temp
33
Outlet Temp
30
Inlet Temp`

kbhardwa-xilinx:
retest this please

dayeh-xilinx:
retest this please.

chvamshi-xilinx (approved):

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
```diff
@@ -776,6 +776,26 @@ static int xclmgmt_read_subdev_req(struct xclmgmt_dev *lro, char *data_ptr, void
 		xclmgmt_subdev_get_data(lro, subdev_req->offset,
 			subdev_req->size, resp, &current_sz);
 		break;
+	case XCL_SDR_BDINFO:
```
houlz0507, line 779: Why can not reuse XCL_BDINFO?

rajkumar-xilinx: If XCL_BDINFO used, then need to add a check to redirect the request to either xmc or hwmon_sdm driver.
Please let me know if adding a check is better? Again this check will be like if hwmon_sdm exist ?

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
```diff
@@ -776,6 +776,26 @@ static int xclmgmt_read_subdev_req(struct xclmgmt_dev *lro, char *data_ptr, void
...
+		current_sz = 4 * 1024;
+		*resp = vzalloc(current_sz);
+		(void) xocl_hwmon_sdm_get_sensors(lro, *resp, XCL_SDR_BDINFO);
+		break;
+	case XCL_SDR_TEMP:
```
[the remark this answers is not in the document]
rajkumar-xilinx: Same like above comment. And, code tries to refresh/update whatever sensors are being read instead of refreshing all sensors.
So, If XLC_SENSOR is used then hwmon_sdm driver need to send all_sensor request to firmware for every sensor read. To avoid it, added these cases. If sensor read request falls in temperature group then hwmon_sdm driver request for temperature sensor data only.

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
houlz0507: Will xmc and hwmon_sdm co-exist? It seems not to me.

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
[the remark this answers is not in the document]
houlz0507: I though we have sensor cache in sdm driver? So in a time period, only the first request read all sensors and the following requests read from cache?

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
rajkumar-xilinx: ok, but how to know which callback to be called if I use XCL_BDINFO for both?

For xmc: we have xocl_xmc_get_sensors()

For hwmon_sdm: we have xocl_hwmon_sdm_get_sensors()

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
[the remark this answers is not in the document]
rajkumar-xilinx: But if cache time expired, then we have to update the sensors in cache right?
Driver updates only the requested sensor group instead of all sensors groups.

**FLow**


Example: sensors are categorized into 4 to 5 groups. Voltage, Current, Power etc.

During proble/init, driver issues all sensor read requests and store them in cache.

Later, If voltage sensor read request came, then driver checks if the cache time expired or not. 
- If expired then it issue xgq requests for voltage sensor group only.
- If not expired, driver returns cache data.

Similarly, for current and other sensors. 

If XCL_SENSOR only used, then hwmon_driver has to issues all_sensors read irrespective of sensor group.

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
houlz0507: I think there are 2 options.
1. use the same device id for  xmc and sdm. You may unify the callback table.
2. check if xocl_hwmon_sdm_get_sensors() returns -NODEV, then call xocl_xmc_get_sensors() if it is.

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
```diff
@@ -776,6 +776,26 @@ static int xclmgmt_read_subdev_req(struct xclmgmt_dev *lro, char *data_ptr, void
 		xclmgmt_subdev_get_data(lro, subdev_req->offset,
 			subdev_req->size, resp, &current_sz);
 		break;
+	case XCL_SDR_BDINFO:
+		current_sz = 4 * 1024;
```
maxzhen, line 780: Can we give '4' and '1024' a meaningful macro name?

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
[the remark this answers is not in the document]
houlz0507: ok. I am not sure if adding mailbox command for this is good. I will leave this question to @maxzhen

maxzhen (changes requested):
1. Did we test with new xclmgmt and old xocl? Also old xclmgmt and new xocl?
2. Now we have 1 sec cache expiration time in both xclmgmt and xocl. So, if we're unlucky, we may see maximum 2 seconds delay for fetching sensor values from xocl. I'm wondering if we should update the cache on mgmt side when the request is from xocl?

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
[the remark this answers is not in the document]
rajkumar-xilinx: Hi @houlz0507 ,

It is better to keep xmc & hwmon_sdm "enum xcl_group_kind" codes as separate for clean flow. Though the functionality of these two drivers are same, they are actually different in implementation & behavior.

Since XMC is used by multiple legacy and ultra scale platforms, i would like to not to use XMC's device ID and related xcl_group_kind codes also.

rajkumar-xilinx:
Hi @maxzhen ,

> 1. Did we test with new xclmgmt and old xocl? Also old xclmgmt and new xocl?
I have not tried this case yet, thanks for pointing it.

> 2. Now we have 1 sec cache expiration time in both xclmgmt and xocl. So, if we're unlucky, we may see maximum 2 seconds delay for fetching sensor values from xocl. I'm wondering if we should update the cache on mgmt side when the request is from xocl?

If the received request is from xocl, then hwmon_sdm driver will check cache timer. If timer expired then it will issue mailbox_request() and in turn it will raise new xgq requests (but not read from mgmtpf cache).

If the received request is from xclmgmt, then hwmon_sdm driver will check cache timer. If timer expired then it will issue new xgq requests. It timer not expired, then it returns mgmtpf cache data.

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
rajkumar-xilinx: i will try it and will update.

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
rajkumar-xilinx: i will update, thank you.

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
rajkumar-xilinx: Hi @houlz0507 , I have added your suggested approach for dealing with BDINFO.

src/runtime_src/core/pcie/driver/linux/xocl/devices.h
```diff
@@ -3605,7 +3611,7 @@ struct xocl_subdev_map {
 	{ XOCL_PCI_DEVID(0x10EE, 0x5031, PCI_ANY_ID, USER_SMARTN) },	\
 	{ XOCL_PCI_DEVID(0x10EE, 0x5029, PCI_ANY_ID, USER_XDMA_VERSAL) },\
 	{ XOCL_PCI_DEVID(0x10EE, 0x5045, PCI_ANY_ID, USER_XDMA_VERSAL) },\
-	{ XOCL_PCI_DEVID(0x10EE, 0x5049, PCI_ANY_ID, VERSAL_USER_RAPTOR2) }, \
+	{ XOCL_PCI_DEVID(0x10EE, 0x5049, PCI_ANY_ID, VERSAL_USER_RAPTOR2_ES3) }, \
```
rajkumar-xilinx, line 3614: Hi @xdavidz , could you please review this change once?

I am trying to disable XMC on vck5000-es3 (device id is 0x5049) .

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
maxzhen: @rajkumar-xilinx , could you confirm that the data format returned from xclmgmt for BDINFO is the same as before (XMC)?

rajkumar-xilinx: Hi @maxzhen , response data format is not same between XMC and hwmon_sdm driver.

XMC: it uses struct xcl_board_info as the response format.
hwmon_sdm: it receives raw data received from vmr. We are not defining xcl_board_info structure here, since it is data driven.

maxzhen: @rajkumar-xilinx , @houlz0507 , I think we should use different sub-op codes for responses from SDR driver (BDINFO, sensors, etc).

Mailbox protocol is b/w xclmgmt and xocl. They may not be always in sync (say in cloud use case). When we making changes in mailbox_proto.h, we're changing the protocol and we should maintain backward compatibility.

A protocol consists of data structure and op-codes. They tie to each other. You can add new op-code and new data structures or obsolete old op-codes and remove them over time. But you don't want to map same op-code to incompatible data structures, which may confuse the peer when the two sides are out of sync.

By "incompatible data structure", I mean that the data structure that can't be parsed properly by code before the change.

In this PR, the response data structure is not compatible with the one from XMC, so different op-code works better.

Let me know if you think otherwise...

rajkumar-xilinx: Hi @maxzhen thanks for suggestions. I have updated the changes to use separate & new opcodes for SDR driver.

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
maxzhen: Sorry that I was not clear. I mean we should define macros for these constant values so that we can get better readability and easy to maintain. Also, where do we check if the size of SDR data obtained does not go beyond this current_sz?

maxzhen:
> Hi @maxzhen ,
> 
> > 1. Did we test with new xclmgmt and old xocl? Also old xclmgmt and new xocl?
> >    I have not tried this case yet, thanks for pointing it.
> 
> > 2. Now we have 1 sec cache expiration time in both xclmgmt and xocl. So, if we're unlucky, we may see maximum 2 seconds delay for fetching sensor values from xocl. I'm wondering if we should update the cache on mgmt side when the request is from xocl?
> 
> If the received request is from xocl, then hwmon_sdm driver will check cache timer. If timer expired then it will issue mailbox_request() and in turn it will raise new xgq requests (but not read from mgmtpf cache).
> 
> If the received request is from xclmgmt, then hwmon_sdm driver will check cache timer. If timer expired then it will issue new xgq requests. It timer not expired, then it returns mgmtpf cache data.

OK, I see.
Please update this PR with compatibility test result, too.

src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
rajkumar-xilinx: Hi @maxzhen  Actually, there is a pending AI to get the size of the sensor data first and then issue the request for actual sensor data. So, we are just using 4k as size for now till we fix it.

rajkumar-xilinx:
Added new sysfs node under hwmon.

bash-4.2# cat /sys/bus/pci/devices/0000\:d8\:00.1/hwmon/hwmon4/name
xilinx_vck5000_gen4x8_xdma_base_1_hwmon_sdm_user
bash-4.2# cat /sys/bus/pci/devices/0000\:d8\:00.0/hwmon/hwmon3/name
xilinx_vck5000_gen4x8_xdma_base_1_hwmon_sdm_mgmt

rajkumar-xilinx:
retest this please

src/runtime_src/core/pcie/driver/linux/xocl/devices.h
```diff
@@ -2951,8 +2945,8 @@ struct xocl_subdev_map {
 		.flags = XOCL_DSAFLAG_DYNAMIC_IP |			\
 			XOCL_DSAFLAG_VERSAL_ES3 |			\
 			XOCL_DSAFLAG_VERSAL,				\
-		.subdev_info = RES_USER_VERSAL_VSEC_ES3,		\
-		.subdev_num = ARRAY_SIZE(RES_USER_VERSAL_VSEC_ES3),	\
+		.subdev_info = RES_USER_VERSAL_VSEC,		\
```
xdavidz: I don't understand this changes. The background of VSEC_ES3 was due to a dma bug in xdma that only 2 channels are working for es3 legacy shell. If we switch to regular versal vsec, which will enable 4 channels. Have you tested the dmatest on most recent es3 shell (gen3x16_xdma)?

xdavidz (changes requested):

xdavidz (approved):
no change in devices.h please

maxzhen (approved):

rajkumar-xilinx:
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>


<div style='direction:ltr'>



Test on VCK5000 | Old xocl   +   New xclmgmt | New xocl   +   Old xclmgmt
-- | -- | --
Loading Drivers | XMC is loaded on   xocl side only.       Not observed any   error messages in dmesg. | Not observed any   hangs during xocl driver loading   Observed timeout   messages for SDR request and which are expected. Since, old xclmgmt doesn't   have SDR request implemented.   Dmesg:   [  +0.000011] xocl 0000:d8:00.1:   mailbox.u.9437184 ffff8c0bcb8bf810 _mailbox_request: sending request: 16 via   HW   [ +20.097370] xocl   0000:d8:00.1: mailbox.u.9437184 ffff8c0bcb8bf810 timeout_msg: found waiting   msg time'd out   [  +0.000085] xocl 0000:d8:00.1:  ffff8c082f5ff098 xocl_hwmon_sdm_init_sysfs:   sdr peer request failed, err: -62
Read xmc sysfs   nodes | XMC requests are   passing, but response will be null only. Since, mgmt side there is no XMC   loaded.   Dmesg:   [Feb17 19:48] xocl   0000:d8:00.1: mailbox.u.9437184 ffff8c0bb142ac10 _mailbox_request: sending   request: 10 via HW   [  +0.036992] xclmgmt 0000:d8:00.0:   mailbox.m.9437184 ffff8c0bb16d0410 process_request: received request from   peer: 10, passed on   [  +0.000010] xclmgmt 0000:d8:00.0:   xclmgmt_read_subdev_req: req kind 2   [  +0.000010] xclmgmt 0000:d8:00.0:   mailbox.m.9437184 ffff8c0bb16d0410 mailbox_post_response: posting response   for: 10 via HW | XMC requests are   passing, but response will be null only. Since, mgmt side there is no XMC   loaded.   Dmesg:   [Feb17 20:29]   hwmon hwmon4: hwmon4 ffff8c0c196ca408 xmc_read_from_peer: reading from peer   [  +0.000018] xocl 0000:d8:00.1:   mailbox.u.9437184 ffff8c0c196cb010 _mailbox_request: sending request: 10 via   HW   [  +0.072375] xclmgmt 0000:d8:00.0:   mailbox.m.9437184 ffff8c0bcba94010 process_request: received request from   peer: 10, passed on   [  +0.000010] xclmgmt 0000:d8:00.0:   xclmgmt_read_subdev_req: req kind 0   [  +0.000011] xclmgmt 0000:d8:00.0:   mailbox.m.9437184 ffff8c0bcba94010 mailbox_post_response: posting response   for: 10 via HW



</div>


</body>

</html>

PR #6263, opened 2022-02-11
Move conditional into one-time executed block

stsoe:
Per observation by @herveratigner potentially save few instructions in xclRegRW

Error conditional resulting from MAP_FAILED should be executed once only.

Move check inside initialization block


maxzhen (approved):
@stsoe so we just avoid a comparing to zero, right? do you really see any difference?

stsoe:
> @stsoe so we just avoid a comparing to zero, right? do you really see any difference?

@maxzhen right there is no change otherwise. It's just that @herveratigner observed this when we made the original optimization a month or so ago.

PR #6266, opened 2022-02-11
CR-1121688: Regression: XRT error while loading xclbin (argidx mismatch)

chvamshi-xilinx:
 There is a sanity added in XRT to check whether argument index is same as running index. This doesnt work in all cases.
There is a use case where argument id can be same for multiple entries in xml file
argindex

Removed the Sanity

None

Verified the CR


chvamshi-xilinx:
Hi @stsoe , I am not sure whether this is the right fix or not. As this is blocking multiple X3 issues, I sent this PR so that it will go thru pipeline before you login.

stsoe (commented):
Absolutely not :-)

chvamshi-xilinx:
> Absolutely not :-)

Let me close this :-)

PR #6268, opened 2022-02-12
Fix cu xgq assign issue

mamin506:
Failed to assign CU XGQ issue.

#6160  introduce this issue. This is discovered by run xbutil validate on versal discovery shell.

Add a fake client_ctx when CU XGQ try to open context for CU.
This is a quick fix to make CU XGQ go through on Versal. 

No

xbutil validate on Versal discover shell.

No

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_kds.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_kds.c
@@ -203,7 +203,7 @@ zocl_create_client_context(struct drm_zocl_dev *zdev,
  *
  */
 static struct kds_client_ctx *
-zocl_check_exists_context(struct kds_client *client, uuid_t *id)
+zocl_check_exists_context(struct kds_client *client, const uuid_t *id)
 {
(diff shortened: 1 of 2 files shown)


maxzhen (approved):

src/runtime_src/core/edge/drm/zocl/zocl_kds.c
```diff
@@ -283,11 +283,26 @@ int zocl_add_context_kernel(struct drm_zocl_dev *zdev, void *client_hdl, u32 cu_
 	int ret = 0;
 	struct kds_ctx_info info = { 0 };
 	struct kds_client *client = (struct kds_client *)client_hdl;
+	struct kds_client_ctx *cctx = NULL;
+
```
saifuddin-xilinx, line 287: It's better to check whether any existing context is already present or not. 
Otherwise, it will end up with duplicate context. 
I am fine if this is a temporary fix. We can add the check later also.

saifuddin-xilinx (approved):

PR #6269, opened 2022-02-12
Workaround dead-lock in hardware-emulation shutdown

keryell:

Workaround dead-lock in hardware-emulation shutdown

Just running some simple SYCL code on xsjsycl41 server.

This is solved by an outrageous hack.

The user needs to define `XRT_PCIE_HW_EMU_FORCE_SHUTDOWN` environment variable before running an XRT program.

This is probably not to be merged but to be used as a branch by other people hitting this problem while there is a real fix.

This works for us™. :-)

### DIFF
--- a/src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
+++ b/src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/shim.cxx
@@ -1694,7 +1694,9 @@ uint32_t HwEmShim::getAddressSpace (uint32_t topology)
     }
 
     //bool simDontRun = xclemulation::config::getInstance()->isDontRun();
-    if(!mSimDontRun)
+    /* If XRT_PCIE_HW_EMU_FORCE_SHUTDOWN environment variable
+       is set, skip the dead-locking wait for the simulator ending */
+    if(!mSimDontRun && !std::getenv("XRT_PCIE_HW_EMU_FORCE_SHUTDOWN"))
       while (-1 == waitpid(0, &status, 0));
 
     if(( lWaveform == xclemulation::debug_mode::gui || lWaveform == xclemulation::debug_mode::batch || lWaveform == xclemulation::debug_mode::off)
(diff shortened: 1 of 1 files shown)


sonals:
ok to test

kbhardwa-xilinx:
retest this please

stsoe (commented):
While I understand the desire here, I think this is a hack around an underlying problem.
@akasat @venkatp-xilinx shouldn't you be addressing the hang itself?

stsoe:
retest this please - infra errors

keryell:
Rebased before https://github.com/Xilinx/XRT/pull/6297 to avoid a test not woking with PSkernel.

venkatp-xilinx:
> While I understand the desire here, I think this is a hack around an underlying problem. @akasat @venkatp-xilinx shouldn't you be addressing the hang itself?

Yes Soren, we have a CR for this, and we will handle it as part of the CR fix.

venkatp-xilinx:
Added hemantk-xilinx as reviewer. Basic issue is in the hw_emu model which is being worked upon the hw emulation team. The solution seems very hacky, we do not want to merge the change.

keryell:
> The solution seems very hacky, we do not want to merge the change.

I totally agree, as suggested in the PR description. ;-)

chvamshi-xilinx:
@keryell , Can we close this PR as this is open from long and no updates.

keryell:
Is there any alternative? Is this problem solved otherwise?

chvamshi-xilinx:
> Is there any alternative? Is this problem solved otherwise?

I am not sure whether this is solved or not. @venkatp-xilinx , Can you please take a look and confirm. @keryell , Can you please file a CR and inform @venkatp-xilinx .

venkatp-xilinx:
> Is there any alternative? Is this problem solved otherwise?

This is resolved now, hw_emu model corrected the double free memory corruption issue and with that we would not see this issue any more. As a general fix in driver, we are trying various ways to identity the sock aliveness (when device process abruptly crashes) and do a clear exit from the driver. So we actively working on it and will provide proper solution in master branch shortly.

stsoe:
@keryell Per @venkatp-xilinx  comment this PR is no longer needed?

stsoe:
I believe this has been addressed to some degree :-)

keryell:
> I believe this has been addressed to some degree :-)

Our CI might be wrong then. :-(
I have not seen the rewrite of all the XRT-emulation layer either... :-(

PR #6272, opened 2022-02-14
VITIS-1997 - Support AIE Frequency scaling from command line for Vers…

rbramand-xilinx:
> Added changes in xbutil utility to support AIE clock frequency scaling
> Users can use xbutil subcommand 'Advanced' to set/get AIE clock frequency
> Example outputs:
image
===============================================================================

image

===============================================================================

root@versal-rootfs-common-2022_1:/run/media/mmcblk0p1# xbutil advanced --aie-clock -g
WARNING: `partition` option is not provided, using default partition id value '1'
INFO: Clock frequency of AIE partition(1) is :  312.50 MHz

### DIFF
--- a/src/runtime_src/core/common/query_requests.h
+++ b/src/runtime_src/core/common/query_requests.h
@@ -258,6 +258,8 @@ enum class key_type
   spc_status,
   accel_deadlock_status,
   get_xclbin_data,
+  aie_get_freq,
+  aie_set_freq_req,
 
   boot_partition,
   flush_default_only,
@@ -797,6 +799,27 @@ struct aie_reg_read : request
 
   virtual boost::any
   get(const device*, const boost::any& row, const boost::any& col, const boost::any& reg) const = 0;
+};
+
+struct aie_get_freq : request
+{
+  using result_type = uint64_t;
+  using partition_id_type = uint32_t;
+  static const key_type key = key_type::aie_get_freq;
+
+  virtual boost::any
+  get(const device*, const boost::any& partition_id) const = 0;
+};
+
+struct aie_set_freq_req : request
+{
+  using result_type = bool;
+  using partition_id_type = uint32_t;
+  using freq_type = uint64_t;
+  static const key_type key = key_type::aie_set_freq_req;
+
+  virtual boost::any
+  get(const device*, const boost::any& partition_id, const boost::any& freq) const = 0;
 };
 
 struct graph_status : request
--- a/src/runtime_src/core/common/query_requests.h
+++ b/src/runtime_src/core/common/query_requests.h
@@ -259,7 +259,7 @@ enum class key_type
   accel_deadlock_status,
   get_xclbin_data,
   aie_get_freq,
-  aie_set_freq_req,
+  aie_set_freq,
 
   boot_partition,
   flush_default_only,
@@ -811,12 +811,12 @@ struct aie_get_freq : request
   get(const device*, const boost::any& partition_id) const = 0;
 };
 
-struct aie_set_freq_req : request
+struct aie_set_freq : request
 {
   using result_type = bool;
   using partition_id_type = uint32_t;
   using freq_type = uint64_t;
-  static const key_type key = key_type::aie_set_freq_req;
+  static const key_type key = key_type::aie_set_freq;
 
   virtual boost::any
   get(const device*, const boost::any& partition_id, const boost::any& freq) const = 0;
--- a/src/runtime_src/core/edge/drm/zocl/zocl_aie.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_aie.c
@@ -464,23 +464,21 @@ int zocl_aie_freqscale(struct drm_zocl_dev *zdev, void *data)
 	if(!args->dir) {
 		// Read frequency from requested aie partition
 		ret = aie_partition_get_freq(zdev->aie->aie_dev, &args->freq);
+		mutex_unlock(&zdev->aie_lock);
 		if(ret)
 			DRM_ERROR("Reading clock frequency from AIE partition(%d) failed with error %d\n",
 				args->partition_id, ret);
-		mutex_unlock(&zdev->aie_lock);
 		return ret;
 	} else {
 		// Send Set frequency request for aie partition
 		ret = aie_partition_set_freq_req(zdev->aie->aie_dev, args->freq);
+		mutex_unlock(&zdev->aie_lock);
 		if(ret)
 			DRM_ERROR("Setting clock frequency for AIE partition(%d) failed with error %d\n",
 				args->partition_id, ret);
-		mutex_unlock(&zdev->aie_lock);
 		return ret;
 
 	}
-
-	return 0;
 }
 
 int
--- a/src/runtime_src/core/edge/drm/zocl/zocl_aie.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_aie.c
@@ -465,7 +465,7 @@ int zocl_aie_freqscale(struct drm_zocl_dev *zdev, void *data)
 		// Read frequency from requested aie partition
 		ret = aie_partition_get_freq(zdev->aie->aie_dev, &args->freq);
 		if(ret)
-			DRM_ERROR("Reading frequency from AIE partition %d failed with error %d\n",
+			DRM_ERROR("Reading clock frequency from AIE partition(%d) failed with error %d\n",
 				args->partition_id, ret);
 		mutex_unlock(&zdev->aie_lock);
 		return ret;
@@ -473,7 +473,7 @@ int zocl_aie_freqscale(struct drm_zocl_dev *zdev, void *data)
 		// Send Set frequency request for aie partition
 		ret = aie_partition_set_freq_req(zdev->aie->aie_dev, args->freq);
 		if(ret)
-			DRM_ERROR("Setting frequency request for AIE partition %d failed with error %d\n",
+			DRM_ERROR("Setting clock frequency for AIE partition(%d) failed with error %d\n",
 				args->partition_id, ret);
 		mutex_unlock(&zdev->aie_lock);
 		return ret;
--- a/src/runtime_src/core/edge/drm/zocl/zocl_aie.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_aie.c
@@ -435,6 +435,54 @@ zocl_aie_reset(struct drm_zocl_dev *zdev)
 	return 0;
 }
 
+int zocl_aie_freqscale(struct drm_zocl_dev *zdev, void *data)
+{
+	struct drm_zocl_aie_freq_scale *args= data;
+	int ret = 0;
+
+	mutex_lock(&zdev->aie_lock);
+
+	if (!zdev->aie) {
+		mutex_unlock(&zdev->aie_lock);
+		DRM_ERROR("AIE image is not loaded.\n");
+		return -ENODEV;
+	}
+
+	if (!zdev->aie->aie_dev) {
+		mutex_unlock(&zdev->aie_lock);
+		DRM_ERROR("No available AIE partition.\n");
+		return -ENODEV;
+	}
+
+	if (zdev->aie->partition_id != args->partition_id) {
+		mutex_unlock(&zdev->aie_lock);
+		DRM_ERROR("AIE partition %d does not exist.\n",
+		    args->partition_id);
+		return -ENODEV;
+	}
+
+	if(!args->dir) {
+		// Read frequency from requested aie partition
+		ret = aie_partition_get_freq(zdev->aie->aie_dev, &args->freq);
+		if(ret)
+			DRM_ERROR("Reading frequency from AIE partition %d failed with error %d\n",
+				args->partition_id, ret);
+		mutex_unlock(&zdev->aie_lock);
+		return ret;
+	} else {
+		// Send Set frequency request for aie partition
+		ret = aie_partition_set_freq_req(zdev->aie->aie_dev, args->freq);
+		if(ret)
+			DRM_ERROR("Setting frequency request for AIE partition %d failed with error %d\n",
+				args->partition_id, ret);
+		mutex_unlock(&zdev->aie_lock);
+		return ret;
+
+	}
+
+	return 0;
+}
+
 int
 zocl_aie_graph_alloc_context(struct drm_zocl_dev *zdev, u32 gid, u32 ctx_code,
 	struct sched_client_ctx *client)
--- a/src/runtime_src/core/edge/drm/zocl/zocl_aie.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_aie.c
@@ -3,7 +3,7 @@
  * A GEM style (optionally CMA backed) device manager for ZynQ based
  * OpenCL accelerators.
  *
- * Copyright (C) 2020-2021 Xilinx, Inc. All rights reserved.
+ * Copyright (C) 2020-2022 Xilinx, Inc. All rights reserved.
  *
  * Authors:
  *    Larry Liu <yliu@xilinc.com>
--- a/src/runtime_src/core/edge/user/device_linux.cpp
+++ b/src/runtime_src/core/edge/user/device_linux.cpp
@@ -557,9 +557,9 @@ struct aie_get_freq
   }
 };
 
-struct aie_set_freq_req
+struct aie_set_freq
 {
-  using result_type = query::aie_set_freq_req::result_type;
+  using result_type = query::aie_set_freq::result_type;
 
   static result_type
   get(const xrt_core::device* device, key_type key, const boost::any& partition_id, const boost::any& freq)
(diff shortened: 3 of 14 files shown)


src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -456,6 +456,82 @@ struct aie_reg_read
...
+    // Reading the aie_metadata sysfs.
+    dev->sysfs_get(AIE_TAG, err, value);
+    if (!err.empty())
+      throw xrt_core::query::sysfs_error
+      (err + ", The loading xclbin acceleration image doesn't use the Artificial "
+       + "Intelligent Engines (AIE). No action will be performed.");
+
+    int mKernelFD = open(ZOCL_DEVICE.c_str(), O_RDWR);
+    if (!mKernelFD)
+      throw xrt_core::error(-EINVAL, boost::str(boost::format("Cannot open %s") % ZOCL_DEVICE));
```
stsoe, line 521: This looks like code duplication, please find a way to share in separate function.

AShivangi:
@rbramand-xilinx, in the past we had clock frequency scaling in xbmgmt. Can you please help me understand why this needs to be done through xbutil and not xbmgmt?
We should also look into better/optimized naming of the option. For eg: `xbutil advanced --aie-freq --set-freq --freq <>` has `freq` 3 times. IMO, it should be something like `xbutil advanced --clock --aie --set <>`

kbhardwa-xilinx:
retest this please

rbramand-xilinx:
> @rbramand-xilinx, in the past we had clock frequency scaling in xbmgmt. Can you please help me understand why this needs to be done through xbutil and not xbmgmt? We should also look into better/optimized naming of the option. For eg: `xbutil advanced --aie-freq --set-freq --freq <>` has `freq` 3 times. IMO, it should be something like `xbutil advanced --clock --aie --set <>`

@AShivangi for edge platforms xbmgmt utility is not available so adding this feature to xbutil.  I will make changes in option names.

rozumx (commented):
Rahul, the use case around getting and setting these frequencies  looks strange.   For example, it looks like the user needs to:
1. Preset the desired frequency to use: --freq 100M
2. Set the frequency: --set-freq-reg 

Is there a reason why this is being done with two separate options, instead of one.  For example:
--set-frequency 100M

Could we have a meeting to discuss the use-cases around this functionality.

src/runtime_src/core/tools/xbutil2/OO_AieClockFreq.cpp
```diff
@@ -0,0 +1,149 @@
...
+		//convert freq to hertz(Hz)
+                freq = XBUtilities::string_to_bytes(m_freq);
+            }
+            catch(const xrt_core::error&) {
+                std::cerr << "Freq value provided with `set` option is invalid. Please specify proper units and rerun" << std::endl;
+                std::cerr << "eg: 'B', 'K', 'M', 'G' " << std::endl;
+                throw xrt_core::error(std::errc::operation_canceled);
+            }
+
+            xrt_core::device_query<qr::aie_set_freq_req>(device, m_partition_id, freq) ?
```
rozumx, line 136: Hmm...it might be beneficial to the user if the following code was added:
1. Print out the clock frequency before the change.
2. Change the clock frequency.
3. Validate the clock frequency is running the the value being set.

src/runtime_src/core/tools/xbutil2/OO_AieClockFreq.cpp
```diff
@@ -0,0 +1,149 @@
...
+                std::cerr << "eg: 'B', 'K', 'M', 'G' " << std::endl;
+                throw xrt_core::error(std::errc::operation_canceled);
+            }
+
+            xrt_core::device_query<qr::aie_set_freq_req>(device, m_partition_id, freq) ? 
+            std::cout << boost::format("INFO: Frequency request for aie partition_id %d is submitted successfully\n") %  m_partition_id :
+            std::cout << boost::format("INFO: Frequency request submission for aie partition_id %d failed\n") %  m_partition_id;
+        }
+    } catch (const std::exception& e){
+      std::cerr << boost::format("ERROR: %s\n") % e.what();
```
rozumx, line 141: A better error could be produced here.   Something like,
`ERROR: Setting the AIE partition(1) clock frequency to 100M Hz failed.`
followed by the exception message.

src/runtime_src/core/tools/xbutil2/OO_AieClockFreq.cpp
```diff
@@ -0,0 +1,149 @@
...
+      errorOccured = true;
+    }
+  }
+
+  if (errorOccured)
```
rozumx, line 146: If there is only one device, then this error check can be removed.

src/runtime_src/core/tools/xbutil2/SubCmdAdvanced.cpp
```diff
@@ -78,6 +79,7 @@ SubCmdAdvanced::execute(const SubCmdOptions& _options) const
 // Only defined for embedded platform
 #ifndef ENABLE_NATIVE_SUBCMDS_AND_REPORTS
   subOptionOptions.emplace_back(std::make_shared<OO_AieRegRead>("read-aie-reg"));
```
rozumx, line 81: I would be better if the `read-aie-reg` option be changed to `aie-reg` with a `read `option.   Doing so allows for this operation to be scaled in the future.   

Note: Being that this is in the ADVANCE command, we can change these options without having to consult with the PRT.

rozumx (changes requested):

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -456,6 +470,75 @@ struct aie_reg_read
...
+  get(const xrt_core::device* device, key_type key, const boost::any& partition_id)
+  {
+    result_type freq = 0;
+#if defined(XRT_ENABLE_AIE) && !defined(__AIESIM__)
+    const static std::string ZOCL_DEVICE = "/dev/dri/renderD128";
+    auto fd_obj = aie_get_drmfd(device, ZOCL_DEVICE);
+    if (fd_obj->fd < 0)
+      throw xrt_core::error(-EINVAL, boost::str(boost::format("Cannot open %s") % ZOCL_DEVICE));
+
+    uint32_t part_id = boost::any_cast<uint32_t>(partition_id);
```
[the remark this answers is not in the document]
rbramand-xilinx: query infra currently uses boost::any to hold values casting it using std::any_cast is not working, so using boost::any_cast for now

src/runtime_src/core/tools/xbutil2/SubCmdAdvanced.cpp
rbramand-xilinx: @rozumx I will try to address this in separate PR as this requires document change as well

src/runtime_src/core/edge/user/device_linux.cpp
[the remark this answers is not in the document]
stsoe: Just FYI @rozumx.  Yes, we want to move away from boost::any, boost::filesystem, etc.  The only change made recently was to make sure that public APIs (end-user APIs) do not have compile-time boost dependency.  The public APis moved to std::any while implementation in XRT was unchanged and continues to use boost::any.  We need to schedule work to remove all boost dependencies that can be replaced with std:: functionality

src/runtime_src/core/tools/xbutil2/OO_AieClockFreq.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+  // Exit if neither action or device specified
+  if(m_help || m_device.empty()) {
+    printHelp();
+    return;
+  }
+
+  // Check if set/get is used
+  if(!m_get && m_freq.length() == 0) {
+    std::cerr << "ERROR: Neither `set` nor `get' is used" << std::endl;
```
rozumx, line 109: The message could be a little bit more terse.  Something like:
`ERROR: Missing either a 'set' or 'get' option.`
Then print the help.

src/runtime_src/core/tools/xbutil2/OO_AieClockFreq.cpp
```diff
@@ -0,0 +1,187 @@
...
+OO_AieClockFreq::OO_AieClockFreq( const std::string &_longName, bool _isHidden )
+    : OptionOptions(_longName, _isHidden, "AIE clock frequency operations" )
+    , m_device("")
+    , m_partition_id(1)
+    , m_get(false)
+    , m_freq("")
+    , m_help(false)
+{
+  m_optionsDescription.add_options()
+    ("device,d", po::value<decltype(m_device)>(&m_device)->multitoken(), "The Bus:Device.Function (e.g., 0000:d8:00.0) device of interest")
```
rozumx, line 64: We now only support 1 device.  Please update the option to reflex this (e.g, remove the multitoken).

src/runtime_src/core/tools/xbutil2/OO_AieClockFreq.cpp
```diff
@@ -0,0 +1,187 @@
...
+    }
+  }
+  catch (const std::exception& e){
+    std::cerr << boost::format("ERROR: %s\n") % e.what();
+    throw xrt_core::error(std::errc::operation_canceled);
+  }
+
+  // Do operations on the device collected
+  auto device = deviceCollection.front();
+  if(m_get) {
```
rozumx, line 151: The get option should be its own function.

It is more for encapsulation.

src/runtime_src/core/tools/xbutil2/OO_AieClockFreq.cpp
```diff
@@ -0,0 +1,187 @@
...
+    double freq_part = get_aie_part_freq(device, m_partition_id);
+    std::cout << boost::format("INFO: Clock frequency of AIE partition(%d) is: %.2f MHz\n") % m_partition_id % freq_part ;
+    return;
+  }
+
+  if(m_freq.length() > 0) {
```
rozumx, line 157: The set option should be its own function.

src/runtime_src/core/tools/xbutil2/OO_AieClockFreq.cpp
```diff
@@ -0,0 +1,187 @@
...
+  return freq;
+}
+
+// ----- C L A S S   M E T H O D S -------------------------------------------
+OO_AieClockFreq::OO_AieClockFreq( const std::string &_longName, bool _isHidden )
+    : OptionOptions(_longName, _isHidden, "AIE clock frequency operations" )
+    , m_device("")
+    , m_partition_id(1)
+    , m_get(false)
+    , m_freq("")
```
rozumx, line 60:
```suggestion
    , m_setFreq("")
```
It just makes the code easier to read for the next person.

rozumx (changes requested):
Few requests.

salindac:
retest this please after fixing emulation tql

rozumx (approved):

chvamshi-xilinx (approved):

src/runtime_src/core/common/query_requests.h
```diff
@@ -795,6 +797,27 @@ struct aie_reg_read : request
...
+{
+  using result_type = uint64_t;
+  using partition_id_type = uint32_t;
+  static const key_type key = key_type::aie_get_freq;
+
+  virtual boost::any
+  get(const device*, const boost::any& partition_id) const = 0;
+};
+
+struct aie_set_freq_req : request
```
AShivangi, line 810:
```suggestion
struct aie_set_freq : request
```

src/runtime_src/core/tools/xbutil2/OO_AieClockFreq.cpp
```diff
@@ -0,0 +1,191 @@
...
+#include <boost/format.hpp>
+#include <boost/program_options.hpp>
+
+namespace po = boost::program_options;
+namespace qr = xrt_core::query;
+
+// System - Include Files
+#include <iostream>
+
+#define to_mega(x) x/1000000
```
AShivangi, line 37: why? please convert this into a lambda or a static function

src/runtime_src/core/tools/xbutil2/OO_AieClockFreq.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+// ----- H E L P E R M E T H O D S ------------------------------------------
+static double get_aie_part_freq(const std::shared_ptr<xrt_core::device>& device, uint32_t part_id)
```
AShivangi, line 40: For consistency and readability purpose, I would suggest the following for all functions in this file
```suggestion
static double 
get_aie_part_freq(const std::shared_ptr<xrt_core::device>& device, uint32_t part_id)
```

src/runtime_src/core/tools/xbutil2/OO_AieClockFreq.cpp
```diff
@@ -0,0 +1,191 @@
...
+{
+  double freq = 0;
+  try {
+    freq = static_cast<double>(to_mega(xrt_core::device_query<qr::aie_get_freq>(device, part_id)));
```
AShivangi, line 44: what does mega mean? MB?

src/runtime_src/core/tools/xbutil2/OO_AieClockFreq.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+  // Try to set frequency
+  try {
+    bool status = xrt_core::device_query<qr::aie_set_freq_req>(device, part_id, freq);
+    if(status) {
+      std::cout << boost::format("INFO: Setting clock freq of AIE partition(%d) is successful\n") %  part_id;
+      std::cout << boost::format("Running clock freq of AIE partition(%d) is: %.2f MHz\n") % part_id % get_aie_part_freq(device, part_id);
+    }
+    else
+      std::cout << boost::format("INFO: Setting clock freq of AIE partition(%d) to %s failed\n") %  part_id % setFreq;
```
AShivangi, line 78: Should this be info or error?

src/runtime_src/core/tools/xbutil2/OO_AieClockFreq.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+  // Do operations on the device collected
+  auto device = deviceCollection.front();
+  if(m_get) {
+    double freq_part = get_aie_part_freq(device, m_partition_id);
+    std::cout << boost::format("INFO: Clock frequency of AIE partition(%d) is: %.2f MHz\n") % m_partition_id % freq_part ;
+    return;
+  }
+
+  if(m_setFreq.length() > 0)
```
AShivangi, line 189:
```suggestion
  if(!m_setFreq.empty())
```

AShivangi (changes requested):
The changes look pretty good. Can you please update the PR description as well to reflect the new interface?

src/runtime_src/core/tools/xbutil2/OO_AieClockFreq.cpp
```diff
@@ -0,0 +1,199 @@
...
+
+  // Try to set frequency
+  try {
+    bool status = xrt_core::device_query<qr::aie_set_freq>(device, part_id, freq);
+    if(status) {
+      std::cout << boost::format("INFO: Setting clock freq of AIE partition(%d) is successful\n") %  part_id;
+      std::cout << boost::format("Running clock freq of AIE partition(%d) is: %.2f MHz\n") % part_id % get_aie_part_freq(device, part_id);
+    }
+    else
+      std::cout << boost::format("ERROR: Setting clock freq of AIE partition(%d) to %s failed\n") %  part_id % setFreq;
```
AShivangi, line 86: We should throw errc operation canceled, right?

AShivangi (approved):
Looks good! please do address one more comment I made about the errc tho

chvamshi-xilinx:
retest this please

larry9523 (commented):
Should zocl driver 
1. read the freq up front
2. validate the freq before calling into aie driver to change it?

rbramand-xilinx:
> Should zocl driver
> 
> 1. read the freq up front
> 2. validate the freq before calling into aie driver to change it?

Hi @larry9523, I dint understand, what do you mean by validating frequency here?

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -494,6 +511,80 @@ struct aie_reg_read
   }
 };
 
+static std::unique_ptr<drm_fd> aie_get_drmfd(const xrt_core::device* device, std::string dev_path)
```
stsoe, line 514: Newline before function name; it's close to impossible to scan files that don't line up functions on the left.
const std::string&

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -55,6 +56,22 @@ using key_type = query::key_type;
 xclDeviceHandle handle;
 xclDeviceInfo2 deviceInfo;
 
+struct drm_fd
+{
+  int fd;
+  drm_fd(const std::string file_path, int flags)
```
stsoe, line 62: `const std::string& `  Always pass 'complex' objects by const reference
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-in

stsoe (changes requested):
Sorry for the late response.  Just few minor items in one file.
Thanks for making all the other changes, looks much better now.

src/runtime_src/core/edge/drm/zocl/zocl_aie.c
```diff
@@ -435,6 +435,54 @@ zocl_aie_reset(struct drm_zocl_dev *zdev)
...
+		ret = aie_partition_set_freq_req(zdev->aie->aie_dev, args->freq);
+		if(ret)
+			DRM_ERROR("Setting clock frequency for AIE partition(%d) failed with error %d\n",
+				args->partition_id, ret);
+		mutex_unlock(&zdev->aie_lock);
+		return ret;
+
+	}
+
+	return 0;
```
chvamshi-xilinx, line 483: we never reach here.. better to club if and else return stuff together.

stsoe (approved):

larry9523 (approved):

PR #6275, opened 2022-02-15
CR-1120510 : AIE report produces non-UTF-8 characters in error_msg

saifuddin-xilinx:
Issue : 
While generating aie_status_edge.json which comes from the XRT device info reports under the hood. The third line of the file contains an error_msg string. However, this string contains characters that are not UTF-8 and therefore, parsing this file throws an exception.

CR:
https://jira.xilinx.com/browse/CR-1120510

Added a destructor as "zynq_device" class and reset the sysfs file name. 
Without that sysfs file is holding some garbage value and the same is reflected on the report.

### DIFF
--- a/src/runtime_src/core/edge/user/zynq_dev.cpp
+++ b/src/runtime_src/core/edge/user/zynq_dev.cpp
@@ -151,3 +151,8 @@ zynq_device *zynq_device::get_dev()
 zynq_device::zynq_device(const std::string& root) : sysfs_root(root)
 {
 }
+
+zynq_device::~zynq_device()
+{
+	sysfs_root = "";
+}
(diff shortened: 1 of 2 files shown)


dayeh-xilinx:
retest this please.

keryell:
While I understand that this might fix a bug, you initialize a member variable in the destructor, and just after the call to the destructor, the variable member will be destructed anyway so this initialization should be useless.
Except if the bug is somewhere else and someone has captured the address of this zombie string and is using it later...
I am also curious about why XRT should produce UTF-8 at the first place. Is there any UTF-8 support at all?

stsoe (changes requested):
This change does not make any sense.  Why would sysfs_root be accessed after the zynq_device has been closed?

salindac:
retest please

chvamshi-xilinx:
Same here. Please close the PR and check who is holding the device handle after close. I assume, this must be from XDP somewhere.

PR #6276, opened 2022-02-15
Use new xbutil commands to enable host memory

mariodruiz:

Update the documentation to use new `xbutil` commands 

Documentation was referring to legacy tools

Updated commands to new switches

None


stsoe:
@mariodruiz The source for the edited page is actually this one: https://github.com/Xilinx/XRT/blob/master/src/runtime_src/doc/toc/hm.rst

@uday610 FYI

stsoe (changes requested):
Per comment

mariodruiz:
I knew I was doing something wrong. I'll open a new PR, I somehow broke the history

PR #6277, opened 2022-02-15
CR-1090469 ASTeR TC 21.03 - Improve error message for p2p

dbenusov-xilinx:
https://jira.xilinx.com/browse/CR-1090469
When sudo xbutil configure --p2p enable -d <BDF> is ran on an unsupported platform, the error messaging is unclear.  This is also problematic for automated testing parsers the expected string can move around.

Change exception handling for p2p errors to shorten the output message. This was modeled after the XBUtilities::sudo_or_throw function which only prints out an error message and the operation cancelled message.

It may be more prudent to set the catch for the xrt_core::error to something more specific, something we did not intend to get caught may end up in there. Comment below with suggestions! Perhaps using a more specific type of error?

Manually tested on U30 (No P2P) and U280 (P2P present)

U30 Testing:
```
root@xsjsagarw50:/proj/xsjhdstaff6/dbenusov/XRT# xbutil examine
System Configuration
  OS Name              : Linux
  Release              : 5.4.0-96-generic
  Version              : #109-Ubuntu SMP Wed Jan 12 16:49:16 UTC 2022
  Machine              : x86_64
  CPU Cores            : 16
  Memory               : 15671 MB
  Distribution         : Ubuntu 20.04.1 LTS
  GLIBC                : 2.31
  Model                : Super Server

XRT
  Version              : 2.13.0
  Branch               : CR-1090469
  Hash                 : 9309f5e1ca602dfdc79be5f17d5afb4cb25e5d31
  Hash Date            : 2022-02-15 14:01:47
  XOCL                 : 2.13.0, a1dfb3b2a3efe42fb0c59fdb9afe692403787561
  XCLMGMT              : 2.13.0, a1dfb3b2a3efe42fb0c59fdb9afe692403787561

Devices present
  [0000:b3:00.1] : xilinx_u200_gen3x16_xdma_base_1 user(inst=130)
  [0000:18:00.1] : xilinx_u30_gen3x4_base_2 user(inst=129)
  [0000:17:00.1] : xilinx_u30_gen3x4_base_2 user(inst=128)

root@xsjsagarw50:/proj/xsjhdstaff6/dbenusov/XRT# xbutil configure --p2p enable -d 17:00
ERROR: Config P2P failed : P2P is not supported: Operation canceled
root@xsjsagarw50:/proj/xsjhdstaff6/dbenusov/XRT# echo $?
1
root@xsjsagarw50:/proj/xsjhdstaff6/dbenusov/XRT# xbutil configure --p2p disable -d 17:00
ERROR: Config P2P failed : P2P is not supported: Operation canceled
root@xsjsagarw50:/proj/xsjhdstaff6/dbenusov/XRT# echo $?
1
```

U280 Testing
```
root@xsjpranjal01:/proj/xsjhdstaff6/dbenusov/XRT# xbutil examine
System Configuration
  OS Name              : Linux
  Release              : 5.4.0-97-generic
  Version              : #110-Ubuntu SMP Thu Jan 13 18:22:13 UTC 2022
  Machine              : x86_64
  CPU Cores            : 16
  Memory               : 15695 MB
  Distribution         : Ubuntu 20.04 LTS
  GLIBC                : 2.31
  Model                : Precision 5820 Tower

XRT
  Version              : 2.13.0
  Branch               : CR-1090469
  Hash                 : 9309f5e1ca602dfdc79be5f17d5afb4cb25e5d31
  Hash Date            : 2022-02-15 14:01:47
  XOCL                 : 2.13.376, 9309f5e1ca602dfdc79be5f17d5afb4cb25e5d31
  XCLMGMT              : 2.13.376, 9309f5e1ca602dfdc79be5f17d5afb4cb25e5d31

Devices present
  [0000:65:00.1] : xilinx_u280_xdma_201920_3 user(inst=130)
  [0000:17:00.1] : xilinx_u250_gen3x16_base_3 user(inst=128)

root@xsjpranjal01:/proj/xsjhdstaff6/dbenusov/XRT# xbutil configure --p2p enable -d 65:00
Please WARM reboot the machine to enable P2P now.
root@xsjpranjal01:/proj/xsjhdstaff6/dbenusov/XRT# xbutil configure --p2p disable -d 65:00
Please WARM reboot the machine to disable P2P now.
```

### DIFF
--- a/src/runtime_src/core/common/ishim.h
+++ b/src/runtime_src/core/common/ishim.h
@@ -431,15 +431,30 @@ struct shim : public DeviceType
   void
   p2p_enable(bool force) override
   {
-    if (auto ret = xclP2pEnable(DeviceType::get_device_handle(), true, force))
-      throw system_error(ret, "failed to enable p2p");
+    auto ret = xclP2pEnable(DeviceType::get_device_handle(), true, force);
+    switch (ret) {
+      case -EINVAL:
+      case -ENODEV:
+        throw system_error(ret, "failed to enable p2p");
+        break;
+      default:
+        break;
+    }
   }
+      
 
   void
   p2p_disable(bool force) override
   {
-    if (auto ret = xclP2pEnable(DeviceType::get_device_handle(), false, force))
-      throw system_error(ret, "failed to disable p2p");
+    auto ret = xclP2pEnable(DeviceType::get_device_handle(), false, force);
+    switch (ret) {
+      case -EINVAL:
+      case -ENODEV:
+        throw system_error(ret, "failed to disable p2p");
+        break;
+      default:
+        break;
+    }
   }
 
   void
--- a/src/runtime_src/core/pcie/linux/shim.cpp
+++ b/src/runtime_src/core/pcie/linux/shim.cpp
@@ -1147,7 +1147,8 @@ int shim::p2pEnable(bool enable, bool force)
         mDev->sysfs_put("p2p", "p2p_enable", err, "0");
     }
     if (!err.empty()) {
-        throw std::runtime_error("P2P is not supported");
+        std::cerr << "ERROR: Config P2P failed : P2P is not supported" << std::endl;
+        throw xrt_core::error(std::errc::operation_canceled);
     }
 
     if (force) {
@@ -2664,6 +2665,9 @@ int xclP2pEnable(xclDeviceHandle handle, bool enable, bool force)
   try {
     xocl::shim *drv = xocl::shim::handleCheck(handle);
     return drv ? drv->p2pEnable(enable, force) : -ENODEV;
+  }
+  catch(const xrt_core::error&) {
+    throw xrt_core::error(std::errc::operation_canceled);
   }
   catch (const std::exception& ex) {
     xrt_core::send_exception_message(ex.what());
--- a/src/runtime_src/core/pcie/linux/shim.cpp
+++ b/src/runtime_src/core/pcie/linux/shim.cpp
@@ -1134,22 +1134,19 @@ int shim::p2pEnable(bool enable, bool force)
         return -EINVAL;
 
     ret = check_p2p_config(mDev, err);
-    if (ret == P2P_CONFIG_ENABLED && enable) {
+    if (ret == P2P_CONFIG_ENABLED && enable)
         throw std::runtime_error("P2P is already enabled");
-    } else if (ret == P2P_CONFIG_DISABLED && !enable) {
+    else if (ret == P2P_CONFIG_DISABLED && !enable)
         throw std::runtime_error("P2P is already disabled");
-    }
 
     /* write 0 to config for default bar size */
-    if (enable) {
+    if (enable)
         mDev->sysfs_put("p2p", "p2p_enable", err, "1");
-    } else {
+    else
         mDev->sysfs_put("p2p", "p2p_enable", err, "0");
-    }
-    if (!err.empty()) {
-        std::cerr << "ERROR: Config P2P failed : P2P is not supported" << std::endl;
-        throw xrt_core::error(std::errc::operation_canceled);
-    }
+
+    if (!err.empty())
+        throw xrt_core::error(std::errc::operation_canceled, "Config P2P failed : P2P is not supported");
 
     if (force) {
         dev_fini();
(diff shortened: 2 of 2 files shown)


src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1147,7 +1147,8 @@ int shim::p2pEnable(bool enable, bool force)
         mDev->sysfs_put("p2p", "p2p_enable", err, "0");
     }
     if (!err.empty()) {
-        throw std::runtime_error("P2P is not supported");
+        std::cerr << "ERROR: Config P2P failed : P2P is not supported" << std::endl;
+        throw xrt_core::error(std::errc::operation_canceled);
```
stsoe, line 1151: The exception should be handled by caller.  In this case xclP2PEnable (which an implementation of C-API) does correctly catch the exception and send the exception message.   

Don't attempt to handle the exception here at this level only to throw another exception.  

If the exception message and or exception type is not as desired then change it on the original throw line.

I agree that you should throw xrt_core::error with an error code, as you do on line 1151, but add the message paylload and remove the std::cerr.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -2665,6 +2666,9 @@ int xclP2pEnable(xclDeviceHandle handle, bool enable, bool force)
     xocl::shim *drv = xocl::shim::handleCheck(handle);
     return drv ? drv->p2pEnable(enable, force) : -ENODEV;
   }
+  catch(const xrt_core::error&) {
+    throw xrt_core::error(std::errc::operation_canceled);
+  }
```
stsoe, line 2671: This is not friendly to the caller.  You are hiding the actual exception.

stsoe (changes requested):
The code should not be littered with try/catch, the try/catch should be in a central place as much as possible.  In this case the central place is xclP2PEnable, which already has a try/catch because it is a C-API implementation

src/runtime_src/core/pcie/linux/shim.cpp
dbenusov-xilinx: Will do thanks for the tip

src/runtime_src/core/pcie/linux/shim.cpp
dbenusov-xilinx: I moved the exception parsing out of the shim.cpp file and partly into the ishim.h file. It is very strange how we throw exceptions and return error values from p2pEnable, is that the expected style? It seems like most of the functions in shim.cpp return error codes, but p2pEnable throws exceptions and returns some codes. Maybe this should be looked at some more? Anyways let me know if the current changes makes a bit more sense.

src/runtime_src/core/common/ishim.h
```diff
@@ -431,15 +431,30 @@ struct shim : public DeviceType
   void
   p2p_enable(bool force) override
   {
-    if (auto ret = xclP2pEnable(DeviceType::get_device_handle(), true, force))
-      throw system_error(ret, "failed to enable p2p");
+    auto ret = xclP2pEnable(DeviceType::get_device_handle(), true, force);
+    switch (ret) {
```
dbenusov-xilinx, line 435: These two values stem from shim.cpp:1134 and shim.cpp:2662. This matches processing for set_cma (ishim.h:461) but it is mixed with some exception handling returns which makes the default case necessary. I will be sure to add a few comments if you like how this is formatted

src/runtime_src/core/common/ishim.h
```diff
@@ -431,15 +431,30 @@ struct shim : public DeviceType
...
+      case -EINVAL:
+      case -ENODEV:
```
dbenusov-xilinx, line 437: These two values stem from shim.cpp:1134 and shim.cpp:2662. This matches processing for set_cma (ishim.h:461) but it is mixed with some exception handling returns which makes the default case necessary. I will be sure to add a few comments if you like how this is formatted

stsoe: I don't see what this new code is doing that the old one didn't.

src/runtime_src/core/common/ishim.h
```diff
@@ -431,15 +431,30 @@ struct shim : public DeviceType
...
-    if (auto ret = xclP2pEnable(DeviceType::get_device_handle(), false, force))
-      throw system_error(ret, "failed to disable p2p");
+    auto ret = xclP2pEnable(DeviceType::get_device_handle(), false, force);
+    switch (ret) {
+      case -EINVAL:
+      case -ENODEV:
+        throw system_error(ret, "failed to disable p2p");
+        break;
+      default:
+        break;
```
stsoe, line 456: What really change here?  I don't see any change how this code behaves.  The end result is the same.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -2661,14 +2658,8 @@ int xclInternalResetDevice(xclDeviceHandle handle, xclResetKind kind)
...
 int xclP2pEnable(xclDeviceHandle handle, bool enable, bool force)
 {
-  try {
     xocl::shim *drv = xocl::shim::handleCheck(handle);
     return drv ? drv->p2pEnable(enable, force) : -ENODEV;
-  }
-  catch (const std::exception& ex) {
-    xrt_core::send_exception_message(ex.what());
-    return -ENODEV;
-  }
```
stsoe, line 2671: No. this will not work.  This can be called from host C application and now it propagates an exception.

src/runtime_src/core/common/ishim.h
dbenusov-xilinx: The old code would print out an additional error message when the throw returned from the function.

PR #6281, opened 2022-02-16
Use new xbutil commands to enable host memory

mariodruiz:
Addressing comments on #6276 

Update the documentation to use new xbutil commands

Documentation was referring to legacy tools

Updated commands to new switches

None

### DIFF
--- a/src/runtime_src/doc/toc/hm.rst
+++ b/src/runtime_src/doc/toc/hm.rst
@@ -90,7 +90,7 @@ For the platform supporting the host memory access feature, we can observe the f
      - **Max Shared Host Memory**: The maximum host memory supported by the platform.
      - **Shared Host Memory**: The host memory specified for this card (by ``xbutil configure --host-mem``)
 
-Assuming the platform supported maximum host memory is 16GB, the following output will be observed when the card is configured for 4GB host memory.
+Assuming the platform supported maximum host memory is 16GB, the following output will be observed when the card is configured for 1GB host memory, no xclbin loaded.
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


uday610:
Thanks a lot @mariodruiz , I am reviewing it in a while!

mariodruiz:
BTW, the commands in this page https://xilinx.github.io/XRT/master/html/cloud_vendor_support.html also need updating. However, I do not know what the new commands are, the migration guide does not cover this part.

uday610:
@AShivangi , can you please review and check the output shown by >`xbutil examine -r pcie-info -d 0000:17:00.1` .  When I checked with master branch XRT the    `Enabled Host Memory`   is shown as a SIZE instead of a number that is shown from this PR update. So which should be correct and which one should be bug? Seems like there is a discrepancy between the last release and master branch.

AShivangi:
@uday610, I am talking to @chienwei-lan about this and I'll have an update soon

AShivangi:
Hi @uday610, there is no bug. Here is how the host mem works:
1. You can enable host-mem any time you want and it will be reflected in the report. So when an xclbin that supports HOST[0] is _not_ loaded and 1G host mem is enabled, the output would look like:
 ```
 Shared Host Memory   : 1G  
 Max Shared Host Memory : 0 Byte
 Enabled Host Memory  : 0 Byte
```
2. When you load an xclbin with the host mem support, the `Max Shared Host Mem` gets populated like:
 ```
 Shared Host Memory   : 1G  
 Max Shared Host Memory : 16 GB 
 Enabled Host Memory  : 0 Byte
```
3. Now when you run an application which exercises HOST[0], `Enabled Host Memory` is populated:
 ```
 Shared Host Memory   : 1G  
 Max Shared Host Memory : 16 GB 
 Enabled Host Memory  : 1G
```

uday610 (approved):

AShivangi (approved):

chvamshi-xilinx:
ok to test

PR #6284, opened 2022-02-16
CR-1121688 XRT error while loading xclbin (argidx mismatch)

stsoe:
Support xclbin where indexed arguments have multiple components.

XRT triggered error when loading an xclbin where indexed arguments
have multiple components.  This case is not supported by OpenCL nor by
native XRT APIs, but used to be supported by semi-deprecated xcl APIs.

A recent change #6210 in hw emulation to consolidate xclbin parsing with
core XRT, triggered an extra DRC check that prevents multiple kernel
arguments with same index.  Managed kernel execution is treated as a
function call and each argument to the function is determined based on
its index in xclbin XML metadata.

This PR adds a simple merge routine to the xclbin parser that combines
multi-component arguments into one component, which then allows core
XRT to operate under the one-argument one-index assumption.

Unmanaged kernel execution is unaffected by this change.  Reading and
writing registers of the IP does not use the argument meta data.

Low

Regression testing on HW

### DIFF
--- a/src/runtime_src/core/common/xclbin_parser.cpp
+++ b/src/runtime_src/core/common/xclbin_parser.cpp
@@ -308,6 +308,58 @@ get_portname_width_map(const pt::ptree& xml_kernel)
   return pwmap;
 }
 
+// Merge multi-component args into the first component of the argument
+//
+// Pre-condition:
+//  - args is sorted based on argidx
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx (approved):
Looks good.

PR #6303, opened 2022-02-19
CR-1121536 ASTeR TC 17.06 - vck5000 - crash occurs when flashing is i…

xdavidz:
…nterrupted with SIGINT/SIGKILL

This is to avoid panic when submitted command gets killed, it's cmd pointer stays in the linkedlist and itself gets freed earlier.
So, each time we remove their cmd pointer from the list before free the cmd.

I think the xgq.c can be optimized with better code structure with better xgq_cmd alloc/free api. @xuhz 

Tested with multiple times ctrl+c, no panics anymore.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
@@ -114,7 +114,7 @@ struct xocl_xgq_vmr_cmd {
 	struct timer_list	xgq_cmd_timer;
 	struct xocl_xgq_vmr	*xgq_vmr;
 	u64			xgq_cmd_timeout_jiffies; /* timout till */
-	uint32_t		xgq_cmd_rcode;
+	int			xgq_cmd_rcode;
 	/* xgq complete command can return in-line data via payload */
 	struct xgq_cmd_cq_default_payload	xgq_cmd_cq_payload;
 };
@@ -236,7 +236,7 @@ static int complete_worker(void *data)
 		if (xgq->xgq_polling) {
 			usleep_range(1000, 2000);
 		} else {
-			wait_for_completion_interruptible(&xgq->xgq_irq_complete);
+			wait_for_completion_killable(&xgq->xgq_irq_complete);
 		}
 
 		if (kthread_should_stop()) {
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -1447,12 +1474,14 @@ static int xgq_collect_sensors(struct platform_device *pdev, int sid,
...
 	if (wait_for_completion_killable(&cmd->xgq_cmd_complete)) {
-		XGQ_ERR(xgq, "submit cmd killed");
-		goto done;
+		XGQ_ERR(xgq, "submitted cmd killed");
+		xgq_submitted_cmd_remove(xgq, cmd);
 	}
 
 	ret = cmd->xgq_cmd_rcode;
 
+	printk("DZ__ %d\n", (int)(0xffff));
```
houlz0507, line 1483: debug print? :)

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -729,8 +753,8 @@ static ssize_t xgq_transfer_data(struct xocl_xgq_vmr *xgq, const void *buf,
 
 	/* wait for command completion */
 	if (wait_for_completion_killable(&cmd->xgq_cmd_complete)) {
-		XGQ_ERR(xgq, "submit cmd killed");
-		goto done;
+		XGQ_ERR(xgq, "submitted cmd killed");
+		xgq_submitted_cmd_remove(xgq, cmd);
```
houlz0507, line 757: why 'goto done' is not needed here? Maybe you can have a unified code to call xgq_submitted_cmd_remove() once instead of calling it on every error branch?

xdavidz: how to call it once? you mean a callback API?

We indeed need to improve the code structure but maybe not now because it will introduce too many code changes.

The goto is removed, so that the xgq_cmd_rcode can be checked after xgq_submitted_cmd_remove. It should return -EIO. If we goto done, we return 0 which is incorrect.

houlz0507: Ah. Okay, they are in different functions. I thought they are in the same function. :)

houlz0507 (approved):

maxzhen (approved):

xilinxgitops:
retest this please

dayeh-xilinx:
retest this please.

PR #6309, opened 2022-02-22
CR-1122901 hwmon_sdm: create sysfs entries for board info.

rajkumar-xilinx:
After receiving board info response from VMR, create all possible
sysfs nodes and register them with hwmon_sdm driver


rajkumar-xilinx:
bash-4.2# cd /sys/bus/pci/devices/0000\:d8\:00.0/hwmon_sdm.m.55574528/

bash-4.2# ls
active_msp_ver  bd_part_num  fan_presence  mac_addr1  modalias  pcie_info  revision    subsystem       uevent
bd_name         driver       mac_addr0     mfg_date   oem_id    power      serial_num  target_msp_ver  uuid

bash-4.2# cat active_msp_ver
0x333434
bash-4.2# cat bd_name
VCK5000-PP
bash-4.2# cat bd_part_num
05025-04
bash-4.2# cat fan_presence
1
bash-4.2# cat mac_addr0
0xa4003cf60c350a00
bash-4.2# cat mac_addr1
0x3df60c350a00
bash-4.2# cat mfg_date
0xcd63d7
bash-4.2# cat oem_id
0x10da
bash-4.2# cat pcie_info
0xe00ee104850ee10
bash-4.2# cat revision
A
bash-4.2# cat serial_num
XFL1TQYAVK2A
bash-4.2# cat target_msp_ver
0x1f0404
bash-4.2# cat uuid
0xa1469de82aa714f3

chvamshi-xilinx (approved):
Looks good. 
Hopefully we are using same sysfs names .They are under xmc subfolder earlier, now they would be under sdm.

rajkumar-xilinx:
> Looks good. Hopefully we are using same sysfs names .They are under xmc subfolder earlier, now they would be under sdm.

Yes, I matched the sysfs names w.r.t xmc driver.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
```diff
@@ -755,10 +966,11 @@ static int hwmon_sdm_update_sensors_by_type(struct platform_device *pdev,
 	if (!sdm->privileged)
 		return hwmon_sdm_read_from_peer(pdev, repo_type);
 
-	sdm->sensor_data[repo_id] = (char*)kzalloc(sizeof(char) * resp_size,
-                                               GFP_KERNEL);
+	if (!sdm->sensor_data[repo_id])
+		sdm->sensor_data[repo_id] = (char*)kzalloc(sizeof(char) * RESP_LEN, GFP_KERNEL);
```
maxzhen, line 970: Can we avoid this memory allocation as well? Where is this memory freed? In general, the less memory allocation, the better in driver... memory allocation is slow and difficult to manage.

maxzhen (changes requested):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
```diff
@@ -722,6 +930,11 @@ static int hwmon_sdm_probe(struct platform_device *pdev)
 		sdm->privileged = true;
 	}
 
+	sdm->bdinfo = (struct xocl_sdr_bdinfo*)devm_kzalloc(&sdm->pdev->dev,
```
maxzhen, line 933: why not embed this data structure into struct xocl_hwmon_sdm so that you don't need to worry about managing it? At least, you need to free it during hwmon_sdm_remove()?

maxzhen (changes requested):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
rajkumar-xilinx: Hi Max, sensor_data size is known at runtime by issuing sensor size request before the actual request. Since, this work is pending, we are allocating 4KB memory by default for now. 
And I agree that we need to free the memory in device_remove () function. But, I replaced kzalloc() with devm_kzalloc() API now, so it is not required to free the memory.  Because, when the device is detached or the driver for the device is unloaded, that memory is freed automatically.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
rajkumar-xilinx: IMO, it is better to have separate structure for board sensor data for easy maintenance and easy to track what are all sensors can be received for this BDINFO request.

Now, I replaced kzalloc() with devm_kzalloc(), it is not required to free the memory. It will be freed automatically when drivers are unloaded or when the device is detached from the system.
Thank you for pointing this :)

maxzhen: you can have separate data structure, but you don't have to dynamically allocate, right? can you just put the data structure directly into the struct xocl_hwmon_sdm (v.s. putting a pointer over there)?

maxzhen (changes requested):

maxzhen (changes requested):
I may have missed it, but where is the lock for protecting members in sdm? you may have one thread accessing sysfs node while you're still working on previous one, right?

rajkumar-xilinx:
> I may have missed it, but where is the lock for protecting members in sdm? you may have one thread accessing sysfs node while you're still working on previous one, right?

Yes, I initialized lock in probe, but not used it. I added sdm lock in the show() API now. Please review.

rajkumar-xilinx:
> You can have separate data structure, but you don't have to dynamically allocate, right?

Yes, I removed dynamic allocation of bdinfo. Thanks.

dayeh-xilinx:
retest this please.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
```diff
@@ -271,7 +292,9 @@ static ssize_t hwmon_sensor_show(struct device *dev,
 	char output[64];
 	uint32_t uval = 0;
 
+	mutex_lock(&sdm->sdm_lock);
 	get_sensors_data(sdm->pdev, repo_id);
+	mutex_unlock(&sdm->sdm_lock);
```
maxzhen, line 297: I'm not sure how the locking schema is designed here, but after you drop the lock here, you still accessing sdm->sensor_data. Can it be changed now by someone else at the same time?

maxzhen (changes requested):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
rajkumar-xilinx: yes, the sensor_data can be accessed by other thread/process. To fix it, driver shall copies the required data from sensor_data and then release the lock.

maxzhen: The fix will make sure sysfs node entry points are multi-thread safe. But, when SDM runs in xclmgmt, there is one more entry point - sensor reading from xocl side. We need to also make sure that entry point is multi-thread safe and also make sure when both entry points are accessed at the same time, it is absolutely safe. I believe the locking schema needs to be designed and tested based on this.

rajkumar-xilinx:
Hi @maxzhen 
I tried to fix using memcpy for more than 4 bytes case, but could not able to get proper results.
So, I added a TODO: fix comment for case where buf_len is > 4 in the code. I will fix it in next coming PR and I am already working on it. It is taking more time than what I was expected.

In the meantime, can we please got ahead reviewing this PR? Thank you.

>>>We need to also make sure that entry point is multi-thread safe and also make sure when both entry points are accessed at the same time, it is absolutely safe. I believe the locking schema needs to be designed and tested based on this.

Verification is in progress now.

maxzhen:
> 

I see that the lock is only acquired in hwmon_sensor_show(). But this entry point is only for sysfs node access, right? How does it work when it handles request from xocl ( I assume that hwmon_sensor_show() is not called in that case)?

rajkumar-xilinx:
> > 
> 
> I see that the lock is only acquired in hwmon_sensor_show(). But this entry point is only for sysfs node access, right? How does it work when it handles request from xocl ( I assume that hwmon_sensor_show() is not called in that case)?

hwmon_sensor_show() API is common for both xclmgmt & xocl hwmon sysfs node access. So, this function will be invoked in both these cases
- /sys/bus/pci/devices/0000\:d8\:00.0/hwmon/hwmon3/*
- /sys/bus/pci/devices/0000\:d8\:00.1/hwmon/hwmon5/*

Below is the sequence:
1. user trying to access xocl's hwmon sysfs node. Ex: cat /sys/bus/pci/devices/0000\:d8\:00.1/hwmon/hwmon5/curr1_input
2. hwmon_sdm driver's API hwmon_sensor_show() is invoked.
3. **Acquire lock**
4. check if the sensor_data is active or expired
    a) If active (i.e timer is not expired) then Goto step (5)
    b) If expired, prepare mailbox request, receive response, update timer. Goto step (5)
5. Read required data from sensor_data[] and store it into buf.
6. **Release lock**
7. Return from hwmon_sensor_show()

Above steps are same for xclmgmt's hwmon sysfs read request except step 4 (b). Here, it directly requests XGQ/VMR for sensor_data.

maxzhen (approved):

maxzhen:
There is still race condition in the code that needs to be addressed. I've talked to @rajkumar-xilinx offline and it will be addressed in the very next PR. Merging it now to unblock the board qualification.

PR #6310, opened 2022-02-22
Support multiple run object completion

stsoe:
XMA multi-thread access to execwait using shared device handle.
XMA checking multiple run objects for completion after one execwait call.

This PR adds an internal APIs to safefly call exec_wait from any host
thread using same device object.

The PR also improves xrt::run::status() to guarantee that the internal
state of the run object reflects possible command completion, thus
allow the run object to be reused for subsequent execution of the
kernel.

### DIFF
--- a/src/runtime_src/core/common/api/command.h
+++ b/src/runtime_src/core/common/api/command.h
@@ -1,19 +1,5 @@
-/**
- * Copyright (C) 2020 Xilinx, Inc
- *
- * Licensed under the Apache License, Version 2.0 (the "License"). You may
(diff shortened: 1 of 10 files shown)


chvamshi-xilinx (approved):
Looks good. Thanks Soren. This is much more simpler now. Users can directly call xrt::run::state() and reuse if state is completed.

vboggara-xilinx (approved):
Looks good, great to see this feature. provided testcase is very helpful.

dayeh-xilinx:
retest this please.

dayeh-xilinx:
retest this please.

PR #6318, opened 2022-02-23
CR-1119177: Fixing cu's with no interrupt case in kds

chvamshi-xilinx:
If there is any CU which doesn't have interrupts enabled, KDS should start a global polling thread to serve the commands. This is not happening today 

We changed  "poling thread per cu" to  to "single polling thread for all cus"  due to heavy host cpu utilization. This change might caused this issue

Starting a polling thread if any of the CU doesnt support interrupt

No

Verified the CR and other simple cases

No


mamin506 (approved):

dayeh-xilinx:
retest this please.

PR #6327, opened 2022-02-25
CR-1123564 Handled the condition when we do not have any PL kernels i…

venkatp-xilinx:
Handled the condition when we do not have any PL kernels in the application.

Able to run the sw_emu designs that does have only AIE kernels, no PL kernels.

We started to get the designs which does have only AIE kernels and PS on x86 feature verification found this issue

Having the safe check on ip_layout ptr while accessing the APIs of that object 

Nope

Ran the canary suite and also ran the PS on x86 designs.

### DIFF
--- a/src/runtime_src/core/common/device.cpp
+++ b/src/runtime_src/core/common/device.cpp
@@ -185,8 +185,10 @@ register_axlf(const axlf* top)
   }
   catch (const query::no_such_key&) {
     auto ip_layout = get_axlf_section<const ::ip_layout*>(IP_LAYOUT);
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/device.cpp
```diff
@@ -185,8 +185,10 @@ register_axlf(const axlf* top)
   }
```
rozumx, line 185: Copyright date line 2 needs to be updated.

rozumx (approved):

stsoe (approved):

chvamshi-xilinx (approved):

PR #6330, opened 2022-02-25
Decouple iops test from BIST

chienwei-lan:
Decoupling iops test from bist. 

If users like to get the iops number, they should call xbutil validate --run iops explicitly.


AShivangi (approved):

maxzhen:
@chvamshi-xilinx mini pipeline infra issue?

eshwar-xilinx:
Hi @chvamshi-xilinx , most of other PRs are also facing same issue. Seems like a github network issue.
I will try re-running the PRs

PR #6337, opened 2022-02-28
CR-1111929 P2P required IO space message

dbenusov-xilinx:
https://jira.xilinx.com/browse/CR-1111929
When we enable p2p, often time after warm reboot the machine become unresponsive. I see lot of questions, confusions from the user. Currently we only show p2p is enabled or disabled through xbutil. I propose we will show memory requirement too. This would help the user upfront if his machine/server is capable enough or not. He can then move to another server if required. 

Added checks if a factory image is installed and display the appropriate messaging.

This does not fix the issue of reboot if the user has not looked at the IO space in the platform report. @AShivangi suggested that a check should be built into the P2P enable command that outputs a warning instead. I think that would be good as a future enhancement.

On u280 (p2p supported)
```
Devices present
BDF             :  Shell                      Platform UUID  Device ID
[0000:65:00.1]  :  xilinx_u280_xdma_201920_3  0x5e278820     user(inst=128)

dbenusov@xsjpranjal01:/proj/xsjhdstaff6/dbenusov/XRT$ xbutil examine -d 65:00

-----------------------------------------------
1/1 [0000:65:00.1] : xilinx_u280_xdma_201920_3
-----------------------------------------------
Platform
  XSA Name               : xilinx_u280_xdma_201920_3
  Platform UUID          : 0x5e278820
  FPGA Name              : xcu280-fsvh2892-2L-e
  JTAG ID Code           : 0x14b7d093
  DDR Size               : 34359738368 Bytes
  DDR Count              : 2
  Mig Calibrated         : true
  P2P Status             : disabled
  P2P IO space required  : 64 GB

Mac Addresses            : 01:02:03:04:05:06
                         : 01:02:03:04:05:07

Xclbin UUID
  00000000-0000-0000-0000-000000000000

Compute Units
  PL Compute Units
    Index   Name                                              Base_Address    Usage   Status

  PS Compute Units
    Index   Name                                              Base_Address    Usage   Status
```
On u30 (p2p not supported)
```
Devices present
BDF             :  Shell                            Platform UUID                         Device ID
[0000:b3:00.1]  :  xilinx_u200_gen3x16_xdma_base_1  A2D4F3CF-5B7A-0B7B-70F9-DA589CB5B3CD  user(inst=130)
[0000:18:00.1]  :  xilinx_u30_gen3x4_base_2         6E7C97F7-949F-5106-3075-A77784C30A4B  user(inst=129)
[0000:17:00.1]  :  xilinx_u30_gen3x4_base_2         6E7C97F7-949F-5106-3075-A77784C30A4B  user(inst=128)

dbenusov@xsjsagarw50:~$ xbutil examine -d 18:00

1/1 [0000:18:00.1] : xilinx_u30_gen3x4_base_2
Platform
  XSA Name               : xilinx_u30_gen3x4_base_2
  Platform UUID          : 6E7C97F7-949F-5106-3075-A77784C30A4B
  FPGA Name              :
  JTAG ID Code           : 0x14a5c093
  DDR Size               : 0 Bytes
  DDR Count              : 0
  Mig Calibrated         : true
  P2P Status             : not supported

Xclbin UUID
  00000000-0000-0000-0000-000000000000

Compute Units
  PL Compute Units
    Index   Name                                              Base_Address    Usage   Status

  PS Compute Units
    Index   Name                                              Base_Address    Usage   Status


```

### DIFF
--- a/src/runtime_src/core/common/info_platform.cpp
+++ b/src/runtime_src/core/common/info_platform.cpp
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2021 Xilinx, Inc
+ * Copyright (C) 2021-2022 Xilinx, Inc
  *
  * Licensed under the Apache License, Version 2.0 (the "License"). You may
  * not use this file except in compliance with the License. A copy of the
--- a/src/runtime_src/core/common/info_platform.cpp
+++ b/src/runtime_src/core/common/info_platform.cpp
@@ -108,9 +108,9 @@ add_p2p_config(const xrt_core::device* device, ptree_type& pt)
 {
   try {
     ptree_type pt_p2p;
-    auto config = xrt_core::device_query<xq::p2p_config>(device);
-    auto config_map = xrt_core::query::p2p_config::to_map(config);
-    for(auto& pair : config_map)
+    const auto config = xrt_core::device_query<xq::p2p_config>(device);
+    const auto config_map = xrt_core::query::p2p_config::to_map(config);
+    for(const auto& pair : config_map)
       pt_p2p.add(pair.first, xrt_core::utils::unit_convert(pair.second)); // Turn bytes into GB
     pt.put_child("p2p", pt_p2p);
   }
--- a/src/runtime_src/core/common/info_platform.cpp
+++ b/src/runtime_src/core/common/info_platform.cpp
@@ -103,33 +103,15 @@ add_mig_info(const xrt_core::device* device, ptree_type& pt)
   }
 }
 
-static std::map<std::string, long long> 
-p2p_convert_config(xrt_core::query::p2p_config::result_type& config)
-{
-  std::map<std::string, long long> map;
-  for (auto& str : config) {
-    // str is in key:value format obtained from p2p_config query
-    auto pos = str.find(":");
-    std::string key = str.substr(0, pos);
-    try {
-      long long value = std::stoll(str.substr(pos + 1));
-      map[key] = value;
-    } catch (const std::exception&) {
-      // Failed to parse a non long long BAR value. Dont parse it into the map and move on!
-    }
-  }
-  return map;
-}
-
 void
 add_p2p_config(const xrt_core::device* device, ptree_type& pt)
 {
   try {
     ptree_type pt_p2p;
     auto config = xrt_core::device_query<xq::p2p_config>(device);
-    auto config_map = p2p_convert_config(config);
+    auto config_map = xrt_core::query::p2p_config::to_map(config);
     for(auto& pair : config_map)
-      pt_p2p.add(pair.first, std::to_string(pair.second/1024/1024/1024)); // Turn bytes into GB
+      pt_p2p.add(pair.first, xrt_core::utils::unit_convert(pair.second)); // Turn bytes into GB
     pt.put_child("p2p", pt_p2p);
   }
   catch (const xq::exception&) {
--- a/src/runtime_src/core/common/info_platform.cpp
+++ b/src/runtime_src/core/common/info_platform.cpp
@@ -124,13 +124,17 @@ p2p_convert_config(xrt_core::query::p2p_config::result_type& config)
 void
 add_p2p_config(const xrt_core::device* device, ptree_type& pt)
 {
-  ptree_type pt_p2p;
-  auto config = xrt_core::device_query<xq::p2p_config>(device);
-  auto config_map = p2p_convert_config(config);
-  for(auto& pair : config_map)
-    pt_p2p.add(pair.first, std::to_string(pair.second/1024/1024/1024)); // Turn bytes into GB
-
-  pt.put_child("p2p", pt_p2p);
+  try {
+    ptree_type pt_p2p;
+    auto config = xrt_core::device_query<xq::p2p_config>(device);
+    auto config_map = p2p_convert_config(config);
+    for(auto& pair : config_map)
+      pt_p2p.add(pair.first, std::to_string(pair.second/1024/1024/1024)); // Turn bytes into GB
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/common/info_platform.cpp
```diff
@@ -103,6 +103,50 @@ add_mig_info(const xrt_core::device* device, ptree_type& pt)
...
+
+void
+add_p2p_config(const xrt_core::device* device, ptree_type& pt)
+{
+  try {
+    ptree_type pt_p2p;
+    auto config = xrt_core::device_query<xq::p2p_config>(device);
+    auto config_map = p2p_convert_config(config);
+    for(auto& pair : config_map)
+      pt_p2p.add(pair.first, std::to_string(pair.second/1024/1024/1024)); // Turn bytes into GB
```
AShivangi, line 132: Can you create a common function for this in XBUtilities like `bytes to string(uint64_t bytes)` which takes the value and convert to Byte, KB, MB, etc. I feel like we have such a function somewhere in the codebase since we parse memory sizes for the memory report. Maybe start looking there.

src/runtime_src/core/common/info_platform.cpp
```diff
@@ -103,6 +103,50 @@ add_mig_info(const xrt_core::device* device, ptree_type& pt)
   }
 }
 
+static std::map<std::string, long long> 
+p2p_convert_config(xrt_core::query::p2p_config::result_type& config)
```
AShivangi, line 107: this function has been copied over from somewhere else? If so, it is better to include that file
if not then I think the correct place for this is in qr::p2p_config as a "to_string" function or something similar.

AShivangi (changes requested):

src/runtime_src/core/common/info_platform.cpp
dbenusov-xilinx: One already existed as xrt_core::utils::unit_convert. Very handy!

src/runtime_src/core/common/query_requests.cpp
```diff
@@ -28,46 +27,42 @@ to_string(xrt_core::query::p2p_config::value_type value)
...
-  else if (rbar != -1 && rbar > bar) {
-    return {xrt_core::query::p2p_config::value_type::reboot, 
-            "Warning:Please WARM reboot to enable p2p now."};
-  }
-  else if (remap > 0 && remap != bar) {
-    return {xrt_core::query::p2p_config::value_type::error,
-            "Error:P2P config failed. P2P remapper is not set correctly"};
-  }
-  else if (bar == exp_bar) {
+  if (config_map.find("bar") == config_map.end())
```
rozumx, line 56: This nested if loop is not needed for if the condition matches it returns.

Please remove replace the "else" with a new line.

src/runtime_src/core/tools/common/ReportPlatforms.cpp
```diff
@@ -23,6 +23,7 @@
 
 // 3rd Party Library - Include Files
 #include <boost/property_tree/json_parser.hpp>
+#include <boost/optional/optional.hpp>
```
rozumx, line 26: Please make sure that the #includes are in alphabetical order.   
For short lists like this one, it doesn't matter, for for longer ones, it helps quickly recognize and remove duplicates.

src/runtime_src/core/tools/common/ReportPlatforms.cpp
```diff
@@ -72,6 +73,10 @@ ReportPlatforms::writeReport( const xrt_core::device* /*_pDevice*/,
     _output << boost::format("  %-23s: %s \n") % "Mig Calibrated" % pt_status.get<std::string>("mig_calibrated");
     _output << boost::format("  %-23s: %s \n") % "P2P Status" % pt_status.get<std::string>("p2p_status");
 
+    boost::optional<const boost::property_tree::ptree&> pt_config = pt_platform.get_child_optional("config.p2p");
```
rozumx, line 76: Optional does work here.  Unfortunately, in the past not ALL versions of gcc correctly compiles this statement correctly.  While I don't think we are support those older OSes any more, if this code should happen to fail on other OSes, the following solution works as well:

```
const boost::property_tree::ptree ptEmpty;
const boost::property_tree::ptree pt_& pt_config = pt_platform.get_child("config.p2p", ptEmpty);

if (!pt_config.empty()) {
   ...
```

src/runtime_src/core/common/query_requests.cpp
```diff
@@ -28,46 +27,42 @@ to_string(xrt_core::query::p2p_config::value_type value)
...
+  std::map<std::string, int64_t> config_map;
+  for (auto& str : config) {
+    // str is in key:value format obtained from p2p_config query
+    const auto pos = str.find(":");
+    const std::string config_item = str.substr(0, pos);
+    try {
+      const int64_t value = std::stoll(str.substr(pos + 1));
+      config_map[config_item] = value;
+    } catch (const std::exception&) {
+      // Failed to parse a non long long BAR value. Dont parse it into the map and move on!
```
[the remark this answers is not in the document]
dbenusov-xilinx: At the moment there is an additional item in the sysfs node that does not follow the pattern of key:value. Here is the CR: https://jira.xilinx.com/browse/CR-1123432. Though I agree with the exception being thrown, I had that originally (we do not even need the try) but the strange p2p config value for the memory bank broke that plan.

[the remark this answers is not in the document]
dbenusov-xilinx: Converting back to the if/else chain to protect our data contents

rozumx (approved):

PR #6339, opened 2022-03-01
Cleanup xdp code with a mix of tabs and spaces 

hackwa:
Current xdp code mostly uses spaces, but some places have tabs in them making the code harder to read
This also removes unused legacy defines in xclperf.h file

Capture
.


jvillarre (approved):
This pull request does not change xclperf.h as that has to go through the deprecation process when changed.

PR #6341, opened 2022-03-01
CR-1121712: Address seg fault at program teardown in profiling

jvillarre:
The XDP profiling plugins instantiate static objects when the modules are loaded and these static objects are destroyed at the application end.  During application runs, profiling callbacks are hit when the user calls XRT APIs (such as OpenCL APIs or Native XRT APIs) which interact with the XDP static objects.  A test case exposed the problem where users' code additionally had static objects that called XRT APIs in their destructor, which in turn hit the profiling callbacks and caused seg faults if they happened after the profiling objects had been destroyed.  Since the ordering of how static objects is destroyed is different based on the compiler and OS, there can be no implied ordering and all of the static objects' destructors must work regardless of the order in which they are called.

This pull request adds guards to all profiling plugins' callbacks that could be triggered by the user from their own static object destructors, and makes sure that data is only tracked while the objects are alive.

### DIFF
--- a/src/runtime_src/xdp/profile/plugin/hal/hal_plugin.cpp
+++ b/src/runtime_src/xdp/profile/plugin/hal/hal_plugin.cpp
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2020 Xilinx, Inc
+ * Copyright (C) 2016-2022 Xilinx, Inc
  *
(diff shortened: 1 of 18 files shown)


IshitaGhosh (approved):

dayeh-xilinx:
retest this please. infra error(ddr retention test case failing on an AMD host)

pgschuey (approved):

PR #6347, opened 2022-03-02
VITIS 1753 Refactor XRT driver logs

eshwar-xilinx:
xocl/xclmgmt drivers exposes too many information to dmesg, many of the which are only for developers debug purpose, and end users may never need them or they don't even understand the meaning of those messages.

Due to that, some critical and useful warnings/errors which would otherwise have been easy to see are hidden in the flood of dmesg. This makes hard for the users report what has been wrong when something happens.
The PR is to cleanup the dmesg, and put those debug messages for developers to debugfs.

As the starting step, all message calls are converted such that only xocl_err/warn/info or mgmt_info/warn/err are present. Subsequent changes will be made in future PRs

Ease of Use

 xocl_xdev_info, xocl_xdev_err, xocl_xdev_dbg calls are converted to xocl_info, xocl_err, xocl_dbg calls respectively to keep track of what is displayed in dmesg and debugfs
 userpf_info, userpf_err, userpf_dbg calls are converted to xocl_info, xocl_err, xocl_dbg calls respectively to keep track of what is displayed in dmesg and debugfs
 
```
In the end, following calls will be present :
1. xocl_info, xocl_err, xocl_dbg, xocl_warn
2. mgmt_info, mgmt_err, mgmt_dbg, mgmt_warn
```

Low Risk

Built XRT locally with changes

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/axigate.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/axigate.c
@@ -85,7 +85,7 @@ static int axigate_freeze(struct platform_device *pdev)
 done:
 	mutex_unlock(&gate->gate_lock);
 
-	xocl_xdev_info(xdev, "freeze gate %s level %d",
+	xocl_info(XDEV2DEV(xdev), "freeze gate %s level %d",
 			gate->ep_name, gate->level);
 	return 0;
 }
@@ -111,7 +111,7 @@ static int axigate_free(struct platform_device *pdev)
 
 done:
 	mutex_unlock(&gate->gate_lock);
-	xocl_xdev_info(xdev, "free gate %s level %d", gate->ep_name,
+	xocl_info(XDEV2DEV(xdev), "free gate %s level %d", gate->ep_name,
 			gate->level);
 	return 0;
 }
@@ -126,7 +126,7 @@ static int axigate_reset(struct platform_device *pdev)
 	reg_wr(gate, 0x1, iag_wr);
 	mutex_unlock(&gate->gate_lock);
 
-	xocl_xdev_info(xdev, "ep_name %s level %d", gate->ep_name, gate->level);
+	xocl_info(XDEV2DEV(xdev), "ep_name %s level %d", gate->ep_name, gate->level);
 	return 0;
 }
 
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/feature_rom.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/feature_rom.c
@@ -595,7 +595,7 @@ static int get_header_from_peer(struct feature_rom *rom)
 
 	memcpy(&rom->header, header, sizeof(*header));
 
-	xocl_xdev_dbg(xdev, "Searching CDMA in dtb.");
+	xocl_dbg(XDEV2DEV(xdev), "Searching CDMA in dtb.");
 	offset = xocl_fdt_path_offset(xdev, XDEV(xdev)->fdt_blob,
 				      "/" NODE_ENDPOINTS "/" RESNAME_KDMA);
 	if (offset < 0)
@@ -604,7 +604,7 @@ static int get_header_from_peer(struct feature_rom *rom)
 	io_off = xocl_fdt_getprop(xdev, XDEV(xdev)->fdt_blob, offset,
 				  PROP_IO_OFFSET, NULL);
 	if (!io_off) {
-		xocl_xdev_err(xdev, "dtb maybe corrupted\n");
+		xocl_err(XDEV2DEV(xdev), "dtb maybe corrupted\n");
 		return -EINVAL;
 	}
 	res.start = be64_to_cpu(io_off[0]);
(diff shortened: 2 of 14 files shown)


eshwar-xilinx:
The changes proposed are according to the below confluence page :

https://confluence.xilinx.com/display/XSW/XRT+drivers+dmesg+cleanup

chvamshi-xilinx (approved):
Hope our next step to review each and every xocl_* and mgmt* messages and refine.

maxzhen (commented):
I see that some debug information is in either dmesg OR debug fs. Maybe we should make sure that all msg will first go to debugfs and, then, some will also be seen in dmesg? Thus, debugfs will contain complete debug msgs in the correct order of time. Maybe I missed the reason why, for some msgs, we do not wish to put into debug fs?

chvamshi-xilinx:
> I see that some debug information is in either dmesg OR debug fs. Maybe we should make sure that all msg will first go to debugfs and, then, some will also be seen in dmesg? Thus, debugfs will contain complete debug msgs in the correct order of time. Maybe I missed the reason why, for some msgs, we do not wish to put into debug fs?

@houlz0507  might know the background .

houlz0507:
I agree with Max that debugfs should have all the logs.  And also, there might be still some critical "info" needs to be printed to dmesg. Especially those information dumps after error happens.

chvamshi-xilinx:
> I agree with Max that debugfs should have all the logs. And also, there might be still some critical "info" needs to be printed to dmesg. Especially those information dumps after error happens.



> I agree with Max that debugfs should have all the logs. And also, there might be still some critical "info" needs to be printed to dmesg. Especially those information dumps after error happens.

Regarding all logs in debugfs, -> I hope there is no downside of adding all the logs in debugfs, Change is simple,
Regarding critical info printing after error -> If there are some errors , we might be already using xocl_err/xocl_warn. They would still come in dmesg.

chvamshi-xilinx:
> > > I agree with Max that debugfs should have all the logs. And also, there might be still some critical "info" needs to be printed to dmesg. Especially those information dumps after error happens.
> > 
> > 
> > > I agree with Max that debugfs should have all the logs. And also, there might be still some critical "info" needs to be printed to dmesg. Especially those information dumps after error happens.
> > 
> > 
> > Regarding all logs in debugfs, -> I hope there is no downside of adding all the logs in debugfs, Change is simple, Regarding critical info printing after error -> If there are some errors , we might be already using xocl_err/xocl_warn. They would still come in dmesg.
> 
> I think that this PR is implementing the mechanism - what msg goes to where. As to specific msgs that needs to be in dmesg, yes, it should be printed out by calling xocl_err or xocl_warn. If that is not the case today, we should fix them. If @houlz0507 can identify these cases now, we could also fix them in this PR. If not, we may fix them in follow-up PRs.
> 
> However, there is a big difference b/w debugfs and dmesg. Msgs in debugfs only will be difficult to retrieve after reboot due to system crash, right? While dmesg log will be easy and all our pipeline teams and end users knows how to fetch them. So, maybe we can hold off the change until after we branch so that we don't make this change in the last minute of this release?

Thats a good point @maxzhen . Let me check whether we can hold off this change.

chvamshi-xilinx:
retest this please

eshwar-xilinx:
Hi @maxzhen @houlz0507 @chvamshi-xilinx 

In the new commit, I have made changes such that the displaying mechanism is unaltered.
All the extra calls ( userpf_info, userpf_err, userpf_dbg, xocl_xdev_info, xocl_xdev_err, xocl_xdev_dbg) are converted to xocl_info, xocl_err and xocl_dbg. I have changed the PR message accordingly.

If we need to change how information is displayed in future PRs, it will be a one line change in xocl_drv.h (for xocl_info/warn/err messages) and mgmtpf/mgmt-core.h( for mgmt_info/warn/err messages).
We can refine the content of the messages in the future commits/PRs.

chvamshi-xilinx:
> Hi @maxzhen @houlz0507 @chvamshi-xilinx
> 
> In the new commit, I have made changes such that the displaying mechanism is unaltered. All the extra calls ( userpf_info, userpf_err, userpf_dbg, xocl_xdev_info, xocl_xdev_err, xocl_xdev_dbg) are converted to xocl_info, xocl_err and xocl_dbg. I have changed the PR message accordingly.
> 
> If we need to change how information is displayed in future PRs, it will be a one line change in xocl_drv.h (for xocl_info/warn/err messages) and mgmtpf/mgmt-core.h( for mgmt_info/warn/err messages). We can refine the content of the messages in the future commits/PRs.

Hi @eshwar-xilinx , Thanks for making this change.
I had a discussion with Anup today. We have moved this to 2022.2. I will mark this PR for 2022.2

houlz0507:
I think we may be careful to make this change. If system crash, we probably will loss debugfs messages. It could make debugging very difficult. There are something I can think of:
1. make sure we keep important debug message still printed to dmesg
2. build a script/tool to dump debugfs message in a file/console (just like dmesg -w). Thus, we can ask customer to use it and reproduce problem
3. fetch debugfs buffer from vmcore. Could it be a simple string search?
I am not very confident for this change before we have a good debugging approach.

PR #6360, opened 2022-03-03
CR-1124025 Remove flush_default_only restriction when upgrading from backup

AShivangi:
Allow operations even if the backup image is booted

This was introduced during the initial implementation of a/b boot 

remove error throw

none

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
@@ -444,7 +444,7 @@ update_default_only(xrt_core::device* device, bool value)
       if(boost::iequals(vmr_stat.get<std::string>("label"), "Boot on default")) {
         // if backup is booted, then do not proceed
         if(std::stoi(vmr_stat.get<std::string>("value")) != 1) {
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/tools/xbmgmt2/SubCmdProgram.cpp
```diff
@@ -444,8 +444,7 @@ update_default_only(xrt_core::device* device, bool value)
       if(boost::iequals(vmr_stat.get<std::string>("label"), "Boot on default")) {
         // if backup is booted, then do not proceed
         if(std::stoi(vmr_stat.get<std::string>("value")) != 1) {
-          std::cout << "ERROR: Please load default boot to use this option" <<std::endl;
-          throw xrt_core::error(std::errc::operation_canceled);
+          std::cout << "Backup image booted. Action will be performed only on deafult image." <<std::endl;
```
rozumx, line 447:
```suggestion
          std::cout << "Backup image booted. Action will be performed only on default image.\n";
```

rozumx (changes requested):

rozumx (approved):

PR #6362, opened 2022-03-03
CR-1118968 Disable PDI reload on warm reboot

xdavidz:
- we should not touch the GPIO after hot_reset because touching those GPIO will cause warm reboot hung.
- we defer this operation till we start hot_reset from host side
- 
- got agreement with lizhi that we should move those loading APU into xgq_probe
- in the future, we can make this async but for now, let's use the straight way
- 
tested on multiple servers
1) hot reset 2) warm reboot

make sure card can comes back after warm reboot


maxzhen (approved):

houlz0507 (approved):

xdavidz:
retest is please

xdavidz:
retest this please

PR #6316, opened 2022-02-23
Initial check-in of PS kernel support on VCK5000 Discovery

jeffli-xilinx:
#### Problem solved by the commit
Initial Check-In for PS Kernel support with XGQ on VCK5000 Discovery Platform

#### Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered
New feature to enable PS kernel

#### How problem was solved, alternative solutions (if any) and why they were rejected
Added separate SCU sub-devices, SCU contexts, cu_domain field, and added support throughout KDS for the new SCU_DOMAIN.

#### Risks (if any) associated the changes in the commit
Will not be backward compatible with old PS kernel flow on U30

#### What has been tested and how, request additional testing if necessary
Simple hello world

PR #6368, opened 2022-03-04
Updated petalinux build to 2022.1_02210750

jeffli-xilinx:
Petalinux 2022.1 02160715 has been removed from the build area.
The build was not retained with a JIRA ticket.

PR https://github.com/Xilinx/XRT/pull/6316 changed the Petalinux build
Once the build was removed build_edge.sh fails due to petalinux not found

Moved to 02210750 build

None

Tested on vck5000 discovery.

None

### DIFF
--- a/build/petalinux.build
+++ b/build/petalinux.build
@@ -1,1 +1,4 @@
-PETALINUX="/proj/petalinux/2022.1/petalinux-v2022.1_02160715/tool/petalinux-v2022.1-final"
+# When updating Petalinux build please file a SH ticket to retain the build
+# https://jira.xilinx.com/secure/CreateIssue!default.jspa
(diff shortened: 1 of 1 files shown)


xdavidz (approved):
Off the topic, why we change the source code every time? 
Can we have 2 options:
1) -daily_latest (by default)
2) -TA a specific TA

maxzhen (approved):

maxzhen:
> Off the topic, why we change the source code every time? Can we have 2 options:
> 
>     1. -daily_latest (by default)
> 
>     2. -TA a specific TA

daily latest is a moving target and it may not work for us. so we have only one option - a specific TA path. we'll have to explicitly update the path whenever we need.

PR #6376, opened 2022-03-07
Fix 1098396 : Skip adding rollover results as new profile monitors have 64 bit counter and unlikely to overflow

IshitaGhosh:
New version of profile monitors have 64 bit counters and are unlikely to overflow. So, no need to add the rollover results. This was causing unexpected increment in CU exec cycles and thus causing decreasing through put numbers reported in CR 1098396.
I found another issue during investigation. When profile_api s are enabled and executed, the profile monitor counters were getting cleared at the end of application due to call to "stopCounter". This is not desired behavior. The profile monitor counters are expected to hold their value even after end of application till the next load/reload of xclbin or board reset. This issue is also fixed here.


IshitaGhosh:
retest this please

jvillarre (approved):

PR #6384, opened 2022-03-09
Fix multiple PS kernel instance issue and some compile warning on zocl

mamin506:
1. When open SCU context on zocl, the CU index is always 0. This will lead to open context failure when open more than 1 scu context.
2. Cleanup some compile warnings, like unused variable, mixed declarations and code.

#6316 introduced the bug.

Assign cu_idx local variable.

No.

vck5000

### DIFF
--- a/src/runtime_src/core/common/drv/kds_core.c
+++ b/src/runtime_src/core/common/drv/kds_core.c
@@ -1641,8 +1641,9 @@ int kds_cfg_update(struct kds_sched *kds)
 			if (!xcu)
 				continue;
 			if (!xcu->info.intr_enable) {
-				kds->cu_intr = false;
 				u32 cu_idx = xcu->info.cu_idx;
+
+				kds->cu_intr = false;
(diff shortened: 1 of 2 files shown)


jeffli-xilinx (approved):

chvamshi-xilinx (approved):

src/runtime_src/core/common/drv/kds_core.c
```diff
@@ -1641,8 +1641,9 @@ int kds_cfg_update(struct kds_sched *kds)
 			if (!xcu)
 				continue;
 			if (!xcu->info.intr_enable) {
-				kds->cu_intr = false;
 				u32 cu_idx = xcu->info.cu_idx;
+
+				kds->cu_intr = false;
```
chvamshi-xilinx, line 1646: Just curious, why do we need to change the order ?

mamin506: @chvamshi-xilinx  This is to cleanup the compile warning about "mixed declarations and code" which is not allow in C90 standard.

Thanks for merging the change.

Issue #6282, opened 2022-02-16
Contributing a docker image to build XRT?


mgehre-amd:
I wanted to build XRT locally without installing the required dependencies into my system. Thus I created a small Dockerfile
```
FROM ubuntu:18.04

ENV LANG C.UTF-8
ENV DEBIAN_FRONTEND="noninteractive"

COPY ./src/runtime_src/tools/scripts/xrtdeps.sh .
RUN apt-get update \
    && ./xrtdeps.sh \
    && rm -rf /var/lib/apt/lists/* xrtdeps.sh
```
This is then run via
```
docker build . -t xrt-build
docker run --user $UID -i -t -v $PWD:$PWD -w $PWD/build xrt-build /bin/bash ./build.sh
```
in the top level folder. The resulting build is in `build/Debug` and  `build/Release`  on the host.

Is there interest in adding this to the repository and allowing others to build the same way?

wilderfield:
Hi @mgehre-amd,

I know that internally @NaAl-Xlnx leverages Docker for building XRT.

I think your suggestion is generally good for building XRT without disturbing a system’s dependencies. Is that your intent? Or, do you also intend to run applications from within this docker?

I think your Dockerfile could be improved with a build argument to support CentOS / Ubuntu 18 / Ubuntu 20 / Ubuntu X.

My only worry is that this method may not build and/or install the kernel modules successfully. The kernel modules are needed for actual communication with the FPGA.

There is a separate project that maintains Docker images with released XRT builds. Even there users are instructed to install XRT on the machine first (to bring in kernel modules).

 https://github.com/Xilinx/Xilinx_Base_Runtime

mgehre-amd:
Hi,
yes, this docker image is only meant for building. It's basically an extended version of `build.sh` which also takes care of installing dependencies in a way that doesn't pollute the users system.
It's true that I didn't use the kernel modules with my XRT build. I only tried sw_emu and hw_emu (via v++).

bryanloz-xilinx:
@mgehre-amd your idea inspired me today. Thanks.

Would you mind reviewing: #6380

bryanloz-xilinx:
@mgehre-amd @uday610 I think we can close this issue per: https://github.com/Xilinx/XRT/blob/master/src/runtime_src/tools/scripts/build_xrt_with_docker.sh

uday610:
Thanks

PR #6386, opened 2022-03-09
Add build_xrt_w_docker.sh

bryanloz-xilinx:
Build XRT with Docker, to avoid disturbing dependencies of system, and/or if you don't have sudo.

To build XRT with docker:
`./build_xrt_w_docker.sh`

This PR overrides https://github.com/Xilinx/XRT/pull/6380
It addresses issue: #6282 
This issue was just a feature request

In this solution, I use a singular bash script to create a temporary docker image (using xrtdeps.sh)
Then I use that docker image to build (using build.sh)
Build outputs are deposited in the usual place (XRT/build/Release XRT/build/Debug).

None

Tested build.

How should we advertise this tangential method of building?


bryanloz-xilinx:
This solution was inspired by @varunsh-xilinx

src/runtime_src/tools/scripts/build_xrt_w_docker.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (C) 2019-2022 Xilinx, Inc. All rights reserved.
+#
+
+# Must be Ubuntu 18.04 or Ubuntu 20.04
+BASE_IMAGE=$1:-ubuntu:18.04
```
stsoe, line 8: BASE_IMAGE is unused.

src/runtime_src/tools/scripts/build_xrt_w_docker.sh
```diff
@@ -0,0 +1,30 @@
...
+echo "Creating xrt-build-ubuntu Docker Image with xrtdeps.sh"
+
+CONTAINER_ID=$(docker run -d -e DEBIAN_FRONTEND=noninteractive -v $REPO_DIR:$REPO_DIR -w $XRTDEPS_DIR ubuntu:18.04 /bin/bash -c " apt-get update && ./xrtdeps.sh")
+
+docker wait $CONTAINER_ID
+
+docker commit $CONTAINER_ID xrt-build-ubuntu
+
+echo "Building XRT inside of Docker container"
+echo "Outputs will be in: $BUILD_DIR"
```
stsoe, line 27: You are at most creating a directory under BUILD_DIR, right?
Will that directory conflict with my native host folders Debug, Release, etc?  IN other words should you add a subfolder 'docker-<image>' maybe?

stsoe (changes requested):
Please add comments.

src/runtime_src/tools/scripts/build_xrt_w_docker.sh
bryanloz-xilinx: Thanks, good catch.

src/runtime_src/tools/scripts/build_xrt_w_docker.sh
bryanloz-xilinx: @stsoe no build directory is created.

The native host folder is mounted into the container `-v $REPO_DIR:$REPO_DIR`. This means that the absolute path of XRT will be mounted... for instance... it may be `/home/bryanloz/XRT`. This path will be visible both outside, and inside the container. `$BUILD_DIR` is just used to communicate the working directory.

The build outputs will get generated as usual in `/home/bryanloz/XRT/build/Debug` and `/home/bryanloz/XRT/build/Release`

In this approach the docker container will work seamlessly with the native host filesystem.

stsoe: @bryanloz-xilinx My point was that I may want to build XRT on my native host as well as docker.  I don't want docker to re-use my native builds Release and Debug directories, so move the docker build dirs one level down under a docker-<image> subdir is what I suggested.

stsoe: Bryan, please also rename with the script :-) what's the point of saving 3 characters?

bryanloz-xilinx:
> Please add comments.

Comments added

src/runtime_src/tools/scripts/build_xrt_w_docker.sh
bryanloz-xilinx: Script is renamed.

Docker is just calling XRT's `XRT/build/build.sh`, the output will go wherever that script wants to put it.

Is there a way to tell `XRT/build/build.sh` to put the build products somewhere else? If so, I can accommodate.

bryanloz-xilinx: Is it safe to play with these variables?
https://github.com/Xilinx/XRT/blob/master/build/build.sh#L188-L190

stsoe: Ahh yeah, got you about just using build.sh default locations.
Yes, to accommodate DevOps custom build dirs, env variables can be set prior to invoking build.sh.
DEBUG_DIR,  REL_DIR,  EDGE_DIR, so yes, use these.

src/runtime_src/tools/scripts/build_xrt_with_docker.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (C) 2019-2022 Xilinx, Inc. All rights reserved.
```
stsoe, line 4:
```suggestion
# Copyright (C) 2022 Xilinx, Inc. All rights reserved.
```

stsoe (changes requested):
Alright.  Thanks for the comments. Let's get the build dir fixed up to not clobber native build dirs and be specific to the docker image.  Also, please and option for just building debug or optimized (this is build.sh -dbg, or build.sh -opt).

bryanloz-xilinx:
@stsoe I tried using those environment (`DEBUG_DIR, REL_DIR, EDGE_DIR`) variables to steer the output, but I started seeing CMake errors. I think build.sh uses relative paths that don't work well with these variables.
https://github.com/Xilinx/XRT/blob/master/build/build.sh#L251

I view controlling the build output directory as overkill. I think its sort of nice that the outputs end up right where they are commonly expected.

src/runtime_src/tools/scripts/build_xrt_with_docker.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (C) 2022 Xilinx, Inc. All rights reserved.
+#
+
+# Must be Ubuntu 18.04 or Ubuntu 20.04
+# This will define the linux distro used inside of the docker container
+BASE_IMAGE=${1:-ubuntu:18.04}
```
rozumx, line 9: Comment on line #7 states that this docker is 18.04 or 20.04.   

How would one select one or the other? 

I ask for for this script only seems to support 18.04.

rozumx (approved):

src/runtime_src/tools/scripts/build_xrt_with_docker.sh
bryanloz-xilinx: @rozumx if invoked like this: `./build_xrt_with_docker.sh ubuntu:20.04` it would use the newer version of ubuntu.

However, based on Soren's earlier request (support dbg or opt), I may need to add real argument parsing.

bryanloz-xilinx:
@stsoe @rozumx 

After my latest commit, I've added argument parsing, such that a user can ask for only debug or only optimized versions.

The only feature I haven't implemented, is specifying a different output directory. I'm hoping we can leave that for future enhancement.

stsoe (approved):

PR #6394, opened 2022-03-10
CR-1114407 False positive on host_mem for u250_xdma_201830_2

chienwei-lan:
A false positive of host memory bandwidth test result. 

We check the entry "shared_host_mem" to decide whether we should test hostmem_bandwidth test, it sometime will be N/A if the shell has no address translator IP or not enable HOST[0].

we should check the entry "enabled_host_mem" to decide whether we should test the hostmem_bandwidth test.

The entry "enabled_host_mem" is determined by the existence of address translator IP and HOST[0].

If it has non-zero value, it means the shell has address translator IP, HOST[0] is enabled and user enables host memory by xbutil configure --host-mem

### DIFF
--- a/src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
+++ b/src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
@@ -1165,6 +1165,7 @@ hostMemBandwidthKernelTest(const std::shared_ptr<xrt_core::device>& _dev, boost:
   auto shared_host_mem = xrt_core::device_query<xrt_core::query::shared_host_mem>(_dev);
 
   try {
+    /* check if shell supports host mem */
     xrt_core::device_query<xrt_core::query::enabled_host_mem>(_dev);
   } catch(...) {
     if (shared_host_mem) {
--- a/src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
+++ b/src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
@@ -1162,16 +1162,16 @@ m2mTest(const std::shared_ptr<xrt_core::device>& _dev, boost::property_tree::ptr
 void
 hostMemBandwidthKernelTest(const std::shared_ptr<xrt_core::device>& _dev, boost::property_tree::ptree& _ptTest)
 {
-  uint64_t shared_host_mem = 0;
+  uint64_t enabled_host_mem = 0;
   try {
-    shared_host_mem = xrt_core::device_query<xrt_core::query::shared_host_mem>(_dev);
+    enabled_host_mem = xrt_core::device_query<xrt_core::query::enabled_host_mem>(_dev);
   } catch(...) {
     logger(_ptTest, "Details", "Address translator IP is not available");
     _ptTest.put("status", test_token_skipped);
     return;
   }
 
-  if (!shared_host_mem) {
+  if (!enabled_host_mem) {
       logger(_ptTest, "Details", "Host memory is not enabled");
       _ptTest.put("status", test_token_skipped);
       return;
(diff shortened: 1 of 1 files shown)


AShivangi (commented):
Does this solve the issue? 
If I understand correctly, enabled_host_mem only exists if the device supports host-mem. Let's take the case when the device doesn't support host-mem and the user successfully enabled host-mem. 
1. Will this sysfs node be available in that case?
2. If it is available, we need to throw an error and not skip. 

I believe xbutil needs some combination of "did user enable hostmem" and "does device support host-mem" to be able to determine whether to skip or throw an error.

chienwei-lan:
> enabled_host_mem only exists if the device supports host-mem. Let's take the case when the device doesn't support host-mem and the user successfully enabled host-mem.

Will this sysfs node be available in that case?
A: 1. If the shell doesn't have address translator IP, the node will not be available.
    2. If the shell has address translator IP, but HOST[0] is not used, the node will return 0
    3. The node only has a valid value (non-zero value) under these conditions 
      a. shell has address translator IP 
      and b. HOST[0] is enabled 
      and c. user enabled host-mem
    
If it is available, we need to throw an error and not skip.

chienwei-lan:
Discussed with Shivangi offline, if the user reserves the host memory by xbutil configure --host-mem and the device does not support address translation, XRT will error out instead of skip.

Test 10 [0000:65:00.1]    : hostmem-bw

    Description           : Run 'bandwidth kernel' when host memory is enabled
    Details               : Host memory is not enabled
Test 1 [0000:65:00.1]     : hostmem-bw
    Description           : Run 'bandwidth kernel' when host memory is enabled
    Xclbin                : /opt/xilinx/firmware/u50/gen3x16-xdma/base/test/bandwidth.xclbin
    Testcase              : /opt/xilinx/xrt/test/hostmemory.exe
    Details               : Throughput (Type: HOST) (Bank count: 1): 23291.6MB/s
Test 1 [0000:65:00.1]     : hostmem-bw

    Description           : Run 'bandwidth kernel' when host memory is enabled
    Details               : Address translator IP is not available

Negative Test
Test 1 [0000:65:00.1]     : hostmem-bw
    Description           : Run 'bandwidth kernel' when host memory is enabled
    Details               : Host memory is enabled, device does not support the feature
    Test Status           : [FAILED]

dayeh-xilinx:
retest this please.

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -1162,12 +1162,18 @@ m2mTest(const std::shared_ptr<xrt_core::device>& _dev, boost::property_tree::ptr
 void
 hostMemBandwidthKernelTest(const std::shared_ptr<xrt_core::device>& _dev, boost::property_tree::ptree& _ptTest)
 {
-  uint64_t shared_host_mem = 0;
+  auto shared_host_mem = xrt_core::device_query<xrt_core::query::shared_host_mem>(_dev);
+
   try {
-    shared_host_mem = xrt_core::device_query<xrt_core::query::shared_host_mem>(_dev);
+    xrt_core::device_query<xrt_core::query::enabled_host_mem>(_dev);
```
AShivangi, line 1168: let's add some comments of what we are checking
"// check if shell supports host mem"

AShivangi (approved):
please add a few comments for the future developer. The rest looks good, thanks

rozumx (approved):
Thank you Shivangi and Chien for your discussion around this issue.  While it did slow down this PR, it was the right thing to do.

maxzhen:
Marking as "do-not-merge" for now. There is no such concept as "host-mem supported by shell" any more. Xbutil validate code can never ever know if a shell can support host mem. So, it cannot decide whether to SKIP or FAIL. The test case's host code may decide. But in our case, there is no such test case at all, so validate can't tell.

PR #6396, opened 2022-03-10
CR-1125346 : VCK190_DFX: Second app is hanging on board

saifuddin-xilinx:
Issue : 
on DFX platform, We took two applications that are independently passing on vck190_dfx platform on board. Tried running two applications  one after the other without rebooting the board to check the true DFX functionality and observed that the second application is hanging. 

Root cause:
Cleanup is missing. While loading the next xclbin the cleanup should be done for the CUs.

Test: 
Working fine with this changes


chvamshi-xilinx (approved):
This is good, but it is better to have init call in loadxclbin which can take care of this. Please make that change in next PR.

mamin506 (approved):

PR #6401, opened 2022-03-11
CR-1120043 : Enhanced xbutil2 to show core information which is not enabled but buffer exists

saifuddin-xilinx:
This PR cover both the following CRs:
[CR-1120043] Missing "locks" section in aie_status_edge.json when Core is unused but buffers exists
[CR-1114191] Please remove the sentence "Pl Kernel Instances in Graph:" from xbutil report.

In AIE, there are some cores which is not enabled but memory buffer exists on those cores. 
In AIE metadata "dma_columns", "dma_row", which tells about all the cores where memory buffer present. 

We traverse all of that and show the lock, DMA information of those cores, even though those cores are not enable. 

Output 
After the regular xbutil2 output the following output is appended. 
Core [Buffer Only]
image

### DIFF
--- a/src/runtime_src/core/common/info_aie.cpp
+++ b/src/runtime_src/core/common/info_aie.cpp
@@ -420,6 +420,21 @@ populate_aie_shim(const xrt_core::device *device, const std::string& desc)
         ....
 */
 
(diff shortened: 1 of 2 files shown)


saifuddin-xilinx:
Will create a new one.

PR #6402, opened 2022-03-11
Uses Linux macros to swap bytes

DavidMarec:
[

Use linux macros to swap bytes.

Make it more portable.

use swab32s  if binary contents are swapped by bitstream generators, whatever the host endianness is.

Bitstream content are swapped , even if the host runs big endian.

Checkout on ZCU102 evaluation board, with custom dpu xclbins.

](https://github.com/Xilinx/XRT.git)


maxzhen (approved):

maxzhen:
ok to test

chvamshi-xilinx (approved):

PR #6407, opened 2022-03-11
CR-1125596 set fifo full flag only in hw flow

hackwa:
Problem : device trace full markers are being set in hw_emulation
Fix: Trace offloaders only set fifo_full flag for hw flow

Also includes some rearranging of code to make it cleaner.


dayeh-xilinx:
retest this please.

jvillarre (approved):

PR #6409, opened 2022-03-12
Disabling AIE resources and AIE error notification for AIE-ML

chvamshi-xilinx:
AIE-ML doesnt support AIE resources and error management, Disabled both of them for AIE gen-2

NA

 Disabled both of them for AIE gen-2 after reading xclbin

No

Verified both AIE-1 and AIE-2 

NA


chvamshi-xilinx:
Better to add ENUMs for AIE versions . Will add in next PR

PR #6412, opened 2022-03-14
Enabled xclbinutil for edge

chvamshi-xilinx:
xclbinutil is currently not enabled for edge. This is useful while debugging on target. We have removed the native build checks to build xclbinutil

NA

Enabled xclbinutil for edge

No

Verified on edge Target 

NA


stsoe (approved):

rozumx (approved):
This is a pleasant surprise. 
Thank you for doing this.

Regarding openssl, it looks like that this library may have already been apart of the edge environment.  If so, this is also a pleasant surprise for I was unware of that.

PR #6415, opened 2022-03-14
CR-1125635 dmesg log shows DDR data retention is disable while actually it is not

chienwei-lan:
Driver emits a log in dmesg indicating that DDR data retention feature is disabled which is not.


maxzhen (approved):

juliank-xilinx (approved):
The new message looks good to me, for the specific scope of CR-1125635. Thanks.

PR #6417, opened 2022-03-15
CR-1123904 XRT - Request a hot reset if XMC packet error received during SC upgrade

chienwei-lan:
During an SC upgrade, if XRT receives a CMC_HOST_MSG_SAT_FW_WRITE_FAIL(XMC Packet Error 4), it will resend the failed segment again which is not ideal. We should halt the upgrade and hot reset the board instead of resending the firmware segment.

XRT provides two ways to upgrade SC, one is via host by mapping the bar, the other one is via xmc driver.

1. Remove the retry part, when error happens during SC upgrade, we should stop upgrading SC instead of resending 
2. Set xmc state to XMC_STATE_ERROR in xmc driver and promote hot reset in dmesg


maxzhen (approved):

AShivangi (approved):
one comment, the rest looks good.

rozumx (changes requested):
Some minor changes.

rozumx (approved):

PR #6429, opened 2022-03-16
CR-1125922: Made sure warning pops up when erroneous aie_trace_start_time is added

nishraptor:
Added warning in case Regex cannot match start_time string at all.


nishraptor:
@pgschuey

pgschuey (approved):

jvillarre (approved):

PR #6430, opened 2022-03-16
setting versal DMA channel number should be after xdma_open

xuhz:
a workaround was introduced to set dma channel number as 2 for versal. But the setting before xdma_open will mess up interrupt allocation, so the setting should still be after xdma_open.


houlz0507 (approved):

maxzhen (approved):

dayeh-xilinx:
retest this please.

PR #6433, opened 2022-03-16
CR-1111561 platform meta data not written to flash with new commands

AShivangi:
flash the shell through driver so that the metadata gets written on the device

This has been ported from xbmgmt1 and is needed for removing legacy tools support

<to-be-added>

None

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/flash/flasher.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/flash/flasher.cpp
@@ -97,7 +97,7 @@ Flasher::E_FlasherType Flasher::getFlashType(std::string typeStr)
  * upgradeFirmware
  */
 int Flasher::upgradeFirmware(E_FlasherType flash_type,
-    firmwareImage *primary, firmwareImage *secondary)
+    firmwareImage *primary, firmwareImage *secondary, firmwareImage* stripped)
 {
     int retVal = -EINVAL;
 
@@ -112,11 +112,11 @@ int Flasher::upgradeFirmware(E_FlasherType flash_type,
         }
         else if(secondary == nullptr)
         {
-            retVal = xspi.xclUpgradeFirmware1(*primary);
+            retVal = xspi.xclUpgradeFirmware1(*primary, *stripped);
         }
         else
         {
-            retVal = xspi.xclUpgradeFirmware2(*primary, *secondary);
+            retVal = xspi.xclUpgradeFirmware2(*primary, *secondary, *stripped);
         }
 
         // program icap controller for webstar flow. Required only for U.2
--- a/src/runtime_src/core/tools/xbmgmt2/flash/flasher.h
+++ b/src/runtime_src/core/tools/xbmgmt2/flash/flasher.h
@@ -79,7 +79,7 @@ public:
 
     Flasher(unsigned int index);
     E_FlasherType getFlashType(std::string typeStr = "");
-    int upgradeFirmware(E_FlasherType flash_type, firmwareImage* primary, firmwareImage* secondary);
+    int upgradeFirmware(E_FlasherType flash_type, firmwareImage* primary, firmwareImage* secondary, firmwareImage* stripped);
     int upgradeBMCFirmware(firmwareImage* bmc);
     void readBack(const std::string& output);
     bool isValid(void) { return m_device != nullptr; }
--- a/src/runtime_src/core/tools/xbmgmt2/flash/xspi.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/flash/xspi.cpp
@@ -546,12 +546,12 @@ int XSPI_Flasher::xclTestXSpi(int index)
 }
 
 //Method for updating QSPI (expects 1 MCS files)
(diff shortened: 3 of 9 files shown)


src/runtime_src/core/tools/xbmgmt2/flash/flasher.cpp
```diff
@@ -97,7 +97,7 @@ Flasher::E_FlasherType Flasher::getFlashType(std::string typeStr)
  * upgradeFirmware
  */
 int Flasher::upgradeFirmware(E_FlasherType flash_type,
-    firmwareImage *primary, firmwareImage *secondary)
+    firmwareImage *primary, firmwareImage *secondary, firmwareImage* stripped)
```
rozumx, line 100: We have run out of time for this release so what I am about to mentioned cannot be done for this release....
This code should have been updated to pass references or use shared pointers.

src/runtime_src/core/tools/xbmgmt2/flash/flasher.h
```diff
@@ -79,7 +79,7 @@ class Flasher
```
rozumx, line 79: Copyright needs to be updated. A new line after 3 should be added to reflect AMD copyright. For example:
`Copyright (C) 2022 Advanced Micro Devices, Inc.`

src/runtime_src/core/tools/xbmgmt2/flash/xspi.cpp
```diff
@@ -546,12 +546,12 @@ int XSPI_Flasher::xclTestXSpi(int index)
 }
```
rozumx, line 546: Copyright needs to be updated. A new line after 3 should be added to reflect AMD copyright. For example:
`Copyright (C) 2022 Advanced Micro Devices, Inc.`

In addition, please remove the authors, there is no need to have this in the code base.

src/runtime_src/core/tools/xbmgmt2/flash/xspi.cpp
```diff
@@ -1854,6 +1854,26 @@ static int writeToFlash(std::FILE *flashDev, int slave,
     return ret;
 }
 
+static int 
+readFromFlash(std::FILE *flashDev, int slave,
```
rozumx, line 1858:
```suggestion
readFromFlash(std::FILE *flashDev, int secondary,
```

src/runtime_src/core/tools/xbmgmt2/flash/xspi.cpp
```diff
@@ -1854,6 +1854,26 @@ static int writeToFlash(std::FILE *flashDev, int slave,
     return ret;
 }
 
+static int
```
rozumx, line 1857:
```suggestion
void
```
The code should have been updated to throw and exception instead of using a return value.   Do you have time to clean this up?  If not, then could you create CR, assign it to yourself for 2022.2, with a high priority value.

src/runtime_src/core/tools/xbmgmt2/flash/xspi.cpp
```diff
@@ -2239,3 +2281,148 @@ int XSPI_Flasher::upgradeFirmware2Drv(std::istream& mcsStream0,
     // Disable bitstream guard.
     return removeBitstreamGuard(mDev.get(), mFlashDev, bsGuardAddr);
 }
+
+struct flash_data_ident {
+    char fdi_magic[7];
```
rozumx, line 2286: Magic number

src/runtime_src/core/tools/xbmgmt2/flash/xspi.cpp
```diff
@@ -2239,3 +2281,148 @@ int XSPI_Flasher::upgradeFirmware2Drv(std::istream& mcsStream0,
...
+flash_xrt_data_get_parity32(const unsigned char *buf, size_t n)
+{
+	uint32_t parity = 0;
+
+	for (size_t len = 0; len < n; len += 4) {
+		uint32_t tmp = 0;
+		size_t thislen = n - len;
+
+		/* One word at a time. */
+		if (thislen > 4)
```
rozumx, line 2309: Magic numbers

src/runtime_src/core/tools/xbmgmt2/flash/xspi.h
```diff
@@ -42,8 +42,8 @@ class XSPI_Flasher
  public:
```
rozumx, line 42: Copyright needs to be updated. A new line after 3 should be added to reflect AMD copyright. For example:
`Copyright (C) 2022 Advanced Micro Devices, Inc.`

rozumx (changes requested):
This code added is written in a C style and not C++.  For example:
1. Functions returning error values instead of throwing exceptions.
2. Use of pointers instead of shared pointers or references.
3. Magic numbers

While I understand that this code came from the legacy code base, it should have been updated to C++ when it was migrated.

AShivangi:
I have updated the copyrights and simple changes, however, code cleanup will take up some time. I have created a CR for the cleanup for early next release

rozumx (approved):
Thank you for making these changes, as well as, creating the CR to address this c style code being added to the C++ source code.  I am looking forward to seeing how you transform this code shortly after the release.

PR #6434, opened 2022-03-17
CR-1119432 xbmgmt configure missing data printouts

dbenusov-xilinx:
https://jira.xilinx.com/browse/CR-1119432
This is the second half of the CR after the issue was not initially solved! Specifically the VCK5000 was not outputting data correctly. This extended to the U30 as well after further testing.

The exception handling for the missing sysfs nodes output nothing instead of printing a not supported message. This has been present for a long while.

Moved printouts and set initial values for output to better mirror the state of the devices when data is requested.

None, just adding output. There is a separate issue of the security and runtime clock nodes not outputting any data if the value which should be noted in another CR.

On u280 with full support:
```
root@xsjpranjal01:~# xbmgmt configure --showx -d 65:00
0000:65:00.0
  Security level                   :
  Runtime clock scaling enabled    :
  Scaling threshold power override : 0
  Scaling threshold temp override  : 0
  Data retention                   : disabled

```
On vck5000 with no support:
```
dbenusov@xsjsagarw50:~$ xbmgmt configure --showx -d 65:00
0000:65:00.0
  Security level                   : Not supported
  Runtime clock scaling enabled    : Not supported
  Scaling threshold power override : Not supported
  Scaling threshold temp override  : Not supported
  Data retention                   : Not supported
```
On u30 with limited support:
```
dbenusov@xsjsagarw50:~$ xbmgmt configure --showx -d 17:00
0000:17:00.0
  Security level                   :
  Runtime clock scaling enabled    : Not supported
  Scaling threshold power override : Not supported
  Scaling threshold temp override  : Not supported
  Data retention                   : disabled
```
*Note the blank values! That will be a different CR that is soon to be filed.
None

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
@@ -407,6 +407,12 @@ SubCmdConfigure::execute(const SubCmdOptions& _options) const
 
     std::shared_ptr<xrt_core::device>& workingDevice = deviceCollection[0];
 
+    // If in factory mode the device is not ready for use
+    if (xrt_core::device_query<xrt_core::query::is_mfg>(workingDevice.get())) {
+      std::cout << boost::format("ERROR: Device is in factory mode and cannot be configured\n");
+      throw xrt_core::error(std::errc::operation_canceled);
+    }
+
     // Load Config commands
     // -- process "input" option -----------------------------------------------
     if (!path.empty()) {
--- a/src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
@@ -183,8 +183,8 @@ show_device_conf(xrt_core::device* device)
   if (is_mfg || is_recovery)
     throw xrt_core::error(std::errc::operation_canceled, "This operation is not supported with manufacturing image.\n");
 
-  boost::format sysfs_format("  %-33s: %s\n");
-  const std::string not_supported = "Not supported";
+  boost::format node_data_format("  %-33s: %s\n");
+  std::string not_supported = "Not supported";
   std::string sec_level = not_supported;
   try {
     sec_level = xrt_core::device_query<xrt_core::query::sec_level>(device);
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
```diff
@@ -183,46 +183,54 @@ show_device_conf(xrt_core::device* device)
   if (is_mfg || is_recovery)
     throw xrt_core::error(std::errc::operation_canceled, "This operation is not supported with manufacturing image.\n");
 
+  boost::format sysfs_format("  %-33s: %s\n");
```
AShivangi, line 186: sysfs is a linux specific term, can you rename it to something more generic like fmt_output

AShivangi (approved):
Looks good!

src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
```diff
@@ -183,46 +183,54 @@ show_device_conf(xrt_core::device* device)
   if (is_mfg || is_recovery)
     throw xrt_core::error(std::errc::operation_canceled, "This operation is not supported with manufacturing image.\n");
 
+  boost::format node_data_format("  %-33s: %s\n");
+  std::string not_supported = "Not supported";
+  std::string sec_level = not_supported;
```
AShivangi, line 188:
```suggestion
  uint16_t sec_level = not_supported;
```

src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
```diff
@@ -183,46 +183,54 @@ show_device_conf(xrt_core::device* device)
   if (is_mfg || is_recovery)
     throw xrt_core::error(std::errc::operation_canceled, "This operation is not supported with manufacturing image.\n");
 
+  boost::format node_data_format("  %-33s: %s\n");
+  std::string not_supported = "Not supported";
+  uint16_t sec_level = not_supported;
```
AShivangi, line 188:
```suggestion
  std::string sec_level = not_supported;
```

rozumx (approved):

PR #6436, opened 2022-03-17
CR-1112575 Enable Software resettable kernels on Edge platforms

rbramand-xilinx:
KDS support for abort command from user space is enabled in common code, this PR adds code to enable this feature in edge platforms.

Not a Bug.

We pass down the sw_reset bit to CU subdevice when xclbin is loaded, KDS will submit abort command to each CU and CU will be reset.

None

Tested on zcu102_base Vitis Embedded platform with abort test case (XRT/tests/xrt/abort) and things are working as expected

NA


chvamshi-xilinx (approved):
Thanks for making changes. Advantage of having common scheduler..

stsoe (approved):

mamin506 (approved):

PR #6448, opened 2022-03-18
xclbinutil : Merged perfoce changes into github.

rozumx:
The underlying code had a dependency regarding the order the files were linked.  Depending on their order, supporting section class objects were not registered to the base Section class.  For Vitis gradle builds, this caused an issue for the linking order was non-deterministic.   This change, made already in perforce, address this issue.

This is has always been apart of this code.

n/a

Low - This code has been running in Vitis for several months

Tested in the Vitis flow, manually tested, and unit-tested.

n/a

### DIFF
--- a/src/runtime_src/tools/xclbinutil/Section.cxx
+++ b/src/runtime_src/tools/xclbinutil/Section.cxx
@@ -1,5 +1,6 @@
 /**
- * Copyright (C) 2018 - 2021 Xilinx, Inc
+ * Copyright (C) 2018 - 2021 Xilinx, In
+ * Copyright (C) 2022 Advanced Micro Devices, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"). You may
  * not use this file except in compliance with the License. A copy of the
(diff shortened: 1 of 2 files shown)


src/runtime_src/tools/xclbinutil/Section.cxx
```diff
@@ -31,16 +32,15 @@ namespace XUtil = XclBinUtilities;
...
-std::map<enum axlf_section_kind, bool> Section::m_mapIdToSubSectionSupport;
-std::map<enum axlf_section_kind, bool> Section::m_mapIdToSectionIndexSupport;
+Section::FactoryData& Section::factoryData()
+{
+  // We allocate an object instead of using a local static to avoid potential
+  // order problems during teardown destruction. One might suggest this is a
+  // memory leak, but it will be reclaimed by the OS when the program exits.
+  // See https://isocpp.org/wiki/faq/ctors#construct-on-first-use-v2
+  static FactoryData* pData = new FactoryData();
+  return *pData;
```
stsoe, line 42: Not following what the problem might be.  If some static global object some place is referencing the FactoryData object then that would be bad and wrong.   By allocating on the heap, proper correct use of the 'singleton' is not enforced and just allows code to abuse global objects.

PR #6449, opened 2022-03-18
Add API to assign arguments from a void* + size_t

Ralender:
Expose a type-erased API to assign arguments.

### DIFF
--- a/src/runtime_src/core/include/xrt/xrt_kernel.h
+++ b/src/runtime_src/core/include/xrt/xrt_kernel.h
@@ -503,6 +503,21 @@ public:
   get_ert_packet() const;
   /// @endcond
 
+public:
+  // Use at your own risk, prefer documented type-safe arguments
+  void
+  set_arg(int index, const void* value, size_t bytes)
+  {
+    set_arg_at_index(index, value, bytes);
+  }
+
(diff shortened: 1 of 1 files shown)


stsoe (commented):
@Ralender Why would we want this API to be exposed?
It is private on purpose, so as to force usage through compile time types.

Ralender:
> @Ralender Why would we want this API to be exposed? It is private on purpose, so as to force usage through compile time types.

I need this because I am integrating XRT into a framework that already transforms a scalar argument into a void* + size because it is pretty generic.

keryell:
@Ralender Perhaps you could add in the function comments that, when possible, the type-safe API taking normal types should be used instead of the `void*` one. This is to address some internal discussion with @stsoe.

stsoe:
I understand the need.   Can we move these two functions under a public section immediately above the private section *and* not document this functions other than
`// Use at own risk, prefer documented type-safe arguments`
I really would prefer for these functions not to show up in the public generated documentation.

stsoe:
ok to test

stsoe (approved):

xilinxgitops:
retest this please

dayeh-xilinx:
retest this please.

PR #6460, opened 2022-03-21
Fix reading aie clock frequency

rbramand-xilinx:
Returning proper value when 'xbutil advanced --aie-clock --get' is called

Aie clock frequency is shown 0 every time because of wrong return value in PR - https://github.com/Xilinx/XRT/pull/6272

Returning proper value solved the issue

none

Tested on vck190 board with aie test case
root@versal-rootfs-common-20221:~# xbutil advanced --aie-clock --get 
WARNING: 'partition' option is not provided, using default partition id value '1'
INFO: Clock frequency of AIE partition(1) is: **1250.00 MHz**

NA


chvamshi-xilinx:
retest this please

PR #6461, opened 2022-03-21
CR-1126093 - PS Kernel - int32_t data type not recognized

rozumx:
When parsing the readelf (DWARF) metadata, the xclbinutil parser did not take into accout that DW_AT_typedef primitive argument size could be at a lower nested node.  This PR fixes this issue by parsing these metadata nodes.

The algorithm was created initially to no look at lower level nodes.  It was only when users started to explore the various levels of argument indirections (e.g., uint32_t -> __int32_t -> int) was this issue discovered.

Updated the algorihtm to support recursion of these nodes.

Low to none

Manual testing against user supplies test case

n/a


AShivangi (approved):

rozumx:
please retest

PR #6462, opened 2022-03-21
CR-1126642 Added support for user-specified AIE interface tile channels

pgschuey:
Added support for the following interface tile metric settings:
[Debug]
aie_profile = true
// For all slaves on channel 3
aie_profile_interface_metrics = input_bandwidths:3
// For all masters on channel 4
aie_profile_interface_metrics = output_bandwidths:4


pgschuey:
please re-test

jvillarre (approved):

PR #6464, opened 2022-03-21
CR 1123673 - ERROR: Property 'reg' data value '0x0000000000920000' exceeds the maximum uint64_t storage space.

rozumx:
There was a false error that occured when checking the return value from std::strtoul().  This function only sets the `errno `variable if an range error occured.  In this case, a previous math function was called that set this value.  It unfortunately wasn't cleared after being resolved, hence this issue.

This issue has been in the code base since the original creation of the partition_metadata section (many years).

The fix was to set the errno value to zero prior to calling std::strtol().  

None

Manual testing

None


rozumx:
please retest

PR #6466, opened 2022-03-21
CR-1119433 Improve error message for xbmgmt configure commands

dbenusov-xilinx:
Running "sudo xbmgmt configure --runtime_clk_scale enable" returns:
```
XRT build version: 2.13.260
Build hash: 6820e65a07311528963d49cf3195821069a72ed4
Build date: 2021-12-16 21:24:14
Git branch: master
PID: 30334
UID: 0
[Wed Jan 12 10:59:12 2022 GMT]
HOST: xniengxbb39
[xbmgmt] ERROR: Failed to find subdirectory for xmc under /sys/bus/pci/devices/0000:c1:00.0
```
This is applicable to the retention option and all the hidden options for the xbmgmt configure command.

If a sysfs node is missing when the xbmgmt configure command is run an access error would occur vs a graceful stop to the program.

Added try/catch handling for any missing sysfs nodes and printed a more helpful message within xbmgmt configure.

None, added better error handling.

Tested on vck5000
Runtime Clock Scaling
```
root@xsjsagarw50:~# xbmgmt configure -d 65:00 --runtime_clk_scale enable
ERROR: Device does not support runtime clock scaling

root@xsjsagarw50:~# echo $?
1
```
Security
```
root@xsjsagarw50:~# xbmgmt configure -d 65:00 --security enable
ERROR: Device does not support security

root@xsjsagarw50:~# echo $?
1
```
Threshold Power Override
```
root@xsjsagarw50:~# xbmgmt configure -d 65:00 --cs_threshold_power_override enable
ERROR: Device does not support threshold power override

root@xsjsagarw50:~# echo $?
1
```
Threshold Temperature Override
```
root@xsjsagarw50:~# xbmgmt configure -d 65:00 --cs_threshold_temp_override enable
ERROR: Device does not support threshold temp override

root@xsjsagarw50:~# echo $?
1
```
Clock Scaling Option Reset
```
root@xsjsagarw50:~# xbmgmt configure -d 65:00 --cs_reset enable
ERROR: Device does not support clock scaling option reset

root@xsjsagarw50:~# echo $?
1
```
Memory Retention
```
root@xsjsagarw50:~# xbmgmt configure -d 65:00 --retention ENABLE
ERROR: Device does not support memory retention

root@xsjsagarw50:~# echo $?
1
```
Non-root attempts to configure
```
dbenusov@xsjpranjal01:/proj/xsjhdstaff6/dbenusov/XRT$ xbmgmt configure -d 65:00 --retention enable
ERROR: Updating memory retention requires sudo
dbenusov@xsjpranjal01:/proj/xsjhdstaff6/dbenusov/XRT$ xbmgmt configure -d 65:00 --cs_reset enable
ERROR: Updating device configuration requires sudo
```
None

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
@@ -287,7 +287,7 @@ update_device_conf(xrt_core::device* device, const std::string& value, config_ty
         xrt_core::device_update<xrt_core::query::xmc_scaling_reset>(device, value);
         break;
     }
-  } catch (const std::exception& ex) {
+  } catch (const std::exception&) {
     std::cerr << boost::format("ERROR: Device does not support %s\n\n") % config_to_string(cfg);
     throw xrt_core::error(std::errc::operation_canceled);
   }
@@ -305,7 +305,7 @@ memory_retention(xrt_core::device* device, bool enable)
   try {
     auto value = xrt_core::query::data_retention::value_type(enable);
     xrt_core::device_update<xrt_core::query::data_retention>(device, value);
-  } catch (const std::exception& ex) {
+  } catch (const std::exception&) {
     std::cerr << boost::format("ERROR: Device does not support memory retention\n\n");
     throw xrt_core::error(std::errc::operation_canceled);
   }
--- a/src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
@@ -45,7 +45,7 @@ enum class config_type {
 
 static
 std::string
-to_string(config_type value)
+config_to_string(config_type value)
 {
   static std::map<config_type, std::string> config_map =
   {
@@ -53,7 +53,7 @@ to_string(config_type value)
    { config_type::clk_scaling,                "runtime clock scaling" },
    { config_type::threshold_power_override,   "threshold power override" },
    { config_type::threshold_temp_override,    "threshold temp override" },
-   { config_type::reset,                      "reset" },
+   { config_type::reset,                      "clock scaling option reset" },
   };
 
   return config_map[value];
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
```diff
@@ -280,8 +302,13 @@ static void
 memory_retention(xrt_core::device* device, bool enable)
 {
   XBU::sudo_or_throw("Updating memory retention requires sudo");
+  try {
   auto value = xrt_core::query::data_retention::value_type(enable);
```
dbenusov-xilinx, line 306: Add an indent here

AShivangi:
retest this please

src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
```diff
@@ -43,6 +43,22 @@ enum class config_type {
     reset
 };
 
+static
+std::string
+config_to_string(config_type value)
```
rozumx, line 48: While this code is indeed functionality, it will have issues scaling.  For example, if there is a new enumerated config type, it will:
1. Create a new empty entry into the map.
2. Return back that empty string name.

Instead, this code should be updated to help the next developer discover that the need to update this collection.  

Some ideas to think about:
1. One might original think that making the std::map const would address the issue, it will not and the build will fail for [] cannot be const for the reason mentioned earlier.
2. Make the collection a vector of pairs and create a lamda to search for the key, if not found  then throw an exception.
3. Create a switch statement.  The compilers are smart enought to recongize when a enumeration is missing and will produce an compile error. 

Each solutions has its own pros and cons.  I will let you decide which to choose.

rozumx (changes requested):

src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
dbenusov-xilinx: Makes sense to me. I was looking up some good ways to do this but the best choice was not obvious, thank you for giving me a write-up about which ones are better/worse.
I went with a switch case as I prefer the simpler syntax

rozumx (approved):

src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
```diff
@@ -43,6 +43,31 @@ enum class config_type {
...
+    case config_type::threshold_power_override:
+      os << "threshold power override";
+      break;
+    case config_type::threshold_temp_override:
+      os << "threshold temp override";
+      break;
+    case config_type::reset:
+      os << "clock scaling option reset";
+      break;
+    default:
```
rozumx, line 65: While I have approved the PR, this default case is not needed.  The compiler will detect a missing enumeration.

PR #6468, opened 2022-03-22
CR-1126538 ASTeR TC1.06 - u50 - cannot upgrade from factory SC version, xocl and SC left in abnormal state

chienwei-lan:
Compare to normal SC,  SC in factory mode takes way longer time to be ready. 

It takes nearly 30 seconds. However, we only wait 5 secs in xbmgmt2.

Don't have the issue in xbmgmt, it goes through xmc driver to flash sc firmware and waits up to 60 secs.

Low

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/flash/xmc.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/flash/xmc.cpp
@@ -464,7 +464,10 @@ int XMC_Flasher::sendPkt(bool print_dot)
 
 int XMC_Flasher::waitTillIdle()
 {
-    // In total, wait for 5000 * 10ms
+    /* In total, wait for 5000 * 10ms
+     * SC in factory mode takes nearly 30 seconds to be idle
+     * We relax the waiting time here to up to 50 seconds.
+     */
     int retry = 5000;
     unsigned int err = 0;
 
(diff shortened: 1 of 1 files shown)


AShivangi (commented):
Now we wait for 50 seconds? isn't that a little too long?

maxzhen:
> Now we wait for 50 seconds? isn't that a little too long?

If the HW takes 30 seconds to finish (this is the fact based on real test result), it is normal practice to wait for twice as long before you claim timeout. And this is why driver waits for 60 seconds.
Total wait time is 50 seconds does not mean you always wait for 50 seconds. If HW finish sooner, you return sooner. 50 seconds is set here to make sure you don't claim time out prematurely when HW is still doing the job.
What is your concern for setting the maximum waiting time to 50 seconds? @AShivangi

AShivangi:
> > Now we wait for 50 seconds? isn't that a little too long?
> 
> If the HW takes 30 seconds to finish (this is the fact based on real test result), it is normal practice to wait for twice as long before you claim timeout. And this is why driver waits for 60 seconds. Total wait time is 50 seconds does not mean you always wait for 50 seconds. If HW finish sooner, you return sooner. 50 seconds is set here to make sure you don't claim time out prematurely when HW is still doing the job. What is your concern for setting the maximum waiting time to 50 seconds? @AShivangi

My concern was that we increased the wait time 10 folds without an explanation. As long as we have an explanation it's fine. @chienwei-lan, can you please add a comment about the HW taking 30 seconds?

rozumx (approved):

rozumx:
What is being changes is the timeout value.  Since the code checks this value 100 times a second, as soon as the the SC is ready it will nicly continue.  Again, what is being taken into account are unexpected delays that might occur.  For example, programming several devices in parallel.  This timeout should not happen, but if it does, it indicates something bad has happened, so waiting 2x longer is good.

AShivangi (approved):

dayeh-xilinx:
retest this please.

PR #6474, opened 2022-03-23
Resolved the result mismatch issue for CR-1126579

subeeshm-xilinx:
Resolved the result mismatch for P2P testcases by adding condition in xclCopyBO such that the control will not go to m2m logic when P2P transfer is involved
This is an issue which was caught as part of the CR-1126579
Low Risk
We have run the Canary suite and also verified various combination testcases of P2P & M2M manually
No impact on Documentation


chvamshi-xilinx (approved):

dayeh-xilinx:
retest this please.

subeeshm-xilinx:
@stsoe Will this be merged automatically to 2022.1 branch?

PR #6482, opened 2022-03-23
Remove unused 21.2 user mode PS kernel code

stsoe:
The POC code added in 21.2 related to ps kernel can be removed
after Vitis-3342.


jeffli-xilinx (approved):

rozumx (approved):
I always like removal of code.  Makes code reviews so much easier :^).

PR #6492, opened 2022-03-25
CR-1120378 : RPM packager using wrong python version (was: xbtop doesn't run on CentOS 7)

rozumx:
Debugging issue with regards to with enabling python3 scripts to be packaged with CMake.  In this case, it is using the python2 to validate that the sources are valid (instead of python3).

When packaging a RPM package on CentOS, the wrong version of python (e.g., python2) was being used to compile (e.g., validate) the python scripts.   For this reason, CentOS packaging of xbtop was previously disabled.

This PR both addresses this issue (e.g., python3 is now used) and re-enables packaging the xbtop for CentOS packages.

CR-1120378

Updated CPack to instruct the rpm packager to use python3.

None

Manual testing the the XRT-Pipeline

n/a


src/CMake/cpackLin.cmake
```diff
@@ -19,6 +19,9 @@ SET(CPACK_ARCHIVE_COMPONENT_INSTALL ON)
 SET(CPACK_DEB_COMPONENT_INSTALL ON)
 SET(CPACK_RPM_COMPONENT_INSTALL ON)
 
+# Force python byte compilation to use python3
```
stsoe, line 22: I think your commit description is better than this comment line.
I don't really know what this line does by looking at it, but you explained it in the description.  Could you add more details here?

rozumx:
retest this please

AShivangi (approved):
@uday610, we are reenabling xbtop on CenOS. Please keep an eye on this PR and update the documentation accordingly.

stsoe (approved):

uday610:
@AShivangi , so we dont need to mention any limitation (that needed to copy python file etc) for Centos right?

dayeh-xilinx:
retest this please.

PR #6501, opened 2022-03-28
CR-1126396 xbtop output does not look correct

dbenusov-xilinx:
https://jira.xilinx.com/browse/CR-1126396
When xbtop is running for a u55c board, the output looks keep rolling and most data is not visible

Discovered on cards that have a large memory bank to display after running a validation test

The problem was solved by splitting up the data displays onto multiple pages the users can go through. The memory report was consolidated into two internal reports rather than three.

Risk of data not being presented properly using xbtop if an unexpected case appears. Possible crashes in these scenarios as well. Resizing the window does not dynamically reprint data, the program must be stopped and restarted.

Remory report pages:
image
image
image
Resized terminal window:
image
Report with single page:
image
Report with no xclbin loaded:
image
image
image
Updated memory utilization to have percentages:
image

None, unless we would like to document the paging feature.

### DIFF
--- a/src/runtime_src/core/common/info_memory.cpp
+++ b/src/runtime_src/core/common/info_memory.cpp
@@ -468,7 +468,7 @@ scheduler_update_stat(const xrt_core::device *device)
     // lock xclbin
     std::string xclbin_uuid = xrt_core::device_query<xq::xclbin_uuid>(dev);
     // dont open a context if xclbin_uuid is empty
-    if(xclbin_uuid.empty())
+    if(xq::xclbin_uuid::is_empty(xclbin_uuid))
       return;
     auto uuid = xrt::uuid(xclbin_uuid);
     dev->open_context(uuid.get(), std::numeric_limits<unsigned int>::max(), true);
--- a/src/runtime_src/core/common/info_memory.cpp
+++ b/src/runtime_src/core/common/info_memory.cpp
@@ -467,8 +467,8 @@ scheduler_update_stat(const xrt_core::device *device)
   try {
     // lock xclbin
     std::string xclbin_uuid = xrt_core::device_query<xq::xclbin_uuid>(dev);
-    // dont open a context if xclbin_uuid is empty
-    if(xq::xclbin_uuid::is_empty(xclbin_uuid))
+    // dont open a context if xclbin_uuid is empty or is all zeros
+    if (xclbin_uuid.empty() || !xrt::uuid(xclbin_uuid))
       return;
     auto uuid = xrt::uuid(xclbin_uuid);
     dev->open_context(uuid.get(), std::numeric_limits<unsigned int>::max(), true);
--- a/src/runtime_src/core/common/query_requests.h
+++ b/src/runtime_src/core/common/query_requests.h
@@ -29,6 +29,7 @@
 #include <stdexcept>
 #include <boost/any.hpp>
 #include <boost/format.hpp>
+#include <boost/algorithm/string.hpp>
 
 struct debug_ip_data;
 
@@ -698,6 +699,12 @@ struct xclbin_uuid : request
 
   virtual boost::any
   get(const device*) const = 0;
+  
+  static bool
+  is_empty(const std::string& uuid) 
+  {
+    return (boost::equals(uuid, "00000000-0000-0000-0000-000000000000") || boost::equals(uuid, ""));
+  }
 };
 
 // dtbo_path is unique path used by libdfx library to load bitstream and device tree
--- a/src/runtime_src/core/common/query_requests.h
+++ b/src/runtime_src/core/common/query_requests.h
@@ -699,12 +699,6 @@ struct xclbin_uuid : request
 
   virtual boost::any
   get(const device*) const = 0;
-  
-  static bool
-  is_empty(const std::string& uuid) 
-  {
-    return (boost::equals(uuid, "00000000-0000-0000-0000-000000000000") || boost::equals(uuid, ""));
-  }
 };
 
 // dtbo_path is unique path used by libdfx library to load bitstream and device tree
--- a/src/runtime_src/core/common/query_requests.h
+++ b/src/runtime_src/core/common/query_requests.h
@@ -29,7 +29,6 @@
 #include <stdexcept>
 #include <boost/any.hpp>
 #include <boost/format.hpp>
-#include <boost/algorithm/string.hpp>
 
 struct debug_ip_data;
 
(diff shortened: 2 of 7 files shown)


dbenusov-xilinx:
> Looks good to me. Can you test when no xclbin is loaded? just a sanity check Also, should "Utilization" have "%" in it?

Good suggestions! I added a test for no xclbin and found a bug present in xbtop which I fixed in here. Also added the '%' character into the utilization printout

src/runtime_src/core/common/info_memory.cpp
```diff
@@ -468,7 +468,7 @@ scheduler_update_stat(const xrt_core::device *device)
     // lock xclbin
     std::string xclbin_uuid = xrt_core::device_query<xq::xclbin_uuid>(dev);
     // dont open a context if xclbin_uuid is empty
-    if(xclbin_uuid.empty())
+    if(xq::xclbin_uuid::is_empty(xclbin_uuid))
       return;
```
stsoe, line 472: Space between `if` and `(` always please.

Do just
```
if (xclbin_uuid.empty() || !xrt::uuid(xclbin_uuid))
  return;
```

src/runtime_src/core/common/query_requests.h
```diff
@@ -698,6 +699,12 @@ struct xclbin_uuid : request
 
   virtual boost::any
   get(const device*) const = 0;
+  
+  static bool
+  is_empty(const std::string& uuid) 
+  {
+    return (boost::equals(uuid, "00000000-0000-0000-0000-000000000000") || boost::equals(uuid, ""));
+  }
```
stsoe, line 707: Remove

stsoe (changes requested):

rozumx (changes requested):
Please schedule a demo.

src/runtime_src/core/common/query_requests.h
dbenusov-xilinx: And here Shivangi and I though we invented the wheel! Cool feature with the boolean overload on the uuid :)

src/runtime_src/core/common/query_requests.h
```diff
@@ -29,6 +29,7 @@
 #include <stdexcept>
 #include <boost/any.hpp>
 #include <boost/format.hpp>
+#include <boost/algorithm/string.hpp>
```
stsoe, line 32: Is this used?

stsoe (changes requested):

stsoe (approved):
Python is beyond me and you got rid of all C++ changes :-)

AShivangi (approved):

src/runtime_src/core/common/info_memory.cpp
```diff
@@ -467,8 +467,8 @@ scheduler_update_stat(const xrt_core::device *device)
   try {
     // lock xclbin
     std::string xclbin_uuid = xrt_core::device_query<xq::xclbin_uuid>(dev);
-    // dont open a context if xclbin_uuid is empty
-    if(xclbin_uuid.empty())
+    // dont open a context if xclbin_uuid is empty or is all zeros
+    if (xclbin_uuid.empty() || !xrt::uuid(xclbin_uuid))
```
rozumx, line 471: What was this change required?

src/runtime_src/core/tools/xbtop/ReportMemory.py
```diff
@@ -28,108 +29,186 @@ def update(self, dev):
...
+            XBUtil.print_warning(term, lock, start_y + offset, "Data unavailable. Acceleration image not loaded")
+            return offset + 1
+
+        for i in range(self.report_length):
+            # The current element to be parsed depends on what page has been requested
+            index = i + (page * self.report_length)
+            # Ensure that our index does not exceed the input data size. This may happen on the last page
+            if (index < len(memories)):
+                memory = memories[index]
+                size = int(memory['range_bytes'], 16)
```
rozumx, line 72: Magic number

src/runtime_src/core/tools/xbtop/ReportMemory.py
```diff
@@ -28,108 +29,186 @@ def update(self, dev):
...
+            if (index < len(memories)):
+                line.append(str(index))
+                memory = memories[index]
+                line.append(memory['tag'])
+                line.append(memory['type'])
+                line.append(memory['extended_info']['temperature_C'] if 'temperature_C' in memory['extended_info'] else "--")
+                line.append(XBUtil.convert_size(int(memory['range_bytes'],16)))
+                line.append(XBUtil.convert_size(int(memory['extended_info']['usage']['buffer_objects_count'],0)))
+                
+                size = int(memory['range_bytes'], 16)
```
rozumx, line 124: Magic number

src/runtime_src/core/tools/xbtop/ReportMemory.py
```diff
@@ -28,108 +29,186 @@ def update(self, dev):
...
-            line.append(XBUtil.convert_size(int(dma_metrics[i]['card_to_host_bytes'], 16)))
-            data.append(line)
+            # The current element to be parsed depends on what page has been requested
+            index = i + (page * self.report_length)
+            # Ensure that our index does not exceed the input data size. This may happen on the last page
+            if (index < len(dma_metrics)):
+                dma_metric = dma_metrics[index]
+                line.append(str(i))
+                line.append(dma_metric['channel_id'])
+                line.append(XBUtil.convert_size(int(dma_metric['host_to_card_bytes'], 16)))
```
rozumx, line 178: Magic number

src/runtime_src/core/tools/xbtop/xbtop.py
```diff
@@ -66,7 +67,8 @@ def running_counter(term, lock, len_x):
 # Key table
 def print_footer(term, lock, y_len):
     footer_buf = [
-        XBUtil.fg.yellow + "'n'/'p' - Next/Previous Report" + XBUtil.fg.reset,
+        XBUtil.fg.yellow + "'p'/'n' - Previous/Next Report" + XBUtil.fg.reset,
```
rozumx, line 70: Thank you for swapping the 'p' and 'n'.  It helps keep the menu and the key description consistent.

rozumx (approved):
Overall looks good, though it could some some additional comments.

src/runtime_src/core/common/info_memory.cpp
dbenusov-xilinx: There was an bug on the dynamic report page that would cause a crash due to the xclbin uuid being all 0s. Since the uuid is not "empty" the software in C++ and Python would attempt to open context on a non existent file which caused strange issues. C++ it would crash when opening the context, in Python is would crash when closing which failed to catch for one reason or another.

PR #6507, opened 2022-03-29
Fix calculation of frequency value in setting aie clock frequency

rbramand-xilinx:
When aie clock freq is set using xbutil we pass a string value (eg: 100M), earlier string_to_bytes is used for calculation which uses 1024 as conversion factor. Modified string_to_bytes function to convert frequency as well.

https://github.com/Xilinx/XRT/pull/6272 introduced this bug

Fixed frequency calculation by using correct conversion factor

NA

Tested setting aie clock frequency on vck190 board and behavior is as expected

NA

### DIFF
--- a/src/runtime_src/core/tools/common/XBUtilities.h
+++ b/src/runtime_src/core/tools/common/XBUtilities.h
@@ -90,7 +90,7 @@ namespace XBUtilities {
   enum class unit
   {
     bytes,
-    hertz
+    Hertz
   };
   uint64_t
   string_to_base_units(std::string str, const unit& conversion_unit);
--- a/src/runtime_src/core/tools/common/XBUtilities.h
+++ b/src/runtime_src/core/tools/common/XBUtilities.h
@@ -87,8 +87,13 @@ namespace XBUtilities {
   std::string
   string_to_UUID(std::string str);
 
-  uint64_t 
-  string_to_bytes(std::string str);
+  enum class unit
+  {
+    bytes,
+    hertz
+  };
+  uint64_t
+  string_to_base_units(std::string str, const unit& conversion_unit);
 
   inline bool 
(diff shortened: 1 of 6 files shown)


chvamshi-xilinx (approved):

rbramand-xilinx:
> I have https://jira.xilinx.com/browse/CR-1127118 for next release to get rid of this function. @stsoe, what should be the correct way to handle this?

@AShivangi, But in my case I need frequency conversion at runtime based on user input, the CR you are talking about is for constant values whose conversion can be moved to compile time

rbramand-xilinx:
@AShivangi this fix is for 2022.1, so can we go ahead with this change for now and discuss in next release for the fix when you handle your CR

src/runtime_src/core/tools/common/XBUtilities.h
```diff
@@ -87,8 +87,13 @@ namespace XBUtilities {
   std::string
   string_to_UUID(std::string str);
 
-  uint64_t 
-  string_to_bytes(std::string str);
+  enum class unit
+  {
+    bytes,
+    hertz
```
rozumx, line 93: Hertz should be capatialized for it is named after a person: Heinrich Rudolf Hertz.

rozumx (approved):

src/runtime_src/core/tools/common/XBUtilities.h
rbramand-xilinx: Thanks for pointing it out @rozumx , I have made changes

AShivangi:
> @AShivangi this fix is for 2022.1, so can we go ahead with this change for now and discuss in next release for the fix when you handle your CR

Sure

AShivangi (approved):

PR #6509, opened 2022-03-29
Get CU functional from kernel xml metadata

chienwei-lan:

Parse the kernel xml metadata to get the value of functional and add it to struct kernel_properties

### DIFF
--- a/src/runtime_src/core/common/xclbin_parser.cpp
+++ b/src/runtime_src/core/common/xclbin_parser.cpp
@@ -230,7 +230,7 @@ kernel_max_ctx(const ip_data& ip)
 }
 
 //Get the cu functional from kernel xml entry
-static uint32_t
+static size_t
 get_functional(const pt::ptree& xml_kernel, const std::string& element)
 {
   for (auto& elem : xml_kernel) {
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/common/xclbin_parser.cpp
```diff
@@ -229,6 +229,20 @@ kernel_max_ctx(const ip_data& ip)
...
 
+//Get the cu functional from kernel xml entry
+static uint32_t
+get_functional(const pt::ptree& xml_kernel, const std::string& element)
+{
+  for (auto& elem : xml_kernel) {
+    if (elem.first != element)
+      continue;
+
+    return convert(elem.second.get<std::string>("<xmlattr>.functional"));
```
stsoe, line 240: The convert function returns size_t, maybe `static_cast<uint32_t>(convert(...))` just to be explicit.
Without cast may fail on some compiler?

stsoe (changes requested):

src/runtime_src/core/common/xclbin_parser.cpp
chienwei-lan: Yes, compiler does complain about the data type conversion may lose data.

I just changed it from uint32_t to size_t

rozumx (approved):

PR #6537, opened 2022-04-01
Fix SOM multiple bitstream programming for full PL DFx shells

rbramand-xilinx:
In SOM we have added a new feature to program multiple xclbins for full PL DFx shells by using device tree overlay, when new xclbin with OVERLAY section is loaded overlay that is already applied is removed. zocl node (zyxclmm_drm) is part of Overlay for this kind of platforms, so zocl node is removed and zocl module is unloaded which causes kernel crash. This PR fixes this kernel crash by checking zyxclmm_drm is present before accessing it. 

When loading different xclbins with overlay sections it was discovered that during driver unload we are accessing zyxclmm_drm device tree node without checking for its presence.

Added check for NULL pointer

None as we are just checking for NULL pointer

Tested on SOM board by loading different xclbins in a loop for 200 times, it did not cause any crash

NA

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/cu.c
+++ b/src/runtime_src/core/edge/drm/zocl/cu.c
@@ -291,7 +291,8 @@ static int cu_remove(struct platform_device *pdev)
 		free_irq(zcu->irq, zcu);
 
 	zdev = zocl_get_zdev();
-	zocl_kds_del_cu(zdev, &zcu->base);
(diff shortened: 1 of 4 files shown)


chvamshi-xilinx (approved):

saifuddin-xilinx (approved):
I believe, this is a workaround. We need to revisit this solution for a proper fix.

PR #6542, opened 2022-04-01
hwmon_sdm: maintain separate cache timer for each SDR type

rajkumar-xilinx:

At present, hwmon_sdm driver is having single cache timer for all SDR types. It will not work for below case.
1. repo_id = 0x2 is received in get_sensors_data() function
2. It checks if the cache timer is expired. If yes, it will update the sensor list for repo_id = 0x2 and reset cache timer.
3. Within cache timer limits, a new repo_id say 0x3 is received in get_sensors_data() function
4. Since, timer is not expired, it returns 0 without updating sensor list for repo_id 0x3. This is an Issue.

This PR will fixes this scenario by adding cache timer for each SDR type.

Added changes in hwmon_sdm driver.

NA
xbutil & xbmgmt examine reports and able to see all SDR types sensors are getting updated if cache timer expired.

NA

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
@@ -71,7 +71,7 @@ struct xocl_hwmon_sdm {
 
 	struct mutex            sdm_lock;
 	u64                     cache_expire_secs;
-	ktime_t                 cache_expires;
+	ktime_t                 cache_expires[SDR_TYPE_MAX];
 };
 
 #define SDM_BUF_IDX_INCR(buf_index, len, buf_len) \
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx (approved):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
```diff
@@ -180,10 +180,10 @@ static int get_sdr_type(enum xcl_group_kind kind)
 	return type;
 }
 
-static void update_cache_expiry_time(struct xocl_hwmon_sdm *sdm)
+static void update_cache_expiry_time(struct xocl_hwmon_sdm *sdm, uint8_t repo_id)
 {
-	sdm->cache_expires = ktime_add(ktime_get_boottime(),
-                                   ktime_set(sdm->cache_expire_secs, 0));
+	sdm->cache_expires[repo_id] = ktime_add(ktime_get_boottime(),
+                                      ktime_set(sdm->cache_expire_secs, 0));
```
stsoe, line 186: What guarantee is there that repo_id is with the range of the array?

stsoe (changes requested):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
rajkumar-xilinx: Hi Soren, 

While parsing sensor buffer in hwmon_sdm driver, it defined  sdr_get_id() API which takes repo_type input and returns "repo_id" value which is in the range of [0, SDR_TYPE_MAX). This same value is being passed to this update_cache_expiry_time() also. 

For safe side, I added a if condition for repo_id in hwmon_sensor_show() if it is in valid range or not. Please review it again, thank you.

stsoe (approved):

maxzhen (approved):

PR #6543, opened 2022-04-01
Update xclLoadXclbin api documentation to include about OVERLAY section

rbramand-xilinx:
Updated xclLoadXclbin api documentation to include OVERLAY section loading use case for edge platforms

NA

updated documentation in xrt.h file

NA

NA

NA

Sample o/p after making this change:
image

### DIFF
--- a/src/runtime_src/core/include/xrt.h
+++ b/src/runtime_src/core/include/xrt.h
@@ -335,9 +335,8 @@ xclGetErrorStatus(xclDeviceHandle handle, struct xclErrorStatus *info);
  * @buffer:        Pointer to device image (xclbin) in memory
  * Return:         0 on success or appropriate error number
  *
- * Download FPGA image (AXLF) to the device. The PR bitstream is
- * encapsulated inside xclbin as a section. xclbin may also contains
- * other sections which are suitably handled by the driver.
+ * | Download FPGA image (AXLF) to the device. The PR bitstream is encapsulated inside xclbin as a section. xclbin may also contains other sections which are suitably handled by the driver.
+ * | **NOTE**: This api also downloads OVERLAY(dtbo) section (Applicable only for Edge Platforms)
  */
 XCL_DRIVER_DLLESPEC
 int
--- a/src/runtime_src/core/include/xrt.h
+++ b/src/runtime_src/core/include/xrt.h
@@ -335,8 +335,9 @@ xclGetErrorStatus(xclDeviceHandle handle, struct xclErrorStatus *info);
  * @buffer:        Pointer to device image (xclbin) in memory
  * Return:         0 on success or appropriate error number
  *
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/include/xrt.h
```diff
@@ -335,9 +335,8 @@ xclGetErrorStatus(xclDeviceHandle handle, struct xclErrorStatus *info);
  * @buffer:        Pointer to device image (xclbin) in memory
  * Return:         0 on success or appropriate error number
  *
- * Download FPGA image (AXLF) to the device. The PR bitstream is
- * encapsulated inside xclbin as a section. xclbin may also contains
- * other sections which are suitably handled by the driver.
+ * | Download FPGA image (AXLF) to the device. The PR bitstream is encapsulated inside xclbin as a section. xclbin may also contains other sections which are suitably handled by the driver.
+ * | **NOTE**: This api also downloads OVERLAY(dtbo) section (Applicable only for Edge Platforms)
```
stsoe, line 339: Please reformat.  Do proper capitalization and spaces, don't exceed 80 chars.

stsoe (changes requested):

src/runtime_src/core/include/xrt.h
```diff
@@ -335,9 +335,9 @@ xclGetErrorStatus(xclDeviceHandle handle, struct xclErrorStatus *info);
  * @buffer:        Pointer to device image (xclbin) in memory
  * Return:         0 on success or appropriate error number
  *
- * Download FPGA image (AXLF) to the device. The PR bitstream is
- * encapsulated inside xclbin as a section. xclbin may also contains
- * other sections which are suitably handled by the driver.
+ * | Download FPGA image (AXLF) to the device. The PR bitstream is encapsulated inside xclbin as a section. 
+ * | xclbin may also contains other sections which are suitably handled by the driver.
+ * | **NOTE**: This API also downloads OVERLAY (dtbo) section. (Applicable only for Edge Platforms)
```
stsoe, line 340: Sorry, the lines are too long.  This is a public exported header files, it must be formatted properly.  What's the deal with the '|' character?   I know the original comment didn't read well either but at least we should fix that.

```suggestion
 * Download FPGA image (AXLF) to the device. The PR bitstream is
 * encapsulated inside xclbin as a section.  The xclbin may also
 * contain other sections, which are suitably handled by the driver.
 *
 * This API also downloads OVERLAY (dtbo) section (Edge only).
```

rbramand-xilinx: Hi @stsoe , I am trying to get NOTE in the new line. '|' is used for new line. With the suggested change its not possible, so I sent the commit this way. I will do some more trials and update

stsoe: Please just commit my change.  I don't see the need for a NOTE.

rbramand-xilinx: Thanks soren, I will commit the changes as suggested

PR #6544, opened 2022-04-01
Cherry-pick: Resolved the result mismatch issue for CR-1127553 (#6538)

subeeshm-xilinx:
Resolved the issue by pointing the p2p file descriptor to appropriate offsets
This is an issue which was caught as part of the CR-1127553
Low Risk
We have verified various combination testcases of P2P manually
No impact on Documentation


stsoe (commented):
@subeeshm-xilinx Please use git cherry-pick -x <commit> when cherry picking and don't bother with any additional verbiage in the PR description.

subeeshm-xilinx:
Thanks @stsoe Was not aware of this option. Will try doing it from next time.

stsoe:
> Thanks @stsoe Was not aware of this option. Will try doing it from next time.

Please close and redo.

PR #6557, opened 2022-04-04
Autorestart changes for edge platforms

ramyam-xilinx:

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/xbmgmt
+++ b/src/runtime_src/core/tools/xbmgmt2/xbmgmt
@@ -3,10 +3,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 #
-# Copyright (C) 2020-2021 Xilinx, Inc.
-#
-# xbmgmt loader which automatically switches between legacy and next-gen
-# xbmgmt utility. Starting with 2021.1 the default is next-gen xbmgmt utility.
+# Copyright (C) 2020-2022 Xilinx, Inc.
 #
 
 # -- Detect a Windows environment and automatically switch to the .bat file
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/edge/sw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -840,6 +858,89 @@ namespace xclcpuemhal2 {
...
+      xclRegRead_RPC_CALL(xclRegRead,baseAddr,offset,buff.data(),size,0,0);
+      auto tmp_buff = reinterpret_cast<uint32_t*>(buff.data());
+      *datap = tmp_buff[0];
+    }
+    else {
+      uint32_t * tmp_buff = reinterpret_cast<uint32_t*>(buff.data());
+      tmp_buff[0] = *datap;
+      xclRegWrite_RPC_CALL(xclRegWrite,baseAddr,offset,tmp_buff,0,0);
+    }	
+    return 0;
```
stsoe, line 926: Hmm, why not just use a uint32_t for buff and pass the address of buff to the RPC calls? E.g. something like this:
```
uint32_t buff;   
...
if (rd) {
  ...
  xclRegRead_RPC_CALL(,,,&buff, sizeof(buff), 0, 0);
}
...
```
You could even just use the `datap` directory assuming it is not null, which isn't checked anyway (in the read case).

stsoe (changes requested):

src/runtime_src/core/edge/sw_em/generic_pcie_hal2/shim.cxx
ramyam-xilinx: Hi,
Since the xclRegRead_RPC_CALL accepts  bytes, I have used std::array<char, REG_BUFF_SIZE> buff;
I am seeing compilation issues, when we are using as uint32_t.
I have accidentally closed this PR. I have recreated a new one with similar changes.

PR #6559, opened 2022-04-04
CR-1127863 VMR easy of use - debug mode - task management

xdavidz:

Update 04/15:
- simplify the rmgmt tasks with no timeout value, just a flag to indicate this task can be started or not.
- add ospi_flash_safe_write, which can safely write data size smaller than sector size by caching existing data and write it back.
- xgq will only init basic service (vmr_control) and then allow either rmgmt or vmc not to provide service but host driver can still talk to xgq for minimum service. Thus xgq service will keep alive.

A new shell flash will automatically set all values to 0 so not affect on regular production shell. 

CR-1127863

image

N/A

Tested on 0328 shell. 
See debug value, after hot reset check log the value is taking effect.
Reset back to 0, everything becomes back to normal mode.

Rerun xbutil validate passed without any issue.

04/15 Update:
simplify it again with no delay value needed. only a flag which can disable the task.
more code changed in VMR side, but seems cleaner than having 2 delay value hard-coded for rmgmt and vmc.

https://confluence.xilinx.com/pages/viewpage.action?pageId=302345261#XGQ1.0CommandandQueueSpecification(WIP)-GetLogPage

### DIFF
--- a/src/runtime_src/core/include/xgq_cmd_vmr.h
+++ b/src/runtime_src/core/include/xgq_cmd_vmr.h
@@ -87,6 +87,7 @@ enum xgq_cmd_vmr_control_type {
 	XGQ_CMD_BOOT_DEFAULT	= 0x1,
 	XGQ_CMD_BOOT_BACKUP	= 0x2,
 	XGQ_CMD_PROGRAM_SC	= 0x3,
+	XGQ_CMD_VMR_DEBUG	= 0x4,
 };
 
 /**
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/include/xgq_cmd_vmr.h
```diff
@@ -193,6 +194,8 @@ struct xgq_cmd_vmr_control_payload {
 	uint32_t req_type:8;
 	uint32_t debug_level:3;
 	uint32_t rsvd:21;
+	uint32_t debug_rmgmt_delay;
+	uint32_t debug_vmc_delay;
```
houlz0507, line 198: There could be more task in future? How about use one field and encode the different VMR apps

xdavidz: agree, I can make the change like:
u32 debug_rmgmt:1
u32 debug_vmc:1
u32 debug_rsve:30
u32 debug_value;

xdavidz:
@stsoe since it is XGQ interface change, simply drop your attention per your previous request. Please feel free to add more comments if you have.

maxzhen (approved):

houlz0507 (approved):

xdavidz:
Need sometime to clean this up with even simpler design.

houlz0507 (approved):

PR #6564, opened 2022-04-05
CR-1127564 Azure: bypass unattested xclbin check for internal testing

rajkumar-xilinx:
Since the standalone ZT servers does not have the same infrastructure of Azure,
this change facilitates verifying xclbin download without recompiling XRT.

User has to add the environment settings in following way in Guess machine

1. Run command "systemctl edit mpd"
2. add below lines in file opened
  [Service]
  Environment="ALLOW_UNATTESTED_XCLBIN=true"
3. save the file
4. restart mpd service using command "systemctl restart mpd"

Allow unattested xclbin in internal testing since the internal servers doesn't have same infra like Azure

Added change in Azure driver

NA
xclbin download in Azure setup

NA

### DIFF
--- a/src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.cpp
+++ b/src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.cpp
@@ -1,5 +1,5 @@
 /*
- * Partial Copyright (C) 2019-2021 Xilinx, Inc
+ * Partial Copyright (C) 2019-2022 Xilinx, Inc
  *
  * Microsoft provides sample code how RESTful APIs are being called
  *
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.cpp
```diff
@@ -225,6 +225,8 @@ static size_t WriteCallback(void *contents, size_t size, size_t nmemb, void *use
 int AzureDev::azureLoadXclBin(const xclBin *buffer)
 {
     char *xclbininmemory = reinterpret_cast<char*> (const_cast<xclBin*> (buffer));
+    bool allow_unattested_xclbin = false;
```
chvamshi-xilinx: declare it wherever it is being used

src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.cpp
```diff
@@ -234,8 +236,13 @@ int AzureDev::azureLoadXclBin(const xclBin *buffer)
         return -E_EMPTY_SN;
     std::cout << "LoadXclBin FPGA serial No: " << fpgaSerialNumber << std::endl;
 
+    char* env = getenv("AZURE_UNATTESTED_XCLBIN");
```
chvamshi-xilinx: use ALLOW_UNATTESTED_XCLBIN

chvamshi-xilinx (changes requested):

chvamshi-xilinx (approved):

PR #6567, opened 2022-04-05
CR-1127707 xbmgmt dump -d 0000:41:00.0 -f does not work for u25

xuhz:
xbmgmt dump -f gets stuck for u25
CR-1127707 filed by DSV
If read flash is the very first cmd to access flash, some required initialization is not done
risk is low. qspips flash/erase work as expected, the change just fix the issue that read is the first cmd to access flash
both xbmgmt and xbflash have the same issue.
N/A

### DIFF
--- a/src/runtime_src/core/pcie/tools/xbflash.qspi/xqspips.cpp
+++ b/src/runtime_src/core/pcie/tools/xbflash.qspi/xqspips.cpp
@@ -1,5 +1,5 @@
 /**
- * Copyright (C) 2016-2018 Xilinx, Inc
+ * Copyright (C) 2016-2022 Xilinx, Inc
  * Author(s) : Min Ma
  *
  * Licensed under the Apache License, Version 2.0 (the "License"). You may
--- a/src/runtime_src/core/pcie/tools/xbflash.qspi/xqspips.cpp
+++ b/src/runtime_src/core/pcie/tools/xbflash.qspi/xqspips.cpp
@@ -499,6 +499,24 @@ int XQSPIPS_Flasher::xclReadBack(std::string output, unsigned base, unsigned tot
     unsigned size = 0;
     int beatCount = 0;
 
+    initQSpiPS();
+
+    uint32_t StatusReg = XQSpiPS_GetStatusReg();
+
+    if (StatusReg == 0xFFFFFFFF) {
+        std::cout << "[ERROR]: Read PCIe device return -1. Cannot get QSPI status." << std::endl;
+        exit(-EOPNOTSUPP);
+    }
+
+    /* Make sure it is ready to receive commands. */
+    resetQSpiPS();
+    XQSpiPS_Enable_GQSPI();
+
+    if (!getFlashID()) {
+        std::cout << "[ERROR]: Could not get Flash ID" << std::endl;
+        exit(-EOPNOTSUPP);
+    }
+
     std::ofstream of_flash;
     of_flash.open(output, std::ofstream::out);
     if (!of_flash.is_open()) {
--- a/src/runtime_src/core/tools/xbmgmt2/flash/xqspips.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/flash/xqspips.cpp
@@ -482,6 +482,24 @@ unsigned int XQSPIPS_Flasher::getFlashSize()
 
 void XQSPIPS_Flasher::readBack(const std::string& output, unsigned int base)
 {
+    initQSpiPS();
+
+    uint32_t StatusReg = XQSpiPS_GetStatusReg();
+
+    if (StatusReg == 0xFFFFFFFF) {
+        std::cout << "[ERROR]: Read PCIe device return -1. Cannot get QSPI status." << std::endl;
+        return;
+    }
+
+    /* Make sure it is ready to receive commands. */
+    resetQSpiPS();
+    XQSpiPS_Enable_GQSPI();
+
+    if (!getFlashID()) {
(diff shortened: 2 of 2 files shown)


src/runtime_src/core/pcie/tools/xbflash.qspi/xqspips.cpp
```diff
@@ -499,6 +499,24 @@ int XQSPIPS_Flasher::xclReadBack(std::string output, unsigned base, unsigned tot
     unsigned size = 0;
```
rozumx, line 499: Copyright date (line 2) needs to be updated.

src/runtime_src/core/pcie/tools/xbflash.qspi/xqspips.cpp
```diff
@@ -499,6 +499,24 @@ int XQSPIPS_Flasher::xclReadBack(std::string output, unsigned base, unsigned tot
...
     int beatCount = 0;
 
+    initQSpiPS();
+
+    uint32_t StatusReg = XQSpiPS_GetStatusReg();
+
+    if (StatusReg == 0xFFFFFFFF) {
```
rozumx, line 506: This code is dangerous for it makes a lot of assumptions.  For example, what happens of StatusReg becomes a uint64_t.  It would be better to replace:
`0xFFFFFFFF`
with:
`std::numeric_limits<decltype(StatusReg)>::max()`

src/runtime_src/core/tools/xbmgmt2/flash/xqspips.cpp
```diff
@@ -482,6 +482,24 @@ unsigned int XQSPIPS_Flasher::getFlashSize()
```
rozumx, line 482: Update copyright date line 2.

src/runtime_src/core/tools/xbmgmt2/flash/xqspips.cpp
```diff
@@ -482,6 +482,24 @@ unsigned int XQSPIPS_Flasher::getFlashSize()
 
 void XQSPIPS_Flasher::readBack(const std::string& output, unsigned int base)
 {
+    initQSpiPS();
+
+    uint32_t StatusReg = XQSpiPS_GetStatusReg();
+
+    if (StatusReg == 0xFFFFFFFF) {
```
rozumx, line 489: This code is dangerous for it makes a lot of assumptions.  For example, what happens of StatusReg becomes a uint64_t.  It would be better to replace:
`0xFFFFFFFF`
with:
`std::numeric_limits<decltype(StatusReg)>::max()`

src/runtime_src/core/tools/xbmgmt2/flash/xqspips.cpp
```diff
@@ -482,6 +482,24 @@ unsigned int XQSPIPS_Flasher::getFlashSize()
...
+        std::cout << "[ERROR]: Read PCIe device return -1. Cannot get QSPI status." << std::endl;
```
rozumx, line 490: Anytime xbmgmt produces an error, it should return a non-zero error value.

This function should be updated to reflect that.   If there isn't time in the release to do this, then please create a CR so that this can be done in the next release.

rozumx (changes requested):

src/runtime_src/core/pcie/tools/xbflash.qspi/xqspips.cpp
xuhz: flash spec already determines all registers are 32bit, and the register read/write functions defined are also based on that. and the statusReg variable is also declared as u32

src/runtime_src/core/pcie/tools/xbflash.qspi/xqspips.cpp
xuhz: will change

src/runtime_src/core/tools/xbmgmt2/flash/xqspips.cpp
xuhz: will change

src/runtime_src/core/tools/xbmgmt2/flash/xqspips.cpp
xuhz: ditto

src/runtime_src/core/tools/xbmgmt2/flash/xqspips.cpp
xuhz: agree.
in today's code, the caller from xbmgmt is also void, and hence it makes no sense to return errno.
I can file CR tracking this. whoever changes the xbmgmt can change this API also

src/runtime_src/core/pcie/tools/xbflash.qspi/xqspips.cpp
xuhz: and this happens to be 0xffffffff, which means all bits in the register is 1
the value may be a number other than all 1s -- it all depends on how spec defines.
The original code was not written by me -- i just copied it to here. but I don't think using a ::max makes sense here

AShivangi (approved):

rozumx (approved):
Approved on the condition that a CR be created to have xbmgmt return back an non-zero error value when this error condition is encountered.

src/runtime_src/core/tools/xbmgmt2/flash/xqspips.cpp
xuhz: https://jira.xilinx.com/browse/CR-1129024
is logged

PR #6570, opened 2022-04-05
CR-1125301 remove legacy xbutil/xbmgmt tools

AShivangi:
2nd iteration of removing legacy code

None 

Locally:
```
$ xbutil --legacy
--------------------------------------------------
Error: Obsoleted option
       The --legacy option is no longer supported.
--------------------------------------------------
$ echo $?
1

```


stsoe (commented):
Nice!  Can you also check core/pcie/common/dd.* which I believe is used only by xbutil; in fact I have a local commit in my clone that moves dd.* to pcie/tools/xbutil, now it can be removed.

chvamshi-xilinx:
Hi @AShivangi ,
Can you please remove the edge version of xbutil aswell. src/runtime_src/core/edge/tools/

stsoe (approved):

rozumx (approved):

PR #6572, opened 2022-04-06
Stop iteration slots on open_context returning EAGAIN

stsoe:
Per discussion with team, open_context on a CU should stop slot
iteration when first EAGAIN is encountered from driver.

On EAGAIN, stop current iteration, re-fresh slots, and try the new
slot. EAGAIN implies that driver has loaded existing xclbin into a
new slot and that current process should stop requesting open context
on any existing slots.

The algorithm reloads the new slot info from driver on EAGAIN, and
then runs again on just the new slots if any.

None, affects EAGAIN return code only, which is currently unused in Alveo and Edge.

Spec slides have been updated.


xilinxgitops:
retest this please

PR #6583, opened 2022-04-07
Improve modularity of some core CMake files

stsoe:
Small step towards better structured makefiles using more modern CMake
design.  Use library targets and properties with targets; avoid
variables.


stsoe:
retest this please - u50 error on single test

rozumx (approved):

PR #6592, opened 2022-04-12
Updating petalinux to 2022.1_stable_latest

jeffli-xilinx:
Updating petalinux version to 2022.1_stable_latest as 0318 build has been removed.

PR#6535 had updated to 0318 petalinux, but that build was not kept


rozumx (approved):
Ideally this value should be settable by the user via a command line option or something similiar. 
Of course, it the user doesn't do so, then this value would be the default value.

PR #6593, opened 2022-04-12
CR-1128307 : Fix dangling pointer dereference issue while deleting kds client context in embedded platforms

rbramand-xilinx:
Fix null pointer dereference which causes kernel crash in embedded platforms when kds client contexts are removed

PR- https://github.com/Xilinx/XRT/pull/6160 introduced this bug, it was discovered using sprite test case - uram_ecc_global_int

we are freeing client but not deleting linked list node, added changes to delete node

no

Ran application 1000 times, no issue is found

NA


chvamshi-xilinx:
Merging as this is needed for 2022.1

PR #6598, opened 2022-04-12
CR-1128706 XRT failure on kvm systems

AShivangi:
The postinst script calls xbmgmt configure to start daemons for kvm. However, there is a small bug in the implementation. The host name is not updated in the configuration

I removed the check which prevents this from happening.

low to none

Manual testing on KVM and bare metal and pipeline testing

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
@@ -298,8 +298,7 @@ update_daemon_config(const std::string& host = "")
   if (!cfile)
     throw xrt_core::system_error(std::errc::invalid_argument, "Missing '" + std::string(config_file) + "'.  Cannot update");
 
-  if(host.empty())
-    cfg.host = host;
+  cfg.host = host;
   // update the configuration file
   cfile << boost::str(boost::format("%s\n") % cfg);
   std::cout << boost::format("Successfully updated the Daemon configuration.\n");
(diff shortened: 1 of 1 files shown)


AShivangi:
Approved by Sumit and Anup. Please merge it when the build passes.

keryell (commented):
Interesting bug. :-)
Could you fix the PR title and the commit to have a correct description of what is done?
I guess the goal is not to have a failure on (uppercase) KVM systems. :-)

rozumx (approved):

src/runtime_src/core/tools/xbmgmt2/SubCmdConfigure.cpp
```diff
@@ -298,8 +298,7 @@ update_daemon_config(const std::string& host = "")
   if (!cfile)
     throw xrt_core::system_error(std::errc::invalid_argument, "Missing '" + std::string(config_file) + "'.  Cannot update");
 
-  if(host.empty())
-    cfg.host = host;
+  cfg.host = host;
```
chvamshi-xilinx, line 301: We should still add a check. Please update cfg.host only if host is not empty.

chvamshi-xilinx (approved):

chvamshi-xilinx:
Please cherry pick to master

AShivangi:
> Please cherry pick to master

@chvamshi-xilinx, this is correct in master. While creating a PR for 2022.1, I messed up.

chvamshi-xilinx:
Thanks for the confirmation..

PR #6605, opened 2022-04-13
CR-1128704: Removed mention of --legacy from xbutil/xbmgmt rst files

uday610:

Removing mention of --legacy as this is completed removed
[CR-1128704]


maxzhen:
This is a documentation only change. How come the pipeline build failed??

maxzhen:
Pipeline team confirmed that this is a infra issue. Merging it now without waiting for next build since this is a doc only change.

Issue #6574, opened 2022-04-06
GCC built int define __GNUC__ used to control Linux specific code


kiwichris:
GCC does not means Linux. For example in `src/runtime_src/core/common/config_reader.cpp`:
```
#ifdef __GNUC__
# include <linux/limits.h>
# include <sys/stat.h>
#endif
```
This break the portability for the code. Why not key the defined on something specific to Linux?

kiwichris:
Many thanks for the change and sorting this out.

PR #6621, opened 2022-04-20
Don't use __GNUC__ to imply Linux

stsoe:
Use __linux__ in place of __GNUC__ where Linux is the differentiator.

Addresses issue #6574.  There are still instance of __GNUC__ left, but primarily for GCC specific attributes.

### DIFF
--- a/src/runtime_src/core/common/config_reader.cpp
+++ b/src/runtime_src/core/common/config_reader.cpp
@@ -30,7 +30,7 @@
 #include <boost/filesystem/operations.hpp>
 #include <boost/format.hpp>
 
-#ifdef __GNUC__
+#ifdef __linux__
 # include <linux/limits.h>
 # include <sys/stat.h>
 #endif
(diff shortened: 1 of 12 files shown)


AShivangi (approved):

maxzhen (approved):
Change looks OK. @stsoe , just want to confirm the thoughts behind this change... so we're going to use OS names to differentiate both compiler difference and OS difference (GUN tool chain can also run on Windows)? I also see "_WINDOWS" and "_WIN32" in the code, are we merging them?

stsoe:
> Change looks OK. @stsoe , just want to confirm the thoughts behind this change... so we're going to use OS names to differentiate both compiler difference and OS difference (GUN tool chain can also run on Windows)? I also see "_WINDOWS" and "_WIN32" in the code, are we merging them?

Compiler differences should be specific to compiler and continue to use __GNUC__, if supported only with gcc.  OS specifics should be specific to OS.  Yes, you are right windows in not correctly used either.

PR #6635, opened 2022-04-26
Support for new memory size notation of the platform

venkatp-xilinx:
Support of new memory size notation in sw_emu edge and pcie drivers

Recent updates to Vivado in extracting the memory size has resulted this size

Enabled the support for new memory notation as per the platform update

No

Ran the Canary and failed pipeline test case has been verified

No

### DIFF
--- a/src/runtime_src/core/edge/common_em/config.cxx
+++ b/src/runtime_src/core/edge/common_em/config.cxx
@@ -464,99 +464,28 @@ namespace xclemulation{
     return environmentNameValueMap;
   }
 
-  //initialize memMap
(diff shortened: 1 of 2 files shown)


venkatp-xilinx:
> Hi @venkatp-xilinx , This is not scalable, Please use user_literal method to convert https://en.cppreference.com/w/cpp/language/user_literal

@chvamshi-xilinx, I am unable to apply the user_literal concept here, however updated the logic to be more scalable and allows various other sort of notations. It applies for both the hw_emu and sw_emu.

PR #6639, opened 2022-04-27
CR-1128896 xbmgmt examine -r vmr should return non-zero when fail

rchane:
https://jira.xilinx.com/browse/CR-1128896
A vck5000 with problems was returning 0, which was undesired behavior. While we want to maintain
a 0 return code for failures for non-vmr-supporting cards, we have to account for this scenario where
the vck5000 should have had information.
Check if any of the non-verbose labels were not found when they should have been. If so, throw an error,
as there was an issue with retrieving vmr info. An alternate solution involved editing the ptree to fit verbose 
vmr info in a subtree (instead of being top-level with the rest) and detecting its contents, but it had a ripple 
effect that was too risky.
This is a fix for a card that was in an unexpected state. It isn't as robust of a solution as a dedicated hardware flag
indicating if all vmr info is good to go/supported, which could be implemented down the line.
Testing Sample Outputs
Card that doesn't support vmr:
```
rchane@xsjsagarw50:/proj/rdi/staff/rchane/XRT$ xbmgmt examine -r vmr -d 17:00

1/1 [0000:17:00.0] : xilinx_u30_gen3x4_base_2
  Information Unavailable
rchane@xsjsagarw50:/proj/rdi/staff/rchane/XRT$ echo $?
0
```
Card that supports vmr and info is successfully retrieved:
```
rchane@xsjsagarw50:/proj/xsjhdstaff4/rchane/XRT$ xbmgmt examine -r vmr -d 65:00

-------------------------------------------------------
1/1 [0000:65:00.0] : xilinx_vck5000_gen4x8_xdma_base_2
-------------------------------------------------------
Vmr Status
  Build flags          :  default full build
  Vitis version        :  2022.1
  Git hash             :  8d589a2c17d0cf27ef2268dc14fc338707deeb18
  Git branch           :  2022.1
  Git hash date        :  Thu, 24 Mar 2022 21

rchane@xsjsagarw50:/proj/xsjhdstaff4/rchane/XRT$ echo $?
0
```
Card that supports vmr and info is NOT successfully retrieved (In this sample, only "build flags" was found. It's also possible that no labels at all are found):
```
rchane@xsjsagarw50:/proj/rdi/staff/rchane/XRT$ xbmgmt examine -r vmr -d 65:00

-------------------------------------------------------
1/1 [0000:65:00.0] : xilinx_vck5000_gen4x8_xdma_base_2
-------------------------------------------------------
Vmr Status
  Build flags          :  default full build
Vmr
  ERROR: Incomplete Information
rchane@xsjsagarw50:/proj/rdi/staff/rchane/XRT$ echo $?
1
```
None

### DIFF
--- a/src/runtime_src/core/tools/common/ReportVmrStatus.cpp
+++ b/src/runtime_src/core/tools/common/ReportVmrStatus.cpp
@@ -50,11 +50,17 @@ ReportVmrStatus::writeReport( const xrt_core::device* /*_pDevice*/,
   boost::property_tree::ptree pt_empty;
   const boost::property_tree::ptree& ptree = pt.get_child("vmr", pt_empty);
 
+  //Only vck5000 cards support vmr
   output << "Vmr Status" << std::endl;
   if (ptree.empty()) {
     output << "  Information Unavailable" << std::endl;
     return;
   }
+
+  //vmr should be supported, but info was unavailable
+  const boost::property_tree::ptree& version_ptree = pt.get_child("vmr.vmr_version", pt_empty);
+  if (version_ptree.empty())
+    throw std::runtime_error("Information Unavailable");
 
   //list of non verbose labels
   const std::vector<std::string> non_verbose_labels = { 
@@ -65,7 +71,7 @@ ReportVmrStatus::writeReport( const xrt_core::device* /*_pDevice*/,
       "git hash date" 
     };
 
-  for(auto& ks : ptree) {
+  for(auto& ks : version_ptree) {
     const boost::property_tree::ptree& vmr_stat = ks.second;
     const auto it = std::find_if(non_verbose_labels.begin(), non_verbose_labels.end(),
                       [&vmr_stat](const auto& str) { return boost::iequals(vmr_stat.get<std::string>("label"), str); });
--- a/src/runtime_src/core/tools/common/ReportVmrStatus.cpp
+++ b/src/runtime_src/core/tools/common/ReportVmrStatus.cpp
@@ -50,18 +50,12 @@ ReportVmrStatus::writeReport( const xrt_core::device* /*_pDevice*/,
   boost::property_tree::ptree pt_empty;
   const boost::property_tree::ptree& ptree = pt.get_child("vmr", pt_empty);
 
-  //Only vck5000 cards support vmr
+
+  output << "Vmr Status" << std::endl;
   if (ptree.empty()) {
     output << "  Information Unavailable" << std::endl;
     return;
   }
-
-  //vmr should be supported, but info was unavailable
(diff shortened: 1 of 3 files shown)


stsoe:
whitelist this please

stsoe:
ok to test

src/runtime_src/core/tools/common/ReportVmrStatus.cpp
```diff
@@ -50,30 +50,36 @@ ReportVmrStatus::writeReport( const xrt_core::device* /*_pDevice*/,
...
+    if (XBUtilities::getVerbose() || it != non_verbose_labels.end())
+      aggregated_report << fmt_basic % vmr_stat.get<std::string>("label") % vmr_stat.get<std::string>("value");
+    if (it != non_verbose_labels.end())
+      non_verbose_labels.erase(it);
   }
+  //After removing found non-verbose labels, if there are any left, throw an error.
+  if (!non_verbose_labels.empty())
+    throw std::runtime_error("Information Unavailable");
+  //If no errors in retrieving all non-verbose labels, output as normal.
+  output << aggregated_report.str();
```
AShivangi, line 83: move this line above the if statement (L80)

src/runtime_src/core/tools/common/ReportVmrStatus.cpp
```diff
@@ -50,30 +50,36 @@ ReportVmrStatus::writeReport( const xrt_core::device* /*_pDevice*/,
...
-    else if(it != std::end(non_verbose_labels))
-      output << fmt_basic % vmr_stat.get<std::string>("label") % vmr_stat.get<std::string>("value");
+    if (XBUtilities::getVerbose() || it != non_verbose_labels.end())
+      aggregated_report << fmt_basic % vmr_stat.get<std::string>("label") % vmr_stat.get<std::string>("value");
+    if (it != non_verbose_labels.end())
+      non_verbose_labels.erase(it);
   }
+  //After removing found non-verbose labels, if there are any left, throw an error.
+  if (!non_verbose_labels.empty())
+    throw std::runtime_error("Information Unavailable");
```
AShivangi, line 81: instead of throwing "Information Unavailable", we can simply throw the error code, We could also add an error message which is more descriptive like "Incomplete info" or "corrupted data"

AShivangi (approved):
Thanks for creating this PR and adding a description! It looks good to me

@rozumx, I talked to Ryan offline to understand the issue here. In short, there is a scenario where the "vmr_version" sysfs node is not fully populated. According to David, this means that something is wrong with the vmr. He said that we can add a hw flag later on, but this workaround works for now. I have asked Ryan to work with David to create a CR for the hw flag.

Please let us know if you have questions/concerns about this fix and we can schedule a meeting. 
Thanks!

dayeh-xilinx:
retest this please.

rozumx (approved):

PR #6644, opened 2022-04-28
Coverity fix to intialize the hal major and minor version to default …

venkatp-xilinx:
…value.

Coverity reported an issue regarding the default initialization of the hal major and minor version

Not a bug. Issue reported by Coverity

Initialized with proper value

No

Ran the canaries

No


chvamshi-xilinx (approved):
Are we using this anywhere?

PR #6647, opened 2022-04-28
Add useful getters to xrt::kernel

Ralender:
I needed this while building interop support for our SYCL plugin.

@stsoe : do you want those to be documented or not ?

### DIFF
--- a/src/runtime_src/core/include/xrt/xrt_kernel.h
+++ b/src/runtime_src/core/include/xrt/xrt_kernel.h
@@ -700,6 +700,20 @@ public:
   uint32_t
   read_register(uint32_t offset) const;
 
+  /**
+   * get_name() - Return the name of the kernel
+   */
+  XCL_DRIVER_DLLESPEC
+  std::string 
+  get_name() const;
+  
+  /**
+   * get_xclbin() - Return the xclbin containing the kernel
+   */
+  XCL_DRIVER_DLLESPEC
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/include/xrt/xrt_kernel.h
```diff
@@ -701,6 +701,9 @@ class kernel
   read_register(uint32_t offset) const;
 
 public:
+  std::string get_name() const;
+  xrt::xclbin get_xclbin() const;
```
stsoe, line 705: Yes, please document doxygen style since these are public APIs.  Also please follow the suggested convention:
```suggestion
  XCL_DRIVER_DLLESPEC
  std::string 
  get_name() const;
  
  XCL_DRIVER_DLLESPEC
  xrt::xclbin 
  get_xclbin() const;
```
I am adamant about coding to the left; I simply hate scanning code for the function name somewhere after the return type which varies in length always.

stsoe (changes requested):
Thank you.  Makes sense.

src/runtime_src/core/include/xrt/xrt_kernel.h
stsoe: ok to test

stsoe (approved):

stsoe:
ok to test

PR #6648, opened 2022-04-29
Add missing #include <stdexcept>

keryell:
There was a throw std::runtime_error which was undefined when
compiling with g++ 11.2 on Ubuntu 21.10.


stsoe (approved):
Thank you!

stsoe:
ok to test

PR #6652, opened 2022-04-29
Issue #6619 Implement xrt::device equality

### LINKED ISSUE #6619: why aren't xrt::device uniqued ?
for example:
```
#include <xrt/xrt_kernel.h>
#include <xrt.h>
#include <cassert>

int main() {
auto a = xrt::device(0);
auto b = xrt::device(0);
assert(static_cast<xclDeviceHandle>(a) == static_cast<xclDeviceHandle>(b));
assert(a.get_handle() == b.get_handle());
}
```
both `a` and `b` refer to the same device but the asserts fail.

stsoe:
Two xrt::device objects referring to same physical device
now compare equal.

Equality fails on two xrt::device objects even if both refer to same underlying physical device

Fixes #6619

Implement operator== to compare the device id with which the device object was constructed.

Behavior changes because the previous default operator== compared the xrt::device opaque implementation handle and would return false when after this change the comparison may return true if the handle implementation refers to same physical device.

Updated unit test accordingly

Per doxygen

### DIFF
--- a/src/runtime_src/core/common/api/xrt_device.cpp
+++ b/src/runtime_src/core/common/api/xrt_device.cpp
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright (C) 2020-2022 Xilinx, Inc. All rights reserved.
+// Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
 
 // This file implements XRT xclbin APIs as declared in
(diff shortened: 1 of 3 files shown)


stsoe:
retest this please

dayeh-xilinx:
retest this please.

keryell (approved):
Thanks!
@Ralender this is for you.

uday610 (approved):
Kool, Thanks

PR #6653, opened 2022-04-30
Multislot fixes for multiple xclbin loaded parallelly

saifuddin-xilinx:
**Issue :** 
Multi slot code is working for single slot only. But while testing with multiple xclbin, it's failing.

**Root Cause:**
The root cause of this is locking. We haven't use the locks efficiently to support multiple xclbin loaded parallelly. 
Also many places we have taken spinlocks in non-atomic context. 

**Test:** 
I have a customized test case for multiple xclbin (multi slot verification) with multiple PL kernels (i.e. vadd, vsub, vmul, vdiv).
I have different xclbin for each kernels. The test application is forking multiple child processes. 
Each child process is loading/reloading a different xclbin to a different slot parallelly.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
@@ -371,8 +371,10 @@ get_next_free_apt_index(struct drm_zocl_dev *zdev)
 {
 	int apt_idx = 0;
 
+	BUG_ON(!mutex_is_locked(&zdev->cu_subdev.lock));
+
 	for (apt_idx = 0; apt_idx < MAX_APT_NUM; ++apt_idx) {
-		if (zdev->apertures[apt_idx].addr == EMPTY_APT_VALUE)
+		if (zdev->cu_subdev.apertures[apt_idx].addr == EMPTY_APT_VALUE)
 			return apt_idx;
 	}
 
@@ -391,10 +393,12 @@ update_max_apt_number(struct drm_zocl_dev *zdev)
 {
 	int apt_idx = 0;
 
-	zdev->num_apts = 0;
+	BUG_ON(!mutex_is_locked(&zdev->cu_subdev.lock));
+
+	zdev->cu_subdev.num_apts = 0;
 	for (apt_idx = 0; apt_idx < MAX_APT_NUM; ++apt_idx) {
-		if (zdev->apertures[apt_idx].addr != 0)
-			zdev->num_apts = apt_idx + 1;
+		if (zdev->cu_subdev.apertures[apt_idx].addr != 0)
+			zdev->cu_subdev.num_apts = apt_idx + 1;
 	}
 }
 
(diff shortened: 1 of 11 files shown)


chvamshi-xilinx (approved):

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
```diff
@@ -992,29 +1013,27 @@ zocl_load_aie_only_pdi(struct drm_zocl_dev *zdev, struct axlf *axlf,
...
-		CLEAR(slot->connectivity);
-	}
-	if (slot->topology) {
+	if (slot->topology)
 		vfree(slot->topology);
-		CLEAR(slot->topology);
-	}
-	if (slot->axlf) {
+	if (slot->axlf)
 		vfree(slot->axlf);
```
mamin506, line 1027: vfree can ignore NULL pointer. We don't need to check if it is NULL.

See mm/vmalloc.c:
/**
 * vfree - Release memory allocated by vmalloc()
 * @addr:  Memory base address
 *
 * Free the virtually continuous memory area starting at @addr, as obtained
 * from one of the vmalloc() family of APIs.  This will usually also free the
 * physical memory underlying the virtual allocation, but that memory is
 * reference counted, so it will not be freed until the last user goes away.
 *
 * If @addr is NULL, no operation is performed.
 *
 * Context:
 * May sleep if called *not* from interrupt context.
 * Must not be called in NMI context (strictly speaking, it could be
 * if we have CONFIG_ARCH_HAVE_NMI_SAFE_CMPXCHG, but making the calling
 * conventions for vfree() arch-dependent would be a really bad idea).
 */
 void vfree(const void *addr)

mamin506, line 1027: vfree can ignore NULL pointer. Maybe we don't need to check if it is NULL?

See mm/vmalloc.c:
/**
 * vfree - Release memory allocated by vmalloc()
 * @addr:  Memory base address
 *
 * Free the virtually continuous memory area starting at @addr, as obtained
 * from one of the vmalloc() family of APIs.  This will usually also free the
 * physical memory underlying the virtual allocation, but that memory is
 * reference counted, so it will not be freed until the last user goes away.
 *
 * If @addr is NULL, no operation is performed.
 *
 * Context:
 * May sleep if called *not* from interrupt context.
 * Must not be called in NMI context (strictly speaking, it could be
 * if we have CONFIG_ARCH_HAVE_NMI_SAFE_CMPXCHG, but making the calling
 * conventions for vfree() arch-dependent would be a really bad idea).
 */
 void vfree(const void *addr)

mamin506 (approved):
Just minor suggestion change.

src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
saifuddin-xilinx: Fixed. Thanks

PR #6665, opened 2022-05-05
Make XRT compile on Ubuntu 22.04

Ralender:
- Fix cmake for finding python libs
  findPythonLibs is deprecated since cmake 3.12
  replace it with findPython3 for newer versions
- silence OpenSSL 3 deprecation warnings
  MD5 used in container.cpp
  and SHA256 in azure.cpp

I didn't analyze if MD5 was used for security purposes. If it is the case it should be replaced by SHA2 or better.


stsoe (approved):

stsoe:
ok to test

Ralender:
I noticed that dbg.sh doesn't work on Ubuntu 22.04.

src/runtime_src/tools/xclbinutil/CMakeLists.txt
```diff
@@ -3,6 +3,10 @@
 #
 # Signing xclbin images are currently only support on Linux
 
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.19)
+  cmake_policy(SET CMP0110 NEW)
```
keryell, line 7: Explain the magic with some comments.

src/python/pybind11/CMakeLists.txt
```diff
@@ -1,23 +1,36 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright (C) 2020-2021 Xilinx, Inc. All rights reserved.
 #
+if (${CMAKE_VERSION} VERSION_LESS "3.12")
```
keryell, line 4: Explain the magic with some comments.

keryell (changes requested):
That sounds good but you should take some rationale from your PR description into some comments in the code.

src/python/pybind11/CMakeLists.txt
```diff
@@ -2,22 +2,37 @@
...
 #
 
-find_package(PythonLibs 3.4.0 REQUIRED)
-if (PythonLibs_FOUND)
-  message("-- Python libs version: ${PYTHONLIBS_VERSION_STRING}")
-  message("-- PYTHON_INCLUDE_PATH ${PYTHON_INCLUDE_PATH}")
-  string(REPLACE "." ";" PYTHONLIBS_VERSION_LIST ${PYTHONLIBS_VERSION_STRING})
-  list(GET PYTHONLIBS_VERSION_LIST 0 PYTHONLIBS_VERSION_MAJOR)
-  list(GET PYTHONLIBS_VERSION_LIST 1 PYTHONLIBS_VERSION_MINOR)
+# find_package(PythonLibs is depracted as of cmake 3.12.
```
keryell, line 5: Using a spell-checker is even better :-)

keryell (commented):
Thanks

PR #6675, opened 2022-05-06
Add RTEMS support for xbutil2 main as a command

kiwichris:
With RTEMS there are some constraints:

1. Applications are all statically linked so there can only be a single `main()`. We need to change the name of xbutil2's `main` to something else.
2. Commands on RTEMS are C functions. The function need external C linkage.

This change is a bit messy but it is simple. I could not see an easy compiler trick to do this without touching the code.


kiwichris:
The patch does not seem to be here. The branch in my repo is empty. I am not sure what happened. 

Ah for some reason the change is on the next pull request. I will close this request.

Issue #6672, opened 2022-05-06
Repo has mixed line endings


kiwichris:
Is this known about and expected?

To see a list run (on FreeBSD): `find . -type f -exec file \{\} \; | grep CRLF`

I am seeing in my clone of the repo the following files with `CRLF`:
```
src/runtime_src/core/edge/sw_em/generic_pcie_hal2/swscheduler.cxx                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               
src/runtime_src/core/edge/sw_em/generic_pcie_hal2/swscheduler.h
src/runtime_src/core/edge/test/add1_zgem.cpp
src/runtime_src/core/edge/test/add1_zgem_64.cpp
src/runtime_src/core/edge/user/test/xclHALProxy2.cpp
src/runtime_src/core/edge/user/test/xclHALProxy2.h
src/runtime_src/core/edge/user/test/zynq_user_test_add1.cpp
src/runtime_src/core/include/xgq_resp_parser.h
src/runtime_src/core/pcie/driver/aws/kernel/mgmt/10-awsmgmt.rules
src/runtime_src/core/pcie/driver/aws/kernel/mgmt/Makefile
src/runtime_src/core/pcie/driver/aws/kernel/mgmt/mgmt-cw.h
src/runtime_src/core/pcie/driver/linux/xocl/subdev/nifd.c
src/runtime_src/core/pcie/driver/linux/xocl/subdev/ulite.c
src/runtime_src/core/pcie/driver/linux/xocl/userpf/99-xocl.rules
src/runtime_src/core/pcie/emulation/cpu_em/generic_pcie_hal2/swscheduler.cxx
src/runtime_src/core/pcie/emulation/cpu_em/generic_pcie_hal2/swscheduler.h
src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/mbscheduler.cxx
src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/mbscheduler.h
src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/mbscheduler_hwemu.cxx
src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/mbscheduler_hwemu.h
src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/xgq_hwemu.cxx
src/runtime_src/core/pcie/emulation/hw_em/generic_pcie_hal2/xgq_hwemu.h
[...]

PR #6681, opened 2022-05-07
Convert CRLF to LF (#6672)

kiwichris:
This pull request converts any CRLF files to LF

The issue is #6672. I found this when attempting to modify CRLF files on FreeBSD. Git diff showed the issue.

I converted the files with:
```
#! /bin/sh
set -e
files=$(find . -type f -exec file \{\} \; | grep CRLF | sed -e 's/^\.\///' -e 's/\:.*$//')
for f in $files
do
  tr -d '\r' < $f > $f.tmp
  rm $f
  mv $f.tmp $f
done
exit 0
```


stsoe (commented):
Jeez, how do we prevent this mess in general?  
Next question, does it really matter?   I always ignore white spaces when doing git diff and in fact you can be specific about what git diff should ignore.

stsoe:
ok to test

kiwichris:
> Jeez, how do we prevent this mess in general?

I am not sure because this is the first time I have encountered a line ending problem on a git repo this size. We do not see this on RTEMS nor other large projects. Maybe we have a Unix base of users? 

In the past found it happened when developers clone on one system then edit on other via a networked file system and their editor is not sensitive to the line endings and does not automatically adjust.
 
> Next question, does it really matter?
This depends on a range of factors such and the life span of the code base, where it is used and by whom. I believe it is important and more so in public facing community focused projects. It provides a sure footing for those who come to the project and want to contribute and it lowers the bar for those passing by and dropping in a simple but important fix.

> I always ignore white spaces when doing git diff and in fact you can be specific about what git diff should ignore.

This is a good thing to do. I think a tool like `clang-format` is great because it manages all this and more.

dayeh-xilinx:
retest this please.

kiwichris:
> retest this please.

I cannot find anything that specifies what this means? Is there documentation on what is required?

This change effects a number of files across a number of places and I suspect the areas touched will be beyond my ability to build or test.

If I could get some constructive guidance it would be appreciated.

Thanks

stsoe:
> I cannot find anything that specifies what this means? Is there documentation on what is required?
> 
> This change effects a number of files across a number of places and I suspect the areas touched will be beyond my ability to build or test.
> 
> If I could get some constructive guidance it would be appreciated.
> 
> Thanks

Internal CI related

keryell:
That is massive. :-)
Good IDE & editor configurations and `clang-format` sound like a good idea.

PR #6683, opened 2022-05-09
Better management of padding in mailbox messages

richardtobin-xlnx:
Interoperation between the XOCL driver and the X3 FW is simplified by eliminating this padding.  The mailbox protocol is also made more explicit, using fixed-width types and deliberate padding instead of architecture/compiler padding and type sizing.

CR-1130511.

We originally had a workaround in the FW to account for the fact that the last 4 bytes of data from the XOCL driver were padding.  This workaround was found to be inadequate when this padding straddles the last two packets of a message.  Fixing it in the FW was going to be convoluted, requiring more careful tracking of how many bytes the host had sent and potentially ignoring the entire last packet of a stream if it contains only padding.  Rather than add more kludges on top of the existing kludge, it seemed much easier to try to fix the problem at source, by eliminating the padding.  The changes on the XOCL and FW sides were both trivial in comparison to the alternative

The original incarnation of this PR moved the padding to the front of the data stream, but this would cause interoperability problems if different versions of XOCL and XCLMGMT were mixed and matched.  The revised PR resolves this problem by simply eliminating the padding - it's redundant and the drivers are not reliant on it being there.

Interop with X3, which involves sending several mailbox requests and transferring a full XCLBIN works fine.  Waiting to see if the Github CI flags any problems

None, because this protocol is not fully documented to begin with!

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/include/mailbox_proto.h
+++ b/src/runtime_src/core/pcie/driver/linux/include/mailbox_proto.h
@@ -346,13 +346,13 @@ struct xcl_mailbox_clock_freqscaling {
 #define XCL_MB_REQ_FLAG_REQUEST		(1 << 1)
 /**
  * struct mailbox_req - mailbox request message header
- * @req: opcode
+ * @req: opcode (see enum xcl_mailbox_request)
  * @flags: flags of this message
  * @data: payload of variable length
  */
 struct xcl_mailbox_req {
 	uint64_t flags;
-	enum xcl_mailbox_request req;
+	uint32_t req;
 	char data[1]; /* variable length of payload */
 };
 
--- a/src/runtime_src/core/pcie/driver/linux/include/mailbox_proto.h
+++ b/src/runtime_src/core/pcie/driver/linux/include/mailbox_proto.h
@@ -19,6 +19,7 @@
 
 #ifndef __KERNEL__
 #include <stdint.h>
+#include <stddef.h>
 #endif
 
 /*
@@ -355,6 +356,12 @@ struct xcl_mailbox_req {
 	uint32_t req;
 	char data[1]; /* variable length of payload */
 };
+
+/**
+ * Definition to use for sizing mailbox request buffers
+ * e.g. size = XCL_MAILBOX_REQ_SIZE + message_length
+ */
+#define XCL_MAILBOX_REQ_SIZE (offsetof(struct xcl_mailbox_req, data))
 
 /**
  * struct sw_chan - mailbox software channel message metadata. This defines the
--- a/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
@@ -1126,7 +1126,7 @@ void xclmgmt_connect_notify(struct xclmgmt_dev *lro, bool online)
 	size_t data_len = 0, reqlen = 0;
 
 	data_len = sizeof(struct xcl_mailbox_peer_state);
-	reqlen = sizeof(struct xcl_mailbox_req) + data_len;
+	reqlen = XCL_MAILBOX_REQ_SIZE + data_len;
 	mb_req = vzalloc(reqlen);
 	if (!mb_req)
 		return;
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/dna.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/dna.c
@@ -86,7 +86,7 @@ static void xlnx_dna_read_from_peer(struct platform_device *pdev)
 	size_t resp_len = sizeof(struct xcl_dna);
 	size_t data_len = sizeof(struct xcl_mailbox_subdev_peer);
 	struct xcl_mailbox_req *mb_req = NULL;
-	size_t reqlen = sizeof(struct xcl_mailbox_req) + data_len;
+	size_t reqlen = XCL_MAILBOX_REQ_SIZE + data_len;
 	xdev_handle_t xdev = xocl_get_xdev(pdev);
 
 	mb_req = vmalloc(reqlen);
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/firewall.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/firewall.c
@@ -190,7 +190,7 @@ static void fw_read_from_peer(struct platform_device *pdev)
 	size_t resp_len = sizeof(struct xcl_firewall);
 	size_t data_len = sizeof(struct xcl_mailbox_subdev_peer);
 	struct xcl_mailbox_req *mb_req = NULL;
-	size_t reqlen = sizeof(struct xcl_mailbox_req) + data_len;
+	size_t reqlen = XCL_MAILBOX_REQ_SIZE + data_len;
 	xdev_handle_t xdev = xocl_get_xdev(pdev);
 
 	xocl_info(&pdev->dev, "reading from peer");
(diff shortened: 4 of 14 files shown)


src/runtime_src/core/pcie/driver/linux/include/mailbox_proto.h
```diff
@@ -346,14 +346,15 @@ struct xcl_mailbox_clock_freqscaling {
...
  * @flags: flags of this message
  * @data: payload of variable length
  */
 struct xcl_mailbox_req {
 	uint64_t flags;
-	enum xcl_mailbox_request req;
-	char data[1]; /* variable length of payload */
+	uint32_t req;
+	uint32_t padding;
+	char data[0]; /* variable length of payload */
```
arpitxilinx, line 357: This change will also require change in xclmgmt2 driver on host in Azure. This will require MS to modify the driver and update on all the hosts in all the Data Centers across the world. Although this is not an immediate impact, as this change doesn't affect 2022.1, but going forward it would impact them without adding additional functionality for the current Azure deployment. I am not oppose to the change, if it makes thing simpler for X3 firmware, but someone have to get the buy-in from MS, as to the impact going forward. 
Also, any reason why char data[1] was changed to data[0]? With current change, it won't compile with Visual Studio compilers. Please move it back to data[1].

arpitxilinx (changes requested):

src/runtime_src/core/pcie/driver/linux/include/mailbox_proto.h
richardtobin-xlnx: > In cloud use cases, the version of xclmgmt running in host may not be the same as xocl which runs in guest (such as in Azure).

I submitted the patch in the hope that the situation described wouldn't be a valid usecase for the drivers.  If it is, then I agree that the patch in its current form cannot be applied.  Removing the new padding member completely would help, but Arpit raises issues with the data[0] change.

> This change will also require change in xclmgmt2 driver on host in Azure... someone have to get the buy-in from MS, as to the impact going forward.

This sounds impractical - is that the point you're making here?  If so, then it's another large, probably impassible barrier.

> Also, any reason why char data[1] was changed to data[0]? With current change, it won't compile with Visual Studio compilers. Please move it back to data[1].

The data[0] is part of the fix for the issue.  As described in the CR, the driver allocates its buffers using calculations of the form `sizeof(struct xcl_mailbox_req) + message_length_bytes`, (with the structure size being 16) but starts storing the message inside the data member of `struct xcl_mailbox_req` instead of at the end (at byte offset 12), which leaves 4 bytes at the end of the buffer unused.  These four bytes are awkward to handle in the X3 FW because we don't have the luxury of buffering full messages in RAM like the host drivers - we have to process message data in packets at a time, and expect the last packet of the message to be the end of the data.   With this padding at the end of the message, that expectation does not hold.  Changing to data[0] gobbles up those 4 bytes and makes FW's life easier.

However, if needs to compile for Windows too, we need to find another way.  I've been thinking about this and another way to gobble the extra 4 bytes and maintain compatibility between older and newer XRT drivers would be to change how the drivers calculate their buffer sizes i.e. instead of modifying the structure definition and breaking things, we could leave it alone and change `sizeof(struct xcl_mailbox_req) + message_length_bytes` into `offsetof(struct xcl_mailbox_req, data) + message_length_bytes`  This would eliminate the padding at the end (making FW's life easy) and keep the message data position in the buffer consistent between different drivers.  The only difference is that the buffers allocated by the drivers would be 4 bytes smaller, but since drivers receiving messages 1) learn the necessary buffer size from the first packet and 2) access the data by casting it to `struct xcl_mailbox_req*` and dereferencing the data member, things should work ok.

The main disadvantage of this approach is that every request buffer sizing statement in the drivers would need to be fixed, so the patch to fix the issue would touch more places.  It would also look a little unconventional.

The main question from me now is whether this approach is acceptable to you and Max?

richardtobin-xlnx: In the meantime, I've given proposal a shot and it works with X3, so I've updated this PR with a replacement patch.  Please let me know if you think this revised solution is acceptable.

richardtobin-xlnx:
> @richardtobin-xlnx, I can see that we have checks in the code to make sure the total size of the msg is not less than size of the request data structure. With your change, the old code runs on peer side will still check sizeof(struct xcl_mailbox_req), but the new code on the other side will send 4 bytes less. Wouldn't it break the compatibility of the protocol here?

@maxzhen Yes it would, I missed those checks in my skimming of the code.  I changed a couple of them in my most recent update, but missed at least one in mgmt-core.c, and there may be others.  While I could continue to drive the CI pipeline to find them all and tweak them to match up, the point stands that if there are minimum transfer size is changed, it creates an incompatibility between older and newer versions of the code.

I think I could get around this problem by making sure that I always send `sizeof(struct xcl_mailbox_req)` bytes of data for requests that have no message content e.g. post notifications.  That would preserve the contract about the minimum message size.  However, that leaves the XRT drivers with a mix of `XCL_MAILBOX_REQ_SIZE` and `sizeof(struct xcl_mailbox_req)` which is annoyingly inconsistent and would probably trip someone up eventually.

> I'm not sure I fully understand your problem here. The last 4 bytes contains non-valid data, so can X3 fw just drop it? You don't need to receive the full msg to know that you need to drop them, right?

The FW can drop the data if it must, but it is somewhat inconvenient because of how the code there is currently organized, combined with the RAM limitations that mean we can't buffer a message's content in full.  These circumstances create some unpleasant hoops for us to jump through, and when I looked at what would be involved with changing the XRT side, it seemed to be the simpler and cleaner solution - and from an implementation standpoint, it is.  However, the issues surrounding backward compatibility are not easily resolved, and carry much more weight than having some neater code in the FW.

I knew it might turn out this way, but still thought it was worth a shot.  At this point, I think I've run out of ideas.  If you'd be willing to consider the change I proposed above (changing some things back to guarantee that the minimum message size conditions are not affected), then I am interested in giving this approach one more try.  However, if that's not an acceptable solution for you, I'll discard this PR and close the CR.

maxzhen:
@richardtobin-xlnx , @arpitxilinx and I had offline discussion and we don't see any better options either. The backward compatibility has to be maintained due to the wide deployment of XRT today (on AWS, Azure, etc) and is going to be more important now after we enabled it on X3 firmware. We can't shift data structure members or reduce the size. The data structure started with "char data[0];", but has to be changed to "char data[1];" as @arpitxilinx has explained. We understand that this will cause more memory to be allocated than necessary, but we can't reduce the size back then. So, I think that we can't take your change even today....

The enabling of mailbox protocol on X3 also means that we now have a case where two ends of mailbox may run on different CPU architectures. It is important now to make sure our data structure uses explicit types. So, part of your change still applies. You could modify your PR and keep those changes. Or if you decide to close it, I can make these changes for you to review later. Please let us know what you'd like to do here...

richardtobin-xlnx:
> @richardtobin-xlnx , @arpitxilinx and I had offline discussion and we don't see any better options either. The backward compatibility has to be maintained due to the wide deployment of XRT today (on AWS, Azure, etc) and is going to be more important now after we enabled it on X3 firmware. We can't shift data structure members or reduce the size. The data structure started with "char data[0];", but has to be changed to "char data[1];" as @arpitxilinx has explained. We understand that this will cause more memory to be allocated than necessary, but we can't reduce the size back then. So, I think that we can't take your change even today....

@maxzhen and @arpitxilinx: Thank you for taking the time to consider this issue.  I've come to the same conclusion - the X3 FW will have to adapt.  While this is unfortunate, it comes with the upside that I now have quite a bit of confidence that the padding won't get unintentionally removed by a future cleanup.

> The enabling of mailbox protocol on X3 also means that we now have a case where two ends of mailbox may run on different CPU architectures. It is important now to make sure our data structure uses explicit types. So, part of your change still applies. You could modify your PR and keep those changes. Or if you decide to close it, I can make these changes for you to review later. Please let us know what you'd like to do here...

I think it's better to close the PR, because there's a good bit more stuff in mailbox_proto.h that needs to be changed - there are enums, shorts, unsigned ints and uncounted padding bytes all over it, and the first commit of my PR only addresses a single instance.  I've played in your sandbox enough for the timebeing but if you are willing to take this on, I am happy to review.

PR #6687, opened 2022-05-11
Handle CRLF to LF automatically

stsoe:
Avoid CRLF mess as per #6681 and issue #6672.
Let Git do what it thinks is best.
See https://git-scm.com/docs/gitattributes

Ran `git add --renormalize .`  with the `.gitattributes`  in a workspace prior to #6681 and it converted the same files.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto


keryell (approved):
This is depressingly too easy. :-)

stsoe:
retest this please

stsoe:
retest this please

PR #6692, opened 2022-05-11
CID-240280 overrunning array

mamin506:
Overrunning array issue reported by CID

Because 'index' was uint8_t, if it was assigned as -EAGAIN. The caller will get a result of 245 instead of a negative error code.
Define index as int8_t should fix this issue. 

If cu_bitmap/scu_bitmap is not corrupted, then index will never be -EAGAIN, which is fine.

U50 validate.

No


dayeh-xilinx:
retest this please. fixed git fetch submodule infra error.

maxzhen (approved):

PR #6693, opened 2022-05-11
CR-1127851 Need to check memory connection for DDR calibration

dbenusov-xilinx:
https://jira.xilinx.com/browse/CR-1127851
DDR calibration will be timeout if there is not DDR/PLRAM bank connected.

Previously the tools didn’t support NULL kernels or Kernels which don’t connect to memory. An AR was requested in the past in [CR-1083004](https://jira.xilinx.com/browse/CR-1083004) as we didn’t believe this would be a common use case.

It looks like the tools now support this ability. As this is an uncommon Use Case it looks like we never through to qualify the DDR calibration status in XRT on whether DDR is in use by the xclbin.

Added a check before MIG calibration to verify if any DDR or PLRAM memories are present and connected.

None, fixing a bug for a new tool feature.

Tested on a u250 using the xclbin provided by the reporter.
```
dbenusov@xsjpranjal01:/proj/rdi/staff/dbenusov$ xbutil program --user kernel.xclbin -d 02:00
INFO: xbutil program succeeded on 0000:02:00.1
dbenusov@xsjpranjal01:/proj/rdi/staff/dbenusov$ echo $?
0
dbenusov@xsjpranjal01:/proj/rdi/staff/dbenusov$ dmesg | grep 'No DDR or PLRAM banks connected. Skipping MIG Calibration'
[ 1638.140879] No DDR or PLRAM banks connected. Skipping MIG Calibration
[ 2963.434743] No DDR or PLRAM banks connected. Skipping MIG Calibration
```

None


houlz0507 (approved):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
```diff
@@ -650,6 +651,22 @@ static int calibrate_mig(struct icap *icap)
 {
 	int i;
 
+	/* Check for any DDR or PLRAM banks that are in use */
+	bool is_memory_bank_connected = false;
+
+	/* If a DDR or PLRAM bank is found no need to keep searching */
+	for (i = 0; (i < icap->mem_topo->m_count) && (!is_memory_bank_connected); i++) {
+		struct mem_data* mem_bank = &icap->mem_topo->m_mem_data[i];
+		if ((IS_DDR(mem_bank->m_tag) || IS_PLRAM(mem_bank->m_tag)) && (mem_bank->m_used != 0))
```
maxzhen, line 660: We don't need to search for HBM?

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
```diff
@@ -650,6 +651,22 @@ static int calibrate_mig(struct icap *icap)
 {
 	int i;
 
+	/* Check for any DDR or PLRAM banks that are in use */
+	bool is_memory_bank_connected = false;
+
+	/* If a DDR or PLRAM bank is found no need to keep searching */
+	for (i = 0; (i < icap->mem_topo->m_count) && (!is_memory_bank_connected); i++) {
+		struct mem_data* mem_bank = &icap->mem_topo->m_mem_data[i];
+		if (MEM_NEEDS_CALIBRATION(mem_bank->m_tag) && (mem_bank->m_used != 0))
```
houlz0507, line 660: PLRAM is DRAM type. So I suggest to change as below:
if ((convert_mem_type(mem_topo->m_mem_data[i].m_tag) == MEM_DRAM ||
     mem_topo->m_mem_data[i].m_type == MEM_DRAM) && (mem_bank->m_used != 0))
@maxzhen Please comment if you have better thought.

src/runtime_src/core/pcie/driver/linux/xocl/xocl_drm.h
```diff
@@ -24,8 +25,14 @@
 
 typedef void (*xocl_execbuf_callback)(unsigned long data, int error);
 
-#define IS_HOST_MEM(m_tag)	(!strncmp(m_tag, "HOST[0]", 7))
-#define IS_PLRAM(m_tag)		(!strncmp(m_tag, "PLRAM[", 6))
+#define IS_HOST_MEM(m_tag)				(!strncmp(m_tag, "HOST[0]", 7))
+#define IS_PLRAM(m_tag)					(!strncmp(m_tag, "PLRAM[", 6))
+#define IS_DDR(m_tag)					(!strncmp(m_tag, "DDR[", 4))
```
houlz0507, line 30: The IS_DDR(m_tag) is incorrect because the legacy platform (u200) use different tag. Please revert this change.

src/runtime_src/core/pcie/driver/linux/xocl/xocl_drm.h
```diff
@@ -24,8 +25,14 @@
...
+/* 
+ * Returns if the memory bank associated with the given tag should be 
+ * calibrated. Currently only DDR and  PLRAM must be calibrated.
+ */
+#define MEM_NEEDS_CALIBRATION(m_tag)	(IS_DDR(m_tag) || IS_PLRAM(m_tag))
```
houlz0507, line 35: This needs to be removed too. Please see my recommended change in icap.c.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
```diff
@@ -650,6 +651,22 @@ static int calibrate_mig(struct icap *icap)
...
+			is_memory_bank_connected = true;
+	}
+
+	// If no DDR or PLRAM banks are in use there is nothing to calibrate
+	if (!is_memory_bank_connected) {
+		pr_info("No DDR or PLRAM banks connected. Skipping MIG Calibration\n");
```
houlz0507, line 666: please do not use pr_info but use ICAP_INFO which prints more device info

houlz0507 (changes requested):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
maxzhen: We should not use m_type from xclbin directly. It is not accurate and sometimes completely wrong. We should just use m_tag and convert it to type. So, I think we should extend convert_mem_type() and use it everywhere else for type conversion. Can we just convert "PLRAM" to DRAM?

src/runtime_src/core/pcie/driver/linux/xocl/xocl_drm.h
maxzhen: We should keep this macro and just fix the logic. We should only use m_tag to tell the type. The m_type in xclbin shall be ignored completely for historical reason.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
dbenusov-xilinx: There are instances where PLRAM needs to be considered separate from DRAM so one function would not cut it. However, we could make a new function that returns an enumerated version of the tag. That would work very nicely. We end up with a new enum that describes the type of memory tag we have read.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
```diff
@@ -1267,21 +1290,50 @@ static void icap_clean_bitstream_axlf(struct platform_device *pdev)
 	icap_clean_axlf_section(icap, PARTITION_METADATA);
 }
 
-static uint32_t convert_mem_type(const char *name)
+enum MEM_TAG convert_mem_tag(const char *name)
```
houlz0507, line 1293: This function is shared by all the sub device drivers. Please move it to xocl_xclbin.c

src/runtime_src/core/pcie/driver/linux/xocl/xocl_drm.h
```diff
@@ -24,8 +25,15 @@
...
-#define IS_PLRAM(m_tag)		(!strncmp(m_tag, "PLRAM[", 6))
+enum MEM_TAG {
+	MEM_TAG_DDR = 0,
+	MEM_TAG_PLRAM,
+	MEM_TAG_HOST,
+	MEM_TAG_HBM,
+	MEM_TAG_INVALID
+};
+
+enum MEM_TAG convert_mem_tag(const char *name);
```
houlz0507, line 36: Please move the enum and convert_mem_tag() to xocl_xclbin.h

dayeh-xilinx:
test this please.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
```diff
@@ -1267,21 +1290,50 @@ static void icap_clean_bitstream_axlf(struct platform_device *pdev)
...
 	else if (!strncasecmp(name, "bank", 4))
-		mem_type = MEM_DRAM;
+		mem_tag = MEM_TAG_DDR;
+	else if (!strncasecmp(name, "HOST[0]", 7))
+		mem_tag = MEM_TAG_HOST;
+
+	return mem_tag;
+}
+
+static enum MEM_TYPE convert_mem_type(const char *name)
```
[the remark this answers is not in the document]
dbenusov-xilinx: I am not sure this is a good idea, it is very convenient to group the memory tags by their types and perform actions on them as a group rather than spelling out each individual name. For example one of the remaining uses of the convert_mem_type function perform actions depends on the ECC memory type (icap.c::icap_save_calib). If we remove this function we would need to list out every memory tag that is a DRAM type. We would end up with a macro to group the types together and at that point we might as well keep the function around.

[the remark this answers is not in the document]
houlz0507: m_type is not accurate and confusing. We should use tag string to category/group banks. you already created a tag enum to which can category/group tag strings, correct? Why would we still need to group by m_type? 
For icap_save_calib, isn't MEM_TAG_DDR good enough? convert_mem_tag() returns MEM_TAG_DDR for both 'bank' and 'DDR', correct?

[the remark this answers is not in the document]
dbenusov-xilinx: Okay, I though once we converted the memory types they were valid. I'll refactor the functions to use the tag instead.

houlz0507 (approved):

xilinxgitops:
retest this please

stsoe (approved):

PR #6696, opened 2022-05-12
Use explicit types in mailbox protocol

maxzhen:
After mailbox is supported on X3, the two ends of mailbox communication channel may run on different CPU architectures. It is important to use explicit type to avoid any mis-match in mailbox protocol data structures.
N/A
Use explicit types
Should not change any data structure in binary
Tested on U55n with xbutil validate
N/A

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/include/mailbox_proto.h
+++ b/src/runtime_src/core/pcie/driver/linux/include/mailbox_proto.h
@@ -353,7 +353,7 @@ struct xcl_mailbox_clock_freqscaling {
 struct xcl_mailbox_req {
 	uint64_t flags;
 	int32_t req;
-	uint8_t data[4]; /* variable length of payload, explicitly padded to int32_t */
+	int8_t data[4]; /* variable length of payload, explicitly padded to int32_t */
 };
 
 /**
@@ -373,7 +373,7 @@ struct xcl_sw_chan {
 	uint64_t sz;
 	uint64_t flags;
 	uint64_t id;
-	uint8_t data[4]; /* variable length of payload, explicitly padded to int32_t */
+	int8_t data[4]; /* variable length of payload, explicitly padded to int32_t */
 };
 
 /**
--- a/src/runtime_src/core/pcie/driver/linux/include/mailbox_proto.h
+++ b/src/runtime_src/core/pcie/driver/linux/include/mailbox_proto.h
@@ -136,6 +136,7 @@ struct xcl_board_info {
 	int8_t		exp_bmc_ver[256];
 	uint32_t	mac_contiguous_num;
 	int8_t		mac_addr_first[6];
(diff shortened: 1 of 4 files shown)


xuhz (approved):

arpitxilinx (approved):

richardtobin-xlnx (changes requested):
There's still some silent padding:

https://godbolt.org/z/9aM67raoK

The above compiles the header for both 64-bit and 32-bit mode, and also compiles it with  structure packing enabled.  The sizeof() operator should give consistent results for all three compilations to ensure that all padding is present and accounted for.  It does not - the discrepancies are as follows:
 - `struct xcl_board_info` (1432 on both 64-bit and machines, but is 1430 when packed)
 - `struct xcl_sensor` (240 on 64-bit machine vs 236 on 32-bit machine)
 - `struct xcl_firewall` (104 on 64-bit machine vs 100 on 32-bit machine vs 98 when packed)
 - `struct xcl_sw_chan` (32 on 64-bit machine vs 28 on 32-bit machine)

Technically the 32-bit compilation is redundant for detecting the padding - I included it to show how it's inadequate.

Please fix up the listed structure definitions to make their tail padding explicit.

src/runtime_src/core/pcie/driver/linux/include/mailbox_proto.h
```diff
@@ -338,22 +338,22 @@ struct xcl_mailbox_bitstream_kaddr {
...
+ * @req: opcode (enum xcl_mailbox_request)
  * @flags: flags of this message
  * @data: payload of variable length
  */
 struct xcl_mailbox_req {
 	uint64_t flags;
-	enum xcl_mailbox_request req;
-	char data[1]; /* variable length of payload */
+	int32_t req;
+	int8_t data[4]; /* variable length of payload, explicitly padded to int32_t */
```
richardtobin-xlnx, line 356: If I was doing this, I would have just added the padding as a separate definition - same comment for `struct xcl_sw_chan`.

maxzhen: Hi @richardtobin-xlnx ,

This is what I initially thought about. But if I keep "data" as the last member, it is more consistent with the fact that it is the starting point of a variable length chunk of payload in the same data structure buffer. Adding another member after it confuses me. Anyway, I changed the type (vs making it an array) but still keep it as the last one.

I also fixed the other data structures that you pointed out. Thanks for pointing me to the compiler tool. It was very helpful. Now I can see that all data structure size are the same without or without packing on both 32 and 64 bit platforms.

Please take another look.

maxzhen:
retest this please - known issue.

maxzhen:
retest this please

richardtobin-xlnx (approved):
Looks ok now, thanks

PR #6421, opened 2022-03-15
Add support for multiple xclbins in user mode XRT

stsoe:
#### Problem solved by the commit
Support for multiple xclbins

#### How problem was solved, alternative solutions (if any) and why they were rejected
This is a monster PR with many changes.   The changes are described in each commit to this PR.
Briefly:

- shim level changes to add xclOpenContextByName
- core infrastructure changes to manage multiple xclbins assigned to a slot
- core infrastructure changes to encode slot index in BO allocation flags
- leverage driver data for 'kds_cu_info' and 'xclbin_slots'
- compute unit indices are per 'kds_cu_info' which is assumed to be valid for a given cu only after a context is successfully acquired on that cu.

#### Risks (if any) associated the changes in the commit
This is a significant departure from single xclbin assumption.   For existing flows that support 
only a single xclbin, the xclbin will be managed in a single slot.

#### What has been tested and how, request additional testing if necessary
The shim level 'noop' driver was updated to support multiple xclbins that can be loaded into
slots.   This was leveraged during testing and debugging of this feature.

PR #6698, opened 2022-05-13
New API xrt::hw_context

stsoe:
New concept to represent hardware resources (columns, regions, PS
processes) used by an xclbin.  A hardware context is used by an
application to control where execution takes place.  In a multi-xclbin
scenario, the hardware context is specific to one xclbin even if that
xclbin is loaded multiple times.

The feature is optional for legacy applications / platforms.

Constructed from a device and an xclbin uuid that has been registered
with the device (driver)

A context is 1-1 with a slot index. A resource resolver assigns a
hardware context to hardware resources, maybe sharing, or fails.
On success, the driver assigns a unique slot index to the context.

A kernel object is created from a hardware context. CU context
creation no longer needs EAGAIN as opening of CU context always
succeeds if the hardware context was successfully created. The rather
complicated code for EAGAIN per #6421 has been removed in this PR.

A hardware context can be shared between threads. A process
can construct multiple hardware contexts.

### DIFF
--- a/tests/xrt/multi_xclbin/simple.cpp
+++ b/tests/xrt/multi_xclbin/simple.cpp
@@ -12,13 +12,10 @@ multiple xclbins
 
 #include "experimental/xrt_hw_context.h"
 
-#include <array>
-#include <atomic>
-#include <cstdlib>
-#include <fstream>
+#include <algorithm>
+#include <stdexcept>
 #include <iostream>
-#include <list>
-#include <thread>
+#include <string>
 #include <vector>
 
 #ifdef _WIN32
--- a/tests/xrt/multi_xclbin/simple.cpp
+++ b/tests/xrt/multi_xclbin/simple.cpp
@@ -1,0 +1,128 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
+/****************************************************************
+This example illustrates the use of xrt::hw_context for working with
+multiple xclbins
+
+% g++ -g -std=c++17 -I$XILINX_XRT/include -L$XILINX_XRT/lib -o simple.exe simple.cpp -lxrt_coreutil -luuid -pthread
+****************************************************************/
+#include "xrt/xrt_bo.h"
+#include "xrt/xrt_device.h"
+#include "xrt/xrt_kernel.h"
+
+#include "experimental/xrt_hw_context.h"
+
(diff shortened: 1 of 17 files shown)


tests/xrt/multi_xclbin/simple.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+#include "experimental/xrt_hw_context.h"
+
+#include <array>
+#include <atomic>
+#include <cstdlib>
+#include <fstream>
+#include <iostream>
+#include <list>
+#include <thread>
```
keryell, line 21: Eek! Be modest in your inclusions. :-) Or is this from the `#include <C++>` diversity&inclusion group?

tests/xrt/multi_xclbin/simple.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+#ifdef _WIN32
+# pragma warning ( disable : 4267 )
+#endif
+
+// Kernel specifics
+// void addone (__global ulong8 *a, __global ulong8 * b, unsigned int  elements)
+// addone(a, b, ELEMENTS)
+// The kernel is compiled with 8 CUs same connectivity.
+static constexpr size_t elements = 16;
```
keryell, line 32: Do you really need all these `static` everywhere in a such single-file example? Same for the functions.
Nowadays the anonymous `namespace` might be more fashionable for this anyway.

tests/xrt/multi_xclbin/simple.cpp
```diff
@@ -0,0 +1,128 @@
...
+static constexpr size_t array_size = 8;
+static constexpr size_t maxcus = 8;
```
keryell, line 34:
```suggestion
constexpr size_t max_cus = 8;
```

keryell (changes requested):
Quite interesting feature!

keryell:
@Ralender context! How does this impact the future of https://github.com/triSYCL/sycl/pull/170 ?

stsoe:
@saifuddin-xilinx @chvamshi-xilinx   I added you both as reviewer, most FYI.  I don't think the changes in this PR affects vck5000 or builtin xclbin right now.   We will meet some time this week to summarize the new APIs.  The slides (spec) are a bit like work-in-progress, but you should get the the idea by checking the slide about the new xcl APIs that are being added.

saifuddin-xilinx (approved):

PR #6700, opened 2022-05-15
VITIS-5012 C++ API for Asynchronous GMIO transfer in AIE

sarab96:
New feature: Asynchronous GMIO txfer in AIE with C++ API
None
None. Possible to use other xcl or C++ APIs
Tested locally on VCK190
To be updated


stsoe (commented):
Let's look at this together.  I imagine async_bo_impl to be a `struct async_bo_something` under class xrt::bo.   I don't think it need to be a class exposed at xrt:: namespace level.

sarab96:
Will submit new PR after changes as per internal discussion

PR #6704, opened 2022-05-16
Remove deprecated QDMA streaming APIs

stsoe:
Per deprecation notice in 2.11, the shim level streaming APIs are now removed.


sonals (approved):
Thanks for making the changes

Issue #6649, opened 2022-04-29
Compilation warning error with g++ 11.2 `src/runtime_src/core/pcie/emulation/cpu_em/generic_pcie_hal2/shim.cxx`


keryell:
Compilation error with g++ 11.2
```
/var/tmp/rkeryell/SYCL/XRT/src/runtime_src/core/pcie/emulation/cpu_em/generic_pcie_hal2/shim.cxx: In member function ‘int xclcpuemhal2::CpuemShim::xclRegRW(bool, uint32_t, uint32_t, uint32_t*)’:
/var/tmp/rkeryell/SYCL/XRT/src/runtime_src/core/pcie/emulation/cpu_em/generic_pcie_hal2/shim.cxx:983:14: error: ‘*(unsigned int*)((char*)&buff + offsetof(std::array<char, 4>,std::array<char, 4>::_M_elems[0]))’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  983 |       *datap = tmp_buff[0];
      |       ~~~~~~~^~~~~~~~~~~~~
```
Honestly, considering the state of this code, compiling with `-Werror` is challenging...
So for now I am using the shortcut from https://github.com/Xilinx/XRT/issues/6640#issuecomment-1112659501

bryanloz-xilinx:
@uday610 Looks like we can close this one.

PR #6713, opened 2022-05-19
Fix a warning

Ralender:
I prevented the warning from appearing but it is likely a real bug. since `xclRegRead_RPC_CALL` will not write anything to the buffer in case of error so `*datap = tmp_buff[0];` was reading uninitialized data in that case before this patch.

@stsoe is it possible to stop forcing to compile with -Werror. because it makes XRT depend on a specific compiler version. which makes it harder to compile XRT anywhere. warnings should be fixed but often like here the person facing the issue is not the one able to fixe the fundamental issue.


keryell:
I guess this is related to my problem with https://github.com/Xilinx/XRT/issues/6649

stsoe:
ok to test

PR #6715, opened 2022-05-19
CR-1110481 Enhance xbutil p2p test commands

dbenusov-xilinx:
https://jira.xilinx.com/browse/CR-1110481
At the moment the p2p command does not generate write combine as it writes one byte at a time.

Changed the p2p test to write in 1024 byte blocks.

The p2p is is slightly slower due to the increased overhead of the new tests.

On a u50 using Ubuntu 20.04
```
root@xsjminm50:~# xbutil validate -r p2p -d 03:00
Starting validation for 1 devices

Validate Device           : [0000:03:00.1]
    Platform              : xilinx_u50_gen3x16_xdma_base_5
    SC Version            : 5.2.4
    Platform ID           : 44654095-25B4-C06A-EC6D-0B479D3FEBE8
[PASSED] :  < 31s >
Test 1 [0000:03:00.1]     : p2p
    Details               : HBM[0] validated
                            HBM[1] validated
                            HBM[2] validated
                            HBM[3] validated
                            HBM[4] validated
                            HBM[5] validated
                            HBM[6] validated
                            HBM[7] validated
                            HBM[8] validated
                            HBM[9] validated
                            HBM[10] validated
                            HBM[11] validated
                            HBM[12] validated
                            HBM[13] validated
                            HBM[14] validated
                            HBM[15] validated
                            HBM[16] validated
                            HBM[17] validated
                            HBM[18] validated
                            HBM[19] validated
                            HBM[20] validated
                            HBM[21] validated
                            HBM[22] validated
                            HBM[23] validated
                            HBM[24] validated
                            HBM[25] validated
                            HBM[26] validated
                            HBM[27] validated
                            HBM[28] validated
                            HBM[29] validated
                            HBM[30] validated
                            HBM[31] validated
Validation completed. Please run the command '--verbose' option for more details
```

None


src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -428,13 +428,18 @@ p2ptest_set_or_cmp(char *boptr, size_t size, char pattern, bool set)
...
+  const size_t write_size = 1024;
   assert((size % stride) == 0);
   for (size_t i = 0; i < size; i += stride) {
-    if (set) {
-      boptr[i] = pattern;
-    }
-    else if (boptr[i] != pattern) {
-      return false;
+    if (set)
+      // Write a large block of datato trigger a write combine
```
keryell, line 435:
```suggestion
      // Write a large block of data to trigger a write combine
```

dbenusov-xilinx: Good catch!

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -428,13 +428,30 @@ p2ptest_set_or_cmp(char *boptr, size_t size, char pattern, bool set)
...
+  std::memset(valid_data, pattern, write_size);
+
   assert((size % stride) == 0);
   for (size_t i = 0; i < size; i += stride) {
-    if (set) {
-      boptr[i] = pattern;
-    }
-    else if (boptr[i] != pattern) {
-      return false;
+    // Due to test timeout we can only have one large write to test write combines
```
rozumx, line 438: Did you look into refactoring the test timeout code base?

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -428,13 +428,30 @@ p2ptest_set_or_cmp(char *boptr, size_t size, char pattern, bool set)
...
+    // The first write in a memory segment is a large write combine while all others are single bytes
+    if (set)
+      if(i == 0)
+        // Write a large block of data to trigger a write combine
+        std::memcpy(&(boptr[i]), valid_data, write_size);
```
rozumx, line 443: What happens if "size" is 100 bytes in size?

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -428,13 +428,30 @@ p2ptest_set_or_cmp(char *boptr, size_t size, char pattern, bool set)
...
   for (size_t i = 0; i < size; i += stride) {
-    if (set) {
-      boptr[i] = pattern;
-    }
-    else if (boptr[i] != pattern) {
-      return false;
+    // Due to test timeout we can only have one large write to test write combines
+    // The first write in a memory segment is a large write combine while all others are single bytes
+    if (set)
+      if(i == 0)
```
rozumx, line 441: This code is screaming to be its own method for 1 flow is using a character (e.g., pattern) to initialize the memory and the other a "buffer" of data.

Yes, in this case it is the same character, but this code should be refactor to separate the difference algorithm to allow for future different memory pattern blocks.

rozumx (changes requested):

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
dbenusov-xilinx: Performing only large writes of 1024 bytes would cause the test would take over eight minutes for a u280! I figured the better choice is to shorten the test.

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
dbenusov-xilinx: Let me know if you like the changes!

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
dbenusov-xilinx: I added a new assert into the writing function to validate size and the validate buffer size

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -424,22 +424,35 @@ free_unmap_bo(xclDeviceHandle handle, xclBufferHandle boh, void * boptr, size_t
  * helper function for P2P test
  */
 static bool
-p2ptest_set_or_cmp(char *boptr, size_t size, char pattern, bool set)
+p2ptest_set_or_cmp(char *boptr, size_t size, char *valid_data, size_t valid_data_size, bool set)
```
rozumx, line 427: The code below is using a vector of chars.  Instead of breaking in to a buffer and size, having an argument (e.g., `std::vector<char> & validData`), would be a better and more scalable solution. 

```suggestion
  p2ptest_set_or_cmp(char *boptr, size_t size, const std::vector<char> & validData, bool set)
```

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -424,22 +424,35 @@ free_unmap_bo(xclDeviceHandle handle, xclBufferHandle boh, void * boptr, size_t
...
 {
   int stride = xrt_core::getpagesize();
 
   assert((size % stride) == 0);
+  assert(size > valid_data_size);
```
rozumx, line 432: Using the vector, this code becomes:
```suggestion
  assert(size > validData.size());
```

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -424,22 +424,35 @@ free_unmap_bo(xclDeviceHandle handle, xclBufferHandle boh, void * boptr, size_t
...
   for (size_t i = 0; i < size; i += stride) {
-    if (set) {
-      boptr[i] = pattern;
-    }
-    else if (boptr[i] != pattern) {
-      return false;
-    }
+    if (set)
+      std::memcpy(&(boptr[i]), valid_data, valid_data_size);
```
rozumx, line 435: Using the vector, this code becomes:
```suggestion
      std::memcpy(&(boptr[i]), validData.data(), validData.size());
```

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -424,22 +424,35 @@ free_unmap_bo(xclDeviceHandle handle, xclBufferHandle boh, void * boptr, size_t
...
+    else if (std::memcmp(&(boptr[i]), valid_data, valid_data_size) != 0) // Verify the written bytes
```
rozumx, line 436: No need to check for not zero.
```suggestion
    else if (std::memcmp(&(boptr[i]), validData.data(), validData.size())) // Verify the written bytes
```

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -424,22 +424,35 @@ free_unmap_bo(xclDeviceHandle handle, xclBufferHandle boh, void * boptr, size_t
...
+ */
+static bool
+p2ptest_set_or_cmp(char *boptr, size_t size, char pattern, bool set)
+{
+
+  // Generate the valid data vector using the given pattern
+  std::vector<char> valid_data;
+  valid_data.push_back(pattern);
+
+  return p2ptest_set_or_cmp(boptr, size, valid_data.data(), 1, set);
```
rozumx, line 453: Magic numbers are bad.
```suggestion
  return p2ptest_set_or_cmp(boptr, size, valid_data.data(), valid_data.size(), set);
```

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -451,6 +464,20 @@ p2ptest_chunk(xclDeviceHandle handle, char *boptr, uint64_t dev_addr, uint64_t s
   if (xrt_core::posix_memalign(reinterpret_cast<void **>(&buf), xrt_core::getpagesize(), size))
     return false;
 
+  // Generate the valid data vector
+  const size_t valid_data_size = 1024;
+  char valid_data[valid_data_size];
```
rozumx, line 469: Please use a vector of chars.

rozumx (changes requested):

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -451,6 +464,20 @@ p2ptest_chunk(xclDeviceHandle handle, char *boptr, uint64_t dev_addr, uint64_t s
...
+  std::memset(valid_data, 'A', valid_data_size);
+
+  // Perform a memory write larger than 512 bytes to trigger a write combine
+  auto buf_size = xrt_core::getpagesize() * 1;
```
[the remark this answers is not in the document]
dbenusov-xilinx: If more than one write was desired the number could be changed. I am not sure if I should include this at all. A variable would definitely be a better idea

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -424,18 +424,17 @@ free_unmap_bo(xclDeviceHandle handle, xclBufferHandle boh, void * boptr, size_t
  * helper function for P2P test
  */
 static bool
-p2ptest_set_or_cmp(char *boptr, size_t size, char pattern, bool set)
+p2ptest_set_or_cmp(char *boptr, size_t size, std::vector<char>& valid_data, bool set)
```
rozumx, line 427:
```suggestion
p2ptest_set_or_cmp(char *boptr, size_t size, const std::vector<char>& valid_data, bool set)
```

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -424,18 +424,17 @@ free_unmap_bo(xclDeviceHandle handle, xclBufferHandle boh, void * boptr, size_t
...
   assert((size % stride) == 0);
+  assert(size > valid_data.size());
   for (size_t i = 0; i < size; i += stride) {
-    if (set) {
-      boptr[i] = pattern;
-    }
-    else if (boptr[i] != pattern) {
-      return false;
-    }
+    if (set)
```
rozumx, line 434: This code could be refactored to be more readable.  For example:

```
    if (!set)    // Validating memory
      return std::equal(valid_data.begin(), valid_data.end(), &(boptr[i]));

    // Setting memory
    std::memcpy(&(boptr[i]), valid_data.data(), valid_data.size());
```

rozumx (changes requested):

rozumx (approved):

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -427,18 +427,17 @@ free_unmap_bo(xclDeviceHandle handle, xclBufferHandle boh, void * boptr, size_t
  * helper function for P2P test
  */
 static bool
-p2ptest_set_or_cmp(char *boptr, size_t size, char pattern, bool set)
+p2ptest_set_or_cmp(char *boptr, size_t size, const std::vector<char>& valid_data, bool set)
 {
   int stride = xrt_core::getpagesize();
 
   assert((size % stride) == 0);
+  assert(size > valid_data.size());
```
chvamshi-xilinx, line 435: is it >= or just >?

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -427,18 +427,17 @@ free_unmap_bo(xclDeviceHandle handle, xclBufferHandle boh, void * boptr, size_t
...
   for (size_t i = 0; i < size; i += stride) {
-    if (set) {
-      boptr[i] = pattern;
-    }
-    else if (boptr[i] != pattern) {
+    if (set)
+      std::memcpy(&(boptr[i]), valid_data.data(), valid_data.size());
```
chvamshi-xilinx, line 438: This function is writing character(s) at each page boundary. is it expected? You just want to verify only only some characters at page_boundary to reduce runtime of this test? If yes, better to add one comment explaining this

if valid_data size is more than page size, then you are overwriting boptr. ( Thats not the case here in Linux as you are sending maximum 1k data size whereas page size is 4k. Better to add explicit comment)

chvamshi-xilinx (changes requested):

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
dbenusov-xilinx: The purpose of this function is to write/check some piece of data in all pages within the given size constraint. The size given by the user determines how many pages will be tested. It would be better to pass the number of pages to test.

I will put a check for that, I could see an issue with writing over the boundary

chvamshi-xilinx (approved):
Looks good.
Thanks

PR #6723, opened 2022-05-23
Remove virtual dtor from xrt::detail::pimpl

stsoe:
XRT native APIs are not polymorphic and cannot be destructed through
the pimpl base class.

A virtual pimpl base class adds a vtable pointer to each native class
object, and this vtable is not useful at all.  In fact it creates link
problems at least on windows if the first function is inline.

Remove the virtual dtor from xrt::detail::pimpl.

This change breaks ABI compatibility of any XRT API object that is implemented using the pimpl class.
Only two experimental APIs fall under this category
* `xrt::xclbin`
* `xrt::ip`

Applications using these would have to be recompiled.   The change in this PR is justified by the pimpl class being used only by experimental APIs.

Verify compilation works.

### DIFF
--- a/src/runtime_src/core/include/xrt/detail/pimpl.h
+++ b/src/runtime_src/core/include/xrt/detail/pimpl.h
@@ -28,11 +28,11 @@ namespace xrt { namespace detail {
 template<typename ImplType>
 class pimpl
 {
+protected:
+  ~pimpl() = default;
+
 public:
   pimpl() = default;
-
-  virtual
-  ~pimpl() = default;
 
   explicit
   pimpl(std::shared_ptr<ImplType> handle)


src/runtime_src/core/include/xrt/detail/pimpl.h
```diff
@@ -28,12 +28,12 @@ namespace xrt { namespace detail {
 template<typename ImplType>
 class pimpl
 {
+protected:
+  ~pimpl() = default;
```
keryell, line 32: Do you need it at all?

stsoe: Likely not.  But at least protected now, not that it matters really.

PR #6730, opened 2022-05-25
CR-1129798 No Error Message for mixed versions of XRT in release 2022.1

rchane:
https://jira.xilinx.com/browse/CR-1129798
In 2022.1, an error message is no longer given to warn users of mixed version usage. This was noticed while running `xbutil examine` on a container with a different XRT version than its host.
The previous warning was re-adapted to current ptree structure and command flags. The logic underneath is essentially the same as [before legacy tools were removed](https://github.com/Xilinx/XRT/pull/6570) to only produce the error message in the given conditions.
Users must ensure that their XRT versions match to use xbutil properly without throwing this error (unless they use --force, but they will still see the warning message).
EDIT: Updated error message/showing usage with --force.
```
rchane@xsjbrd102bx:/proj/rdi/staff/rchane/XRT$ xbutil examine
XRT build version: 2.14.0
Build hash: 6bd8485bd340219087c5c08363ffedf13ec4e51b
Build date: 2022-07-20 11:46:14
Git branch: CR-1129798
PID: 317247
UID: 90979
[Wed Jul 20 19:47:22 2022 GMT]
HOST: xsjbrd102bx
EXE: /proj/xsjhdstaff4/rchane/XRT/build/Debug/opt/xilinx/xrt/bin/unwrapped/xbutil2
[xbutil] ERROR: Unexpected XRT Driver version (unknown) was found. Expected 2.14.0, to match XRT tools.

rchane@xsjbrd102bx:/proj/rdi/staff/rchane/XRT$ xbutil examine --force
Unexpected XRT Driver version (unknown) was found. Expected 2.14.0, to match XRT tools.
Bypassing mismatch with --force.
System Configuration
  OS Name              : Linux
  Release              : 5.4.0-100-generic
  Version              : #113-Ubuntu SMP Thu Feb 3 18:43:29 UTC 2022
  Machine              : x86_64
  CPU Cores            : 6
[more examine info...]
```
N/A

### DIFF
--- a/src/runtime_src/core/tools/common/XBMain.cpp
+++ b/src/runtime_src/core/tools/common/XBMain.cpp
@@ -97,7 +97,7 @@ void  main_(int argc, char** argv,
   }
 
   // If not asking for version or help in xbutil, ensure the versions of XRT and xbutil match
-  if (!bHelp && isUserDomain)
+  if (!bHelp && !bForce && isUserDomain)
     XBU::xrt_xbutil_version_cmp();
  
   // -- Enable/Disable helper "global" options
--- a/src/runtime_src/core/tools/common/XBMain.cpp
+++ b/src/runtime_src/core/tools/common/XBMain.cpp
@@ -95,6 +95,10 @@ void  main_(int argc, char** argv,
     std::cout << XBU::get_xrt_pretty_version();
     return;
   }
+
+  // If not asking for version or help in xbutil, ensure the versions of XRT and xbutil match
+  if (!bHelp && isUserDomain)
+    XBU::xrt_xbutil_version_cmp();
  
   // -- Enable/Disable helper "global" options
   XBU::disable_escape_codes( bBatchMode );
--- a/src/runtime_src/core/tools/common/XBMain.cpp
+++ b/src/runtime_src/core/tools/common/XBMain.cpp
@@ -97,8 +97,8 @@ void  main_(int argc, char** argv,
   }
 
   // If not asking for version or help in xbutil, ensure the versions of XRT and xbutil match
-  if (!bHelp && !bForce && isUserDomain)
-    XBU::xrt_xbutil_version_cmp();
+  if (!bHelp && isUserDomain)
+    XBU::xrt_xbutil_version_cmp(bForce);
  
   // -- Enable/Disable helper "global" options
   XBU::disable_escape_codes( bBatchMode );
--- a/src/runtime_src/core/tools/common/XBUtilities.cpp
+++ b/src/runtime_src/core/tools/common/XBUtilities.cpp
@@ -248,20 +248,27 @@ str2index(const std::string& str, bool _inUserDomain)
 }
 
 void
-XBUtilities::xrt_xbutil_version_cmp()
+XBUtilities::xrt_xbutil_version_cmp(bool force)
 {
   boost::property_tree::ptree pt_xrt;
   xrt_core::get_xrt_info(pt_xrt);
-  boost::property_tree::ptree empty_ptree;
+  const boost::property_tree::ptree empty_ptree;
 
-  std::string xrt_version = pt_xrt.get<std::string>("version", "N/A");
+  std::string xrt_version = pt_xrt.get<std::string>("version", "<unknown>");
   const boost::property_tree::ptree& available_drivers = pt_xrt.get_child("drivers", empty_ptree);
-  for(auto& drv : available_drivers) {
+  for(const auto& drv : available_drivers) {
     const boost::property_tree::ptree& driver = drv.second;
-    std::string drv_name = driver.get<std::string>("name", "N/A");
-    std::string drv_version = driver.get<std::string>("version", "N/A");
-    if (drv_name.compare("xocl") == 0 && drv_version.compare("unknown") != 0 && xrt_version.compare(drv_version) != 0)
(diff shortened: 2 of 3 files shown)


src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -241,6 +241,24 @@ str2index(const std::string& str, bool _inUserDomain)
...
+  boost::property_tree::ptree empty_ptree;
+
+  std::string xrt_version = pt_xrt.get<std::string>("version", "N/A");
+  const boost::property_tree::ptree& available_drivers = pt_xrt.get_child("drivers", empty_ptree);
+  for(auto& drv : available_drivers) {
+    const boost::property_tree::ptree& driver = drv.second;
+    std::string drv_name = driver.get<std::string>("name", "N/A");
+    std::string drv_version = driver.get<std::string>("version", "N/A");
+    if (drv_name.compare("xocl") == 0 && drv_version.compare("unknown") != 0 && xrt_version.compare(drv_version) != 0)
+      throw std::runtime_error("Mixed versions of XRT and xbutil are not supported. Please install matching versions of XRT and xbutil.");
```
rozumx, line 258: This solution does report the current build / version, but still doesn't help the user in understanding what is different.

It would be better the difference in versions was reported.   Something similar to but not limited to:

- Unknown driver version

- Unexpected XRT Driver version (2.13.0) found.  Was expected 2.14.0 which compatible with this version of XRT tools.

rozumx (dismissed):

src/runtime_src/core/tools/common/XBUtilities.cpp
rchane: Hi Stephen,

Thank you for the feedback; I have updated the error messages! However, while doing so, I realized that drivers with unknown versions were being let through without throwing an error. Do we want to continue allowing this? When throwing an error for unknown driver version, pipeline builds ended up failing due to the issue, though I am not entirely sure if this is expected behavior.

chvamshi-xilinx:
retest this please

chvamshi-xilinx (changes requested):

src/runtime_src/core/tools/common/XBMain.cpp
```diff
@@ -95,6 +95,10 @@ void  main_(int argc, char** argv,
     std::cout << XBU::get_xrt_pretty_version();
     return;
   }
+
+  // If not asking for version or help in xbutil, ensure the versions of XRT and xbutil match
+  if (!bHelp && isUserDomain)
```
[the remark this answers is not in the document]
rchane: Sounds good! As for the latter question, I believe the CR specifically wanted the xbutil message to return, but we can also do this for xbmgmt if desired.

chvamshi-xilinx (approved):

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -241,6 +241,24 @@ str2index(const std::string& str, bool _inUserDomain)
...
+{
+  boost::property_tree::ptree pt_xrt;
+  xrt_core::get_xrt_info(pt_xrt);
+  boost::property_tree::ptree empty_ptree;
+
+  std::string xrt_version = pt_xrt.get<std::string>("version", "N/A");
+  const boost::property_tree::ptree& available_drivers = pt_xrt.get_child("drivers", empty_ptree);
+  for(auto& drv : available_drivers) {
+    const boost::property_tree::ptree& driver = drv.second;
+    std::string drv_name = driver.get<std::string>("name", "N/A");
```
rozumx, line 255: Hmm...this causes the error message to be rather crypted by stating the version is "N/A".  Consider something a little more useful.  Possible "unknown".
```suggestion
    std::string drv_name = driver.get<std::string>("name", "<unknown>");
```

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -241,6 +241,24 @@ str2index(const std::string& str, bool _inUserDomain)
...
+    std::string drv_version = driver.get<std::string>("version", "N/A");
```
rozumx, line 256:
```suggestion
    std::string drv_version = driver.get<std::string>("version", "<unknown>");
```

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -241,6 +241,24 @@ str2index(const std::string& str, bool _inUserDomain)
...
+    if (drv_name.compare("xocl") == 0 && drv_version.compare("unknown") != 0 && xrt_version.compare(drv_version) != 0)
+      throw std::runtime_error(boost::str(boost::format("Unexpected XRT Driver version (%s) was found. Expected %s, to match XRT tools.") % drv_version % xrt_version));
```
rozumx, line 258: Breaking this line into two lines would help it be more readable.  For example:
```
const auto & errMsg = boost::format("Unexpected XRT Driver version (%s) was found. Expected %s, to match XRT tools.") % drv_version % xrt_version;
throw std::runtime_error(errMsg.str());
```

rozumx (changes requested):

chvamshi-xilinx:
This looks good now, Can you please rebase your branch so that build gets triggered.

dayeh-xilinx:
retest this please.

chvamshi-xilinx:
Merging as all comments are addressed.

PR #6731, opened 2022-05-25
Use hwctx when xrt::kernel implementation opens compute units

stsoe:
This is a continuation of changes in #6727.  

Use of xrt::hw_context will move all the way to xclOpenContextByName, but changes are done in
small incremental steps.

This PR also removes the virtual CU context from the xrt::kernel
constructor.  Opening a virtual CU context makes no sense when the
kernel constructor anyway opens regular CU context or fails if no
regular CU can be opened.

The virtual CU context creation was removed from xrt::kernel constructor.  It appeared to not 
be needed and commit logs are clear why it was originally added few years back.

Regression HW tests


vboggara-xilinx (approved):
Changes are looks good to me.

Issue #6640, opened 2022-04-27
XRT failing to compile on Ubuntu 22.04


keryell:
Besides mild problem https://github.com/Xilinx/XRT/issues/6098 there are some failures due to warnings as errors
```
[...]
/home/rkeryell/Xilinx/Projects/Vitis/XRT/src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.cpp: In member function ‘int AzureDev::Sha256AndSplit(const string&, std::vector<std::__cxx11::basic_string<char> >&, std::string&)’:
/home/rkeryell/Xilinx/Projects/Vitis/XRT/src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.cpp:580:21: error: ‘int SHA256_Init(SHA256_CTX*)’ is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
  580 |     if (!SHA256_Init(&context)) {
      |          ~~~~~~~~~~~^~~~~~~~~~
In file included from /home/rkeryell/Xilinx/Projects/Vitis/XRT/src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.cpp:26:
/usr/include/openssl/sha.h:73:27: note: declared here
   73 | OSSL_DEPRECATEDIN_3_0 int SHA256_Init(SHA256_CTX *c);
      |                           ^~~~~~~~~~~
/home/rkeryell/Xilinx/Projects/Vitis/XRT/src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.cpp:590:26: error: ‘int SHA256_Update(SHA256_CTX*, const void*, size_t)’ is deprecated: Since OpenSSL 3.0 [-Werror=deprecated-declarations]
  590 |         if(!SHA256_Update(&context, segment.c_str(), segment.size()))
      |             ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/rkeryell/Xilinx/Projects/Vitis/XRT/src/runtime_src/core/pcie/tools/cloud-daemon/azure/azure.cpp:26:
/usr/include/openssl/sha.h:74:27: note: declared here
   74 | OSSL_DEPRECATEDIN_3_0 int SHA256_Update(SHA256_CTX *c,
      |                           ^~~~~~~~~~~~~
[...]

PR #6737, opened 2022-05-30
Issue #6640 Option to disable warnings as error 

stsoe:
Provide build.sh option `-disable-werror` to disable warnings
as errors.

Issue #6640 

Not willing to disable `-Werror` by default, since I can't seem to
find a good way to have it enabled for all internal builds including
all developers.

Fixes https://github.com/Xilinx/XRT/issues/6640, maybe not exactly per comments, but close enough.


keryell (commented):
Thanks!

PR #6741, opened 2022-05-31
fix CR-1130364

mamin506:
Some internal users want to access some CU register in a shared context. They still want to use KDS/ERT. But XRT xclRegRead() doesn't allow read with a shared context.

Not a bug.

We don't want to change the old behavior too much. So, when CU is opened in an exclusive context, everything will be the same.

What's changed:
1. When open a CU in a shared context, call xclIPSetReadRange() to set a read only range on the CU. This will allow xclRegRead() to read this range even KDS/ERT is still running.
2. The new API is added in deprecated/xrt.h, it will be move to the internal header file when the internal user move to XRT APIs.

Low

u50/vck5000 xbutil validate, local xclRegRead() test cases.

No

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
@@ -432,6 +432,8 @@ static const struct drm_ioctl_desc xocl_ioctls[] = {
 			  DRM_AUTH|DRM_UNLOCKED|DRM_RENDER_ALLOW),
 	DRM_IOCTL_DEF_DRV(XOCL_FREE_CMA, xocl_free_cma_ioctl,
 			  DRM_AUTH|DRM_UNLOCKED|DRM_RENDER_ALLOW),
+	DRM_IOCTL_DEF_DRV(XOCL_SET_READ_RANGE, xocl_set_range_ioctl,
+			  DRM_AUTH|DRM_UNLOCKED|DRM_RENDER_ALLOW),
 
 /* LINUX KERNEL-SPACE IOCTLS - The following entries are meant to be
  * accessible only from Linux Kernel and need be grouped to at the end
--- a/src/runtime_src/core/pcie/linux/device_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/device_linux.cpp
@@ -677,6 +677,58 @@ struct accel_deadlock_status
   }
 };
 
+struct cu_size
+{
+  using result_type = query::cu_size::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type key, const boost::any& cu_idx)
+  {
+    const auto index = boost::any_cast<uint32_t>(cu_idx);
+    const auto cu_subdev = "CU[" + std::to_string(index) + "]";
+    auto pdev = get_pcidev(device);
+    std::string errmsg;
+    result_type size;
+
+    pdev->sysfs_get(cu_subdev, "size", errmsg, size, static_cast<uint32_t>(0));
+    if (!errmsg.empty())
+      throw xrt_core::query::sysfs_error(errmsg);
+
+    return size;
+  }
+};
+
+struct cu_read_range
+{
+  using result_type = query::cu_read_range::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type key, const boost::any& cu_idx)
+  {
+    using tokenizer = boost::tokenizer< boost::char_separator<char> >;
+    const auto index = boost::any_cast<uint32_t>(cu_idx);
+    const auto cu_subdev = "CU[" + std::to_string(index) + "]";
+    auto pdev = get_pcidev(device);
+    std::string errmsg;
+    std::string range_str;
+    result_type range = {0, 0};
+
+    // The read_range is printing in formatted string
+    // Format: "%d %d"
+    pdev->sysfs_get(cu_subdev, "read_range", errmsg, range_str);
+    if (!errmsg.empty())
+      throw xrt_core::query::sysfs_error(errmsg);
+
+    tokenizer tokens(range_str);
+    const int radix = 16;
+    tokenizer::iterator tok_it = tokens.begin();
+    range.start = std::stoul(std::string(*tok_it++), nullptr, radix);
+    range.end = std::stoul(std::string(*tok_it++), nullptr, radix);
+
+    return range;
+  }
+};
+
 // Specialize for other value types.
 template <typename ValueType>
 struct sysfs_fcn
@@ -1080,6 +1132,9 @@ initialize_query_table()
   emplace_sysfs_get<query::hwmon_sdm_mac_addr1>                ("hwmon_sdm", "mac_addr1");
   emplace_sysfs_get<query::hwmon_sdm_fan_presence>             ("hwmon_sdm", "fan_presence");
   emplace_sysfs_get<query::hwmon_sdm_revision>                 ("hwmon_sdm", "revision");
+
+  emplace_func4_request<query::cu_size,                        cu_size>();
+  emplace_func4_request<query::cu_read_range,                  cu_read_range>();
 }
 
 struct X { X() { initialize_query_table(); }};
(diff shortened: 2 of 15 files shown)


src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -677,6 +677,58 @@ struct accel_deadlock_status
...
+
+    tokenizer tokens(range_str);
+    const int radix = 16;
+    tokenizer::iterator tok_it = tokens.begin();
+    range.start = std::stoul(std::string(*tok_it++), nullptr, radix);
+    range.end = std::stoul(std::string(*tok_it++), nullptr, radix);
+
+    return range;
+  }
+};
```
stsoe, line 730: This doesn't feel right.  The parsing should be at coreutil level in query_request.
This file should not need changes beyond adding (emplacing) the two new sysfs query requests.
You may want to check 
core/common/query.h:85 (get(..., modifier, const std::string& modifier);
core/common/query_requests.h:2410 (for a request that uses subdev modifier)
The idea would then be to simply call `get(...,modifier::subdev, subdev)` where subdev is for example `CU[0]` or what have you.  The query_request structure would have a parse function in cu_read_range case.

I didn't exactly work through this, but what you have here doesn't scale very well beyond Alveo Linux.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -656,8 +656,8 @@ shim::~shim()
     dev_fini();
 
     for (auto p : mCuMaps) {
```
stsoe, line 658: Given the changes to the changes to mCuMaps, I think this should be 
```suggestion
    for (const auto& p : mCuMaps) {
```

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1810,10 +1810,10 @@ int shim::xclCloseContext(const uuid_t xclbinId, unsigned int ipIndex)
 
     if (ipIndex < mCuMaps.size()) {
 	    // Make sure no MMIO register space access when CU is released.
-	    uint32_t *p = mCuMaps[ipIndex].first;
+	    uint32_t *p = mCuMaps[ipIndex].addr;
 	    if (p) {
```
stsoe, line 1814:
```suggestion
	    if (auto p = mCuMaps[ipIndex].addr) {
```

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1810,10 +1810,10 @@ int shim::xclCloseContext(const uuid_t xclbinId, unsigned int ipIndex)
...
-		(void) munmap(p, mCuMaps[ipIndex].second);
-		mCuMaps[ipIndex] = std::make_pair<uint32_t*, uint32_t>(nullptr, 0);
+		(void) munmap(p, mCuMaps[ipIndex].size);
```
stsoe, line 1815: It it safe to ignore non-zero return value?
If yes, then let's now use 
`std::ignore = munmap(...)`

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -2032,44 +2032,68 @@ int shim::xclRegRW(bool rd, uint32_t ipIndex, uint32_t offset, uint32_t *datap)
...
+  auto& cumap = mCuMaps[ipIndex];  // {base, size, start, end}
 
-  if (cumap.first == nullptr) {
-    auto cu_subdev = "CU[" + std::to_string(ipIndex) + "]";
-    uint32_t size = 0;
-    std::string errmsg;
-    mDev->sysfs_get<uint32_t>(cu_subdev, "size", errmsg, size, 0);
-    if (size <= 0) {
+  if (cumap.addr == nullptr) {
+    auto size = xrt_core::device_query<xrt_core::query::cu_size>(mCoreDevice, ipIndex);
```
stsoe, line 2038: This is where you should use the modifier per earlier comment:
```
namespace xq = xrt_core::query;
auto size = xrt_core::device_query<...>(mCoreDevice, xq::modifier::subdev, "CU[" + std::to_string(ipIndex) + "]");
```

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -2032,44 +2032,68 @@ int shim::xclRegRW(bool rd, uint32_t ipIndex, uint32_t offset, uint32_t *datap)
...
+    if (size <= 0 || size > 0x10000) {
       xrt_logmsg(XRT_ERROR, "%s: incorrect cu size %d", __func__, size);
       return -EINVAL;
     }
+    auto range = xrt_core::device_query<xrt_core::query::cu_read_range>(mCoreDevice, ipIndex);
```
stsoe, line 2043: ditto, something like:
```
auto raw = ... (as above) ...;
auto range = xq::cu_read_range::to_range(raw);
```

src/runtime_src/core/pcie/linux/device_linux.cpp
mamin506: Let me double check. Thanks for the pointer!

src/runtime_src/core/pcie/linux/shim.cpp
mamin506: Good eye. Fixed.

src/runtime_src/core/pcie/linux/shim.cpp
mamin506: It was ignore from day one. This is safe to ignore. Will use std::ignore

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -2032,44 +2032,68 @@ int shim::xclRegRW(bool rd, uint32_t ipIndex, uint32_t offset, uint32_t *datap)
...
+
+  if (offset >= cumap.size) {
     xrt_logmsg(XRT_ERROR, "%s: invalid CU offset: %d", __func__, offset);
     return -EINVAL;
   }
 
+  if ((cumap.start != 0xFFFFFFFF) && ((cumap.start > offset) || (cumap.end < offset))) {
+    xrt_logmsg(XRT_ERROR, "%s: CU offset %d out of read range, %d, %d", __func__, offset, cumap.start, cumap.end);
+    return -EINVAL;
+  }
```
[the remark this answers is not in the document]
mamin506: Sure, will merge.

dayeh-xilinx:
retest this please.

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
```diff
@@ -432,6 +432,8 @@ static const struct drm_ioctl_desc xocl_ioctls[] = {
 			  DRM_AUTH|DRM_UNLOCKED|DRM_RENDER_ALLOW),
 	DRM_IOCTL_DEF_DRV(XOCL_FREE_CMA, xocl_free_cma_ioctl,
 			  DRM_AUTH|DRM_UNLOCKED|DRM_RENDER_ALLOW),
+	DRM_IOCTL_DEF_DRV(XOCL_SET_READ_RANGE, xocl_set_range_ioctl,
```
maxzhen, line 435: The name here is too generic. It might be better to say "set_cu_readonly_range"?

dayeh-xilinx:
retest this please.

dayeh-xilinx:
retest this please.

src/runtime_src/core/common/query_requests.cpp
```diff
@@ -160,4 +161,20 @@ to_map(const result_type& value)
...
 
+xrt_core::query::cu_read_range::range_data
+xrt_core::query::cu_read_range::
+to_range(const std::string& range_str)
+{
+  using tokenizer = boost::tokenizer< boost::char_separator<char> >;
+  xrt_core::query::cu_read_range::range_data range = {0, 0};
+
+  tokenizer tokens(range_str);
+  const int radix = 16;
```
rozumx, line 172:
```suggestion
  constexpr int radix = 16;
```

rozumx (approved):

PR #6748, opened 2022-06-03
Remove warning messages from AIE_profile/trace_settings processing

IshitaGhosh:
Warnings for missing configs related to runtime metrics under AIE_profile/trace_settings are not required now. As the new support is incomplete and existing tests are using old configs under Debug, the warnings from new settings are not expected now.


jvillarre (approved):
Approved as part of ongoing work to integrate several features.  This portion is undergoing specific testing and the warnings will be re-integrated along with additional features soon.

PR #6749, opened 2022-06-06
Deprecate profiling header files in core/include

jvillarre:
The header files xclperf.h, xcl_app_debug.h, and xcl_axi_checkers.h in core/include were being exposed to the user in the distribution directory, but were from the original SDAccel iteration of the tools and are full of definitions and declarations that are not used and not necessary.  Additionally, the file xcl_perfmon_parameters.h had a lot of unused definitions and was organized in a confusing manner.

This pull request removes xcl_perfmon_parameters.h and moves the three released headers to the deprecated directory.  All of the information still used by the different parts of the code base has been reorganized and moved into the newly introduced core/include/xdp directory.  All references to the original constructs have been replace with references to the new equivalent constructs in the core/include/xdp directory.

Additionally, there was a single function (xclDebugReadIPStatus) in xrt.h that was exposed to the user but should not be called from the user's code, so it has been moved to deprecated/xrt.h.  The function itself is still used in application debug (when reading our IP from gdb commands) but it is linked dynamically via dlsym and doesn't include the header.  After the deprecated function has been removed, I will replace the implementation.

Originally, I had intended to completely separate out the deprecated files from our internal code base.  Unfortunately, because I had to keep the deprecated function xclDebugReadIPStatus visible to the user in deprecated/xrt.h, I had to include the deprecated declaration file deprecated/xcl_app_debug.h in deprecated/xrt.h.  This in turn also included deprecated/xclperf.h, and since xrt.h (which is used many places in our internal code) also includes deprecated/xrt.h, all of the old contents of the deprecated files were reintroduced.

In order to keep the deprecated function declaration visible to users for a release, I had to completely rename all constructs and constants in all of our code base.  All of the internal code now should be using the new constructs defined in the core/include/xdp library instead.

Low risk that some replacement from old construct to new construct was incorrect, which would result in incorrect profiling, debug, and trace results.  Profiling and AXI checker IP access or display via xbutil and the XDP library needs to be verified in all situations.

Application debug and basic PL device trace functionality has been tested on Alveo hardware.  Additional testing will be covered by internal regressions but the daily results should be monitored for a while to ensure all emulation and hardware flows on all types of boards remain correct.

We will have to mention in the release notes that the function "xclDebugReadIPStatus" and the header files xclperf.h, xcl_app_debug.h, and xcl_axi_checker_codes.h have been deprecated and will be removed in a future release.


hackwa (approved):
looks good!

chvamshi-xilinx (approved):

pgschuey (approved):
looks good

IshitaGhosh (approved):
Changes look good.

venkatp-xilinx (approved):

PR #6750, opened 2022-06-06
CR-1131656 Soft lockup is observed when running XRT in polling mode

chienwei-lan:
In the ERT polling mode, it's possible for the scheduler to keep running without yielding the CPU.

After 20s without yielding, we may see the soft lockup warning.

To avoid the warning, the scheduler will voluntarily yield after a few rounds of polling.
 
For more details, please refer to CR-1131656


saifuddin-xilinx (approved):
The issue is resolved with this changes. Please cherry pick the same with 2021.2 branch.

maxzhen (approved):

PR #6756, opened 2022-06-08
CR-1125485 keep reference count in driver to allow multiple context

Zenfendson:
CR-1125485 keep reference count in driver to allow multiple context

This is not a bug 

Add per client cu and scu refcnt

Low

Run testcase in CR-1125485 on U50

No

### DIFF
--- a/src/runtime_src/core/common/drv/kds_core.c
+++ b/src/runtime_src/core/common/drv/kds_core.c
@@ -334,26 +334,6 @@ kds_test_and_refcnt_decr(struct kds_client *client, int domain, int cu_idx)
 }
 
 static int
-kds_client_get_cu_refcnt(struct kds_client *client, int domain, int idx)
-{
-	int ret = 0;
-	u32 *refs = NULL;
-
-	refs = kds_client_domain_refcnt(client, domain);
-	if (!refs) {
-		return -EINVAL;
-	}
-	ret = MAX_CUS;
-	mutex_lock(&client->refcnt->lock);
-	if (idx > -1 && idx < MAX_CUS)
-		ret = refs[idx];
-	else
-		kds_err(client, "Client context cu index out of range");
-	mutex_unlock(&client->refcnt->lock);
-	return ret;
-}
-
-static int
 kds_client_set_cu_refs_zero(struct kds_client *client, int domain)
 {
 	u32 *dst = NULL;
@@ -1093,7 +1073,6 @@ _kds_fini_client(struct kds_sched *kds, struct kds_client *client,
 		 struct kds_client_ctx *cctx)
 {
 	struct kds_ctx_info info;
-	u32 count;
 	u32 count_idx;
 
 	kds_info(client, "Client pid(%d) has %d opening context for %d slot",
@@ -1117,7 +1096,6 @@ _kds_fini_client(struct kds_sched *kds, struct kds_client *client,
 			kds_del_context(kds, client, &info);
 		}
 		kds_info(client,"Removing CU Domain[%d] CU Index [%d]",info.cu_domain,info.cu_idx);
-		count = kds_client_get_cu_refcnt(client, DOMAIN_PS, count_idx);
 		count_idx += 1;
 	};
 	kds_client_set_cu_refs_zero(client, DOMAIN_PS);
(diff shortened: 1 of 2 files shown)


stsoe:
add to whitelist

stsoe:
ok to test

mamin506 (approved):
LGTM

mamin506:
@Zenfendson Can you compare the result of "xbutil validate -d <bdf> -u iops" w/ and w/o your change?

src/runtime_src/core/common/drv/kds_core.c
```diff
@@ -255,13 +255,35 @@ kds_scu_config(struct kds_cu_mgmt *scu_mgmt, struct kds_command *xcmd)
...
+static bool
+kds_test_refcnt(struct kds_client *client, int domain, u32 idx)
+{
+	u32 *refs;
+	bool is_set = false;
+
+	if (domain == DOMAIN_PL)
+		refs = client->refcnt->cu_refs;
+	else
+		refs = client->refcnt->scu_refs;
```
maxzhen, line 272: The code here seems to assume that there can only be two type of domains. I'd think that it'll be better to explicitly check against domains you can handle here and return / print errors, if you see unexpected domains. Next time when we extend domain types, we'll not silently corrupt scu array.

src/runtime_src/core/common/drv/kds_core.c
```diff
@@ -477,6 +495,42 @@ kds_submit_ert(struct kds_sched *kds, struct kds_command *xcmd)
 	return 0;
 }
 
+static int
+kds_test_and_refcnt_incr(struct kds_client *client, int domain, int cu_idx)
+{
+	int prev;
+	u32 *refs;
+
+	if (domain == DOMAIN_PL || domain == DOMAIN_VIRT)
```
maxzhen, line 504: do we have a specific domain for virtual cu? if we do, we map cu_refs to it. But in line 271, we map scu_refs to it? Maybe we should have a function to map cu_/scu_refs to domain and be consistent with our design?

src/runtime_src/core/common/drv/kds_core.c
```diff
@@ -934,9 +978,14 @@ int kds_init_client(struct kds_sched *kds, struct kds_client *client)
 	if (!client->stats)
 		return -ENOMEM;
 
+	client->refcnt = kzalloc(sizeof(struct kds_client_cu_refcnt), GFP_KERNEL);
+	if (!client->refcnt) {
+		return -ENOMEM;
```
maxzhen, line 983: leaking client->stats?

maxzhen (changes requested):

Zenfendson:
@mamin506 after running 10 times of the command, the median iops with my driver is 538170 and the median iops without my driver is 541908, which have no significantly impact on iops.

salindac:
retest this please

maxzhen (approved):

jeffli-xilinx (approved):

rozumx (approved):

Zenfendson:
> Looks good. I can't tell if maybe already covered. But it would be helpful to a logged message when the reference count is unexpected different from 0 at process exit.

Well, when the process exits, the reference count per client will be eliminated and set to 0. The memory will then be released so I think such exception will not happen.

stsoe:
> > Looks good. I can't tell if maybe already covered. But it would be helpful to a logged message when the reference count is unexpected different from 0 at process exit.
> 
> Well, when the process exits, the reference count per client will be eliminated and set to 0. The memory will then be released so I think such exception will not happen.

My point was that if you find yourself having to reset to 0, then something is wrong in user land.  Of course in abnormal termination you need to do this reset, but in the normal case, user space should always release a CU context it has acquire.  That's why a warning to this effect could be useful.

mamin506:
> 

Hi @stsoe , if user space doesn't release a CU context it has acquired. There is a warning in dmesg, for example:
  " Client pid(1001) has 2 opening context for 0 slot"
  
Is this what you want? Or do you suggest to print more details, like
" Client pid(1001) has 2 opening context for 0 slot
      CU(0) refcnt is 2
      CU(1) refcnt is 1
"

PR #6759, opened 2022-06-10
hwmon_sdm shall requests SDR data from VMR for dynamic fields

rajkumar-xilinx:

hwmon_sdm driver requests SDR data for all sensor fields during driver loading time. In next sensor readings, It should not request for the same data for static fields of SDR. Only sensor's value, max, average and status fields values gets changed and all other fields are static. So, add changes in hwmon_sdm driver to fix this case.
hwmon_sdm shall requests SDR data from VMR for sensor's dynamic fields and should skip for other fields.
Added changes in hwmon_sdm driver to handle this case when user tries to read the sensors.
NA
xbutil examine reports, hwmon_sdm & hwmon sysfs access.
NA


rozumx (approved):

rajkumar-xilinx:
Hi @maxzhen please review this PR and let me know, thank you.

maxzhen (approved):

PR #6763, opened 2022-06-13
CR-1128916 Download xclbin failed for pre-emptible kernel

rbramand-xilinx:
While downling xclbin in pre-emptible kernel copy_from_user fails to copy buffers with size greater than 4K bytes, fixed it by copying 4K chunks at a time when preemption is enabled

Fixed copying xclbin sections from user pointer failures 

copying 4K chunk buffers at a time solved this issue

none as the change is added under if preemption condition

tested yolov4 application with preemption enabled and the test passes

NA

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_xclbin.c
@@ -285,6 +285,7 @@ zocl_read_sect(enum axlf_section_kind kind, void *sect,
 	uint64_t offset;
 	uint64_t size;
 	void **sect_tmp = (void *)sect;
+	uint64_t copy_size, copy_offset, chunk_size;
 	int err = 0;
 
 	err = xrt_xclbin_section_info(axlf_full, kind, &offset, &size);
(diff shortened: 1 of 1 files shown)


maxzhen:
This is just a workaround till we find the real root cause, right?

mamin506:
Hi @rbramand-xilinx , I understand what you are doing. Can you explain why copy_from_user() doesn't work with CONFIG_PREEMPT when size is large than 4K? I didn't get any helpful information from internet and Linux source code. Please share your knowledge.

rbramand-xilinx:
Hi @mamin506 , @maxzhen, @houlz0507 
copy_from_user inside zocl_read_sect function fails for all sections that have size greater than 4K(PAGE_SIZE) bytes when **Preemption** is enabled (AIE_RESOURCES and AIE_METADATA in this case). Also I came across some google pages where they are talking about copy_from_user and 4k (page size) buffers eg: https://www.spinics.net/lists/newbies/msg00058.html . So I made some experiments like copying 4k chunks at a time and copy_from_user passes so I made this change. There are not many articles about copy_from_user behavior when preemption is enabled. 
Also user application uses threads and reading xclbin sections fails only when xclbin is programmed from application not from command line. Please let me know your thoughts on this.

larry9523:
> 



> Hi @mamin506 , @maxzhen, @houlz0507 copy_from_user inside zocl_read_sect function fails for all sections that have size greater than 4K(PAGE_SIZE) bytes when **Preemption** is enabled (AIE_RESOURCES and AIE_METADATA in this case). Also I came across some google pages where they are talking about copy_from_user and 4k (page size) buffers eg: https://www.spinics.net/lists/newbies/msg00058.html . So I made some experiments like copying 4k chunks at a time and copy_from_user passes so I made this change. There are not many articles about copy_from_user behavior when preemption is enabled. Also user application uses threads and reading xclbin sections fails only when xclbin is programmed from application not from command line. Please let me know your thoughts on this.

Can you please elaborate "copy_from_user inside zocl_read_sect function fails for all sections that have size greater than 4K(PAGE_SIZE) bytes"? What failure are we talking about? I understand that if the user buffer crosses 4K boundary, copy_from_user may sleep. But it does not necessarily means we can not pass user buffer that cross page boundary. Just make sure we are not under atomic context.

rbramand-xilinx:
> > 
> 
> > Hi @mamin506 , @maxzhen, @houlz0507 copy_from_user inside zocl_read_sect function fails for all sections that have size greater than 4K(PAGE_SIZE) bytes when **Preemption** is enabled (AIE_RESOURCES and AIE_METADATA in this case). Also I came across some google pages where they are talking about copy_from_user and 4k (page size) buffers eg: https://www.spinics.net/lists/newbies/msg00058.html . So I made some experiments like copying 4k chunks at a time and copy_from_user passes so I made this change. There are not many articles about copy_from_user behavior when preemption is enabled. Also user application uses threads and reading xclbin sections fails only when xclbin is programmed from application not from command line. Please let me know your thoughts on this.
> 
> Can you please elaborate "copy_from_user inside zocl_read_sect function fails for all sections that have size greater than 4K(PAGE_SIZE) bytes"? What failure are we talking about? I understand that if the user buffer crosses 4K boundary, copy_from_user may sleep. But it does not necessarily means we can not pass user buffer that cross page boundary. Just make sure we are not under atomic context.

copy_from_user is returning number of bytes it is failing to copy and this happens only when preemption is enabled. It is failing only for those sections whose size is greater than 4KB, so I have added to copy 4K chunks at a time.
Also we are not under atomic context while doing the copying part.

rbramand-xilinx:
sure @larry9523 I will debug further on why it fails and get back.

maxzhen:
Marking this PR as do-not-merge until it is root caused.

chvamshi-xilinx:
Please raise a PR after figuring out reason behind this issue. Closing this PR

PR #6764, opened 2022-06-13
Doesn't copy entire extra cu mask

mamin506:
When extra_cu_mask is not zero, KDS doesn't copy all CU mask to xcmd.

@pranjalv-xilinx found this issue.

Just use correct size.

This will impact CU when index exceed 35.

U50 xbutil validate

No


dayeh-xilinx:
retest this please.

PR #6765, opened 2022-06-13
xclbinutil - Added initial AIE Partition support

rozumx:
Work Done
---------
+ Create a new xclbin AIE Partition section that includes dynamic arrays and multi-level nodes
+ Created unit tests for the new AIE Partition section
+ Removed previous code that would extract AIE Partition information from the AIE_METADATA section.
+ Updated XUtil::as_vector_simple to be more resilient with missing property child nodes
+ Refactored structure initializers from "{0}" to "{}".
+ Bug Fix: Section file name value used to indicate which file is being read or written to was sometimes wasn't being set.
+ Remove unnecessary creation of Section objects when adding / removing a section by name.
+ Minor cleanup of some code formatting
+ Clean up some error messages

#### Problem solved by the commit
New functionality

#### Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered
n/a
#### How problem was solved, alternative solutions (if any) and why they were rejected
n/a
#### Risks (if any) associated the changes in the commit
Low risk for this is new functionality.  In addition, unit tests were added as well to detect potential future code change issues.

#### What has been tested and how, request additional testing if necessary
Unit tests added.

#### Documentation impact (if any)
n/a

PR #6770, opened 2022-06-16
Add XCLBIN parser of AIE_PARTITION with CDO groups

larry9523:
Extend AIE_PARTITION parser after https://github.com/Xilinx/XRT/pull/6765

Extend aie_partition_obj to include PDI blobs and CDO groups' metadata

Tested on hw_emu with a new XCLBIN that AIE Partition metadata along with PDIs and CDO groups' metadata can be correctly retrieved.

### DIFF
--- a/src/runtime_src/core/common/xclbin_parser.h
+++ b/src/runtime_src/core/common/xclbin_parser.h
@@ -19,6 +19,7 @@
 #include "config.h"
 #include "cuidx_type.h"
 #include "core/include/xclbin.h"
+#include "core/include/xrt/xrt_uuid.h"
 
 #include <array>
 #include <limits>
@@ -116,7 +117,7 @@ struct aie_cdo_group_obj
 // @pdi: PDI blob
 struct aie_pdi_obj
 {
-  xuid_t uuid;
+  xrt::uuid uuid;
   std::vector<aie_cdo_group_obj> cdo_groups;
   std::vector<uint8_t> pdi;
 };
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/common/xclbin_parser.h
```diff
@@ -95,16 +95,44 @@ struct softkernel_object
...
+};
+
+// struct aie_pdi_obj - wrapper for an AIE PDI object
+//
+// @uuid: PDI UUID
+// @cdo_groups: Array of CDO groups
+// @pdi: PDI blob
+struct aie_pdi_obj
+{
+  xuid_t uuid;
```
stsoe, line 119: I wonder who the consumer is, but given that this is C++ code I think it should be xrt::uuid instead?
xrt::uuid can be constructed from an xuid_t and is of course doing the same copy you are explicitly doing.

stsoe (changes requested):
Looks good.  One suggestion.

dayeh-xilinx:
retest this please. boardfarm host e344 has power issue.

src/runtime_src/core/common/xclbin_parser.h
larry9523: The consumer will be user space c++ code, e.g. hw_emu, windows user space code. I updated it use xrt::uuid. Thanks.

stsoe (approved):

pranjalv-xilinx (approved):

dayeh-xilinx:

PR #6773, opened 2022-06-16
fixing sysroot generation failure for vck5000

chvamshi-xilinx:
sysroot generation was failing in vck5000

NA

We should either enable openssh or dropbear. Sysroot generation fails if we enable both openssh and dropbear.

None

Generated sysroot and verified

NA


dayeh-xilinx:

PR #6774, opened 2022-06-16
CR-1130285 PIPELINE faile about XGQ  command version 0.1

mamin506:
1. remove debug code and disable cid in ERT fw
2. don't use write combine mapping

CR-1130285

So far, this is only reproducible in PIPELINE. But we don't see it happen recently.
We suspect this is related to write combine mapping of CQ. Let's see if that error still happen.

No

u50 validate test. After disable write combine, the iops test result is the same.

No


maxzhen (approved):

dayeh-xilinx:

PR #6784, opened 2022-06-20
VITIS-5847 Resilient VMR - put apu firmware download in a worker

xdavidz:

We block in xgq probe till apu firmware download is finished. This will increase drive attach time when there are multiple cards because the driver attach is not in parallel.

We can put the apu firmware download in a worker. And only the worker finished, we notify the peer (xocl) to mark itself ready.

This will also compatible with legacy ALEVO card, since there is not xgq, simple just ignore the -ENODEV.

N/A
N/A
Test if ALEVO cards have regression with this change.

Tested on 2 cards config.

with the fix: real    0m32.965s
without the fix: real    2m14.281s


xdavidz:
hold this PR until it is necessary to improve. There is not direct complain about driver loading slow, instead user complained XRT installation on mult-card is slower than single card. https://jira.xilinx.com/browse/CR-1122975. We can keep an eye when it is necessary to fix. Close it for now.

PR #6800, opened 2022-06-22
Add support for latest Amazon Linux(AL2) Image 

rbramand-xilinx:
1. Fix kernel module compilation on latest AL2.
2. Change package name to include amzn. Original package has just _2 in package name eg: xrt_202220.2.14.161_2-x86_64-xrt.rpm which makes it hard to recognize as amazon package. With changes in this pr package name will be changed to eg: xrt_202220.2.14.0_amzn2-x86_64-xrt.rpm
3. Make xrtdeps run without user intervention.

1. Getting module path and exporting the variable to so that module path will not be altered in subshells of make and include paths will be proper
2. Used /etc/os-release to get linux flavor name and added it to package name. This is done only for amazon linux OS, rest of Oses can easily be distinguished by their current names.  

none

Tested package creation and installation on AL2 vms and it succeeded. 

We need to document the change in package name for Amazon Linux OS.


chvamshi-xilinx (approved):

maxzhen (approved):

src/CMake/cpackLin.cmake
```diff
@@ -155,7 +155,16 @@ else ()
   SET (CPACK_GENERATOR "TGZ")
 endif()
 
-SET(CPACK_PACKAGE_FILE_NAME "${CPACK_PACKAGE_NAME}_${XRT_VERSION_RELEASE}.${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}_${CPACK_REL_VER}-${CPACK_ARCH}")
+if (${LINUX_FLAVOR} MATCHES "^Amazon")
+  execute_process(
+    COMMAND bash "-c" "cat /etc/os-release | grep '^ID=' | awk -F= '{print $2}' | tr -d '\"'"
```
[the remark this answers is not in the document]
rbramand-xilinx: @keryell, I have added changes as per comments

dayeh-xilinx:
retest this please.

stsoe (approved):
LGTM

src/CMake/cpackLin.cmake
```diff
@@ -155,7 +155,20 @@ else ()
...
 endif()
 
-SET(CPACK_PACKAGE_FILE_NAME "${CPACK_PACKAGE_NAME}_${XRT_VERSION_RELEASE}.${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}_${CPACK_REL_VER}-${CPACK_ARCH}")
+# On Amazon Linux CPACK_REL_VER is just '2' and it is hard to
+# distinguish it as AL package, so adding CPACK_FLOVOR (eg: amzn) 
+# to package name
+if (${LINUX_FLAVOR} MATCHES "^Amazon")
+  execute_process(
+    COMMAND awk -F= "$1==\"ID\" {print $2}" /etc/os-release
+    COMMAND tr -d "\""
```
keryell, line 164: Since you are using the AWK programming language, you should not need `tr` too, for example by using the `sub()` function.
But this is not fundamental.

keryell (approved):

src/CMake/cpackLin.cmake
rbramand-xilinx: Hi @keryell, in both cases o/p is different -> lsb_release --id --short gives **Amazon** as o/p but all the standard packages uses **amzn** in their versions, so I used o/p from file '/etc/os-release'. 

eg: yum info git
Name        : git
Arch        : x86_64
Version     : 2.32.0
Release     : **1.amzn2.0.1**
Size        : 365 k
Repo        : installed
From repo   : **amzn2-core**
Summary     : Fast Version Control System
URL         : https://git-scm.com/
License     : GPLv2

But as it is just a part of package name we can give any name we want, o/p from lsb_release --id should also do

rbramand-xilinx: We are trying to add OS names to other distribution(ubuntu,centos,rhel) packages as well, while handling them we can decide whether to use info from lsb_release or /etc/os-release file

keryell: But what I do not understand is what is not already in https://cmake.org/cmake/help/latest/command/cmake_host_system_information.html#command:cmake_host_system_information

chvamshi-xilinx: That will add additional requirement of cmake 3.22+. 
We can either use existing approach or we can even use lsb_release --id --short in the package name.

keryell: Do you mean the information you want is not available in an older version of CMake, like 3.10?

chvamshi-xilinx: Yes @keryell  . Distribution information is added in 3.22.

keryell: Ah, too bad!

rbramand-xilinx: Yeah some of the centos and rhel OSes we support still doesn't have this support, they use cmake of 3.17 version

chvamshi-xilinx: @rbramand-xilinx , So, whatever you have implemented is good then. Merging the change..

PR #6802, opened 2022-06-22
Minor update to run summary format

jvillarre:
When profiling or debug features are enabled, a run summary file is produced with pointers to all of the generated data files that contain profiling or debug information.  This run summary file is consumed by tools, like Vitis Analyzer, which then can make choices on how to present the raw information to the user.

In some cases, the user may run an AIE application in XRT and enable AIE trace generation, but the AIE trace files are not listed in the run summary file.  When this happens, the tool that consumes the run summary has no knowledge that the design was an AIE application and may not display the information correctly.

This pull request updates the run summary being generated to a new minor version and adds a generic "flags" element to the JSON.  The only flag currently supported is "aie" which will be added whenever an AIE profiling/trace/debug option is enabled.  This allows the tool that consumes the run summary to determine to display AIE layouts even if no AIE files are directly pointed to by the run summary.  Additional flags may be added in the future to help downstream tools make display decisions.

Each AIE XDP plugin now registers the fact that the application is AIE based.  At the end of the application, when the run_summary is being generated we check this information and add the flag if appropriate.

The generation of the new run summary and consumption by downstream tools has been tested.

### DIFF
--- a/src/runtime_src/xdp/profile/database/static_info_database.cpp
+++ b/src/runtime_src/xdp/profile/database/static_info_database.cpp
@@ -119,6 +119,17 @@ namespace xdp {
     applicationStartTime = t ;
   }
 
+  bool VPStaticDatabase::getAieApplication() const
+  {
+    return aieApplication;
+  }
+
+  void VPStaticDatabase::setAieApplication()
+  {
+    std::lock_guard<std::mutex> lock(summaryLock);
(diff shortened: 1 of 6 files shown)


pgschuey (approved):

IshitaGhosh (approved):
Change looks good

PR #6815, opened 2022-06-28
Revert "CR-1120194: Addressed hang issue for external IO example, specific thread polls the client file descriptor periodically"

chvamshi-xilinx:
Reverts Xilinx/XRT#6769

Reverting as per the request.


keryell:
All this story is complex.
Perhaps using higher-level libraries like https://www.boost.org/doc/libs/1_79_0/doc/html/interprocess.html to re-engineer the next version instead of handling every thing low-level by hand?

PR #6820, opened 2022-06-29
Vitis-6101 Support read and write directly from/to device buffer

stsoe:
This PR extends the semantics of xclUnmgdPwrite/read into the native
xrt::bo APIs.  The xcl APIs are deprecated and should not be used.

To avoid of host memcpy with the usage of xclSynBO(), Vitis AI runtime
(VART) currently leverages xclUnmgdPwrite() and xclUnmgdPread() to
perform PCIe DMA data transfer between host and device so as to
achieve better performance.

Extend xrt::bo::read() and xrt::bo::write() to write directly to
device buffer if buffer object has not host backing storage.

None, exposes previously unsupported functionality

Test case updated for quick sanity check

Inline doxygen updated.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -263,7 +263,7 @@ public:
     return export_handle;
   }
 
-  void
+  virtual void
   write(const void* src, size_t sz, size_t seek)
   {
     if (sz + seek > size)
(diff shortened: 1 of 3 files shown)


vboggara-xilinx (approved):

maxzhen (approved):
I remembered that Lizhi tested it and found that for non-4k aligned buffer, DMA works, but performance is noticeably slower. Not sure if it worth a warning msg when we see an unaligned buffer pointer passed in...

PR #6831, opened 2022-07-03
hw ctx support for PCI shim and xocl driver 

saifuddin-xilinx:
I have added support for new hw context for pci shim and xocl driver, 
The changes are backward comparability. The existing code is minimal modified to make sure it will not break the existing flow. 
Basic validate is working fine.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/include/xocl_ioctl.h
+++ b/src/runtime_src/core/pcie/driver/linux/include/xocl_ioctl.h
@@ -77,6 +77,8 @@
  * 11   Send an execute job to a compute unit  DRM_IOCTL_XOCL_EXECBUF         drm_xocl_execbuf
  * 12   Register eventfd handle for MSIX       DRM_IOCTL_XOCL_USER_INTR       drm_xocl_user_intr
  *      interrupt
+ * 13   Register a specifi xclbin image with   DRM_IOCTL_XOCL_REGISTER_AXLF   drm_xocl_axlf
+ *      the device
  * 13   Update device view with a specific     DRM_IOCTL_XOCL_READ_AXLF       drm_xocl_axlf
  *      xclbin image
  * 14   Obtain info of bo                      DRM_IOCTL_XOCL_INFO            drm_xocl_info_bo
@@ -152,6 +154,8 @@ enum drm_xocl_ops {
 	DRM_XOCL_EXECBUF,
 	/* Register eventfd for user interrupts */
 	DRM_XOCL_USER_INTR,
+	/* Register xclbin/axlf */
+	DRM_XOCL_REGISTER_AXLF,
 	/* Read xclbin/axlf */
 	DRM_XOCL_READ_AXLF,
 	/* Hot reset request */
@@ -500,6 +504,9 @@ struct drm_xocl_pread_bo {
 enum drm_xocl_ctx_code {
 	XOCL_CTX_OP_ALLOC_CTX = 0,
 	XOCL_CTX_OP_FREE_CTX,
+	XOCL_CTX_OP_ALLOC_HW_CTX,
+	XOCL_CTX_OP_FREE_HW_CTX,
+	XOCL_CTX_OP_OPEN_CU_CTX,
 	XOCL_CTX_OP_OPEN_UCU_FD,
 };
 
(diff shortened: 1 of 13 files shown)


src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1393,6 +1393,123 @@ xclLoadXclBin(const xclBin *buffer)
   return ret;
 }
 
+/*
+ * xclRegisterAxlf()
+ */
+int shim::xclRegisterAxlf(const axlf *buffer)
```
stsoe, line 1399: This looks a lot like xclLoadAxlf()?  Should we refactor and share code?

stsoe (commented):
I am not reviewing xocl change, but the shim changes looks good to me. We will have to do some refactoring or remove the legacy code path if possible.

src/runtime_src/core/pcie/linux/shim.cpp
saifuddin-xilinx: Yes. You are correct. Actually I don't want to disturb the existing code. Hence added extra function for registration. 
Anyway I will cleanup once the changes are stable

src/runtime_src/core/pcie/driver/linux/include/xocl_ioctl.h
```diff
@@ -520,10 +528,11 @@ enum drm_xocl_ctx_code {
 struct drm_xocl_ctx {
 	enum drm_xocl_ctx_code op;
 	xuid_t   xclbin_id;
+	char	 cu_name[64];
```
maxzhen, line 531: This is a little awkward... sometimes, cu_index is suffice, sometimes you need cu_name as well to uniquely identify a CU. The code becomes less intuitive and hard to maintain.

@stsoe @saifuddin-xilinx , the reason we need the cu_name is because that, in a multi-slot case, <uuid, cu_index> may refer to a wrong CU since xclbin may be swapped, right? I'm wondering if this is the time to introduce explicit xclbin-lockdown operation (v.s. implicitly locking down by opening the first ctx)? So, user space code will:
1. lock down xclbin in one particular slot
2. find out cu index through sysfs node
3. still talks to driver using cu index
4. unlock xclbin when everything is done

Of course, driver shall check and make sure xclbin is locked down before servicing any ctx request

Or, we separate data structure for ctx allocation and free. In allocation, we always require xclbin uuid and cu_name. In free, we always need the handle back (no cu_index and xclbin uuid any more).

Let me know what you think?

saifuddin-xilinx: Hi Max, 
Actually for legacy context we need xclbin_id and cu_index for a context. That is already present in this data structure. 
Now for the new hw_context we need slot_handle(slot id) and cu_name for a context. Instead of creating a new data structure I have used existing and added cu_name  filed. 

As we are using exchanging slot id between user space and kernel space, I believe we don't need to use sysfs entry to get cu_index. Using slot_id and cu_name driver can maintain this.

src/runtime_src/core/pcie/linux/shim.h
```diff
@@ -157,6 +158,7 @@ class shim
   register_xclbin(const xrt::xclbin&);
 
 private:
+  bool legacyCtxSupport;
```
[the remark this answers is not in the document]
saifuddin-xilinx: From application user can either use new h_w context or legacy LoadXclbin.
User application should not mix them. 
If application call register then it should use hw_context, which is responsible for load xclbin.
if application call load xclbin then it should use legacy opencontext/closecontext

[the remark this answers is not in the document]
maxzhen: Yes, you're right that user app should not mix them, but what if they do? Instead, we should add the check in the driver so that we can make sure driver won't crash even when user app does something wrong.

chvamshi-xilinx:
Hi @saifuddin-xilinx , As discussed offline,

Both legacy and new applications should enter driver using following same calls
1) RegisterXclbin ioctl -> This caches the xclbin information
2) OpenHwContext ioctl -> This downloads the xclbin and returns a new context

In Legacy Applications, XRT shim takes care of calling these two APIs from xclLoadXclbin
In New Applications, Application has to call these APIs explicitly. 

Open_cu_context is going to be based on hardware context & cu name (Earlier it was based on UUID and cu index)

saifuddin-xilinx:
A new PR has been created for the same 
https://github.com/Xilinx/XRT/pull/6860

Issue #6832, opened 2022-07-03
Update build.rst for C++17


timsnyder-siv:
https://github.com/Xilinx/XRT/blame/master/src/runtime_src/doc/toc/build.rst#L16-L35

I think you guys require C++17 based on the fact that I'm getting messages like this from cmake:
```
CMake Error in /home/centos/XRT/tests/validate/ps_aie_test/CMakeLists.txt:
  Target "ps_aie.exe" requires the language dialect "CXX17" (with compiler
  extensions), but CMake does not know the compile flags to use to enable it.
```

I'm able to get cmake to generate after activating devtoolset-9 but I thought you might want to update your docs to reflect your C++ requirement.

stsoe:
@uday, the documentation here is out-of-date: https://xilinx.github.io/XRT/master/html/build.html
Will you update?

PR #6833, opened 2022-07-04
Automatically handle devtoolset-9

bryanloz-xilinx:
Ease of use enhancement.   
No one has to remember to do `scl enable devtoolset-9 bash`.  

`source scl_source enable devtoolset-9`  

We don't need special install instructions for building on RHEL/CentOS anymore.   
@uday610 @stsoe

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -3,11 +3,25 @@
 set -e
 
 OSDIST=`grep '^ID=' /etc/os-release | awk -F= '{print $2}' | tr -d '"'`
+VERSION=`grep '^VERSION_ID=' /etc/os-release | awk -F= '{print $2}' | tr -d '"'`
+MAJOR=${VERSION%.*}
 BUILDDIR=$(readlink -f $(dirname ${BASH_SOURCE[0]}))
 CORE=`grep -c ^processor /proc/cpuinfo`
 CMAKE=cmake
 CMAKE_MAJOR_VERSION=`cmake --version | head -n 1 | awk '{print $3}' |awk -F. '{print $1}'`
 CPU=`uname -m`
+
+# Documentation states that devtoolset-9 is a requirement for building on RHEL/CentOS 7.x
+if [[ ( $OSDIST == "rhel" || $OSDIST == "centos" ) && $MAJOR -lt "8" ]]; then
+    echo "Setting up devtoolset-9"
+    {
+        # Try
+	source scl_source enable devtoolset-9
+    } || {
+        # Except
+	echo "devtoolset-9 is not installed, please run xrtdeps.sh" && exit 1
+    }
+fi
 
 if [[ $CMAKE_MAJOR_VERSION != 3 ]]; then
     if [[ $OSDIST == "centos" ]] || [[ $OSDIST == "amzn" ]] || [[ $OSDIST == "rhel" ]] || [[ $OSDIST == "fedora" ]]; then
--- a/build/build.sh
+++ b/build/build.sh
@@ -3,23 +3,26 @@
 set -e
 
 OSDIST=`grep '^ID=' /etc/os-release | awk -F= '{print $2}' | tr -d '"'`
-VERSION=`grep '^VERSION_ID=' /etc/os-release | awk -F= '{print $2}' | tr -d '"'`
-MAJOR=${VERSION%.*}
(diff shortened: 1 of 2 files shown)


bryanloz-xilinx:
Latest commit addresses documentation and issue: https://github.com/Xilinx/XRT/issues/6832

build/build.sh
```diff
@@ -3,12 +3,26 @@
...
+if [[ ( $OSDIST == "rhel" || $OSDIST == "centos" ) && $MAJOR -lt "8" ]]; then
+    echo "Setting up devtoolset-9"
+    {
+        # Try
+	source scl_source enable devtoolset-9
+    } || {
+        # Except
+	echo "devtoolset-9 is not installed, please run xrtdeps.sh" && exit 1
+    }
+fi
```
stsoe, line 24: At the very least this should be conditioned on the version of gcc,  it is possible that a user has installed a more modern version of gcc or modified their env to point to a local install of later gcc.   Enforcing devtoolset-9 is not a good idea..

Personally, I believe the user should set the devtoolset themselves if necessary.  Also 7.x is completely legacy.

src/runtime_src/doc/toc/build.rst
```diff
@@ -26,13 +26,8 @@ for the tools and libraries XRT depends on. If any system libraries
...
 
-On RHEL7.x/CentOS7.x use devtoolset to switch to C++14 devlopment
-environment. This step is not applicable to Ubuntu, which already has
-C++14 capable GCC.
-
-::
-
-   scl enable devtoolset-9 bash
+On RHEL7.x/CentOS7.x devtoolset-9 is required to switch to a C++17 devlopment
+environment.
```
stsoe, line 30: We should be more vague, and only suggest devtoolset-9.  C++17 is a requirement, but devtoolset-9 is not.

stsoe (changes requested):

build/build.sh
bryanloz-xilinx: @stsoe Based on your suggestion, I have modified this logic... now devtoolset-9 is not a requirement... but it is a fallback.

This PR is just for being kind to our existing users who work on RHEL/CentOS 7.x. Which is still a lot of people. For instance the FPGA Developer AMI on AWS is still CentOS 7.5.

Once we drop support for this legacy OS, we can just remove this logic.

The GCCVERSION extraction I added could be useful in the future as well.

bryanloz-xilinx: https://aws.amazon.com/marketplace/pp/prodview-gimv3gqbpe57k

src/runtime_src/doc/toc/build.rst
bryanloz-xilinx: updated

build/build.sh
```diff
@@ -3,12 +3,29 @@
...
+    echo "Trying devtoolset-9"
+    {
+        # Try
+	source scl_source enable devtoolset-9
+    } || {
+        # Except
+	echo "Could not find a suitable gcc version" && exit 1
+    }
+fi
+
```
stsoe, line 28: And what if I am compiling with clang?

@bryanloz-xilinx,  I don't think we should do this.  If you can get CMake to error out when the picked up C++ compiler doesn't support c++17, then that would be great.  But I don't think changing the build script is not a good idea.    Developers / and folks who build XRT should just follow the instructions that you nicely changed.   You can add your own checks in your own set up scripts.

stsoe: Sure. Thanks!
We do have at least one external user building with clang but undoubtedly on an obsolete OS.
https://github.com/Xilinx/XRT/pull/6391

stsoe: CXX 17 is already a strict requirement and the error message is quite specific about this.

bryanloz-xilinx: > CXX 17 is already a strict requirement and the error message is quite specific about this.

For some reason, I thought the errors were more catastrophic looking in the past, but I agree, they look okay in the modern day:

```
CMake Error in runtime_src/core/pcie/linux/CMakeLists.txt:
  Target "xrt_core" requires the language dialect "CXX17" (with compiler
  extensions), but CMake does not know the compile flags to use to enable it.
```

stsoe:
Not wanted.

PR #6835, opened 2022-07-08
Add petalinux in apu package name

rbramand-xilinx:
Added petalinux string in apu package name, apu packages will now look like -> **xrt-apu_202220.2.14.192_petalinux_all.deb** and **xrt-apu_202220.2.14.192_petalinux.noarch.rpm**
At present adding this change only when XRT_VERSION_PATCH is defined so that package name generated from xrt pipeline is only affected.

Renaming the package after creation as package version should not change from 2.14.x

none

NA

### DIFF
--- a/src/runtime_src/tools/scripts/pkgapu.sh
+++ b/src/runtime_src/tools/scripts/pkgapu.sh
@@ -297,6 +297,13 @@ fi
 dodeb $INSTALL_ROOT
 dorpm $INSTALL_ROOT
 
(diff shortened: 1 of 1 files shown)


rozumx (changes requested):

F.Y.I. - Upon further reflection regarding my earlier comment about adding the arm64 architecture target, I was mistaken and it does not apply to this use-case.   That said all looks food from my perspective.

rozumx (approved):
Looks good.  Thank you for making this change.

PR #6836, opened 2022-07-08
Vitis-4626 Deprecated all XCL APIs

stsoe:
Move content of include/xrt.h to include/deprecated/xrt.h

The APIs are deprecated and are for internal use only.
The replacement APIs are xrt native C++ APIs.

Post deprecation the appropriate APIs will move to a private none
exported header.

Documentation must be updated to remove XCL


stsoe:
retest this please - doesn't look like a legit failure

PR #6838, opened 2022-07-11
Add sc_pid to SCU subdevice sysfs to support PS kernel stat gathering

jeffli-xilinx:
Allows query of PS kernel pid

None

Expose pid through sysfs node on SCU platform subdevice

None

PS kernel validate tests

Example sysfs node:
/sys/devices/platform/ert_hw/SCU.4.auto/sc_pid

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/scu.c
+++ b/src/runtime_src/core/edge/drm/zocl/scu.c
@@ -32,8 +32,8 @@ struct zocl_scu {
 	 * soft cu pid and parent pid. This can be used to identify if the
 	 * soft cu is still running or not. The parent should never crash
 	 */
-	uint32_t		sc_pid;
-	uint32_t		sc_parent_pid;
+	u32		sc_pid;
+	u32		sc_parent_pid;
 };
 
 static ssize_t debug_show(struct device *dev,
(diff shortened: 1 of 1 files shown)


rozumx (commented):
Jeff,
Could you update the PR description with an example directory structure.

src/runtime_src/core/edge/drm/zocl/scu.c
```diff
@@ -92,11 +92,33 @@ stat_show(struct device *dev, struct device_attribute *attr, char *buf)
 }
 static DEVICE_ATTR_RO(stat);
 
+ssize_t show_sc_pid(u32 sc_pid, char *buf)
+{
+	ssize_t sz = 0;
+	char *fmt = "%u\n";
```
dbenusov-xilinx, line 98:
```suggestion
	char *fmt = "PID:%u\n";
```
To match all the other files would it be a good idea to have the format like this? Even if it is one element.

src/runtime_src/core/edge/drm/zocl/scu.c
```diff
@@ -92,11 +92,33 @@ stat_show(struct device *dev, struct device_attribute *attr, char *buf)
...
+
+static ssize_t
+sc_pid_show(struct device *dev, struct device_attribute *attr, char *buf)
+{
+	struct platform_device *pdev = to_platform_device(dev);
+	struct zocl_scu *scu = platform_get_drvdata(pdev);
+
+	return show_sc_pid(scu->sc_pid, buf);
+}
+static DEVICE_ATTR_RO(sc_pid);
```
dbenusov-xilinx, line 114:
```suggestion
static DEVICE_ATTR_RO(status);
```
If there is every additional data would we want to create a different node? We could place it all into one attribute file to reduce the amount of files to access. Renaming the attribute to `status` mimics the behavior of the `/proc/` entries. The PID of the process in `/proc/` is also stored in the `status` attribute so the logic matches up. Any additional information we would ever need to add could be easily parsed into the PS kernel I am developing by adding another row of data.

dbenusov-xilinx (changes requested):

chvamshi-xilinx (approved):

dbenusov-xilinx (approved):
Thank you!

PR #6840, opened 2022-07-12
XBUtilties::xclbin_lock should not use xcl APIs

stsoe:
Change xcl APIs to core XRT device APIs

xbutil2 and xbmgmt2 should be rid of xclAPI dependencies.

This is just one small change.  Subcmd validate need fixing.


stsoe:
retest this please - know edge error

PR #6841, opened 2022-07-13
add query request to know if APU/ERT is connected

mamin506:
Add query request to identify if APU/ERT is connected. The "connected" status means that host was able to communicate with APU/ERT. It doesn't mean that APU/ERT is still alive at the moment.

This can be used to tell that if APU package is loaded from the userpf perspective.

### DIFF
--- a/src/runtime_src/core/common/query_requests.cpp
+++ b/src/runtime_src/core/common/query_requests.cpp
@@ -177,4 +177,24 @@ to_range(const std::string& range_str)
   return range;
 }
 
+xrt_core::query::ert_status::ert_status_data
+xrt_core::query::ert_status::
+to_ert_status(const result_type& strs)
+{
+  using tokenizer = boost::tokenizer< boost::char_separator<char> >;
+  xrt_core::query::ert_status::ert_status_data ert_status = {0};
+
+  for (auto& line : strs) {
+    // Format on each line: "<name>: <value>"
+    boost::char_separator<char> sep(":");
+    tokenizer tokens(line, sep);
+    auto tok_it = tokens.begin();
+    if (line.find("Connected:") != std::string::npos) {
+      ert_status.connected = std::stoi(std::string(*(++tok_it)));
+    }
+  }
+
+  return ert_status;
+}
+
 }} // query, xrt_core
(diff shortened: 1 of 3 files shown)


maxzhen (approved):

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -306,6 +306,39 @@ struct sdm_sensor_info
...
+{
+  using result_type = query::ert_status::result_type;
+
+  static result_type
+  get(const xrt_core::device* device, key_type)
+  {
+    using tokenizer = boost::tokenizer< boost::char_separator<char> >;
+    auto pdev = get_pcidev(device);
+    std::vector<std::string> status;
+    result_type ertStatus = {0};
```
stsoe, line 319: No mixed case, please

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -306,6 +306,39 @@ struct sdm_sensor_info
...
+
+      auto tok_it = tokens.begin();
+      if (line.find("Connected:") != std::string::npos) {
+        ertStatus.connected = std::stoi(std::string(*(++tok_it)));
+        std::cout << line << std::endl;
+      }
+    }
+
+    return ertStatus;
+  }
```
stsoe, line 339: Similar to some other recent query addition, I don't think this scales,  the query_requests should be as simple as possible gathering the raw data from sysfs using emplace_sysfs_get, parsing of the raw data should be done upstream in query_requests.cpp.     Is that possible?

stsoe (changes requested):

src/runtime_src/core/pcie/linux/device_linux.cpp
mamin506: I will update this PR and respect the rules.

src/runtime_src/core/common/query_requests.cpp
```diff
@@ -177,4 +177,24 @@ to_range(const std::string& range_str)
...
+{
+  using tokenizer = boost::tokenizer< boost::char_separator<char> >;
+  xrt_core::query::ert_status::ert_status_data ert_status = {0};
+
+  for (auto& line : strs) {
+    // Format on each line: "<name>: <value>"
+    boost::char_separator<char> sep(":");
+    tokenizer tokens(line, sep);
+    auto tok_it = tokens.begin();
+    if (line.find("Connected:") != std::string::npos) {
```
rozumx, line 192: Curly brackets are not needed.

rozumx (approved):
minor code changes.

PR #6854, opened 2022-07-15
Remove scheduler init from shim as driver is doing the same

saifuddin-xilinx:
scheduler::init can not be skipped from shim.  New KDS fixes this by ensuring that scheduler init is done by driver itself along with icap download.


stsoe (approved):
Ah, I see, I was hoping the function itself could have been removed, but it is used in emulation as well so we can't get rid of it, I guess.

maxzhen (approved):

chvamshi-xilinx (approved):

PR #6855, opened 2022-07-17
CR-1135117 when hardware is tripped, avoid dmesg being flooded by 65535

xdavidz:
xgq cids

stop xgq services when cid 0xffff is received.

CR-1135117 

when there is hardware error, like downloading xclbin cause the hardware issue, cid becomes 0xffff,
the dmesg can be flooded with full of cid is 65535.

N/A

tested with vck5000 and v70 shell(currently has xclbin download issue)

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
@@ -76,7 +76,7 @@
 #define	XGQ_DEV_NAME "ospi_xgq" SUBDEV_SUFFIX
 
 static DEFINE_IDR(xocl_xgq_vmr_cid_idr);
-#define XOCL_VMR_INVALID_CID	0xFFFF;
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -76,6 +76,7 @@
 #define	XGQ_DEV_NAME "ospi_xgq" SUBDEV_SUFFIX
 
 static DEFINE_IDR(xocl_xgq_vmr_cid_idr);
+#define XOCL_VMR_INVALID_CID	0xFFFF;
```
xuhz, line 79: who uses this new introduced define?

xdavidz: I will fix this, great catch!

xuhz (approved):

maxzhen (approved):

PR #6859, opened 2022-07-18
skeleton aie trace plugin with x86 support

hackwa:
Add a new aie trace plugin for x86 trace using pskernels. This will replace the original aie trace plugin in future with support for edge.
NA
NA
Low
tested on a pcie design and tested compilation for edge


stsoe (approved):

jvillarre (approved):
Approved as this is the skeleton code we will be building additional functionality into in subsequent pull requests.

PR #6861, opened 2022-07-19
Vitis-5925: Improve bandwidth calculation in profiling summary

jvillarre:
When enabled, the profile summary reports kernel to global memory data transfers, and gives an average bandwidth utilization.  This bandwidth utilization was calculated using assumptions on maximum bandwidth between the kernel and memory that did not take into account differences of connections, such as bit-width and clock rate.

This pull request improves the kernel to global memory table by reporting the achieved bandwidth that was observed as well as the ideal bandwidth on an ideal connection.  The ideal bandwidth is based on the maximum bit-width of a connection and the maximum clock speed of that connection, while the achieved bandwidth is based on the actual connection's bit-width, actual connection's clock rate, and the memory it is connected to.

Incorrect or misleading values in the profile summary tables were discovered by manual inspection and comparison with expected values.

The profiling library now parses additional information in the system metadata section in the xclbin to determine the characteristics of each port of a compute unit.  This includes bit-width, clock speed, and the memory connection.  This information is now collected as a new Port struct which is stored as part of a compute unit struct in the profiling side.

This implementation had the additional benefit of collecting the arguments associated with each port which was previously only available in OpenCL flows, so this information is now available in Native XRT flows as well.

Low risk as the bandwidth numbers previously reported were unclear, and if the metadata is not available then numbers will not be reported.

The changes have been run on various different hardware configurations of ports in order to see that the calculations are working as expected.  Full testing on large regression suites will be necessary to test platform specific information.

Documentation on the profiling table columns will have to be updated to reflect the new fields and what they mean.


pgschuey (approved):

src/runtime_src/xdp/profile/database/static_info_database.cpp
```diff
@@ -1428,57 +1403,213 @@ namespace xdp {
...
+        for (auto& kernel : region.second.get_child("kernels")) {
+          auto kernelName = kernel.second.get<std::string>("name", "");
+          for (auto& port : kernel.second.get_child("ports")) {
+            auto portName = port.second.get<std::string>("name");
+            auto portType = port.second.get<std::string>("port_type");
+            if (portName == "S_AXI_CONTROL" || portType == "stream")
+              continue;
+
+            auto portWidth = port.second.get<std::string>("data_width");
+	    std::transform(portName.begin(), portName.end(), portName.begin(),
```
hackwa, line 1517: looks misaligned. Mix of tabs and spaces?

jvillarre: Changed these lines to spaces to align in the latest commit

hackwa (approved):
looks good!

dayeh-xilinx:
retest this please.

dayeh-xilinx:
retest this please.

PR #6871, opened 2022-07-24
Vitis-6279 Support xclIPSetReadRange in xrt::kernel for reading from shared CUs

stsoe:
Add XRT C++ API for setting compute unit read range.

Feature request.

xrt::set_read_range(const xrt::kernel& kernel, ...)
This function defines the read-only register range on a compute unit
associated with a kernel if and only if (1) the kernel has exactly
one compute unit and (2) the compute unit is opened in shared mode.
The read range allows xrt::kernel::read_register to read directly
from compute unit registers even if the compute unit is shared between
this process and another or between multiple kernel objects.

Added shim level support in xrt_core::device::ishim and defined in
Alveo Linux shim.  No other shims support this feature currently.

None unless the new APIs is used.

Nothing has been tested.  Feature requestor will have to test.

None, this is an undocumented feature.

### DIFF
--- a/src/runtime_src/core/pcie/linux/device_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/device_linux.cpp
@@ -50,28 +50,27 @@ namespace {
 namespace query = xrt_core::query;
 using pdev = std::shared_ptr<pcidev::pci_device>;
 using key_type = query::key_type;
-const static int LEGACY_COUNT = 4;
 
 static int
-get_render_value(const std::string dir)
+get_render_value(const std::string& dir)
 {
   static const std::string render_name = "renderD";
-  int instance_num = INVALID_ID;
-  std::string sub;
+  int instance_num = INVALID_ID; // argh, what is this?
 
   boost::filesystem::path render_dirs(dir);
   if (!boost::filesystem::is_directory(render_dirs))
     return instance_num;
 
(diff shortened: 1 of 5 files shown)


src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -55,17 +55,17 @@ const static int LEGACY_COUNT = 4;
...
 
   boost::filesystem::path render_dirs(dir);
   if (!boost::filesystem::is_directory(render_dirs))
     return instance_num;
- 
+
   boost::filesystem::recursive_directory_iterator end_iter;
     for(boost::filesystem::recursive_directory_iterator iter(render_dirs); iter != end_iter; ++iter) {
-      if (iter->path().filename().string().compare(0,render_name.size(),render_name)== 0) {   
+      if (iter->path().filename().string().compare(0,render_name.size(),render_name)== 0) {
```
keryell, line 68: While you are here, contribute to a better world by using `==` instead of compare. :-)

stsoe: > While you are here, contribute to a better world by using `==` instead of `compare`. :-)
haha, sure, didn't see that :-)

mamin506 (approved):

maxzhen (approved):

PR #6872, opened 2022-07-25
CR-1132776: Fixing the kernel data transfers table in the profile summary

jvillarre:
The profile summary table for kernel data transfers was incorrectly reporting reads as writes on some designs.

The root cause of the misreported rows was in how the counter results read from hardware were accessed.  This pull request now takes advantage of the stored slot index in each monitor in order to correctly access the counter values read from hardware.


IshitaGhosh (approved):
Change looks good.

PR #6834, opened 2022-07-04
CR-1133835 Print progress report when flashing SC using xgq's

rbramand-xilinx:
#### Problem solved by the commit
When flashing SC using xgq's on vck5000 kind of devices continuous dots will be printed to indicate progress

#### How problem was solved, alternative solutions (if any) and why they were rejected
Flashing SC is made as blocking call and while we wait progress is reported by printing continuous dots. Percentage is not printed at present as we dont have any info from VMR on how much progress is being made.

#### Risks (if any) associated the changes in the commit
none

#### What has been tested and how, request additional testing if necessary
Tested by flashing on vck5000 device and the result is as expected.

eg: time ./build/Release/opt/xilinx/xrt/bin/unwrapped/xbmgmt2 program --base all --image xilinx_vck5000_gen4x8_qdma_base_1 -d --force

Device : [0000:d8:00.0]
Current Configuration
  Platform              : xilinx_vck5000_gen4x8_qdma_base_1
  SC Version           : 4.4.33
  Platform ID          : 0xc727c8224d876f34
Incoming Configuration
  Deployment File          : partition.xsabin
  Deployment Directory : /lib/firmware/xilinx/c727c8224d876f34d05896abf1aaff02
  Size                              : 7,646,083 bytes
  Timestamp                   : Wed Jun 29 13:47:23 2022
  Platform                       : xilinx_vck5000_gen4x8_qdma_base_1
  SC Version                    : 4.4.33
  Platform UUID        : C727C822-4D87-6F34-D058-96ABF1AAFF02

Actions to perform:
  [0000:d8:00.0] : Program base (FLASH) image
  [0000:d8:00.0] : Program Satellite Controller (SC) image

Are you sure you wish to proceed? [Y/n]: Y (Force override)

INFO: Forcing flashing of the Satellite Controller (SC) image (Force flag is set).
[0000:d8:00.0] : Updating Satellite Controller (SC) firmware flash image
**..........................................................................................**

INFO: Forcing flashing of the base (e.g., shell) image (Force flag is set).
#### Documentation impact (if any)
NA

PR #6874, opened 2022-07-27
add sysfs node to tell if device can support ps kernel

mamin506:
@dbenusov-xilinx is asking for a way to tell if a device can support PS kernel or not.

In this change, I only mark VERSAL ES3 shell. U30 type shell will not show PS kernel support.
With the query request from #6841 and this one, xbutil is able to know that if shell has APU and if APU is loaded. 

Update: #6834 add similar sysfs node in the mgmtpf side and add query request. This PR is for userpf.

Low

vck5000

No

### DIFF
--- a/src/runtime_src/core/common/query_requests.h
+++ b/src/runtime_src/core/common/query_requests.h
@@ -287,7 +287,6 @@ enum class key_type
   cu_size,
   cu_read_range,
 
-  can_ps_kernel,
   noop
 };
 
@@ -3104,23 +3103,6 @@ struct cu_read_range : request
 
   static range_data
   to_range(const std::string& str);
-};
-
-// query can_ps_kernel sysfs node on userpf.
-// this can tell if the device can support PS kernel or not.
-struct can_ps_kernel : request
-{
-  using result_type = bool;
-  static const key_type key = key_type::can_ps_kernel;
-
-  virtual boost::any
-  get(const device*) const = 0;
-
-  static std::string
-  to_string(result_type value)
-  {
-    return value ? "true" : "false";
-  }
 };
 
 } // query
--- a/src/runtime_src/core/common/query_requests.h
+++ b/src/runtime_src/core/common/query_requests.h
@@ -3106,6 +3106,8 @@ struct cu_read_range : request
   to_range(const std::string& str);
 };
 
+// query can_ps_kernel sysfs node on userpf.
+// this can tell if the device can support PS kernel or not.
 struct can_ps_kernel : request
 {
   using result_type = bool;
@@ -3119,7 +3121,7 @@ struct can_ps_kernel : request
   {
     return value ? "true" : "false";
   }
-}
+};
 
 } // query
 
--- a/src/runtime_src/core/common/query_requests.h
+++ b/src/runtime_src/core/common/query_requests.h
@@ -287,6 +287,7 @@ enum class key_type
   cu_size,
   cu_read_range,
 
+  can_ps_kernel,
   noop
 };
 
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/common/query_requests.h
```diff
@@ -3105,6 +3106,21 @@ struct cu_read_range : request
   to_range(const std::string& str);
 };
 
+struct can_ps_kernel : request
```
stsoe, line 3109: A comment preceding this struct would be helpful.
We have not been good at commenting the query structs, but let's at least do it for new ones.

stsoe (changes requested):

chvamshi-xilinx:
There is already a query to check whether device is versal or not.  https://github.com/Xilinx/XRT/pull/6841 & https://github.com/Xilinx/XRT/blob/master/src/runtime_src/core/common/query_requests.h#L2513 check should be sufficient to check whether device supports PS kernel or not.

mamin506:
> 

Thanks @chvamshi-xilinx, but it is implement on the mgmtpf side. Let me double check with @dbenusov-xilinx. If he need it to be at the userpf. I can implement it. Anyway, there is not need to have a new query request. I will use "is_versal".

stsoe (approved):

mamin506:
@rozumx / @stsoe , based on what @chvamshi-xilinx points out. There is already a "is_versal" query request implement for the same purpose. I don't need to introduce a new one. Just add "versal" sysfs node in the xocl.

kbhardwa-xilinx:
retest this please

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_sysfs.c
```diff
@@ -765,6 +765,18 @@ static ssize_t host_mem_size_show(struct device *dev,
 }
 static DEVICE_ATTR_RO(host_mem_size);
 
+static ssize_t versal_show(struct device *dev,
+	struct device_attribute *attr, char *buf)
+{
+	struct xocl_dev *xdev = dev_get_drvdata(dev);
+	uint32_t val = 0;
+
+	val = (XOCL_DSA_IS_VERSAL_ES3(XDEV(xdev))) ? 1 : 0;
```
dbenusov-xilinx, line 774: This is slightly different from the implementation of the mgmtpf. Are they reporting the same information? If so they should both be the same. Maybe changing the mgmtpf portion to use XOCL_DSA_IS_VERSAL_ES3 instead of the bit comparison?

dbenusov-xilinx (changes requested):

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_sysfs.c
mamin506: Oh.. I didn't take a close look at the mgmt implementation. It should use macro which is more readable. I will update it.

mamin506: @dbenusov-xilinx The XOCL_DSA_IS_VERSAL_ES3 macro is only for userpf. There is no similar macro defined in mgmtpf. Actually, the coding style in mgmtpf side is always use bit comparison... I will not make that change.

On the userpf side, XOCL_DSA_IS_VERSAL_ES3 is for the discovery and newer versal shell. That is the target shell for PS kernel. Unless we want to run PS kernel on legacy versal shell. Otherwise, I think current solution is sufficient.

maxzhen (approved):
we can add non-es3 support as well, otherwise, looks good.

dayeh-xilinx:
retest this please.

dayeh-xilinx:
retest this please.

rozumx (approved):

mamin506:
retest this please.

dbenusov-xilinx (approved):

PR #6876, opened 2022-07-28
VITIS-5884 XGQ: Add clock throttling feature support on versal platforms

rajkumar-xilinx:
Adding clock throttling feature support on versal platforms. Added all required changes in tools, user space driver, and in xgq driver. This PR enables user to check the clock throttling(scaling) feature through xbutil/xbmgmt tools "cmc" report and also through XGQ's sysfs node.

Adding clock scaling feature support in XRT
VITIS-5870
Added all required support in XRT to enable clock scaling feature support. End user can read clock scaling feature default settings through xbutil/xbmgmt tool's reports, also can configure clock throttling limits through XGQ's sysfs node.
NA
xbutil/xbmgmt reports, reading/configuring clock scaling feature limits.
NA


rajkumar-xilinx:
Since the supported changes are not ready in VMR, below results are observed. Also, tried with temporary data set in XGQ driver manually in the clock_scaling's opcode response.

**Reading clock scaling default & override settings**
bash-4.2# cat /sys/bus/pci/devices/0000\:d8\:00.0/xgq_vmr.m.54525952/clk_scaling_stat_raw
HAS_CLOCK_THROTTLING:0
CLOCK_THROTTLING_ENABLED:0
POWER_SHUTDOWN_LIMIT:0
TEMP_SHUTDOWN_LIMIT:0
POWER_THROTTLING_LIMIT:0
TEMP_THROTTLING_LIMIT:0
POWER_THROTTLING_OVRD_LIMIT:0
TEMP_THROTTLING_OVRD_LIMIT:0
POWER_THROTTLING_OVRD_ENABLE:0
TEMP_THROTTLING_OVRD_ENABLE:0
CLOCK_THROTTLING_MODE:0

bash-4.2# xbmgmt examine -r cmc -d

-------------------------------------------------------
1/1 [0000:d8:00.0] : xilinx_vck5000_gen4x8_qdma_base_1
-------------------------------------------------------
CMC
  Heartbeat information unavailable
  Runtime Clock Scaling :
    Not supported

**configuring clock scaling settings**
**Enable feature**
echo "1,0,0" > /sys/bus/pci/devices/0000\:d8\:00.0/xgq_vmr.m.54525952/clk_throttling_configure
**Disable feature**
echo "0,0,0" > /sys/bus/pci/devices/0000\:d8\:00.0/xgq_vmr.m.54525952/clk_throttling_configure
**Enable feature & set power limit to 50, temp limit to 45**
echo "1,50,45" > /sys/bus/pci/devices/0000\:d8\:00.0/xgq_vmr.m.54525952/clk_throttling_configure

kbhardwa-xilinx:
retest this please

kbhardwa-xilinx:
retest this please

rajkumar-xilinx:
Verified the PR on U2 board
image

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -126,6 +126,18 @@ struct xgq_worker {
 	struct xocl_xgq_vmr	*xgq_vmr;
 };
 
+struct xgq_clk_scaling_info {
```
xdavidz, line 129: This struct should not be here, please move this to xgq_cmd_vmr.h

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -151,6 +163,8 @@ struct xocl_xgq_vmr {
 	struct semaphore 	xgq_data_sema;
 	struct semaphore 	xgq_log_page_sema;
 	struct xgq_cmd_cq_default_payload xgq_cq_payload;
+	//clock scaling structure to save all feature details
+	struct xgq_clk_scaling_info xgq_cs_struct;
```
xdavidz, line 167: We should not have this specific additional struct iin xocl_xgq_vmr.
We should have a scaling_payload added in xgq_cmd_cq in the xgq_cmd_vmr.h.
Example code:
struct xgq_cmd_cq {
           hdr;
           union {
                cq_default_payload;
                cq_log_payload; 
                cq_cls_payload; //clock scaling payload? you can have your preferred name
            }

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -1657,6 +1671,118 @@ static int vmr_status_query(struct platform_device *pdev)
...
 
+static void clk_scaling_cq_result_copy(struct xocl_xgq_vmr *xgq,
+                                       struct xocl_xgq_vmr_cmd *cmd)
+{
+	struct xgq_cmd_cq_default_payload *payload =
+		(struct xgq_cmd_cq_default_payload *)&cmd->xgq_cmd_cq_payload;
+	struct xgq_cmd_cq_clk_scaling_payload *cs_payload=
+		(struct xgq_cmd_cq_clk_scaling_payload *)&xgq->xgq_cq_payload;
+
+	mutex_lock(&xgq->xgq_lock);
```
xdavidz, line 1682: you don't have to copy the payload again, because it has been copied out automatically into the xgq_cmd_cq_payload.

you can directly use any of those by:
((struct xgq_cmd_cq_cls_payload *)&cmd->xgq_cmd_cq_payload)->has_clk_scaling;
etc.

xdavidz: See examples in clock and log_page_payload.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
xdavidz, line 167: We don't have to keep extra storage for cs. The xgq_cq_payload has everything cached already.
Check vmr_status_show for a reference.
You actually have already casted to it in the clk_scaling_cq_result_copy.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -2072,6 +2198,94 @@ static ssize_t vmr_debug_type_store(struct device *dev,
...
+
+static ssize_t clk_scaling_configure_store(struct device *dev,
+                                           struct device_attribute *attr,
+                                           const char *buf, size_t count)
+{
+	struct xocl_xgq_vmr *xgq = platform_get_drvdata(to_platform_device(dev));
+	uint8_t enable = 0;
+	uint16_t pwr = 0;
+	uint8_t temp = 0;
+	char* pch = buf;
```
xdavidz, line 2232: what is pch?
Could you please make those variables a little readable?

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -2072,6 +2198,94 @@ static ssize_t vmr_debug_type_store(struct device *dev,
...
+		pch = strsep(&end, ",");
+		if ((kstrtou8(pch, 10, &temp) == -EINVAL) ||
+			(temp > xgq->xgq_cs_struct.temp_scaling_limit)) {
+			XGQ_ERR(xgq, "Invalid temp override limit %u provided, wereas max limit is %u",
+					temp, xgq->xgq_cs_struct.temp_scaling_limit);
+			return -EINVAL;
+		}
+		pch = end;
+	}
+	mutex_lock(&xgq->xgq_lock);
```
xdavidz, line 2272: This is very good, maybe the vmr_status also need mutex to protect. Otherwise, the return payload might be overrided to corrupted data. I will fix my code.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -2072,6 +2198,94 @@ static ssize_t vmr_debug_type_store(struct device *dev,
...
+	ret = clk_scaling_configure_op(xgq->xgq_pdev,
+                                   XGQ_CMD_CLK_THROTTLING_AID_CONFIGURE,
+                                   enable, pwr, temp);
+	if (!ret) {
+		xgq->xgq_cs_struct.clk_scaling_en = enable;
+		if (pwr)
+			xgq->xgq_cs_struct.pwr_scaling_ovrd_limit = pwr;
+		if (temp)
+			xgq->xgq_cs_struct.temp_scaling_ovrd_limit = temp;
+	}
```
xdavidz, line 2282: see vmr_status_query. We should collect data or copy data out from here.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
rajkumar-xilinx: Hi @xdavidz thanks for review. I have taken care of your suggestions. Please review it again, thank you.

src/runtime_src/core/include/xgq_cmd_common.h
```diff
@@ -91,6 +91,7 @@ enum xgq_cmd_opcode {
 	XGQ_CMD_OP_LOAD_APUBIN		= 0xd,
 	XGQ_CMD_OP_VMR_CONTROL		= 0xe,
 	XGQ_CMD_OP_PROGRAM_SCFW		= 0xf,
+	XGQ_CMD_OP_CLK_THROTTLING	= 0x10,
```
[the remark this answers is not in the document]
rajkumar-xilinx: yes, Sandeep Prepared PR https://github.com/Xilinx/VMR/pull/191/files

rajkumar-xilinx:
I have verified the PR using VMR.elf (local drop from Sandeep) which has clock scaling support. Able to read clock scaling default configurations during driver load time and dmesg will show if the clock scaling feature is supported|enabled.
By default, clock scaling is not enabled.

> [  271.563056] xclmgmt 0000:d8:00.0: xgq_vmr.m.54525952 ffff95238c286010 xgq_vmr_probe: clock scaling feature is supported, and enable status: 0

Enable & set overrides
echo "1,200,80" > /sys/bus/pci/devices/0000\:d8\:00.0/xgq_vmr.m.54525952/clk_scaling_configure

-bash-4.2# xbmgmt examine -r cmc -d

[0000:d8:00.0] : xilinx_vck5000_gen4x8_qdma_base_1

CMC
  Heartbeat information unavailable
  Runtime Clock Scaling :
    Critical threshold (clock shutdown) limits:
      Power : 69 W
      Temperature : 100 C
    Throttling threshold limits:
      Power : 290 W
      Temperature : 95 C
    Power threshold override:
      Override : true
      Override limit : 200 W
    Temperature threshold override:
      Override : true
      Override limit : 80 C

-bash-4.2# cat /sys/bus/pci/devices/0000\:d8\:00.0/xgq_vmr.m.54525952/clk_scaling_stat_raw
HAS_CLOCK_THROTTLING:1
CLOCK_THROTTLING_ENABLED:1
POWER_SHUTDOWN_LIMIT:69
TEMP_SHUTDOWN_LIMIT:100
POWER_THROTTLING_LIMIT:290
TEMP_THROTTLING_LIMIT:95
POWER_THROTTLING_OVRD_LIMIT:200
TEMP_THROTTLING_OVRD_LIMIT:80
POWER_THROTTLING_OVRD_ENABLE:200
TEMP_THROTTLING_OVRD_ENABLE:80
CLOCK_THROTTLING_MODE:2

**Disable clock scaling and  recheck the status using cmc report**
cs_disable

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
rajkumar-xilinx: I added a comment just before the function as requested. please review.

src/runtime_src/core/include/xgq_cmd_vmr.h
```diff
@@ -286,6 +317,22 @@ struct xgq_cmd_cq_data_payload {
...
+ *
+ * clock scaling status
+ */
+struct xgq_cmd_cq_clk_scaling_payload {
+	uint8_t has_clk_scaling:1;
+	uint8_t clk_scaling_mode:2;
+	uint8_t clk_scaling_en:1;
+	uint8_t rsvd:4;
+	uint16_t temp_shutdown_limit:8;
+	uint16_t temp_scaling_limit:8;
```
xdavidz, line 331: there is a pad uint8_t. Because it is u16 aligned for this structure. The risk here is that the sizeof(struct *) might be different than we think it should be. So please add a padding.

xdavidz (approved):

rajkumar-xilinx:
Verified the PR using latest qdma base-1 package taken from 2022.2_daily_latest builds.

  Platform             : xilinx_vck5000_gen4x8_qdma_base_1
  SC Version           : 4.4.35
  Platform UUID        : 9A3B9518-4E5A-B131-5FC0-D76AFF4C8991
  Interface UUID       : 527F6489-4668-77D5-E224-FDB2330ACB6B

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -451,6 +451,111 @@ struct hotplug_offline
...
+    std::vector<std::string> stats;
+    std::string errmsg;
+    result_type ctStats;
+
+    pdev->sysfs_get("xgq_vmr", "clk_scaling_stat_raw", errmsg, stats);
+    if (!errmsg.empty()) {
+      // Backward compatibilty check.
+      // Read XMC sysfs nodes
+      data_type data = { 0 };
+	  uint32_t uint_op = 0;
```
rozumx, line 473: Format issue. Line shouldn't be indented.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -451,6 +451,111 @@ struct hotplug_offline
...
+        data.pwr_scaling_ovrd_limit = uint_op;
+      pdev->sysfs_get<uint32_t>("xmc", "scaling_threshold_power_override_en", errmsg, uint_op, EINVAL);
+      if (errmsg.empty())
+        data.pwr_scaling_ovrd_enable = uint_op ? true : false;
+      pdev->sysfs_get<uint32_t>("xmc", "scaling_threshold_temp_override_en", errmsg, uint_op, EINVAL);
+      if (errmsg.empty())
+        data.temp_scaling_ovrd_enable = uint_op ? true : false;
+      ctStats.push_back(data);
+      return ctStats;
+	}
```
rozumx, line 507: Format Issue: There should not be an indent.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -451,6 +451,111 @@ struct hotplug_offline
...
+  get(const xrt_core::device* device, key_type)
+  {
+    auto pdev = get_pcidev(device);
+
+    std::vector<std::string> stats;
+    std::string errmsg;
+    result_type ctStats;
+
+    pdev->sysfs_get("xgq_vmr", "clk_scaling_stat_raw", errmsg, stats);
+    if (!errmsg.empty()) {
```
rozumx, line 469: If I understand correctly, if there is an error message, this code is assuming that the "old XMC sysfs nodes" should be used.   

If so this is not a clean solution.  Possible alternatives:
1. See if pdev->sysfs_get() retruns a err value.  If so, use this to determine if it is the error of interest.
2. Query the driver to determine which version is support and make a decision with this value.

The solutions that involve:
A. Examine the errmsg for the error of interest. 
B. Examining if stats is empty.

Are side effect kludges that should not be done.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -451,6 +451,111 @@ struct hotplug_offline
...
+      // Backward compatibilty check.
+      // Read XMC sysfs nodes
+      data_type data = { 0 };
```
rozumx, line 472:
```suggestion
      data_type data = {};
```

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -451,6 +451,111 @@ struct hotplug_offline
...
+	  uint32_t uint_op = 0;
+
```
rozumx, line 474: This body of code should be its own separate function.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -451,6 +451,111 @@ struct hotplug_offline
...
+      if (errmsg.empty())
+        data.pwr_scaling_ovrd_enable = uint_op ? true : false;
+      pdev->sysfs_get<uint32_t>("xmc", "scaling_threshold_temp_override_en", errmsg, uint_op, EINVAL);
+      if (errmsg.empty())
+        data.temp_scaling_ovrd_enable = uint_op ? true : false;
+      ctStats.push_back(data);
+      return ctStats;
+	}
+
+    data_type data = { 0 };
```
rozumx, line 509:
```suggestion
    data_type data = {};
```

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -451,6 +451,111 @@ struct hotplug_offline
...
+    int id = 0;
+    //parse one line at a time
+    // The clk_scaling_stat_raw is printing in formatted string of each line
+    // Format: "%s:%u"
+    // Using colon ":" as separator.
+    for (auto& stat_raw : stats) {
```
rozumx, line 515:
```suggestion
    for (const auto& stat_raw : stats) {
```

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -451,6 +451,111 @@ struct hotplug_offline
...
+        data.temp_scaling_ovrd_enable = uint_op ? true : false;
+      ctStats.push_back(data);
+      return ctStats;
+	}
+
+    data_type data = { 0 };
+    int id = 0;
+    //parse one line at a time
+    // The clk_scaling_stat_raw is printing in formatted string of each line
+    // Format: "%s:%u"
```
rozumx, line 513: Thank you for the comment and example.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -451,6 +451,111 @@ struct hotplug_offline
...
+    // Using colon ":" as separator.
+    for (auto& stat_raw : stats) {
+      std::vector<std::string> stat;
+      boost::split(stat, stat_raw, boost::is_any_of(":"));
+      uint16_t value = std::stoi(std::string(stat.at(1)));
+      switch (id) {
```
rozumx, line 519: This switch statement isn't very scalable and it assumes that the stat array is order.  Meaning the first item is support and the last is temp_scaling_ovrd_enable.  

Minimally, there should be a helper function:
```
uint16_t get_value(const std::string & keyValue)
{
      std::vector<std::string> stat;
      boost::split(stat, keyValue, boost::is_any_of(":"));
     if (stat.size() != 2) {
        const auto errMsg = boost::format("Error: KeyValue pair doesn't meet expected format '<key>:<value>': '%s'") % keyValue;
        throw std::runtime_error(errMsg.str())
      return = std::stoi(std::string(stat.at(1)));
}
```
And it would be to initialize each of the variables:
```
data.support = get_value(stats[0]) ? true : false;
data.enable = get_value(stats[1]) ? true : false;
data.pwr_shutdown_limit = get_value(stats[2]);
```

Of course, it would be better if the key was use to data drive the variables.  This can be done without the need of a switch statement.

rozumx (changes requested):

maxzhen (approved):

xdavidz (commented):
@rajkumar-xilinx please address all comments and put an ETA if possible. The VMR side is ready, but needs to be re-sync with this after the xgq_cmd_xxx header files are merged. Before we decide to switch to submodule, we still manually copy those files over to VMR. So the XRT PR goes first then VMR PR.

rajkumar-xilinx:
Hi @rozumx could you please review this PR again. I have taken care of your review comments. Thank you.

src/runtime_src/core/pcie/linux/device_linux.cpp
rajkumar-xilinx: > @rajkumar-xilinx please address all comments and put an ETA if possible. The VMR side is ready, but needs to be re-sync with this after the xgq_cmd_xxx header files are merged. Before we decide to switch to submodule, we still manually copy those files over to VMR. So the XRT PR goes first then VMR PR.

Hi @xdavidz I have addressed all review comments, requested Stephen to review the PR again. Thank you.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -1677,6 +1680,111 @@ static int vmr_status_query(struct platform_device *pdev)
...
+	cmd->xgq_cmd_cb = xgq_complete_cb;
+	cmd->xgq_cmd_arg = cmd;
+	cmd->xgq_vmr = xgq;
+
+	payload = &(cmd->xgq_cmd_entry.clk_scaling_payload);
+	payload->aid = aid;
+	if (aid == XGQ_CMD_CLK_THROTTLING_AID_CONFIGURE)
+	{
+		payload->scaling_en = enable ? 1 : 0;
+		if (pwr_ovrd_limit)
```
rozumx, line 1731: The logic here has a different logic then expected.   Is the intention truly:
1. If the pwr_ovrd_limit is zero, honor the previous value of psr_scaling_ovrd_limit.
2. If the pwr_ovrd_limit is non-zero, over the previous value of psr_scaling_ovrd_limt with this one.

If so, another possible way of writing this code could be:
`   payload->psr_scaling_ovrd_limit = psr_ovrd_limit ? psr_ovr_limit :  payload->psr_scaling_ovrd_limit`

If not, the why couldn't the code be writing as follows:
`   payload->psr_scaling_ovrd_limit = psr_ovrd_limit`

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -1677,6 +1680,111 @@ static int vmr_status_query(struct platform_device *pdev)
...
+			payload->pwr_scaling_ovrd_limit = pwr_ovrd_limit;
+		if (temp_ovrd_limit)
```
rozumx, line 1733: See above comment regarding `payload->psr_scaling_ovrd_limit`

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -1677,6 +1680,111 @@ static int vmr_status_query(struct platform_device *pdev)
...
+			payload->temp_scaling_ovrd_limit = temp_ovrd_limit;
+	}
+
+	hdr = &(cmd->xgq_cmd_entry.hdr);
+	hdr->opcode = XGQ_CMD_OP_CLK_THROTTLING;
+	hdr->state = XGQ_SQ_CMD_NEW;
+	hdr->count = sizeof(*payload);
```
rozumx, line 1740: The `count `name and `sizeof(*payload`) don't seem to align.

One is normally referring to a 'number of elements'  and the other the size (bytes) to represent the items.  

For example, if one has:
`uint32_t myArray[10];`
The `count `(e.g., number of elements would be 10) and the `size `(e.g., number of bytes) would be 40.

If this code is truly what is desired, then a simply comment explaining why would greatly be helpful.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -2092,6 +2200,116 @@ static ssize_t vmr_debug_type_store(struct device *dev,
 }
 static DEVICE_ATTR_WO(vmr_debug_type);
 
+static ssize_t clk_scaling_stat_raw_show(struct device *dev,
```
rozumx, line 2203: This is a VERY dangerous function for it could crash the system if the the buf is smaller then the strings being printed. 

In addition, it looks like it isn't being used.  Is this still needed?

rozumx (approved):
Upon second review of this PR, I did notices a few logic issues that should be validated to make sure that indeed this is the desired behavior.  Nothing major and the PR filer would know best, hence the approval.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
rajkumar-xilinx: Hi @rozumx thank you for the review. count represents the size in bytes only, copying it's definition from xgq_cmd_vmr.h file below.
 "@count:  how many data returned in bytes".

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
rajkumar-xilinx: It is used to read all the default configurations of clock throttling feature, override limits etc using xgq's sysfs node "clk_scaling_stat_raw". Also, this sysfs node used in device_linux.cpp file to get all feature stats in single call.

bash-4.2# cat /sys/bus/pci/devices/0000:d8:00.0/xgq_vmr.m.54525952/clk_scaling_stat_raw
HAS_CLOCK_THROTTLING:1
CLOCK_THROTTLING_ENABLED:1
POWER_SHUTDOWN_LIMIT:69
TEMP_SHUTDOWN_LIMIT:100
POWER_THROTTLING_LIMIT:290
TEMP_THROTTLING_LIMIT:95
POWER_THROTTLING_OVRD_LIMIT:200
TEMP_THROTTLING_OVRD_LIMIT:80
POWER_THROTTLING_OVRD_ENABLE:200
TEMP_THROTTLING_OVRD_ENABLE:80
CLOCK_THROTTLING_MODE:2

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
rajkumar-xilinx: payload will have clk throttling feature enable, and override limits. Intention here is to send the values only if override limits are valid (non zeros).

chvamshi-xilinx:
Hi @rajkumar-xilinx , As discussed offline, Please address if there is any cleanup required in next PR

PR #6882, opened 2022-08-02
Add accessor for type of xrt::xclbin::ip

stsoe:
Add accessor for type of xrt::xclbin::ip

Per need.  

Define only types that make sense for compute units.  All
types (even undefined in class enum) are comparable with static_cast
to int and correspond to IP_TYPE from xclbin.h

None, new API.

Low level test part of this pull request

### DIFF
--- a/src/runtime_src/core/common/api/xrt_xclbin.cpp
+++ b/src/runtime_src/core/common/api/xrt_xclbin.cpp
@@ -263,6 +263,12 @@ public:
     return m_args[argidx];
   }
 
+  xrt::xclbin::ip::ip_type
+  get_type() const
+  {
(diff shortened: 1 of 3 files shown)


vboggara-xilinx (approved):

uday610 (commented):
@stsoe , in which use-case xrt::ip will represent a ps-kernel? Are we envisioning a user-managed ps-kernel? It is for built-in ps-kernel that would be handled by XDP?

stsoe:
> @stsoe , in which use-case xrt::ip will represent a ps-kernel? Are we envisioning a user-managed ps-kernel? It is for built-in ps-kernel that would be handled by XDP?

It is not the xrt::ip object it is the xrt::xclbin::ip object that represents an IP in the xclbin, so this is meta data access to determine the type of compute unit.

PR #6889, opened 2022-08-03
CR-1094631 --batch option does not remove all escape characters

dbenusov-xilinx:
https://jira.xilinx.com/browse/CR-1094631

The xbmgmt2 validate --batch option does not remove all escape characters

The log below shows the following are still present:-

ESC[38;5;2m
ESC[39m
ESC[F
ESC[2K

The progress bar that was previously used introduced escape characters.

A busy bar replaced the progress bar. The new bar has a clear separation of logic and does not introduce escape characters.

Things look slightly different. PRT may need to review these updates. This is an example of the new output for all tests that require a loading bar @uday610 

CR-1094631

Manual testing with u50 and vck5000 validation tests.

None


rozumx (approved):

For future submissions, please separate format and functionality changes into two separate pull requests.  It helps in understanding what really was changed.

chvamshi-xilinx (approved):

PR #6413, opened 2022-03-14
ESE-52 XRT failed to find drm device in edge platforms

rbramand-xilinx:
#### Problem solved by the commit

> DRM device file path is hardcoded as "/dev/dri/renderD128" in edge platforms
> There are other drivers(eg: display) which also use DRM framework and create drm userspace device, so we cannot guarantee zocl drm device is 'renderD128' everytime

#### Documentation impact (if any)
None

PR #6890, opened 2022-08-04
Fix exception thrown when '/dev/dri/by-path' doesnot exist on edge platforms

rbramand-xilinx:
On Edge platforms when /dev/dri/by-path does not exist exception is thrown, fixed that issue.

PR - https://github.com/Xilinx/XRT/pull/6413 introduced this bug

Fixed it by catching exception and adding default path renderD128 when exception is thrown

none

tested on versal device.

NA


chvamshi-xilinx:
retest this please

PR #6894, opened 2022-08-05
CR 1135678 Optimized the access to the ddr via noc using mmap files

venkatp-xilinx:
Reading the large amounts of date from host to device is taking long time. Bypassed the TLM transaction and writing to the noc mmap file created directly from the host memory. This reduced the simulation time from hours to mins 

This is not a bug, this is a enhancement request.

This is the best solution to transfer the data from the host mem to noc accessible ddr in fast manner

No, currently this change is under ENV control

Ran the impactful design and performance is as expected. Canary is clean with this change.

nope


xilinxgitops:
retest this please

PR #6895, opened 2022-08-05
Fixed CR-1136864 issue for vck5000 device

saifuddin-xilinx:
vck5000 validation test is failing for new kds cu context changes. 
Root cause:
Some issue with ZOCL apth for vck5000 devices. 

Working fine now with validation test.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_kds.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_kds.c
@@ -313,18 +313,22 @@ int zocl_add_context_kernel(struct drm_zocl_dev *zdev, void *client_hdl, u32 cu_
 	}
 	uuid_copy(cctx->xclbin_id, &uuid_null);
 
+	/* Multiple CU context can be active. Initializing CU context list */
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx (approved):

mamin506 (approved):

src/runtime_src/core/edge/drm/zocl/zocl_kds.c
```diff
@@ -341,23 +345,31 @@ int zocl_del_context_kernel(struct drm_zocl_dev *zdev, void *client_hdl, u32 cu_
 
 	mutex_lock(&client->lock);
 	cctx = zocl_check_exists_context(client, &uuid_null);
-	if (cctx == NULL)
+	if (cctx == NULL) {
+		mutex_unlock(&client->lock);
 		return -EINVAL;
+	}
```
xdavidz, line 352: I have a question about this code. It seems that the cctx is allocated from the zocl_check_exists_context and should be freed before returning this call. But there arem many places that we don't free the cctx. Is that expected behavior?

PR #6896, opened 2022-08-05
CR-1136383: Incorrect error message when enabling device level profiling

jvillarre:
In order for the profiling library to communicate correctly with the profiling IP in the PL, the profiling library must verify that the xclbin was built with a recent tool version.  When device level profiling is turned on, some xclbins built with recent tools were incorrectly reporting that they needed to be rebuilt in order to enable device level profiling.

The bug was discovered by running applications that loaded xclbins that did not have the optional BUILD_METADATA section with the xrt.ini switch "device_trace=fine" set.  Even though the xclbin did not contain any profiling IP and was built with the most recent version of the tools, the error message was being output because the BUILD_METADATA section was missing.

The function that checks if an xclbin is valid for device level profiling has changed its implementation to use the major and minor number in the header of the axlf instead of looking into the BUILD_METADATA section.

Tested on a recent xclbin and an xclbin built using 2019.1 tools.

No documentation impact.

### DIFF
--- a/src/runtime_src/xdp/profile/database/static_info_database.cpp
+++ b/src/runtime_src/xdp/profile/database/static_info_database.cpp
@@ -1222,8 +1222,6 @@ namespace xdp {
   }
 
   // ************************************************************************
-  // ***** Functions for  *****
-
 
   bool VPStaticDatabase::validXclbin(void* devHandle)
   {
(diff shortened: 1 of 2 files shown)


rozumx (approved):

IshitaGhosh (approved):
Changes look good.

PR #6901, opened 2022-08-08
CR-1135432 fix decimal positions for electrical report

rajkumar-xilinx:

Using static_cast<double> shows the value in decimal points, but it is not 3 digit precision by default. So, fixed decimal point issue using std::setprecision() in sensor code for electrical report.
CR-1135432
Added changes in common sensor code to fix this issue.
NA
examine's electrical report
NA


rajkumar-xilinx:
**Without this PR**
before

**With this PR**
image

src/runtime_src/core/common/sensor.cpp
```diff
@@ -167,9 +168,12 @@ read_data_driven_electrical(const std::vector<xq::sdm_sensor_info::data_type>& c
      * So, actual sensor value => 12000 * 10 ^ (-3) = 12 Volts.
      */
     auto unitm = pow(10, tmp.unitm);
-    pt.put("voltage.volts", static_cast<double>(tmp.input) * unitm);
-    pt.put("voltage.max", static_cast<double>(tmp.max) * unitm);
-    pt.put("voltage.average", static_cast<double>(tmp.average) * unitm);
+    stream << std::fixed << std::setprecision(3) << static_cast<double>(tmp.input) * unitm;
+    pt.put("voltage.volts", stream.str());
+    stream << std::fixed << std::setprecision(3) << static_cast<double>(tmp.max) * unitm;
```
chvamshi-xilinx, line 173: Don't you need to reset the stream? It gets concatenated to existing stream object.

rajkumar-xilinx: taken care of this. Thank you

chvamshi-xilinx: Better to write simple function to get the value. I see the same line in 6 different places.

dayeh-xilinx:
retest this please.

src/runtime_src/core/common/sensor.cpp
rajkumar-xilinx: In format_base10_shiftdown3() API, decimal_shift value is 1000 and it is constant. In our case, this value may not be 1000 and this value is derived from unitm for us.
So, I can't directly use this function here.

chvamshi-xilinx (approved):

rajkumar-xilinx:
retest this please

src/runtime_src/core/common/sensor.cpp
chvamshi-xilinx: As discussed, i was asking you to create one more generic function. Latest changes looks good

src/runtime_src/core/common/utils.cpp
```diff
@@ -21,6 +21,7 @@
 #include "utils.h"
 #include <atomic>
 #include <cstdint>
+#include <cmath>
```
keryell, line 24: Sort.

rozumx (approved):
Glad to see that the data is correctly be fixed in the underlying property tree.

While this isn't an issue, in the future the reports might be needing a different precision value.  If this should occur, please update the report's format routines and NOT how the values are added to property tree.

src/runtime_src/core/common/utils.cpp
rajkumar-xilinx: Done

dayeh-xilinx:
retest this please.

PR #6904, opened 2022-08-10
PRE/POST kernel support

chienwei-lan:
To support the pre/post kernel and parse the information from aie partition json file, we have to add two more types: CT_PRE and CT_POST


dayeh-xilinx:
retest this please. Failed for a known issue - singletask.lock file not found. [CR-1113069]

chienwei-lan:
The format of cdo type hasn't been finalized yet, close this PR.

PR #6905, opened 2022-08-10
set cu read only range on edge

mamin506:
Add xclIPSetReadRange() for the edge platforms.

Low

zcu104

No.


stsoe (approved):
We need to wire this in to device_linux.{h,cpp} as was done for pcie/linux/shim.

mamin506:
> We need to wire this in to device_linux.{h,cpp} as was done for pcie/linux/shim.

I tried it. But the edge device_linux.{h,cpp} doesn't support subdev. So, it is not able to get sysfs node from a CU subdev. I need help for supporting subdev in device_linux.

On edge platform, it is allow user to access CU register via xclRegRead/xclRegWrite when context is shared. This change is still needed, since this API let DPU team be able to compile their code on edge platform.

stsoe:
> I tried it. But the edge device_linux.{h,cpp} doesn't support subdev. So, it is not able to get sysfs node from a CU subdev. I need help for supporting subdev in device_linux.
> 
> On edge platform, it is allow user to access CU register via xclRegRead/xclRegWrite when context is shared. This change is still needed, since this API let DPU team be able to compile their code on edge platform.

What does subdev have to do with this.  device_linux simply needs to call the shim API you are adding so that XRT coreutil can use the function without linking with libxrt_core.  DPU team should be be using shim directly.

chvamshi-xilinx:
> > I tried it. But the edge device_linux.{h,cpp} doesn't support subdev. So, it is not able to get sysfs node from a CU subdev. I need help for supporting subdev in device_linux.
> > On edge platform, it is allow user to access CU register via xclRegRead/xclRegWrite when context is shared. This change is still needed, since this API let DPU team be able to compile their code on edge platform.
> 
> What does subdev have to do with this. device_linux simply needs to call the shim API you are adding so that XRT coreutil can use the function without linking with libxrt_core. DPU team should be be using shim directly.

@mamin506 , 
There are two things
1) wiring this new api in device_linux.h /.cpp . This is required to compile code successfully when app is using xrt native api.
2) Using this range value in xclRegRead/xclRegWrite.

1st one should be straight forward and you can add it easily,
2nd one is tricky as you don't have cu subdevice, I would suggest you to create one sysfs node which contains name, index, range etc instead of differrent sysfs nodes sothat you can tokenize and use it in xclRegWrite/xclRegRead call
.

mamin506:
> What does subdev have to do with this? device_linux simply needs to call the shim API you are adding so that XRT coreutil can use the function without linking with libxrt_core. DPU team should be be using shim directly.

You are right. Wired this API to ishim in device_linux.

chvamshi-xilinx (approved):
As discussed offline, please file a CR to use this information in xclRegRead/xclRegWrite

PR #6912, opened 2022-08-12
Add experimental support for QoS specification in hwctx

stsoe:
Allow applications to create a hwctx specifying the necessary QoS.
This is *very* preliminary experimental support subject to change w/o
warning.

QoS for a hwctx is supposed to guide the resource solver to find
available HW resources sufficient to statisfy the specified QoS or
fail if no resources are available.  For platforms that support
multiple programmable regions (slots), the QoS can be used to
determine if an existing slot can be shared with another context or if
a new region (slot) needs to be programmed.

QoS is a set of metrics defined as a key-value pair represented as
std::map.  The accepted keys are communicated internally in this
version and their value type (represented as uint32_t) is also
internally documented.

The QoS is bound to evolve as more platform drivers support our
concept of xclbin sharing per QoS specifications, so view this PR as
work-in-progress trial feature until we determine exactly what metrics
are needed can be supported.

This PR changes the type of the QoS parameter and adds a second
parameter for legacy support of CU shared/exclusive access mode.  The
vast number of files changed is due to the extra access mode parameter
and the type changes of the QoS paramter. The QoS type is aliased in
one place so that future change to some more intelligent
representation should trigger ripple effects.

The changes has been tested with standard hw regression tests that
exercise legacy OpenCL test cases.

QoS is internal for now

### DIFF
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -23,7 +23,8 @@ namespace xrt {
 //
 class hw_context_impl
 {
-  using qos_type = xrt::hw_context::qos;
+  using qos_type = xrt::hw_context::qos_type;
+  using access_mode = xrt::hw_context::access_mode;
 
   // Managed context handle with exceptional handling for flows
   // or applications that use load_xclbin (legacy Alveo)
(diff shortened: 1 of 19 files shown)


dayeh-xilinx:
retest this please.

PR #6916, opened 2022-08-12
Created a Daemon to modprobe/rmmod vdu drivers based on request.

chvamshi-xilinx:
in new platform, vdu drivers would be in blacklist at boot time. These modules should be loaded once dynamic region is loaded. As there is no easy way, we have to create a daemon which is going to do this job
This is not a clean solution, but there is no other option to support this usecase.

NA

Created a daemon which waits for the requests from user to modprobe vdu modules. this request can come by user via PS Kernel.

not a scalable solution.

vdk5000 qdma to verify the flows.

NA


stsoe (approved):

rozumx (approved):

chvamshi-xilinx:
retest this  please

PR #6918, opened 2022-08-14
Fix CU context manager to be hwctx aware

stsoe:
For platforms that support multi-xclbin and/or multi hardware contexts
for same or different xclbin, each unique hardware context must open
its corresponding compute unit contexts.

The reason for the changes is xocl adding support for creating
multiple hardware contexts for same xclbin.  This exposed a bug in
the compute unit context handing where the context manager did not
differentiate between different context.  This lead to compute units
contexts being closed prematurely during destruction of first hardware
context.

The changes in this PR manages compute unit context per hwctx and
calls open_cu_context for each used compute unit within a hardware
context regardless of the hardware contexts referring to same physical
hardware slot (something user space XRT doesn't know or care about).

Testing was done with standard opencl hw regression tests.

No end-user changes.

### DIFF
--- a/src/runtime_src/core/common/api/context_mgr.cpp
+++ b/src/runtime_src/core/common/api/context_mgr.cpp
@@ -4,28 +4,18 @@
 #define XRT_CORE_COMMON_SOURCE // in same dll as coreutil
 #define XRT_API_SOURCE         // in smae dll as coreutil
 #include "context_mgr.h"
+#include "hw_context_int.h"
 #include "core/common/cuidx_type.h"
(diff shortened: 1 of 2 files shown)


stsoe:
retest this please - petalinux vitis error

saifuddin-xilinx (approved):

chvamshi-xilinx:
retest this please

stsoe:
retest this please

stsoe:
retest this please - emulation timeout not present yesterday

PR #6925, opened 2022-08-16
CR-1136416 Dumped the deadlock message on to the console and created …

venkatp-xilinx:
CR-1136416 Dumped the deadlock message on to the console and createdthe file pl_deadlock_diagnosis.txt as requested in application path.

Its a new request to dump the HLS kernel deadlock message on the user console and create a file using that message to show that as part of the vitis analyzer

This is part of the new PR PL deadlock detection

Its a enhancement

No

tested the deadlock detection designs 

No


kbhardwa-xilinx:
retest this please

PR #6926, opened 2022-08-16
sed apt -> apt-get for docker build purposes

mahdiaminik:
"apt" is not suitable for docker build

### DIFF
--- a/src/runtime_src/tools/scripts/xrtdeps.sh
+++ b/src/runtime_src/tools/scripts/xrtdeps.sh
@@ -532,7 +532,7 @@ install()
     fi
 
     if [ $FLAVOR == "ubuntu" ] && [ $MAJOR == 20 ]; then
-	apt-get install clang-tidy
+	apt-get install -y clang-tidy
     fi
 
     # Enable EPEL on CentOS/RHEL
--- a/src/runtime_src/tools/scripts/xrtdeps.sh
+++ b/src/runtime_src/tools/scripts/xrtdeps.sh
@@ -373,7 +373,7 @@ update_package_list()
 validate()
 {
     if [ $FLAVOR == "ubuntu" ] || [ $FLAVOR == "debian" ]; then
-        #apt -qq list "${UB_LIST[@]}"
+        #apt-get -qq list "${UB_LIST[@]}"
         dpkg -l "${UB_LIST[@]}" > /dev/null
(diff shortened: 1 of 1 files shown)


keryell:
Curious. Why is it not suitable for Docker? On which OS?
On my Ubuntu 22.04 laptop I can see:
```
rkeryell@rk-xsj:~ (master)$ dpkg -S /usr/bin/apt
apt: /usr/bin/apt
rkeryell@rk-xsj:~ (master)$ dpkg -S /usr/bin/apt-get 
apt: /usr/bin/apt-get
```
so both `apt` and `apt-get` come from the same package. In my case I cannot see why I would use the older and longer `apt-get`.

mahdiaminik:
When you use `FROM baseimage \n RUN cmd` in Dockerfile and build process instead of running `docker run -ti baseimage cmd` docker daemon creates something called "**intermediate container**" to create layers.
Intermediate containers are not created in "**tty**" sandbox.
So when you run `apt update` instead of `apt-get update` you will face the message :
- "apt does not have a stable CLI interface. Use with caution in scripts."

[https://docs.docker.com/develop/develop-images/dockerfile_best-practices/](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/)

mahdiaminik:
also in line 535 :
`apt install clang-tidy`
the parameter `-y` is missing.

keryell:
> also in line 535 : `apt install clang-tidy` the parameter `-y` is missing.

Good point. Can you indent the line while adding it?
It is unclear to me why Ubuntu 20.x requires `clang-tidy` at the first place. I use 22.04 currently. But this is another story...

mahdiaminik:
> > also in line 535 : `apt install clang-tidy` the parameter `-y` is missing.
> 
> Good point. Can you indent the line while adding it? It is unclear to me why Ubuntu 20.x requires `clang-tidy` at the first place. I use 22.04 currently. But this is another story...

In git graph it can be checked that why this line is added, but for answering these build issues I suggest add a `Dockerfile` in the repository (something like [https://github.com/Xilinx/XRT/issues/6942](https://github.com/Xilinx/XRT/issues/6942) , without `sed` command of course) and add `docker build` in the CI pipeline (in this case `.travis.yml` file)

stsoe:
ok to test

dayeh-xilinx:
retest this please.

PR #6936, opened 2022-08-19
Use the correct property for reporting firewall status

richardtobin-xlnx:

The firewall status report generated by xbutil examine is incorrect - instead of decoding the details of the firewall trip based on the value of the firewall status register, the firewall level is being used - the latter is supposed to help identify which firewall instance tripped, not provide the details of the trip cause.  For this we need to display and decode the firewall status property.

By using the correct property.

Perhaps there are people who are expecting the current behaviour and think it to be correct.  Those people would be surprised.

Just code inspection.  I'm not set up to build xbutil from source - I know the CI pipeline will do it for me.  However, the change should be verified by someone who can apply the patch and trip a firewall.


maxzhen:
ok to test

chvamshi-xilinx (approved):

PR #6939, opened 2022-08-19
CR-1138133 xbutil advanced --write-mem --fill shows help menu on success

rchane:
https://jira.xilinx.com/browse/CR-1138133
The issue was first brought up in https://jira.xilinx.com/browse/CR-1136750 before this CR was created. 
`xbutil advanced --write-mem --fill` now properly returns after finishing successfully. The help menu's redundant `size` arg was removed (non-required arguments will not be printed beforehand if they are positional arguments). In addition, a bug where `--write-mem`'s help menu returned a non-zero exit code has been fixed.
This affects the display of the "usage" information in the help menus for commands.
Testing was done on vck5000.
```
rchane@xsjsagarw50:/proj/rdi/staff/rchane/XRT$ xbutil advanced --write-mem --help

DESCRIPTION: Write to a given memory address

USAGE: xbutil advanced --write-mem [--help] [-i arg] [-f arg] [--count arg] -d arg address size
  <address>       - Base address to start from
  <size>          - Block size (bytes) to write

OPTIONS:
  -i, --input        - Input file
  -d, --device       - The Bus:Device.Function (e.g., 0000:d8:00.0) device of interest
  --count            - Number of blocks to write
  -f, --fill         - The byte value to fill the memory with
  --help             - Help to use this sub-command


GLOBAL OPTIONS:
rchane@xsjsagarw50:/proj/rdi/staff/rchane/XRT$ xbutil advanced --write-mem --fill 0xaa -d 0000:65:00.1 0x0 1024
Memory write succeeded
```
N/A


chvamshi-xilinx (commented):
Hi @rchane , is there any reason for choosing address & size as positional arguments?

rchane:
Hello @chvamshi-xilinx, address & size were positional arguments previously, so I left them as such!

chvamshi-xilinx:
> Hello @chvamshi-xilinx, address & size were positional arguments previously, so I left them as such!

Ok. Thanks Ryan. This PR looks good if we are considering these as positional arguments.
Can you please consider converting these position arguments to regular arguments in your next PR as this is anyway undocumented feature.

chvamshi-xilinx (approved):

rchane:
The commit CR number is incorrect, but the CR link/PR title has been updated to reflect the correct CR.

src/runtime_src/core/tools/xbutil2/OO_MemWrite.cpp
```diff
@@ -205,6 +205,7 @@ OO_MemWrite::execute(const SubCmdOptions& _options) const
       addr += size;
     }
     std::cout << "Memory write succeeded" << std::endl;
+    return;
```
rozumx, line 208: Why is there a need to add `return` here?

src/runtime_src/core/tools/xbutil2/OO_MemWrite.cpp
```diff
@@ -58,16 +58,16 @@ OO_MemWrite::execute(const SubCmdOptions& _options) const
 {
   XBU::verbose("SubCommand option: write mem");
 
-  // Parse sub-command ...
```
rozumx, line 61: Could you elaborate why the following changes were required.   These changes bind's this code to the "--help" option name and no longer relies on the "m_help" variable.

Looking over the comments it is mentioned that:
_In addition, a bug where --write-mem's help menu returned a non-zero exit code has been fixed._

I assume that this is the bug that has been addressed, but not 100% sure.

rozumx (changes requested):

src/runtime_src/core/tools/xbutil2/OO_MemWrite.cpp
rchane: Without `return`, the code previously went on after a successful write to `printHelp()` and throws an error.

src/runtime_src/core/tools/xbutil2/OO_MemWrite.cpp
rchane: Yes, that is the bug! `process_arguments` was originally throwing an error when using `xbutil advanced --write-mem --help` due to the missing `--address` option. I then tried to make the help option check consistent with OO_MemRead.cpp. From my understanding, "m_help" is not set until `process_arguments` is called.

rozumx (approved):

PR #6940, opened 2022-08-19
CR-1138144 and CR-1138164: Improve profile summary table column headers

jvillarre:
Adjusting the column header names on the kernel to global memory data transfer table in the profile summary to better reflect the values being presented.  Also adds two new columns to show the maximum bandwidths possible on the current port so the user can compare their performance.

Documentation that describes the table will have to be updated.


pgschuey (approved):
Looks good!

PR #6944, opened 2022-08-21
Memory manager changes for multislot support

saifuddin-xilinx:
This is the changes to support multislot. 
Current memory manager support of XOCL driver doesn't support multislot. 

Changes are as follwos:
- Memory Manager need to be updated for the multi-slot solution. 
- Current memory manager only supports single xclbin and do cleanup for every reload of xclbin.
- The solution should be enhanced so that memory manager should in active and functional while some xclbin is reloading while other xclbin is active. 
- Assumption is that memory topology should be same for all the xclbins, loading for a specific device, only memory connection can be differed. 
- The single slot solution should work without any further modification.  The same can be extended to multislot solution.


saifuddin-xilinx:
retest this please

saifuddin-xilinx:
retest this please

saifuddin-xilinx:
Retest this please

PR #6952, opened 2022-08-22
1138498 : Remove finer granularity configs for AIE Profile/trace which are not supported in 2022.2

IshitaGhosh:
The following AIE Profile/Trace configurations are not supported in 2022.2 due to unavailability of metadata in XRT
AIE_profile_settings : graph_based_interface_tile_metrics, graph_based_mem_tile_metrics, tile_based_mem_tile_metrics
AIE_trace_settings : tile_based_interface_tile_metrics, tile_based_mem_tile_metrics

For AIE_profile_settings.graph_based_interface_tile_metrics, specific graph or specific port could not be supported in XRT 2022.2. This is because interfaces could not be associated with graphs. Also, currently this config could be supported only for "all" graphs and "all" ports. However, that does not add much value as user can use 
AIE_profile_settings.tile_based_interface_tile_metrics = all:<metric> 
for such scenario. So any support for graph_based_interface_tile_metrics is removed in 2022.2.

Also, configs related to mem_tile are for AIE2 only and will not be supported in 2022.2 XRT.

So, get methods for the above configs, their usage and entries in INI parameters for Guidance are removed.


dayeh-xilinx:
retest this please. Github fetch error.

dayeh-xilinx:
retest this please.

jvillarre (approved):

PR #6959, opened 2022-08-24
Fixed ffi_call return code type which was causing stack smashing error

jeffli-xilinx:
Fix PS kernel crash due to not sufficient size for the return code

Use ffi_sarg which will hold signed return value from ffi_call

None

PS kernel iops test

None

### DIFF
--- a/src/runtime_src/core/edge/skd/xrt_skd.cpp
+++ b/src/runtime_src/core/edge/skd/xrt_skd.cpp
@@ -221,7 +221,10 @@ namespace xrt {
       }
 
       ffi_call(&cif,FFI_FN(kernel), &kernel_return, ffi_arg_values);
-      args_from_host[return_offset] = (uint32_t)kernel_return;
+      if (kernel_return >= 0)
+	args_from_host[return_offset] = static_cast<uint32_t>(kernel_return);
+      else
+	args_from_host[return_offset] = 255; // Exit status out of range if returning negative value
 
       // Unmap Buffers
       for(auto i:bo_list) {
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/edge/skd/xrt_skd.cpp
```diff
@@ -167,7 +167,7 @@ namespace xrt {
   XCL_DRIVER_DLLESPEC
   void
   skd::run() {
-    int32_t kernel_return = 0;
+    ffi_sarg kernel_return = 0;
```
rozumx, line 170: This makes sense to match the argument type one line 215 for the function, ffi_call().
The next line 216 has the following code:
`args_from_host[1] = (uint32_t)kernel_return;`
which uses a "C-Style cast".  It should instead use a reinterpret cast:
`args_from_host[1] = reinterpret_cast<uint32_t>(kernel_return);`

rozumx (changes requested):

src/runtime_src/core/edge/skd/xrt_skd.cpp
```diff
@@ -221,7 +221,10 @@ namespace xrt {
       }
 
       ffi_call(&cif,FFI_FN(kernel), &kernel_return, ffi_arg_values);
-      args_from_host[return_offset] = (uint32_t)kernel_return;
+      if (kernel_return >= 0)
```
rozumx, line 224: My I suggest the following:
`args_from_host[return_offset] = (kernel_return >= 0) ? static_cast<uint32_t>(kernel_return) : 255;`

rozumx (approved):
Suggested refactoring code change regarding a more simplified assignment.

PR #6963, opened 2022-08-26
fix bug at cu-stats VITIS-3304

Zenfendson:
Average sq length value in the output of stat sysfs node is always zero

The average submitted length is always zero in all test case. #6933  cu stats introduced the bug 

The branch condition inside the cu_stats_timer which enables the sq statistic update is not right. Correct it by using the right branch condition.

Low

Tested on iops, mem-bw and verify testcases.

No


Zenfendson:
Not caused by my code. Retest this please.

mamin506 (approved):

dayeh-xilinx:
retest this please.

PR #6964, opened 2022-08-26
CR-1138003 P2p tests are hung on board. Regression

stsoe:
Don't share hw_queue between devices.  A hardware queue is for
submitting commands for execution on a device.  Alas a queue must be
specific to a device.

Bug was found in p2p regression testing that use multiple devices.

Fix is to manage hw_queues per device, then per hwctx.  Prior to
change queues were managed only per hwctx, but the underlying hwctx
handle returned by driver is not necessarily unique across
devices. And for platforms that do not support hwctx, the internally
created hwctx share same ctx handle so no matter what the queue
handling has to per device then per ctx.

Standard OpenCL regressions.

### DIFF
--- a/src/runtime_src/core/common/api/hw_queue.cpp
+++ b/src/runtime_src/core/common/api/hw_queue.cpp
@@ -392,17 +392,22 @@ public:
 };
 
 // For time being there is only one hw_queue per hw_context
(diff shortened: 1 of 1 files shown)


dayeh-xilinx:
retest this please.

PR #6972, opened 2022-08-29
CR-1138782 Pass a valid char buffer as argument to required qdma APIs

rajkumar-xilinx:
Pass a valid character buffer to qdma_queue_stop() & qdma_queue_remove() APIs,
which will have QDMA driver's return message either success or error message.

QDMA driver APIs expects a empty char buffer so that it can store error message or success message, and the same can be read in client driver (subdev/qdma.c). So, pass a valid char buffer to required QDMA APIs.
CR-1138782
Added fix in qdma driver by passing a valid char buffer to qdma APIs.
NA
xocl & xclmgmt load & unload drivers test.
NA

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/qdma.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/qdma.c
@@ -438,24 +438,26 @@ static void free_channels(struct platform_device *pdev)
 	struct mm_channel *chan;
 	u32	write, qidx;
 	int i, ret = 0;
+	char	ebuf[MM_EBUF_LEN + 1];
 
 	qdma = platform_get_drvdata(pdev);
 	if (!qdma || !qdma->channel)
 		return;
 
 	for (i = 0; i < qdma->channel * 2; i++) {
+		memset(&ebuf, 0, sizeof (ebuf));
 		write = i / qdma->channel;
 		qidx = i % qdma->channel;
 		chan = &qdma->chans[write][qidx];
 
 		channel_sysfs_destroy(chan);
 
-		ret = qdma_queue_stop(qdma->dma_hndl, chan->queue, NULL, 0);
+		ret = qdma_queue_stop(qdma->dma_hndl, chan->queue, ebuf, MM_EBUF_LEN);
 		if (ret < 0) {
 			xocl_err(&pdev->dev, "Stopping queue for "
 				"channel %d failed, ret %x", qidx, ret);
 		}
-		ret = qdma_queue_remove(qdma->dma_hndl, chan->queue, NULL, 0);
+		ret = qdma_queue_remove(qdma->dma_hndl, chan->queue, ebuf, MM_EBUF_LEN);
 		if (ret < 0) {
 			xocl_err(&pdev->dev, "Destroy queue for "
 				"channel %d failed, ret %x", qidx, ret);
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/qdma.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/qdma.c
@@ -438,14 +438,20 @@ static void free_channels(struct platform_device *pdev)
 	struct mm_channel *chan;
 	u32	write, qidx;
 	int i, ret = 0;
-	char	ebuf[MM_EBUF_LEN + 1];
+	char *ebuf;
 
 	qdma = platform_get_drvdata(pdev);
 	if (!qdma || !qdma->channel)
 		return;
 
+	ebuf = devm_kzalloc(&pdev->dev, MM_EBUF_LEN, GFP_KERNEL);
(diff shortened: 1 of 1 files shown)


rajkumar-xilinx:
**Before this PR, error messages from dmesg log during rmmod of xocl:**

> [163055.554630] xocl 0000:d9:00.1:  ffff9d115f3840b0 __xocl_subdev_destroy: Destroy subdev dma.qdma, cdev           (null)
> [163055.554711] xocl:qdma_queue_stop: invalid argument: buf=          (null), buflen=0
> [163055.554719] xocl 0000:d9:00.1: dma.qdma.u.5242881 ffff9d08dcee2010 free_channels: Stopping queue for channel 0 failed, ret ffffffea
> [163055.574324] xocl:qdma_queue_remove: invalid argument: buf=          (null), buflen=0
> [163055.574329] xocl 0000:d9:00.1: dma.qdma.u.5242881 ffff9d08dcee2010 free_channels: Destroy queue for channel 0 failed, ret ffffffea

**Using this PR,  dmesg log: no error/warning messages seen now**

> [  +0.000066] xocl 0000:d8:00.1:  ffff9cabeeeb5098 __xocl_subdev_destroy: Destroy subdev dma.qdma, cdev           (null)
> [  +0.001389] xocl 0000:d8:00.1:  ffff9cabeeeb5098 __xocl_subdev_destroy: Destroy subdev rom, cdev           (null)
>

chvamshi-xilinx (approved):

rajkumar-xilinx:
Hi @maxzhen could you please review this PR and let me know. Thank you.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/qdma.c
```diff
@@ -438,24 +438,26 @@ static void free_channels(struct platform_device *pdev)
 	struct mm_channel *chan;
 	u32	write, qidx;
 	int i, ret = 0;
+	char	ebuf[MM_EBUF_LEN + 1];
```
maxzhen, line 441: Stack size for kernel thread is generally very limited. On Linux, it's probably just 4K or 8K. It is a bad idea to put a big array on stack like this since you'll run out of stack pretty quick. So, please kzalloc() the buffer and you don't even need memset() later. Or use a much smaller buffer here since no one seems to consume the content after the call anyway?

maxzhen (changes requested):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/qdma.c
rajkumar-xilinx: Hi @maxzhen MM_EBUF_LEN is defined as 256 bytes only and the same buffer is being used for all qdma channels in a loop. It is not tied with any single channel. So, buffer contents cleared using memset(). Even, kzalloc() used, buffer contents has to be cleared as it is. Please let me know if you still want to use kzalloc (dynamic) way. Thank you.

maxzhen: I see why you need to memset to clear it. It is normally more efficient for the API to clear it internally since they know when it is needed and how much space it needs to be cleared. But it is what it is.
The concern is that 256 bytes sounds too big for kernel stack. Please dynamically alloc it or put it in some global data structure so that it does not show up in the stack.

rajkumar-xilinx: Hi @maxzhen I have updated the code to use dynamic alloc for this error buffer. This buffer is used only to store return message from qdma driver, and this buffer should be of some size, but not null.
For each channel, we are using this same buffer hence we are clearing the contents to make sure the buffer doesn't contain previous channel's return message.

rajkumar-xilinx:
Verified the changes on vck5000. No error messages shown from qdma driver now during unloading xocl driver.

image

src/runtime_src/core/pcie/driver/linux/xocl/subdev/qdma.c
```diff
@@ -438,29 +438,39 @@ static void free_channels(struct platform_device *pdev)
...
 		}
-		ret = qdma_queue_remove(qdma->dma_hndl, chan->queue, NULL, 0);
+		ret = qdma_queue_remove(qdma->dma_hndl, chan->queue, ebuf, MM_EBUF_LEN);
 		if (ret < 0) {
 			xocl_err(&pdev->dev, "Destroy queue for "
-				"channel %d failed, ret %x", qidx, ret);
+				"channel %d failed, ret: %x, ebuf: %s", qidx, ret, ebuf);
 		}
 	}
+	if (ebuf)
```
maxzhen, line 472: You don't really need this check since you will not reach here if ebuf is NULL, right?

maxzhen (approved):

rajkumar-xilinx:
retest it please

PR #6977, opened 2022-08-29
Add profiling compile time guards

jvillarre:
This pull request adds (but does not define) compile time guards to different sections of the profiling code base that should be compiled in the main branch, but might want to be excluded in future specialized local branches.

### DIFF
--- a/src/runtime_src/xdp/profile/device/ioctl_monitors/ioctl_add.cpp
+++ b/src/runtime_src/xdp/profile/device/ioctl_monitors/ioctl_add.cpp
@@ -16,8 +16,7 @@
  * under the License.
  */
 
-#ifndef _WIN32
-#ifndef SKIP_IOCTL
+#if !defined(_WIN32) && !defined(SKIP_IOCTL)
 
 #include <sys/fcntl.h>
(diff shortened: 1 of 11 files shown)


src/runtime_src/xdp/profile/device/ioctl_monitors/ioctl_add.cpp
```diff
@@ -16,6 +17,7 @@
  */
 
 #ifndef _WIN32
+#ifndef SKIP_IOCTL
```
stsoe, line 20: Isn't this the same as:
```
#if !defined(_WIN32) && !defined(SKIP_IOCTL)
...
#endif
```

stsoe (approved):
One comment

IshitaGhosh (approved):
Changes look good

src/runtime_src/xdp/profile/device/ioctl_monitors/ioctl_add.cpp
jvillarre: You are correct, and I have modified the code in the last commit to clean up all these locations with the single line.

PR #6978, opened 2022-08-30
CR-1074663 Provide help info when xbmgmt cannot see xclmgmt

rchane:
https://jira.xilinx.com/browse/CR-1074663 xbmgmt only says version is unknown when it cannot see xclmgmt
This was first discovered after updating from legacy versions of XRT commands.
Suggestions were added in a warning for users when xclmgmt version is unknown, as discussed in the CR comments. 
This affects output for `xbmgmt examine` and `xbutil examine` when xclmgmt version is unknown.
Output for `xbmgmt examine` when xclmgmt version is unknown:
```
rchane@xsjsagarw50:/proj/rdi/staff/rchane/XRT$ xbmgmt examine
System Configuration
  OS Name              : Linux
  Release              : 5.4.0-100-generic
  Version              : #113-Ubuntu SMP Thu Feb 3 18:43:29 UTC 2022
  Machine              : x86_64
  CPU Cores            : 16
  Memory               : 46911 MB
  Distribution         : Ubuntu 20.04.1 LTS
  GLIBC                : 2.31
  Model                : Super Server

XRT
  Version              : 2.14.0
  Branch               : CR-1074663
  Hash                 : 9889e1f8ca1163102603b1977002a97df3827a29
  Hash Date            : 2022-08-30 11:24:30
  XOCL                 : 2.14.239, 4929243eac42cf6b9ced7b33b8f6aa431a4ca663
  XCLMGMT              : unknown, unknown
WARNING: xclmgmt version is unknown. Is xclmgmt driver loaded? Or is MSD/MPD running?

Devices present
  0 devices found
BDF  :  Shell  Platform UUID  Device ID  Device Ready*  

* Devices that are not ready will have reduced functionality when using XRT tools
```
N/A


src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -262,12 +262,10 @@ XBUtilities::xrt_xbutil_version_cmp(bool force)
...
     if (drv_name.compare("xocl") == 0 && drv_version.compare("unknown") != 0 && xrt_version.compare(drv_version) != 0) {
       const auto & errMsg = boost::str(boost::format("Unexpected XRT Driver version (%s) was found. Expected %s, to match XRT tools.") % drv_version % xrt_version);
-      if (force) {
-        std::cout << errMsg << std::endl;
-        std::cout << "Bypassing mismatch with --force." << std::endl;
-      } else {
-        throw std::runtime_error(errMsg);
-      }
+      if (force)
+        std::cout << boost::format("%s\nBypassing mismatch with --force.") % errMsg << std::endl;
```
dbenusov-xilinx, line 266: Using `--force` to bypass will have secondary effects for certain commands. For instance, `xbmgmt program --force` causes all permission requests to be selected as yes. Would it be possible to use a add a new flag? If we do not want to add a new flag, is this a good trade off?

src/runtime_src/core/tools/common/ReportHost.cpp
```diff
@@ -99,6 +99,8 @@ ReportHost::writeReport(const xrt_core::device* /*_pDevice*/,
       boost::algorithm::to_upper(drv_name);
       _output << boost::format("  %-20s : %s, %s\n") % drv_name
           % driver.get<std::string>("version", "N/A") % driver.get<std::string>("hash", "N/A");
+      if (boost::iequals(drv_name, "xclmgmt") && boost::iequals(driver.get<std::string>("version", "N/A"), "unknown"))
+        _output << "WARNING: xclmgmt version is unknown. Is xclmgmt driver loaded? Or is MSD/MPD running?" << std::endl;
```
dbenusov-xilinx, line 103: Why is detection of the fault in xbmgmt a warning and in xbutil an error? I think it would be better if both tools generated a warning. That would get rid of the need to use the `--force` flag.

dbenusov-xilinx (changes requested):

src/runtime_src/core/tools/common/ReportHost.cpp
rchane: Both xbutil and xbmgmt show "WARNING: xclmgmt version is unknown [...]" underneath xclmgmt details, and the --force flag doesn't actually apply to this warning.

src/runtime_src/core/tools/common/XBUtilities.cpp
rchane: I see! @chvamshi-xilinx, do you believe we should continue using the --force flag here?

rozumx: @dbenusov-xilinx it sounds like you believe that `--force` operation definition only applies to user conditional inputs.   It is not.   --force was mention to be an option to force the code to procedure forward across issues like this.  So I do believe that @rchane is using it correctly.

As for adding any new flags for messages like this, we should discuss this further to understand all of the use-cases.

rozumx (approved):

dayeh-xilinx:
retest this please.

chvamshi-xilinx (approved):

chvamshi-xilinx:
Hi @dbenusov-xilinx , 
Are you expecting more changes to be required in this PR. If not, please approve sothat we can merge this change

dbenusov-xilinx (approved):

dbenusov-xilinx (approved):

PR #6982, opened 2022-09-01
CR-1132437 ASTeR TC22.06 - x3522/u280 - host mem enable segmentation fault

mamin506:
CR-1132437, segmentation fault when enable host mem  

I believe this bug is introduced since #5747. In the CR, the reporter has a script to keep enable and disable host mem. The segmentation fault issue will happen by chance.

The issue is cma_info.user_addr[i] is uninitialized in "munmap((void*)cma_info.user_addr[i], page_sz);".
In most of the cases, munmap() returns error. But sometimes, munmap() can destroy the stack frame. I guess this is because of that user_addr is allocated by alloca(). Maybe the uninitialized value is related to stack frame. BTW, I tried simply replace alloca() by malloc(). Then I don't see segmentation fault.

Base on what I have found, I make two changs:
1.  In `man alloca`, it says _"If the allocation causes stack overflow, program behavior is undefined."_. I thought that unless there is a strong reason, we should avoid use alloca(). So, I remove it. 
2. I use std::vector<uint64_t> for user_addr and initialized it.

Low

On u280, keep enable and disable host mem in a infinit loop. It runs 2 hours without any issue. In before, it can be reproduce in less than 2 minutes.

No

### DIFF
--- a/src/runtime_src/core/pcie/linux/shim.cpp
+++ b/src/runtime_src/core/pcie/linux/shim.cpp
@@ -1216,6 +1216,7 @@ int shim::cmaEnable(bool enable, uint64_t size)
         uint64_t allocated_size = 0;
         uint32_t page_num = size >> 30;
         drm_xocl_alloc_cma_info cma_info = {0};
+        uint64_t user_addr[page_num] = {0};
 
         /* We check the sysfs node host_mem_size first before going forward
          * If the same size of host memory chunk is allocated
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1216,6 +1216,7 @@ int shim::cmaEnable(bool enable, uint64_t size)
         uint64_t allocated_size = 0;
         uint32_t page_num = size >> 30;
         drm_xocl_alloc_cma_info cma_info = {0};
+        uint64_t user_addr[page_num] = {0};
```
stsoe, line 1219: This is not standard, please use std::vector

stsoe (changes requested):

src/runtime_src/core/pcie/linux/shim.cpp
mamin506: I forgot this rule. std::vector is even better. Thanks for pointing it out. I will fix this.

keryell:
It is not that arrays are not standard.
C99 VLA are a GNU extension in C++ and are actually optional in C since C11.

maxzhen (approved):

chienwei-lan (approved):

stsoe (approved):
The change is simple, the function is complicated.  Too many indents, too many braces, just hard to follow.  Anyway, this is not related to the PR.

PR #6908, opened 2022-08-11
Support for newest kernels and GCC

goekce:
#### Problem solved by the commit

1. realloc error handling
2. GCC 12.1.1-related warnings
   1. nullptr->member accesses 
   2. `new`-`delete` type mismatches
3. Rename dma-buf-map to iosys-map in the driver code

Additional details can be found in the commit messages.

##### Additional info on `realloc`

realloc() returns a NULL if reallocation was not successful. This was not handled properly (by returning `new_size` after conditional execution `free(temp)`.

More info on `realloc`:

- [`realloc` from C17 standard](https://web.archive.org/web/20181230041359if_/http://www.open-std.org/jtc1/sc22/wg14/www/abq/c17_updated_proposed_fdis.pdf#subsection.7.22.3)
- [SO: How to handle realloc if it fails](https://stackoverflow.com/a/1986572)

#### How problem was solved, alternative solutions (if any) and why they were rejected

N/A

#### Risks (if any) associated the changes in the commit

I do not know.

#### What has been tested and how, request additional testing if necessary

<s>I did not carry any tests. I request additional testing.</s>. I tested the changes on on U280 using `xbutil validate` using tests 1 to 7.

#### Documentation impact (if any)

I do not know.

PR #6983, opened 2022-09-01
[split #6908 1/3] Fix realloc error handling

goekce:
I split my PR #6908 into three which was already partly reviewed. This is part 1/3.

### DIFF
--- a/src/runtime_src/core/edge/sw_em/generic_pcie_hal2/shim.cxx
+++ b/src/runtime_src/core/edge/sw_em/generic_pcie_hal2/shim.cxx
@@ -214,10 +214,15 @@ namespace xclcpuemhal2 {
     }
     if (buf_size < new_size)
     {
-      void *temp = buf;
-      buf = (void*) realloc(buf,new_size);
-      if (!buf) // prevent leak of original buf
-        free(temp);
+      void *result = realloc(buf, new_size);
+      // If realloc was unsuccessful, then give up and deallocate.
+      if (result == nullptr)
+      {
+        free(buf);
+        buf = nullptr;
+        return 0;
+      }
+      buf = result;
       return new_size;
     }
     return buf_size;
--- a/src/runtime_src/core/edge/sw_em/generic_pcie_hal2/shim.cxx
+++ b/src/runtime_src/core/edge/sw_em/generic_pcie_hal2/shim.cxx
@@ -216,7 +216,7 @@ namespace xclcpuemhal2 {
     {
       void *result = realloc(buf, new_size);
       // If realloc was unsuccessful, then give up and deallocate.
-      if (result == nullptr)
+      if (!result)
       {
         free(buf);
         buf = nullptr;
--- a/src/runtime_src/core/pcie/emulation/cpu_em/generic_pcie_hal2/shim.cxx
+++ b/src/runtime_src/core/pcie/emulation/cpu_em/generic_pcie_hal2/shim.cxx
@@ -119,7 +119,7 @@ namespace xclcpuemhal2
         free(buf);
         buf = nullptr;
         return 0;
-	  }
+      }
       buf = result;
       return new_size;
     }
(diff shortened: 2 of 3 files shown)


src/runtime_src/core/edge/sw_em/generic_pcie_hal2/shim.cxx
```diff
@@ -214,10 +214,15 @@ namespace xclcpuemhal2 {
     }
     if (buf_size < new_size)
     {
-      void *temp = buf;
-      buf = (void*) realloc(buf,new_size);
-      if (!buf) // prevent leak of original buf
-        free(temp);
+      void *result = realloc(buf, new_size);
+      // If realloc was unsuccessful, then give up and deallocate.
+      if (result == nullptr)
```
stsoe, line 219:
```suggestion
      if (!result)
```
Checking against nullptr is just silly and C++ or C style at all.

src/runtime_src/core/pcie/emulation/cpu_em/generic_pcie_hal2/shim.cxx
```diff
@@ -112,10 +112,15 @@ namespace xclcpuemhal2
     }
     if (buf_size < new_size)
     {
-      void *temp = buf;
-      buf = (void*)realloc(temp, new_size);
-      if (!buf) // prevent leak of original buf
-        free(temp);
+      void *result = realloc(buf, new_size);
+      // If realloc was unsuccessful, then give up and deallocate.
+      if (result == nullptr)
```
stsoe, line 117: ditto

src/runtime_src/core/pcie/emulation/hw_em/alveo_shim/shim.cxx
```diff
@@ -154,10 +154,15 @@ namespace xclhwemhal2 {
     }
     if (buf_size < new_size)
     {
-      void *temp = buf;
-      buf = (void*) realloc(temp,new_size);
-      if (!buf) // prevent leak of original buf
-        free(temp);
+      void *result = realloc(buf, new_size);
+      // If realloc was unsuccessful, then give up and deallocate.
+      if (result == nullptr)
```
stsoe, line 159: ditto

stsoe (changes requested):

src/runtime_src/core/edge/sw_em/generic_pcie_hal2/shim.cxx
goekce: AFAIK `nullptr` was introduced in C++11 to distinguish it from arithmetic zero (as in `NULL`). This avoids problems when we have two overloaded functions where one function has an `int` as an argument and the other a pointer as an argument. I thought it is better to be more clear in the if statement. But [I read that `nullptr` evaluates to `false` in a boolean context](https://stackoverflow.com/a/57527724), so your solution is shorter. Thanks!

Can you elaborate on why is this silly? I am just curious.

stsoe: Sure it's not really a big deal.   Absolutely nullptr in place of NULL of course.

But checking against nullptr is redundant and doesn't make the code more readable, IMO.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-if

goekce:
Note that there are further expressive checks against `nullptr` in these files which seem to me as obvious null pointer checks.

stsoe:
> Note that there are further expressive checks against `nullptr` in these files which seem to me as obvious null pointer checks.

Yes, the code is not perfect for sure :-(  I just happened to see this change.  Thanks!

src/runtime_src/core/edge/sw_em/generic_pcie_hal2/shim.cxx
goekce: After reading the guideline your comment makes more sense, thanks @stsoe

goekce:
I am not very-versed with the PR reviews in Github. I have seen that Github still requires a review from @stsoe even I implemented the feedback so I rerequested a review.

stsoe (approved):
Thank you.  I added back @chvamshi-xilinx and @venkatp-xilinx as code owners.
Please sign-off your changes, per[ ](https://github.com/Xilinx/XRT/blob/master/CONTRIBUTING.rst)
Thank you for your efforts, much appreciated.

chvamshi-xilinx (approved):

venkatp-xilinx (approved):

stsoe:
ok to test

PR #6984, opened 2022-09-01
[split #6908 2/3] Fix modern gcc warnings

goekce:
I split my PR #6908 into three which was already partly reviewed. This is part 2/3.

### DIFF
--- a/src/runtime_src/xocl/api/clReleaseCommandQueue.cpp
+++ b/src/runtime_src/xocl/api/clReleaseCommandQueue.cpp
@@ -37,6 +37,13 @@ static cl_int
 clReleaseCommandQueue(cl_command_queue command_queue)
 {
   validOrError(command_queue);
+  // validOrError checks for nullptr only in the debug build. Suppress
+  // collateral GCC warnings (e.g. stringop-overflow in smart pointer methods)
+  // in case of a non-debug build.
+  #ifdef NDEBUG
+  if (!command_queue) __builtin_unreachable();
+  #endif
+
   if (xocl(command_queue)->release())
     delete xocl(command_queue);
   return CL_SUCCESS;
--- a/src/runtime_src/xocl/api/clReleaseCommandQueue.cpp
+++ b/src/runtime_src/xocl/api/clReleaseCommandQueue.cpp
@@ -37,7 +37,7 @@ static cl_int
 clReleaseCommandQueue(cl_command_queue command_queue)
 {
   validOrError(command_queue);
-  if (xocl(command_queue) != nullptr && xocl(command_queue)->release())
+  if (xocl(command_queue)->release())
     delete xocl(command_queue);
   return CL_SUCCESS;
 }
--- a/src/runtime_src/xocl/api/clReleaseCommandQueue.cpp
+++ b/src/runtime_src/xocl/api/clReleaseCommandQueue.cpp
@@ -37,7 +37,7 @@ static cl_int
 clReleaseCommandQueue(cl_command_queue command_queue)
 {
   validOrError(command_queue);
-  if (xocl(command_queue)->release())
+  if (xocl(command_queue) != nullptr && xocl(command_queue)->release())
     delete xocl(command_queue);
   return CL_SUCCESS;
 }
--- a/src/runtime_src/xocl/api/clReleaseContext.cpp
+++ b/src/runtime_src/xocl/api/clReleaseContext.cpp
@@ -36,7 +36,7 @@ static cl_int
 clReleaseContext(cl_context  context )
 {
   validOrError(context);
-  if (xocl(context)->release())
+  if (xocl(context) != nullptr && xocl(context)->release())
     delete xocl(context);
   return CL_SUCCESS;
 }
--- a/src/runtime_src/xocl/api/clReleaseContext.cpp
+++ b/src/runtime_src/xocl/api/clReleaseContext.cpp
@@ -36,7 +36,7 @@ static cl_int
 clReleaseContext(cl_context  context )
 {
   validOrError(context);
-  if (xocl(context) != nullptr && xocl(context)->release())
+  if (xocl(context)->release())
     delete xocl(context);
   return CL_SUCCESS;
 }
--- a/src/runtime_src/xocl/api/clReleaseContext.cpp
+++ b/src/runtime_src/xocl/api/clReleaseContext.cpp
@@ -36,6 +36,13 @@ static cl_int
 clReleaseContext(cl_context  context )
 {
   validOrError(context);
+  // validOrError checks for nullptr only in the debug build. Suppress
+  // collateral GCC warnings (e.g. stringop-overflow in smart pointer methods)
+  // in case of a non-debug build.
+  #ifdef NDEBUG
+  if (!context) __builtin_unreachable();
+  #endif
+
   if (xocl(context)->release())
     delete xocl(context);
   return CL_SUCCESS;
--- a/src/runtime_src/xocl/api/clReleaseKernel.cpp
+++ b/src/runtime_src/xocl/api/clReleaseKernel.cpp
@@ -37,7 +37,7 @@ static cl_int
 clReleaseKernel(cl_kernel kernel)
 {
   validOrError(kernel);
-  if (xocl(kernel) != nullptr && xocl(kernel)->release())
+  if (xocl(kernel)->release())
     delete xocl(kernel);
   return CL_SUCCESS;
 }
--- a/src/runtime_src/xocl/api/clReleaseKernel.cpp
+++ b/src/runtime_src/xocl/api/clReleaseKernel.cpp
@@ -37,6 +37,13 @@ static cl_int
 clReleaseKernel(cl_kernel kernel)
 {
   validOrError(kernel);
+  // validOrError checks for nullptr only in the debug build. Suppress
+  // collateral GCC warnings (e.g. stringop-overflow in smart pointer methods)
+  // in case of a non-debug build.
+  #ifdef NDEBUG
+  if (!kernel) __builtin_unreachable();
+  #endif
+
   if (xocl(kernel)->release())
     delete xocl(kernel);
   return CL_SUCCESS;
--- a/src/runtime_src/xocl/api/clReleaseKernel.cpp
+++ b/src/runtime_src/xocl/api/clReleaseKernel.cpp
@@ -37,7 +37,7 @@ static cl_int
 clReleaseKernel(cl_kernel kernel)
 {
   validOrError(kernel);
-  if (xocl(kernel)->release())
+  if (xocl(kernel) != nullptr && xocl(kernel)->release())
     delete xocl(kernel);
   return CL_SUCCESS;
 }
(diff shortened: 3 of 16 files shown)


src/runtime_src/xocl/api/clReleaseCommandQueue.cpp
```diff
@@ -37,7 +37,7 @@ static cl_int
 clReleaseCommandQueue(cl_command_queue command_queue)
 {
   validOrError(command_queue);
-  if (xocl(command_queue)->release())
+  if (xocl(command_queue) != nullptr && xocl(command_queue)->release())
```
stsoe, line 40: validOrError checks this, for better or worse it is an OpenCL implementation idiom to not have error checks outside of validOrError.  Please revert.

src/runtime_src/xocl/api/clReleaseContext.cpp
```diff
@@ -36,7 +36,7 @@ static cl_int
 clReleaseContext(cl_context  context )
 {
   validOrError(context);
-  if (xocl(context)->release())
+  if (xocl(context) != nullptr && xocl(context)->release())
```
stsoe, line 39: ditto

src/runtime_src/xocl/api/clReleaseKernel.cpp
```diff
@@ -37,7 +37,7 @@ static cl_int
 clReleaseKernel(cl_kernel kernel)
 {
   validOrError(kernel);
-  if (xocl(kernel)->release())
+  if (xocl(kernel) != nullptr && xocl(kernel)->release())
```
stsoe, line 40: ditto

src/runtime_src/xocl/api/clReleaseMemObject.cpp
```diff
@@ -38,8 +38,7 @@ static cl_int
 clReleaseMemObject(cl_mem memobj)
 {
   validOrError(memobj);
-
-  if (!xocl(memobj)->release())
+  if (xocl(memobj) != nullptr && !xocl(memobj)->release())
```
stsoe, line 41: ditto

src/runtime_src/xocl/api/clReleaseProgram.cpp
```diff
@@ -36,8 +36,7 @@ static cl_int
 clReleaseProgram(cl_program program)
 {
   validOrError(program);
-
-  if (xocl::xocl(program)->release())
+  if (xocl::xocl(program) != nullptr && xocl::xocl(program)->release())
```
stsoe, line 39: ditto

src/runtime_src/xocl/api/clReleaseSampler.cpp
```diff
@@ -36,7 +36,7 @@ static cl_int
 clReleaseSampler(cl_sampler sampler)
 {
   validOrError(sampler);
-  if (xocl(sampler)->release())
+  if (xocl(sampler) != nullptr && xocl(sampler)->release())
```
stsoe, line 39: ditto

stsoe (changes requested):
The effort here is good, but sadly most, if not all instances, of `std::shared_ptr<T[]>` should be replaced with `std::vector<T>`.   The changes in this PR clearly illustrates that C++ has a flaw in that it should never have had shared_ptr support `T[]`, because developers forget.

keryell:
So is it a false positive from g++ 12.1 ?

stsoe:
> So, is it a false positive from g++ 12.1 ?

It's not a false positive that the ptr can be null if the OpenCL checks are disabled.  However, default behavior is to validate all OpenCL APIs calls per OpenCL spec and this includes checking for valid objects.     If checks are disabled and objects are null, then the program is in error and a segmentation fault is warranted, hiding it behind a conditional and returning CL_SUCESSS is definitely not correct.

goekce:
Soren is right, the warnings do not happen in Debug build, but in Release. I give a hint to the compiler only in case of a non-debug build now.

<details><summary>FYI the warning that I was getting is here</summary>
<code>
<pre>

In file included from /usr/include/c++/12.2.0/bits/shared_ptr_atomic.h:33,
                 from /usr/include/c++/12.2.0/memory:78,
                 from /usr/include/boost/property_tree/ptree_fwd.hpp:18,
                 from /git-repos/XRT/src/runtime_src/core/common/config_reader.h:24,
                 from /git-repos/XRT/src/runtime_src/xrt/util/config_reader.h:20,
                 from /git-repos/XRT/src/runtime_src/xrt/config.h:20,
                 from /git-repos/XRT/src/runtime_src/xocl/config.h:40,
                 from /git-repos/XRT/src/runtime_src/xocl/api/clReleaseCommandQueue.cpp:18:
In member function ‘std::__atomic_base<_IntTp>::__int_type std::__atomic_base<_IntTp>::operator--() [with _ITp = unsigned int]’,
    inlined from ‘bool xocl::refcount::release()’ at /git-repos/XRT/src/runtime_src/xocl/core/refcount.h:65:13,
    inlined from ‘cl_int xocl::clReleaseCommandQueue(cl_command_queue)’ at /git-repos/XRT/src/runtime_src/xocl/api/clReleaseCommandQueue.cpp:40:35,
    inlined from ‘cl_int clReleaseCommandQueue(cl_command_queue)’ at /git-repos/XRT/src/runtime_src/xocl/api/clReleaseCommandQueue.cpp:53:39:
/usr/include/c++/12.2.0/bits/atomic_base.h:393:34: error: ‘unsigned int __atomic_sub_fetch_4(volatile void*, unsigned int, int)’ writing 4 bytes into a region of size 0 overflows the destination [-Werror=stringop-overflow=]
  393 |       { return __atomic_sub_fetch(&_M_i, 1, int(memory_order_seq_cst)); }
      |                ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
</pre></code>
</details>

goekce:
> The effort here is good, but sadly most, if not all instances, of `std::shared_ptr<T[]>` should be replaced with `std::vector<T>`. The changes in this PR clearly illustrates that C++ has a flaw in that it should never have had shared_ptr support `T[]`, because developers forget.

According to [this comment](https://stackoverflow.com/questions/8947579/why-isnt-there-a-stdshared-ptrt-specialisation#comment11204653_8947700) `shared_ptr<T[]>` can lead to less overhead compared to `vector<T>`. So there is a rationale to use them.

stsoe:
> > The effort here is good, but sadly most, if not all instances, of `std::shared_ptr<T[]>` should be replaced with `std::vector<T>`. The changes in this PR clearly illustrates that C++ has a flaw in that it should never have had shared_ptr support `T[]`, because developers forget.

Problem is developers nearly always forget `[]` so I wish it wasn't used.  But I hear you on the very slight overhead.

goekce:
> Problem is developers nearly always forget `[]` so I wish it wasn't used 

Makes sense. However that should be part of the coding style.

For what it's worth, modern GCC integrates many checks regarding pointers, so the memory management errors could be mitigated easier.

I will leave my patch untouched then.

src/runtime_src/xocl/api/clReleaseCommandQueue.cpp
```diff
@@ -37,6 +37,13 @@ static cl_int
 clReleaseCommandQueue(cl_command_queue command_queue)
 {
   validOrError(command_queue);
+  // validOrError checks for nullptr only in the debug build. Suppress
+  // collateral GCC warnings (e.g. stringop-overflow in smart pointer methods)
+  // in case of a non-debug build.
+  #ifdef NDEBUG
+  if (!command_queue) __builtin_unreachable();
+  #endif
```
stsoe, line 45: I don't really like this conditional code.  Prefer to disable the stringop-overflow in the file itself.   It's just not very useful code checking for a condition that has already been checked even in optimized compile, .e.g. validOrError is always enabled, and users have to jump through hoops to disable it.

stsoe (changes requested):

rozumx (approved):
**_Pre-Approved_**

With regards to the PR, I concur with all of the comments / change requests made by Soren and have no other ones to add.

Instead of waiting for Soren to to approve the PR prior to my approval, I am pre-approving the PR on the condition that Soren's requests are satisfied.

dayeh-xilinx:
retest this please. Jenkins server restart.

src/runtime_src/xocl/api/clReleaseCommandQueue.cpp
goekce: >  Prefer to disable the stringop-overflow in the file itself

If disabled, then gcc cannot check for stringop-overflow in debug (i.e., non-optimized) build.

>  It's just not very useful code checking for a condition that has already been checked even in optimized compile, .e.g. validOrError is always enabled, and users have to jump through hoops to disable it.

As I mentioned in [my previous comment](https://github.com/Xilinx/XRT/pull/6984#issuecomment-1235504053), nullptr check (probably `validOrError`) does not seem to be active in the release (i.e., optimized) build. Please correct me if I am wrong.

So I still don't see the advantage of disabling the warning on the file level. The behavior of the code seem to differ between release and debug builds.

goekce:
@stsoe just a friendly reminder that you did not respond to my comment above in your review, in case you thought was waiting on me.

src/runtime_src/xocl/api/clReleaseCommandQueue.cpp
stsoe: Thanks @goekce for pinging me.

Definitely validOrError is active in release builds. There is config run-time option to disable validOrError.  I do not understand why gcc would complain?  IMO it smells like an compiler optimization bug (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=104475). 

I am able to reproduce the gcc warning with a small stand-alone program, but it really just confirms that optimization is wrong.

I understand the work-around (it is mentioned in the gcc bug report also), but it will not work as is since XRT is not for GCC only.   Also we would have to fix the comment given that validOrError does indeed check for nullptr both in DEBUG and NDEBUG, but we are facing a compiler bug.

Let's keep the conditional macro code in one file.  I suggest adding `xocl_or_error()` in `src/runtime_src/xocl/core/object.h` and replace the `xocl(...)` calls in the `ZZZRelease.cpp` files with `xocl_or_error(...)` when we want to enforce undefined behavior when the cl object is null.

```
template <typename CLTYPE>
typename detail::cl_object_traits<CLTYPE>::xocl_type*
xocl_or_error(CLTYPE c)
{
  auto x = detail::cl_object_traits<CLTYPE>::get_xocl(c);
#ifdef __GNUC__
  if (!x)
    __builtin_unreachable();
#endif
  return x;
}

```

src/runtime_src/xocl/api/clReleaseContext.cpp
```diff
@@ -36,6 +36,13 @@ static cl_int
 clReleaseContext(cl_context  context )
 {
   validOrError(context);
+  // validOrError checks for nullptr only in the debug build. Suppress
+  // collateral GCC warnings (e.g. stringop-overflow in smart pointer methods)
+  // in case of a non-debug build.
+  #ifdef NDEBUG
+  if (!context) __builtin_unreachable();
```
keryell, line 43: Why all these weird code?
```suggestion
  if (!context) std::terminate();
```

stsoe:
@goekce The warning aspect of this PR was fixed in 7302.  If any other issues, please resubmit.  Thanks.

PR #6991, opened 2022-09-01
Generate PDB files in release mode on Windows

stsoe:
Enable generation of PDB file in release mode and consolidate all PDB
files into single PDB_OUTPUT_DIRECTORY for easier distribution.

TBD if PDB files are to be released to field and if so, how.


dayeh-xilinx:
retest this please.

src/CMake/nativeWin.cmake
```diff
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright (C) 2019-2022 Xilinx, Inc. All rights reserved.
+# Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
```
jhoyamd, line 3: Please remove AMD copyright.

stsoe: No, please see separate email :-)

PR #7005, opened 2022-09-08
CR-1140025 warn if writing more than 10mb aie trace data

hackwa:
application seems to hang if amount of trace is large
NA
tested on txchain


dayeh-xilinx:
retest this please.

jvillarre (approved):

PR #7009, opened 2022-09-08
Update skd.service to automatically restart on failure

jeffli-xilinx:
skd not running after petalinux boot in some cases.

Noticed sometimes skd do not start after petalinux boot

Updated skd.service to restart skd on exit

None

vck5000 / v70

None

### DIFF
--- a/src/runtime_src/core/edge/skd/main.cpp
+++ b/src/runtime_src/core/edge/skd/main.cpp
@@ -70,6 +70,12 @@ int main(int argc, char *argv[])
 
   handle = initXRTHandle(0);
   if (!handle) {
+    syslog(LOG_INFO, "Fail to init XRT first time\n");
+  }
+  // Retry XRT init after delay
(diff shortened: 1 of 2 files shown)


rozumx (changes requested):
With regards to the Daemon starting up with the APU makes sense and this should always occur.

As for restarting the SDK if it should crash / fail, I understand why you would like to do this, but I don't think this is a good idea.  Let me explain further.

The SDK code base needs to perform without ANY issues.  If there is an issue, we need to know about it and address it.   This change in behavior will "mask" SDK issues as users PS Kernel issues. 
 This will resulting in days of lost productive and possible SDK issues not being reported.

At our next opportunity lets to discuss why we should enable this service.

rozumx (approved):
Thank you not automatically restarting the daemon.

dayeh-xilinx:
retest this please.

PR #7023, opened 2022-09-12
CR1132359 compute board power using existing voltage & current sensor readings

rajkumar-xilinx:

At present, driver issues power SDR request to VMR firmware for board power data. But, the computed board power using existing voltage & current sensor readings doesn't match with the received power data from VMR firmware.
So, to fix this problem, measuring the board power using existing readings of voltage & current sensor values.
CR1132359
Added changes to measure board power using existing readings of voltage & current sensor values.
NA
xbutil examine reports.
NA


dayeh-xilinx:
retest this please.

chvamshi-xilinx:
Hi @rajkumar-xilinx ,
As discussed offline, Please close this PR. We shouldnt be hardcoding the sensor values in the userspace for Versal class devices.

PR #7027, opened 2022-09-13
CR-1140644 missing xgq_vmr_log_dump_debug for some failure scenarios

xdavidz:

The current xgq only dump VMR logs for xclbin and pdi download failures. but missing for any other VMR return code.
The contract between host and VMR is that any non-zero return code is an error, so we should dump the debug logs whenever we get non-zero return code.

CR-1140644

dump the log when see rcode != 0

N/A

tested on v70 vmr

N/A


maxzhen (approved):

dayeh-xilinx:
retest this please.

PR #7028, opened 2022-09-13
Fix for CR-1139265: xbmgmt incorrect BDF causes seg fault

vboggara-xilinx:
Fix for CR-1139265: XRT - Running 'xbmgmt examine' with incorrect BDF causes seg fault
CR-1139265
added null check for get() in bdf struct
Low
Ran bugcase locally with this change, testcase passed.

### DIFF
--- a/src/runtime_src/core/pcie/linux/device_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/device_linux.cpp
@@ -118,7 +118,7 @@ struct bdf
   static result_type
   get(const xrt_core::device* device, key_type)
   {
-    auto pdev = get_pcidev(device);    
+    auto pdev = get_pcidev(device);
     return std::make_tuple(pdev->domain, pdev->bus, pdev->dev, pdev->func);
   }
 };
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -116,6 +116,8 @@ struct bdf
   get(const xrt_core::device* device, key_type)
   {
     auto pdev = get_pcidev(device);
+    if (!pdev)
+      throw xrt_core::error("Invalid device handle");
```
stsoe, line 120: Wouldn't it make more sense to nave get_pcidev throw?  There are several other calls to get_pcidev.

stsoe (changes requested):

src/runtime_src/core/pcie/linux/device_linux.cpp
stsoe: Also, please use `git commit -s ...` to sign the commit.

vboggara-xilinx: sure soren.

stsoe (approved):

chvamshi-xilinx (approved):

chvamshi-xilinx:
retest this please

PR #7049, opened 2022-09-16
Fixed CR-1137646 : zocl_create_client() panic for missing NULL pointer checking

saifuddin-xilinx:
Fixed this issue. The actual issue is that there is no null pointer checking for zdev in zocl_create_client() of zocl driver.
Added the same.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/scu.c
+++ b/src/runtime_src/core/edge/drm/zocl/scu.c
@@ -174,9 +174,11 @@ static const struct attribute_group scu_attrgroup = {
 static int configure_soft_kernel(u32 cuidx, char kname[64], unsigned char uuid[16])
 {
 	struct drm_zocl_dev *zdev = zocl_get_zdev();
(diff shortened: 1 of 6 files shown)


src/runtime_src/core/edge/drm/zocl/scu.c
```diff
@@ -230,6 +233,7 @@ static int scu_probe(struct platform_device *pdev)
 	memcpy(&zcu->base.info, info, sizeof(struct xrt_cu_info));
 
 	zdev = zocl_get_zdev();
+	BUG_ON(!zdev);
```
mamin506, line 236: Why use if statement in cu.c but BUG_ON here?

mamin506 (approved):
Overall this looks good to me.

mamin506:
Add @jeffli-xilinx  seems this changed soft kernel related code.

jeffli-xilinx (approved):

PR #7053, opened 2022-09-16
Enable /sdl compiler option on Windows

stsoe:
Per request enable
https://docs.microsoft.com/en-us/cpp/build/reference/sdl-enable-additional-security-checks?view=msvc-170


stsoe:
retest this please - petalinux does not even include the change makefile

dayeh-xilinx:
retest this please.

rozumx (approved):
Being that warnings are treated as errors, I do believe our current settings cover these issues.

That said, I have no issues with adding additional (or even redundant) checks like this.

PR #7055, opened 2022-09-19
Fix RC when accessing singleton's sessions vector

Qizot:
There are 2 threads `xma_thread1` and `xma_thread2` reading from a vector of sessions inside the `g_xma_singleton`. The access was not guarded by any mutex while other threads could access the sessions vector and push a new element, this sometimes was causing a race condition where the thread was looping over elements and a new elements was being added causing a vector resize. This automatically invalidated any existing vector iterator (which the threads were using for reading from vector) which resulted in a segfault when trying to dereference it.  

The bug has been discovered when running a lot of quite short-lived sessions which caused the application to exit due to
segfaults.
Locking a mutex before accessing a list of sessions.
Probably none, the mutex could be replaced with `std::shared_mutex` as there are probably more reads than writes but with that low frequency of locking it, it is probably not worth it.

### DIFF
--- a/src/xma/src/xmaapi/xmaapi.cpp
+++ b/src/xma/src/xmaapi/xmaapi.cpp
@@ -265,13 +265,13 @@ void xma_thread2(uint32_t hw_dev_index) {
     bool desired = true;
     auto xrt_device_obj = g_xma_singleton->hwcfg.devices[hw_dev_index].xrt_device;
     while (!g_xma_singleton->xma_exit) {
+        std::lock_guard<std::mutex> lock(g_xma_singleton->m_mutex);
         if (g_xma_singleton->cpu_mode == XMA_CPU_MODE2) {
             std::this_thread::sleep_for(std::chrono::milliseconds(3));
         } else {
             xrt_device_obj.get_handle()->exec_wait(100);
         }
 
-        std::lock_guard<std::mutex> lock(g_xma_singleton->m_mutex);
         for (auto& itr1: g_xma_singleton->all_sessions_vec) {
             if (g_xma_singleton->xma_exit) {
                 break;
--- a/src/xma/src/xmaapi/xmaapi.cpp
+++ b/src/xma/src/xmaapi/xmaapi.cpp
@@ -108,7 +108,9 @@ void xma_thread1() {
             //bool expected = false;
             //bool desired = true;
             XmaHwSessionPrivate *slowest_session = nullptr;
-            uint32_t sessioin_cmd_busiest_val = 0;
+            uint32_t session_cmd_busiest_val = 0;
+            std::lock_guard lock(g_xma_singleton->m_mutex);
+
             for (auto& itr1: g_xma_singleton->all_sessions_vec) {
                 if (g_xma_singleton->xma_exit) {
                     break;
@@ -129,8 +131,8 @@ void xma_thread1() {
                     continue;
                 }
                 if (priv1->kernel_complete_total > 127) {
-                    if (priv1->cmd_busy > sessioin_cmd_busiest_val) {
-                        sessioin_cmd_busiest_val = priv1->cmd_busy;
+                    if (priv1->cmd_busy > session_cmd_busiest_val) {
+                        session_cmd_busiest_val = priv1->cmd_busy;
                         slowest_session = priv1;
                     }
                 }
@@ -210,6 +212,8 @@ void xma_thread1() {
         }
     }
     //Print all stats here
+    std::lock_guard lock(g_xma_singleton->m_mutex);
+
     xclLogMsg(NULL, XRT_INFO, "XMA-Session-Stats", "=== Session CU Command Relative Stats: ===");
     for (auto& itr1: g_xma_singleton->all_sessions_vec) {
         xclLogMsg(NULL, XRT_INFO, "XMA-Session-Stats", "--------");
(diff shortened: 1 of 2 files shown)


vboggara-xilinx (approved):
hope XMA test suite is working with this change.

src/xma/src/xmaapi/xmaapi.cpp
```diff
@@ -267,6 +271,7 @@ void xma_thread2(uint32_t hw_dev_index) {
             xrt_device_obj.get_handle()->exec_wait(100);
         }
 
+        std::lock_guard<std::mutex> lock(g_xma_singleton->m_mutex);
```
dbenusov-xilinx, line 274: Would you want to protect the cpu_mode check above? It is not atomic. It may be worth just updating that variable to be atomic too.

Qizot: I'm not that familiar with the codebase and I'm not sure what the `cpu_mode` check really does. I guess it did not affect my issue.

dbenusov-xilinx: No worries, the reason I pointed it out is I see the mutex lock you added as a check on the data inside the singleton. Mainly you use it to defend the vector of XmaSessions which makes perfect sense.

However, in `struct XmaSingleton` (xmaapi.h:41) the cpu_mode is not defined as atomic.

```
typedef struct XmaSingleton
{
    XmaHwCfg          hwcfg;
    bool              xma_initialized;
    bool              kds_old;
    uint32_t          cpu_mode;
    std::mutex            m_mutex;
    std::atomic<uint32_t> num_decoders;
 ... More atomics ...
    std::atomic<uint32_t> num_of_sessions;
    std::vector<XmaSession> all_sessions_vec;// XMASessions
...
} XmaSingleton;
```

I believe we should use the mutex to defend access to all non-atomic variables of the struct.

Alternatively, you could make `cpu_mode` an atomic value which ensure its value in the check. And extending that change to all of the primitive types (except the `reserved` member variable) is a good idea since this is struct use between multiple threads.

Qizot: The variable is not atomic but it is not being changed in other threads, it is only set once in `xma_initialize` so the if statement here is secure, no potential race conditions.

vstreame:
@vboggara-xilinx This is causing segfaults for us and blocking us from using Xilinx in prod. Do you happen to know when this may be merged and then released?

Thanks!

dbenusov-xilinx (changes requested):

Qizot:
@dbenusov-xilinx I've moved the lock acquisition before the cpu_mode read as you wished. But when reading the comment of the `XMA_CPU_MODE2` flag being `#define XMA_CPU_MODE2           2  //High cpu load + high performance` I'm not sure if that is good idea to lock and then sleep for 3 milliseconds. Maybe it should lock, check the flag, unlock, sleep and lock afterwards?

dbenusov-xilinx:
That is a really good point I did not notice that, good catch. Try making the cpu_check an atomic variable and moving the lock_guard back to its original spot. Hopefully everything should work the same :)

dbenusov-xilinx (approved):
Thanks for the changes :)

dayeh-xilinx:
retest this please.

src/xma/src/xmaapi/xmaapi.cpp
```diff
@@ -210,6 +212,8 @@ void xma_thread1() {
         }
     }
     //Print all stats here
+    std::lock_guard lock(g_xma_singleton->m_mutex);
```
vboggara-xilinx, line 215: @Qizot, can we move this line to 104 line? 

if we move this to line number 104, lock_gaurd at line number 112 is not needed.

src/xma/src/xmaapi/xmaapi.cpp
vboggara-xilinx, line 274: can't we move this to before line number 267?

src/xma/src/xmaapi/xmaapi.cpp
Qizot: I think that it is best to keep it that way to keep the lock congestion low. If we unlock the mutex regularly we can let other threads do something with the singleton as well.

src/xma/src/xmaapi/xmaapi.cpp
Qizot: Same reasoning as above 👆

src/xma/src/xmaapi/xmaapi.cpp
vboggara-xilinx: ok.

vboggara-xilinx:
changes looks good to me. we can merge this change.

src/xma/src/xmaapi/xmaapi.cpp
keryell, line 274:
```suggestion
        std::lock_guard lock(g_xma_singleton->m_mutex);
```
In C++17 we trust.

Qizot: Why though? The rest of the code uses the lock guard like this, want to stay consistent.

vboggara-xilinx (approved):

Qizot:
Can this finally be merged if all tests are passing?

PR #7060, opened 2022-09-20
Multislot mm support

saifuddin-xilinx:
Memory manager changes for multislot support. This PR has following changes:
-> Introduce a memory manager.
   --> Memory manager initialization is done in two phases. memory manager init is done only once for a device.
         --> Phase 1 : In the drm init state, read the memory information from platform. This is done even before first xclbin is 
                               loaded. Here We read all the available banks and create a drm_mm_init for <start_addr - end_addr>.
                               Also allocated memory for all the stats. This phase is required if build in PS kernel is loading first which 
                               doesn't have memory information.
                               
             --> Phase 2 : This is done while first XCLBIN is loading, which has memory topology information. In this phase we also 
                               initialized p2p memory. Some platform doesn't have memory information. For that case Phase 1 is also 
                               done here. Here memory initialization is done based on memory topology.

-> xocl_init_mem is done everytime when a new xclbin is loaded. Now this is a link list which has only stats information and memory is allocated based on the start end address of that bank from memory manager. 

-> Removed wrapper data structure. I believe we don't need this.

image

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
@@ -54,6 +54,7 @@
 
 static char driver_date[9];
 
+int xocl_cleanup_mem_nolock(struct xocl_drm *drm_p);
 static int xocl_cleanup_memory_manager(struct xocl_drm *drm_p);
 static int xocl_init_drm_memory_manager(struct xocl_drm *drm_p);
 
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
@@ -925,10 +925,9 @@ static int xocl_cleanup_drm_memory_manager(struct xocl_mm *xocl_mm)
 	return 0;
 }
 
-static int xocl_init_drm_mm(struct xocl_drm *drm_p, uint64_t mm_start_addr,
-		uint64_t mm_end_addr, uint32_t m_count)
+static int xocl_init_drm_mm(struct xocl_drm *drm_p, struct xocl_mm *xocl_mm,
+		uint64_t mm_start_addr, uint64_t mm_end_addr, uint32_t m_count)
 {
-	struct xocl_mm *xocl_mm = drm_p->xocl_mm;
 	size_t size = 0;
 	int err = 0;
 	int i = 0;
@@ -987,7 +986,6 @@ static int xocl_init_drm_mm(struct xocl_drm *drm_p, uint64_t mm_start_addr,
 
 error:
 	xocl_cleanup_drm_memory_manager(xocl_mm);
-	drm_p->xocl_mm = NULL;
 	return err;
 }
 
@@ -1015,7 +1013,6 @@ static int xocl_init_drm_memory_manager(struct xocl_drm *drm_p)
                 return -ENOMEM;
 	}
 
-	drm_p->xocl_mm = xocl_mm;
         /* Initialize with max and min possible value */
         mm_start_addr = 0xffffFFFFffffFFFF;
         mm_end_addr = 0;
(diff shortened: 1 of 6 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
```diff
@@ -928,19 +900,316 @@ int xocl_cleanup_mem(struct xocl_drm *drm_p)
...
+	}
+
+	drm_mm_init(xocl_mm->mm, mm_start_addr, (mm_end_addr - mm_start_addr));
+	if (!xocl_mm->mm) {
+		xocl_err(drm_p->ddev->dev,
+				"init memory manager failed");
+		err = -EINVAL;
+		goto error;
+	}
+
```
mamin506, line 978: This is my last comment in the older PR.
"Line 966 already check if xocl_mm->mm is NULL. Why check this again? Do you mean drm_mm_init() will free and set xocl_mm->mm to NULL if initialized incorrectly?"

mamin506 (changes requested):
Just move a comment from PR#6958.
Please double check why mini-pipeline and XRT pipeline failed. Then I will review the rest of the code.

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
saifuddin-xilinx: We may not need to check again. drm_mem_init doesn't set anything and not return any error code. 
We should continue without any check.

saifuddin-xilinx:
retest this please

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
```diff
@@ -639,257 +659,209 @@ void xocl_drm_fini(struct xocl_drm *drm_p)
...
 					   start_addr, end_addr, 0);
 #endif
 
 	return ret;
 }
 
-int xocl_mm_insert_node(struct xocl_drm *drm_p, u32 mem_id,
-			struct drm_mm_node *node, u64 size)
+int xocl_mm_insert_node(struct xocl_drm *drm_p, unsigned memidx,
+			uint32_t slotidx, struct drm_mm_node *node, u64 size)
```
houlz0507, line 763: This is confusing. The slotidx is per xclbin, correct? It is never used with xocl_mm_insert_*. Is it possible that bo carries the incorrect slotidx which belongs to other xclbin?

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
```diff
@@ -639,257 +659,209 @@ void xocl_drm_fini(struct xocl_drm *drm_p)
...
                 return -EINVAL;
 
 	ret = XOCL_GET_GROUP_TOPOLOGY(drm_p->xdev, grp_topology);
         if (ret)
                 return 0;
 
-	if (grp_topology->m_mem_data[mem_id].m_type == MEM_PS_KERNEL) {
-		ret = xocl_mm_insert_node_range_all(drm_p, mem_id, 
+	if (grp_topology->m_mem_data[memidx].m_type == MEM_PS_KERNEL) {
+		ret = xocl_mm_insert_node_range_all(drm_p, &memidx,
```
houlz0507, line 778: Why is memidx changed inside insert? It does not look right.

maxzhen (approved):

chvamshi-xilinx (approved):

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
saifuddin-xilinx: The slot id is required for status update. Along with the device memory stats we are also maintaining the stats for each slot.

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
saifuddin-xilinx: We need to send this for PS kernel. For PS kernel we need to find the available memory bank to allocate memory. 
Hence we need to pass memidx to get final bank where the memory has been allocated

houlz0507: with line 777, it uses m_mem_data[memidx]. what if  778 returns different memidx?  Which one would be correct?

saifuddin-xilinx: As per PS kernel the memidx is dummy. 
             {
 20                 "m_type": "MEM_PS_KERNEL",
 21                 "m_used": "1",
 22                 "m_sizeKB": "0x0",
 23                 "m_tag": "MEM_PS_KERNEL",
 24                 "m_base_address": "0x0"
 25             }

It just tells that this is the request from a PS kernel. We need to create memory from actual BANKs. Hence memidx will be updated from where the actual memory has been allocated.

houlz0507: Please add comment on this.

houlz0507 (approved):

mamin506 (approved):

PR #7063, opened 2022-09-21
CR-1141332 and CR-1141338: Fix issues in Kernel to Global Memory data transfer table in profile summary

jvillarre:
The "Data Transfers: Kernel to Global Memory" table in the profile summary was reporting incorrect numbers for the number of transactions as well as sometimes not showing the argument associated with a set of transfers.  Both of these issues are fixed in this pull request.

The incorrect number of transactions displayed was due to the number of transactions being treated and output as a double value.  When millions of transactions were recorded, the value was output in exponential notation which was incorrectly parsed by post-processing tools and made it appear that only 1 transaction was recorded.  This was introduced earlier in some refactoring that was done and noticed when a stress test was run.

The issue with the missing argument was due to the code doing a string comparison to see if an argument was going to a memory resource.  The string comparison needed an exact match, so if it was compared to a range of memory resources it would fail even if the memory should have been contained in the range.  The argument missing issue was introduced when the argument matching was moved from being OpenCL specific to all applications.  It was found via internal stress testing.

The first issue was fixed by treating and printing the number of transactions as a uint64_t instead of a double, and explicitly casting it as a double when we need to.

The second issue was fixed by introducing a new function to check if a memory is either the same or is contained in a range.  For example, we now match that DDR[1] is part of DDR[0:2] and add the argument to the table.

Low risk as this attacks a very specific issue on a single table in the profile summary.

The stress test cases have been run to verify the changes.

No documentation impact.

### DIFF
--- a/src/runtime_src/xdp/profile/database/static_info/pl_constructs.cpp
+++ b/src/runtime_src/xdp/profile/database/static_info/pl_constructs.cpp
@@ -21,6 +21,83 @@
 
 #include "xdp/profile/database/static_info/pl_constructs.h"
 
+// Anonymous namespace for local static helper functions
+namespace {
+
+  // This function will check if the spTag a particular port or argument
+  // is connected to belongs to a particular memory resource.  If the strings
+  // are an exact match, or if the memory resource is a range that the
+  // spTag is a part of, we return true.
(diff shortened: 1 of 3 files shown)


pgschuey (approved):

dayeh-xilinx:
retest this please.

PR #6862, opened 2022-07-19
CR-1131075 Support CU interrupt to APU

mamin506:
#### Problem solved by the commit
Support CU completion interrupt on APU

#### Risks (if any) associated the changes in the commit
No

#### What has been tested and how, request additional testing if necessary
vck5000 xbutil validate test, zcu104

#### Documentation impact (if any)
No

============================================
vck5000 xbutil validate iops result
```
$xbutil validate -d 04:00 -r iops
Starting validation for 1 devices

Validate Device           : [0000:04:00.1]
    Platform              : xilinx_vck5000_gen4x8_qdma_base_1
    SC Version            : 4.4.33
    Platform ID           : 2A6AB1C4-DFA1-27FD-7F24-E8F121039A2E
-------------------------------------------------------------------------------
Test 1 [0000:04:00.1]     : iops
    Details               : IOPS: 483580 (verify)
    Test Status           : [PASSED]
-------------------------------------------------------------------------------
Validation completed. Please run the command '--verbose' option for more details
```

PR #7066, opened 2022-09-22
Fix Unable to free interrupt after driver removal in edge platforms

rbramand-xilinx:

Interrupt requested is not freed even after driver removal in edge platforms because new intc platform device remove function is being called before cu platform device remove function which has free_irq function. This PR solves that issue.

PR - https://github.com/Xilinx/XRT/pull/6862 introduced this bug and it was caught in internal testing

Corrected the order of platform devices removal in zocl driver removal function

Low

1. Load xclbin  -> creates cu where irq is registered
2. rmmod zocl  ->  destroys cu and irq is freed
3. modprobe zocl and load xclbin  -> cu is created and irq registration does not fail after this fix
Tested this flow using vck190 hw_emu test case

NA


chvamshi-xilinx (approved):

dayeh-xilinx:
retest this please.

mamin506 (approved):

PR #7068, opened 2022-09-22
xclbinutil - Added kernel_id support

chienwei-lan:
Added kernel_id in extended-data for Pre/Post kernel pdi swapping enhancement.

Extract the value of dpu_kernel_id@kernel.json => kernel_id@EMBEDDED_METADATA

New functionality

Low risk, modified the metadata and passed the xclbinutil unit test

### DIFF
--- a/src/runtime_src/core/common/xclbin_parser.cpp
+++ b/src/runtime_src/core/common/xclbin_parser.cpp
@@ -251,7 +251,7 @@ get_kernel_id(const pt::ptree& xml_kernel, const std::string& element)
     if (elem.first != element)
       continue;
 
-    return convert(elem.second.get<std::string>("<xmlattr>.kernel_id"));
+    return convert(elem.second.get<std::string>("<xmlattr>.dpu_kernel_id"));
   }
 
   return 0;
@@ -875,7 +875,7 @@ get_aie_partition(const axlf* top)
       auto cdop = reinterpret_cast<const cdo_group*>(topbase + aiepdip->cdo_groups.offset + j * sizeof(cdo_group));
       auto kernel_idp = reinterpret_cast<const uint64_t*>(topbase + cdop->dpu_kernel_ids.offset);
       for (uint32_t k = 0; k < cdop->dpu_kernel_ids.size; ++k)
-        dpu_kernel_ids.push_back(*(kernel_idp++));
+        dpu_kernel_ids.push_back(kernel_idp[k]);
 
       pdiobj.cdo_groups.emplace_back<aie_cdo_group_obj>({topbase + cdop->mpo_name, cdop->cdo_type, cdop->pdi_id, std::move(dpu_kernel_ids)});
     }
--- a/src/runtime_src/core/common/xclbin_parser.cpp
+++ b/src/runtime_src/core/common/xclbin_parser.cpp
@@ -243,6 +243,20 @@ get_functional(const pt::ptree& xml_kernel, const std::string& element)
   return 0;
 }
 
+//Get the cu kernel id from kernel xml entry
+static size_t
+get_kernel_id(const pt::ptree& xml_kernel, const std::string& element)
+{
+  for (auto& elem : xml_kernel) {
+    if (elem.first != element)
+      continue;
+
+    return convert(elem.second.get<std::string>("<xmlattr>.kernel_id"));
+  }
+
+  return 0;
+}
+
 // Determine the address range from kernel xml entry
 static size_t
 get_address_range(const pt::ptree& xml_kernel)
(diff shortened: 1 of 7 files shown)


dayeh-xilinx:
retest this please.

src/runtime_src/core/common/xclbin_parser.cpp
```diff
@@ -857,10 +871,13 @@ get_aie_partition(const axlf* top)
     pdiobj.pdi.resize(aiepdip->pdi_image.size);
     memcpy(pdiobj.pdi.data(), topbase + aiepdip->pdi_image.offset, pdiobj.pdi.size());
     for (uint32_t j = 0; j < aiepdip->cdo_groups.size; j++) {
+      std::vector<uint64_t> dpu_kernel_ids;
       auto cdop = reinterpret_cast<const cdo_group*>(topbase + aiepdip->cdo_groups.offset + j * sizeof(cdo_group));
+      auto kernel_idp = reinterpret_cast<const uint64_t*>(topbase + cdop->dpu_kernel_ids.offset);
```
rozumx, line 876: Being that this points to an array of 64-bit values, it might be easier to work with the array instead of pointers.  
It will make reading line 878 easier to read.  Something like:
`dpu_kernel_ids.push_back(kernel_idp[k]);`

src/runtime_src/tools/xclbinutil/SectionAIEPartition.cxx
```diff
@@ -103,7 +103,9 @@ SectionAIEPartition::subSectionExists(const std::string& /*sSubSectionName*/) co
 static const std::vector<std::pair<std::string, CDO_Type>> CTTypes = {
   { "UNKNOWN", CT_UNKNOWN },
   { "PRIMARY", CT_PRIMARY },
-  { "LITE", CT_LITE }
+  { "LITE", CT_LITE },
+  { "PRE", CT_PREPOST },
```
rozumx, line 107: Why are the PRE and POST associated with the same enum value.   Shouldn't this be "PREPOST" instead?

Being that this is data driven, "PRE" and "POST" entries will be mapped to CT_PREPOST, unfortunately in reverse CT_PREPOST gets mapped to "PRE".

src/runtime_src/tools/xclbinutil/KernelUtilities.cxx
```diff
@@ -129,7 +129,15 @@ void buildXMLKernelEntry(const boost::property_tree::ptree& ptKernel,
   const boost::property_tree::ptree& ptExtendedData = ptKernel.get_child("extended-data", ptEmpty);
   if (!ptExtendedData.empty()) {
     boost::property_tree::ptree ptEntry;
-    ptEntry.add_child("<xmlattr>", ptExtendedData);
+    boost::property_tree::ptree ptNewExtendedData;
+    const std::string& kernel_id = ptExtendedData.get<std::string>("dpu_kernel_id", "");
```
rozumx, line 133: Why the filter as opposed to using the ptExtendedData directly?  

The idea here was to take the "extended-data" from the input JSON file and added it directly to the XML's extended data.   By doing so there isn't a need to have this functionality.

Side note, if there is a need to support a given input schema, please do let me know and a rapidSchema validator can be added.

rozumx (changes requested):

src/runtime_src/tools/xclbinutil/KernelUtilities.cxx
chienwei-lan: We have discussed and we don't like the name dpu_kernel_id while already has type="dpu" in somewhere.

So are we OK with dpu_kernel_id here?
<kernel name="PP_FD_PRE" language="c" type="dpu">
         <extended-data functional="1“ kernel_id ="0x100">

dayeh-xilinx:
retest this please.

dayeh-xilinx:
retest this please.

src/runtime_src/tools/xclbinutil/SectionAIEPartition.cxx
chienwei-lan: Have checked with DPU team, they are good with "PRE_POST" for all pre and post process kernel.

Updated the code

src/runtime_src/core/common/xclbin_parser.cpp
chienwei-lan: done

stsoe (approved):

larry9523 (approved):

chienwei-lan:
retest this please.

rozumx (approved):

PR #7071, opened 2022-09-24
CR-1129189 Throw exception on abnormal command completion

stsoe:
New `xrt::run::wait2()` API was added to throw an exception when command completes abnormally (state different from `ERT_CMD_STATE_COMPLETED`).

The existing `run::wait()` API cannot be changed without breaking ABI, so new function had to be added.

The new `wait2()` API can be given an optional timeout value in which case the API returns `std::cv_status::timeout` on the API timeout, or `std::cv_status::no_timeout` on successful command completion.

An `xrt::exception` base class was added and a derived `xrt::run::command_error` exception defined.  This was done so that the abnormal command state can be captured if an exception is thrown by the wait2() API.  The exception class is insulated, which may seem like an overkill, but it to facilitate otherwise ABI breaking changes.

Mostly changes to existing internal wait functions with adjustments to preserve behavior of old wait() API.  Inherently a risk here, so this PR should not be merged before XRT:master separates from 2022.2.

All OpenCL tests.
Unit test case `tests\xrt\abort` has been adjusted to support wait2() , while still supporting legacy wait().

Yes, should be documented.


src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -3184,6 +3242,33 @@ set_arg_at_index(int index, const xrt::bo& glb)
...
+{
+  return m_impl->m_state;
+}
+
+const char*
+run::command_error::
+what() const noexcept
+{
+  return m_impl->m_message.c_str();
+}
```
sonals, line 3268: Shouldn't we also include string representation of the `m_state`? I just looked up the following examples:

1. fs exceptions report reasons for the failure as well: https://en.cppreference.com/w/cpp/filesystem/copy_file
2. system error translates the meaning of error code: https://en.cppreference.com/w/cpp/error/system_error

stsoe: Yeah, it's just painful to translate the error codes, we have no such map yet.

stsoe: Added string representation at throw time.

stsoe:
retest this please

dayeh-xilinx:
retest this please.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -159,6 +159,29 @@ debug_cmd_packet(const std::string& msg, const ert_packet* pkt)
   return fnm;
 }
 
+std::string
+cmd_state_to_string(ert_cmd_state state)
+{
+  static const std::map<ert_cmd_state, const char*> ert_cmd_state_string {
+   {ERT_CMD_STATE_NEW, "ERT_CMD_STATE_NEW"},
```
keryell, line 166: I guess that all these top-level uppercase symbols comes from the compatibility with C headers too?

keryell (changes requested):
Interesting feature.

uday610 (approved):

src/runtime_src/core/common/api/xrt_kernel.cpp
stsoe: Yes.  :-(

tests/xrt/abort/main.cpp
```diff
@@ -160,8 +173,34 @@ timeout(xrt::run run, int hang, int timeout_ms)
   }
 }
 
+// Test wait with timeout with or without hanging kernel using run::wait2()
+static void
+timeout2(xrt::run run, int hang, int timeout_ms)
+{
+  auto status = run.wait2(timeout_ms * 1ms);
+  std::cout << "timeout: wait2 completed with status (" << cv_status_to_string(status) << ")\n";
```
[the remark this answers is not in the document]
stsoe: I have looked at this for a long time, and I cannot see what clang-format would do about it?  I have not been successful in getting clang-tidy to like my style :-)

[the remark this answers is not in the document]
keryell: Just that the line looks too long.

vboggara-xilinx (approved):

keryell (approved):
Cool!

PR #7077, opened 2022-09-28
CR-1142046 Fix APU boot on v70 obfuscated device

rbramand-xilinx:

Because of id code change APU boot fails on V70 obfuscated device, changed id code to fix it

Changed id code that is used in bif file to generate pdi according to new V70 obfuscated device. Generating new apu package is also an option but all V70 devices will be changed in future.

Using new APU package on non obfuscated V70 devices may cause APU boot failures but it is expected.

Tested APU booting on V70 device and it is successful

NA


chvamshi-xilinx (approved):

chvamshi-xilinx:
retest this please

chvamshi-xilinx:
retest this please

PR #7004, opened 2022-09-08
Add EXTERNAL_SHIM cmake flag to conditionally compile libxrt_core.so

maxzhen:
#### Problem solved by the commit
SHIM layer (libxrt_core.so) can be provided by a repo which uses xrt as a git submodule. Adding EXTERNAL_SHIM cmake flag to conditionally compile libxrt_core.so, so that the xrt built-in libxrt_core.so target does not conflict with the one provided by the repo.
#### Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered
N/A
#### How problem was solved, alternative solutions (if any) and why they were rejected
This is just a quick short term solution to unblock external shim layer development work.
#### Risks (if any) associated the changes in the commit
Low.
#### What has been tested and how, request additional testing if necessary
Build XRT in normal flow.
Build external libxrt_core.so in separate repo.
#### Documentation impact (if any)
N/A

PR #7087, opened 2022-10-04
Make XRT SHIM extend'able to support new platform drivers on Linux

maxzhen:
This PR is to enable extension to XRT SHIM to support new drivers for new platforms on Linux. The implementation of the new SHIM is from another repository.
The basic idea behind it is to introduce a pci_driver class which can scan the entire system and provide a list of supported devices as pci_device instances. Current system singleton instance can have multiple pci_driver instances. Each of them can help with discovering the kind of devices it can support. System instance will add all of them to a single pci_device list for future use.
After this change, adding support for a new driver requires adding a new class derived from base pci_driver class, which can provide a list of instances of its own pci_device class (derived from base pci_device class). A device class (derived from device_linux class) can be created through it to provide driver specific ishim implementation and query key table.
Ideally, this new driver class and new matching SHIM implementation will be provided in its own standalone .so as a plug-in for XRT. and will be loaded by libxrt_core.so during process startup. In one Linux system, there will be one XRT instance with multiple driver .so libs. This PR only introduces the pci_driver class. The support for .so driver plugin can be done in a follow up PR.

Note that this PR reverted changes from https://github.com/Xilinx/XRT/pull/7004, which was a short term change and is not needed any more.
N/A
See description above
New feature added. May cause regression.
Built, installed XRT package and tested on U55N on a Ubuntu 20.04 box with xbutil examine/validate.
Also, tested, together with the extension part from a different repo on new platform.
N/A


stsoe (commented):
Still reviewing.  Please update copyrights and use the SPDX form.

// Copyright (C) zzzz-2022 Xilinx, Inc. All rights reserved.
// Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.

src/runtime_src/core/pcie/linux/pcidrv.h
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
+
+#ifndef _XCL_PCIDRV_H_
+#define _XCL_PCIDRV_H_
+
+#include "pcidev.h"
```
stsoe, line 7: Looks like a circular dependency here.  pcidrv.h -> pcidev.h -> pcidrv.h

src/runtime_src/core/pcie/linux/pcidrv.h
```diff
@@ -0,0 +1,38 @@
...
+class pci_driver
+{
+public:
+  virtual
+  const std::string
+  name(void) const = 0;
+
+  virtual
+  bool
+  is_user(void) const = 0;
```
stsoe, line 25: Someone is implementing these, do you want to comment what they are and what they are suppose to return?

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -36,22 +38,32 @@
...
 
+xrt_core::system_linux* singleton = nullptr;
+
 // Singleton registers with base class xrt_core::system
-// during static global initialization.  If statically
+// during static global initialization. If statically
 // linking with libxrt_core, then explicit initialiation
 // is required
-static xrt_core::system_linux*
+static inline xrt_core::system_linux*
```
stsoe, line 47: inline should be compilers decision, this is definitely not critical.

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -36,22 +38,32 @@
...
+
+  if (singleton)
+    return singleton;
+
+  throw std::runtime_error("system_linux singleton is not initialized");
 }
 
 // Dynamic linking automatically constructs the singleton
+// Do not instantiate singleton, if SHIM is extended from outside.
+#ifndef EXTERNAL_SHIM
```
stsoe, line 61: didn't you remove EXTERNAL_SHIM?

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -98,26 +110,76 @@ static std::string machine_info()
...
 
+void
+system_linux::
+register_driver(std::shared_ptr<pcidrv::pci_driver> driver)
+{
+  driver_list.push_back(driver);
+  if (driver->is_user())
+    driver->scan_devices(user_ready_list, user_nonready_list);
+  else
+    driver->scan_devices(mgmt_ready_list, mgmt_nonready_list);
```
stsoe, line 125: Move the push_back to the end so as not to change driver_list state if scan_devices throws?
Also as you are correctly passing a copy of the shared pointer you should use driver_list.emplace_back(std::move(driver))  but of course not until after using the `driver` variable.

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -98,26 +110,76 @@ static std::string machine_info()
...
+
+size_t
+system_linux::
+get_num_dev_total(bool is_user) const
+{
+  if (is_user)
+    return user_ready_list.size() + user_nonready_list.size();
+  return mgmt_ready_list.size() + mgmt_nonready_list.size();
+}
+
```
stsoe, line 160: Are all these list accesses thread safe?  Could a list be in process of being changed / updated by another thread?

src/runtime_src/core/pcie/linux/system_linux.h
```diff
@@ -46,6 +46,32 @@ class system_linux : public system_pcie
...
+
+private:
+  std::vector<std::shared_ptr<pcidrv::pci_driver>> driver_list;
+
+  std::vector<std::shared_ptr<pcidev::pci_device>> user_ready_list;
+  std::vector<std::shared_ptr<pcidev::pci_device>> user_nonready_list;
+
+  std::vector<std::shared_ptr<pcidev::pci_device>> mgmt_ready_list;
+  std::vector<std::shared_ptr<pcidev::pci_device>> mgmt_nonready_list;
 };
```
stsoe, line 75: v

src/runtime_src/core/pcie/linux/system_linux.cpp
stsoe, line 160: Are all these list accesses thread safe?  Could a list be in process of being changed / updated by another thread?

maxzhen:
> Still reviewing. Please update copyrights and use the SPDX form.
> 
> // SPDX-License-Identifier: Apache-2.0 // Copyright (C) zzzz-2022 Xilinx, Inc. All rights reserved. // Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.

Will go through and fix them.

src/runtime_src/core/pcie/linux/pcidev.h
```diff
@@ -64,9 +66,14 @@ struct fdt_header {
     uint32_t size_dt_struct;
 };
 
+// Forward declaration
+namespace pcidrv {
+class pci_driver;
+};
```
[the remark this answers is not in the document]
maxzhen: Not sure if adding OS name in namespace is a good idea... it makes sense to have xrt_core::pci::drv and xrt_core::pci::dev. They are at the lowest level of SW stack closer to KMD. xrt_core::device_pcie can stay the same since it is upper level of device model providing ishim APIs. What do you think?

src/runtime_src/core/pcie/linux/pcidrv.h
maxzhen: Right, the forward declaration should be enough. I might have forgot to remove the included header after I added it. Will fix.

src/runtime_src/core/pcie/linux/pcidrv.h
maxzhen: Yes, will add comment.

src/runtime_src/core/pcie/linux/system_linux.cpp
maxzhen: Sure, will remove 'inline'.

src/runtime_src/core/pcie/linux/system_linux.cpp
maxzhen: This is still needed until we can really support driver plug-in .so libs. Today, the system_linux class instance implemented here only knows about pci_driver_xocl and pci_driver_xclmgmt. A derived system class instance knows new drivers and should register itself as the singleton. This code is in the extended part in external repo, where EXTERNAL_SHIM will be defined.

src/runtime_src/core/pcie/linux/system_linux.cpp
maxzhen: Sure, will fix.

src/runtime_src/core/pcie/linux/system_linux.cpp
maxzhen: We use the index into the list as the id of the device. This is still the case even after this change The device list can only be created in the constructor of system class singleton. I think we want to provide a stable id to device mapping, so once created, the list should not be modified. It seems that there is no race here?

src/runtime_src/core/pcie/linux/pcidev.h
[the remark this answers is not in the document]
stsoe: xrt_core::pci::drv/dev works for me.

src/runtime_src/core/pcie/linux/system_linux.h
maxzhen: After a user get a shared ptr of a dev through system_linux, it's shared ownership between the user and the list. If we ever want to be able to dynamically remove a dev from the list, we can do it by removing the dev from the list without keeping track of the users by ourselves. Of course, we need to add a lock to protect these lists and stop using index as id by then.

src/runtime_src/core/pcie/linux/system_linux.cpp
maxzhen: Duplicated comments?

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -98,26 +110,76 @@ static std::string machine_info()
...
+system_linux::
+system_linux()
+{
+  if (singleton)
+    throw std::runtime_error("More than one SHIM registered!");
+  singleton = this;
```
keryell, line 166: Argh. Avoid teleportation.
Move this logic to the singleton handler instead as Kate Gregory says.

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -98,26 +110,76 @@ static std::string machine_info()
...
 void
 system_linux::
 get_xrt_info(boost::property_tree::ptree &pt)
 {
   boost::property_tree::ptree _ptDriverInfo;
-  _ptDriverInfo.push_back( std::make_pair("", driver_version("xocl") ));
-  _ptDriverInfo.push_back( std::make_pair("", driver_version("xclmgmt") ));
+
+  for (auto drv : driver_list)
+    _ptDriverInfo.push_back(std::make_pair("", driver_version(drv->name())));
```
keryell, line 179:
```suggestion
    _ptDriverInfo.push_back({ "", driver_version(drv->name()) });
```

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -98,26 +110,76 @@ static std::string machine_info()
...
 
 void
 system_linux::
 get_xrt_info(boost::property_tree::ptree &pt)
 {
   boost::property_tree::ptree _ptDriverInfo;
-  _ptDriverInfo.push_back( std::make_pair("", driver_version("xocl") ));
-  _ptDriverInfo.push_back( std::make_pair("", driver_version("xclmgmt") ));
+
+  for (auto drv : driver_list)
```
keryell, line 178: Do you need a copy of `drv` here or `auto& drv` is fine?

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -279,3 +335,25 @@ get_device_id_from_bdf(const std::string& bdf)
 } // pcie_linux
 
 } // xrt_core
+
+namespace pcidev {
+
+size_t
+get_dev_ready(bool user)
+{
+  return singleton_instance()->get_num_dev_ready(user);
```
keryell, line 344: `singleton_instance()` is not an informative name.

src/runtime_src/core/pcie/tools/cloud-daemon/container/container.h
```diff
@@ -22,7 +22,7 @@
 #include "xclhal2.h"
 #include "core/pcie/driver/linux/include/mailbox_proto.h"
 #include "core/pcie/driver/linux/include/mgmt-ioctl.h"
-#include "core/pcie/linux/scan.h"
+#include "core/pcie/linux/pcidev.h"
```
keryell, line 25: Sort.

src/runtime_src/core/pcie/tools/cloud-daemon/pciefunc.h
```diff
@@ -21,7 +21,7 @@
 
 #include <string>
 #include <mutex>
-#include "core/pcie/linux/scan.h"
+#include "core/pcie/linux/pcidev.h"
```
keryell, line 24: Sort.

keryell (changes requested):
Quite interesting.

src/runtime_src/core/pcie/linux/pcidev.h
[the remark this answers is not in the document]
stsoe: @keryell That we won't do, `namespace xrt` is for publically exported XRT

src/runtime_src/core/pcie/linux/pcidrv.h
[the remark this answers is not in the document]
stsoe: @keryell I strongly believe in scanning for important information by just glancing at the left margin but modifier should be at same line as type.  In this case I would prefer:
```
virtual bool
is_user() const = 0;
```

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -36,22 +38,32 @@
...
+// during static global initialization. If statically
 // linking with libxrt_core, then explicit initialiation
 // is required
-static xrt_core::system_linux*
+static inline xrt_core::system_linux*
 singleton_instance()
 {
-  static xrt_core::system_linux singleton;
-  return &singleton;
+  if (!singleton)
```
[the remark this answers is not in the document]
stsoe: @keryell The singleton system object is the registration with core XRT.  It is initialized when libxrt_core is loaded at runtime.   As you know we have different backends (different libxrt_core).    It does not have to be atomic.

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -36,22 +38,32 @@
...
+    static xrt_core::system_linux s;
```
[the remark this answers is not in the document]
stsoe: @keryell See other comment.   The singleton system object is derived from core XRT base class and functions as a "factory" of xrt::device objects for a particular xrt_core  shim library.

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -36,22 +38,32 @@
...
+  throw std::runtime_error("system_linux singleton is not initialized");
 }
 
 // Dynamic linking automatically constructs the singleton
+// Do not instantiate singleton, if SHIM is extended from outside.
+#ifndef EXTERNAL_SHIM
 struct X
 {
   X() { singleton_instance(); }
 } x;
```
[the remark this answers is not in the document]
stsoe: @keryell Yeah, static singleton creation when library is loaded at runtime.   Unfortunately the code has to accomodate two entries into the singleton creation.  (1) when xrt_core is loaded at run-time and devices constructed through XRT C++ functions (e.g. xrt::device), or (2) when user's host application links with libxrt_core and uses the shim API xclOpen().  But it all has to play together, so a little strange indeed.

src/runtime_src/core/pcie/linux/system_linux.cpp
stsoe: @keryell It isn't.

src/runtime_src/core/pcie/linux/system_linux.cpp
stsoe: @keryell Yeah, I think that logic is there already.  It is strictly enforced that only one singleton exists and this is handled at base class level.  One problem we are catching is when users link the "xcl" application against xrt_core and then runs hw or sw emulation which tries to load libxrt_hwemu or swemu at runtime; this would imply two singletons, which is currently not supported because two core libraries cannot be loaded at the same time as long as we support xcl APIs.

maxzhen:
Closing for now while working on addressing code review comments, so adding more intermediate commits won't waste pipeline resources.

src/runtime_src/core/pcie/linux/pcidev.h
```diff
@@ -26,6 +26,8 @@
 #include <sys/stat.h>
 #include <fcntl.h>
 #include <mutex>
+#include "pcidrv.h"
+#include "device_linux.h"
```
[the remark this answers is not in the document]
maxzhen: Code changed, nothing to sort.

src/runtime_src/core/pcie/linux/pcidev.h
```diff
@@ -64,9 +66,14 @@ struct fdt_header {
...
 
+// Forward declaration
+namespace pcidrv {
+class pci_driver;
+};
+
 namespace pcidev {
 
-// One PCIE function on FPGA board
+// One PCIE function on FPGA or AIE device
```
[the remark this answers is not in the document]
maxzhen: Sure.

src/runtime_src/core/pcie/linux/pcidev.h
```diff
@@ -80,19 +87,14 @@ class pci_device
...
   size_t user_bar_size =      0;
-  bool is_mgmt() const
-  {
-      return mgmt;
-  }
+  bool is_mgmt =              false;
   bool is_ready =             false;
 
-  pci_device(const std::string& drv_name, const std::string& sysfs_name);
+  pci_device(const pcidrv::pci_driver* driver, const std::string& sysfs_name);
```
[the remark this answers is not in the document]
maxzhen: Sure.

src/runtime_src/core/pcie/linux/pcidev.h
```diff
@@ -142,29 +144,42 @@ class pci_device
...
+  // On legacy Alveo device, this handle can be used to lookup a device instance and
+  // make xcl HAL API calls.
+  // On new platforms, this handle can only be used to look up a device. HAL API calls
+  // through it are not supported any more.
+  virtual
+  xrt_core::device::handle_type
+  create_shim(xrt_core::device::id_type id) const;
 
 private:
   int map_usr_bar(void);
```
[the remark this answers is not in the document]
maxzhen: sure.

src/runtime_src/core/pcie/linux/pcidrv.cpp
```diff
@@ -0,0 +1,52 @@
...
+void
+pci_driver::
+scan_devices(std::vector<std::shared_ptr<pcidev::pci_device>>& ready_list,
+             std::vector<std::shared_ptr<pcidev::pci_device>>& nonready_list) const
+{
+  namespace bfs = boost::filesystem;
+  const std::string drv_root = "/sys/bus/pci/drivers/";
+  const std::string drvpath = drv_root + name();
+
+  if(!bfs::exists(drvpath))
```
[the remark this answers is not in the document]
maxzhen: Replaced with std::filesystem calls.

src/runtime_src/core/pcie/linux/pcidrv.cpp
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
+
+#include <boost/filesystem.hpp>
```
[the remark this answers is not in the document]
maxzhen: replaced with #include <filesystem>

src/runtime_src/core/pcie/linux/pcidrv.cpp
```diff
@@ -0,0 +1,52 @@
...
+      // devnode to make sure the device is really assigned to docker.
+      if (!bfs::exists(pf->get_subdev_path("", -1)))
+        continue;
+
+      // Insert detected device into proper list.
+      if (pf->is_ready)
+        ready_list.push_back(pf);
+      else
+        nonready_list.push_back(pf);
+    } catch (std::invalid_argument const& ex) {
```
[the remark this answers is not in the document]
maxzhen: Sure.

src/runtime_src/core/pcie/linux/pcidrv.h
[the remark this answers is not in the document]
maxzhen: Sure.

src/runtime_src/core/pcie/linux/pcidrv_xclmgmt.h
```diff
@@ -0,0 +1,25 @@
...
+
+class pci_driver_xclmgmt : public pci_driver
+{
+public:
+  const std::string
+  name(void) const override
+  { return "xclmgmt"; }
+
+  bool
+  is_user(void) const override
```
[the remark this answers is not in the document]
maxzhen: Sure.

src/runtime_src/core/pcie/linux/pcidrv_xocl.h
```diff
@@ -0,0 +1,25 @@
...
+
+class pci_driver_xocl : public pci_driver
+{
+public:
+  const std::string
+  name(void) const override
+  { return "xocl"; }
+
+  bool
+  is_user(void) const override
```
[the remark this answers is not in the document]
maxzhen: Sure.

src/runtime_src/core/pcie/linux/shim.cpp
```diff
@@ -1011,7 +1011,7 @@ void shim::xclSysfsGetDeviceInfo(xclDeviceInfo2 *info)
     mDev->sysfs_get("", "link_width_max", errmsg, info->mPCIeLinkWidthMax, static_cast<unsigned short>(-1));
 
     //dont try to get any information which needs mailbox communication when device is not ready.
-    if(!mDev->is_mgmt() && !mDev->is_ready)
+    if(!mDev->is_mgmt && !mDev->is_ready)
```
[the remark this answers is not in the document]
maxzhen: Yeah, the code were copied over...

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -169,7 +231,7 @@ get_device_id(const std::string& bdf) const
     return system::get_device_id(bdf);
     
   unsigned int i = 0;
-  for (auto dev = pcidev::get_dev(i); dev; i++, dev = pcidev::get_dev(i)) {
+  for (auto dev = get_pcidev(i); dev; i++, dev = get_pcidev(i)) {
```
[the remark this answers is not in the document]
maxzhen: sure.

src/runtime_src/core/pcie/linux/system_linux.cpp
maxzhen: Sure.

src/runtime_src/core/pcie/linux/system_linux.cpp
maxzhen: Should be a reference, not a copy.

src/runtime_src/core/pcie/linux/system_linux.cpp
maxzhen: Changed to "singleton_system_linux"

src/runtime_src/core/pcie/tools/cloud-daemon/container/container.h
maxzhen: By header path length? By header filename?

src/runtime_src/core/pcie/tools/cloud-daemon/pciefunc.h
maxzhen: By header path length? By header filename?

src/runtime_src/core/pcie/linux/pcidev.h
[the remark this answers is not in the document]
stsoe: By name in groups.

src/runtime_src/core/pcie/linux/pcidrv.cpp
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
+
+#include <filesystem>
```
stsoe, line 4: I don't think std::filesystem works on Ubuntu18.04?

src/runtime_src/core/pcie/linux/pcidev.h
```diff
@@ -142,43 +129,93 @@ class pci_device
...
-  int poll(int devhdl, short events, int timeoutMilliSec);
-  virtual void *mmap(int devhdl, size_t len, int prot, int flags, off_t offset);
-  virtual int munmap(int devhdl, void* addr, size_t len);
-  int flock(int devhdl, int op);
-  int get_partinfo(std::vector<std::string>& info, void *blob = nullptr);
-  std::shared_ptr<pcidev::pci_device> lookup_peer_dev();
+  get_subdev_path(const std::string& subdev, uint32_t idx) const;
+
+  virtual int
+  pcieBarRead(uint64_t offset, void *buf, uint64_t len);
```
stsoe, line 135: `void *` -> `void*` all over the place,

src/runtime_src/core/pcie/linux/pcidev.h
```diff
@@ -142,43 +129,93 @@ class pci_device
...
+  open(const std::string& subdev, uint32_t idx, int flag) const;
+
+  virtual void
+  close(int devhdl) const;
+
+  virtual int
+  ioctl(int devhdl, unsigned long cmd, void *arg = nullptr) const;
+
+  virtual int
+  poll(int devhdl, short events, int timeoutMilliSec);
```
stsoe, line 153:
```suggestion
  poll(int devhdl, short events, int timeout_ms);
```
No camelCase, please.

src/runtime_src/core/pcie/linux/pcidev.h
```diff
@@ -142,43 +129,93 @@ class pci_device
...
+get_dev_ready(bool user = true);
+
+std::shared_ptr<dev>
+get_dev(unsigned index, bool user = true);
+
+int
+get_axlf_section(const std::string& filename, int kind, std::shared_ptr<char>& buf);
+
+int
+get_uuids(std::shared_ptr<char>& dtbbuf, std::vector<std::string>& uuids);
```
stsoe, line 204: Looks like these functions are returning status?   If so change to exception and return what is being asked for.  shared_ptr<char> does not look right, maybe std::vector<char>?  What do these  function actually do and where is they used?

src/runtime_src/core/pcie/linux/pcidev.h
```diff
@@ -142,43 +129,93 @@ class pci_device
...
+int
+check_p2p_config(const std::shared_ptr<dev>& dev, std::string &err);
 
-} // namespace pcidev
+} } // namespace xrt_core :: pci
 
 // For print out per device info
 std::ostream&
-operator<< (std::ostream& stream, const std::shared_ptr<pcidev::pci_device>& dev);
+operator<< (std::ostream& stream, const std::shared_ptr<xrt_core::pci::dev>& dev);
```
stsoe, line 219: Unless ownership is shared within this function, passing const std::shared_ptr is not necessary.  It looks like this function should take `const xrt_core::pci::dev*` and that caller should `.get()` and pass the pointer.

src/runtime_src/core/pcie/linux/pcidrv.cpp
```diff
@@ -0,0 +1,54 @@
...
+      auto pf = create_pcidev(path.filename().string());
+
+      // In docker, all host sysfs nodes are available. So, we need to check
+      // devnode to make sure the device is really assigned to docker.
+      if (!std::filesystem::exists(pf->get_subdev_path("", -1)))
+        continue;
+
+      // Insert detected device into proper list.
+      if (pf->is_ready)
+        ready_list.push_back(pf);
```
stsoe, line 38:
```suggestion
        ready_list.push_back(std:move(pf));
```

src/runtime_src/core/pcie/linux/pcidrv.cpp
```diff
@@ -0,0 +1,54 @@
...
+      else
+        nonready_list.push_back(pf);
```
stsoe, line 40:
```suggestion
        nonready_list.push_back(std::mov(pf));
```

src/runtime_src/core/pcie/linux/pcidrv.cpp
```diff
@@ -0,0 +1,54 @@
...
+    } catch (const std::invalid_argument& ex) {
```
stsoe, line 41:
```suggestion
    } 
    catch (const std::invalid_argument& ex) {
```

src/runtime_src/core/pcie/linux/pcidrv_xclmgmt.h
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
+
+#ifndef _XCL_PCIDRV_XCLMGMT_H_
+#define _XCL_PCIDRV_XCLMGMT_H_
+
+#include "pcidrv.h"
+
```
stsoe, line 8:
```suggestion
#include <string>

```

src/runtime_src/core/pcie/linux/pcidrv_xclmgmt.h
```diff
@@ -0,0 +1,25 @@
...
+namespace xrt_core { namespace pci {
+
+class drv_xclmgmt : public drv
+{
+public:
+  const std::string
```
stsoe, line 14:
```suggestion
  std::string
```
const here doesn't do anything at all.

src/runtime_src/core/pcie/linux/pcidrv_xocl.h
```diff
@@ -0,0 +1,25 @@
...
+#define _XCL_PCIDRV_XOCL_H_
+
+#include "pcidrv.h"
+
+namespace xrt_core { namespace pci {
+
+class drv_xocl : public drv
+{
+public:
+  const std::string
```
stsoe, line 14:
```suggestion
  std::string
```

src/runtime_src/core/pcie/linux/pcidev.h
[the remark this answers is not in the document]
maxzhen: Not sure I follow? Do you mean 'device_linux.h' should go before 'pcidrv.h' due to letter 'd' is ahead of 'p'? I just failed to see order from the original code...

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -36,22 +38,32 @@
...
+xrt_core::system_linux* singleton = nullptr;
+
 // Singleton registers with base class xrt_core::system
-// during static global initialization.  If statically
+// during static global initialization. If statically
 // linking with libxrt_core, then explicit initialiation
 // is required
 static xrt_core::system_linux*
-singleton_instance()
+singleton_system_linux()
```
stsoe, line 48: Let's rewrite this function a bit tomorrow.  It is obfuscated from all the comments right now, but doesn't look quite right.

src/runtime_src/core/pcie/linux/pcidrv.cpp
maxzhen: Indeed! Isn't it part of standard C++17? I'm confused. I'll change it back. Sigh...

src/runtime_src/core/pcie/linux/pcidev.h
maxzhen: I can clean it up.

src/runtime_src/core/pcie/linux/pcidev.h
maxzhen: Not my code, but I can clean it up.

src/runtime_src/core/pcie/linux/pcidev.h
maxzhen: I have no idea what they are doing and I prefer not to touch them. I merely followed your suggestion and push function name to the very left.

src/runtime_src/core/pcie/linux/pcidev.h
maxzhen: Ancient code, prefer not to touch.

src/runtime_src/core/pcie/linux/pcidrv.cpp
maxzhen: Sure.

src/runtime_src/core/pcie/linux/pcidrv.cpp
maxzhen: Sure.

src/runtime_src/core/pcie/linux/pcidrv.cpp
maxzhen: Oh, 'catch' statement should start a new line? OK.

src/runtime_src/core/pcie/linux/pcidrv_xclmgmt.h
maxzhen: Sure.

src/runtime_src/core/pcie/linux/pcidrv_xclmgmt.h
maxzhen: It means that the content of the string can't be changed, no?

src/runtime_src/core/pcie/linux/pcidrv_xocl.h
maxzhen: ditto.

src/runtime_src/core/pcie/linux/system_linux.cpp
maxzhen: Ha! I thought I followed the sample code in system class...

src/runtime_src/core/pcie/linux/pcidev.h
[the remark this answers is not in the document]
stsoe: See email with link to guidelines.

src/runtime_src/core/pcie/linux/pcidrv.cpp
stsoe: The compile on Ubuntu18.04 is not C++17 std compliant for at least std::filesystem.   We need to get rid of this, it is the hold out for general move from boost::filesystem -> std::filesystem

src/runtime_src/core/pcie/linux/pcidev.h
stsoe: Thanks.  I know that it is copied code, but when we are in the neighborhood, we might as well clean it :-)

src/runtime_src/core/pcie/linux/pcidev.h
stsoe: That's how I noticed them :-)  Sure leave, but someone has to take a look one day.

src/runtime_src/core/pcie/linux/pcidrv.cpp
stsoe: Just my preference in general to be at the left as much as possible.
```
if (...) {
  stmt1;
  stmt2;
}
else {
  stmt3;
  stmt4
}
```

```
try {
  ...;
}
catch (...) {
}
```

src/runtime_src/core/pcie/linux/pcidrv_xclmgmt.h
stsoe: Return-by-value, which is correct here, but the rvalue will be assigned to some lvalue which caller can change.

maxzhen: This string contains the name of the driver/module and should never be changed, even by the caller. So, I can return const string for this?

src/runtime_src/core/pcie/linux/pcidev.h
maxzhen: They won't be implemented in newly derived AIE class anyway :-).

src/runtime_src/core/pcie/linux/pcidrv.cpp
maxzhen: I have been trying not to make too much change to the existing code which is not really necessary. But, this one was too attempting, hehe...

src/runtime_src/core/pcie/linux/pcidev.h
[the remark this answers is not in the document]
maxzhen: Understood now. Will them move to the front.

src/runtime_src/core/pcie/linux/pcidrv_xclmgmt.h
stsoe: You are returning a copy, the caller owns the copy and can do with it what they want.
std::string mycpy = name();  // will be fine, and mycpy can be changed
auto mycpy = name();  // will be fine, but auto does not capture cv-qualifiers so same as above

src/runtime_src/core/pcie/tools/cloud-daemon/container/container.h
keryell: By lexicographic alphabetic order.
Header path length would be ambiguous. :-)

src/runtime_src/core/pcie/tools/cloud-daemon/pciefunc.h
keryell: By lexicographic alphabetic order.
Header path length would be ambiguous. :-)

src/runtime_src/core/pcie/linux/pcidrv.cpp
keryell: Who would use a 4+ year old system anyway? :-)
Just add a more modern compiler to the install dependency?
According to https://en.cppreference.com/w/cpp/compiler_support/17#C.2B.2B17_library_features this is implemented in `g++-8`, `clang++-7` and above.
Ubuntu 18.04 might come with an old `g++` but you could install `g++-8` and/or `clang++-10` package on it, with their C++ standard libraries.
https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=gcc-8&searchon=names

src/runtime_src/core/pcie/linux/system_linux.cpp
stsoe: You don't need to change anything in regards to construction of the system singleton.

src/runtime_src/core/pcie/linux/pcidev.h
maxzhen: It turned out that they are not even used. I'll just remove them completely.

src/runtime_src/core/pcie/tools/cloud-daemon/pciefunc.h
maxzhen: Sure. And also follow Soren's grouping rule.

src/runtime_src/core/pcie/tools/cloud-daemon/container/container.h
maxzhen: Sure. And also follow Soren's grouping rule.

src/runtime_src/core/pcie/linux/pcidrv.cpp
maxzhen: Sure.

src/runtime_src/core/pcie/linux/system_linux.cpp
maxzhen: As discussed offline, we'll do it this way for now until we have our driver plug-in implemented.

dayeh-xilinx:
retest this please.

stsoe (approved):

maxzhen:
retest this please - why it's not even in the pipeline queue?

stsoe (commented):
Hey @maxzhen .  I think we should run this PR through regression testing before merge.
I can run my standard OpenCL tests which still functions to catch fundamental errors.  Just caught one on my #7100 PR.  I can do this testing and report back here.

stsoe:
> Hey @maxzhen . I think we should run this PR through regression testing before merge. I can run my standard OpenCL tests which still functions to catch fundamental errors. Just caught one on my #7100 PR. I can do this testing and report back here.

Regression are clean.

maxzhen:
> > Hey @maxzhen . I think we should run this PR through regression testing before merge. I can run my standard OpenCL tests which still functions to catch fundamental errors. Just caught one on my #7100 PR. I can do this testing and report back here.
> 
> Regression tests are clean.

Thanks, @stsoe!

src/runtime_src/core/pcie/linux/debug.cpp
```diff
@@ -1,37 +1,6 @@
...
- * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
- * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
- * License for the specific language governing permissions and limitations
- * under the License.
- */
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2015-2017 Xilinx, Inc
+// Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
 
 #include "core/include/xdp/aim.h"
```
rozumx, line 5: It is best to sort the include files by name.

src/runtime_src/core/pcie/linux/pcidev.cpp
```diff
@@ -1,21 +1,6 @@
...
- * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
- * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
- * License for the specific language governing permissions and limitations
- * under the License.
- */
-
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2016-2020 Xilinx, Inc
+// Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
 #include <stdexcept>
```
rozumx, line 4: Sort include files.

src/runtime_src/core/pcie/linux/pcidev.h
```diff
@@ -78,22 +69,18 @@ class pci_device
...
-  uint16_t device_id =        INVALID_ID;
   uint32_t instance =         INVALID_ID;
   std::string sysfs_name =    ""; // dir name under /sys/bus/pci/devices
   int user_bar =              0;  // BAR mapped in by tools, default is BAR0
   size_t user_bar_size =      0;
-  bool is_mgmt() const
-  {
-      return mgmt;
-  }
+  bool is_mgmt =              false;
```
rozumx, line 78: Naming convention.  Being that these variables are member variables of the class, normally an "m_" is added to them. 

Is there a reason why this isn't done in this "container" class?

src/runtime_src/core/pcie/linux/pcidev.cpp
```diff
@@ -467,57 +428,43 @@ get_subdev_path(const std::string& subdev, uint idx)
...
+dev::
+open(const std::string& subdev, int flag) const
 {
   return open(subdev, 0, flag);
 }
 
-pci_device::
-pci_device(const std::string& drv_name, const std::string& sysfs) : sysfs_name(sysfs)
+dev::
+dev(const drv& driver, const std::string& sysfs) : sysfs_name(sysfs)
```
rozumx, line 458: The constructor should be initializing all of the member variables.

src/runtime_src/core/pcie/linux/pcidev.h
```diff
@@ -78,22 +69,18 @@ class pci_device
...
   bool is_ready =             false;
 
-  pci_device(const std::string& drv_name, const std::string& sysfs_name);
-  virtual ~pci_device();
+  dev(const drv& driver, const std::string& sysfs_name);
```
rozumx, line 81: The compiler will create a default constructor.  So it might be best if you explicitly prevent one from begin created.  Something like:
`dev() = delete;`

OR

If you expect one to be created, then create definition and body that will initializes the class and have ALL other constructors use it.

src/runtime_src/core/pcie/linux/pcidev.h
```diff
@@ -78,22 +69,18 @@ class pci_device
 
   uint16_t domain =           INVALID_ID;
   uint16_t bus =              INVALID_ID;
-  uint16_t dev =              INVALID_ID;
+  uint16_t dev_no =           INVALID_ID;
   uint16_t func =             INVALID_ID;
-  uint16_t vendor_id =        INVALID_ID;
-  uint16_t device_id =        INVALID_ID;
   uint32_t instance =         INVALID_ID;
   std::string sysfs_name =    ""; // dir name under /sys/bus/pci/devices
```
rozumx, line 75: By default strings are initialized to empty strings.  So this extra assignment isn't needed.
```suggestion
  std::string sysfs_name;; // dir name under /sys/bus/pci/devices
```

src/runtime_src/core/pcie/linux/pcidrv_xclmgmt.h
```diff
@@ -0,0 +1,26 @@
...
+#include "pcidrv.h"
+#include <string>
+
+namespace xrt_core { namespace pci {
+
+class drv_xclmgmt : public drv
+{
+public:
+  std::string
+  name() const override
```
rozumx, line 16: formatting suggestion:
```suggestion
  name() const override { return "xclmgmt"; }
```

Make reading the code a little bit easier.

src/runtime_src/core/pcie/linux/pcidrv_xclmgmt.h
```diff
@@ -0,0 +1,26 @@
...
+  { return "xclmgmt"; }
+
+  bool
+  is_user() const override
```
rozumx, line 20: formatting suggestion:
```suggestion
  is_user() const override  { return false; }
```

src/runtime_src/core/pcie/linux/pcidrv_xocl.h
```diff
@@ -0,0 +1,26 @@
...
+#include "pcidrv.h"
+#include <string>
+
+namespace xrt_core { namespace pci {
+
+class drv_xocl : public drv
+{
+public:
+  std::string
+  name() const override
```
rozumx, line 16: Please see earlier formatting suggestions.

src/runtime_src/core/pcie/linux/pcidrv_xocl.h
```diff
@@ -0,0 +1,26 @@
...
+  { return "xocl"; }
+
+  bool
+  is_user() const override
```
rozumx, line 20: Please see earlier formatting suggestions.

src/runtime_src/core/pcie/linux/pcidev.h
rozumx, line 78: Instead of being `public`, could these variables be `protected`?

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -5,7 +5,9 @@
 #include "device_linux.h"
```
rozumx, line 5: Please sort the include list.

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -98,26 +112,76 @@ static std::string machine_info()
...
 
+void
+system_linux::
+register_driver(std::shared_ptr<pci::drv> driver)
+{
+  if (driver->is_user())
+    driver->scan_devices(user_ready_list, user_nonready_list);
+  else
+    driver->scan_devices(mgmt_ready_list, mgmt_nonready_list);
+  driver_list.emplace_back(std::move(driver));
```
rozumx, line 127: Format: Insert new line
```suggestion

    driver_list.emplace_back(std::move(driver));
```

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -98,26 +112,76 @@ static std::string machine_info()
...
+{
+  if (is_user) {
+    if (index < user_ready_list.size())
+      return user_ready_list[index];
+    return user_nonready_list[index - user_ready_list.size()];
+  }
+
+  if (index < mgmt_ready_list.size())
+    return mgmt_ready_list[index];
+  return mgmt_nonready_list[index - mgmt_ready_list.size()];
```
rozumx, line 142: Format: Insert new line.

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -98,26 +112,76 @@ static std::string machine_info()
...
+}
+
+size_t
+system_linux::
+get_num_dev_ready(bool is_user) const
+{
+  if (is_user)
+    return user_ready_list.size();
+  return mgmt_ready_list.size();
```
rozumx, line 151: Inserting the new line does make the code easier to read.

rozumx (changes requested):

src/runtime_src/core/pcie/linux/debug.cpp
maxzhen: Sorted by name and group.

src/runtime_src/core/pcie/linux/pcidev.cpp
maxzhen: Sorted by name and group.

src/runtime_src/core/pcie/linux/pcidev.h
maxzhen: added "m_" prefix.

src/runtime_src/core/pcie/linux/pcidev.cpp
maxzhen: add "m_user_bar_map = reinterpret_cast<char *>(MAP_FAILED);" in the constructor.

src/runtime_src/core/pcie/linux/pcidev.h
maxzhen: added "dev() = delete" to prevent default constructor from being created by compiler.

src/runtime_src/core/pcie/linux/pcidev.h
maxzhen: removed.

src/runtime_src/core/pcie/linux/pcidrv_xclmgmt.h
maxzhen: sure.

src/runtime_src/core/pcie/linux/pcidrv_xclmgmt.h
maxzhen: sure.

src/runtime_src/core/pcie/linux/pcidrv_xocl.h
maxzhen: sure.

src/runtime_src/core/pcie/linux/pcidrv_xocl.h
maxzhen: sure.

src/runtime_src/core/pcie/linux/pcidev.h
maxzhen: looks like it's used in shim class, so can't make it protected.

src/runtime_src/core/pcie/linux/system_linux.cpp
maxzhen: sorted.

src/runtime_src/core/pcie/linux/system_linux.cpp
maxzhen: sure.

src/runtime_src/core/pcie/linux/system_linux.cpp
maxzhen: sure.

src/runtime_src/core/pcie/linux/system_linux.cpp
maxzhen: sure.

chvamshi-xilinx (approved):
Looks good.

rozumx (approved):
Thank you for making the changes.

PR #7088, opened 2022-10-04
workaround for CR-1141322

mamin506:
DPU hang with XRT 2022.2

In XRT 2022.2, KDS launch a kthread which can polling CU and wait interrupt from CU.
The process is,
1. Once submit a command to CU. The thread will polling CU for a while, like 30 times.
2. If CU is done very fast, this thread can get CU is done and notify host quickly.
3. If CU is slow, the thread will stop polling and wait for CU interrupt (with 0.5 second timeout).

But in this new process, the DPU will get stuck in ap_start statue. After many experiment, it seems DPU doesn't like write GIE (Global Interrupt Enable) register after ap_start is set.

Because this is related to the DPU kernel behavior. There is no final solution yet.
This is a workaround in CU subdevice, which will force the thread to use interrupt and never polling.
This workaround only impact Edge platform.

Low

ZCU102 DPU test case

No

### DIFF
--- a/src/runtime_src/core/common/drv/include/xrt_cu.h
+++ b/src/runtime_src/core/common/drv/include/xrt_cu.h
@@ -367,7 +367,8 @@ struct xrt_cu {
 	struct timer_list	  timer;
 	atomic_t		  tick;
 	u32			  start_tick;
-	
+	u32			  force_intr;
+
 	struct xrt_cu_stats        stats;
 	/**
 	 * @funcs:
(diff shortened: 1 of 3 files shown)


maxzhen (approved):

chvamshi-xilinx (commented):
Looks good Min.
We should clone this CR and use the new CR for 2023.1 to analyze and do the right fix. 
I am not sure whether this is related to DPU Kernel. or it is a real issue.

PR #7091, opened 2022-10-05
Add compile-time guard for device profiling

jvillarre:
Adds a missing compile time guard to protect sections of the profiling code that should not include Ioctl based control of profiling monitors when building on Linux.

Low risk as this guard is not set in the master branch, but is intended to be used in the future if necessary.

None


IshitaGhosh (commented):
Changes look good.

PR #7090, opened 2022-10-05
CR-1142217 xball xbutil validate throws error KeyError: 'XILINX_XRT'

rchane:
#### Problem solved by the commit
https://jira.xilinx.com/browse/CR-1142217 Using sudo xball (not just validate) was throwing an error when looking for XILINX_XRT variable.
#### Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered
This is a specific use case where xrt is installed in a /bin dir that can be found by sudo, where all tools except for xball work with sudo (otherwise, we expect to see `sudo: xball: command not found`). The issue arose [here](https://github.com/Xilinx/XRT/commit/ff35e8a78cba106779e842cda213b3c38ed1e64f) when XILINX_XRT was used instead of XRT_BIN_DIR.
#### How problem was solved, alternative solutions (if any) and why they were rejected
XRT_BIN_DIR is used again as an argument.
#### Risks (if any) associated the changes in the commit
N/A
#### What has been tested and how, request additional testing if necessary
Testing was done on a vck5000.
```
rchane@xsjsagarw50:/proj/rdi/staff/rchane/XRT$ export PATH=$PATH:$RCHOME/build/Debug/opt/xilinx/xrt/bin
rchane@xsjsagarw50:/proj/rdi/staff/rchane/XRT$ echo $PATH
/proj/xsjhdstaff4/rchane/.vscode-server/bin/74b1f979648cc44d385a2286793c226e611f59e7/bin/remote-cli:/home/rchane/bin:/usr/local/bin:/mis/TREE/bin:/usr/bin:/bin:/usr/ucb:/proj/rdi/staff/rchane/XRT/build/Debug/opt/xilinx/xrt/bin
rchane@xsjsagarw50:/proj/rdi/staff/rchane/XRT$ sudo env "PATH=$PATH" xball xbutil examine
[sudo] password for rchane: 
Discovering installed devices....


=====================================================================
1 / 3 [0000:65:00.1] : xilinx_vck5000_gen4x8_qdma_base_2
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Command: /proj/rdi/staff/rchane/XRT/build/Debug/opt/xilinx/xrt/bin/xbutil --device 0000:65:00.1 examine

WARNING: Unexpected xocl version (2.14.340) was found. Expected 2.15.0, to match XRT tools.

---------------------------------------------------
[...]

PR #7094, opened 2022-10-06
CR-1142217 Change cmd to use xrt_bin_dir instead of XILINX_XRT (#7090)

rchane:
(cherry picked from commit f623347121679faf55f1f88055019a81dbf16367)

https://jira.xilinx.com/browse/CR-1142217 sudo xball was throwing a key error.


chvamshi-xilinx (approved):

anupcs-xlnx (approved):
Affects only the xball script and does not risk XRT itself. Ok to backport to 2022.2.

PR #7096, opened 2022-10-06
Enable cgroup v2 in petalinux

jeffli-xilinx:
Added cgroup CPUShare and MemoryLimit to skd.service

Signed-off-by: Jeff Lin <jeffylin@xilinx.com>

Enabled cgroup v2 and added CPUShare of 1024 and MemoryLimit of 1GB for skd

None.  This is enhancement for skd.

None

Test skd on vck5000 and v70

None

### DIFF
--- a/src/runtime_src/tools/scripts/apu_recipes/skd.service
+++ b/src/runtime_src/tools/scripts/apu_recipes/skd.service
@@ -10,7 +10,6 @@ User=softkernel
 ExecStart=/usr/bin/skd
 RemainAfterExit=on
 StandardOutput=journal+console
-CPUShare=1024
 MemoryLimit=1G
 
 [Install]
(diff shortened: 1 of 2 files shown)


rozumx (changes requested):
Could you help me understand what we needed to enable cgroups?

I ask that being that this is being enabled for the first time, I wonder if the other daemons / processes have CPUShared set for this value is relative to each other.  What is the impact on the VDU drivers?

Also, how did you determine the memory limit of the soft kernel daemon (SKD)?  Being that this SKD creates and starts the PS kernels, does the sum of all of the  PS kernel memory need to be lower then the MemoryLimit value?

Note
Can these values be obtain an OS call (instead of examining the service file)?  If so, then you should work with @dbenusov-xilinx  to update the PS Kernel report to recode this information (along with how much memory is currently being consumed).

rozumx (approved):

chvamshi-xilinx (approved):

PR #7098, opened 2022-10-07
 fixing qdma msix vector index issue

chvamshi-xilinx:
Latest Platform is using MSIX vectors from 9 to 12 for the command queue XGQs. Currently we kept max msix vector as 8. Changed that to 16
Host is still not getting interrupts, but that could be a hardware issue.

it never worked.

Increased the maximum msix user interrupt vector index from 8 to 16

No

Veirifed xbutil validate on V70

NA


maxzhen (approved):

xdavidz:
@chvamshi-xilinx can you please check if this fix works with vck5000 qdma2 shell?

chvamshi-xilinx:
Hi @xdavidz , I have already verified this change with older shell as latest shell has some issue with command submitting.
I have used 1 week old shell.

PR #7101, opened 2022-10-07
Added and updated latest theme files

ryanvergel:


uday610 (approved):
Thank you

stsoe (changes requested):
Same comments as  for #7102

PR #7102, opened 2022-10-07
Added latest theme files

ryanvergel:


uday610 (approved):
Thank you

stsoe (changes requested):
Binary files and most files are missing copyright.
At least one file claims to be from some other place, but no copyright included.
Some files are missing a newline at end.

uday610:
Why it is failing binary cheks? I see it is giving error for png file, many PNG files are already there inside the toc directory from long time

PR #7103, opened 2022-10-08
Enable xrt run object to return PS kernel return code

jeffli-xilinx:
Add new function return_code to XRT run object to enable PS kernel return code to be passed back to host.

New feature for PS kernel

Added new function to return return code if kernel type is ps.

None

Tested on vck5000 with PS kernel

Need to document new return_code function for XRT run object

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -745,6 +745,16 @@ public:
     return state;
   }
 
+  // Return kernel return code from command object for PS kernels
+  int
+  get_return_code() const
+  {
+    auto pkt = get_ert_packet();
+    int ret;
+    ert_read_return_code(pkt, ret);
+    return ret;
+  }
+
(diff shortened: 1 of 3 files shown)


manikandan-xilinx:
retest this please.

stsoe (commented):
Looks good.

The ert_read_ret_code is definitely not an int.
It is a uint32_t, but if we want to interpret that as an int32_t then that's fine, but I think that is what we should return then.

jeffli-xilinx:
> Looks good.
> 
> The ert_read_ret_code is definitely not an int. It is a uint32_t, but if we want to interpret that as an int32_t then that's fine, but I think that is what we should return then.

I had missed the return type.
I have made the fix and updated the code.
Thanks!

chvamshi-xilinx (approved):

stsoe (approved):

stsoe:
retest this please - not clear why it failed, all lights are green

uday610 (approved):

mamin506 (approved):

PR #7104, opened 2022-10-11
Support external SHIM plug-in in Linux

maxzhen:

This is a follow-up change for PR#7087. It fully enables external driver plug-in .so in Linux. A driver plugin should be implemented as below:
1. Make sure it calls xrt_core::pci::register_driver() during library initialization phase to register itself with libxrt_core.so.
2. Should implement inherited xrt_core::pci::drv class, which hands out inherited xrt_core::pci::dev object during scanning of the system.
3. Should implement inherited xrt_core::device_linux class to provide ishim and query table implementation specific to the driver.
4. Should implement an internal "shim" class for the meat of ishim implementation. The xrt core framework requires an opaque pointer (shim handle) to an instance of it per xrt_core::device. (this requirement can be lifted if framework is refactored and do not need this handle in the future)
5. In order for libxrt_core.so to discover and load all driver plug-ins at run time, they should be copied to the same directory as libxrt_core.so and named as libxrt_driver_xxx.so.\<xrt-core-lib-major-version\>

When XRT package is installed, by default, it can see devices driven by xocl and xclmgmt. When a driver plug-in .so is installed, it can see more devices driven by other drivers.
New feature
N/A
N/A
+ Built and installed XRT package and run xbutil examine and validate on a U55N card.
+ Built and installed a driver plug-in from another repo for a new platform and tested it.
Internal feature, no doc for now.


chvamshi-xilinx:
Hi @maxzhen , Overall it looks good. I have couple of questions

- You have mentioned in the PR description that driver plugin must register itself with libxrt_core.so. libxrt_core.so is mentioned in some other places in PR description, is it libxrt_core or libxrt_coreutil?
- Application should link with libxrt_coreutil to use external shim?
- how does coreutil knows which ishim interface to talk to?

src/runtime_src/core/common/module_loader.cpp
```diff
@@ -146,6 +147,29 @@ shim_path()
...
 }
 
+static std::vector<std::string>
+driver_plugin_paths()
+{
+  std::vector<std::string> ret;
+  std::vector<bfs::path> vec{
+    bfs::directory_iterator(shim_path().parent_path()),
+    bfs::directory_iterator()
+  };
```
stsoe, line 157: Do we really need a heap allocated vector for this?   
Other patterns in XRT use a while-loop, I see the appeal of this a set that can be iterated, new pattern to me for sure, but you should use std::array in this case.

src/runtime_src/core/common/module_loader.cpp
```diff
@@ -193,4 +217,13 @@ shim_loader()
   load_library(path.string());
 }
 
+driver_loader::
+driver_loader()
+{
+  auto paths = driver_plugin_paths();
+
+  for (auto& p : paths)
+    load_library(p);
```
stsoe, line 226:
```suggestion
  for (const auto& p : driver_plugin_paths())
    load_library(p);
```

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -168,12 +147,24 @@ get_num_dev_total(bool is_user) const
...
+  pci::register_driver(std::make_shared<pci::drv_xclmgmt>());
 
-  register_driver(std::make_shared<pci::drv_xocl>());
-  register_driver(std::make_shared<pci::drv_xclmgmt>());
+  // Load driver plug-ins. Don't need to die on a plug-in loading failure.
+  try {
+    xrt_core::driver_loader plugins;
+  }
+  catch (const std::runtime_error& err) {
+    std::cerr << "WARNING: " << err.what() << std::endl;
```
stsoe, line 159: Use our message system

maxzhen:
> Hi @maxzhen , Overall it looks good. I have couple of questions
> 
>     * You have mentioned in the PR description that driver plugin must register itself with libxrt_core.so. libxrt_core.so is mentioned in some other places in PR description, is it libxrt_core or libxrt_coreutil?

Driver plug-in should call xrt_core::pci::register_driver(). This function will put the driver in a list which is only known by libxrt_core, not libxrt_coreutil.

> 
>     * Application should link with libxrt_coreutil to use external shim?

Apps should link in the same way (with libxrt_coreutil). These changes are transparent to them.

> 
>     * how does coreutil knows which ishim interface to talk to?

As it does today, it should get an xrt_core::device from xrt_core::system and start using the ishim interface through this device object. Under the hood, the call will be handled by the driver plug-in (indirectly through object derived from xrt_core::device_linux). Libxrt_coreutil does not need to be aware of driver plug-ins specifically.

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -168,12 +147,24 @@ get_num_dev_total(bool is_user) const
...
+  }
+
+  for (auto driver : driver_list) {
```
stsoe, line 162:
```suggestion
  for (const auto& driver : driver_list) {
```

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -168,12 +147,24 @@ get_num_dev_total(bool is_user) const
 system_linux::
 system_linux()
 {
-  if (singleton)
-    throw std::runtime_error("More than one SHIM registered!");
-  singleton = this;
+  // Add built-in driver to the list.
+  pci::register_driver(std::make_shared<pci::drv_xocl>());
+  pci::register_driver(std::make_shared<pci::drv_xclmgmt>());
```
stsoe, line 152: It would make the code more readable if this just directly emplaced the driver_list.  I was scratching my head figuring out how the driver_list (iterated below) was populated.

stsoe (changes requested):

src/runtime_src/core/common/module_loader.cpp
maxzhen: I think a while-loop would do. Will remove this vector.

src/runtime_src/core/common/module_loader.cpp
maxzhen: Sure.

src/runtime_src/core/pcie/linux/system_linux.cpp
maxzhen: Sure.

src/runtime_src/core/pcie/linux/system_linux.cpp
maxzhen: Sure.

src/runtime_src/core/pcie/linux/system_linux.cpp
maxzhen: Sure. Also changed the comments a little bit. Hope it will make code more clearer.

stsoe (approved):

stsoe (approved):

src/runtime_src/core/common/module_loader.cpp
```diff
@@ -146,6 +147,27 @@ shim_path()
...
+driver_plugin_paths()
+{
+  std::vector<std::string> ret;
+  bfs::directory_iterator p{shim_path().parent_path()};
+
+  // All driver plug-ins are in the same directory as shim .so and with below prefix and suffix.
+  const std::string pre = "libxrt_driver_";
+  const std::string suf = std::string(".so.") + XRT_VERSION_MAJOR;
+  while (p != bfs::directory_iterator{}) {
+    std::string name = p->path().filename().string();
```
rozumx, line 160: This can be const and auto since the type is obvious.
```suggestion
    const auto name = p->path().filename().string();
```

src/runtime_src/core/common/module_loader.cpp
```diff
@@ -146,6 +147,27 @@ shim_path()
...
+    if (name.size() > pre.size() + suf.size() &&
```
rozumx, line 161: Helps the next developer to understand what is happening here.
```suggestion
    if (name.size() > (pre.size() + suf.size()) &&
```

maxzhen:
retest this please

rozumx (approved):
Code looks good.  
I have a few minor requests, but nothing to hold up the merge.

PR #7109, opened 2022-10-14
64-bit BO handle in ishim on Linux

maxzhen:

Today, a BO in Linux XRT is represented by a handle of xclBufferHandle type, which is essentially a 32-bit unsigned integer. This works fine since, so far, all BOs' book keeping data structure is maintained by xocl through DRM in Linux kernel. And this 32-bit handle is exactly the DRM BO handle passed up from xocl through entire XRT user space software stack. It is even exposed to end user via XRT's HAL APIs.

As support for new platforms are added to XRT, new drivers and SHIM implementation are introduced where a BO's book keeping data structure may start having user space part alongside the kernel part maintained by new driver through DRM. However, a 32-bit handle can only help to identify the kernel part, not the user part of the data structure. Hence, it is desired to change the handle to be a 64-bit integer or void *, so that it can hold:
+ a 32-bit DRM handle
+ an opaque pointer to user space data structure containing the DRM handle alongside other members for user space code

Or, in user space, we can maintain a map from a 32-bit integer to a pointer and keep using the 32-bit handle as-is. But this solution is not ideal because:
+ For each BO processing, say exec_buf(), we need to search two mapping tables one in user space and one in kernel via DRM. This is really inefficient.
+ It is good to have this flexibility to be able to hide the implementation detail in SHIM from the rest of software stack. Thus, we can keep the upper stack as OS and platform/driver independent.

In Windows XRT, xclBufferHandle has already been defined as void *. However, we can't just simply do the same in Linux since the xclBufferHandle is exposed to end user through HAL APIs, which we can't change. We do not support HAL API on Windows, so we don't have such restriction there.

Considering that HAL APIs are obsolete, we do not have to support them in the SHIM for new platforms. So, as long as we have a unified BO handle representation at ishim level, we are good. The new SHIM implementation will only extend ishim, not adding support for HAL APIs. And ishim is internal to XRT which can be changed at any time.

In this PR, a new BO handle type (xrt_buffer_handle) is introduced replacing xclBufferHandle in ishim API interface. The code using ishim will start using xrt_buffer_handle as BO handle. When the ishim implementation calls into existing SHIM on legacy platforms, xrt_buffer_handle will be explicitly converted to xclBufferHandle. There should be no change to existing SHIM/HAL APIs. When a BO is created via HAL APIs on existing platforms by xocl, its handle will be explicitly converted to xrt_buffer_handle right before it gets passed over to ishim users. So, ishim users should not see xclBufferHandle any more.

New SHIM and driver implementation will extend ishim and can convert xrt_buffer_handle to its own data type as required.

As mentioned above, xrt_buffer_handle can be defined as uint64_t or void *. But, the latter seems better because:
+ It is compatible with Windows' implementation
+ Compiler cannot perform implicit conversion between xrt_buffer_handle and xclBufferHandle, so I get to examine every place the conversion is needed to make sure I don't miss any.

N/A
See PR description for alternative solutions.
Some code calls both ishim and HAL APIs, hopefully the conversion is done properly for them, but there may still be bugs.
Some of these code are in current SHIM implementation, so does not apply to new platform anyway. Some of them are in utility code which part also does not apply to new platform.
All platform independent code should start using new type, if change is missed, it's a bug.
Built and installed XRT package and ran xbutil examine/validate on U55n.
N/A

### DIFF
--- a/src/runtime_src/core/common/api/hw_queue.cpp
+++ b/src/runtime_src/core/common/api/hw_queue.cpp
@@ -317,7 +317,7 @@ public:
   std::cv_status
   wait(size_t timeout_ms) override
   {
-    return m_device->wait_command(m_qhdl, XRT_NULL_BO, static_cast<int>(timeout_ms))
+    return m_device->wait_command(m_qhdl, XRT_INVALID_BUFFER_HANDLE, static_cast<int>(timeout_ms))
       ? std::cv_status::no_timeout
       : std::cv_status::timeout;
   }
--- a/src/runtime_src/core/common/api/kernel_int.h
+++ b/src/runtime_src/core/common/api/kernel_int.h
@@ -37,8 +37,8 @@ namespace xrt_core { namespace kernel_int {
 void
 copy_bo_with_kdma(const std::shared_ptr<xrt_core::device>& core_device,
                   size_t sz,
-                  xclBufferHandle dst_bo, size_t dst_offset,
-                  xclBufferHandle src_bo, size_t src_offset);
+                  xrt_buffer_handle dst_bo, size_t dst_offset,
+                  xrt_buffer_handle src_bo, size_t src_offset);
 
 XRT_CORE_COMMON_EXPORT
 std::vector<const xclbin::kernel_argument*>
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -142,7 +142,7 @@ public:
   static constexpr uint64_t no_addr = std::numeric_limits<uint64_t>::max();
   static constexpr uint32_t no_group = std::numeric_limits<uint32_t>::max();
   static constexpr bo::flags no_flags = static_cast<bo::flags>(std::numeric_limits<uint32_t>::max());
-  static constexpr xclBufferHandle null_bo = XRT_NULL_BO;
+  static constexpr xrt_buffer_handle null_bo = XRT_INVALID_BUFFER_HANDLE;
   static constexpr xclBufferExportHandle null_export = XRT_NULL_BO_EXPORT;
 
 private:
@@ -167,20 +167,20 @@ protected:
   // deliberately made protected, this is a file-scoped controlled API
   std::shared_ptr<xrt_core::device> device;     // NOLINT device where bo is allocated
   std::vector<std::shared_ptr<bo_impl>> clones; // NOLINT local m2m clones if any
-  xclBufferHandle handle = null_bo;             // NOLINT driver bo handle
+  xrt_buffer_handle handle = null_bo;           // NOLINT driver bo handle
   size_t size = 0;                              // NOLINT size of buffer
   mutable uint64_t addr = no_addr;              // NOLINT bo device address
-  mutable uint32_t grpid = no_group;             // NOLINT memory group index
+  mutable uint32_t grpid = no_group;            // NOLINT memory group index
   mutable bo::flags flags = no_flags;           // NOLINT flags per bo properties
   mutable xclBufferExportHandle export_handle = null_export; // NOLINT export handle if exported
   bool free_bo;                                 // NOLINT should dtor free bo
 
 public:
   explicit bo_impl(size_t sz)
-    : handle(XRT_NULL_BO), size(sz), free_bo(false)
+    : handle(XRT_INVALID_BUFFER_HANDLE), size(sz), free_bo(false)
   {}
 
-  bo_impl(xclDeviceHandle dhdl, xclBufferHandle bhdl, size_t sz)
+  bo_impl(xclDeviceHandle dhdl, xrt_buffer_handle bhdl, size_t sz)
     : device(xrt_core::get_userpf_device(dhdl)), handle(bhdl), size(sz), free_bo(true)
   {}
 
@@ -242,7 +242,7 @@ public:
     clones.push_back(std::move(clone));
   }
 
-  xclBufferHandle
+  xrt_buffer_handle
   get_xcl_handle() const
   {
     return handle;
@@ -564,7 +564,7 @@ class buffer_ubuf : public bo_impl
   void* ubuf;
 
 public:
-  buffer_ubuf(xclDeviceHandle dhdl, xclBufferHandle bhdl, size_t sz, void* buf)
+  buffer_ubuf(xclDeviceHandle dhdl, xrt_buffer_handle bhdl, size_t sz, void* buf)
     : bo_impl(dhdl, bhdl, sz)
     , ubuf(buf)
   {}
@@ -586,7 +586,7 @@ class buffer_hbuf : public bo_impl
   xrt_core::aligned_ptr_type hbuf;
 
 public:
-  buffer_hbuf(xclDeviceHandle dhdl, xclBufferHandle bhdl, size_t sz, xrt_core::aligned_ptr_type&& b)
+  buffer_hbuf(xclDeviceHandle dhdl, xrt_buffer_handle bhdl, size_t sz, xrt_core::aligned_ptr_type&& b)
     : bo_impl(dhdl, bhdl, sz), hbuf(std::move(b))
   {}
 
@@ -606,7 +606,7 @@ class buffer_kbuf : public bo_impl
   void* hbuf;
 
 public:
-  buffer_kbuf(xclDeviceHandle dhdl, xclBufferHandle bhdl, size_t sz)
+  buffer_kbuf(xclDeviceHandle dhdl, xrt_buffer_handle bhdl, size_t sz)
     : bo_impl(dhdl, bhdl, sz), hbuf(device->map_bo(handle, true))
   {}
 
@@ -714,7 +714,7 @@ public:
 class buffer_dbuf : public bo_impl
 {
 public:
-  buffer_dbuf(xclDeviceHandle dhdl, xclBufferHandle bhdl, size_t sz)
+  buffer_dbuf(xclDeviceHandle dhdl, xrt_buffer_handle bhdl, size_t sz)
     : bo_impl(dhdl, bhdl, sz)
   {}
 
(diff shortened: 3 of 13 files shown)


dayeh-xilinx:
retest this please.

chvamshi-xilinx (approved):
Looks good. Description of this PR is very good.

src/runtime_src/core/include/xrt/xrt_bo.h
```diff
@@ -44,7 +44,7 @@ typedef uint32_t xrtMemoryGroup;
  * @details
  * Use when constructing xrt::bo from xclBufferHandle
  */
-struct xcl_buffer_handle { xclBufferHandle bhdl; };
+struct xcl_buffer_handle { xrt_buffer_handle bhdl; };
```
stsoe, line 47: Oops, we can't do this.  This is a binary incompatibility in an end user exposed API.
Granted the API that uses this type is undocumented and explicitly states that it's a bad idea to mix xcl objects with xrt objects, but still we need to think how to make this not break the ABI.

stsoe (changes requested):
This looks great overall.   But there is one problem as I noted.
I didn't think through a solution.
...
I think you can revert this one change, and do conversion from xclBufferHandle to xrt_buffer_handle in the implementation of the xrt::bo constructor.

src/runtime_src/core/include/xrt/xrt_bo.h
maxzhen: Updated the PR with the fix we had discussed. Please take a look.

stsoe (approved):
Looks good, once build passes let me run through the OpenCL suite of regressions.

src/runtime_src/core/common/api/xrt_bo.cpp
```diff
@@ -1171,7 +1173,8 @@ fill_copy_pkt(const xrt::bo& dst, const xrt::bo& src, size_t sz,
 #ifndef _WIN32
```
rozumx, line 1173: At a later date, the code should be refactored to call a common XRT helper routine to do this functionality.   
As it currently is implemented, this flow is not support to run on windows.

src/runtime_src/core/common/api/kernel_int.h
```diff
@@ -37,8 +37,8 @@ namespace xrt_core { namespace kernel_int {
 void
```
rozumx, line 37: The copyright date (line 2) needs to be updated.

src/runtime_src/core/common/ishim.h
```diff
@@ -335,52 +335,52 @@ struct shim : public DeviceType
...
   xclBufferExportHandle
-  export_bo(xclBufferHandle bo) const override
+  export_bo(xrt_buffer_handle bo) const override
   {
-    auto ehdl = xclExportBO(DeviceType::get_device_handle(), bo);
+    auto ehdl = xclExportBO(DeviceType::get_device_handle(), to_xclBufferHandle(bo));
     if (ehdl == XRT_NULL_BO_EXPORT)
       throw system_error(EINVAL, "Unable to export BO");
     if (ehdl < 0) // system error code
       throw system_error(ENODEV, "Unable to export BO");
```
rozumx, line 371: This error is different from line 369, yet the message is the same.  One might wish to update the message to be more relevant.

src/runtime_src/core/common/ishim.h
```diff
@@ -391,38 +391,39 @@ struct shim : public DeviceType
   }
 
   void
-  copy_bo(xclBufferHandle dst, xclBufferHandle src, size_t size, size_t dst_offset, size_t src_offset) override
+  copy_bo(xrt_buffer_handle dst, xrt_buffer_handle src, size_t size, size_t dst_offset, size_t src_offset) override
   {
-    if (auto err = xclCopyBO(DeviceType::get_device_handle(), dst, src, size, dst_offset, src_offset))
+    if (auto err = xclCopyBO(DeviceType::get_device_handle(),
```
rozumx, line 396: There is a lot going on in this line, please consider reformatting it.
```suggestion
    auto err = xclCopyBO(DeviceType::get_device_handle(),to_xclBufferHandle(dst), to_xclBufferHandle(src), size, dst_offset, src_offset);
    if (err)
```

src/runtime_src/core/common/ishim.h
```diff
@@ -391,38 +391,39 @@ struct shim : public DeviceType
...
+      to_xclBufferHandle(dst), to_xclBufferHandle(src), size, dst_offset, src_offset))
       throw system_error(err, "unable to copy BO");
   }
 
   void
-  sync_bo(xclBufferHandle bo, xclBOSyncDirection dir, size_t size, size_t offset) override
+  sync_bo(xrt_buffer_handle bo, xclBOSyncDirection dir, size_t size, size_t offset) override
   {
-    if (auto err = xclSyncBO(DeviceType::get_device_handle(), bo, dir, size, offset))
+    if (auto err = xclSyncBO(DeviceType::get_device_handle(), to_xclBufferHandle(bo), dir, size, offset))
```
rozumx, line 404: See earlier comment regarding reformatting the line.

src/runtime_src/core/include/deprecated/xrt.h
```diff
@@ -80,6 +80,34 @@ typedef unsigned int xclBufferHandle;
...
 
+/*
+ * typedef xrt_buffer_handle - opaque buffer handle
+ *
+ * A buffer handle of xrt_buffer_handle kind is obtained by allocating buffer
+ * objects through ISHIM API. The buffer handle is used by XRT ISHIM APIs
+ * that operate on on buffer objects.
+ */
+typedef void * xrt_buffer_handle;
+#define XRT_INVALID_BUFFER_HANDLE	NULL
```
rozumx, line 91: Hmm...would `nullptr `be better here. 

Note: If the code is used in C code then the answer would be no :^).

src/runtime_src/core/common/api/hw_queue.cpp
```diff
@@ -317,7 +317,7 @@ class qds_device : public hw_queue_impl
   std::cv_status
   wait(size_t timeout_ms) override
   {
-    return m_device->wait_command(m_qhdl, XRT_NULL_BO, static_cast<int>(timeout_ms))
+    return m_device->wait_command(m_qhdl, XRT_INVALID_BUFFER_HANDLE, static_cast<int>(timeout_ms))
```
rozumx, line 320: Hmm, would `reinterpret_cast `work here instead, being that `size_t` is signed.
```suggestion
    return m_device->wait_command(m_qhdl, XRT_INVALID_BUFFER_HANDLE, reinterpret_cast<int>(timeout_ms))
```

src/runtime_src/core/common/api/xrt_bo.cpp
```diff
@@ -854,8 +856,8 @@ class buffer_sub : public bo_impl
 class buffer_xbuf : public bo_impl
 {
 public:
-  buffer_xbuf(xclDeviceHandle dhdl, xclBufferHandle bhdl)
-    : bo_impl(dhdl, xcl_buffer_handle{bhdl})
+  buffer_xbuf(xclDeviceHandle dhdl, xcl_buffer_handle xhdl)
+    : bo_impl(dhdl, xhdl)
   {}
```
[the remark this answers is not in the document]
stsoe: I don't like this, empty is empty and `{}` is clearly deliberate.  Most constructor bodies should be empty anyway.

src/runtime_src/core/common/api/xrt_bo.cpp
stsoe, line 1173: Sorry, didn't mean to remove Stephen's comment.   He said this should be refactored to call common code as it is not supported on Windows.  I wanted to reply, I didn't understand the comment?

src/runtime_src/core/include/deprecated/xrt.h
stsoe: The answer is no for exactly the reason you mention ;-)

src/runtime_src/core/common/api/hw_queue.cpp
stsoe: Oh no, static_cast is what you want here, it declares the intent of narrowing the type.  Don't use reinterrpret_cast unless you really have to.

src/runtime_src/core/common/api/xrt_bo.cpp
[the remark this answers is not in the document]
maxzhen: OK, not going to change.

src/runtime_src/core/common/bo_cache.h
```diff
@@ -39,7 +39,7 @@ class bo_cache {
   // Helper typedef for std::pair. Note the elements are const so that the
```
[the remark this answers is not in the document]
maxzhen: Fixed.

src/runtime_src/core/common/api/kernel_int.h
maxzhen: Fixed.

src/runtime_src/core/common/ishim.h
maxzhen: Fixed both line 369 and 380 to mention bad handle.

src/runtime_src/core/common/ishim.h
maxzhen: Broke it into two lines.

src/runtime_src/core/common/ishim.h
maxzhen: Broke into two lines.

src/runtime_src/core/common/api/xrt_bo.cpp
```diff
@@ -167,20 +167,20 @@ class bo_impl
   // deliberately made protected, this is a file-scoped controlled API
```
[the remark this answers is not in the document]
maxzhen: @stsoe will fix them in a follow-up PR. Not going to change it in this pull request, then.

vboggara-xilinx (approved):

PR #7114, opened 2022-10-18
One command monitor thread per xrt_core::device

stsoe:
Ensure that only one legacy command monitor thread (kds_device) is created for platforms that do not implement create_hw_queue.

The legacy command monitor handles command completion for the enqueued command execution, mostly OpenCL or native XRT with callback, latter rarely used.  The APIs used by the monitor are legacy execbuf and execwait() APIs that operate at the device level.

Testing of #7712 found that when a platform supports hardware contexts, but not hardware queue, it ends up creating multiple kds_devices (command monitor threads) one for each unique hwctx handle.  The command monitor does not support a single thread creating multiple instances and also it is wasteful to use multiple monitor threads for same device when execbuf and execwait calls go through the xrt_core::device object anyway.

This PR thus adds logic to ensure that only one kds_device is created for each xrt_core::device.

For platforms that support hardware queue, new queue specific APIs are used to manage command execution and as such require a command monitor per hardware queue, so no changes are made here.  Only legacy non hardware queue flow is changed.

### DIFF
--- a/src/runtime_src/core/common/api/hw_queue.cpp
+++ b/src/runtime_src/core/common/api/hw_queue.cpp
@@ -508,8 +508,24 @@ namespace {
 // For time being there is only one hw_queue per hw_context
 // Use static map with weak pointers to implementation.
 using hwc2hwq_type = std::map<xcl_hwctx_handle, std::weak_ptr<xrt_core::hw_queue_impl>>;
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx (approved):

chvamshi-xilinx:
retest this please

vboggara-xilinx (approved):

PR #7119, opened 2022-10-21
Improve performance of exec_wait

stsoe:
Use a per-device thread for calling xrt_core::device::exec_wait. The thread synchronizes with waiting execution threads notifying all threads upon any command completing per exec_wait.

Performance testing of multiple threads running simple [run, wait] loops of same kernel run object showed that a multi-threaded application with n-threads had poor performance over n-process single thread application.  The problem was narrowed down to a strange combination of mutex lock interacting with waiting threads to ensure mutual exclusion when calling exec_wait.  The waiting treated a mutex lock more or less as queue when in fact it doesn't at all behave a such and some threads got starved out waiting in vain.

The change in this PR moves exec_wait into a dedicated thread that calls exec_wait and notifies all waiting threads upon any command completing.  This leads to all waiting threads simultanously getting kicked to go check the status of the command they are waiting for.  If a threads command is not complete, the thread simply waits to get notified again.

Complete revamp of exec_wait command completion so quite risky.

OpenCL regression test and misc XRT tests, validate iops, etc.

### DIFF
--- a/src/runtime_src/core/common/api/hw_queue.cpp
+++ b/src/runtime_src/core/common/api/hw_queue.cpp
@@ -16,12 +16,16 @@
 #include "core/include/ert.h"
 #include "core/include/xcl_hwqueue.h"
 
+#include <atomic>
 #include <algorithm>
+#include <chrono>
 #include <condition_variable>
 #include <map>
 #include <memory>
 #include <mutex>
 #include <thread>
+
+using namespace std::chrono_literals;
 
 ////////////////////////////////////////////////////////////////
 // Main command execution interface for scheduling commands for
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/hw_queue.cpp
```diff
@@ -478,13 +549,20 @@ class kds_device : public hw_queue_impl
...
       if (exec_wait(timeout_ms) == std::cv_status::timeout)
         return std::cv_status::timeout;
     }
 
+    // One command is completed, give the runner a chance
+    // for entering sleep at next iteration.  This variable
+    // update doesn't have to be modified under a mutex lock
+    // since the exec_wait_runner will at most complete one
+    // more loop before noticing the change.
+    --m_commands_in_flight;
```
maxzhen, line 564: What if there are more than one thread sharing the same device object and one is doing submit() and one is doing wait()? Will this create a race condition?

stsoe:
Obsolete.

PR #7128, opened 2022-10-25
CR-1140807 PLM Reports from VMR

prapulkrishnamurthy:
Feature to forward the PLM logs from VMR to Host

NA

Improvised API version of same feature need to be implemented.
Both Syst DTB and PLM sysfs should be able to poll for data instead of sending command at VMR_Probe.
Buffer optimizations can be done for Bin attribute nodes.

The sysfs dump include dummy residual bytes from RPU shared memory at the end of full PLM data untill 16kth byte.
This is not optimized and left for user end to decide.

Testing along side VMR involves sysfs dump for this node.
The output looks like attached.
image

NA

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
@@ -2994,13 +2994,9 @@ static int xgq_vmr_probe(struct platform_device *pdev)
 	xgq->xgq_cmd_id = 0;
 	xgq->xgq_vmr_system_dtb = NULL;
 	xgq->xgq_vmr_system_dtb_size = 0;
-<<<<<<< HEAD
 	xgq->xgq_halted = true;
-=======
 	xgq->xgq_vmr_plm_sync = NULL;
 	xgq->xgq_vmr_plm_sync_size = 0;
-
->>>>>>> CR-1140807_PLM Initial Draft
 
 	mutex_init(&xgq->xgq_lock);
 	mutex_init(&xgq->clk_scaling_lock);
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
@@ -2741,6 +2741,7 @@ static ssize_t vmr_system_dtb_read(struct file *filp, struct kobject *kobj,
 	if (off + count > size)
 		count = size - off;
 	memcpy(buf, blob + off, count);
+	mutex_unlock(&xgq->xgq_lock);
 
 	ret = count;
 out:
@@ -2772,7 +2773,7 @@ static ssize_t vmr_plm_sync_read(struct file *filp, struct kobject *kobj,
 	mutex_lock(&xgq->xgq_lock);
 	blob = xgq->xgq_vmr_plm_sync;
 	size = xgq->xgq_vmr_plm_sync_size;
-	mutex_unlock(&xgq->xgq_lock);
+	
 
 	if (off >= size)
 		goto out;
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
@@ -2734,8 +2734,10 @@ static ssize_t vmr_system_dtb_read(struct file *filp, struct kobject *kobj,
 	blob = xgq->xgq_vmr_system_dtb;
 	size = xgq->xgq_vmr_system_dtb_size;
 
-	if (off >= size)
+	if (off >= size){
+		mutex_unlock(&xgq->xgq_lock);
 		goto out;
+	}
 
 	if (off + count > size)
 		count = size - off;
@@ -2744,7 +2746,6 @@ static ssize_t vmr_system_dtb_read(struct file *filp, struct kobject *kobj,
 
 	ret = count;
 out:
-	mutex_unlock(&xgq->xgq_lock);
 	return ret;
 }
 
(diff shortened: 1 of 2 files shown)


stsoe:
Add to whitelist please

stsoe:
Ok to test

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -2637,6 +2645,44 @@ static struct bin_attribute bin_attr_vmr_system_dtb = {
...
+	struct xocl_xgq_vmr *xgq =
+		dev_get_drvdata(container_of(kobj, struct device, kobj));
+	unsigned char *blob;
+	size_t size;
+	ssize_t ret = 0;
+
+	mutex_lock(&xgq->xgq_lock);
+	blob = xgq->xgq_vmr_plm_sync;
+	size = xgq->xgq_vmr_plm_sync_size;
+	mutex_unlock(&xgq->xgq_lock);
```
maxzhen, line 2660: What if xgq->xgq_vmr_plm_sync and xgq->xgq_vmr_plm_sync_size changed right after this line? Is it still safe to use them at line 2667?

maxzhen (changes requested):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
xdavidz: good catch, we should fix this and the system_dtb sysfs. 
in real case, this is reading some logs or static device tree metadata back from VMR devices, so unlikely it will break anything, but still lets move the code below into the mutex. Say all the way too the memcpy at 2667.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -2924,6 +2978,14 @@ static int xgq_vmr_probe(struct platform_device *pdev)
 		XGQ_WARN(xgq, "cannot load system dtb from shell, ret: %d", ret);
 		ret = 0;
 	}
+
+	ret = xgq_log_page_plm_sync(pdev, &xgq->xgq_vmr_plm_sync,
```
xdavidz, line 2982: This is wrong location, we should not query the plm only once when driver probe.
We actually should query it from the sysfs node read.

This is the difference between system_dtb and plm_log. 

I think even the system_dtb should not be in the probe. I will fix this issue soon.

xdavidz: Overall, seems ok to me. You can leave the code as is. I will refactor this and system_dtb together.

xdavidz (changes requested):
please fix the issue commented by Max.

prapulkrishnamurthy:
retest this please

maxzhen (approved):

stsoe (approved):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -2725,14 +2733,14 @@ static ssize_t vmr_system_dtb_read(struct file *filp, struct kobject *kobj,
 	mutex_lock(&xgq->xgq_lock);
 	blob = xgq->xgq_vmr_system_dtb;
 	size = xgq->xgq_vmr_system_dtb_size;
-	mutex_unlock(&xgq->xgq_lock);
 
 	if (off >= size)
 		goto out;
```
xdavidz, line 2738: if goto out, you should release the lock

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -2752,6 +2760,44 @@ static struct bin_attribute bin_attr_vmr_system_dtb = {
...
+		dev_get_drvdata(container_of(kobj, struct device, kobj));
+	unsigned char *blob;
+	size_t size;
+	ssize_t ret = 0;
+
+	mutex_lock(&xgq->xgq_lock);
+	blob = xgq->xgq_vmr_plm_sync;
+	size = xgq->xgq_vmr_plm_sync_size;
+
+	if (off >= size)
```
xdavidz, line 2776: same here, if goto out, you should release the lock

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -2752,6 +2761,45 @@ static struct bin_attribute bin_attr_vmr_system_dtb = {
...
+		goto out;
+
+	if (off + count > size)
+		count = size - off;
+	memcpy(buf, blob + off, count);
+	mutex_unlock(&xgq->xgq_lock);
```
xdavidz, line 2783: here, u will unlock twice. it will cause potential undefined behavior or make debugging harder. Please clean up your code.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq.c
```diff
@@ -2725,17 +2733,18 @@ static ssize_t vmr_system_dtb_read(struct file *filp, struct kobject *kobj,
...
 	size = xgq->xgq_vmr_system_dtb_size;
-	mutex_unlock(&xgq->xgq_lock);
 
 	if (off >= size)
 		goto out;
 
 	if (off + count > size)
 		count = size - off;
 	memcpy(buf, blob + off, count);
+	mutex_unlock(&xgq->xgq_lock);
```
xdavidz, line 2743: here, u will unlock twice. it will cause potential undefined behavior or make debugging harder. Please clean up your code.

xdavidz (approved):

PR #7132, opened 2022-10-26
Force xrt.h to be C compatible

maxzhen:

Make sure xrt.h is C compatible, always.
N/A
Added a tiny C program including xrt.h and compile it during build of XRT.
Low
Build XRT with good and bad xrt.h to make sure the added test works.
N/A


maxzhen:
Sigh...looks like xrt.h is not C compatible on Windows. I'll only enable this test on Linux.

src/runtime_src/core/include/unittests/xrt_h_c_compilation/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+# SPDX-License-Identifier: Apache-2.0                                           
+# Copyright (C) 2022, Advanced Micro Devices, Inc. All rights reserved.
+
+add_executable(main main.c)
+target_include_directories(main
+  PRIVATE
+  ${CMAKE_CURRENT_SOURCE_DIR}/../../
+  )
```
stsoe, line 8:
```suggestion
```
Remove, the default include search path when building XRT has `<PATH>/src/runtime_src`

stsoe (changes requested):

src/runtime_src/core/include/unittests/xrt_h_c_compilation/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+# SPDX-License-Identifier: Apache-2.0                                           
+# Copyright (C) 2022, Advanced Micro Devices, Inc. All rights reserved.
+
+add_executable(main main.c)
```
stsoe, line 4:
```suggestion
add_library(unit_cpps OBJECT main.c)
```
You only care about compiling the object file, not linking the executable, so I think this will do?

src/runtime_src/core/include/unittests/xrt_h_c_compilation/main.c
```diff
@@ -0,0 +1,14 @@
...
+
+/*
+ * Test C compilation with xrt.h header file
+ */
+#include "core/include/xrt.h"
+
+int main()
+{
+  return 0;
+}
```
stsoe, line 14:
```suggestion
```
No need for main() if all we do is compile  objects.

src/runtime_src/core/include/unittests/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (C) 2022, Advanced Micro Devices, Inc. All rights reserved.
+
+if (NOT WIN32)
+  add_subdirectory(xrt_h_c_compilation)
```
stsoe, line 5: I suggest you rename `xrt_h_c_compilation` to `xrt_c_compilation`.  I have several other header files I would like to add to this test.

stsoe (changes requested):

stsoe (approved):

PR #7001, opened 2022-09-07
Fix build issues when creating deb packages

hkhonming:
#### Problem solved by the commit
Fixed build failures for build_deb_edge.sh

#### How problem was solved, alternative solutions (if any) and why they were rejected
1. debian packaging skills is outdated and I just synchronise it with the latest packaging script
2. Fix the bug in dkms-edge.cmake introduced after the removal of cu_plram.c
3. Update to build on jammy by default 

#### Risks (if any) associated the changes in the commit
Most changes are specific to debian packaging and so it should be safe for other usage. The dkms-edge.cmake is an obvious bug after a removal of the cu_plram.c and so the fix is pretty straightforward too.

#### What has been tested and how, request additional testing if necessary
Tested to generate .deb file in 2022.1 branch head. If we need to cherrypick the change to master branch. Probably we need to test it too since I am not sure if there are other changes in master branch that require packaging script change.

#### Documentation impact (if any)
If there are document on how to create .deb files. We should update it and remove the need of the sysroot, as well as mentioning we are building jammy by default now.

PR #7148, opened 2022-11-02
CR-1121893 - Fix debian packages creation for Edge(zynqmp) platforms

rbramand-xilinx:
Fixed build failure in creation of debian packages for edge platforms using sbuild mechanism.

Discovered in pipeline regression

As XRT has standard packages as dependencies, instead of creating sysroot from custom tar and providing it as an input to build script we have added changes to use standard mechanism. 
Setting up environment and creation of sysroot, creation of packages are explained in below confluence page:
https://confluence.xilinx.com/display/DCG/Creation+of+Debian+packages+for+Edge+Platforms+using+sbuild

Low

Tested creation of debian packages on 20.04 and 22.04 Ubuntu machines using both jammy and focal sysroots.

Example of created packages format:
1. xrt_embedded_202310.2.15.0_20.04-arm64.deb  - user space package
2. xrt_zocl_dkms_202310.2.15.0_20.04-arm64.deb  - zocl driver package which is built on target using dkms flow (like DC platform packages)

We need to discuss on updating xrt docs with the build steps.

### DIFF
--- a/build/build_edge_deb.sh
+++ b/build/build_edge_deb.sh
@@ -14,7 +14,7 @@ usage()
     echo "  options:"
     echo "          -help                           Print this usage"
     echo "          -aarch                          Architecture <aarch32/aarch64>"
-    echo "          -sysroot_name                   Name of sysroot folder created for compilation"
+    echo "          -dist                           Distribution for the package build <focal/jammy>"
     echo "          -clean, clean                   Remove build directories, pass 'aarch' option to it"
     echo ""
 }
@@ -30,13 +30,16 @@ SAVED_OPTIONS=$(set +o)
 set +x
 # Get the canonical file name of the current script
 THIS_SCRIPT=`readlink -f ${BASH_SOURCE[0]}`
- 
+
 PROGRAM=`basename $0`
 
 # XRT Version variables
 XRT_MAJOR_VERSION=2
-XRT_MINOR_VERSION=13
-RELEASE_VERSION=202210
+XRT_MINOR_VERSION=15
+RELEASE_VERSION=202310
+
+# Default distribution
+DIST=jammy
 
 # Pick XRT_VERSION_PATCH from Env variable
 if [ -z $XRT_VERSION_PATCH ]; then
@@ -57,13 +60,13 @@ while [ $# -gt 0 ]; do
                 -help | --help )
                         usage_and_exit 0
                         ;;
+                -dist | --dist )
+                        shift
+                        DIST=$1
+                        ;;
                 -aarch | --aarch )
                         shift
                         AARCH=$1
-                        ;;
-                -sysroot_name | --sysroot_name )
-                        shift
-                        SYSROOT=$1
                         ;;
                 -clean | clean | --clean )
                         clean=1
@@ -81,6 +84,18 @@ done
 THIS_SCRIPT_DIR=$(dirname "$THIS_SCRIPT")
 XRT_DIR=`readlink -f $THIS_SCRIPT_DIR/../`
 DEBIAN=`readlink -f $THIS_SCRIPT_DIR/debian`
+
+if [ -z $DIST ]; then
+    error "-dist is required option"
+fi
+
+if [[ $DIST == "jammy" ]]; then
+    OS_VERSION="22.04"
+elif [[ $DIST == "focal" ]]; then
+    OS_VERSION="20.04"
+else
+    error "$DIST is not a valid dist option"
+fi
 
 if [ -z $AARCH ]; then
     error "-aarch is required option"
@@ -101,11 +116,6 @@ if [[ $clean == 1 ]]; then
     exit 0
 fi
 
-# Sanity Check
-if [ -z $SYSROOT ] ; then
-    error "Please provide the required option 'sysroot_name'"
-fi
-
 if [ ! -d $DEBIAN ]; then
     error "$DEBIAN is not accessible"
 fi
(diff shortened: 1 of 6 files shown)


build/build_edge_deb.sh
```diff
@@ -30,13 +30,16 @@ SAVED_OPTIONS=$(set +o)
...
- 
+
 PROGRAM=`basename $0`
 
 # XRT Version variables
 XRT_MAJOR_VERSION=2
-XRT_MINOR_VERSION=13
-RELEASE_VERSION=202210
+XRT_MINOR_VERSION=15
+RELEASE_VERSION=202310
```
keryell, line 39: I have the feeling that these values are more general than the scope of this file and should come from a single-truth file.

build/build_edge_deb.sh
```diff
@@ -82,6 +85,18 @@ THIS_SCRIPT_DIR=$(dirname "$THIS_SCRIPT")
...
+if [ -z $DIST ]; then
+    error "-dist is required option"
+fi
+
+if [[ $DIST == "jammy" ]]; then
+    OS_VERSION="22.04"
+elif [[ $DIST == "focal" ]]; then
+    OS_VERSION="20.04"
+else
+    error "$DIST is not a valid dist option"
```
keryell, line 97: Why? I want my freedom!

build/build_edge_deb.sh
```diff
@@ -82,6 +85,18 @@ THIS_SCRIPT_DIR=$(dirname "$THIS_SCRIPT")
...
 DEBIAN=`readlink -f $THIS_SCRIPT_DIR/debian`
 
+if [ -z $DIST ]; then
+    error "-dist is required option"
+fi
+
+if [[ $DIST == "jammy" ]]; then
+    OS_VERSION="22.04"
+elif [[ $DIST == "focal" ]]; then
+    OS_VERSION="20.04"
```
keryell, line 95: All this looks very ad-hoc and we should use some Debian/Ubuntu recipe for this.

keryell (changes requested):
That looks good but could be generalized.
@Ralender can you try this?

keryell:
@hkhonming could you (or someone else from Canonical) review this, as it seems related to https://github.com/Xilinx/XRT/pull/7001 ?

rbramand-xilinx:
> @hkhonming could you (or someone else from Canonical) review this, as it seems related to #7001 ?

Yes this PR is a clone of pr #7001, but I have made some changes on top of it.

build/build_edge_deb.sh
rbramand-xilinx: At present we tested only on these two versions.

hkhonming: I agree with keryell's comment. Given the context here that the script is used to create a Debian packages, we should allow the user to specify any possible Debian/Ubuntu distribution which is known to sbuild. So probably we could just delegate the distribution name checking to sbuild.

keryell (commented):
Almost there!

build/build_edge_deb.sh
```diff
@@ -82,6 +85,18 @@ THIS_SCRIPT_DIR=$(dirname "$THIS_SCRIPT")
 XRT_DIR=`readlink -f $THIS_SCRIPT_DIR/../`
 DEBIAN=`readlink -f $THIS_SCRIPT_DIR/debian`
 
+if [ -z $DIST ]; then
```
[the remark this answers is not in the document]
rbramand-xilinx: I will fix it, thanks @keryell

rbramand-xilinx:
We are adding OS_VERSION to generated debs to make them similar to all our other packages (dc platforms)
eg of deb package we already ship - xrt_202220.2.14.365_**18.04**-amd64-xrt.deb

src/CMakeLists.txt
```diff
@@ -92,10 +92,10 @@ set (CMAKE_INSTALL_DEFAULT_COMPONENT_NAME "xrt")
 #set (XRT_DEV_COMPONENT "xrt-dev")
 set (XRT_DEV_COMPONENT "xrt")
 
-# Version adjusted to 2.15 for 2023.1
-set(XRT_VERSION_RELEASE 202310)
-SET(XRT_VERSION_MAJOR 2)
-SET(XRT_VERSION_MINOR 15)
+# Set XRT version variables passed from command line
```
chvamshi-xilinx, line 95: How does this work for windows? are these variable set in windows?

build/build.sh
```diff
@@ -207,6 +207,14 @@ edge_dir=${EDGE_DIR:-Edge}
 # Disable with '-disable-werror' option.
 cmake_flags+=" -DXRT_ENABLE_WERROR=$werror"
 
+# Add XRT version information
+# XRT_VERSION_RELEASE XRT_VERSION_MAJOR XRT_VERSION_MINOR are read from xrt_version.txt file
+# XRT_VERSION_PATCH is read from environment variable from cmake file
+
+XRT_VERSION_FILE=$BUILDDIR/xrt_version.txt
+source $XRT_VERSION_FILE
+cmake_flags+=" -Dxrt_version_release=${XRT_VERSION_RELEASE} -Dxrt_version_major=${XRT_VERSION_MAJOR} -Dxrt_version_minor=${XRT_VERSION_MINOR}"
```
chvamshi-xilinx, line 216: windows build need to use similar approach.  how does build_edge.sh gets the XRT version? 
It is better to define these variables in cmake itself so that, things will work even if someone calls cmake directly instead of build.sh.

chvamshi-xilinx (changes requested):

build/build_edge_deb.sh
rbramand-xilinx: Hi @keryell, there many flows and maintaining single xrt version file is not working out for all flows as some use cmake and some flows use shell scripts. 
The variables in this script are needed for changelog file.

build/build.sh
rbramand-xilinx: Hi @chvamshi-xilinx, made changes as suggested please review.

chvamshi-xilinx (approved):
Looks good.

keryell (approved):
Thanks!

stsoe (approved):

PR #7153, opened 2022-11-04
Fixed the memory manager issue

saifuddin-xilinx:
[CR-1144880](https://jira.xilinx.com/browse/CR-1144880)
Xbutil validate hangs on U55n/U55c/U50

Validation test is failing for U50 devices. 
This issue is introduced For multislot memory manager changes. 

Rootcause: 
xocl_addr_translator_enable_remap will fail if base address is "0". We should not initialize address translation for host memory BANK if it is not used. Also the host memory bank address translation should be done for each xclbin, hence we need to do this under memory initialization.

Fixed this issue.


saifuddin-xilinx:
retest this please

dayeh-xilinx:
retest this please.

maxzhen (approved):

PR #7155, opened 2022-11-06
system_dtb and plm_log are not being updated everytime

xdavidz:

This fixes several issues during recent code review.
1) xgq.c is confusing, it should be xgq_vmr.c which is the same subdev name we used.
2) plm_sync is confusing, it is actually plm_log we collect from the vmr.
3) we should not only load device tree (system_dtb) and (plm_log) during probe time, instead, we should load up-to-date data every time user requests it.

CR-1135130
tested with v70 shell

https://confluence.xilinx.com/pages/viewpage.action?pageId=302345261#XGQ1.0CommandandQueueSpecification(WIP)-GetLogPage


maxzhen (approved):
Finally, it's not xgq.c any longer :-).

PR #7157, opened 2022-11-08
Xrt mm changes

saifuddin-xilinx:
Fixed memory manager issue for multislot. 
For multi slot changes some of the issue introduced.
This PR consists of following  fixes:
1) Memory topology is not consistent between multiple xclbin 
re-initialize the memory manager if xclbin memory topology changes. 
2) Static memory to maintain stats of memory topology
3) Avoid initialization of dummy memory banks entry (size=0)
4) PS kernel validate issue fix

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
@@ -659,40 +659,23 @@ void xocl_drm_fini(struct xocl_drm *drm_p)
 void xocl_mm_get_usage_stat(struct xocl_drm *drm_p, u32 ddr,
 	struct drm_xocl_mm_stat *pstat)
 {
-	struct xocl_mm *xocl_mm = drm_p->xocl_mm;
-	struct drm_xocl_mm_stat *mm_stat = NULL;
-
-	if (!xocl_mm) {
-        	xocl_err(drm_p->ddev->dev, "Invalid memory manager");
-		return;
-	}
+	struct drm_xocl_mm_stat *mm_stat = &drm_p->mm_usage_stat[ddr];
 
 	if (!pstat) {
         	xocl_err(drm_p->ddev->dev, "Invalid memory stats");
 		return;
 	}
 
-	mm_stat = &drm_p->mm_usage_stat[ddr];
-
-	if (mm_stat->is_used) {
-		pstat->memory_usage = mm_stat ? mm_stat->memory_usage : 0;
-		pstat->bo_count = mm_stat ? mm_stat->bo_count : 0;
-	}
+	pstat->memory_usage = mm_stat->is_used ? mm_stat->memory_usage : 0;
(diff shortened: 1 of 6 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
```diff
@@ -672,9 +672,12 @@ void xocl_mm_get_usage_stat(struct xocl_drm *drm_p, u32 ddr,
 		return;
 	}
 
-	mm_stat = xocl_mm->mm_usage_stat[ddr];
-	pstat->memory_usage = mm_stat ? mm_stat->memory_usage : 0;
-	pstat->bo_count = mm_stat ? mm_stat->bo_count : 0;
+	mm_stat = &drm_p->mm_usage_stat[ddr];
+
+	if (mm_stat->is_used) {
```
maxzhen, line 677: Don't you need to check if mm_stat is NULL and assign 0 to usage and bo_count if it is NULL or not used? The original code does check. And if you check it here, there is no need to check inside this if statement.

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
```diff
@@ -1124,28 +1098,48 @@ static int xocl_init_memory_manager(struct xocl_drm *drm_p)
...
+	}
+
+	/* Memory manager initialize should be done only once */
+	if (drm_p->xocl_mm_done && (drm_p->xocl_mm != NULL)) {
+		/* Validate the new memory topology with existing memory manager */
+		if ((drm_p->xocl_mm->start_addr != mm_start_addr) ||
+				(drm_p->xocl_mm->end_addr != mm_end_addr) ||
+				(drm_p->xocl_mm->m_count != topo->m_count)) {
+			xocl_warn(drm_p->ddev->dev, "Memory Topology doesn't "
+				"consistent in xclbins. Re-initializing Memory Manager");
```
maxzhen, line 1134: This is confusing... At line 1127, the comment says that the memory manager should be done only once. But, here you're doing it again? Is this expected? If so, why issue an warning? If not, then is it really OK to re-init it? Claiming that an entity can only be initialized once is big decision to make and has big impact to future code maintenance. Please think through.

maxzhen (changes requested):

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drm.c
maxzhen: I missed the line 675, so mm_state will never be NULL. So, please just change the next two lines and check if mm_stat->is_used, instead of checking mm_stat is NULL.

saifuddin-xilinx: Fixed

saifuddin-xilinx:
retest this please

maxzhen (approved):

saifuddin-xilinx:
retest this please

PR #7165, opened 2022-11-11
Construct xrt::bo associated with hwctx

stsoe:
Add xrt::bo constructors that take hwctx.  

The goal is to remove the need for encoding 'slotidx' in the grpid flags passed during construction and instead make the context explicit during construction.

This PR does not remove the slotidx encoding, this has to be a coordinated change between XRT and various shim implementations so this is for follow up PRs.

But this PR does define new ishim APIs for context aware buffer allocation so shim implementation can be changed now to essentially ignore the slotidx encoded in the grpid.

Ultimately all XRT applications should construct a hardware context even in on legacy platforms where only one physical slot is supported.  The hwctx should be used explicitly in the kernel and bo constructors.  Shim level APIs operating on hwctx or using hwctx will be passed a hwctx_handle.  The hwctx_handle is currently typedef as a uint8_t for the urpose of encoding in flags, but once encoding is no longer needed the hwctx_handle can be changed to some opaque handle type.

No real flow changes for legacy applications, so risk is low

OpenCL regression and simple test modified with this PR

Ultimately yes,  we want to promote the use of hwctx for all flows


maxzhen (approved):

dayeh-xilinx:
retest this please.

PR #7167, opened 2022-11-15
Use hwctx handle at shim level if supported

stsoe:
This PR corrects a problem associated with hwctx and hwctx handles, where legacy shim APIs were (are) passed xrt::hw_context objects rather than the underlying handle.   

This PR works around an inconsistency in how XRT coreutil and XRT core (shim) exchange handles.

Some objects managed by XRT coretutil are wrappers of shim opaque handles that provides a hook into a kernel driver or memory managed by shim.  These opaque handles are created through shim APIs when the corresponding XRT coreutil object is created.

The interface between XRT coreutil and XRT core is through the shim opaque handle if one such is involved.  This means that if shim returns a handle, then shim gets this handle in all API that operate on the the data associated with the handle.  

The fix is only partial in that shim implementations that do not support creation of hw_context still are passed an internally created xrt::hw_context that wraps information (xclbin, qos, mode) needed by shim to do its work.  Once all shims support hw_context creation, some wrapper and try/catch code is this PR can be removed.

Low risk.  Existing shim APIs are still called with the changes in this PR.

OpenCL regressions.

None.

### DIFF
--- a/src/runtime_src/core/common/api/context_mgr.cpp
+++ b/src/runtime_src/core/common/api/context_mgr.cpp
@@ -99,7 +99,8 @@ public:
       if (m_cv.wait_for(ul, 100ms) == std::cv_status::timeout)
         throw std::runtime_error("aquiring cu context timed out");
     }
-    auto ipidx = m_device->open_cu_context(hwctx, ipname);
+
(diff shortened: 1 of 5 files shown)


maxzhen (approved):
Thanks for making this change.

chvamshi-xilinx (approved):

PR #7173, opened 2022-11-16
Refactoring of XDP CMake structure

jvillarre:
This breaks up the single CMake file used for building all of the profiling/debug plugins into individual make files for each part and removes unused definitions, making it easier to add new plugins to be built on different architectures.

Low risk as there is no algorithmic change, only the build process.

The code has been built for edge, x86 Linux, and Windows and the plugins have been verified to be built in the correct location.

No documentation impact.


hackwa (approved):
Splitting the cmake into smaller manageable chunks looks good.

IshitaGhosh (approved):
The change looks good.

PR #7177, opened 2022-11-17
Cleanup statically cached device entries from hw queue map

stsoe:
Ensure that cached device entries are removed when device is destructed.

Applications that repeatedly create and delete xrt::device objects can end up as bogus dead entries in a static map that manages hardware queue entries.  While the stale entries are not referenced, they can end up accumulating 8 bytes leaked memory that is not reclaimed until process exit where the map is destructed.

This PR fixes this problem by having the xrt_core::~device destructor remove the device entry from the static map.

No obvious risks. It is just a stale entry that is removed from a map.  No side effects.

OpenCL regressions

### DIFF
--- a/src/runtime_src/core/common/api/hw_queue.cpp
+++ b/src/runtime_src/core/common/api/hw_queue.cpp
@@ -25,8 +25,6 @@
 #include <mutex>
 #include <thread>
 
-#include <iostream>
-
 using namespace std::chrono_literals;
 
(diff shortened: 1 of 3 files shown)


chvamshi-xilinx (approved):

vboggara-xilinx (approved):

stsoe:
retest this please - infra

PR #7182, opened 2022-11-21
Fix wrong driver minor version for 2022.2 branch

rbramand-xilinx:

XRT driver version is 2.15.0 for 2022.2 branch, updated it to 2.14.0 which is proper one.

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/Makefile
+++ b/src/runtime_src/core/edge/drm/zocl/Makefile
@@ -30,7 +30,8 @@ ifneq ($(cflags_zocl),)
 endif
 
 # XRT driver version used in apu reporting
-# Version is specific to release
+# Version info is repeated here as Petalinux doesn't have access to 
+# user space top level CMake file. Version is specific to release
 XRT_DRIVER_VERSION ?= 2.14.0
 
 # Add patch number in version if 'XRT_VERSION_PATCH' env variable is defined
--- a/src/runtime_src/core/edge/drm/zocl/Makefile
+++ b/src/runtime_src/core/edge/drm/zocl/Makefile
@@ -29,8 +29,9 @@ ifneq ($(cflags_zocl),)
 	ccflags-y += $(cflags_zocl)
 endif
 
+# XRT driver version used in apu reporting
 # Version is specific to release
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/edge/drm/zocl/Makefile
```diff
@@ -30,7 +30,7 @@ ifneq ($(cflags_zocl),)
 endif
 
 # Version is specific to release
-XRT_DRIVER_VERSION ?= 2.15.0
+XRT_DRIVER_VERSION ?= 2.14.0
```
keryell, line 33: I guess this should be defined in a top-level file instead of being buried in a lot of files all over the place.
@uday610 @stsoe

chvamshi-xilinx:
retest this please

src/runtime_src/core/edge/drm/zocl/Makefile
rbramand-xilinx: @keryell this change is for driver, user space uses cmake flow where we have top level CMake file with this info. We build this driver using petalinux which will not have access to top level CMake.

keryell: Thanks for this information. Then why not explaining this in this file with a comment so that everyone can understand the purpose?

rbramand-xilinx: sure, I will add a comment. Thanks

keryell: Thanks. But I was thinking to a comment explaining the root cause you give in your answer. Like:
```suggestion
# The version is repeated here to be used by the device driver in the case of a PetaLinux build
# when the user-space top-level CMake configuration file specifying the version is not available.
XRT_DRIVER_VERSION ?= 2.14.0
```
To be truncated to 80 characters as usual.
Otherwise I will ask again the question about it the next time I read this. ;-)

rbramand-xilinx: Added the comment explaining that we cant use top level CMake. Made sure characters in line are < 80 : )

keryell (approved):
Thank you for the educational explanation.

dayeh-xilinx:
retest this please.

PR #7190, opened 2022-11-29
CR-1145946 - Fix seg fault due to wrong device handle in native C api's

rbramand-xilinx:

When XRT native C api's are called with xrtdevicehandle, we are using wrong api to get core device out of it which is causing seg fault. Fixed this issue.
Minor Fix: xbutil examine fails in edge platforms because of newly introduced device_query_default function.

It was reproduced in SSW_Embedded pipeline.

Used correct api for mapping

Low

Tested AIE application with C api's using hw_emu and sw_emu and both tests passed

NA


chvamshi-xilinx (approved):

manikandan-xilinx:
retest this please.

stsoe (approved):

vboggara-xilinx (approved):

PR #7197, opened 2022-11-30
AIE Profiling Plugin Refactor 

nishraptor:

Refactored the AIE Profiling Plugin to support both x86 and edge flows. 

The old AIE Profile plugin only supported edge flows. The Plugin was re-architected to be able to now support x86 flows as well.

The AIE Profiling plugin was split into edge and x86 implementations, with common functionality shared in the plugin file. The edge solution is now functional, with the x86 implementation to be completed in a future pull request. Additional changes include the default metricsets for unspecified tiles. If AIE profile is turned on, core, interface and memory metrics are defaulted to all:heat_map, all:conflicts, and all:input_bandwidths if they are not specified. For tile based metrics, any tiles that are not specified explicitly by the user get defaulted to the above metric sets.

The CMakeLists.txt was changed to build this new plugin instead of the old AIE Profile plugin. 

The edge flow was tested with various runtime metrics on multiple designs.


hackwa (approved):
Few code style issues but overall structure looks good.

src/runtime_src/xdp/profile/plugin/aie_profile_new/edge/aie_profile.h
```diff
@@ -0,0 +1,114 @@
...
+      AieProfile_EdgeImpl(VPDatabase* database, std::shared_ptr<AieProfileMetadata> metadata);
+
+      ~AieProfile_EdgeImpl() = default;
+
+      void updateDevice();
+      void poll(uint32_t index, void* handle);
+      bool checkAieDevice(uint64_t deviceId, void* handle);
+
+      bool setMetricsSettings(uint64_t deviceId, void* handle);
+      std::vector<tile_type> getAllTilesForCoreMemoryProfiling(const XAie_ModuleType mod,
```
hackwa, line 47: These functions don't make sense here as x86 will also need this metadata. As we discussed in meeting, xAie_ModuleType isn't needed here and this can be refactored to be at a common place. Please makes these changes when implementing x86 side of things.

jvillarre (approved):
Looks good as the first part of the rearchitecture of this plugin, with some small issues discussed internally to be addressed in future pull requests.

IshitaGhosh (approved):
Changes look good as first phase checkin for AIE Profiling Plugin Refactor

PR #7202, opened 2022-12-01
CR-1143061 and CR-1141964: Missing arguments in kernel to global memory data transfer table

jvillarre:
For a subset of platforms, the kernel to global memory data transfer table in the profile summary does not correctly list the arguments associated with a port.  This resolves the issue and correctly adds the arguments to the table.

When determining if an argument belongs to a particular port, we compare the memory resource destination of the port with the sptag from the metadata.  On some some platforms, the sptag uses different terminology that must be adjusted for.  This was discovered by a full suite of regressions on multiple platforms.

The comparison function was expanded to account for older platforms where DDR is specified as "bank" and for Versal platforms where DDR connections go through the NoC.

Low risk as this is localized to the comparison function for the arguments used in this one profile table.

Both failing test cases have been tested in hardware emulation.

No documentation impact.

### DIFF
--- a/src/runtime_src/xdp/profile/database/static_info/pl_constructs.cpp
+++ b/src/runtime_src/xdp/profile/database/static_info/pl_constructs.cpp
@@ -24,6 +24,18 @@
 // Anonymous namespace for local static helper functions
 namespace {
 
+  static std::string convertBankToDDR(const std::string& memory)
+  {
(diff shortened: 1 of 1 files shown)


hackwa (approved):
looks good

IshitaGhosh (approved):

PR #7213, opened 2022-12-07
Cleanup of AIE Profile Plugin

nishraptor:
Cleaned up the AIE Profiling plugin and moved functionality from the edge implementation to the shared metadata class.

Removed xrt_core::edge dependencies and  moved parsing functionality to the shared metadata class.

Tested on 2 separate designs.


jvillarre (approved):
Moving the common constructs to a common location looks good.

PR #7215, opened 2022-12-08
PS Kernel Cleanup

jeffli-xilinx:
Moved pscontext out of sk_types.h and into include/experimental
Switched to virtual destructor for pscontext class
Cleaned up PS kernel destructor sequence

VCK5000 and V70


stsoe (changes requested):
This does not look right in regards to pstcontext.h being exported. 
Is it used in end-user applications? 
The header itself is not in shape for export.

PR #7221, opened 2022-12-13
CR-1145195 Fixes related to default clock scaling override values

rajkumar-xilinx:

By default, power & temp scaling override feature disabled, and hence the override values should also be set to 0 initially. But, XRT reports these values as max values (threshold values), which is not correct.
Also, CMC report in "xbmgmt examine" not display the clock scaling feature enable message which causing a little confusion for the user. So, fix this message.
CR-1145195
For scaling override values, keep a copy of local override data and report the same to user whenever requested.
NA

xbmgmt examine's cmc report.
NA

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
@@ -174,6 +174,10 @@ struct xocl_xgq_vmr {
 	size_t			xgq_vmr_plm_log_size;
 	u16			pwr_scaling_threshold_limit;
 	u8			temp_scaling_threshold_limit;
+	u16			pwr_scaling_limit;
+	u8			temp_scaling_limit;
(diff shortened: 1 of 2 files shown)


rajkumar-xilinx:
Default readings, highlighted is the new message seen when clock scaling is enabled.
$xbmgmt examine -r cmc -d
---------------------------------------------------
[0000:d8:00.0] : xilinx_vck5000_gen4x8_qdma_base_2
---------------------------------------------------
CMC
  Heartbeat information unavailable
  Runtime Clock Scaling :
    **Enabled**
    Critical threshold (clock shutdown) limits:
      Power : 300 W
      Temperature : 100 C
    Throttling threshold limits:
      Power : 291 W
      Temperature : 95 C
    Power threshold override:
      Override : false
      Override limit : 0 W
    Temperature threshold override:
      Override : false
      Override limit : 0 C

chvamshi-xilinx (approved):

maxzhen (approved):

PR #7229, opened 2022-12-14
CR-1147558 clock throttling percentage in the dmesg

xdavidz:

This is to support checking of both firewall and clock shutdown, clock throttling.
For clock throttling, it is non critical thus return value is 0.
However, we want to print dmesg by checking the log has data (count > 0).

For firewall and clock shutdown, those are critical and return value is non-zero.
They also have dmesg if count > 0.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
@@ -1275,12 +1275,14 @@ static int xgq_firewall_op(struct platform_device *pdev, enum xgq_cmd_log_page_t
 	}
 
 	if (log_size != 0) {
-		char *log_msg = vmalloc(log_size);
+		char *log_msg = vmalloc(log_size + 1);
 		if (log_msg == NULL) {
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
```diff
@@ -1259,29 +1260,29 @@ static int xgq_firewall_op(struct platform_device *pdev, enum xgq_cmd_log_page_t
...
+
+	if (log_size > len) {
+		XGQ_WARN(xgq, "return log size %d is greater than request %d",
+			log->count, len);
+		/* reset to valid shared memory size */
+		log_size = len;
+	}
+
+	if (log_size != 0) {
+		char *log_msg = vmalloc(log_size);
```
maxzhen, line 1278: do you want to do vmalloc(log_size + 1) and log_msg[log_size] = '\0' before mem copy to make sure the string is always 0-terminated?

maxzhen (changes requested):

maxzhen (approved):

PR #7230, opened 2022-12-15
CR-1147934: Fix seg fault at end of execution when profiling is enabled on some devices

jvillarre:
On some devices and OS configurations, applications were crashing at the end of execution when PL profiling was turned on.  This was due to profiling trying to access information on the device after the hardware shim had been destroyed.  This pull request makes sure that any PL profiling information is offloaded to the profiling library before the shim is destroyed by adding a callback in the shim destructor.  This is necessary as when profiling dumps its information it requires the last data from the device.

This pull request also comments the shim functions and consolidates all of the individual profiling plugin callbacks into a single location in the plugin/xdp directory so they can scale without changing the shim.cpp file.

The bug was discovered through daily regression testing.

Since the order of static objects destruction is undefined, the added callback handles the case when the shim is destroyed before profiling objects are destroyed.  If profiling objects are destroyed first, then these functions simply return when the shim is destroyed.

Low risk as this structure of callbacks was in use for other profiling plugins.  When profiling is not enabled, the callbacks simply return so there is minimal overhead added.

The failing test case on hardware has been verified to be fixed with good profiling data being retrieved after the fix was applied.

No documentation impact.

### DIFF
--- a/src/runtime_src/core/pcie/linux/plugin/xdp/shim_callbacks.h
+++ b/src/runtime_src/core/pcie/linux/plugin/xdp/shim_callbacks.h
@@ -1,0 +1,75 @@
+/**
+ * Copyright (C) 2022 Advanced Micro Devices, Inc. - All rights reserved
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may
+ * not use this file except in compliance with the License. A copy of the
+ * License is located at
+ *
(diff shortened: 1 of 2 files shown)


stsoe (approved):

IshitaGhosh (approved):
The changes look good.

PR #7231, opened 2022-12-15
CR-1148012 fix hwmon device register & unregister sequence

rajkumar-xilinx:

Fix hwmon device registration sequence in hwmon_sdm driver
CR-1148012
Fixed hwmon driver registration sequence in hwmon_sdm driver. 
NA
XRT driver load & unload, reset.
NA

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
@@ -1203,28 +1203,28 @@ static int create_hwmon_sysfs(struct platform_device *pdev)
 	if (!sdm->supported)
 		return 0;
 
-	err = device_create_file(sdm->hwmon_dev, &name_attr.dev_attr);
-	if (err) {
-		xocl_err(&pdev->dev, "create attr name failed: 0x%x", err);
-		return err;
-	}
-
 	sdm->hwmon_dev = hwmon_device_register(&core->pdev->dev);
 	if (IS_ERR(sdm->hwmon_dev)) {
 		err = PTR_ERR(sdm->hwmon_dev);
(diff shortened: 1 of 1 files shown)


maxzhen (approved):

chvamshi-xilinx (approved):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
```diff
@@ -1409,14 +1409,13 @@ static void destroy_hwmon_sysfs(struct platform_device *pdev)
 	if (!sdm->supported)
 		return;
 
-	sysfs_remove_group(&pdev->dev.kobj, &hwmon_sdm_bdinfo_attrgroup);
-
 	if (sdm->hwmon_dev) {
-		hwmon_device_unregister(sdm->hwmon_dev);
 		device_remove_file(sdm->hwmon_dev, &name_attr.dev_attr);
+		hwmon_device_unregister(sdm->hwmon_dev);
```
xdavidz, line 1414: Raj, could you please explain why we need to change the sequence back from unregister first and remove file second? All linux code is doing this sequence: https://elixir.bootlin.com/linux/latest/source/drivers/hwmon/asc7621.c#L1112

I am wondering if we change it back, we might see the occasionally memory corruption again.

I actually tested all the sequence, so if we just do this sequence, we can see the painc oops right a way.
create the file
register
unregister
remove file << panic here

If we swap the unregister and remove file. The panic seems gone. But why we see the panic is my question.

xdavidz: @rajkumar-xilinx I think I kind of understand your are using this in a different way of creating devs, you have 2 devs.... please make sure backport this to 2022.2 after this merged.

xdavidz (approved):

PR #7245, opened 2022-12-20
Revert "CR-1134860 PCIE IO from host to DDR shared memory is too slow"

chvamshi-xilinx:
Reverts Xilinx/XRT#7227


chvamshi-xilinx:
retest this please

PR #7248, opened 2022-12-21
Increase skd cgroup memory limit from 1G to 2G

jeffli-xilinx:
Increase PS kernel service group memory limit from 1G to 2G to accommodate larger cached xclbin sizes.

### DIFF
--- a/src/runtime_src/tools/scripts/apu_recipes/skd.service
+++ b/src/runtime_src/tools/scripts/apu_recipes/skd.service
@@ -7,7 +7,7 @@ Type=forking
 User=softkernel
 ExecStart=/usr/bin/skd
 StandardOutput=journal+console
-MemoryLimit=1G
+MemoryLimit=2G
 
 [Install]
 WantedBy=multi-user.target


stsoe (approved):

jeffli-xilinx:
retest

manikandan-xilinx:
retest this please.

rozumx (commented):
Is this change still needed?

I thought that you were looking into reducing the xclbin container sizes by removing the PDI images.   Even if this wasn't the case, I would expect that the cache would not have duplicate xclbin containers, which looks to be the case.  

Instead of increasing the cache size, would it possible to create reference counts to cache images.  For this use-case, there will be only 1 cache image instead of multiple duplicate cache images, which seems to be the case.

jeffli-xilinx:
> Is this change still needed?
> 
> I thought that you were looking into reducing the xclbin container sizes by removing the PDI images. Even if this wasn't the case, I would expect that the cache would not have duplicate xclbin containers, which looks to be the case.
> 
> Instead of increasing the cache size, would it possible to create reference counts to cache images. For this use-case, there will be only 1 cache image instead of multiple duplicate cache images, which seems to be the case.

There will need to be a separate change to not cache copies of full xclbins.  But I think increasing memory limit is probably still prudent since if we want to support 32 and in the future 64 PS kernels, 1G is probably not enough.

PR #7251, opened 2022-12-22
make bandwidth host.cpp error message clear when buffer mismatch

xdavidz:
for vck5000

This is very unlikely an XRT issue. Instead this should be a shell issue that sometimes the data is not correct in the hardware DDR memory.

However, the bandwidth test cpp code prints the value without casting it to int, thus the cpp will take it as an ASCII and sometimes it is "invisible" when print out. So the fix is to cast the value to it's original int value, it should be from 0-256 based on the design of the bandwidth test.

### DIFF
--- a/tests/validate/bandwidth_test/src/host.cpp
+++ b/tests/validate/bandwidth_test/src/host.cpp
@@ -192,6 +192,9 @@ int main(int argc, char** argv) {
                 std::fill(output_host[i].begin(), output_host[i].end(), 0);
             }
 
+	    for (uint32_t j = 0; j < 256; j+=100)
+		    std::cout << "j:" << j << "=" << (uint32_t)input_host[j] << std::endl;
+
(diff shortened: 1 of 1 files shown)


maxzhen (commented):
I'm confused. If this is a shell issue, how can your code change in the test case fix it?

xdavidz:
> I'm confused. If this is a shell issue, how can your code change in the test case fix it?

The code is to help for narrowing down the shell issue. Otherwise we get error message like:
ERROR : kernel failed to copy entry 4055041 input output

This is useless:
It should be like:
ERROR : kernel failed to copy entry 4055041 input [0-255] output [0-255]. <<- this should be a value of 0-255. Thus, the hardware team can use this to identify what went wrong.

xdavidz:
> Member

Modified the PR's name in case of any further confusion.

PR #7252, opened 2022-12-22
CR-1134860 PCIE IO from host to DDR shared memory is too slow

xdavidz:
Hi @maxzhen  and @houlz0507 ,

This is the re-submitted fix for the map_bar fixes.

What I have missed is the map_bar is still being used for collecting golden version on legacy shells like u250, u50.
I simply rollback the map_bars functions (no changes).
And, only allow it is being called during golden version sysfs retrival.
I think this is safe because there is no other cases that can access this code path.

I don't change the map_bars function so that we can minimize the chance to introduce new regression.

Tested on u250 manually because pipeline doesn't cover this.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-core.c
@@ -188,6 +188,33 @@ failed:
 	return ret;
 }
 
+/*
+ * Legacy code will use pci_iomap(_range) instead of pci_iomap_wc(_range)
+ * thus on linux system will not allow future ioremap_wc to be able to enable
+ * better performance.
+ *
+ * Current code is using ioremap within sub-drivers and managing by either
+ * ioremap_nocache or ioremap_wc. Thus, we should not keep the bar mapped
+ * out of the scope of sub-drivers. Aka, mgmt-core should not map bars.
+ *
+ * However, to support legacy code which is already using mapped bar, we can
+ * map the bar, read the value and unmap the bar.
+ */
+uint32_t mgmt_bar_read32(struct xclmgmt_dev *lro, uint32_t bar_off)
+{
+	int rc = 0;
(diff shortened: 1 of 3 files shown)


maxzhen (commented):
I'm not sure if this will work well. If, after a portion of the BAR is mapped as WC, some code path called MGMT_READ_REG32. The BAR will be mapped as UC? Or the MGMT_READ_REG32 call will fail since there is some overlapping of the mapping? If the BAR could be remapped to UC while a portion of it is WC, then the user of the WC portion will be interfered? Or maybe I'm missing something here...

xdavidz:
> I'm not sure if this will work well. If, after a portion of the BAR is mapped as WC, some code path called MGMT_READ_REG32. The BAR will be mapped as UC? Or the MGMT_READ_REG32 call will fail since there is some overlapping of the mapping? If the BAR could be remapped to UC while a portion of it is WC, then the user of the WC portion will be interfered? Or maybe I'm missing something here...

@maxzhen Let's discuss your concern tomorrow at work. I think the difference with the new proposal is that we map and **unmap** for one MGMT_READ_REG32. Since the bar is unmapped, then the later map with WC will work well. 

The pro is: it is clear that when we don't mix UC and WC. We map and unmap entire bar for MGMT_READ_REG32; then we only map certain range(shared memory for versal case) as WC. 

The con is: It is not efficient but it doesn't change legacy behavior of the MGMT_READ_REG32.  The concern is that in the future someone keeps using this API, that is a bad thing that we want to avoid.

I would like to discuss with you all options and then we can proceed. My point is mixing UC and WC might not be an ideal solution. The behavior of mixing mapping UC and WC might be undefined or very error-prone across different linux kernels. I just don't know those details. But clearly, we map UC when we want to use UC and we map WC when we want to use WC, no overlap gives me a little peace in mind.

xdavidz:
close this due to this depends on UI team to change flash code, they don't have bandwidth for this release. We will resubmit separate PR when decencies are cleared.

PR #7257, opened 2023-01-04
Updated init-apu.service to ensure skd.service starts after init-apu

jeffli-xilinx:
Addresses CR-1141939

Signed-off-by: Jeff Lin <jeffylin@xilinx.com>

Sometimes skd service could not start if user soft kernel has not been created yes due to race condition between init-apu service and skd service

init-apu.service was added ensure all initialization is done before apu-boot which signals to rpu that apu has completed boot process.  However, there was no clear signal to indicate skd service need to run after init-apu service has completed.
Thus occasionally skd would not start which lead to xclbin download failure with ps kernels.

Added Before keyword to ensure init-apu is run before skd.service, and also make it an exec so it is only considered done after the script has finished.

None

Tested on vck5000 and v70

None

### DIFF
--- a/src/runtime_src/tools/scripts/apu_recipes/init-apu.service
+++ b/src/runtime_src/tools/scripts/apu_recipes/init-apu.service
@@ -1,7 +1,9 @@
 [Unit]
 Description=APU init Daemon
+Before=skd.service
 
 [Service]
+Type=exec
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx (approved):

chvamshi-xilinx (approved):
Thanks for making the change.

PR #7260, opened 2023-01-09
Added support for Profiling PS kernel

nishraptor:
Added support for a profiling PS kernel

This PR adds AIE profiling support for the VCK5000 via PS kernel. 

The profiling PS kernel is scheduled by the host when aie_profile=true. First, the counters are configured on the AIE. Then, during the application run, the profiling PS kernel is repeatedly scheduled to poll the counters and return the data to the host.

Tested on lab8 example on the VCK5000.


src/runtime_src/core/edge/ps_kernels/profiling/aie_profile_config/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (C) 2018-2022 Xilinx, Inc. All rights reserved.
+# Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
+set(SRC_DIR  "${CMAKE_CURRENT_SOURCE_DIR}")
+set(PS_KERNEL_INSTALL_DIR "${XRT_INSTALL_LIB_DIR}/ps_kernels_lib")
+
+include_directories(
+  ${SRC_DIR}
```
stsoe, line 8: Do not current directory to the search path, the compiler automatically searches current directory for includes first.

src/runtime_src/core/edge/ps_kernels/profiling/aie_profile_config/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+# Copyright (C) 2018-2022 Xilinx, Inc. All rights reserved.
+# Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
+set(SRC_DIR  "${CMAKE_CURRENT_SOURCE_DIR}")
+set(PS_KERNEL_INSTALL_DIR "${XRT_INSTALL_LIB_DIR}/ps_kernels_lib")
+
+include_directories(
+  ${SRC_DIR} 
+  ${SRC_DIR}/../../../include
+  ${SRC_DIR}/../../../../
+  ${SRC_DIR}/../../../edge/include
```
stsoe, line 11: Please use XRT_SOURCE_DIR as the root directory rather than relative paths.

src/runtime_src/core/edge/ps_kernels/profiling/aie_profile_config/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+  )
+
+file(GLOB SRC_FILES
+  "${SRC_DIR}/*.cpp"
+  )
```
stsoe, line 16: In CMake it is better to explicit in the targets, e.g. add_library of the source files.

src/runtime_src/core/edge/ps_kernels/profiling/aie_profile_config/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+
+set(CMAKE_CXX_FLAGS "-DXAIE_DEBUG -D__PS_ENABLE_AIE__ ${CMAKE_CXX_FLAGS}")
```
stsoe, line 18: Don't change CMAKE flags, add the defines to the target using `target_compile_defintions`

src/runtime_src/core/edge/ps_kernels/profiling/aie_profile_config/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+add_definitions(-DXCLHAL_MAJOR_VER=1 -DXCLHAL_MINOR_VER=0)
```
stsoe, line 19: I don't understand the mix between `set(CMAKE_CXX_FLAGS, ...)` and this `add_definitons`.   Please use just `target_compile_defintions`.

src/runtime_src/core/edge/ps_kernels/profiling/aie_profile_config/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+add_library(aie_profile_config SHARED ${SRC_FILES})
+
+set_target_properties(aie_profile_config PROPERTIES VERSION ${XRT_VERSION_STRING}
```
stsoe, line 22: Good, you are using a target function, exactly what we should use always.

stsoe (changes requested):

src/runtime_src/core/edge/ps_kernels/profiling/aie_profile_config/CMakeLists.txt
nishraptor: Made the changes you requested. Let me know if this looks OK.

stsoe (approved):
Thank you.

jvillarre (approved):
Additions look good to add the new feature and changes of older code do some clean up refactoring as previously scheduled.

PR #7267, opened 2023-01-10
Fix for incorrect xrt rpm name

saumyag-xilinx:
[CR-1149492](https://jira.xilinx.com/browse/CR-1149492)
Fixed the change in XRT rpm name format.
low
Tested xrt package creation on redhat, centos and ubuntu
na

### DIFF
--- a/src/CMake/cpackLin.cmake
+++ b/src/CMake/cpackLin.cmake
@@ -31,6 +31,13 @@ SET(CPACK_RPM_SPEC_MORE_DEFINE "%define __python python3")
 
 if (DEFINED CROSS_COMPILE)
   set(CPACK_REL_VER ${LINUX_VERSION})
+elseif (${LINUX_FLAVOR} MATCHES "^centos")
+  execute_process(
+    COMMAND awk "{print $4}" /etc/redhat-release
+    COMMAND tr -d "\""
+    OUTPUT_VARIABLE CPACK_REL_VER
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+  )
 else()
   execute_process(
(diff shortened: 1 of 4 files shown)


chvamshi-xilinx (approved):

stsoe (commented):
I am wary of redhat-relaese.  I think it is a mistake that we append .x to the package name.   major.x is compatible with major.y where y>=x.   So there is no point including x when we know what the least minor we support is.

chvamshi-xilinx:
> I am wary of redhat-relaese. I think it is a mistake that we append .x to the package name. major.x is compatible with major.y where y>=x. So there is no point including x when we know what the least minor we support is.

That is true Soren, but this new naming convention broke Sprite regressions, Thatswhy we wanted to add it back. Saumya can send another PR after intimating DevOps/DSV about this name change. We also wanted to add OS_DIST as part of the package name. Saumya will do all those changes together

chvamshi-xilinx:
> > I am wary of redhat-relaese. I think it is a mistake that we append .x to the package name. major.x is compatible with major.y where y>=x. So there is no point including x when we know what the least minor we support is.
> 
> That is true Soren, but this new naming convention broke Sprite regressions, Thatswhy we wanted to add it back. Saumya can send another PR after intimating DevOps/DSV about this name change. We also wanted to add OS_DIST as part of the package name. Saumya will do all those changes together

@stsoe , shall i go ahead and merge once build passes as it is blocking sprite regressions

stsoe (approved):

chvamshi-xilinx:
> > @stsoe , shall i go ahead and merge once build passes as it is blocking sprite regressions
> 
> Yes please.

Thanks Soren.

PR #7268, opened 2023-01-10
Updated AIE Trace plugin to support Tile-based Metrics on PCIe

nishraptor:
Updated the AIE Trace plugin to support tile-based metrics

Moved from the old metric parsing to the new tile-based metrics for PCIe boards, as well as new delay options. 

Changed the input data we pass to the AIE Trace PS kernel and modified the PS kernel such that it now uses the new configuration parameters to set tile-based metrics and delay.


pgschuey (approved):
changes look good

PR #7270, opened 2023-01-11
Displaying only valid sensors in the xrt tool report command

saumyag-xilinx:
[CR-1146994](https://jira.xilinx.com/browse/CR-1146994)- Add command option support for sensor status reporting on Versal platforms
We are now displaying only valid sensors through the XRT tool report command, so a check has been added to verify the sensor status and only then add the sensor.
Sensor status byte value represents the following
     0 – Sensor Data not present
     1 - Sensor Data present & valid
     2 – sensor Data not available.
low
Example: On vck5000, using xbutil examine -r thermal -d
Thermals
  Temperature            : Celcius
  PCB                    :     27 C
  device                 :     37 C
  vccint                 :     38 C
  cage_temp_0            :      0 C
  cage_temp_1            :      0 C

Now, the thermal report shows only valid sensor information:
Thermals
  Temperature            : Celcius
  PCB                    :     27 C
  device                 :     37 C
  vccint                 :     38 C

na

### DIFF
--- a/src/runtime_src/core/pcie/linux/device_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/device_linux.cpp
@@ -142,6 +142,7 @@ struct sdm_sensor_info
     result_type output;
     std::vector<std::string> stats;
     std::string errmsg;
+    const std::string SD_PRESENT_CHECK = "1";
 
     // The voltage_sensors_raw is printing in formatted string of each line
     // Format: "%s,%u,%u,%u,%u"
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -142,6 +142,7 @@ struct sdm_sensor_info
     result_type output;
     std::vector<std::string> stats;
     std::string errmsg;
+    const std::string SD_PRESENT_CHECK = "1";
```
stsoe, line 145:
```suggestion
    constexpr const char* sd_present_check = "1"; 
```
Plus 1 for symbolic constant.
No upper case for variables. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rl-all-caps
Plus, might as well use compile time constant here since string buys nothing for the comparison done later.

stsoe (changes requested):

src/runtime_src/core/pcie/linux/device_linux.cpp
saumyag-xilinx: Thank you for suggesting this @stsoe. Done.

stsoe:
retest this please

dayeh-xilinx:
retest this please.

PR #7277, opened 2023-01-12
Modify xrt_hwctx_handle to support all the variants of a hwctx

stsoe:
This PR changes xrt_hwctx_handle into a class that supports holding either an class derived from xrt_hwctx_handle, a void*, or a slotidx.

This change is necessary for shims that migrate towards representing a hwctx as an opaque handle. 

By converting xrt_hwctx_handle into a base class, a shim derived class can redefine methods for accessing private details, e.g. slotidx, of a context handle.

The slotidx is currently used by xrt::kernel::group_id to encode in flags used when constructing an xrt::bo.  A new hwctx aware xrt::bo constructor has been added, but legacy applications do not use this constructor and continue to rely on the encoding of slotidx.

Until the encoding can be removed, the class xrt_hwctx_handle has a virtual override that can be redefined by shim to return the slotidx, this allows the existing encoding to continue to be used.

Standard regression tests.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -143,6 +143,12 @@ public:
     return m_device;
   }
 
+  bool
+  valid_hwctx() const
+  {
+    return (m_hwctx) ? true : false;
+  }
+
   xrt_hwctx_handle
   get_hwctx_handle() const
   {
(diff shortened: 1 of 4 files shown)


dayeh-xilinx:
retest this please.

maxzhen (approved):

stsoe:
Replaced by #7280

PR #7278, opened 2023-01-13
Fixed CR-1149084, for legacy cu create we also need to reserve the Cu…

saifuddin-xilinx:
…s index

Signed-off-by: Saifuddin <saifuddi@amd.com>

CR-1149084

hw context changes introduce this issue. 

Actually as per the new implementation we need to reserve the CUs index before crate them. 
This is required for both XQG and legacy case. For HW context changes we have done for XQG case but for legacy case  we missed it. 

N/A
Regular validate and this CR
N/A


chvamshi-xilinx (approved):

maxzhen (approved):

dayeh-xilinx:
retest this please.

PR #7282, opened 2023-01-16
CR 1145449 freq scaling tolerance

prapulkrishnamurthy:
Signed-off-by: prapulkrishnamurthy <prapulk@xilinx.com>

Implementation to check the tolerance 
Min Max is not checked as Req Freq is chosen Max directly from topology.

set_and_verify_freq API is implemented in reference to icap older code.

Edge case is not handled which needs to be addressed.

Testing in progress


xdavidz (changes requested):
please rebase your workspace and regenerate the source code.

the frequency table should not be in the master code.
We need a cleaner PR to review.

xdavidz:
messed up worksapce, closing

PR #7283, opened 2023-01-16
CR-1149239 xrt failed to write more registers when ert_start_kernel_cmd count is larger

mamin506:
Signed-off-by: Min Ma <min.ma@amd.com>

CR-1149239

This bug is here at the beginning. Recently, there is a DPU related use case will need to create a command more than 1KB and less than 2KB. So, this is exposed.

Extent the command slow size limited to 4KB.

Low

V70

no


manikandan-xilinx:
retest this please.

dayeh-xilinx:
retest this please.

chvamshi-xilinx (approved):

PR #7299, opened 2023-01-24
Fixing profiling issue with freeze

jvillarre:
Reimplements the refactoring on calling the profiling hooks in the linux shim and removes the hooks in the create_hw_context, which prevents certain test cases from hitting the profiling hooks twice recursively and causing a freeze.

Bug was introduced in Pull Request 7209, and was discovered through internal testing.

The problem was solved by removing the extra profiling hooks and relying on the hooks in xclLoadXclbin.  Other solutions will require a greater understanding of expected usage of create_hw_context on multiple platforms to see if and how profiling hooks can be added there for future designs.

Low risk to current designs, some risk that future designs that do not call loadXclbin will not have device level profiling enabled.

The original profiling and trace test has been verified with the fix.

No documentation impact

### DIFF
--- a/src/runtime_src/core/pcie/linux/shim.cpp
+++ b/src/runtime_src/core/pcie/linux/shim.cpp
@@ -1555,9 +1555,10 @@ int
 shim::
 xclLoadXclBin(const xclBin *buffer)
 {
-  xdp::hal::flush_device(this);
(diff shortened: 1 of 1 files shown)


dayeh-xilinx:
retest this please.

stsoe (approved):

PR #7312, opened 2023-01-27
CR-1144989: Fixing issue where profiling generated incorrect CSV format

jvillarre:
When native XRT API tracing is turned on, the profiling library captures "sync" calls as both an API call and a data transfer, and generates the events in the native_trace.csv file so post-processing can display both the API call and the data transfer.  The two separate end events in the csv file were using the same start event id and end event id, which is not the correct format supported by the CSV file and the post-processing tools.

The bug where multiple end events used the same start and end event ID was discovered with internal testing.

The issue is resolved by creating and storing two separate events for calls to "sync."  This allows the infrastructure to correctly track start and end events for both portions that are eventually displayed on the final visualization.

Low risk, as this is focused on tracing of Native API sync calls.

The original design has been verified.

No documentation impact.


dayeh-xilinx:
retest this please.

IshitaGhosh (approved):
Changes look good

PR #7323, opened 2023-02-01
CR-1151107: Correcting warning message for too large PL trace buffers on Edge

jvillarre:
On Edge systems with limited available memory, it was possible to ask for a PL trace buffer in memory via xrt.ini options that was too big for the available memory yet too small to trigger our warning messages, resulting in a bad alloc.  This pull request adjusts the code so both AIE and PL buffer sizes are checked against the size of the memory available on the device when running on edge instead of using predefined hard coded values.

The issue was discovered through internal testing.

The problem was solved by refactoring the current code used to check if enough space was available for AIE buffers and applying it to the PL buffer portion of the code as well.

Low risk, as it fixes a bad alloc error that only applies when a buffer is requested that is too large.  In the worst case, if we cannot determine the available memory, we will not change the current behavior.

The original failing test has been tested on a vck190 board.

No documentation impact.

### DIFF
--- a/src/runtime_src/xdp/profile/device/device_intf.cpp
+++ b/src/runtime_src/xdp/profile/device/device_intf.cpp
@@ -17,9 +17,25 @@
 
 #define XDP_SOURCE
 
-#include "aieTraceS2MM.h"
-#include "device_intf.h"
-#include "tracedefs.h"
+#include <chrono>
+#include <cmath>
+#include <cstdio>
+#include <cstring>
+#include <iostream>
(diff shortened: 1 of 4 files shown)


nishraptor (approved):

pgschuey (approved):

hackwa (approved):
looks good for checking PL and AIE independently. In case both are used, this might not work. We may have to look at other entries in meminfo in future that tell us amount of free device memory.

PR #7335, opened 2023-02-02
Added support for V70 AIE Profile PS kernel

nishraptor:
Added support for V70 AIE Profile via PS kernel.

Added support for Memtiles in addition to regular tiles for profiling on the V70. A new PS kernel was created for the V70 to configure the AIE2 registers

Breaking current VCK5000 profiling, but this has been tested.

This change has been tested on the VCK5000 to make sure everything working correctly. This requires further testing on the V70.


jvillarre (approved):
Some comments and parts need to be updated (left over from some cut-paste error), but functionally looks correct and necessary for initial testing.

PR #7341, opened 2023-02-03
Trace processor to create events from raw PL trace data and write to output csv

IshitaGhosh:
Add methods in Static DB source to retrieve info on IP, Memory topology, Debug IPs, Clock from xclbin file. This is in addition to methods that retrieve info using device handle.
Add source and Makefile for Trace Processor executable which reads and creates events from raw PL trace data and writes to output.csv

The Makefile for trace processor executable is outside of build infrastructure for XRT and its tools. The new executable is Early Access. So, user needs to manually build it in the source directory using XRT Libraries and Include files.


src/runtime_src/xdp/profile/database/static_info_database.cpp
```diff
@@ -2111,4 +2074,279 @@ namespace xdp {
     commandQueueAddresses.emplace(a) ;
   }
 
+  // This function is called from "trace_processor" tool 
+  // The tool creates events from raw PL trace data
+  void VPStaticDatabase::updateDevice(uint64_t deviceId, std::string xclbinFile)
```
stsoe, line 2079:
```suggestion
  void VPStaticDatabase::updateDevice(uint64_t deviceId, const std::string& xclbinFile)
```

src/runtime_src/xdp/profile/database/static_info_database.cpp
```diff
@@ -2111,4 +2074,279 @@ namespace xdp {
...
+      XclbinInfo* xclbin = deviceInfo[deviceId]->currentXclbin() ;
+      if (xclbin)
+        db->getDynamicInfo().markXclbinEnd(deviceId) ;
+    }
+
+    DeviceInfo* devInfo = nullptr ;
+    auto itr = deviceInfo.find(deviceId);
+    if (itr == deviceInfo.end()) {
+      // This is the first time this device was loaded with an xclbin
+      devInfo = new DeviceInfo();
```
stsoe, line 2102: Don't use raw new and delete

src/runtime_src/xdp/profile/database/static_info_database.cpp
```diff
@@ -2111,4 +2074,279 @@ namespace xdp {
...
+        devInfo->isEdgeDevice = true ;
+      deviceInfo[deviceId] = devInfo ;
+
+    } else {
+      // This is a previously used device being reloaded with a new xclbin
+      devInfo = itr->second ;
+      devInfo->cleanCurrentXclbinInfo() ;
+    }
+    
+    XclbinInfo* currentXclbin = new XclbinInfo() ;
```
stsoe, line 2114: Ditto

jvillarre (approved):

PR #7356, opened 2023-02-09
CR-1152824: Fixing trace buffer allocation to not use full memory resources when user over allocates

jvillarre:
On some configurations, the requested buffer size for device trace information is not able to fit in the memory resources connected to the trace offload infrastructure.  This pull request fixes the warning message emitted and also restricts the total size of the buffer trace allocates to 80% of the total memory resource.  This does not prevent users from overallocating additional buffers on the memory resource, but does leave some room for host buffers if the user has overallocated trace.

This issue was discovered by full regression testing on multiple platforms.

Low risk, as this adjusts a warning message and only changes the amount of trace buffer allocated on designs that were already overallocating the resource.

No documentation impact.


IshitaGhosh (approved):
Changes look good

dayeh-xilinx:
retest this please.

hackwa (approved):
looks good

PR #7360, opened 2023-02-10
xma: fix for segfault issue due to short-lived sessions

vboggara-xilinx:
Signed-off-by: vboggara <vboggara@xilinx.com>

NOTE: Same fix submitted in master branch (https://github.com/Xilinx/XRT/pull/7055), backporting these changes to 2021.1 due to some customer issue as they are facing similar kind of issue(segfault).

There are 2 threads `xma_thread1` and `xma_thread2` reading from a vector of sessions inside the `g_xma_singleton`. The access was not guarded by any mutex while other threads could access the sessions vector and push a new element, this sometimes was causing a race condition where the thread was looping over elements and a new elements was being added causing a vector resize. This automatically invalidated any existing vector iterator (which the threads were using for reading from vector) which resulted in a segfault when trying to dereference it.  

The bug has been discovered when running a lot of quite short-lived sessions which caused the application to exit due to
segfaults.

Locking a mutex before accessing a list of sessions.
Probably none, the mutex could be replaced with `std::shared_mutex` as there are probably more reads than writes but with that low frequency of locking it, it is probably not worth it.
Tested locally some XMA cases with this fix.

### DIFF
--- a/src/xma/include/lib/xmaapi.h
+++ b/src/xma/include/lib/xmaapi.h
@@ -43,7 +43,7 @@ typedef struct XmaSingleton
     XmaHwCfg          hwcfg;
     bool              xma_initialized;
     bool              kds_old;
-    uint32_t          cpu_mode;
+    std::atomic<uint32_t> cpu_mode;
     std::mutex            m_mutex;
     std::atomic<uint32_t> num_decoders;
     std::atomic<uint32_t> num_encoders;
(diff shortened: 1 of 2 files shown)


stsoe (approved):

vboggara-xilinx:
@sarab96 , pl review the changes.

sarab96 (approved):
I'm not familiar with these changes.
Soren's & Vamshi's approval are sufficient for this PR

vboggara-xilinx:
@stsoe , please merge this PR.

PR #7361, opened 2023-02-13
Fix race in bo and device construction

stsoe:
Multiple threads competing to construct the very first xrt::bo object can hit a race in a device query for nodma check.  Additionally, multiple threads creating device objects can hit a race in device map access.

The nodma check is cached in a member variable and this variable is subsequently checked.  The check must be under a mutex lock.

Device construction is inserting core device objects into a static map in core util.  This map needs to be protected not only for insertion but also for reading.

### DIFF
--- a/src/runtime_src/core/common/device.cpp
+++ b/src/runtime_src/core/common/device.cpp
@@ -54,6 +54,7 @@ bool
 device::
 is_nodma() const
 {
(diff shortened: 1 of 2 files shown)


vboggara-xilinx (approved):
changes looks good.

PR #7367, opened 2023-02-14
CR-1142742 No interface counters mentioned in summary.csv when packet counters are used

pgschuey:
Better tracking of AIE resources

AIE counter resources API was using bool instead of uint8_t

Lots of testing on vck190


dayeh-xilinx:
retest this pleases.

jvillarre (approved):

PR #7368, opened 2023-02-15
src/runtime_src/tools/scripts: Add setup.zsh for use with the zsh shell

### LINKED ISSUE #7366: Sourcing setup.sh fails when using zsh
Sourcing setup.sh [0] fails (`dirname: missing operand`) when using zsh as shell; also sourcing setup.csh doesn't work (`/opt/xilinx/xrt/setup.csh:17: bad pattern: path=(/usr/sbin ...`)

Please consider providing a setup.zsh script for zsh users. It can be almost the same as the setup.sh script, just replacing
```
XILINX_XRT=$(readlink -f $(dirname ${BASH_SOURCE[0]}))
```
by
```
XILINX_XRT=$(readlink -f $(dirname ${(%):-%N}))
```
Thanks!

[0] https://github.com/Xilinx/XRT/blob/master/src/runtime_src/tools/scripts/setup.sh

mgehre-amd:
Provides a setup.zsh that can be used when using the zsh shell.

Fixes #7366

Alternatively, we could modify setup.sh to detect which shell is being used,
to provide a single script.

### DIFF
--- a/src/runtime_src/tools/scripts/setup.sh
+++ b/src/runtime_src/tools/scripts/setup.sh
@@ -30,7 +30,15 @@ if [[ $OSDIST == "centos" ]] || [[ $OSDIST == "rhel"* ]]; then
     fi
 fi
 
-XILINX_XRT=$(readlink -f $(dirname ${BASH_SOURCE[0]:-${(%):-%x}}))
+
+if [ -n "$BASH_VERSION" ]; then
+    XILINX_XRT=$(readlink -f $(dirname ${BASH_SOURCE[0]:-${(%):-%x}}))
+elif [ -n "$ZSH_VERSION" ]; then
+    XILINX_XRT=$(readlink -f $(dirname ${(%):-%N}))
+else
+    echo "ERROR: Unsupported shell. Only bash and zsh are supported"
+    return 1
+fi
 
 if [[ $XILINX_XRT != *"/opt/xilinx/xrt" ]]; then
     echo "Invalid location: $XILINX_XRT"


src/runtime_src/tools/scripts/setup.zsh
```diff
@@ -0,0 +1,72 @@
...
+# This script is installed in /opt/xilinx/xrt and must
+# be sourced from that location
+
+# Check OS version requirement
+OSDIST=`cat /etc/os-release | grep -i "^ID=" | awk -F= '{print $2}'`
+if [[ $OSDIST == "centos" ]]; then
+    OSREL=`cat /etc/redhat-release | awk '{print $4}' | tr -d '"' | awk -F. '{print $1*100+$2}'`
+else
+    OSREL=`cat /etc/os-release | grep -i "^VERSION_ID=" | awk -F= '{print $2}' | tr -d '"' | awk -F. '{print $1*100+$2}'`
+fi
```
keryell, line 17: I thought the goal of `/etc/os-release` was to just source it, not parse it...

src/runtime_src/tools/scripts/setup.zsh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (C) 2019-2021 Xilinx, Inc. All rights reserved.
+# Copyright (C) 2023 Advanced Micro Devices, Inc. All rights reserved.
+#
+# Script to setup environment for XRT
+# This script is installed in /opt/xilinx/xrt and must
+# be sourced from that location
```
keryell, line 9: Which means the line 1 should not exist for a second reason.

src/runtime_src/tools/scripts/setup.zsh
```diff
@@ -0,0 +1,72 @@
...
+# 1. This is a workaround for set -e
+# The issue is chaining conditionals with actual commands
+# The issue is caused when sourcing the ${COMP_FILE}.
+# Specifically ${COMP_FILE}::_sysvdirs. Each check in that function
+# will fail the script due to the issues.
+# If set -e is removed from the pipeline then check can be removed
+# 2. Make sure that the shell is bash! The completion may not function
+# correctly or setup on other shells.
+# 3. Make sure the bash completion file exists
+if [[ $- != *e* ]] && [[ "$BASH" == *"/bash" ]] && [ -f "${COMP_FILE}" ]; then
```
keryell, line 51: I am a little bit lost here. I thought this was about `zsh`. Is there anything specific to `zsh` here anyway?

keryell (changes requested):
As you said:
> Alternatively, we could modify setup.sh to detect which shell is being used,
to provide a single script.

That sounds like a great idea to avoid adding to the existing technical debt. :-)

mgehre-amd:
@keryell, thanks for your review. The only difference between setup.zsh and setup.sh is
```
--- src/runtime_src/tools/scripts/setup.zsh     2023-02-15 09:45:01.145363000 +0100
+++ src/runtime_src/tools/scripts/setup.sh      2023-02-15 09:43:31.447311000 +0100
@@ -30,7 +30,7 @@
     fi
 fi

-XILINX_XRT=$(readlink -f $(dirname ${(%):-%N}))
+XILINX_XRT=$(readlink -f $(dirname ${BASH_SOURCE[0]:-${(%):-%x}}))

 if [[ $XILINX_XRT != *"/opt/xilinx/xrt" ]]; then
     echo "Invalid location: $XILINX_XRT"
```
so your other comments apply equally to the setup.sh script.

As a next step, I will merge my changes into the setup.sh script, so we have only one that works for both bash and zsh.

keryell (approved):
Thanks.
Feel free to fix the rest of script if you are motivated. :-)

src/runtime_src/tools/scripts/setup.sh
```diff
@@ -30,7 +30,15 @@ if [[ $OSDIST == "centos" ]] || [[ $OSDIST == "rhel"* ]]; then
...
 fi
 
-XILINX_XRT=$(readlink -f $(dirname ${BASH_SOURCE[0]:-${(%):-%x}}))
+
+if [ -n "$BASH_VERSION" ]; then
+    XILINX_XRT=$(readlink -f $(dirname ${BASH_SOURCE[0]:-${(%):-%x}}))
+elif [ -n "$ZSH_VERSION" ]; then
+    XILINX_XRT=$(readlink -f $(dirname ${(%):-%N}))
+else
+    echo "ERROR: Unsupported shell. Only bash and zsh are supported"
```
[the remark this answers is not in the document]
mgehre-amd: Sorry, I'm not sure if you would like me to change my commit?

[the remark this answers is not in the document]
keryell: Nothing. I was just some wondering.

chvamshi-xilinx:
ok to test

dayeh-xilinx:
retest this please.

mgehre-amd:
Ping @rozumx & @stsoe for review.

stsoe (approved):

PR #7381, opened 2023-02-21
VITIS-6990:xrt::xclbin API to obtain interface_uuid

vboggara-xilinx:
Added below API for xrt::xclbin to obtain interface_uuid 
```
//get_interface_uuid() - Get the interface uuid of the xclbin and this can be a collection of UUIDs and not just one.   
//return  - Collection of UUIDs

XCL_DRIVER_DLLESPEC
std::vector<std::string>  get_interface_uuid() const;

//This API returning std::vector<std::string> instead of std::vector<uuid> because 
// interface_uuild which is coming from PARTITION_METADATA  is not in standard UUID format.
```
https://jira.xilinx.com/browse/VITIS-6990
Low risk as it is new API.
Tested locally some cases.
Yes, Need to add this new API to the XRT::XCLBIN documentation.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_xclbin.cpp
+++ b/src/runtime_src/core/common/api/xrt_xclbin.cpp
@@ -21,10 +21,11 @@
 #define XRT_CORE_COMMON_SOURCE // in same dll as core_common
 #include "core/include/experimental/xrt_xclbin.h"
 
-#include "core/common/system.h"
 #include "core/common/device.h"
 #include "core/common/message.h"
 #include "core/common/query_requests.h"
+#include "core/common/system.h"
+#include "core/common/utils.h"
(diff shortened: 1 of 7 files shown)


stsoe (changes requested):
Couple of issues.  I think we should change xclbinutil to dump the interface_uuid in axlf_header `rom_id` and simply return that.  All the dt parsing code does not belong in coreutil, but is already part of xclbinutil, so it should be able to parse the dt data.  Secondly, the interface uuid should be returned as a uuid not as a string.    Finally,, we only have one interface uuid after the passing of 2RP, so no vector is necessary.  Please correct me if I am wrong here.

vboggara-xilinx:
Closing this pull request and will create new PR with Soren's suggestions.

PR #7385, opened 2023-02-21
CR-1153424 IOMMU Status dump

prapulkrishnamurthy:
Iommu status check from iommu grub logs

Parsed iommu event logs from Dmesg and dumped into service bundle output directory.

Tested on VCK5000 and found the output at dump files in output directory.
image

### DIFF
--- a/src/runtime_src/tools/scripts/service_bundle.sh
+++ b/src/runtime_src/tools/scripts/service_bundle.sh
@@ -59,27 +59,6 @@ done
 #############################
 
 # $1 output directory
-iommu_dump()
-{
-	dmesg | grep iommu > $1/iommu_dump.txt
-}
-
-# $1 output directory
-iommu_groups()
-{
-	rm -r $1/iommu_groups.txt 2> /dev/null
-	export devs=$(find /sys/kernel/iommu_groups/* -maxdepth 0 -type d | sort -V)
-	for g in $devs; do
-		echo "IOMMU Group ${g##*/}:" >> $1/iommu_groups.txt
-		for d in $g/devices/*; do
-			VALUE="$(/sbin/lspci -nns ${d##*/})"
-			echo -e "\t$VALUE"
-			printf "	%s\n" "$VALUE" >> $1/iommu_groups.txt
-		done;
-	done;
-}
-
-# $1 output directory
 core_dumps()
 {
 	if [ -z $ALL_LOG ];then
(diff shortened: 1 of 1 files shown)


src/runtime_src/tools/scripts/service_bundle.sh
```diff
@@ -58,6 +58,12 @@ done
 # Funcs start here
 #############################
```
xdavidz, line 60: this is good, but please add the following script that I attached in the CR's description
```
for g in `find /sys/kernel/iommu_groups/* -maxdepth 0 -type d | sort -V`; do
    echo "IOMMU Group ${g##*/}:"
    for d in $g/devices/*; do
        echo -e "\t$(lspci -nns ${d##*/})"
    done;
done;
```

make this as a function and then dump the info into a separate file
field engineer is using this script.

prapulkrishnamurthy: Added changes as per comments.

src/runtime_src/tools/scripts/service_bundle.sh
```diff
@@ -182,6 +204,8 @@ funcArray=(
 	vmr_logs
 	debug_logs
 	core_dumps
+	iommu_dump
+	iommu_groups
 )
```
xdavidz, line 209: can you please dump the iommu dump and iommu groups into host_info as iommu_dump.log and iommu_groups.log?

xdavidz: either 2 files or 1 file combined with two logs are ok

xdavidz (approved):

xdavidz:
@prapulkrishnamurthy this is approved by me last week. Please ping all stockholders for any other potential concerns or seek for approval. Then, you can ask Max, Vamshi or those who can merge your PR. Note: the PR owner is responsible for moving your PR forward.

rozumx (approved):
Looks good.

I do find it interesting that this script doesn't check to see if the output directory exists.   If in the future you find yourself updating the script, you might wish to look adding this EOU feature.    

Doing so with most likely result in the need of adding a "force" or a "yes" option to create if that situation should arise.

prapulkrishnamurthy:
> Looks good.
> 
> I do find it interesting that this script doesn't check to see if the output directory exists. If in the future you find yourself updating the script, you might wish to look adding this EOU feature.
> 
> Doing so with most likely result in the need of adding a "force" or a "yes" option to create if that situation should arise.

@rozumx thanks for the input and that sounds like a good improvement. But I don't think we need that because the script is designed to dump different logs and logs are allocated for different subfolders within a root folder; as simple as that. But adding more checks like this will complicate the ease of use (For field tester let's say) and we don't have to over-simplify it.

PR #7388, opened 2023-02-22
Disable warnings about anymous structs when compiled with GCC and -Wp…

qberthet:
When XRT is linked to projects compiled with GCC and using -Wpedantic, few warnings concerning anonymous structs are fired:
```
...
/opt/xilinx/xrt/include/xrt_mem.h:59:12: warning: ISO C++ prohibits anonymous structs [-Wpedantic]
   59 |     struct {
...
/opt/xilinx/xrt/include/ert.h:93:12: warning: ISO C++ prohibits anonymous structs [-Wpedantic]
   93 |     struct {
...
```
Added 
```
#if defined(__GNUC__)
# pragma GCC diagnostic push
# pragma GCC diagnostic ignored "-Wpedantic"
#endif
```
to disable -Wpedantic in the two headers causing the warning. 

**Note:** Another approach to fix these warning would have been to prefix the structs definition by `__extension__` intrinsic for GCC only as shown in [this](https://github.com/boostorg/winapi/commit/f004b4753d6fee6f68c90cfc65b075fa0a33d092) commit. This have the advantage of not disabling -Wpedantic for the full file but reduces the code readability. I have choosen the above solution as a similar warning silencing seems to be implemented in both file for WIN32.

-Wpedantic being disabled for these two file, it might silence usefull warning in the future.

The warnings are not fired during compilation anymore and the build is successful.
None that I am aware of.

### DIFF
--- a/src/runtime_src/core/include/ert.h
+++ b/src/runtime_src/core/include/ert.h
@@ -59,6 +59,11 @@
 # pragma warning( disable : 4201 )
 #endif
 
+#if defined(__GNUC__)
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wpedantic"
+#endif
+
 #define to_cfg_pkg(pkg) \
     ((struct ert_configure_cmd *)(pkg))
 #define to_start_krnl_pkg(pkg) \
--- a/src/runtime_src/core/include/ert.h
+++ b/src/runtime_src/core/include/ert.h
@@ -937,6 +937,10 @@ get_size_with_timestamps_or_zero(struct ert_packet *pkt)
 }
 #endif
 
+#if defined(__GNUC__)
+# pragma GCC diagnostic pop
+#endif
+
 #ifdef _WIN32
 # pragma warning( pop )
 #endif
--- a/src/runtime_src/core/include/xrt_mem.h
+++ b/src/runtime_src/core/include/xrt_mem.h
@@ -38,6 +38,11 @@
 # pragma warning( disable : 4201 )
 #endif
 
+#if defined(__GNUC__)
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wpedantic"
+#endif
+
 #ifdef __cplusplus
 # include <cstdint>
 extern "C" {
--- a/src/runtime_src/core/include/xrt_mem.h
+++ b/src/runtime_src/core/include/xrt_mem.h
@@ -114,6 +114,10 @@ enum xclDDRFlags {
 }
 #endif
 
+#if defined(__GNUC__)
+# pragma GCC diagnostic pop
+#endif
+
 #ifdef _WIN32
 # pragma warning( pop )
 #endif


qberthet:
Re-created with signed-off commit

stsoe (changes requested):
Yeah, thanks.  But isn't a pop needed at end of file?  

Looks like the same goes for the windows disabling in the same headers, but changing that may have ripple effects so we will leave them for now.

qberthet:
My bad, you're absolutely right.
PR updated.

Note that windows flags pop are already present at the end of the file.

stsoe (approved):
Great thanks!

keryell (commented):
Naïve question: why not giving a name to the struct. It will replace the lacking comment anyway which is MISSING. :-)

qberthet:
If that's an option for you, this is great, but be aware that this will change the access syntax:

if
```
struct xcl_bo_flags
{
  union {
    uint32_t flags;
    struct {
      uint16_t bank;       // [15-0]
      uint8_t  slot;       // [16-23]
      uint8_t  boflags;    // [24-31]
    };
  };
};
```
is changed for
```
struct xcl_bo_flags
{
  union {
    uint32_t flags;
    struct foo {
      uint16_t bank;       // [15-0]
      uint8_t  slot;       // [16-23]
      uint8_t  boflags;    // [24-31]
    };
  };
};
```
you will need to change all the `xcl_bo_flags.bank` access by `xcl_bo_flags.foo.bank`.
Would that be acceptable ?

stsoe:
> you will need to change all the `xcl_bo_flags.bank` access by `xcl_bo_flags.foo.bank`. Would that be acceptable ?

IMO, not at this point.   I prefer we stick with disabling the warnings.

keryell:
I guess there is another problem with the unnamed `union` too and you do not need the top-level `struct`.
It should be:
```c++
union xcl_bo_flags {
  uint32_t flags;
  struct individual {
    uint16_t bank;       // [15-0]
    uint8_t  slot;       // [16-23]
    uint8_t  boflags;    // [24-31]
  };
};
```

keryell:
Actually the `boflags` was just here because a name conflict due to anonymous struct... :-(
```c++
union xcl_bo_flags {
  uint32_t flags;
  struct individual {
    uint16_t bank;   // [15-0]
    uint8_t  slot;   // [16-23]
    uint8_t  flags;  // [24-31]
  };
};
```

qberthet:
As mentionned in the initial message, another approach could be something like
```
#if defined(__GNUC__)
#define GCC_DETAIL_EXTENSION __extension__
#else
#define GCC_DETAIL_EXTENSION
#endif

struct xcl_bo_flags
{
  GCC_DETAIL_EXTENSION union {
    uint32_t flags;
    GCC_DETAIL_EXTENSION struct {
      uint16_t bank;       // [15-0]
      uint8_t  slot;       // [16-23]
      uint8_t  boflags;    // [24-31]
    };
  };
};
```
Suppress the warnings only on the required lines without changing the access syntax.

stsoe:
Keep changes to a minimum now.  We will stick with the warning push/pop for now.

keryell:
> Keep changes to a minimum now. We will stick with the warning push/pop for now.

Yes, I guess the design train left the station long time ago...

stsoe:
ok to test

stsoe:
retest this please

PR #7396, opened 2023-02-27
CR-1150167 xvc: fix integer overflow case

rajkumar-xilinx:
Fixing integer overflow in xvc driver
CR-1150167
Added fix in xvc driver by adding conditional check on total bits received from user space.
NA
CR-1150167 test application
NA


rajkumar-xilinx:
Hi, I will add Max Zhen once internal review is over. Thank you.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xvc.c
```diff
@@ -170,7 +171,21 @@ static long xvc_ioctl_helper(struct xocl_xvc *xvc, const void __user *arg)
 	}
 
 	total_bits = xvc_obj.length;
-	total_bytes = (total_bits + 7) >> 3;
+
+	/* Fixing integer overflow scenario */
+	if (total_bits + 7 >= UINT_MAX)
```
chvamshi-xilinx, line 176: i think, this wont work, at the end you are actually adding 7 to total_bits which can go out of bounds.
You should not be adding 7 to  check whether it is greater to UINT_MAX. You should be checking the other way,
You just need to add following check and keep other logic same

if (total_bits >= UINT_MAX - 7 )
        {
                total_bits = UINT_MAX - 7 ;
        }
        total_bytes = total_bits + 7 >> 3

rajkumar-xilinx: Hi @chvamshi-xilinx thanks for the review. I will fix the logic. Thank you

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xvc.c
```diff
@@ -170,8 +171,19 @@ static long xvc_ioctl_helper(struct xocl_xvc *xvc, const void __user *arg)
...
 	total_bits = xvc_obj.length;
+
+	/* Fixing integer overflow scenario */
+	if (total_bits >= UINT_MAX - 7)
+		total_bits = UINT_MAX - 7;
+
 	total_bytes = (total_bits + 7) >> 3;
 
+	if (total_bytes == 0) {
+		pr_info("Err: int overflow, op 0x%x, len %u bits, %u bytes.\n",
```
chvamshi-xilinx, line 182: is this error valid if total_bits is zero? would there be a case where total_bits can be zero?

chvamshi-xilinx (approved):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xvc.c
rajkumar-xilinx: Not sure if driver receives total_bits as 0. But, If total_bits is 0, so total_bytes will also be 0. In the next step, driver tries to allocate mem for 0 bytes, which might be an issue. So, added a check for total_bits.

dayeh-xilinx:
retest this please.

chvamshi-xilinx (approved):

maxzhen (approved):

PR #7397, opened 2023-02-27
Fix for CR-1150166 & CR-1150159: adding checks to avoid out of bound heap access

saumyag-xilinx:
https://jira.xilinx.com/browse/CR-1150166
https://jira.xilinx.com/browse/CR-1150159
added necessary checks on header data like sectionOffset to avoid out of bound heap access 
na
tested using poc application provided by CR filer.
na


mamin506 (commented):
We don't want to do check XCLBIN data in driver.
Think about this, if XCLBIN data can go wrong, why not check "m_numSections"? How can we know m_sectionKind is correct? How to verify it? I can easily make more examples.

The xclbin needs to be fixed, if we see m_sectionSize, m_sectionOffset and/or m_length in that xclbin is incorrect. Driver has NO enough knowledge to tell if the data in xclbin is correct or not.

Please rethink about these two CRs.

PR #7409, opened 2023-03-01
Resolving the XDP Coverity issues related to AUTO_CAUSES_COPY

jvillarre:
This changes all of the opportunities identified by Coverity to replace auto with const auto& in the XDP owned code in order to avoid an unnecessary copy.

Low, as the eliminated copy does not effect the usage of the reference.

No documentation impact


hackwa (approved):
looks good!

IshitaGhosh (approved):
Changes look good.

PR #7413, opened 2023-03-02
replace 2&>1 by 2>&1

peutch:
Hi,
I think that more than this change is needed on this line, may be something like '> /dev/null 2>&1' in fact.
Moreover the line 224 is also strange: 2&1> /dev/null
Thanks!

ps: this follows the point here: https://lists.debian.org/debian-qa/2023/02/msg00052.html


stsoe:
ok to test

PR #7414, opened 2023-03-02
VITIS-5914 - Bug fixes and move to latest implementation

rbramand-xilinx:
Fixed coverity issues and bugs related to printing reports
Driver now returns buffer filled and cols used information when we query for aie tiles status, made changes according to that

Iterating over cols_filled bit map and parsing buffer for only those columns

Low

Tested on IPU and was able to print aie core, mem and shim tiles

NA

### DIFF
--- a/src/runtime_src/core/common/asd_parser.cpp
+++ b/src/runtime_src/core/common/asd_parser.cpp
@@ -352,16 +352,29 @@ size()
 
 uint64_t
 aie_core_tile_status::
-size(aie_tiles_info& info)
+size(const aie_tiles_info& info)
 {
   return sizeof(aie_dma_status) * info.core_dma_channels + sizeof(uint32_t) * info.core_events * 2 /*core, mem mode*/ +
       sizeof(uint8_t) * info.core_locks + sizeof(uint32_t) * 4 /*(cs,pc,sp,lr)*/;
 }
 
+// Convert Raw buffer data received from driver to core tile status
+// Sanity checks on buffer size is done in previous calls, so there will be no overflow
+// Buffer format: 
+//        Multiple columns with core, mem, shim tiles information
+// +-----------------------------------+
+// | core rows | mem rows | shim rows  |  col 0
+// |-----------------------------------|
+// | core rows | mem rows | shim rows  |  col 1
+// |-----------------------------------|
+// |          .........                |  col N
+// +-----------------------------------+
 void
 aie_core_tile_status::
-parse_buf(const char* buf, aie_tiles_info& info, std::vector<aie_tiles_status>& aie_status)
+parse_buf(const std::vector<char>& raw_buf, const aie_tiles_info& info, std::vector<aie_tiles_status>& aie_status)
 {
+  auto buf = raw_buf.data();
+
   for (auto &aie_col_status : aie_status) {
     for (auto &core : aie_col_status.core_tiles) {
       // DMA status
@@ -412,7 +425,7 @@ parse_buf(const char* buf, aie_tiles_info& info, std::vector<aie_tiles_status>& 
 }
 
 static bpt::ptree
-get_core_tile_info(aie_core_tile_status& core)
+get_core_tile_info(const aie_core_tile_status& core)
 {
   bpt::ptree pt;
   bpt::ptree core_pt;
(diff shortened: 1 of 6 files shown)


src/runtime_src/core/common/asd_parser.cpp
```diff
@@ -461,20 +462,26 @@ get_core_tile_info(aie_core_tile_status& core)
 
 bpt::ptree
 aie_core_tile_status::
-format_status(std::vector<aie_tiles_status>& aie_status, uint32_t start_col,
-              uint32_t cols, aie_tiles_info& tiles_info)
+format_status(std::vector<aie_tiles_status>& aie_status, uint32_t cols, aie_tiles_info& tiles_info,
```
stsoe, line 465: Should these by reference arguments not be const?  I don;t see them modified?

src/runtime_src/core/common/asd_parser.cpp
```diff
@@ -501,7 +508,7 @@ size(aie_tiles_info& info)
 
 void
 aie_mem_tile_status::
-parse_buf(char* buf, aie_tiles_info& info, std::vector<aie_tiles_status>& aie_status)
+parse_buf(const char* buf, aie_tiles_info& info, std::vector<aie_tiles_status>& aie_status)
```
stsoe, line 511: Ugh, how large is buf, what is buf?  Comments are missing.

stsoe (changes requested):
Functions in this file are magical.  Comments are lacking, raw buf pointers are used as source, but no size check, maybe it's fine because of how the functions are used, but lacking comments it hard to tell.  

It appears at least some references passed as arguments could be `const` ?

dayeh-xilinx:
retest this please.

src/runtime_src/core/common/asd_parser.cpp
rbramand-xilinx: Hi @stsoe, comments were present in .h file, I have added here as well now.
Sanity checks on buffer is already done before calling this function so there will be no overflow.
Thanks

stsoe (approved):

src/runtime_src/core/common/asd_parser.cpp
```diff
@@ -461,20 +475,26 @@ get_core_tile_info(aie_core_tile_status& core)
 
 bpt::ptree
 aie_core_tile_status::
-format_status(std::vector<aie_tiles_status>& aie_status, uint32_t start_col,
-              uint32_t cols, aie_tiles_info& tiles_info)
+format_status(const std::vector<aie_tiles_status>& aie_status, uint32_t cols, const aie_tiles_info& tiles_info,
+              uint32_t cols_filled)
 {
   bpt::ptree pt_array;
+  uint16_t col_count = 0;
```
dbenusov-xilinx, line 482: In the for loop below we use a different type from this one. I know we will never have uint16_t of aie cores (although that would be an interesting day if it does) these values should match.

There are a few places where this change should be made (the other for loops)
```suggestion
  uint32_t col_count = 0;
```

dbenusov-xilinx (approved):
Minor things to fix. Looks good to me!

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -929,18 +929,20 @@ struct aie_tiles_stats
...
+    uint32_t buf_size = data.col_size * data.num_cols;
 
-    std::vector<char> buf(data.col_size * data.num_cols, 0);
+    std::vector<char> buf(buf_size, 0);
 
-    // TODO : Add code to get the data
+    // TODO : Add code to get the data and cols filled info
 
-    return buf;
+    return {buf, cols_filled};
```
keryell, line 945: Or even more modern
```suggestion
    return { .buf = buf, .cols_filled = cols_filled };
```

rbramand-xilinx: Hi @keryell, thanks for the suggestion but this feature needs compiler to have C++20 support. XRT at present uses c++17 as default and windows flows also does not support c++20 yet. So let me create structure and return it for now. Thanks.

chvamshi-xilinx (approved):

src/runtime_src/core/pcie/linux/device_linux.cpp
keryell: Ah, too bad.

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -929,18 +929,23 @@ struct aie_tiles_stats
...
 
   static result_type
   get(const xrt_core::device* device, key_type key, const boost::any& param)
   {
     auto data = boost::any_cast<xrt_core::query::aie_tiles_status_info::parameters>(param);
+    uint32_t cols_filled = 0;
+    uint32_t buf_size = data.col_size * data.num_cols;
 
-    std::vector<char> buf(data.col_size * data.num_cols, 0);
+    std::vector<char> buf(buf_size, 0);
```
keryell, line 941:
```suggestion
    std::vector<char> buf(buf_size);
```

keryell (approved):
Thanks.

PR #7416, opened 2023-03-02
Support hardware queue dependencies

stsoe:
- Add xrt_core::fence and xrt_core::fence_handle internal classes
- Add xrt::run::enqueue() with and without dependencies

New methods xrt::run::enqueue() allows a run object to be enqueued with a dependency on other run objects.  A dependency on a run object is possible only if that run object has itself been enqueued().  In other words, legacy xrt::run::start() has not been changed and a run object started this way cannot be used as dependency on another run object submission.

Basic usage is something like this:
```
xrt::run r1{...};
xrt::run r2{...}
r1.enqueue();
r2.enqueue({r1});  // can be list of dependencies.
```

The second run object is enqueue on hw queue prior to r1 completing execution.

New flow, no particular risk

### DIFF
--- a/src/runtime_src/core/common/api/hw_queue.cpp
+++ b/src/runtime_src/core/common/api/hw_queue.cpp
@@ -7,6 +7,7 @@
 #include "hw_queue.h"
 
 #include "command.h"
+#include "fence.h"
 #include "hw_context_int.h"
 
 #include "core/common/debug.h"
@@ -359,6 +360,14 @@ public:
   virtual std::cv_status
   wait(const xrt_core::command* cmd, size_t timeout_ms) = 0;
 
+  // Enqueue a command returning a fence that can be waited on
+  virtual fence
+  enqueue(xrt_core::command* cmd) = 0;
+
+  // Enqueue a command with dependencies
+  virtual fence
+  enqueue(xrt_core::command* cmd, const std::vector<fence>& waits) = 0;
+
   // Managed start uses command manager for monitoring command
   // completion
   void
@@ -374,6 +383,7 @@ public:
   {
     submit(cmd);
   }
+
 };
 
 // class qds_device - queue implementation for shim queue support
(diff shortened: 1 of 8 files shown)


uday610:
So is `xrt::run::enqueue` and `xrt::run::start `are essentially same with the difference that enqueue can support dependency with another run handle? 
By the way, in the basic usage code example, I assumed there is small typo and read it as `r2.enqueue({r1});`

uday610 (approved):

maxzhen (commented):
So, if we submit command with dependents through hw queue handle, shim has to always return a fence handle? We can't submit commands with some dependents, but don't ask for a fence back?

stsoe:
> So, if we submit command with dependents through hw queue handle, shim has to always return a fence handle? We can't submit commands with some dependents, but don't ask for a fence back?

Correct, that's the way it is coded now.  It's a good observation.  Even if we keep the semantics that run objects started with `start()` cannot be waited on (no fence returned from `start()`), there really is no reason why the `start()` function couldn't accept a wait list of run objects as long as these were enqueued().

maxzhen:
> > So, if we submit command with dependents through hw queue handle, shim has to always return a fence handle? We can't submit commands with some dependents, but don't ask for a fence back?
> 
> Correct, that's the way it is coded now. It's a good observation. Even if we keep the semantics that run objects started with `start()` cannot be waited on (no fence returned from `start()`), there really is no reason why the `start()` function couldn't accept a wait list of run objects as long as these were enqueued().

Yes, so maybe we could extend the start() member function later for this.

maxzhen (approved):

src/runtime_src/core/common/api/hw_queue.cpp
```diff
@@ -359,6 +360,14 @@ class hw_queue_impl : public command_manager::executor
   virtual std::cv_status
   wait(const xrt_core::command* cmd, size_t timeout_ms) = 0;
 
+  // Enqueue a command returning a fence that can be waited on
+  virtual fence
+  enqueue(xrt_core::command* cmd) = 0;
+
+  // Enqueue a command with dependencies
+  virtual fence
+  enqueue(xrt_core::command* cmd, const std::vector<fence>& waits) = 0;
```
keryell, line 369: I think this is too constraining.
You could accept a range instead, to avoid systematically allocating a vector.
Of course since you are only in C++17, an object which work with std::cbegin() and `std::cend()` for example will play the concept of range.
But a generic object does not work well with dynamic polymorphism...
So at the end I do not know how to generalize as is...
Do you have a code example for the use case?

stsoe: This is purely implementation you are showing.  I do have to get from a list of run objects (application level) to internal fence_handles  I don't understand what you mean by constraining?  I will share example with you.  

There are two new APIs added at application level, both are in xrt::run.   They are marked experimental because I am not sure they are what we ultimately want.  I am not worried about the implementation you are showing, that can always change to anything more efficient you might have in mind.   Let's talk.

PR #7420, opened 2023-03-03
CR-1124404 Remove ctypes bindings from python tests

rchane:
https://jira.xilinx.com/browse/CR-1124404 update python designs to not use ctypes
ctypes bindings should be removed from python tests.
ctypes bindings were removed from the 23_bandwidth folder tests and utils_binding.py.
Testcases are being changed. This should not affect platforms with platform.json (we use .exe files instead of these older .py files for them), but it will affect those that don't. This PR is coming in close to 2023.1. 
I have tried testing locally on my u200 to check if tests still work, but please test robustly on other platforms (specifically, ones that actually use these .py files as mentioned above).
We may have to remove the README from https://github.com/Xilinx/XRT/tree/master/tests/python


tests/python/23_bandwidth/host_mem_23_bandwidth.py
```diff
@@ -17,80 +17,57 @@
...
-    try:
-        khandle3 = xrtPLKernelOpen(opt.handle, opt.xuuid, "bandwidth3")
-    except Exception as e:
-        return errno.EOPNOTSUPP
+    d = pyxrt.device(opt.index)
+    xbin = pyxrt.xclbin(opt.bitstreamFile)
+    uuid = d.load_xclbin(xbin)
 
-    kfunc = xrtKernelGetFunc(xrtBufferHandle, xrtBufferHandle, ctypes.c_int, ctypes.c_int)
+    khandle3 = pyxrt.kernel(d, uuid, "bandwidth3", pyxrt.kernel.shared)
```
chvamshi-xilinx, line 65: This was under an exception earlier. Please keep under exception. I dont think all platforms contains bandwidth3 kernel

rchane: Thank you, I've moved it back under the exception.

chvamshi-xilinx (approved):

PR #7424, opened 2023-03-06
Move to latest petalinux to fix apu package build time

rbramand-xilinx:


chvamshi-xilinx:
HI @rbramand-xilinx , Please file a SH ticket to preserve this build.

PR #7428, opened 2023-03-07
Additional profiling Coverity fixes

jvillarre:
Resolve recently introduced Coverity issues in profiling code.

The coverity issues 3109492, 310950, 310951, 310952, 310954, and 310956 were all introduced in pull request 7423 and are related to the addition of xrt_core::message::send in a base profiling class, which could throw an exception in some conditions.

Added try...catch blocks around the sending of the message to catch the possible exception.

Low risk as if the exception is thrown and caught, the only thing is an info message will not be displayed.


hackwa (approved):
looks good!

dayeh-xilinx:
retest this please.

PR #7431, opened 2023-03-08
Fix for CR-1143627: ASTeR TC18.01 - ul3x24 - Low DMA Write and Read bandwidth at 1MB Block size, and Low DMA Write bandwidth at 256MB Block size

saumyag-xilinx:
https://jira.xilinx.com/browse/CR-1143627
Added a readable message for the case when the block-size used is larger than the bank size, so that the reason for XRT skipping the test is clear.
low
Verified by using a block-size larger than the bank memory size:

Verbose: Enabling Verbosity
Test 1 [0000:37:00.1]     : dma 
    Description           : Run dma test
    Details               : Buffer size - '256 GB'                              
                            The bank does not have enough memory to allocate. Use lower
                            'block-size' value. 
                            Buffer size - '256 GB'
                            The bank does not have enough memory to allocate. Use lower
                            'block-size' value. 
                            Buffer size - '256 GB'
                            The bank does not have enough memory to allocate. Use lower
                            'block-size' value. 
                            Buffer size - '256 GB'
                            The bank does not have enough memory to allocate. Use lower
                            'block-size' value. 

### DIFF
--- a/src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
+++ b/src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
@@ -988,8 +988,11 @@ dmaTest(const std::shared_ptr<xrt_core::device>& _dev, boost::property_tree::ptr
 
     // check if the bank has enough memory to allocate
     // m_size is in KB so convert block_size (bytes) to KB for comparison
-    if (mem.m_size < (block_size/1024))
+    if (mem.m_size < (block_size/1024)) {
+      logger(_ptTest, "Details", boost::str(boost::format(
+	"The bank does not have enough memory to allocate. Use lower '%s' value. \n") % "block-size"));
       continue;
+    }
 
     size_t totalSize = 0;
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx (approved):

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -988,8 +988,11 @@ dmaTest(const std::shared_ptr<xrt_core::device>& _dev, boost::property_tree::ptr
 
     // check if the bank has enough memory to allocate
     // m_size is in KB so convert block_size (bytes) to KB for comparison
-    if (mem.m_size < (block_size/1024))
+    if (mem.m_size < (block_size/1024)) {
+      logger(_ptTest, "Details", boost::str(boost::format(
+	"The bank does not have enough memory to allocate. Use lower '%s' value. \n") % "block-size"));
```
chvamshi-xilinx, line 993: By reviewing the log message, it is better to inform the bank index also while printing this message.

PR #7438, opened 2023-03-10
CR-1150171: Fix issue in NIFD driver

jvillarre:
Fixes issues where a single value is copied from user memory into kernel memory multiple times, and where that value was used to increment through an array of kernel memory.  Since there was a possibility the user memory could change between the two reads, this led to unsafe behavior where we could have accessed memory past the allocated array.

The new code reads the critical user value just once, and reuses the value as the bound to iterate through the kernel memory.  This ensures the iteration count will not change and we will not iterate past the memory allocated for the array.

Very low risk as this is an isolated driver not loaded.

No documentation impact.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/nifd.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/nifd.c
@@ -327,14 +327,14 @@ static long readback_variable(struct xocl_nifd* nifd, void __user *arg)
     total_data_payload_size =
       ((num_bits * 2) + result_space_size) * sizeof(unsigned int);
 
-    kernel_memory = (unsigned int *)(vmalloc(total_data_size));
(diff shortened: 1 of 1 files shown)


maxzhen (approved):

dayeh-xilinx:
retest this please.

xdavidz:
> Build Failed! :(

This is weird pass. I found that both the xocl_init_nifd and xocl_fini_nifd are missing in the xocl_drv.h.
But how it can pass on some systems?

+++ b/src/runtime_src/core/pcie/driver/linux/xocl/xocl_drv.h
@@ -2823,6 +2823,9 @@ void xocl_fini_config_gpio(void);
 int __init xocl_init_xgq(void);
 void xocl_fini_xgq(void);
 
+int __init xocl_init_nifd(void);
+void xocl_fini_nifd(void)

PR #7445, opened 2023-03-15
LWPAECGDEV-69: Fixed using xrt::error.to_string() triggers an exception

yvonyuan:
Using xrt::error.to_string() when there is no async error will throw an exception, but we expect to get some information from the returned string, not triggering exception.
[LWPAECGDEV-69](https://ontrack-internal.amd.com/browse/LWPAECGDEV-69) 
Check if error code gets a value firstly when calling xrt::error.to_string(), and return "No async error was detected" when error code is still zero.
Tested `xrt_run_ert_packet.exe -v 3 --op 1 --xrt-err` on a birman board and the results are shown as below:
 
use_xrt_error

### DIFF
--- a/src/runtime_src/core/common/api/xrt_error.cpp
+++ b/src/runtime_src/core/common/api/xrt_error.cpp
@@ -250,7 +250,7 @@ public:
   to_string()
   {
     if (!m_errcode)
-      return std::string("No async error was detected");
+      return "No async error was detected";
 
     auto fmt = boost::format
       ("%s\n"
--- a/src/runtime_src/core/common/api/xrt_error.cpp
+++ b/src/runtime_src/core/common/api/xrt_error.cpp
@@ -249,6 +249,9 @@ public:
   std::string
   to_string()
   {
+    if (!m_errcode)
+      return std::string("No async error was detected");
+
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/xrt_error.cpp
```diff
@@ -249,6 +249,9 @@ class error_impl
   std::string
   to_string()
   {
+    if (!m_errcode)
+      return std::string("No async error was detected");
```
stsoe, line 253:
```suggestion
      return "No async error was detected";
```

stsoe (commented):
LGTM, no need for `std::string` when returning the string.

stsoe:
add to whitelist

src/runtime_src/core/common/api/xrt_error.cpp
yvonyuan: Removed std::string() and its test results remain the same.

yvonyuan:
Hi @stsoe, I re-requested your review, but it automatically removed requests from other reviewers. Could you please add someone to the reviewers again if necessary? I could not add anyone at my end. Thanks.

PR #7463, opened 2023-03-23
CR-1153585 - Fix Mpd msd binaries seg faults with latest XRT

rbramand-xilinx:
Fixed segmentation fault issue when running Mpd Msd cloud daemons

pr - https://github.com/Xilinx/XRT/pull/7104 introduced this bug. It was caught in regressions.

Mpd and Msd are linked to xrt_core and xrt_coreutil statically and this causes [static initialization order fiasco](http://yosefk.com/c++fqa/fqa.html#fqa-10.12) problem where a global object being built references another global that is not yet built. Solved this problem by lazy initialization i.e: initialization on first use.

Low

Tested Mpd Msd connection and ran validate tests on U250 card.

NA

### DIFF
--- a/src/runtime_src/core/common/config_reader.cpp
+++ b/src/runtime_src/core/common/config_reader.cpp
@@ -48,7 +48,7 @@ static std::mutex mutex;
 // values are statically cached, they can be changed only until they
 // have been accessed the very first time.  This map tracks first key
 // access.
-static std::set<std::string>
+static std::set<std::string>&
 get_lock_map()
 {
   static std::set<std::string> lock_map;
--- a/src/runtime_src/core/common/config_reader.cpp
+++ b/src/runtime_src/core/common/config_reader.cpp
@@ -43,25 +43,30 @@ namespace {
 
 namespace key {
 
+static std::mutex mutex;
 // Configuration values can be changed programmatically, but because
 // values are statically cached, they can be changed only until they
 // have been accessed the very first time.  This map tracks first key
 // access.
-static std::set<std::string> locked;
-static std::mutex mutex;
+static std::set<std::string>
+get_lock_map()
+{
+  static std::set<std::string> lock_map;
+  return lock_map;
+}
 
 static void
 lock(const std::string& key)
 {
   std::lock_guard<std::mutex> lk(mutex);
-  locked.insert(key);
+  get_lock_map().insert(key);
 }
 
 static bool
 is_locked(const std::string& key)
 {
   std::lock_guard<std::mutex> lk(mutex);
-  return locked.find(key) != locked.end();
+  return get_lock_map().find(key) != get_lock_map().end();
 }
 
 } // key
--- a/src/runtime_src/core/pcie/linux/pcidev.cpp
+++ b/src/runtime_src/core/pcie/linux/pcidev.cpp
@@ -157,21 +157,16 @@ namespace xrt_core { namespace pci {
 
 namespace sysfs {
 
-static std::string
-get_dev_root()
-{
-  static const std::string dev_root = "/sys/bus/pci/devices/";
-  return dev_root;
-}
+static constexpr const char* dev_root = "/sys/bus/pci/devices/";
 
 static std::string
 get_path(const std::string& name, const std::string& subdev, const std::string& entry)
 {
   std::string subdir;
-  if (get_subdev_dir_name(get_dev_root() + name, subdev, subdir) != 0)
+  if (get_subdev_dir_name(dev_root + name, subdev, subdir) != 0)
     return "";
 
-  auto path = get_dev_root();
+  std::string path = dev_root;
   path += name;
   path += "/";
   path += subdir;
@@ -213,7 +208,7 @@ open(const std::string& name,
   if (path.empty()) {
     std::stringstream ss;
     ss << "Failed to find subdirectory for " << subdev
-       << " under " << get_dev_root() + name << std::endl;
+       << " under " << dev_root + name << std::endl;
     err = ss.str();
   } else {
     fs = open_path(path, err, write, binary);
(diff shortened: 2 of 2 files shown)


src/runtime_src/core/common/config_reader.cpp
```diff
@@ -43,25 +43,30 @@ namespace {
...
 namespace key {
 
+static std::mutex mutex;
 // Configuration values can be changed programmatically, but because
 // values are statically cached, they can be changed only until they
 // have been accessed the very first time.  This map tracks first key
 // access.
-static std::set<std::string> locked;
-static std::mutex mutex;
+static std::set<std::string>
```
stsoe, line 51: This cannot be right?   The map is returned by value alas no changes to the value returned by `get_lock_map()` will ever be recorded.    I am curious how the the static global map in this file was a problem, where was it referenced before initialized?

src/runtime_src/core/pcie/linux/pcidev.cpp
```diff
@@ -157,16 +157,21 @@ namespace xrt_core { namespace pci {
 
 namespace sysfs {
 
-static const std::string dev_root = "/sys/bus/pci/devices/";
+static std::string
+get_dev_root()
+{
+  static const std::string dev_root = "/sys/bus/pci/devices/";
+  return dev_root;
+}
```
stsoe, line 165: I think this should have been a `constexpr const char* dev_root = "..."` ?

stsoe (changes requested):
Ouch.

src/runtime_src/core/common/config_reader.cpp
chvamshi-xilinx: Yes Soren. Function should be
static std::set<std::string>& get_lock_map.

Regarding your question about static global set initialization,
https://github.com/Xilinx/XRT/blob/master/src/runtime_src/core/pcie/linux/system_linux.cpp#L74 gets initialized first which end up accessing the above lock_map, but this lock_map has not been initialized yet.

Here is the stack trace of this crash
```
`
(gdb) bt
#0  0x00007ffff68cf43a in ?? () from /lib64/libstdc++.so.6
#1  0x000000000046fef1 in std::_Rb_tree_iterator<std::string>::operator-- (this=0x7fffffffd8c0) at /opt/rh/devtoolset-9/root/usr/include/c++/9/bits/stl_tree.h:302
#2  0x000000000046f182 in std::_Rb_tree<std::string, std::string, std::_Identity<std::string>, std::less<std::string>, std::allocator<std::string> >::_M_get_insert_unique_pos (this=0x7457c0 <(anonymous namespace)::key::locked>, __k="Runtime.runtime_log") at /opt/rh/devtoolset-9/root/usr/include/c++/9/bits/stl_tree.h:2104
#3  0x000000000046e643 in std::_Rb_tree<std::string, std::string, std::_Identity<std::string>, std::less<std::string>, std::allocator<std::string> >::_M_insert_unique<std::string const&> (this=0x7457c0 <(anonymous namespace)::key::locked>, __v="Runtime.runtime_log") at /opt/rh/devtoolset-9/root/usr/include/c++/9/bits/stl_tree.h:2147
#4  0x000000000046de83 in std::set<std::string, std::less<std::string>, std::allocator<std::string> >::insert (this=0x7457c0 <(anonymous namespace)::key::locked>, 
    __x="Runtime.runtime_log") at /opt/rh/devtoolset-9/root/usr/include/c++/9/bits/stl_set.h:511
#5  0x000000000046c834 in (anonymous namespace)::key::lock (key="Runtime.runtime_log") at /scratch/XRT_Edge/XRT-1/src/runtime_src/core/common/config_reader.cpp:57
#6  0x000000000046d437 in xrt_core::config::detail::get_string_value (key=0x5f9b50 "Runtime.runtime_log", default_value="console")
    at /scratch/XRT_Edge/XRT-1/src/runtime_src/core/common/config_reader.cpp:223
#7  0x0000000000455ee4 in xrt_core::config::get_logging () at /scratch/XRT_Edge/XRT-1/src/runtime_src/core/common/config_reader.h:339
#8  0x0000000000455a3a in xrt_core::message::send (l=xrt::message::error, tag=0x5f6481 "WARNING", msg=0x7474a8 "XILINX_XRT not set")
    at /scratch/XRT_Edge/XRT-1/src/runtime_src/core/common/message.cpp:262
#9  0x0000000000454864 in xrt_core::send_exception_message (msg=0x7474a8 "XILINX_XRT not set", tag=0x5f6481 "WARNING")
    at /scratch/XRT_Edge/XRT-1/src/runtime_src/core/common/error.cpp:28
#10 0x0000000000435866 in xrt_core::system_linux::system_linux (this=0x745360 <(anonymous namespace)::singleton_system_linux()::singleton>)
    at /scratch/XRT_Edge/XRT-1/src/runtime_src/core/pcie/linux/system_linux.cpp:176
#11 0x0000000000434dd6 in (anonymous namespace)::singleton_system_linux () at /scratch/XRT_Edge/XRT-1/src/runtime_src/core/pcie/linux/system_linux.cpp:67
#12 0x0000000000434e3d in (anonymous namespace)::X::X (this=0x7453d0 <(anonymous namespace)::x>)
    at /scratch/XRT_Edge/XRT-1/src/runtime_src/core/pcie/linux/system_linux.cpp:74
#13 0x00000000004371d8 in __static_initialization_and_destruction_0 (__initialize_p=1, __priority=65535)
    at /scratch/XRT_Edge/XRT-1/src/runtime_src/core/pcie/linux/system_linux.cpp:75
#14 0x00000000004371ee in _GLOBAL__sub_I_system_linux.cpp(void) () at /scratch/XRT_Edge/XRT-1/src/runtime_src/core/pcie/linux/system_linux.cpp:375
#15 0x00000000005f34bd in __libc_csu_init ()
#16 0x00007ffff5f964e5 in __libc_start_main () from /lib64/libc.so.6
#17 0x0000000000414c55 in _start ()

`

```
You can easily reproduce this issue by running Debug/opt/xilinx/xrt/bin/msd from your already compiled sandbox.

stsoe: Got it thanks.  Bad idea when static global initialization is accessing static globals.  The message system is not ready before the application has loaded.  

I think this is an error in usage; not sure it makes sense to accommodate access to ini file prior to the application having fully loaded.  I am worried there may be more globals along the send_message path, not today maybe, but in the future accidentally or on purpose?  It's just not acceptable to access globals during static global initialization.  

So IMO it's more a problem in system_linux.cpp and how it is initializing itself.   It is not just that it is calling into the message system (and config_reader), but the constructor is doing a lot more than just that and it is all happening at initialization time.  This was never the case prior to #7104 and this PR merely patches a bigger problem.  

Let it slide for 23.1 but this needs to be added as a revisit in the ishim project.

dayeh-xilinx:
retest this please.

src/runtime_src/core/common/config_reader.cpp
chvamshi-xilinx: I agree @stsoe , this is a bigger problem. How do we prevent this kind of situation in future? just providing a guidance to the developers? or is there any better way?

chvamshi-xilinx (approved):
Approved. We need a generic fix, but this is good for 2023.1

stsoe (approved):

PR #7455, opened 2023-03-21
Added PL Reset support from ZOCL driver

saifuddin-xilinx:
#### Problem solved by the commit
PL kernels can be reset by asserting and de-asserting PMC reset (see reset connection in base platform next page):
devmem 0x00F1260330 32 0xf	
devmem 0x00F1260330 32 0x0

XRT ZOCL driver will do that programmatically.  Basically it does following steps:
1.	Cleanup Outstanding commands/context related to PL.
2.	Cleanup ZOCL data structure related to PL and AIE  of slot 0.
3.	devmem 0x00F1260330 32 0xf & devmem 0x00F1260330 32 0x0 

#### Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered
-- New features 

#### How problem was solved, alternative solutions (if any) and why they were rejected
Added that support to XRT ZOCL driver 

#### Risks (if any) associated the changes in the commit
N/A

#### What has been tested and how, request additional testing if necessary
Basic validation testing. 
Also from different terminal run a test case and reset PL from different terminal

#### Documentation impact (if any)
N/A

PR #7472, opened 2023-03-27
Add changes to support xbutil reset on Edge Platforms

rbramand-xilinx:
Added shim changes to support `xbutil reset ` on edge platforms
PR - https://github.com/Xilinx/XRT/pull/7455 added driver changes to support reset on zocl, this PR is an extension which adds shim level changes to support it

zocl exports sysfs entry zocl_reset and writing '1' to it to initiate reset

Low 

Tested reset functionality on vck190 device

NA

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/zocl_kds.c
+++ b/src/runtime_src/core/edge/drm/zocl/zocl_kds.c
@@ -1165,6 +1165,9 @@ zocl_reset(struct drm_zocl_dev *zdev, const char *buf, size_t count)
 	mutex_lock(&kds->lock);
         /* Find out number of active client and send a signal to it. */
         list_for_each_entry_safe(client, tmp_client, &kds->clients, link) {
+                if (pid_nr(client->pid) == pid_nr(task_tgid(current)))
+                        continue;
+
                 ret = kill_pid(client->pid, SIGTERM, 1);
                 if (ret) {
                         DRM_WARN("Failed to terminate Client pid %d."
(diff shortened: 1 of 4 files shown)


chvamshi-xilinx (changes requested):

chvamshi-xilinx (approved):

src/runtime_src/core/edge/drm/zocl/zocl_kds.c
```diff
@@ -1165,6 +1165,9 @@ zocl_reset(struct drm_zocl_dev *zdev, const char *buf, size_t count)
 	mutex_lock(&kds->lock);
         /* Find out number of active client and send a signal to it. */
         list_for_each_entry_safe(client, tmp_client, &kds->clients, link) {
+                if (pid_nr(client->pid) == pid_nr(task_tgid(current)))
```
mamin506, line 1168: client->pid is "get_pid(task_pid(current))" in kds_core.c. Do we really want to compare with task_tgid()?

mamin506 (approved):
Look good to me. Just a question about PID. Please confirm.

src/runtime_src/core/edge/drm/zocl/zocl_kds.c
saifuddin-xilinx: Actually this checks is to ignore xbutil  as a client. While we are killing the PID of all the active client. We should not kill the xbutil itself, which is requested for reset.

saifuddin-xilinx (approved):

PR #7487, opened 2023-04-03
Fixed CU context related issue.

saifuddin-xilinx:
This problem has been found In [CR-1154047](https://jira.xilinx.com/browse/CR-1154047). 
But it's a generic issue of CU context management.
There is a cu context reference count present for each context. The number of Open/Close should match before free the context. But we are not doing the same from driver. 
Till not this is not an issue as userspace XRT only opening a single context for each CU. But recent changes of Userspace XRT introduced this changes and XRT driver is  started failing .  

-- We are now looping the cu context reference counter till 0, before freeing it up. 
-- The changes are impacted both XOCL/ZOCL driver and KDS

N/A

N/A
-- Specific test case and validation tests
-- V70 Edge and DC test case

N/A

### DIFF
--- a/src/runtime_src/core/common/drv/kds_core.c
+++ b/src/runtime_src/core/common/drv/kds_core.c
@@ -1120,6 +1120,7 @@ static inline void
 _kds_fini_client(struct kds_sched *kds, struct kds_client *client,
 		 struct kds_client_ctx *cctx)
 {
+	int ret = 0;
 	struct kds_client_cu_ctx *cu_ctx = NULL;
 	struct kds_client_cu_ctx *next = NULL;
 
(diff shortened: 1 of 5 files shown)


src/runtime_src/core/common/drv/kds_core.c
```diff
@@ -1249,13 +1253,19 @@ int kds_del_context(struct kds_sched *kds, struct kds_client *client,
...
+	if (!cu_ctx->ref_cnt) {
+		kds_err(client, "No opening CU Context");
+		return -EINVAL;
+	}
+
+	/* Decrement the ref count and check whether any existing context
+	 * present. If no active context exists then delete that cu context.
+	 */
+	if (--cu_ctx->ref_cnt)
+		return -EBUSY;
```
mamin506, line 1265: @saifuddin-xilinx , if ref_cnt is not zero, you can just return 0 instead of -EBUSY.

If you return 0, in xocl_del_context() and xocl_close_cu_context(), you don't need to convert -EBUSY to 0. Please avoid this kind of error code change. It is confusing and not easy to understand. For most of the time, it can be avoided. Usually, if we have to do that, put a comment to explain why it has to be done.

In kds_fini_hw_ctx_client() and _kds_fini_client(), you can use below code.
``` c
                while (cu_ctx->ref_cnt)
			kds_del_context(kds, client, cu_ctx);
```

Usually, kds_del_context will failed in the open context ioctl() code path. User passed in cu index, it might failed and the caller needs to check the return. But when close client and close hwctx, our code is fulling controling. It should not fail. If it failed, that should be a BUG that we need to fix.

mamin506 (approved):

src/runtime_src/core/common/drv/kds_core.c
saifuddin-xilinx: Thanks Min for the suggestion. Now changes are looking more simpler. Please take a look once again.

mamin506 (approved):
Thank you for fixing this. Good job man. 👍 

The logic is clear. 
kds_del_context() will decrement CU ref count by 1. It will release CU context once it drop down to 0.

When kds_fini_hw_ctx_client() is called and there are still opening CU context, kds_del_context will be called and it will make sure CU ref count is 0.

When kds_fini_client is called and there are openming HW context, kds_fini_hw_ctx_client() will be called.

maxzhen (approved):

chvamshi-xilinx (approved):

PR #7490, opened 2023-04-05
Use correct BO handle for export

IshitaGhosh:
When GMIO AIE Trace and PL Device Trace are enabled together, there is a segmentation fault after due to " Unable to export BO while attaching to AIE Driver".

The reported fail is caused by use of incorrect buffer handle with "xclExportBO" for trace buffer used in GMIO AIE Trace.

Access to xrt::bo handle to the allocated buffer is not available in aie_trace_offloader where GMIO AIE Trace buffer is managed and set up using AIE APIs. To solve this new methods are added to Device Interface and XDP Device abstractions which help in calling export_buffer on the allocated xrt::bo .
 
Reported test now passes with the fix and records AIE trace.


jvillarre (approved):

dayeh-xilinx:
retest this please.

PR #7492, opened 2023-04-06
Memory manger issue fixed for xbtest

saifuddin-xilinx:
 -- There is some extra flag in memory manager because of that one test if failing, 
Actually that flag is not allowing to init the memory manager when the platform memory topology and xclbin topology is not machine and xclbin topology has some extra memory banks  available. 
-- This issue has been found in https://jira.xilinx.com/browse/CR-1145960

-- Multislot memory manager changes introduced this  

-- Removed that flag as this will not be required anymore

N/A
xbtest, validate 
N/A


chvamshi-xilinx (approved):

maxzhen (approved):

saifuddin-xilinx:
Verifying more test cases

PR #7496, opened 2023-04-11
fix for channel numbers in MEM tiles

pgschuey:
Fixed issues with parsing and reporting MEM tile channels

For now, metric sets report a single channel, so we should only report one

None

Designs on vek280

MEM tile metric settings only need up to one channel


jvillarre (approved):

jvillarre:
Fix for approved CR-1159512

PR #7503, opened 2023-04-14
Remove redundant xdp::XrtDevice; Rename methods in DeviceIntf and xdp::Device, xdp::HalDevice related to trace buffers

IshitaGhosh:
xdp::XrtDevice abstraction is not used anymore. So removed it.
xdp::HalDevice abstraction returns an index to identify the Buffer Object allocated using its alloc method. This index is used in methods like free, sync, export, get address on device memory space. However, use of these indices were wrongly named as "buffer handle". This may cause wrong assumption and subsequent wrong use where actual buffer handle is expected. So, rename the uses to indicate "index" and not "buffer handle"
Rename methods related to trace buffers in DeviceIntf and xdp::Device/HalDevice.
 

Unit tests using trac ebuffers


jvillarre (approved):

dayeh-xilinx:
retest this please.

dayeh-xilinx:
retest this please.

PR #7506, opened 2023-04-17
Fixed DMA validation test for new hw context

saifuddin-xilinx:
[CR-1159505](https://jira.xilinx.com/browse/CR-1159505) system crash during xrt stress test
-- The root cause of the issue is, as per the current design BO should be associated with a hw context. 
    If a BO is created without a context and there is a chance that the xclbin is replaced then Driver will through a error that BO is present while XCLBIN is replacing.  

-- This issue only happen for stress test. Single validation will not hit this issue.

-- Generic issue of stress test. 
-- Hw context & Multislot design is introduced this issue. 

-- Added hw context creation before allocating a BO. 
-- All legacy APIs need to update/remove. 

-- N/A

xbutil validation  and stress test script 

N/A

### DIFF
--- a/src/runtime_src/core/pcie/common/dmatest.h
+++ b/src/runtime_src/core/pcie/common/dmatest.h
@@ -45,6 +45,7 @@ namespace xcldev {
         using buffer_and_deleter = std::pair<std::unique_ptr<xrt_core::buffer_handle>, xrt_core::aligned_ptr_type>;
         std::vector<buffer_and_deleter> mBOList;
         std::shared_ptr<xrt_core::device> mHandle;
+	std::unique_ptr<xrt_core::hwctx_handle> mhwCtxHandle;
         size_t mSize;
         size_t mTotalSize;
         unsigned mFlags;
@@ -130,10 +131,13 @@ namespace xcldev {
             if (count > 0x40000)
                 count = 0x40000;
 
+	    xrt::hw_context::cfg_param_type cfg_param;
+	    mhwCtxHandle = mHandle->create_hw_context(mHandle->get_xclbin_uuid().get(), cfg_param, xrt::hw_context::access_mode::shared);
+	    mFlags |= ((mFlags & ~XRT_BO_FLAGS_MEMIDX_MASK) | (mhwCtxHandle->get_slotidx() << XRT_BO_FLAGS_SLOTIDX_SHIFT));
             for (long long i = 0; i < count; i++) {
                 // This can throw and callers of DMARunner are supposed to catch this.
                 xrt_core::aligned_ptr_type buf = xrt_core::aligned_alloc(xrt_core::getpagesize(), mSize);
-                auto bo = mHandle->alloc_bo(buf.get(), mSize, mFlags);
+                auto bo = mhwCtxHandle->alloc_bo(buf.get(), mSize, mFlags);
                 if (!bo)
                     break;
                 std::memset(buf.get(), mPattern, mSize);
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/pcie/common/dmatest.h
```diff
@@ -130,10 +131,13 @@ namespace xcldev {
             if (count > 0x40000)
                 count = 0x40000;
 
+	    xrt::hw_context::cfg_param_type cfg_param;
+	    mhwCtxHandle = mHandle->create_hw_context(mHandle->get_xclbin_uuid().get(), cfg_param, xrt::hw_context::access_mode::shared);
```
stsoe, line 135:
```suggestion
	    mhwCtxHandle = mHandle->create_hw_context(mHandle->get_xclbin_uuid().get(), {}, xrt::hw_context::access_mode::shared);
```

src/runtime_src/core/pcie/common/dmatest.h
```diff
@@ -144,7 +148,9 @@ namespace xcldev {
         }
 
         ~DMARunner()
-        {}
+        {
+	    mBOList.clear();
```
stsoe, line 152: Why?

stsoe (changes requested):

stsoe:
Also, please add PR description.

saifuddin-xilinx:
> Also, please add PR description.

Sure, Actually I was just checking the pipeline. Hence, I have added the flag "Please do no Review". :)
I will fix the review comments as suggested. Thanks

src/runtime_src/core/pcie/common/dmatest.h
saifuddin-xilinx: The BOs are getting freed after hw destructor getting called. 
Hence, after hw context destructed there are some BOs are still present.

src/runtime_src/core/pcie/common/dmatest.h
```diff
@@ -130,10 +131,13 @@ namespace xcldev {
             if (count > 0x40000)
                 count = 0x40000;
 
+	    xrt::hw_context::cfg_param_type cfg_param;
+	    mhwCtxHandle = mHandle->create_hw_context(mHandle->get_xclbin_uuid().get(), cfg_param, xrt::hw_context::access_mode::shared);
+	    mFlags |= ((mFlags & ~XRT_BO_FLAGS_MEMIDX_MASK) | (mhwCtxHandle->get_slotidx() << XRT_BO_FLAGS_SLOTIDX_SHIFT));
```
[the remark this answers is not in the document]
saifuddin-xilinx: Thanks. Fixed

src/runtime_src/core/pcie/common/dmatest.h
stsoe: But mBOList is an automatic object, it is destructed when DMARunner goes out of scope.  That's what I meant by "Why?"

saifuddin-xilinx: Buffers are keeping an information about hw context. If buffers are freed before hw context, then we are losing track. We need to make sure all buffers are freed up before hardware context goes out of scope.

stsoe: Ok, got it, but order of destruction is opposite order of construction so it is implicitly correct anyway.  But I like being explicit about it, but that definitely requires a comment in the code before calling clear(), it is not at all obvious.

dayeh-xilinx:
retest this please.

src/runtime_src/core/pcie/common/dmatest.h
saifuddin-xilinx: Added a comments for the same

saifuddin-xilinx:
retest this please

PR #7501, opened 2023-04-13
fix CR-1158004 (#7478)

mamin506:
This is porting from 2021.1 branch.

PR #7507, opened 2023-04-18
CR-1158005 Report if a device is in a bad state

dbenusov-xilinx:
https://jira.xilinx.com/browse/CR-1158005

Not a bug.

Use the sysfs node added in https://github.com/Xilinx/XRT/pull/7501 and associate that status with a hardware context. For legacy devices the status is associated with the single hardware context that represents the device. For devices with multiple hardware contexts the driver will populate the context status.

None

Ubuntu 20.04 U55C
```
dbenusov@xsjdbenusov50:/proj/rdi/staff/dbenusov$ xbutil examine -d 04:00 -r dynamic-regions
WARNING: Unexpected xocl version (2.14.0) was found. Expected 2.16.0, to match XRT tools.

-------------------------------------------------
[0000:04:00.1] : xilinx_u55c_gen3x16_xdma_base_3
-------------------------------------------------
  Hardware Context ID: 0
    Context Status: HEALTHY
    Xclbin UUID: E106E953-CF90-4024-E075-282D1A7D820B
    PL Compute Units
      Index  Name             Base Address  Usage  Status
      0      verify:verify_1  0x800000      1      (IDLE)
```

None.

### DIFF
--- a/src/runtime_src/core/common/info_memory.cpp
+++ b/src/runtime_src/core/common/info_memory.cpp
@@ -612,7 +612,7 @@ populate_hardware_context(const xrt_core::device* device)
     ptree_type pt_hw;
     pt_hw.put("id", boost::algorithm::to_upper_copy(hw.id));
     pt_hw.put("xclbin_uuid", boost::algorithm::to_upper_copy(hw.xclbin_uuid));
-    pt_hw.put("status", boost::algorithm::to_upper_copy(hw.status));
+    pt_hw.put("status", std::to_string(hw.status));
     pt_hw.add_child("compute_units", populate_cus(device, hw.pl_compute_units, hw.ps_compute_units));
     pt.push_back(std::make_pair("", pt_hw));
   }
--- a/src/runtime_src/core/common/info_memory.cpp
+++ b/src/runtime_src/core/common/info_memory.cpp
@@ -593,6 +593,7 @@ populate_hardware_context(const xrt_core::device* device)
     }
 
     try {
+      // For legacy devices  mark the single hardware context status as the device status
       hw_context.context_status = xrt_core::device_query<xq::device_status>(device);
       hw_context.pl_compute_units = xrt_core::device_query<xq::kds_cu_info>(device);
       hw_context.ps_compute_units = xrt_core::device_query<xq::kds_scu_info>(device);
--- a/src/runtime_src/core/common/info_memory.cpp
+++ b/src/runtime_src/core/common/info_memory.cpp
@@ -593,7 +593,7 @@ populate_hardware_context(const xrt_core::device* device)
     }
 
     try {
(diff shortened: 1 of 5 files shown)


src/runtime_src/core/common/info_memory.cpp
```diff
@@ -593,6 +593,7 @@ populate_hardware_context(const xrt_core::device* device)
     }
 
     try {
+      hw_context.context_status = xrt_core::device_query<xq::device_status>(device);
```
stsoe, line 596: The query is for device_status, the result in context_status, which is which and what is what?

dbenusov-xilinx: I will add a comment explaining why. For management devices the device status node states if a device needs to be reset or not due to a hanging kernel. However, a device can only hang when it is has some xclbin loaded! This clearly ties into having the hardware context have some sort of 'hanging' status. For legacy devices this ends up with the context status of the hardware context being the device status! For future devices the status will be returned through the hw_context query and this node will be used for the management partitions.

manikandan-xilinx:
retest this please.

chvamshi-xilinx:
am confused, why are we mapping the hardware context status with device status? device status is not associated with any hardware contexts. We may need to add this information in platform report???

dbenusov-xilinx:
> am confused, why are we mapping the hardware context status with device status? device status is not associated with any hardware contexts. We may need to add this information in platform report???

For management devices the device status node states if a device needs to be reset or not due to a hanging kernel. However, a device can only hang when it is has some xclbin loaded! This clearly ties into having the hardware context have some sort of 'hanging' status. For legacy devices this ends up with the context status of the hardware context being the device status! For future devices each hardware context will have their status will be independently of this node and this node will be used for the management partitions.

chvamshi-xilinx:
> > am confused, why are we mapping the hardware context status with device status? device status is not associated with any hardware contexts. We may need to add this information in platform report???
> 
> For management devices the device status node states if a device needs to be reset or not due to a hanging kernel. However, a device can only hang when it is has some xclbin loaded! This clearly ties into having the hardware context have some sort of 'hanging' status. For legacy devices this ends up with the context status of the hardware context being the device status! For future devices each hardware context will have their status will be independently of this node and this node will be used for the management partitions.

Hi @dbenusov-xilinx , even though we have multiple hardware contexts, we have only one scheduler(kds in alveo) per device. it never happens that, one hardware context is in good state whereas other is in bad state.  Whatever kds is reporting is the device status.
We can have an offline discussion about where to keep the device status information.

dayeh-xilinx:
retest this please.

dayeh-xilinx:
retest this please.

dayeh-xilinx:
retest this please.

PR #7508, opened 2023-04-19
EDGEML-4839 load xrt_core.dll from where xrt_coreutil.dll is

GraceWangA:
By default, load xrt_core.dll in the same folder as xrt_coreutil.dll, instead of C:/Windows/System32/AMD folder.

Not a bug fix, required by usage.  Jira ID: EDGEML-4839

Look for the current executing code(xrt_coreutil.dll) folder. 

XILINX_XRT environment variable can still be used to set xrt_core.dll.  

test with VAI-RT test app by put DLLs in a specific folder other than C:/Windows/System32/AMD.

### DIFF
--- a/src/runtime_src/core/common/dlfcn.h
+++ b/src/runtime_src/core/common/dlfcn.h
@@ -80,6 +80,15 @@ dlsym(void* handle, const char* symbol)
   return ::GetProcAddress(HMODULE(handle),symbol);
 }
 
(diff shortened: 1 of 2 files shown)


stsoe (approved):

stsoe:
ok to test

stsoe:
retest this please - could not lock config file

PR #7515, opened 2023-04-20
CR_1158638_Sysfs node size overflow fix

prapulkrishnamurthy:
Boundary case check improvised to trim the max size of data at sysfs node to be 4K.

Tested and verified on VCK5000 and found the size is within limit.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
@@ -431,13 +431,14 @@ static size_t xgq_vmr_log_dump(struct xocl_xgq_vmr *xgq, int num_recs, char *buf
 				sizeof(log));
 			log_idx = (log_idx + 1) % VMR_LOG_MAX_RECS;
 
-			/* calling call back function */
-			count += dump_cb(xgq, buf ? buf + count : NULL, log.log_buf);
-			if (count > PAGE_SIZE) {
-				XGQ_WARN(xgq, "message size %ld exceeds page %ld",
+			if((PAGE_SIZE) - count < VMR_LOG_ENTRY_SIZE){
+				XGQ_WARN(xgq, "Ignoring messages size %ld exceeds page %ld",
 					count, PAGE_SIZE);
 				break;
 			}
+
+			/* calling call back function */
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
```diff
@@ -431,13 +431,14 @@ static size_t xgq_vmr_log_dump(struct xocl_xgq_vmr *xgq, int num_recs, char *buf
 				sizeof(log));
 			log_idx = (log_idx + 1) % VMR_LOG_MAX_RECS;
 
-			/* calling call back function */
-			count += dump_cb(xgq, buf ? buf + count : NULL, log.log_buf);
```
xdavidz, line 435: since we are copying from log.log_buf, can you use the sizeof(log_buf)? I know it is the same, but it is not intuitive for any other developers to understand what is the size.

xdavidz: and it is natural to use count + sizeof(log.log_buf) < PAGE_SIZE

dbenusov-xilinx:
If we are not trying to upstream this part of the driver, you could also use a binary sysfs node. kds_cu_stats and kds_scu_stats are sysfs nodes that may exceed 4K, so we converted them into binary sysfs nodes. The drawback is that the read function is called once for each 4K block and cannot return data in a single batch. May be worth looking into.

prapulkrishnamurthy:
retest this please

maxzhen (approved):

xdavidz:
> If we are not trying to upstream this part of the driver, you could also use a binary sysfs node. kds_cu_stats and kds_scu_stats are sysfs nodes that may exceed 4K, so we converted them into binary sysfs nodes. The drawback is that the read function is called once for each 4K block and cannot return data in a single batch. May be worth looking into.

We are going to upstream. The vmr_log should not be exceed 4k anytime, this concerns me that something corrupted already. Thus, this is better not exceed 4k.

xdavidz (approved):
I am ok with this. Please get this fixed asap.

prapulkrishnamurthy:
retest this please

prapulkrishnamurthy:
retest this please

dayeh-xilinx:
retest this please.

dayeh-xilinx:
retest this please.

PR #7520, opened 2023-04-25
Adding ability to skip git submodules update

manikandan-xilinx:
Ability to provide options the build.sh to skip the git submodule update if needed

It will help fix the issue "could not lock config file" when multiple parallels builds and run together and we want to avoid git submodule update as in CI it would be done during the workspace sync step itself

This does not have a breaking change. So, If the code is merged we can check in CI if we can ignore git submodule update by passing the -noinit argument


stsoe (approved):

manikandan-xilinx:
retest this please.

dayeh-xilinx:
retest this please.

stsoe:
add to whitelist

stsoe:
ok to test

dayeh-xilinx:
retest this please.

PR #7528, opened 2023-05-02
Cherry pick of #7527

saifuddin-xilinx:

There are multiple issue fixed by this PR

1. [CR-1161789](https://jira.xilinx.com/browse/CR-1161789)
-- Memory allocation failed as for PS kernel the memory size calculation is wrong. We should calculate by byte instead of KB.

2. OOM terminate called after throwing an instance of 'std::bad_alloc
-- The issue is generic. For execbuf we should not check the slot/hw context id in the driver,
-- But current code doesn't check that. As there is always a hw context 0 present we never hit this issue.
-- The issue hit now as for the specific test it's closed the hw context 0 and active hw context are started from 1.

3. For PS kernel case : If existing xclbin present then it's failing.
-- For existing xclbin case we need to return the existing slot it so that a new hw context has been created 2for the same.

-- Multiple regression test cases hit this issue.

-- Fixed all of then

N/A

-- Validation and specific test case

N/A

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_bo.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_bo.c
@@ -559,16 +559,18 @@ __xocl_create_bo_ioctl(struct drm_device *dev, struct drm_file *filp,
 	uint32_t slot_id = 0;
 	int ret;
 
-	/* Currently userspace will provide the corresponding hw context id.
-	 * Driver has to map that hw context to the corresponding slot id.
-	 */
-	hw_ctx_id = xocl_bo_slot_idx(args->flags);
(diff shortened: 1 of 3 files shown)


chvamshi-xilinx:
Hi @saifuddin-xilinx , it is better to cherry-pick this change from master after it merged into master. Keep this PR open to check if there are any pipeline issues.

jeffli-xilinx (approved):

dayeh-xilinx:
retest this please.

saifuddin-xilinx:
retest this please

maxzhen (approved):

chvamshi-xilinx:
retest this please

saifuddin-xilinx:
Created a separate  PR https://github.com/Xilinx/XRT/pull/7534

PR #7532, opened 2023-05-03
Cherry-pick of #7527

saifuddin-xilinx:

Cherry pick of https://github.com/Xilinx/XRT/pull/7527

There are multiple issue fixed by this PR

[CR-1161789](https://jira.xilinx.com/browse/CR-1161789)
-- Memory allocation failed as for PS kernel the memory size calculation is wrong. We should calculate by byte instead of KB.

OOM terminate called after throwing an instance of 'std::bad_alloc
-- The issue is generic. For execbuf we should not check the slot/hw context id in the driver,
-- But current code doesn't check that. As there is always a hw context 0 present we never hit this issue.
-- The issue hit now as for the specific test it's closed the hw context 0 and active hw context are started from 1.

For PS kernel case : If existing xclbin present then it's failing.
-- For existing xclbin case we need to return the existing slot it so that a new hw context has been created 2for the same.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_bo.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_bo.c
@@ -559,16 +559,20 @@ __xocl_create_bo_ioctl(struct drm_device *dev, struct drm_file *filp,
 	uint32_t slot_id = 0;
 	int ret;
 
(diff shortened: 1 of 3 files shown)


saifuddin-xilinx:
Another PR exists 
https://github.com/Xilinx/XRT/pull/7528

PR #7544, opened 2023-05-10
Create xrt/xrt.h

wilderfield:
Provide users with an aggregating header file.

This will allow users who are less familiar with xrt code base to quickly bring all of the APIs into their compilation unit.

Ease of Use Enhancement

### DIFF
--- a/src/runtime_src/core/include/xrt/api.h
+++ b/src/runtime_src/core/include/xrt/api.h
@@ -1,0 +1,15 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2020-2022 Xilinx, Inc.  All rights reserved.
+// Copyright (C) 2022-2023 Advanced Micro Devices, Inc. All rights reserved.
+#ifndef _XRT_API_H_
+#define _XRT_API_H_
+
+#include <xrt/xrt_aie.h>
+#include <xrt/xrt_bo.h>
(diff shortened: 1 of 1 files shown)


stsoe:
> I would prefer `xrt/xrt.h`.

Yes,  I agree.  Even as we have a deprecated include/xrt.h, having xrt/xrt.h does not conflict.  Do watch out for the include guard though.

wilderfield:
> > I would prefer `xrt/xrt.h`.
> 
> Yes, I agree. Even as we have a deprecated include/xrt.h, having xrt/xrt.h does not conflict. Do watch out for the include guard though.
> 
> Personally, I hate these kind of include all buckets, I prefer include-what-you-use, but I know many like a mixed bag of useless stuff.

Would you prefer that I directly add xrt/xrt.h ?
Or do you prefer that I generate this header file at CMake config time by detecting all the header files in */include/xrt ?

The second approach would require no maintenance unless there is a repo restructure.

Once you provide guidance I’ll make new well formed PR.

stsoe:
> Would you prefer that I directly add xrt/xrt.h ?
> Or do you prefer that I generate this header file at CMake config time by detecting all the header files in */include/xrt ?

I prefer explicitly added xrt/xrt.h,

bryanloz-xilinx:
Please close this in favor of #7548

PR #7548, opened 2023-05-12
Add aggregating header file for xrt/ APIs

bryanloz-xilinx:
Ease of use enhancement. Single include can bring in all the pieces a new user needs.

Doesn't need to impact documentation

### DIFF
--- a/src/runtime_src/core/include/xrt/CMakeLists.txt
+++ b/src/runtime_src/core/include/xrt/CMakeLists.txt
@@ -8,7 +8,8 @@ set(XRT_XRT_HEADER_SRC
   xrt_graph.h
   xrt_hw_context.h
   xrt_kernel.h
-  xrt_uuid.h)
+  xrt_uuid.h
+  xrt.h)
 
 
 install (FILES ${XRT_XRT_HEADER_SRC} DESTINATION ${XRT_INSTALL_INCLUDE_DIR}/xrt COMPONENT ${XRT_DEV_COMPONENT})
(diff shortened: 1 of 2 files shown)


stsoe (approved):

uday610 (approved):

wilderfield:
There is some conflict between the old XRT.h and the new one in the unittests…

stsoe:
> There is some conflict between the old XRT.h and the new one in the unittests…

Yes.  Makes sense unfortunately.  I would like to think about getting rid of legacy include/xrt.h since we anyway stating the xcl APIs are no longer available in 23.2.

I am not keen on using a different name for the umbrella xrt.h you are creating.

keryell:
If the old one is `include/xrt.h`, why should it conflict with `include/xrt/xrt.h`?

wilderfield:
> If the old one is `include/xrt.h`, why should it conflict with `include/xrt/xrt.h`?

It must be that the test case being compiled is somehow supplied with both include/ and include/xrt as include paths. There is no other way for this to be a problem, but I haven’t actually debugged yet.

stsoe:
> If the old one is `include/xrt.h`, why should it conflict with `include/xrt/xrt.h`?

Because some xrt/*.h files want to include legacy xrt.h for typedefs that do not belong there but happen to be for historical reasons.
```
% cat xrt/xrt_bo.h
...
#include "xrt.h"   // <-- want legacy xrt.h
...
```
that ends up including the new umbrella.

stsoe:
I am not willing to spend cycles cleaning this up now.  Folks who want umbrellas, can create their own until we have cleaned up our mess for which we have a 23.2 story any way.

wilderfield:
Ok, this can wait then

vboggara-xilinx (approved):

stsoe:
Not ready for this PR

PR #7552, opened 2023-05-16
Added check for cfg_cu payload_size to be within MAX_CQ_SLOT_SIZE in xocl_kds.c

jeffli-xilinx:
PR-7178 added a check in zocl_ctrl_ert.c for XGQ payload_size > MAX_CQ_SLOT_SIZE.  
However, in xocl_kds.c there was no such check, and the payload size is also doubled to account for key-value pair command type.  This then exceeded the MAX_CQ_SLOT_SIZE and caused zocl's ERT not to initialize due to size check.
However, the actual arguments are within the valid slot size.
This change caps the payload_size to MAX_CQ_SLOT_SIZE if it's exceeded, and prints out an error message.
But there is probably a need to revisit key-value pair command buffer size calculation.

PR-7178

Check for payload_size against MAX_CQ_SLOT_SIZE, and set it to MAX_CQ_SLOT_SIZE if it exceeds it.

If a kernel has a very large key-value pair command it could  run into issues when running the actual kernel

V70

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_kds.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_kds.c
@@ -1950,6 +1950,10 @@ xocl_kds_xgq_cfg_cus(struct xocl_dev *xdev, xuid_t *xclbin_id, struct xrt_cu_inf
 		 * approach.
 		 */
 		cfg_cu->payload_size = cfg_cu->payload_size * 2;
+		if(cfg_cu->payload_size > MAX_CQ_SLOT_SIZE) {
+			userpf_err(xdev, "CU Argument Size %x > MAX_CQ_SLOT_SIZE, setting it to MAX_CQ_SLOT_SIZE!", cfg_cu->payload_size);
+			cfg_cu->payload_size = MAX_CQ_SLOT_SIZE;
+		}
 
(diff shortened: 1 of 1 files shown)


dayeh-xilinx:
retest this please.

maxzhen (commented):
do we want to return error in this case?

jeffli-xilinx:
> do we want to return error in this case?

Currently this might not be an actual error since we are doubling the size based on the argument offset to accommodate key-value pair type commands.  So if we are not using key-value pairs then it will not be an actual error

maxzhen (approved):

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_kds.c
```diff
@@ -1950,6 +1950,10 @@ xocl_kds_xgq_cfg_cus(struct xocl_dev *xdev, xuid_t *xclbin_id, struct xrt_cu_inf
 		 * approach.
 		 */
 		cfg_cu->payload_size = cfg_cu->payload_size * 2;
```
chvamshi-xilinx, line 1952: We may need to error out if the payload size is more than MAX_CQ_SLOT_SIZE before doubling. Otherwise, we may not be configuring CU properly.
We also need to ignore the streaming arguments if that is causing an issue.

chvamshi-xilinx (approved):

mamin506 (approved):

PR #7556, opened 2023-05-18
Vitis-7923 HW context independent buffer objects

stsoe:
Default group id to sentinel for host only buffers allocated with xrt::ext::bo consructors.

xrt::ext::bo allocates host only buffers that should not be validated against xclbin metadata when setting these BOs as kernel arguments. By using a sentinel, the validation check performed as part of xrt::run::set_arg can be special cased to not error out.

```
// Allocate a buffer object on a device
xrt::ext::bo bo{device, size, access_mode};

// Allocate a buffer object specific to a hwcxt
xrt::ext::bo bo{hwctx, size, access_mode};
```

The code path in xrt::ext::bo is experimental per
https://github.com/Xilinx/XRT/pull/7517, the changes in this PR do not affect existing code paths.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -1495,10 +1495,21 @@ adjust_buffer_flags(xrt::ext::bo::access_mode access)
   return flags.all;
 }
 
+// xrt::ext::bo is always a host only BO and group is not used default
+// to sentinel that can be special cased when BOs are validated as
+// part of setting kernel arguments.
+static int
+no_group_id()
+{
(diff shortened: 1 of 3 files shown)


vboggara-xilinx (approved):

stsoe:
Invalid pull request.  The change is not right.

Host only buffer objects should specify group_id for host bank.  Problem is that at construction time, the host bank id is unknown if xclbin is not loaded.  For now it is assumed that the group id for the host bank is always '0'.

I may be better to have a sentinel that designates host_bank and then resolve later when BO is set as argument at which point the actual group  id can be obtained from xclbin.

PR #7561, opened 2023-05-19
Vitis-7922 Synchronization of run objects between devices and processes

stsoe:
Expose new experimental API, xrt::fence, for synchronization of xrt::run objects between different devices or processes.

This PR reverts changes made in #7416 and replaces them with explicit xrt::fence synchronization.

Fence objects can be submitted for signaling or for wait. Two new APIs are added to xrt::run to facilitate the synchronization:

- `xrt::run::submit_wait(xrt::fence)` submits a fence for wait to the same hardware queue that executes the command associated with the run object.
- `xrt::run::submit_signal(xrt::fence)` submits a fence for signaling to the same hardware queue that executes the command associated with the run object.

For example, if a fence is used to synchronize to xrt::run objects, then one run object will signal the fence while the other run object will wait on the fence.  The signaling and wait happens in the hardware queues (or kernel driver) so there is CPU wait in the host application.

```
void
t1(xrt::fence f)
{
  r1(a, b, c);
  r1.submit_signal(f);
}

void
t2(xrt::fence f)
{
  r2.submit_wait(f);
  r2(c, d, e);
}

void
foo()
{
  xrt::fence f;
  t1(f);
  t2(f);
  ...;
}
```

Fence objects can be exported from one process and imported in another in the same way as buffer objecs, so multi-process pipelines can be created in this way.

A fence object has state associated with it.  The state is updated every time a fence is waited on and signaled.  It is not possible to wait on the same (as in by reference) fence object in two different run objects, because one would wait on the next signal state, while the other would wait on the next-next signal state.

If two or more run objects must synchronize on a fence, then each run object should use its own fence copied from the same original fence source.

The xrt::fence and APIs added in this PR are experimental.  Both submit_wait and submit_signal are subject to be merged into the xrt::run invocation APIs, but with this PR, the changes are self-contained without affecting any existing APIs.

No risk, new code, not implemented for Alveo or Edge

Standard tests, project specific tests in other repos.

Not at this point


stsoe:
retest this please.

manikandan-xilinx:
retest this please.

stsoe:
retest this please - edge apu, jeez

stsoe:
retest this please - docker run --rm -e ...  edge

stsoe:
retest this please

dayeh-xilinx:
retest this please.

dayeh-xilinx:
retest this please.

dayeh-xilinx:
retest this please.

dayeh-xilinx:
retest this please.

PR #7524, opened 2023-04-28
Vitis-7921 First-class execution buffer

stsoe:
#### Problem solved by the commit
Add first-class concept to represent instructions used by a function to execute on some configured hardware.

#### Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered
In some scenarios, an xclbin configures the hardware with an overlay that is used to execute control code / instructions specific to a function, or put another way, the function executes instructions on the overlay.

Without a first-class instruction buffer, the host code is passing instructions as a semi-random regular kernel argument under the assumption that firmware can locate the instructions from the arguments.

#### How problem was solved, alternative solutions (if any) and why they were rejected
This PR introduces `xrt::module` for representing a binary that contains instructions for one or more functions.  The module is used to create instruction buffers for specific functions and kernel execution will populate an ERT packet for a function with the instruction buffer address explicitly defined.

The ERT packet for kernel execution is the standard `ert_start_kernel_cmd`, but the data payload is reinterpreted as a `ert_dpu_data` struct, which has explicit fields for instruction buffer address and size.

Additional this PR add `xrt::elf` for managing an elf binary that in turn can be used to create an `xrt::module`.  The `xrt::elf` will eventually provide APIs for extracting elf meta-data, like function names, arguments and more.

#### Risks (if any) associated the changes in the commit
None.  Added code is new and not currently used.

PR #7572, opened 2023-06-02
Vitis-7921 First-class execution buffer

stsoe:
Add first-class concept to represent instructions used by a function to execute on some configured hardware.

In some scenarios, an xclbin configures the hardware with an overlay that is used to execute control code / instructions specific to a function, or put another way, the function executes instructions on the overlay.

Without a first-class instruction buffer, the host code is passing instructions as a semi-random regular kernel argument under the assumption that firmware can locate the instructions from the arguments.

This PR introduces xrt::module for representing a binary that contains instructions for one or more functions. The module is used to create instruction buffers for specific functions and kernel execution will populate an ERT packet for a function with the instruction buffer address explicitly defined.

The ERT packet for kernel execution is the standard ert_start_kernel_cmd, but the data payload is reinterpreted as a ert_dpu_data struct, which has explicit fields for instruction buffer address and size.

Additional this PR add xrt::elf for managing an elf binary that in turn can be used to create an xrt::module. The xrt::elf will eventually provide APIs for extracting elf meta-data, like function names, arguments and more.

This PR replaces #7524 with changes per spec:
- Construct xrt::module from elf or userptr.
- Construct xrt::kernel from hwctx and module
- Copy module to sram when creating xrt::run from kernel
- Prep for patching of global argument buffer address in xrt::run::set_arg

None. Added code is new and not currently used.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -1,0 +1,191 @@
+// Copyright (C) 2023 Advanced Micro Devices, Inc. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+#define XCL_DRIVER_DLL_EXPORT  // exporting xrt_module.h
+#define XRT_API_SOURCE         // exporting xrt_module.h
+#define XRT_CORE_COMMON_SOURCE // in same dll as core_common
+#include "experimental/xrt_module.h"
+#include "experimental/xrt_elf.h"
+
+#include "xrt/xrt_bo.h"
+#include "xrt/xrt_hw_context.h"
+#include "xrt/xrt_uuid.h"
+
+#include "module_int.h"
+#include "core/common/error.h"
+
+#include <cstring>
+#include <string>
+
+namespace {
+
+} // namespace
+
+namespace xrt {
+
+// class module_impl - Base class for different implementations
+//
+// An xrt::bo buffer object that represents the instructions and is
+// passed to hw execution.  This buffer object is constructed in
+// context of the kernel object that will use it.
+class module_impl
+{
+protected:
+  xrt::hw_context m_hwctx;
+  xrt::uuid m_cfg_uuid;     // matching hw configuration id
+
+public:
+  module_impl(xrt::hw_context hwctx)
+    : m_hwctx(std::move(hwctx))
+    , m_cfg_uuid(m_hwctx.get_xclbin_uuid())
+  {}
+
+  module_impl(const module_impl* parent)
+    : m_hwctx(parent->m_hwctx)
+    , m_cfg_uuid(parent->m_cfg_uuid)
+  {}
+
+  virtual
+  ~module_impl()
+  {}
+
+  xrt::uuid
+  get_cfg_uuid() const
+  {
+    return m_hwctx.get_xclbin_uuid();
+  }
+
+  xrt::hw_context
+  get_hw_context() const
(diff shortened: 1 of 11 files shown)


dayeh-xilinx:
retest this please.

stsoe:
retest this please - edge em intr

stsoe:
retest this please - docker run --rm

stsoe:
retest this please - edge em failures, but PR is not changing any edge flow

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+  virtual std::pair<const char*, size_t>
+  get_data() const override
+  {
+    return {m_buffer.data(), m_buffer.size()};
+  }
+};
+
+// class module_sram - Create an sram module from parent
+class module_sram : public module_impl
```
maxzhen, line 111: This is only applicable to IPU PHX and STX. There is no SRAM on Turin. Do we want a generic name? Since this is common code, we can support xrt::bo::flags::cacheable flag and malloc normal memory in shim on Turin. The downside would be that we have to make a copy of instruction buffer even on Turin....

src/runtime_src/core/include/ert.h
```diff
@@ -146,6 +146,19 @@ struct ert_start_kernel_cmd {
...
+/**
+ * struct ert_dpu_data - interpretation of data payload for ERT_START_DPU
+ *
+ * The data payload for ERT_START_DPU is interpreted as fixed instruction
+ * buffer address along with instruction count, followed by regular kernel
+ * arguments.
+ */
+struct ert_dpu_data {
+  uint64_t instruction_buffer;  /* buffer address 2 words */
+  uint32_t instruction_count;   /* number of instructions */
```
maxzhen, line 158: is this really number of instructions? if so, then how big is one instruction?

stsoe: No clue.    For raw ptr it was defined as number of instruction words and passed as kernel argument.  I actually think we should make it the size of the buffer object and leave it to KMD to convert to whatever firmware needs.

src/runtime_src/core/common/api/xrt_module.cpp
stsoe: Yeah, it's just a name, completely internal to this one .cpp file.   We can call it whatever we like and we can change the name whenever we want.   We also need to change the "cacheable" flag name, it makes no sense.  This code will change as we adopt it,  I am not changing at this point.

stsoe:
retest this please - docker run --rm

src/runtime_src/core/include/ert.h
maxzhen: Size of the buffer in bytes sounds good :-).

src/runtime_src/core/common/api/xrt_module.cpp
maxzhen: Sounds good.

maxzhen (approved):

dayeh-xilinx:
retest this please.

mamin506 (approved):

uday610 (approved):

dayeh-xilinx:
retest this please.

dayeh-xilinx:
retest this please. failed due to "singletask.lock"

vboggara-xilinx (approved):

PR #7581, opened 2023-06-07
RTP update issue for AIE

saifuddin-xilinx:
https://jira.xilinx.com/browse/CR-1164872
-- We run host.elf several times in a loop in the set-up. The hw runs crash when RTP ports are used. 
-- The issue is observed after 64 iteration 
-- With this changes the issue is not observed.

### DIFF
--- a/src/runtime_src/core/edge/user/aie/common_layer/adf_runtime_api.cpp
+++ b/src/runtime_src/core/edge/user/aie/common_layer/adf_runtime_api.cpp
@@ -439,7 +439,7 @@ err_code graph_api::update(const rtp_config* pRTPConfig, const void* pValue, siz
     // write the new selector value
     driverStatus |= XAie_DataMemWrWord(config_manager::s_pDevInst, selectorTile, pRTPConfig->selectorAddr, selector);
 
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx:
@manikandan-xilinx , looks like you need to backport -noinit change to 2022.2 branch also

manikandan-xilinx:
Hi @chvamshi-xilinx Is it possible to get a list of branches where this change could be needed? So that I'll plan to get this change in the required branches?

chvamshi-xilinx:
> Hi @chvamshi-xilinx Is it possible to get a list of branches where this change could be needed? So that I'll plan to get this change in the required branches?

You need to make this change in all the active branches as you are using same docker script to run pipeline steps.
Current active branches are , 2022.2, 2023.1 and master. As your change is pushed to 2023.1 and master already, 2022.2 is the only active branch which is left.

dayeh-xilinx:
retest this please.

manikandan-xilinx:
retest this please.

saifuddin-xilinx:
No longer required this changes

PR #7587, opened 2023-06-12
Fixed a issue for AIE context for multiple thread

saifuddin-xilinx:
-- There is a issue when from multiple thread is opening a AIE context. 
-- AIE context can be open by once, but because of multiple thread it's failing as the current code is not thread safe. 
-- Added a lock guard to make it thread safe.

### DIFF
--- a/src/runtime_src/core/edge/user/aie/aie.cpp
+++ b/src/runtime_src/core/edge/user/aie/aie.cpp
@@ -32,6 +32,8 @@
 
 namespace zynqaie {
 
+std::mutex am_lock;
+
 Aie::Aie(const std::shared_ptr<xrt_core::device>& device)
 {
     DevInst = {0};
@@ -113,6 +115,10 @@ open_context(const xrt_core::device* device, xrt::aie::access_mode am)
 {
 #ifndef __AIESIM__
   auto drv = ZYNQ::shim::handleCheck(device->get_device_handle());
+  std::lock_guard access_lock_guard(am_lock);
+  if (access_mode != xrt::aie::access_mode::none) {
+	  return;
+  }
 
   auto current_am = drv->getAIEAccessMode();
   if (current_am != xrt::aie::access_mode::none)
@@ -120,18 +126,19 @@ open_context(const xrt_core::device* device, xrt::aie::access_mode am)
 
   int ret = drv->openAIEContext(am);
   if (ret)
-      throw xrt_core::error(ret, "Fail to open AIE context");
+    throw xrt_core::error(ret, "Fail to open AIE context");
 
   drv->setAIEAccessMode(am);
   access_mode = am;
 #endif
 }
 
-bool
+xrt::aie::access_mode
 Aie::
-is_context_set()
+get_access_mode()
 {
-  return (access_mode != xrt::aie::access_mode::none);
+  std::lock_guard access_lock_guard(am_lock);
+  return access_mode;
 }
 
 void
@@ -141,7 +148,7 @@ sync_bo(xrt::bo& bo, const char *gmioName, enum xclBOSyncDirection dir, size_t s
   if (!devInst)
     throw xrt_core::error(-EINVAL, "Can't sync BO: AIE is not initialized");
 
-  if (access_mode == xrt::aie::access_mode::shared)
+  if (get_access_mode() == xrt::aie::access_mode::shared)
     throw xrt_core::error(-EPERM, "Shared AIE context can't sync BO");
 
   auto gmio_itr = gmio_apis.find(gmioName);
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/edge/user/aie/aie.cpp
```diff
@@ -32,6 +32,8 @@
 
 namespace zynqaie {
 
+std::mutex am_lock;
```
keryell, line 35: It looks like there is a comment missing here, right?

src/runtime_src/core/edge/user/aie/aie.cpp
```diff
@@ -113,25 +115,30 @@ open_context(const xrt_core::device* device, xrt::aie::access_mode am)
 {
 #ifndef __AIESIM__
   auto drv = ZYNQ::shim::handleCheck(device->get_device_handle());
+  std::lock_guard access_lock_guard(am_lock);
+  if (access_mode != xrt::aie::access_mode::none) {
```
keryell, line 119: Does this test require to hold the test longer?

keryell: Actually what this uncommented code does?

src/runtime_src/core/edge/user/aie/aie.cpp
```diff
@@ -113,25 +115,30 @@ open_context(const xrt_core::device* device, xrt::aie::access_mode am)
...
   drv->setAIEAccessMode(am);
   access_mode = am;
 #endif
 }
 
-bool
+xrt::aie::access_mode
 Aie::
-is_context_set()
+get_access_mode()
```
keryell, line 138: Curious formatting.
Is this an output of clang-format?

src/runtime_src/core/edge/user/aie/aie.cpp
keryell: I realize that neither the `.h` nor the `.cpp` have comments on their functions. This is not very inclusive for an open-source project.

larry9523:
@saifuddin-xilinx can you please show me the host code which causes the re-entry problem of open_context?. Is the xrt::aie::device created without specifying a access mode? And later on, two threads which share the same xrt::aie::device call open_context? Thought the open_context is a private function.

saifuddin-xilinx:
> @saifuddin-xilinx can you please show me the host code which causes the re-entry problem of open_context?. Is the xrt::aie::device created without specifying a access mode? And later on, two threads which share the same xrt::aie::device call open_context? Thought the open_context is a private function.

Hi Larry, 
As per our offline discussion, I have shared the application. Please take a look. Thanks

src/runtime_src/core/edge/user/aie/aie.cpp
saifuddin-xilinx: I just followed the same as used in this file before. I am not very sure about the formatting.

src/runtime_src/core/edge/user/aie/aie.cpp
saifuddin-xilinx: Added comments

src/runtime_src/core/edge/user/aie/aie.cpp
saifuddin-xilinx: Added comments. Thanks

src/runtime_src/core/edge/user/aie/aie.cpp
```diff
@@ -32,6 +32,12 @@
 
 namespace zynqaie {
 
+/* Added this lock to make AIE context thread safe. If AIE context is opened
```
keryell, line 35:
```suggestion
/* Use a lock to make AIE context thread safe. If AIE context is opened
```

keryell: Just describe the fact. No need to describe the fact that you added it, since it will not be useful for the readers 20 years from now.

src/runtime_src/core/edge/user/aie/aie.cpp
keryell: Yes, if this is the local coding style.

src/runtime_src/core/edge/user/aie/aie.cpp
```diff
@@ -113,25 +119,31 @@ open_context(const xrt_core::device* device, xrt::aie::access_mode am)
...
 
-bool
+xrt::aie::access_mode
 Aie::
-is_context_set()
+get_access_mode()
 {
-  return (access_mode != xrt::aie::access_mode::none);
+  std::lock_guard access_lock_guard(am_lock);
+  return access_mode;
```
keryell, line 146: I am unsure this solves always the problem.
The access to `access_mode` is safe but is it possible to have another thread changing it after `get_access_mode()` returns before taking an action?
There should be a design top-level comment explaining the precondition and invariant of the code.

saifuddin-xilinx:
This is not the expected fix. Will look this issue in a different PR

PR #7591, opened 2023-06-14
AIE trace improvements and new features

pgschuey:
Problem solved by the commit
* Added interface tile trace support
* Added graph-based interface tile profiling support
* Use edge events for memory tile lock stalls start/end
* Creation of timestamp file for system timeline support
* Moved AIE metadata parsing to common APIs
* Improved reporting of memory tiles (MEM -> memory)

* Lots of file changes in this PR
* Interface tile trace not tested since trace streams not yet connected

* Tested on vek280 and vck190

Documentation impact (if any)
* xrt.ini settings for interface tile trace and graph-based interface tile profiling
* System timeline support


src/runtime_src/xdp/profile/plugin/aie_trace/edge/aie_trace.cpp
```diff
@@ -282,90 +315,204 @@ namespace xdp {
...
+    }
+
+    // MEM tiles
+    if (metricSet.find("trace") != std::string::npos) {
+      XAie_EventSelectStrmPort(aieDevInst, loc, rscId, XAIE_STRMSW_SLAVE, TRACE, 0);
+    }
+    else {
+      auto slaveOrMaster = (metricSet.find("output") != std::string::npos) ?
+        XAIE_STRMSW_SLAVE : XAIE_STRMSW_MASTER;
+      XAie_EventSelectStrmPort(aieDevInst, loc, rscId, slaveOrMaster, DMA, channel);
```
hackwa, line 506: This part of the code totally doesn't make sense to me. We reserve a stream switch port and that gives us a resource ID and we map it to the appropriate channel. Why do we completely ignore this ID and trace some hardcoded IDs instead? This seems to completely circumvent the resource manager.

hackwa (changes requested):
I have seen similar code in aie profile and I think this could lead to problems as we are circumventing the resource manager. I found an example of how to get the event IDs after we reserve a stream switch port:
https://gitenterprise.xilinx.com/ai-engine/aie-rt/blob/main/fal/tests/utests/tc/test_ss.cpp#L48

jvillarre (approved):

src/runtime_src/xdp/profile/plugin/aie_trace/edge/aie_trace.cpp
pgschuey: @hackwa Thanks for catching this and the feedback. This will be addressed in a future PR.

PR #7594, opened 2023-06-16
Vitis 6327 Add PS kernel xclbins into APU Package

dbenusov-xilinx:
https://jira.xilinx.com/browse/VITIS-6327

The platform team requested that XRT create the xclbins that contain the PS kernels. They will generate the xclbins that contain the PL portions. This PR addresses adding the PS kernel xclbins into the APU package.

The validation PS kernels were added into the build for edge platforms. The installation directory for all generated ps kernels is `/lib/firmware/xilinx/ps_kernels`. This is the same directory in which the apu files are stored.

An additional json file is included to list out which PL xclbins are required to operate this PS kernel.

Created CMake files that generate the Xclbins using a passed in reference to a VITIS directory. Added logic to install the generated xclbins into `/lib/firmware/xilinx/ps_kernels` within `pkgapu.sh`. xbutil will need to be updated in a 2nd PR.

Only one APU package can be installed at a time. If the time comes where multiple APU packages can be installed we must revist how the PS kernels are packaged.

Additionally, I am concerned about sourcing an internal directory within the yocto build to get access to `xclbinutil`. It would be better to allow this to be set some other way so that non-internal customers can build the apu package.

Ubuntu 20.04
```
// Install the package
dbenusov@xsjdbenusov50:/proj/rdi/staff/dbenusov/XRT/build$ sudo apt-get install ./versal/apu_packages/xrt-apu-vck5000_202310.2.15.0_all.deb
Reading package lists... Done
Building dependency tree
Reading state information... Done
Note, selecting 'xrt-apu-vck5000' instead of './versal/apu_packages/xrt-apu-vck5000_202310.2.15.0_all.deb'
The following NEW packages will be installed:
  xrt-apu-vck5000
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
Need to get 0 B/90.3 MB of archives.
After this operation, 0 B of additional disk space will be used.
Get:1 /proj/xsjhdstaff6/dbenusov/XRT/build/versal/apu_packages/xrt-apu-vck5000_202310.2.15.0_all.deb xrt-apu-vck5000 all 2.15.0 [90.3 MB]
dpkg: warning: failed to open configuration file '/home/dbenusov/.dpkg.cfg' for reading: Permission denied
Selecting previously unselected package xrt-apu-vck5000.
(Reading database ... 258103 files and directories currently installed.)
Preparing to unpack .../xrt-apu-vck5000_202310.2.15.0_all.deb ...
Unpacking xrt-apu-vck5000 (2.15.0) ...
dpkg: warning: failed to open configuration file '/home/dbenusov/.dpkg.cfg' for reading: Permission denied
Setting up xrt-apu-vck5000 (2.15.0) ...

// Validate the contents are installed
dbenusov@xsjdbenusov50:/proj/rdi/staff/dbenusov/XRT/build$ ls /lib/firmware/xilinx/ps_kernels/
ps_aie.xclbin  ps_bandwidth.xclbin  ps_validate.xclbin  test_dependencies.json
```
None.


build/build_edge.sh
```diff
@@ -307,7 +308,8 @@ fi
 if [ -f $SETTINGS_FILE ]; then
     source $SETTINGS_FILE
 fi
-source $PETALINUX/settings.sh 
+source $PETALINUX/settings.sh
+source /proj/xbuilds/2023.2_daily_latest/installs/lin64/Vitis/2023.2/settings64.sh
```
jeffli-xilinx, line 312: Perhaps this should be moved to a separate file like vitis.build similar to petalinux.build?
this way the script just need to refer to the vitis.build and we can change the build version in a separate file.

src/runtime_src/tools/scripts/pkgapu.sh
```diff
@@ -177,7 +178,6 @@ if [[ ! -d $IMAGES_DIR ]]; then
 	error "Please specify the valid path of APU images by -images"
 fi
 IMAGES_DIR=`realpath $IMAGES_DIR`
-source /proj/xbuilds/2022.2_released/installs/lin64/Vitis/2022.2/settings64.sh
```
jeffli-xilinx, line 180: Similarly, we should keep this here since pkgapu.sh could potentially be called independently.
And source it from vitis.build env file.

jeffli-xilinx (changes requested):

dayeh-xilinx:
retest this please.

src/runtime_src/tools/scripts/pkgapu.sh
dbenusov-xilinx: Makes sense! Thanks for the suggestion :)

jeffli-xilinx (approved):

src/runtime_src/core/edge/ps_kernels/xrt/tests/validate/ps_aie_test/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+  DEPENDS ${PS_KERNEL_NAME}
+  )
+
+add_custom_target(${XCLBIN_TARGET} ALL
+  DEPENDS ${XCLBIN_LOCATION}
+  )
+
+install (FILES ${XCLBIN_LOCATION}
+  DESTINATION ${PS_KERNEL_INSTALL_DIR}
+  )
```
sonals, line 37: Can all three PS kernel tests share the same CMakeLists.txt? The three CMakeLists.txt look similar except for the kernel and library name.

dbenusov-xilinx: We could do that. However, since each directory is making a separate library and an xclbin I would prefer to keep them in their own files. As an example all of the AIE profiling ps kernels each have their own CMakeFile. I'll see if I can create a function an reduce the repetition!

src/runtime_src/core/edge/ps_kernels/xrt/tests/validate/ps_bandwidth_test/ps_bandwidth.cpp
```diff
@@ -0,0 +1,101 @@
...
+    return(handles);
+  }
+  
+__attribute__((visibility("default")))
+int bandwidth_kernel(int reps, double *max_throughput, xrtHandles *xrtHandle)
+{
+  double max_throughput_int = 0;
+
+  // Starting at 4K and going up to 16M with increments of power of 2
+  for (uint32_t i = 4 * 1024; i <= 16 * 1024 * 1024; i *= 2) {
```
rozumx, line 43: Thank you for the comment, without it we wouldn't really know the purpose behind this for loop.  That said it, this line should be refactored to something like this:
```
constexpr uint32_t 4K = 4 * 1024;
constexpr uint32_t 16M = 15 * 1024 * 1024;
for (uint32_t i = 4K; i <= 16M; i *= 2)

```
And yes, I know that 4K and 16M are not valid variable names, but you get the idea.

rozumx (approved):
Overall this this looks good.

Regarding the validation tests, it would be great if future PRs has more comments in the cpp source code to assist the next developer in debugging issues.

src/runtime_src/core/edge/ps_kernels/xrt/tests/validate/ps_bandwidth_test/ps_bandwidth.cpp
keryell: I guess someone came with a constexpr UDL library so you can write 4_Ki and 16_Mi https://en.cppreference.com/w/cpp/language/user_literal

chvamshi-xilinx (approved):

sabarinaxil:
retest this please

manikandan-xilinx:
retest this please.

manikandan-xilinx:
retest this please.

src/runtime_src/core/edge/ps_kernels/xrt/tests/validate/ps_bandwidth_test/ps_bandwidth.cpp
```diff
@@ -9,6 +9,14 @@
 
 #include "xrt/xrt_kernel.h"
 
+constexpr unsigned long long int operator "" _Kb(unsigned long long int n) {
+  return n * 1024;
+}
+
+constexpr unsigned long long int operator "" _Mb(unsigned long long int n) {
+  return n * 1024 * 1024;
+}
```
keryell, line 18: What a great idea. :-)
But if we go that path, there is nothing about bit quantity here and you should follow the standard naming: 
https://en.wikipedia.org/wiki/Binary_prefix#kibi
https://en.wikipedia.org/wiki/Binary_prefix#mebi

keryell: While you are on it, to have your colleagues benefiting from it, what about having this it in a private XRT header which is only included by the `.cpp` needing this?

keryell (approved):

src/runtime_src/core/edge/ps_kernels/xrt/tests/validate/ps_bandwidth_test/ps_bandwidth.cpp
dbenusov-xilinx: This is a source executable that does not have access to the majority of the XRT codebase, only a few include files for xrt_* types. If this was inside XRT proper we already have something like this.

sonals (approved):
We should move away from sk_types.h. However, that should be taken up as a separate effort.

PR #7609, opened 2023-06-30
CR-1167063 Fix hwqueue_handle::wait_command() API comment

stsoe:
Remove comment about cmd nullptr waiting for any command to complete.

Throw if nullptr is passed.  This means that OpenCL is not supported with platforms that implement hwqueue. If OpenCL support is needed, the command monitor needs to track order of submitted commands and wait on first submitted command before checking status of all running commands.


stsoe:
retest this please - windows build error unrelated to code changes

maxzhen (approved):

PR #7617, opened 2023-07-06
Obey CMAKE_INSTALL_PREFIX

bryanloz-xilinx:
If you want to install XRT into a user specified directory, XRT's CMake infra will not obey the user.
It will always put it into CMAKE_INSTALL_PREFIX/xrt

This change would make the CMake infra more obedient to the user.

Could potentially break existing flows.


bryanloz-xilinx:
I'm only 90% sure I want to make this happen. This PR also serves as a test to see if this change breaks anything.

stsoe (commented):
This indeed breaks many different repos that use XRT as a submodule.

Also, CMake itself prepends CMAKE_INSTALL_PREFIX to all variables referenced in install directories so now you have two levels of CMAKE_INSTALL_PREFIX.  

I agree it should be possible to set CMAKE_INSTALL_PREFIX to something different from what is hardwired, but the change in this PR will not do it.

Can we close and schedule a more thorough revamp of XRT's CMake infra.? It's long been on my wish list, but no resources assigned.

stsoe:
> I'm only 90% sure I want to make this happen. This PR also serves as a test to see if this change breaks anything.

Please do all testing outside the build pipeline in your own environment.

bryanloz-xilinx:
@stsoe Closing in favor of CMake revamp

PR #7629, opened 2023-07-16
Moving PS Kernel's pscontext to public XRT include area and under XRT namespace

jeffli-xilinx:
Moved pscontext class used in PS kernels to XRT's public include areas and under XRT namespace.
Header is now called xrt_pscontext.h, and sk_types.h should no longer be used.
Also updated all the current PS kernels under XRT repo to use this new header file.

All the existing PS kernel need to be updated to use the new pscontext class under XRT namespace

Tested on v70

### DIFF
--- a/src/runtime_src/core/include/experimental/xrt_pscontext.h
+++ b/src/runtime_src/core/include/experimental/xrt_pscontext.h
@@ -1,0 +1,31 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2023 Advanced Micro Devices, Inc. All rights reserved.
+#ifndef XRT_PSCONTEXT_H_
+#define XRT_PSCONTEXT_H_
+
+#include "xrt/detail/config.h"
+#include "xrt/detail/pimpl.h"
+#include "xclhal2.h"
+
+#ifdef __cplusplus
+#include <memory>
+
+/*
+ * PS Context Data Structure included by user PS kernel code
+ */
+
+namespace xrt {
+
+class pscontext_impl;
+class pscontext : public detail::pimpl<pscontext_impl> {
+public:
+  pscontext() = default;
+};
+
+}
+ 
+#else
+# error xrt_pscontext is only implemented for C++
+#endif // __cplusplus
+
+#endif
--- a/src/runtime_src/core/include/experimental/xrt_pscontext.h
+++ b/src/runtime_src/core/include/experimental/xrt_pscontext.h
@@ -11,7 +11,29 @@
 #include <memory>
 
 /*
- * PS Context Data Structure included by user PS kernel code
+ * PS Context Data Structure to be derived from for user's xrtHandle class
+ * User will need to declare an xrtHandles class derived from xrt::pscontext class
+ * Example:
+ * class xrtHandles : public xrt::pscontext
+ * {
+ * public:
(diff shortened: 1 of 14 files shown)


src/runtime_src/core/include/experimental/xrt_pscontext.h
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2023 Advanced Micro Devices, Inc. All rights reserved.
+#ifndef XRT_PSCONTEXT_H_
+#define XRT_PSCONTEXT_H_
+
+#include "xrt/detail/config.h"
+#include "xrt/detail/pimpl.h"
+#include "xclhal2.h"
```
stsoe, line 8:
```suggestion
```

src/runtime_src/core/include/experimental/xrt_pscontext.h
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * 
+ */
+
+namespace xrt {
+
+class pscontext_impl;
+class pscontext : public detail::pimpl<pscontext_impl> {
+public:
+  pscontext() = default;
```
stsoe, line 44: This does not look right.   Where is the `pscontext_impl` constructed?

src/runtime_src/core/include/experimental/xrt_pscontext.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include "xrt/detail/config.h"
+#include "xrt/detail/pimpl.h"
+#include "xclhal2.h"
+
+#ifdef __cplusplus
+#include <memory>
+
+/*
+ * PS Context Data Structure to be derived from for user's xrtHandle class
```
stsoe, line 14: This comment doesn't reflect usage.  The comment is for xrt::pscontext presumably, but that class is not derived from user's xrt_handle class.   Next line of comment is correct though.

Also can we use xrt_handle rather than camelCase.  There is no camelCase in xrt exposed APIs.

src/runtime_src/core/include/experimental/xrt_pscontext.h
```diff
@@ -0,0 +1,54 @@
...
+ * User will need to declare an xrtHandles class derived from xrt::pscontext class
+ * Example:
+ * class xrtHandles : public xrt::pscontext
+ * {
+ * public:
+ *   xrt::device dhdl;
+ *   xrt::kernel kernel;
+ *   xrtHandles(xclDeviceHandle dhdl_in, const xuid_t xclbin_uuid)
```
stsoe, line 22: Do not use `xclDeviceHandle`  use `xrt::device`.
And use xrt::uuid instead of plain uuid_t (aka xuid_t).

src/runtime_src/core/include/experimental/xrt_pscontext.h
```diff
@@ -0,0 +1,54 @@
...
+ *     : dhdl(dhdl_in)
+ *     , kernel(dhdl,xclbin_uuid,"kernel name")
```
stsoe, line 24:
```suggestion
 *     , kernel(dhdl, xclbin_uuid, "kernel name")
```

stsoe (changes requested):

PR #7632, opened 2023-07-18
Deprecate sk_types.h and move pscontext class to a separate file pscontext.h

jeffli-xilinx:
Move pscontext class out of sk_types.h and make sk_types.h to be deprecated

### DIFF
--- a/src/runtime_src/core/edge/include/sk_types.h
+++ b/src/runtime_src/core/edge/include/sk_types.h
@@ -52,6 +52,10 @@ typedef int (* kernel_t)(void *args, struct sk_operations *ops);
 
 #pragma message ("sk_types.h is deprecated and will be removed from the distribution directory in a future release.  Please use pscontext.h instead.")
 
+/*
+ * Including pscontext.h for backward compatibility with 
+ * PS kernels that were using sk_types.h
+ */
 #include "pscontext.h"
 
 #endif
--- a/src/runtime_src/core/edge/include/sk_types.h
+++ b/src/runtime_src/core/edge/include/sk_types.h
@@ -1,5 +1,6 @@
 /**
  * Copyright (C) 2019-2022 Xilinx, Inc
+ * Copyright (C) 2023 Advanced Micro Devices, Inc
  * Author(s): Min Ma	<min.ma@xilinx.com>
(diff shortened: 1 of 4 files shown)


stsoe (approved):

src/runtime_src/core/edge/include/sk_types.h
```diff
@@ -30,7 +31,8 @@
  * freeBO    : free BO handle.
  * logMsg    : send log messages to XRT driver for saving as per ini settings
  */
-struct sk_operations {
+struct XRT_DEPRECATED
```
sonals, line 34: Should the `XRT_DEPRECATED` come before `struct`?

src/runtime_src/core/edge/include/sk_types.h
```diff
@@ -45,29 +47,11 @@ struct sk_operations {
...
-
-struct pscontext::impl {
-private:
-  bool aie_profile_en;
-};
+#pragma message ("sk_types.h is deprecated and will be removed from the distribution directory in a future release.  Please use pscontext.h instead.")
 
-typedef pscontext* (* kernel_init_t)(xclDeviceHandle device, const uuid_t &uuid);
-typedef int (* kernel_fini_t)(pscontext *xrtHandles);
+#include "pscontext.h"
```
sonals, line 55: Can add a comment here saying this include is for backward compatibility?

sonals (changes requested):
Minor changes

src/runtime_src/core/edge/include/sk_types.h
jeffli-xilinx: I had it before struct, but got warning when compiling saying attribute ignored, and it need to follow struct keyword

src/runtime_src/core/edge/include/sk_types.h
jeffli-xilinx: Comment added

src/runtime_src/core/edge/include/sk_types.h
sonals: Thanks for detail.

sonals (approved):

chvamshi-xilinx (approved):

PR #7640, opened 2023-07-25
CR-1159280: Fixing profiling table to correctly associate argument with connection

jvillarre:
Profiling tables that report data movement between kernels and memory report the original kernel arguments associated with the connection.  With connections to host memory via host bridge, this association was not correctly made and the tables had empty entries.  Additionally, if multiple kernels were present in a design and had arguments with the same name, they were incorrectly being reported as attached to the wrong memory component.

The issue was discovered through comprehensive regression testing.

This change explicitly adds a check for host bridge connections.  Additionally, it changes the argument to memory mapping data structure to make the connection on a compute unit basis rather than a kernel basis, so multiple kernels with identical named arguments don't overwrite each other.

Low risk as this should only affect information reported in a single table.

Tested on original failing test case.

No documentation impact


dayeh-xilinx:
retest this please.

IshitaGhosh (approved):
Change looks good

PR #7641, opened 2023-07-25
Templatize internal utility aligned_alloc

stsoe:
Make aligned alloc return managed ptr of specified type.

Preserve existing usage of default void and exposed deleter, but hide the deleter for specified types.
Feels like this could be more elegant with better use of C++ features.


maxzhen (approved):

dayeh-xilinx:
retest this please.

PR #7647, opened 2023-07-28
Move hwctx initializtion of module to xrt::run

stsoe:
Ensure that creation of instruction buffer object is specific to a xrt::run object.  Treat the module associated with xrt::kernel as a template for all run objects created from the kernel.

The instruction buffer may be address patching for the kernel arguments applied prior to starting the kernel.  Invoking a kernel creates an xrt::run object and kernel arguments are specific to the run.  Setting a argument patches the instruction buffer with the address of the kernel argument buffer objects.


dayeh-xilinx:
retest this please.

manikandan-xilinx:
retest this please.

PR #7649, opened 2023-07-31
Fixed a issue where same client requested multiple times

saifuddin-xilinx:
-- There is a issue where same client requested the XRT driver multiple time. 
-- In our XRT driver we haven't any check before creating a client or destroying a client. 
-- In this PR I have added such checks based on the PID. If same PID is requested then it just increase a reference count and return, While deleting if the reference count is "0" then only it will cleanup that client.

### DIFF
--- a/src/runtime_src/core/common/drv/include/kds_client.h
+++ b/src/runtime_src/core/common/drv/include/kds_client.h
@@ -73,6 +73,7 @@ struct kds_client_cu_refcnt {
  */
 struct kds_client {
 	struct list_head	  link;
+	u32                       client_refcnt;
 	struct device	         *dev;
 	struct pid	         *pid;
 	struct mutex		  lock;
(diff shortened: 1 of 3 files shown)


saifuddin-xilinx:
retest this please

chvamshi-xilinx (approved):

saifuddin-xilinx:
retest this please

maxzhen (commented):
Conceptually, a client is mapped to a fd. Now after this change, multiple fd from same process will share the same client. What does not work before?

Is this change valid for other parts of XRT...Say, if a cu context is opened through one fd, it is considered as "opened" by any fd in that process?

saifuddin-xilinx:
> Conceptually, a client is mapped to a fd. Now after this change, multiple fd from same process will share the same client. What does not work before?
> 
> Is this change valid for other parts of XRT...Say, if a cu context is opened through one fd, it is considered as "opened" by any fd in that process?

Thanks for the feedback. I am checking with vvas team for the application.

saifuddin-xilinx:
No longer required. Application has to take care of opening multiple clients.

PR #7654, opened 2023-08-01
CR-1170195: Fixing issue with seg fault in AIE status/profiling/trace

jvillarre:
Processing the system_metadata section in profiling was crashing on designs that had PL kernels and AIE kernels.  This was because newly introduced code was based on the faulty assumption that only PL kernels and compute units were listed in the specific system_metadata sections that PL profiling was dependent on.  On designs that have AIE kernels in addition to PL kernels, we were going outside the bounds of our internal array.

This bug was introduced in pull request 7640 and it was discovered with testing on AIE designs.

This pull request no longer accesses the problematic array, but instead creates a local map of compute unit ID to compute unit name based solely on the information in the system_metadata section. 

Low risk as this fixes the seg fault and still maintains the functionality, but is dependent on the system metadata section format.

This has been tested on a mixed AIE + PL design on vek280 and a PL only design on the u200.

No documentation impact.

### DIFF
--- a/src/runtime_src/xdp/profile/database/static_info_database.cpp
+++ b/src/runtime_src/xdp/profile/database/static_info_database.cpp
@@ -1409,6 +1409,18 @@ namespace xdp {
       // Temp data structures to hold mappings of each CU's argument to memory
       typedef std::pair<std::string, std::string> fullName;
       std::map<fullName, int32_t> argumentToMemoryIndex;
+      std::map<int, std::string> computeUnitIdToName;
+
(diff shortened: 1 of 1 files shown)


IshitaGhosh (approved):
Change looks good. However, we may need to look deeper for cases where "compute_unit" metadata has argument name but no id name. This can be a corner case, if possible. As we have no knowledge about the schema of "compute_units" metadata we can ignore possibility of such corner cases and submit this change. If such scenario is found, we can handle it later.

PR #7664, opened 2023-08-10
driver plug-in should support non-DRM based device node

maxzhen:
In the future, the driver plug-in will need to support:
- user driver which is under Accel (vs DRM) framework, hence has a different device node path
- mgmt driver which is not called xclmgmt, hence has a different device node path

This pull request makes changes in driver plug-in to make it more flexible for the device node path / name.

### DIFF
--- a/src/runtime_src/core/pcie/linux/pcidev.cpp
+++ b/src/runtime_src/core/pcie/linux/pcidev.cpp
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright (C) 2016-2020, 2023 Xilinx, Inc
+// Copyright (C) 2016-2020 Xilinx, Inc
 // Copyright (C) 2022-2023 Advanced Micro Devices, Inc. All rights reserved.
 #include "pcidev.h"
 #include "pcidrv.h"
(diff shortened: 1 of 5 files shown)


src/runtime_src/core/pcie/linux/pcidev.cpp
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright (C) 2016-2020 Xilinx, Inc
-// Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
+// Copyright (C) 2016-2020, 2023 Xilinx, Inc
```
stsoe, line 2: I don't usually update Xilinx Inc on copyright these days.  Are we supposed to?

stsoe (approved):

dayeh-xilinx:
retest this please.

src/runtime_src/core/pcie/linux/pcidev.cpp
maxzhen: Oh, no. It's automatically done by my vim script. I need to fix it. We should update the AMD one.

keryell: That is a great idea to have such a script anyway!

PR #7665, opened 2023-08-11
Add support for LSP for use by emacs/eglot

sonals:
Language server protocol configuration for clangd and pyright for use with IDEs like emacs/eglot, etc.

NA

Add configuration files for the LSPs so they can resolve all symbols, includes, etc. and give code suggestions.

None

Tested code completion and code analysis with eglot built into Emacs 29.1

### DIFF
--- a/.clangd
+++ b/.clangd
@@ -1,0 +1,2 @@
+CompileFlags:
+  CompilationDatabase: build/Debug
--- a/pyrightconfig.json
+++ b/pyrightconfig.json
@@ -1,0 +1,7 @@
+{
+  "extraPaths": [
+      "src/python",
+      "src/python/pybind11/src",
+      "build/Debug/opt/xilinx/xrt/python"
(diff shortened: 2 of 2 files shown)


sonals:
> 

I recently started using [eglot](https://www.gnu.org/software/emacs/manual/html_node/eglot/Using-Eglot.html) in Emacs and I am impressed with it. It unobtrusively annotates your code with suggestions (like a built-in clang-tidy), completions, documentation, go to definition, find uses of, etc. It scales well even for a huge code base like XRT!

keryell:
After using Clang-based `rtags` for years, I compared `eglot` and `lsp-mode` 3 years ago and I ended up picking `lsp-mode` but perhaps I could look again at `eglot`.

dayeh-xilinx:
retest this please.

stsoe (approved):

PR #7676, opened 2023-08-24
VITIS-9039 Add an IPU specific test to validate

AShivangi:
Add a "Hello World" test for ipu

Added a new test for IPU which will be called if the device name is IPU in "verify". With this change, the user can run `xbutil verify --run quick` without any check failing. 

On windows:
```
 .\xbutil validate -d 00c2:00:01.1 -r quick --verbose
Verbose: Enabling Verbosity
Validate Device           : [00c2:00:01.1]
    Platform              : IPU
    SC Version            :
    Platform ID           : 0x0
Test 1 [00c2:00:01.1]     : aux-connection

    Description           : Check if auxiliary power is connected
    Details               : Aux power connector is not available on this board
Test 2 [00c2:00:01.1]     : pcie-link
    Description           : Check if PCIE link is active
Test 3 [00c2:00:01.1]     : sc-version
    Description           : Check if SC firmware is up-to-date
Test 4 [00c2:00:01.1]     : verify
    Description           : Run 'Hello World' kernel test
    Xclbin                : C:\Windows\System32\DriverStore\FileRepository\ipukmddrv.inf_amd64_f0ee9320079f8f17\1x4.xclbin
    Details               : Kernel name is 'DPU_1x4'
Validation completed
```

On Linux:
```
$ xbutil validate -d b3:00 -r ipu-verify

Validate Device           : [0000:b3:00.1]
    Platform              : xilinx_vck5000_gen4x8_qdma_base_2
    SC Version            : 4.4.35
    Platform ID           : 05DCA096-76CB-730B-8D19-EC1192FBAE3F
Verbose: Enabling Verbosity
Test 1 [0000:b3:00.1]     : ipu-verify 

    Description           : Run 'Hello World' test on IPU
    Details               : /opt/xilinx/firmware/vck5000/gen4x8-qdma/base/test/1x4.xclbin
                            not available. Skipping validation.
Validation completed
```

### DIFF
--- a/src/runtime_src/core/common/system.h
+++ b/src/runtime_src/core/common/system.h
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright (C) 2019 Xilinx, Inc
-// Copyright (C) 2022-2023 Advanced Micro Devices, Inc. All rights reserved.
+// Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
 #ifndef XRT_CORE_SYSTEM_H
 #define XRT_CORE_SYSTEM_H
 
@@ -122,12 +122,6 @@ public:
   program_plp(const device*, const std::vector<char>&, bool) const
   {
     throw std::runtime_error("plp program is not supported");
-  }
-
-  virtual std::string 
-  find_platform_path() const
-  {
-    throw std::runtime_error("find_platform_path is not supported");
   }
 }; // system
 
@@ -263,9 +257,6 @@ get_monitor_access_type();
 XRT_CORE_COMMON_EXPORT
 void
 program_plp(const device* dev, const std::vector<char> &buffer, bool force);
-
-XRT_CORE_COMMON_EXPORT
-std::string find_platform_path();
 } //xrt_core
 
 #endif /* CORE_SYSTEM_H */
--- a/src/runtime_src/core/common/system.h
+++ b/src/runtime_src/core/common/system.h
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright (C) 2019 Xilinx, Inc
-// Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
+// Copyright (C) 2022-2023 Advanced Micro Devices, Inc. All rights reserved.
 #ifndef XRT_CORE_SYSTEM_H
 #define XRT_CORE_SYSTEM_H
 
@@ -122,6 +122,12 @@ public:
   program_plp(const device*, const std::vector<char>&, bool) const
   {
     throw std::runtime_error("plp program is not supported");
+  }
+
+  virtual std::string 
+  find_platform_path() const
+  {
+    throw std::runtime_error("find_platform_path is not supported");
   }
 }; // system
 
@@ -257,6 +263,9 @@ get_monitor_access_type();
 XRT_CORE_COMMON_EXPORT
 void
 program_plp(const device* dev, const std::vector<char> &buffer, bool force);
+
+XRT_CORE_COMMON_EXPORT
+std::string find_platform_path();
 } //xrt_core
 
 #endif /* CORE_SYSTEM_H */
--- a/src/runtime_src/core/tools/common/TestRunner.cpp
+++ b/src/runtime_src/core/tools/common/TestRunner.cpp
@@ -19,6 +19,7 @@
 #include "TestRunner.h"
 #include "core/common/error.h"
 #include "core/common/query_requests.h"
+#include "core/common/module_loader.h"
 #include "core/common/system.h"
 #include "core/tools/common/Process.h"
 #include "tools/common/XBUtilities.h"
@@ -391,11 +392,6 @@ std::string
 TestRunner::findPlatformPath(const std::shared_ptr<xrt_core::device>& _dev,
                  boost::property_tree::ptree& _ptTest)
 {
-#ifdef _WIN32
-  boost::ignore_unused(_dev);
-  boost::ignore_unused(_ptTest);
-  return xrt_core::find_platform_path();
-#else
   //check if a 2RP platform
   const auto logic_uuid = xrt_core::device_query_default<xrt_core::query::logic_uuids>(_dev, {});
 
@@ -406,20 +402,33 @@ TestRunner::findPlatformPath(const std::shared_ptr<xrt_core::device>& _dev,
     auto name = xrt_core::device_query<xrt_core::query::rom_vbnv>(_dev);
     return searchLegacyXclbin(vendor, name, _ptTest);
   }
-#endif
 }
 
 std::string
 TestRunner::findXclbinPath( const std::shared_ptr<xrt_core::device>& _dev,
                 boost::property_tree::ptree& _ptTest)
 {
+  auto xclbin_name = _ptTest.get<std::string>("xclbin", "");
+  std::string xclbin_path;
+#ifdef _WIN32
+  boost::ignore_unused(_dev);
+  try {
+    xclbin_path = xrt_core::environment::xclbin_path(xclbin_name);
+  }
+  catch(const std::exception) {
+    const auto fmt = boost::format("%s not available. Skipping validation.") % xclbin_name;
+    logger(_ptTest, "Details", boost::str(fmt));
+    _ptTest.put("status", test_token_skipped);
+  }
+#else
   const auto platform_path = findPlatformPath(_dev, _ptTest);
-  const auto xclbin_path = _ptTest.get<std::string>("xclbin_directory", platform_path) + _ptTest.get<std::string>("xclbin", "");
+  xclbin_path = _ptTest.get<std::string>("xclbin_directory", platform_path) + xclbin_name;
   if (!boost::filesystem::exists(xclbin_path)) {
     const auto fmt = boost::format("%s not available. Skipping validation.") % xclbin_path;
     logger(_ptTest, "Details", boost::str(fmt));
     _ptTest.put("status", test_token_skipped);
   }
+#endif
   return xclbin_path;
 }
 
--- a/src/runtime_src/core/tools/common/TestRunner.cpp
+++ b/src/runtime_src/core/tools/common/TestRunner.cpp
@@ -410,7 +410,7 @@ TestRunner::findXclbinPath( const std::shared_ptr<xrt_core::device>& _dev,
   auto xclbin_name = _ptTest.get<std::string>("xclbin", "");
   std::string xclbin_path;
 #ifdef _WIN32
-  std::ignore(_dev);
+  boost::ignore_unused(_dev);
   try {
     xclbin_path = xrt_core::environment::xclbin_path(xclbin_name);
   }
--- a/src/runtime_src/core/tools/common/TestRunner.cpp
+++ b/src/runtime_src/core/tools/common/TestRunner.cpp
@@ -32,6 +32,10 @@ namespace XBU = XBUtilities;
 // System - Include Files
 #include <iostream>
 #include <regex>
+#ifdef _WIN32
+#include <cfgmgr32.h> // CM_Get_Device_ID_List_Size
+#endif
+
 
 // ------ L O C A L   F U N C T I O N S ---------------------------------------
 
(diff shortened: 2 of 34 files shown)


src/runtime_src/core/tools/common/tests/TestIPU.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+  xrt::xclbin xclbin{xclbin_path};
+  if(!xclbin) { // if xclbin not found
+    return ptree;
+  }
+
+  // Determine The DPU Kernel Name
+  auto xkernels = xclbin.get_kernels();
+
+  auto xkernel = *std::find_if(xkernels.begin(), xkernels.end(), [](xrt::xclbin::kernel& k) {
```
stsoe, line 45: What if std::find_if returns end() ?

src/runtime_src/core/tools/common/tests/TestIPU.cpp
```diff
@@ -0,0 +1,84 @@
...
+  catch (const std::runtime_error& ex) {
+    logger(ptree, "Error", ex.what());
+    ptree.put("status", test_token_failed);
+    return ptree;
+  }
+
+  xrt::xclbin xclbin{xclbin_path};
+  if(!xclbin) { // if xclbin not found
+    return ptree;
+  }
```
stsoe, line 40: Line 37 throws if file not found.  Maybe include within earlier try?
```suggestion
```

src/runtime_src/core/tools/common/tests/TestIPU.h
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2023 Advanced Micro Devices, Inc. All rights reserved.
+
+#ifndef __TestIPU_h_
+#define __TestIPU_h_
+
+#include "tools/common/TestRunner.h"
+#include "xrt/xrt_bo.h"
```
stsoe, line 8: Doesn't look like bo is needed
```suggestion
#include "xrt/xrt_device.h"

#include <boost/property_tree/ptree.hpp>
#include <memory>
```

src/runtime_src/core/tools/common/tests/TestIPU.cpp
```diff
@@ -0,0 +1,84 @@
...
+    auto name = k.get_name();
+    return name.rfind("DPU",0) == 0; // Starts with "DPU"
+  });
+  auto kernelName = xkernel.get_name();
+  logger(ptree, "Details", boost::str(boost::format("Kernel name is '%s'") % kernelName));
+
+  auto working_dev = xrt::device{dev->get_device_id()};
+  working_dev.register_xclbin(xclbin);
+  xrt::hw_context hwctx{working_dev, xclbin.get_uuid()};
+  xrt::kernel kernel{hwctx, kernelName};
```
stsoe, line 55: I think this is a good way of constructing class objects, but you should at least be consistent and do the same for the xrt::bo s below and device above.

stsoe (changes requested):
Minor tweaks.

src/runtime_src/core/tools/common/tests/TestIPU.cpp
AShivangi: Done

src/runtime_src/core/tools/common/tests/TestIPU.cpp
AShivangi: Fixed

src/runtime_src/core/tools/common/tests/TestIPU.cpp
AShivangi: Good point! Fixed

maxzhen:
Not sure how findXclbinPath() works on Linux? On Linux, where should we put the ipu xclbin?

AShivangi:
> Not sure how findXclbinPath() works on Linux? On Linux, where should we put the ipu xclbin?

From what I understand ipu would have a platform package similar to SSV2 platforms, so findXclbinPath() will go find the ipu platform package and look for the xclbin in it. If that's not the case, we can definitely add some code in this function to search for the xclbin wherever it is placed.

maxzhen:
> > Not sure how findXclbinPath() works on Linux? On Linux, where should we put the ipu xclbin?
> 
> From what I understand ipu would have a platform package similar to SSV2 platforms, so findXclbinPath() will go find the ipu platform package and look for the xclbin in it. If that's not the case, we can definitely add some code in this function to search for the xclbin wherever it is placed.

We may not be able to add some code specific to Linux since this is common code across all OSes. Some kind of SHIM layer API needs to be introduced here if we'd like to do it that way. Let's assume that there will be a platform package on Linux...

AShivangi:
> > > Not sure how findXclbinPath() works on Linux? On Linux, where should we put the ipu xclbin?
> > 
> > 
> > From what I understand ipu would have a platform package similar to SSV2 platforms, so findXclbinPath() will go find the ipu platform package and look for the xclbin in it. If that's not the case, we can definitely add some code in this function to search for the xclbin wherever it is placed.
> 
> We may not be able to add some code specific to Linux since this is common code across all OSes. Some kind of SHIM layer API needs to be introduced here if we'd like to do it that way. Let's assume that there will be a platform package on Linux...

So the way I have the findXclbinPath() code setup right now is, if it is windows, it goes looks at the ipu driver, but if it is linux, it looks in SSV2 platforms `searchSSV2Xclbin ()` if not found, it looks in legacy platforms `searchLegacyXclbin()`. What I meant by "adding code" was that we can have another if statement which looks for the ipu path if it is different than the SSV2 platform. As of now, the testcase skips on Linux since it doesn't finds the xclbin in versal platform pkg (see description)

maxzhen:
> > > > Not sure how findXclbinPath() works on Linux? On Linux, where should we put the ipu xclbin?
> > > 
> > > 
> > > From what I understand ipu would have a platform package similar to SSV2 platforms, so findXclbinPath() will go find the ipu platform package and look for the xclbin in it. If that's not the case, we can definitely add some code in this function to search for the xclbin wherever it is placed.
> > 
> > 
> > We may not be able to add some code specific to Linux since this is common code across all OSes. Some kind of SHIM layer API needs to be introduced here if we'd like to do it that way. Let's assume that there will be a platform package on Linux...
> 
> So the way I have the findXclbinPath() code setup right now is, if it is windows, it goes looks at the ipu driver, but if it is linux, it looks in SSV2 platforms `searchSSV2Xclbin ()` if not found, it looks in legacy platforms `searchLegacyXclbin()`. What I meant by "adding code" was that we can have another if statement which looks for the ipu path if it is different than the SSV2 platform. As of now, the testcase skips on Linux since it doesn't finds the xclbin in versal platform pkg (see description)

I see that searchSSV2Xclbin() searches /opt/xilinx/firmware on Linux. We could place the xclbin there or some similar path, yes.

sonals:
We should stick with _verify_ sub-command? Why call it _ipu-verify_?

src/runtime_src/core/common/system.h
```diff
@@ -257,6 +263,9 @@ get_monitor_access_type();
 XRT_CORE_COMMON_EXPORT
 void
 program_plp(const device* dev, const std::vector<char> &buffer, bool force);
+
+XRT_CORE_COMMON_EXPORT
+std::string find_platform_path();
```
dbenusov-xilinx, line 268: This seems device specific not system specific.

src/runtime_src/core/tools/common/TestRunner.cpp
```diff
@@ -391,6 +391,11 @@ std::string
 TestRunner::findPlatformPath(const std::shared_ptr<xrt_core::device>& _dev,
                  boost::property_tree::ptree& _ptTest)
 {
+#ifdef _WIN32
+  boost::ignore_unused(_dev);
+  boost::ignore_unused(_ptTest);
+  return xrt_core::find_platform_path();
```
dbenusov-xilinx, line 397: We should try and not have any ifdefs in higher level code. We should reexamine how we find xclbins and their paths.

Personally, I think the platform path is something a device should define not a system. Linux will also support IPU one day which will break the current changes. Better to sort this issue out now rather than patching a hole.

src/runtime_src/core/tools/common/tests/TestIPU.cpp
```diff
@@ -0,0 +1,97 @@
...
+// ----- C L A S S   M E T H O D S -------------------------------------------
+TestIPU::TestIPU()
+  : TestRunner("ipu-verify", 
+                "Run 'Hello World' test on IPU",
+                "1x4.xclbin"){}
+
+boost::property_tree::ptree
+TestIPU::run(std::shared_ptr<xrt_core::device> dev)
+{
+  boost::property_tree::ptree ptree = get_test_header();
```
dbenusov-xilinx, line 27: Why is this a built-in test rather than a precompiled executable? I believe we should move towards all tests being executables. It will allow the `validate` command to be much simpler and encourage other parts of AMD to adopt our tool (eventually).

dbenusov-xilinx (changes requested):

src/runtime_src/core/tools/common/tests/TestIPU.cpp
```diff
@@ -0,0 +1,97 @@
...
+namespace XBU = XBUtilities;
+
+#include <boost/filesystem.hpp>
+
+static constexpr size_t host_app = 1; //opcode
+static constexpr size_t buffer_size = 128;
+
+// ----- C L A S S   M E T H O D S -------------------------------------------
+TestIPU::TestIPU()
+  : TestRunner("ipu-verify",
```
dbenusov-xilinx, line 20:
```suggestion
  : TestRunner("verify-ipu", 
```

dbenusov-xilinx, line 20: Is this ipu verify or AIE verify? Is the 1x4 xclbin usable on any other platforms?

AShivangi:
> We should stick with _verify_ sub-command? Why call it _ipu-verify_?

I've made changes to only have verify test. Based on the device, xbutil will run basic Hello World or IPU hello world

manikandan-xilinx:
retest this please. (gitenterprise issue is fixed)

src/runtime_src/core/tools/common/tests/TestIPU.cpp
AShivangi: discussed offline

src/runtime_src/core/tools/common/TestRunner.cpp
```diff
@@ -407,13 +407,27 @@ std::string
 TestRunner::findXclbinPath( const std::shared_ptr<xrt_core::device>& _dev,
                 boost::property_tree::ptree& _ptTest)
 {
+  auto xclbin_name = _ptTest.get<std::string>("xclbin", "");
+  std::string xclbin_path;
+#ifdef _WIN32
+  boost::ignore_unused(_dev);
+  try {
+    xclbin_path = xrt_core::environment::xclbin_path(xclbin_name);
```
dbenusov-xilinx, line 415: Instead of using the ifdefs here could you just use the try catch? The catch can do the legacy method of getting the xclbin path.

src/runtime_src/core/tools/common/tests/TestIPU.cpp
dbenusov-xilinx: I like the idea of writing the tests in the code more now that I thought about it.

src/runtime_src/core/tools/common/tests/TestVerify.cpp
```diff
@@ -16,7 +17,15 @@ TestVerify::TestVerify()
 boost::property_tree::ptree
 TestVerify::run(std::shared_ptr<xrt_core::device> dev)
 {
-  boost::property_tree::ptree ptree = get_test_header();
-  runTestCase(dev, "22_verify.py", ptree);
+  auto device_name = xrt_core::device_query_default<xrt_core::query::rom_vbnv>(dev, "");
+  boost::property_tree::ptree ptree;
+  if(device_name.find("IPU") != std::string::npos) {
```
dbenusov-xilinx, line 22: You should definitely comment that this is a temporary hack! At least I hope it is going to be... 
The idea makes sense to me though. I do see a lot of problems with scaling this solution, especially as we move into multiple device types that would all use the IPU test.

src/runtime_src/core/tools/common/tests/TestIPU.cpp
dbenusov-xilinx: Actually I am still concerned about the name of this test. Have you talked to PRT about the name?

dbenusov-xilinx (changes requested):

src/runtime_src/core/tools/common/TestRunner.cpp
```diff
@@ -407,13 +407,27 @@ std::string
 TestRunner::findXclbinPath( const std::shared_ptr<xrt_core::device>& _dev,
                 boost::property_tree::ptree& _ptTest)
 {
+  auto xclbin_name = _ptTest.get<std::string>("xclbin", "");
+  std::string xclbin_path;
+#ifdef _WIN32
+  boost::ignore_unused(_dev);
```
keryell, line 413: It has been in C++ for a decade
```suggestion
  std::ignore = _dev;
```

src/runtime_src/core/tools/common/tests/TestVerify.cpp
keryell, line 22:
```suggestion
  if (device_name.find("IPU") != std::string::npos) {
```
@stsoe I realize there is no `.clang-format` in this project.
Start with a top-level minimalistic one everybody can agree on and add later some more precise per sub-project?

src/runtime_src/core/tools/common/tests/TestVerify.cpp
```diff
@@ -16,7 +17,15 @@ TestVerify::TestVerify()
...
+    //run ipu verify
+    ptree = std::make_shared<TestIPU>()->run(dev);
```
keryell, line 24: Curious... I have the feeling that once the full expression is evaluated the shared pointer is removed.
If this is the correct expected behavior why writing such a code using dynamic memory and not just
```suggestion
    ptree = TestIPU {}.run(dev);
```
?

src/runtime_src/core/tools/common/tests/TestVerify.cpp
AShivangi: I added a to-do tag. We can tackle this as a part of tests refactoring story

src/runtime_src/core/tools/common/TestRunner.cpp
AShivangi: fixed

src/runtime_src/core/tools/common/tests/TestVerify.cpp
AShivangi: fixed

src/runtime_src/core/tools/common/tests/TestVerify.cpp
AShivangi: Thanks for the suggestion. The shared pointer goes out of scope once the test is run. I've updated the code to what you suggested

manikandan-xilinx:
retest this please.

manikandan-xilinx:
retest this please.

PR #7679, opened 2023-08-28
Use driver store path for xilinx_xrt on  Windows

stsoe:
Implement xrt_core::environment::xilinx_xrt() on Windows to use driver store path if available.  On Linux keep using existing defaults.

On Windows the implementation is:
- If XILINX_XRT is set in environment, then return the value of XILINX_XRT
- Else compute as driver store path if available (new)
- Else return same directory as xrt_coreutil.dll is loaded from

On Linux the implementation is
- If XILINX_XRT is set in environment, then return the value of XILINX_XRT
- Else use /opt/xilinx/xrt (/usr on edge)
- TBD: Else return same directory as libxrt_coreutil.so is loaded from

Add internal API for obtaining absolute path of xclbin file. For Windows, the xclbin path will be the same as xilinx_xrt().

The code for xilinx_xrt() and xclbin_path() is conditional based on platform.  xilinx_xrt() is used to load the xrt_core shim library,

### DIFF
--- a/src/runtime_src/core/common/detail/linux/xilinx_xrt.h
+++ b/src/runtime_src/core/common/detail/linux/xilinx_xrt.h
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2023 Advanced Micro Devices, Inc. All rights reserved.
+#include <boost/filesystem/path.hpp>
+
+namespace xrt_core::detail {
+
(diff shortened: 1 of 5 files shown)


manikandan-xilinx:
retest this please. (windows agents is fixed now)

manikandan-xilinx:
retest this please. (petalinux issue is fixed)

PR #7687, opened 2023-09-05
[CR-1161728]: emconfig.json path issue if host code is in python.

sgundime-xilinx:
Fix:  emconfig.json file path is wrongly deduced with getSelfPath API especially if host code is python script. The current fix ensures the proper current path is chosen if the exe name contains /usr/*/python in it.

Risk: very low

Tests: Canary Run. Results are Clean.

Reviewer: venkatp

### DIFF
--- a/src/runtime_src/core/pcie/emulation/common_em/config.cxx
+++ b/src/runtime_src/core/pcie/emulation/common_em/config.cxx
@@ -102,7 +102,7 @@ namespace xclemulation{
     return defaultValue;
   }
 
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx (changes requested):
Hi @sgundime-xilinx ,
It is better to check the emconfig.json file in the executable path first. If we dont find it there, check in current working directory. We shouldn't be adding special checks for python etc.

dayeh-xilinx:
retest this please.

chvamshi-xilinx (approved):

venkatp-xilinx (approved):

PR #7689, opened 2023-09-05
CR-1173506 Segmentation fault when running VAI application

stsoe:
Make sure xrt::bo shares the core device used by xrt::device even when xrt::bo is created from deprecated xclDeviceHandle.

Creating an xrt::device object under the hood calls xclOpen which creates a shim object. The shim object contructs a shared xrt_core::device object and associates the shim handle (xclDeviceHandle or void*) with a weak ptr to this xrt_core::device.

Because legacy xcl applications explicitly calls xclOpen and xclClose, the xrt_core::device object created by shim must not itself call xclClose.

When an application uses XRT native C++ APIs to create an xrt::device object, this device object calls xclOpen and must therefore also call xclClose.  The call to xclClose is done by rewrapping the shared xrt_core::device object created by xclOpen into a shared pointer with a deleter that calls xclClose.

Because of the legacy xcl support, some deprecated APIs use xclDeviceHandle during construction, but shares ownership with the underlying xrt_core::device object.  The shared ownership is through the association between shim handle (xclDeviceHandle) and xrt_core::device weak ptr.

The problem in CR-1173506 was that the shared ownership established by deprecated xrt::bo constructor is with the shared pointer that does *not* call xclClose and if xrt::device is going out of scope first, then xclClose has been called before the xrt::bo is freed and this results in segmentation fault.  Assumption is that xrt::device is constructed before first deprecated API is used with xclDeviceHandle.  This is a safe assumption because everything starts with a device object.

This PR fixes the cached xrt_core::device weak pointer to be that of the rewrapped pointer created by xrt::device.

Standard old suite of OpenCL testcases.

### DIFF
--- a/src/runtime_src/core/common/system.cpp
+++ b/src/runtime_src/core/common/system.cpp
@@ -131,7 +131,14 @@ get_userpf_device(device::id_type id)
   // but leaves device object alone. The returned device is managed in that
   // it calls xclClose when going out of scope.
   auto close = [] (xrt_core::device* d) { d->close_device(); };
-  return {device.get(), close};
+  std::shared_ptr<xrt_core::device> ptr{device.get(), close};
+
+  // The repackage raw ptr is the one that should be cached so
+  // so that all references to device handles in application code
+  // are tied to the shared ptr that ends up calling xclClose
+  std::lock_guard lk(mutex);
+  userpf_device_map[device->get_device_handle()] = ptr;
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx (approved):

chvamshi-xilinx:
Thanks Soren. Your description in PR very good.

PR #7696, opened 2023-09-06
Add xrt::version APIs

stsoe:
Ability to check XRT version and build at run-time.


src/runtime_src/core/common/api/xrt_version.cpp
```diff
@@ -0,0 +1,76 @@
...
+unsigned int
+build()
+{
+  return XRT_BUILD;
+}
+
+} // namespace xrt::version
+
+////////////////////////////////////////////////////////////////
+// xrt_message C API implmentations (xrt_version.h)
```
keryell, line 46:
```suggestion
// xrt_version C API implementations (xrt_version.h)
// with "C" linkage.
```

stsoe: Duh.

stsoe:
retest this please - seems to build correctly when using build_edge -aarch versal locally

xilinxgitops:
retest this please - debugging issue with Vamshi

keryell:
I am discovering this `sys/sysmacros.h` nightmare. :-(
Is `sys/types.h` required by the XRT API? Perhaps this API enabling file-descriptor sharing?
Could this plague contained in only the internal `.c`/`.cpp` files using it?

stsoe:
> Is `sys/types.h` required by the XRT API? Perhaps this API enabling file-descriptor sharing?
> Could this plague contained in only the internal `.c`/`.cpp` files using it?

I admit I didn't track down where this include came from, but xrt_version.cpp does not include (implicitly) anything but `<string>` and `<iostream>` .   

I suspect iostream is the culprit, and yes, I don't really fancy `iostream` being included especially not in a header file, but that's unfortunately how the version.h file generated by CMake was coded up.  Let me revisit that, and cleanup src/CMake/config/version.h.in and move the definition of the print function into the xrt_version.cpp.  While a proper change, this doesn't actually solve the major minor macro mess.

manikandan-xilinx:
retest this please. (Including hidden files during checkout)

keryell:
Curious.
I cannot imagine C++ `<iostream>` or anything else pulling anything from `sys/` defining some atrocious macros like that.

stsoe:
Well I don't know, but as I said, the macro mess is isolated to `xrt_version.cpp` which includes `gen/version.h` which in turn includes `<string>` and `<iostream>`

PR #7701, opened 2023-09-08
CR-1114740 xbutil examine -d <BDF> -r 'aieshim' console output does n…

saumyag-xilinx:
…ot match with JSON output

fix

https://jira.xilinx.com/browse/CR-1114740
AIE information is not available for alveo and versal data center platforms, so updated the same in code flow.
low
Verified aie reports on v70 and u50.
NA


chvamshi-xilinx:
Hi @saumyag-xilinx ,
I think, you get "no such key" error in json output. We shouldnt show this error to user. We should report that "no aie information available"

saumyag-xilinx:
> Hi @saumyag-xilinx , I think, you get "no such key" error in json output. We shouldnt show this error to user. We should report that "no aie information available"

sure, made the change

chvamshi-xilinx (approved):
This is not ideal, but this is what we need to do it for DC cases

chvamshi-xilinx:
retest this please

src/runtime_src/core/common/asd_parser.cpp
```diff
@@ -728,55 +728,56 @@ get_formated_tiles_info(const xrt_core::device* device, aie_tile_type tile_type,
...
+        break;
+      case aie_tile_type::mem :
+        aie_status = parse_data_from_buf<aie_mem_tile_status>(tiles_status.buf, info, cols_filled);
+        pt = format_aie_info<aie_mem_tile_status>(aie_status, info.cols, info, cols_filled);
+        break;
+      default :
+        throw std::runtime_error("Unknown tile type in formatting Aie tiles status info");
+    }
+  }
+  catch (const std::exception&) {
```
chvamshi-xilinx, line 779: As discussed offline, fill the property tree here with proper message.

chvamshi-xilinx:
Hi @saumyag-xilinx , Please make the suggested changes in next PR as this is required for XDP team.

PR #7702, opened 2023-09-08
CR-1175649 Always wait for command completion through hwqueue in SHIM layer

stsoe:
Today, XRT coreutil can directly check the exec buf state to know if a command is completed without calling into SHIM layer. In the command submission path, however, XRT coreutil will always call into SHIM layer to issue the cmd to HW. There is a little bit asymmetric in this architecture, which makes it harder for SHIM to abstract device specific behavior and hide it from coreutil. It gets leaked to coreutil through the completion code path.

One particular example is that it is difficult for SHIM to insert static probes in both submission and completion code path to do performance profiling since SHIM may not be involved in completion path. And this type of probe is specific to some platforms due to the fact that the driver is not involved during command handling at all.

Minor tweak to call hwqueue_handle::wait() unconditionally from coreutil.  On success,  follow up with check that command state is as expected.  

When this CR is fixed, the workaround implemented in CR-1167065 (#7608 and #7667) can be reverted since CERT will not directly write to the exec buf state field anymore. SHIM will hide it from coreutil. This revert will be handled separately from this PR.

### DIFF
--- a/src/runtime_src/core/common/api/hw_queue.cpp
+++ b/src/runtime_src/core/common/api/hw_queue.cpp
@@ -412,12 +412,17 @@ public:
   std::cv_status
   wait(const xrt_core::command* cmd, size_t timeout_ms) override
   {
+    // Dispatch wait to shim hwqueue_handle rather than accessing
+    // pkt state directly.  This is done to allow shim direct control
+    // over command completion and command state.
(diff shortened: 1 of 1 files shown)


manikandan-xilinx:
retest this please.

maxzhen (approved):

PR #7703, opened 2023-09-08
CR-1174069: Catch exceptions when debug/profiling plugins fail to be loaded

jvillarre:
Users can enable profiling functionality through xrt.ini options, and during the application run different plugin modules will be loaded based on what is requested.  If an option like aie_status is requested on platforms that do not have the corresponding plugin then an exception is thrown when the library is not found.  

If the exception is thrown when running "xbutil examine," then the exception is caught but from the user's perspective it is not explained and the output of xbutil is incorrect.  The issue was discovered by running xbutil on a machine that did not have AIE and tried to load the aie_status debug module.

Added try...catch blocks during the loading of the individual debug/profiling modules.  If the exception is caught at this level, a warning message is issued to the user and the execution continues.

Low risks as the exceptions that were already being generated now only reports a warning instead of throwing it higher.

The original failing test case where the application tries to load a plugin that does not exist on that platform has been tested.

No documentation changes.

### DIFF
--- a/src/runtime_src/core/pcie/linux/plugin/xdp/plugin_loader.cpp
+++ b/src/runtime_src/core/pcie/linux/plugin/xdp/plugin_loader.cpp
@@ -1,5 +1,6 @@
 /**
  * Copyright (C) 2020-2022 Xilinx, Inc
+ * Copyright (C) 2023 Advanced Micro Devices, Inc. - All rights reserved
  *
  * Licensed under the Apache License, Version 2.0 (the "License"). You may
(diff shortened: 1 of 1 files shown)


manikandan-xilinx:
retest this please.

pgschuey (approved):

manikandan-xilinx:
retest this please.

hackwa (approved):

IshitaGhosh (approved):
Changes look good.

PR #7704, opened 2023-09-11
add interface uuid into xgq_vmr host driver for linux upstream project

xdavidz:

For linux upstreaming effort of versal host driver, we cannot use device tree fdt source code. 
Even though we have the fdt metadata sent back to host, we cannot use any fdt source code to decode it.

Luckly, the VMR has already decoded it in the firmware.

Thus, we can add this additional sub-opcode which can retrieve current interface uuid from current shell.
This will benefit the mailbox request from xocl to compare it with incoming xclbin.

We would like to bring up xgq_vmr.c up-to-date with every ops that we support, that's why we also need this code in xgq_vmr.c.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
@@ -177,6 +177,8 @@ struct xocl_xgq_vmr {
 	size_t			xgq_vmr_system_dtb_size;
 	char			*xgq_vmr_plm_log;
 	size_t			xgq_vmr_plm_log_size;
+	char			*xgq_vmr_shell_int_uuid;
+	size_t			xgq_vmr_shell_int_uuid_size;
 	u16			pwr_scaling_threshold_limit;
 	u8			temp_scaling_threshold_limit;
 	u16			pwr_scaling_limit;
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
```diff
@@ -1200,6 +1202,15 @@ static int xgq_refresh_system_dtb(struct xocl_xgq_vmr *xgq)
 		&xgq->xgq_vmr_system_dtb_size, XGQ_CMD_LOG_SYSTEM_DTB, 0, 0);
 }
 
+static int xgq_refresh_shell_int_uuid(struct xocl_xgq_vmr *xgq)
```
maxzhen, line 1205: when do you call this function to reload interface uuid?

xdavidz: We can add a sysfs node to refresh this info.

maxzhen: > We can add a sysfs node to refresh this info.

This does not make sense to me. The refresh should only be called when the shell has changed. Driver knows exactly when the shell can change, so driver should not rely on any other triggers to refresh. We can call it in probe, and after hot reset?

xdavidz: ok, we will add this check in the probe, it is good for both cold reboot and hot reset.

xdavidz:
any comments from @NishadSaraf  and @prapulkrishnamurthy ?

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
xdavidz: fixed

prapulkrishnamurthy (approved):

maxzhen (approved):

NishadSaraf (approved):

PR #7712, opened 2023-09-19
Add platform specific repository for xclbins

stsoe:
For some use flows xclbins are not provided by the application but instead dynamically loaded from a repository created at sw stack install time.  This PR adds support for repository of xclbins though the xrt::xclbin_repository first class object.

Support repository of xclbins, where repository path(s) are one of
- explicit directory path
- platform specific directory where xclbins are installed
- configuration option via xrt.ini

The repository can be iterated for all xclbins is available in the repo or it can be asked to load a specific xclbin.
The built-in repository path is platform specific. 

The PR also changes the behavior of xrt::xclbin constructed from std::string.  The constructor now checks for specified xclbin either as (1) absolute path, (2) relative path to current directory, (3) relative path under user configured repository, or (4) relative path under built-in repository.

- Standard OpenCL tests on Alveo.
- Hand picked windows tests on MCDM.
- Debugging.

See tests/xrt/56_xclbin/main.cpp for sample use.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -3,6 +3,7 @@
 set -e
 
 OSDIST=`grep '^ID=' /etc/os-release | awk -F= '{print $2}' | tr -d '"'`
+VERSION=`grep '^VERSION_ID=' /etc/os-release | awk -F= '{print $2}' | tr -d '"'`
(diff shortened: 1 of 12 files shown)


manikandan-xilinx:
retest this please. (We tried different backend for pcie/hw tests. looks like there was an issue so retest should work)

PR #7714, opened 2023-09-20
Added AIE Profiling on additional Windows Devices

nishraptor:
Added AIE Profiling support on IPU devices. 

A new AIE Profile plugin is compiled only on windows devices. The plugin configure/poll calls are now within the xrt hardware context creation/destruction in order to ensure that the XRT kernels are valid and that the AIE is in a valid state for reading counters after the application has completed. AIE configuration occurs when the hardware context is created, and the counters are profiled once when the hardware context is being destroyed. The raw hw_context_impl void* is passed to the aie_profile callback, which then uses the new method defined in hw_context_int in order to construct a xrt::hw_context from the implementation shared pointer.

Could have effects on V70/VCK5000 or VCK190 AIE profiling.

Built and verified on windows machines


src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -176,4 +202,4 @@ operator xrt_core::hwctx_handle* () const
   return get_handle()->get_hwctx_handle();
 }
 
-} // xrt
+} // xrt
```
stsoe, line 205: missing new line end of file

src/runtime_src/core/common/xdp/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+  PRIVATE
+  ${XRT_SOURCE_DIR}/runtime_src
+  )
+
+if (XDP_MINIMAL_BUILD STREQUAL "yes")
+  target_compile_definitions(core_common_xdp_profile_objects
+    PRIVATE 
+    XDP_MINIMAL_BUILD=1
+    )
+endif()
```
stsoe, line 17: missing new line

src/runtime_src/core/common/api/hw_context_int.h
```diff
@@ -28,6 +30,14 @@ get_core_device_raw(const xrt::hw_context& ctx);
 void
 set_exclusive(xrt::hw_context& ctx);
 
+// Allows the creation of the hardware context from a void pointer
+// to the hardware context implementation. This is used by the
+// XDP plugins in order to initialize when the user creates
+// a hardware context in their host
```
stsoe, line 36: Let's extend this comment to explain why the pointer is a void* rather than a forward declared xrt::hw_context_impl pointer.

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -48,6 +49,17 @@ class hw_context_impl
...
 
+  std::shared_ptr<hw_context_impl>
+  get_shared_ptr()
+  {
+    return shared_from_this();
+  }
+
+  ~hw_context_impl()
+  {
+    xrt_core::xdp::end_poll(this);
```
stsoe, line 60: Comment the asymmetry here. Explain that xrt::hw_context constructor started the polling and why polling was not started from the hw_context_impl constructor (shared ptr not available).   It isn't obvious, so it deserves a comment.

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -127,13 +149,17 @@ namespace xrt {
 hw_context::
 hw_context(const xrt::device& device, const xrt::uuid& xclbin_id, const xrt::hw_context::cfg_param_type& cfg_param)
   : detail::pimpl<hw_context_impl>(std::make_shared<hw_context_impl>(device.get_handle(), xclbin_id, cfg_param))
-{}
+{
+  xrt_core::xdp::update_device(get_handle().get());
```
stsoe, line 153: I am little confused here.  Do we even need shared_from_this, why isn't update_device called with the shared pointer if you need to participate in ownership.  The raw pointer passed here is converted to a shared ptr immediately or later?  Not clear to me.

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -127,13 +149,17 @@ namespace xrt {
...
+}
 
 
 hw_context::
 hw_context(const xrt::device& device, const xrt::uuid& xclbin_id, access_mode mode)
   : detail::pimpl<hw_context_impl>(std::make_shared<hw_context_impl>(device.get_handle(), xclbin_id, mode))
-{}
+{
+  xrt_core::xdp::update_device(get_handle().get());
```
stsoe, line 161: Same comment as above

src/runtime_src/core/common/xdp/profile.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "core/common/module_loader.h"
+
+#include <functional>
+// This file makes the connections between all xrt_coreutil level hooks
+// to the corresponding xdp plugins.  It is responsible for loading all of
+// modules.
+
+namespace xrt_core::xdp::aie::profile {
+
+  void
```
stsoe, line 17: Do not indent within namespace

src/runtime_src/core/common/xdp/profile.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+namespace xrt_core::xdp {
+
+void 
+update_device(void* handle)
+{
+  if (xrt_core::config::get_aie_profile()) {
+    try {
+      xrt_core::xdp::aie::profile::load();
+    } catch (...) {
```
stsoe, line 69:
```suggestion
    } 
    catch (...) {
```

src/runtime_src/core/common/xdp/profile.cpp
```diff
@@ -0,0 +1,83 @@
...
+}
+
+void 
+end_poll(void* handle)
+{
+  if (xrt_core::config::get_aie_profile())
+    xrt_core::xdp::aie::profile::end_poll(handle);
+}
+
+} // end namespace xrt_core::xdp
```
stsoe, line 83: Missing new line

src/runtime_src/core/common/xdp/profile.cpp
```diff
@@ -0,0 +1,83 @@
...
+  void 
+  load()
+  {
+    static xrt_core::module_loader xdp_aie_loader("xdp_aie_profile_plugin",
+                                                  register_callbacks,
+                                                  warning_callbacks);
+  }
+
+  std::function<void (void*)> update_device_cb = nullptr;
+  std::function<void (void*)> end_poll_cb = nullptr;
```
stsoe, line 26:
```suggestion
  std::function<void (void*)> update_device_cb;
  std::function<void (void*)> end_poll_cb;
```

src/runtime_src/core/common/xdp/profile.cpp
```diff
@@ -0,0 +1,83 @@
...
+  void 
+  warning_callbacks()
+  {
+  }
+
+  // Make connections
+  void 
+  update_device(void* handle)
+  {
+    if (update_device_cb != nullptr)
```
stsoe, line 48:
```suggestion
    if (update_device_cb)
```

src/runtime_src/core/common/xdp/profile.cpp
```diff
@@ -0,0 +1,83 @@
...
+      update_device_cb(handle);
+  }
+
+  void 
+  end_poll(void* handle)
+  {
+    if (end_poll_cb != nullptr)
```
stsoe, line 55:
```suggestion
    if (end_poll_cb)
```

src/runtime_src/core/common/xdp/profile.h
```diff
@@ -0,0 +1,37 @@
...
+namespace xrt_core::xdp::aie::profile {
+
+void load();
+void register_callbacks(void* handle);
+void warning_callbacks();
+
+void update_device(void* handle);
+void end_poll(void* handle);
+
+} // end namespace xrt_core::xdp::aie::profile
```
stsoe, line 35: Do these declarations need to be in header file?   I think the implementation is in profile.cpp and the functions are called from profile.cpp only.

stsoe (changes requested):
Some comments stem from misunderstanding, but maybe that can be helped with some more direct inline code comments?

src/runtime_src/core/common/api/hw_context_int.h
nishraptor: Added further explanation

src/runtime_src/core/common/api/xrt_hw_context.cpp
nishraptor: Added explanation comment

stsoe (approved):

jvillarre (approved):

PR #7715, opened 2023-09-20
CR-1156386 Default and Backup FPT sysfs nodes impl

prapulkrishnamurthy:
This is an implementation of sysfs nodes for debug purposes. Sysfs nodes for reading raw meta data from Default FPT and Backup FPT through VMR. 

Added changes in both XRT and corresponding VMR compatible changes. 

Tested on V70 base 3 2023.2 and found the hexdump to be matching between default and backup dumps 
Test results are attached in the VMR PR here:- https://github.com/Xilinx/VMR/pull/335


src/runtime_src/core/include/xgq_cmd_vmr.h
```diff
@@ -116,17 +116,19 @@ enum xgq_cmd_vmr_control_type {
...
+	XGQ_CMD_LOG_INFO		 = 0x2,
+	XGQ_CMD_LOG_AF_CLEAR		 = 0x3,
+	XGQ_CMD_LOG_ENDPOINT		 = 0x4,
+	XGQ_CMD_LOG_TASK_STATS  	 = 0x5,
+	XGQ_CMD_LOG_MEM_STATS		 = 0x6,
+	XGQ_CMD_LOG_SYSTEM_DTB		 = 0x7,
+	XGQ_CMD_LOG_PLM_LOG		 = 0x8,
+	XGQ_CMD_LOG_APU_LOG		 = 0x9,
+	XGQ_CMD_LOG_SHELL_INTERFACE_UUID = 0xa,
+	XGQ_CMD_LOG_DEFAULT_FPT		 = 0xb,
```
prapulkrishnamurthy, line 130: alignment changes made to keep consistency with vmr common header file

xdavidz (changes requested):
This looks ok, but I realized that there is a potential bug in our driver, that every log_page request might return a big size back. We need to valid the size before call vmalloc.

I think just add a code to make sure the returned size should be within request size. Does this make sense?

xdavidz:
> This looks ok, but I realized that there is a potential bug in our driver, that every log_page request might return a big size back. We need to valid the size before call vmalloc.
> 
> I think just add a code to make sure the returned size should be within request size. Does this make sense?

Adjust my request a little bit. Actually, we did have checking code in the log_page request for each set of data.
But given the request off might be not zero. Thus, I don't think we should cache those data unless we have to.

Can you please take a look at the vmr_apu_log_read? We only temporary copy data to buf and free the memory. We don't cache those in the driver.

Including fpt, system dtb, fw_log, plm_log, apu_log. The shell interface uuid is special, this one can be cached.

xdavidz:
ok to test

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
```diff
@@ -3019,6 +3041,100 @@ static ssize_t program_vmr_store(struct device *dev,
...
+                goto out;
+
+        if (off + count > size)
+                count = size - off;
+        memcpy(buf, blob + off, count);
+
+        ret = count;
+out:
+        mutex_unlock(&xgq->xgq_lock);
+        return ret;
```
maxzhen, line 3125: worth creating a function for here and L3062-3078?

maxzhen (approved):

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
xdavidz: there are indeed 3+ places dup code. could prapul fix this?

xdavidz:
reset this please

xdavidz (approved):

prapulkrishnamurthy:
The commit 87af11fa45558202e95af4f3c02349cdb09a36d5 involves eliminating the need for caching sysfs read data in driver as per @xdavidz suggested.

src/runtime_src/core/include/xgq_cmd_vmr.h
xdavidz: look ok as long as the vmr changes made in vmr.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
```diff
@@ -3019,41 +3022,99 @@ static ssize_t program_vmr_store(struct device *dev,
...
+        size_t log_size = 0;
+        ssize_t ret = 0;
 
-	/*
-	 * refresh cached data only if off is 0,
-	 * which indicates that there is a new request.
-	 */
-	if (off == 0)
-		ret = xgq_refresh_system_dtb(xgq);
+	if (off > 0)
```
xdavidz, line 3034: this looks incorrect. why off cannot be bigger than 0? It might be more than 4096 per linux sysfs request for the blob data. we are also able to return data with any offset.

prapulkrishnamurthy: non zero offset means there is some date returned per read request. And this is a more like a boundary check to avoid infinite looping of this API call because after read and return, the offset seeks back to zero and automatically keeps calling this API by behavior of sysfs.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
```diff
@@ -3019,43 +3021,109 @@ static ssize_t program_vmr_store(struct device *dev,
...
-	if (off == 0)
-		ret = xgq_refresh_system_dtb(xgq);
+        /* return count should be less or equal to count */
+        ret = xgq_vmr_default_fpt(xgq, &log_buf, &log_size, off, count);
+        if (ret)
+                return (ret == -EINVAL) ? 0 : ret;;
 
-	if (ret)
-		return ret;
+        if (off  >= log_size)
```
xdavidz, line 3038: nits: extra space in between off and >=

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
```diff
@@ -3019,43 +3021,109 @@ static ssize_t program_vmr_store(struct device *dev,
...
-	/*
-	 * refresh cached data only if off is 0,
-	 * which indicates that there is a new request.
-	 */
-	if (off == 0)
-		ret = xgq_refresh_system_dtb(xgq);
+        /* return count should be less or equal to count */
+        ret = xgq_vmr_default_fpt(xgq, &log_buf, &log_size, off, count);
+        if (ret)
+                return (ret == -EINVAL) ? 0 : ret;;
```
xdavidz, line 3036: nits: I think we can return -EINVAL for any error.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
```diff
@@ -3019,43 +3021,109 @@ static ssize_t program_vmr_store(struct device *dev,
...
+                goto done;
 
-	mutex_lock(&xgq->xgq_lock);
+        if (off + count > log_size)
+                count = log_size - off;
 
-	blob = xgq->xgq_vmr_system_dtb;
-	size = xgq->xgq_vmr_system_dtb_size;
+        memcpy(buf, log_buf + off, count);
+        vfree(log_buf);
```
xdavidz, line 3045: can you please move the vfree after done:?
otherwise, we have memory leak when goto done.

NishadSaraf (approved):

xdavidz (changes requested):
overall look ok to me, those nits are optional.
But the memory leak has to be fixed in all 4 places.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/xgq_vmr.c
```diff
@@ -3019,43 +3021,109 @@ static ssize_t program_vmr_store(struct device *dev,
...
+                goto done;
+
+        if (off + count > log_size)
+                count = log_size - off;
+
+        memcpy(buf, log_buf + off, count);
+        ret = count;
+done:
+	vfree(log_buf);
+        XGQ_INFO(xgq,"Offset  = %d, count = %d, ret = %d\n",(int)off,(int)count,(int)ret);
```
xdavidz, line 3124: nits: to avoid compiling warning, the more cleaner way: offset and count is %ld, ret is %d

xdavidz (approved):

manikandan-xilinx:
retest this please.

PR #7723, opened 2023-09-27
Update loader APIs for XDP Plugins in XRT core  

IshitaGhosh:
"xrt_core::xdp::update_device" API is currently used for loading AIE Profile Plugin only. In future this can be used to load other XDP plugins too.  

Improvement

XDP plugin handler code in xrt_core is improved to extend support to future XDP Plugins

AIE profile plugin on PHX


src/runtime_src/core/common/xdp/aie_profile.h
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: Apache-2.0
```
stsoe, line 1: This used to be in .cpp file (aka an implementation detail), it is now in a header file and the functions should probably be documented.  Splitting up implementation details into several compilation units is not always a good idea unless the pieces are meant to be used in isolation. 

Your plugins could easily have been defined in profile.cpp within different namespaces.   Who is meant to include this header files besides profile.cpp?

IshitaGhosh: We are working on adding more plugins for aie_debug, aie_trace, flexml_timeline. So my thinking was splitting up implementation of loading of separate plugins into different files will improve code structure and readability. Technically they can be in different namespaces in the same file. I can revert the change.

dayeh-xilinx:
retest this please.

stsoe (approved):

jvillarre (approved):

PR #7727, opened 2023-10-02
Removing unnecessary edge-specific include in profiling

jvillarre:
A profiling file that is common to all implementations was erroneously including a file from the core/edge section.  As the functionality should be common there should be no edge specific code used in this file.

This code was recently refactored into a general solution from code that originated in the edge-specific profiling implementation, and the include file was accidentally never removed.

Low as no functionality changed.

Compilation on x86 and edge has been tested.

None.


stsoe (approved):
Thanks!

jvillarre:
retest this please

stsoe:
retest this please - u200

PR #7728, opened 2023-10-02
Add support for trace event logging at API level

stsoe:
Log calls to APIs at begin and end using native platform logging infrastructure.  On windows event tracing is using WPP software tracing.

This PR adds a base trace class that is instantiated into platform specific objects. The windows platform specific implementation is added in a detail directory.

Tracing is integrated into the existing native profiling XDP infrastructure but is enabled using Runtime ini or env variable.

xrt.ini:
[Runtime]
trace_logging = true

or with environment variable XRT_TRACE_LOGGING_ENABLE defined.

Manual testing on windows

### DIFF
--- a/src/runtime_src/core/common/CMakeLists.txt
+++ b/src/runtime_src/core/common/CMakeLists.txt
@@ -23,6 +23,7 @@ add_library(core_common_library_objects OBJECT
   system.cpp
   thread.cpp
   time.cpp
+  trace.cpp
   utils.cpp
   xclbin_parser.cpp
   xclbin_swemu.cpp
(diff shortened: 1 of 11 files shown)


stsoe:
@pranjalv-xilinx You might want to check src/runtime_src/core/common/detail/windows/trace.h

manikandan-xilinx:
retest this please. (seems like be one-off jenkins plugin issue during checkout)

pranjalv-xilinx (approved):

stsoe:
retest this please - Chinmay Khole
  [11:58 AM](https://xilinx.slack.com/archives/CG29V6XPF/p1696359538373779)
We are seeing some issues with U200 boards. 1 board has been disabled and will not run tests. However, latest 2 PR's were already started before other boards could be disabled. So we might see some failures. We  is looking into it.

PR #7729, opened 2023-10-02
Added compile time variable to only aie profile target

nishraptor:
Fixed an issue where Windows builds would fail to build correctly if this compile time flag is not passed. 

Added Compile time variable specifically to aie_profiling only when XDP_MINIMAL_BUILD is set.


stsoe (approved):
Yes, this looks right, except shouldn't it be specific to MSVC (windows)?

nishraptor:
> Yes, this looks right, except shouldn't it be specific to MSVC (windows)?

This is only for when XDP_MINIMAL_BUILD is set. For regular AIE profiling on windows, we will not need to do this.

jvillarre (approved):

PR #7736, opened 2023-10-07
add reference to pcidrv and pcidev

maxzhen:
Need to add below reference since objects can be freed our of order:
pcidev -> pcidrv
device_linux -> pcidev
N/A
N/A
Low risk since there is no code logic change.
Build and tested with xbutil validate.
Manually tested the reference by hacking up code.
N/A

### DIFF
--- a/src/runtime_src/core/pcie/linux/device_linux.h
+++ b/src/runtime_src/core/pcie/linux/device_linux.h
@@ -9,8 +9,14 @@
 #include "core/common/shim/buffer_handle.h"
 #include "core/common/shim/hwctx_handle.h"
 #include "core/pcie/common/device_pcie.h"
+#include "core/pcie/linux/pcidev.h"
 
 namespace xrt_core {
+
+// Forward declaration
+namespace pci {
+class dev;
+}
 
 // concrete class derives from device_pcie, but mixes in
 // shim layer functions for access through base class
@@ -18,6 +24,7 @@ class device_linux : public shim<device_pcie>
 {
 public:
   device_linux(handle_type device_handle, id_type device_id, bool user);
+  ~device_linux();
 
   // query functions
   virtual void read_dma_stats(boost::property_tree::ptree& pt) const;
@@ -86,9 +93,16 @@ public:
   {
     return xrt::shim_int::alloc_bo(get_device_handle(), userptr, size, xcl_bo_flags{flags}.flags);
   }
-  ////////////////////////////////////////////////////////////////
+
+protected:
+  std::shared_ptr<pci::dev>
+  get_dev() const
+  {
+    return m_pcidev;
+  }
 
 private:
+  std::shared_ptr<pci::dev> m_pcidev;
   // Private look up function for concrete query::request
   virtual const query::request&
   lookup_query(query::key_type query_key) const override;
--- a/src/runtime_src/core/pcie/linux/device_linux.h
+++ b/src/runtime_src/core/pcie/linux/device_linux.h
@@ -95,10 +95,10 @@ public:
   }
 
 protected:
-  std::shared_ptr<pci::dev>
+  pci::dev*
   get_dev() const
   {
-    return m_pcidev;
+    return m_pcidev.get();
   }
 
 private:
--- a/src/runtime_src/core/pcie/linux/pcidev.cpp
+++ b/src/runtime_src/core/pcie/linux/pcidev.cpp
@@ -455,28 +455,28 @@ open(const std::string& subdev, int flag) const
 }
 
 dev::
-dev(std::shared_ptr<const drv> driver, const std::string& sysfs)
-  : m_sysfs_name(sysfs)
-  , m_driver(driver)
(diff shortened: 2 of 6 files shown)


src/runtime_src/core/pcie/linux/device_linux.h
```diff
@@ -86,9 +93,16 @@ class device_linux : public shim<device_pcie>
...
     return xrt::shim_int::alloc_bo(get_device_handle(), userptr, size, xcl_bo_flags{flags}.flags);
   }
-  ////////////////////////////////////////////////////////////////
+
+protected:
+  std::shared_ptr<pci::dev>
+  get_dev() const
+  {
+    return m_pcidev;
+  }
```
stsoe, line 102: Suggest to return `pci::dev*` and not the shared pointer
```suggestion
  pci::dev*
  get_dev() const
  {
    return m_pcidev.get();
  }
```
The shared_ptr is for sharing ownership, not of the class I see to this function needs to share ownership of the pcidev.

src/runtime_src/core/pcie/linux/pcidev.cpp
```diff
@@ -455,21 +455,23 @@ open(const std::string& subdev, int flag) const
 }
 
 dev::
-dev(const drv& driver, const std::string& sysfs) : m_sysfs_name(sysfs), m_driver(driver)
+dev(std::shared_ptr<const drv> driver, const std::string& sysfs)
+  : m_sysfs_name(sysfs)
+  , m_driver(driver)
```
stsoe, line 460:
```suggestion
dev(std::shared_ptr<const drv> driver, std::string sysfs)
  : m_sysfs_name(std::move(sysfs))
  , m_driver(std::move(driver))
```

src/runtime_src/core/pcie/linux/pcidev.cpp
```diff
@@ -455,21 +455,23 @@ open(const std::string& subdev, int flag) const
...
 
   if(sscanf(sysfs.c_str(), "%hx:%hx:%hx.%hx", &m_domain, &m_bus, &m_dev, &m_func) < 4)
     throw std::invalid_argument(sysfs + " is not valid BDF");
 
-  m_is_mgmt = !driver.is_user();
+  m_is_mgmt = !driver->is_user();
 
   if (m_is_mgmt) {
     sysfs_get("", "instance", err, m_instance, static_cast<uint32_t>(INVALID_ID));
   } else {
```
stsoe, line 471:
```suggestion
  } 
  else {
```

src/runtime_src/core/pcie/linux/pcidev.h
```diff
@@ -185,12 +185,12 @@ class dev
 
 private:
   int
-  map_usr_bar();
+  map_usr_bar() const;
 
-  std::mutex m_lock;
-  char *m_user_bar_map = reinterpret_cast<char *>(MAP_FAILED);
+  mutable std::mutex m_lock;
+  mutable char *m_user_bar_map = reinterpret_cast<char *>(MAP_FAILED);
```
stsoe, line 191: While mutable mutexes are obvious, m_user_bar_map is less so.  Maybe comment what it is used for and why it is made useable and perceived as constant by const member functions.

stsoe (changes requested):
Looks good.  Few requests.

src/runtime_src/core/pcie/linux/pcidev.cpp
maxzhen: Can't make this change. Xbmgmt or xbutil will segfault. This is the constructor of pcidev object, caller may not expect that the arguments are std::move'd? Say, caller passes in a 'driver', to construct pcidev, then use the same 'driver' to do something else...

src/runtime_src/core/pcie/linux/device_linux.h
maxzhen: fixed.

src/runtime_src/core/pcie/linux/pcidev.cpp
maxzhen: fixed.

src/runtime_src/core/pcie/linux/pcidev.h
maxzhen: comment added.

src/runtime_src/core/pcie/linux/pcidev.cpp
stsoe: Both arguments are passed by value.  `driver` you already passed by value and I changed `sysfs` to by value, so yes, you can and should move them into the data members.  By-value-and-move is btw the recommend way to initialize data member.

stsoe (changes requested):

src/runtime_src/core/pcie/linux/pcidev.cpp
maxzhen: You're right. It is the implementation of the constructor which is not ready for the by-value-and-move change. Now it's fixed.

stsoe (approved):

PR #7745, opened 2023-10-13
Update trace logging to support Linux

stsoe:
Change how events are logged, the logger class is no longer needed for events, but may be used later for collecting stats.

On Linux trace point logging uses DTRACE events enabled via perf. 
On Windows trace point logging uses WPP enabled via xrt.ini or env.

Macros are used to interface with the platform specific logging infrastructure.  XRT code is agnostic to the underlying trace point infrastructure used.

Windows requires specific enablement via xrt.ini or environment variable.  This is documented in code.

The infrastructure itself is not exposed to applications, but of course usage require action from the user to enable and / or collect the trace point data

### DIFF
--- a/build/build_edge.sh
+++ b/build/build_edge.sh
@@ -47,6 +47,7 @@ install_recipes()
         echo "EXTERNALSRC = \"$XRT_REPO_DIR/src\"" >> $XRT_BB
         echo "EXTRA_OECMAKE += \"-DMY_VITIS=$XILINX_VITIS\"" >> $XRT_BB
         echo 'EXTERNALSRC_BUILD = "${WORKDIR}/build"' >> $XRT_BB
+	echo 'DEPENDS += " systemtap"' >> $XRT_BB
         echo 'PACKAGE_CLASSES = "package_rpm"' >> $XRT_BB
         echo 'LICENSE = "GPLv2 & Apache-2.0"' >> $XRT_BB
         echo 'LIC_FILES_CHKSUM = "file://../LICENSE;md5=de2c993ac479f02575bcbfb14ef9b485 \' >> $XRT_BB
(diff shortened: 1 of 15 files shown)


maxzhen (approved):

sonals (approved):
This is good, bringing Windows and Linux tracing as much closer as possible.

One question, we would create and destroy the trace object wherever trace point has been added even if tracing has not been enabled via perf. I would imagine that cost must be small?

stsoe:
> One question, we would create and destroy the trace object wherever trace point has been added even if tracing has not been enabled via perf. I would imagine that cost must be small?

If tracing is not enabled (and on Linux always), the code should optimize away.  On Linux something like
```
int bar(int y)
{
  XRT_TRACE_POINT_SCOPE(bar);
  XRT_TRACE_POINT_LOG(bar, y);
  { // 2. scope
    XRT_TRACE_POINT_SCOPE1(bar2, y);
  }
  return y;
}
```

results in this assembler code
```
00000000000011d0 <_Z3bari>:
    11d0:	f3 0f 1e fa          	endbr64 
    11d4:	89 f8                	mov    %edi,%eax
    11d6:	90                   	nop
    11d7:	90                   	nop
    11d8:	90                   	nop
    11d9:	90                   	nop
    11da:	90                   	nop
    11db:	c3                   	retq   
    11dc:	0f 1f 40 00          	nopl   0x0(%rax)
```
 I have not looked at Windows assembler code.

sonals:
Thanks for the ASM dump. I knew that the tracepoint themselves are NOPs but looks like the compiler is able to optimize away the struct constructor and destructor, possibly because they do nothing but call the DTRACE perf call. This is perfect!

stsoe:
retest this please - infra

manikandan-xilinx:
retest this please.

manikandan-xilinx:
retest this please. (rerun once without PACT just to isolate it's not due to PACT related infra changes for HW tests)

stsoe:
retest this please - apu failure

manikandan-xilinx:
retest this please. (updated the XILINX_VITIS mount point)

manikandan-xilinx:
retest this please. (retstored TA 2019.2_released)

PR #7746, opened 2023-10-13
VITIS-9655 Move code from host files to Test*.cpp

rchane:
https://jira.xilinx.com/browse/VITIS-9655
To allow for future JSON-configuration of test cases for devices, black-box tests from tests/validate were moved over to Test*.cpp files.
Tests were ported. Some tests required changes to use XRT native APIs for buffer objects and the like instead of OCL and are noted below.
Ported Test Rundown
TestPsPlVerify was previously done by Daniel. TestVCUKernel wasn't ported (it also would have needed an XRT API update). 
The `tests/validate` folder has been removed. 
The BusyBar has been moved out of TestPsPlVerify and has been implemented for the ported tests.
New TODO: The Hostmem-bw test seems to need a fix, as indicated above.
Large code change to tests. 
Note: Most of the lines deleted were from dependency files for VCUKernel (esp. the input file, 220k lines!)
Please test thoroughly. I have tested locally on vck5000 and u30 for the tests in the chart above. 
EDIT: Tested on XRT-IPU as well. Builds properly with the same final output.


stsoe:
@rchane One note here.  Please ensure that when this is clean in XRT that the changes will build with XRT-IPU and XRT-MCDM.

dayeh-xilinx:
retest this please.

src/runtime_src/core/tools/common/tests/TestBandwidthKernel.cpp
```diff
@@ -25,8 +36,241 @@ TestBandwidthKernel::run(std::shared_ptr<xrt_core::device> dev)
...
 }
+
+static bool 
+is_emulation() {
+  bool ret = false;
+  char* xcl_mode = std::getenv("XCL_EMULATION_MODE");
+  if (xcl_mode != nullptr) {
+    ret = true;
+  }
+  return ret;
```
AShivangi, line 50:
```suggestion
  return (std::getenv("XCL_EMULATION_MODE") != nullptr) ? true : false;
```

src/runtime_src/core/tools/common/tests/TestBandwidthKernel.cpp
```diff
@@ -25,8 +36,241 @@ TestBandwidthKernel::run(std::shared_ptr<xrt_core::device> dev)
...
+}
+
+void
+TestBandwidthKernel::runTest(std::shared_ptr<xrt_core::device> dev, boost::property_tree::ptree& ptree)
+{
+  const auto bdf_tuple = xrt_core::device_query<xrt_core::query::pcie_bdf>(dev);
+  const std::string bdf = xrt_core::query::pcie_bdf::to_string(bdf_tuple);
```
AShivangi, line 57: Are these used anywhere?

src/runtime_src/core/tools/common/tests/TestBandwidthKernel.cpp
```diff
@@ -25,8 +36,241 @@ TestBandwidthKernel::run(std::shared_ptr<xrt_core::device> dev)
...
+  const std::string test_path = findPlatformPath(dev, ptree);
```
AShivangi, line 58: I guess this is used with platform.json? Please define it closer to the actual implementation

src/runtime_src/core/tools/common/tests/TestBandwidthKernel.cpp
```diff
@@ -25,8 +36,241 @@ TestBandwidthKernel::run(std::shared_ptr<xrt_core::device> dev)
...
+  const std::string b_file = findXclbinPath(dev, ptree); // bandwidth.xclbin
+  bool flag_s = false;
```
AShivangi, line 60: I don't think we need this anymore

src/runtime_src/core/tools/common/tests/TestBandwidthKernel.cpp
```diff
@@ -25,8 +36,241 @@ TestBandwidthKernel::run(std::shared_ptr<xrt_core::device> dev)
...
+  if (flag_s || retVal == EOPNOTSUPP) {
+    ptree.put("status", test_token_skipped);
+    return;
+  } else if (retVal != EXIT_SUCCESS) {
+    logger(ptree, "Error", "Unknown error validating bandwidth xclbin");
+    ptree.put("status", test_token_failed);
+    return;
+  }
+
+  int num_kernel = 0, num_kernel_ddr = 0;
```
AShivangi, line 80:
```suggestion
  int num_kernel = 0;
  int num_kernel_ddr = 0;
```

src/runtime_src/core/tools/common/tests/TestBandwidthKernel.cpp
```diff
@@ -25,8 +36,241 @@ TestBandwidthKernel::run(std::shared_ptr<xrt_core::device> dev)
...
+    // Here Kernel object is created by specifying kernel name along with
+    // compute unit.
+    // For such case, this kernel object can only access the specific
+    // Compute unit
+    krnls[i] = xrt::kernel(device, xclbin_uuid, krnl_name_full.c_str());
+  }
+
+  double max_throughput = 0;
+  int reps = 10000;
+  if (num_kernel_ddr) {
```
AShivangi, line 127: As a coding practice, if you have more than 2 nested statements, you should re-write that piece of code. There is a better way to do it. Always follow "early exit"

src/runtime_src/core/tools/common/tests/TestBandwidthKernel.cpp
```diff
@@ -25,8 +36,241 @@ TestBandwidthKernel::run(std::shared_ptr<xrt_core::device> dev)
...
+
+      if (is_emulation()) {
+        reps = 2; // reducing the repeat count to 2 for emulation flow
+        // Running only upto 8K for emulation flow
+        if (data_size > 8 * 1024) {
+          break;
+        }
+      }
+
+      unsigned int vector_size_bytes = data_size;
```
AShivangi, line 140: there is too much happening in one func. Try to split it up in "init_buffer", "init_bo" etc. Look at this sample test case from Kelly: https://gitenterprise.xilinx.com/XRT/testcases/blob/master/xrt_qor_throughput/src/host_hal.cpp. This testcase is much more readable

src/runtime_src/core/tools/common/TestRunner.cpp
```diff
@@ -85,6 +99,33 @@ TestRunner::TestRunner (const std::string & test_name,
...
+  // Start the test process
+  std::thread test_thread([&] { runTestInternal(dev, result, this, is_thread_running); });
+  // Wait for the test process to finish
+  while (is_thread_running) {
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+    try {
+      busy_bar.check_timeout(MAX_TEST_DURATION);
+    } catch (const std::exception&) {
+      test_thread.detach();
+      throw;
```
AShivangi, line 120: What is this throw for?

AShivangi (changes requested):
I just reviewed TestRunner and TestBandwidth.cpp for now. Let's start with a single test case and see how it goes

src/runtime_src/core/tools/common/tests/TestBandwidthKernel.cpp
```diff
@@ -25,8 +36,241 @@ TestBandwidthKernel::run(std::shared_ptr<xrt_core::device> dev)
...
+
+  auto retVal = validate_binary_file(b_file);
+  if (flag_s || retVal == EOPNOTSUPP) {
+    ptree.put("status", test_token_skipped);
+    return;
+  } else if (retVal != EXIT_SUCCESS) {
+    logger(ptree, "Error", "Unknown error validating bandwidth xclbin");
+    ptree.put("status", test_token_failed);
+    return;
+  }
```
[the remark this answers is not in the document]
rchane: Thank you simplifying the logic here! The `validate_binary_file` function checks the stream like you said (returns EOPNOTSUPP if stream isn't good, EXIT_SUCCESS if it is); I think we use it in a few places, so that's why it was made a dedicated function. Should I go ahead and just check if the stream is good here?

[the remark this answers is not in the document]
AShivangi: yeah, 
i think  just do
```
if (!b_file.good())
  //skip
```

src/runtime_src/core/tools/common/TestRunner.cpp
rchane: I'll catch this in SubCmdValidate and fail the test if there's an exception like this (e.g. there's an issue with checking timeout).

chvamshi-xilinx (approved):

AShivangi (approved):
I think this looks good for first iteration. From what I understand, you have tested your changes with XRT-IPU and MCDM and everything works.

PR #7749, opened 2023-10-17
Test build, comment change only

stsoe:
Why is #7748 and #7745 failing?  Is it really because of xrtdeps.sh change adding a dependency on systemtap-sdt-dev?


stsoe:
No longer needed

PR #7756, opened 2023-10-18
VITIS-10045 - Fix a crash in xclbinUtil on windows platform

dezhiAmd:
Fix a crash in xclbinUtil
When running
xclbinutil --dump-section :json:json.txt --input vadd.xclbin
on windows OS, application crash
The root cause is that a local variable inside a function which will be released when out-of-scope is used as a return value by reference

Signed-off-by: Dezhi Liao [dezhi.liao@amd.com](mailto:dezhi.liao@amd.com)


stsoe (approved):

rozumx (approved):
Thank you for recognizing and addressing this issue.

stsoe:
Ok to test

stsoe:
Add to whitelist

PR #7743, opened 2023-10-12
VITIS-9943 Update XRT utilities to JSON-based infrastructure

dbenusov-xilinx:
#### Problem solved by the commit
https://jira.xilinx.com/browse/VITIS-8047

With the addition of different device categories, not all options are applicable to all devices as was the case in the past. This PR attempt to fix this issue by using a JSON configuration which states which options are applicable to certain devices.

This also updates SubCmdValidate to use the new JSONConfigurable class to interface with the xbutil JSON file!

#### Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered
Devices must have the display only the applicable options in the help menu and restrict execution during operation.

#### How problem was solved, alternative solutions (if any) and why they were rejected
Implemented using an existing JSONConfigurable base class and various edites to the Help menu classes to display the options.

#### Risks (if any) associated the changes in the commit
Changes to the help menu may confuse some people. Some options may break (hopefully not I tested them as best I could).

#### What has been tested and how, request additional testing if necessary
Ubuntu 22.04 U55c
OptionOptions Testing

Without device
```
dbenusov@xsjdbenusov50:/proj/rdi/staff/dbenusov/XRT$ xbutil configure --help

COMMAND: configure

DESCRIPTION: Device and host configuration.

USAGE: xbutil configure [ --host-mem | --p2p | --performance ] [--help] [-d arg]

OPTIONS:
 Common:
  -d, --device       - The Bus:Device.Function (e.g., 0000:d8:00.0) device of interest
  --help             - Help to use this sub-command
 AIE:
  --performance      - Change performance mode of the device
 Alveo:
  --host-mem         - Controls host-mem functionality
  --p2p              - Controls P2P functionality


GLOBAL OPTIONS:
  --verbose          - Turn on verbosity
  --batch            - Enable batch mode (disables escape characters)
  --force            - When possible, force an operation
```

With device (Alveo)
```
dbenusov@xsjdbenusov50:/proj/rdi/staff/dbenusov/XRT$ xbutil configure -d --help

COMMAND: configure

DESCRIPTION: Device and host configuration.

USAGE: xbutil configure [ --host-mem | --p2p ] [--help] [-d arg]

OPTIONS:
  -d, --device       - The Bus:Device.Function (e.g., 0000:d8:00.0) device of interest
  --help             - Help to use this sub-command
  --host-mem         - Controls host-mem functionality
  --p2p              - Controls P2P functionality


GLOBAL OPTIONS:
  --verbose          - Turn on verbosity
[...]

PR #7536, opened 2023-05-04
System class unification

vboggara-xilinx:
#### Problem solved by the commit
The system class boots up the system.  At core library load time, libxrt_core (shim library) instantiates concrete device factory class that are inserted into a singleton system class in xrt_coreutil.   When user host code creates an xrt::device object, then implementation goes through the singleton system class, find the proper device factory, and creates the device object.   The concrete device class is instantiated by the device factory at the shim level (libxrt_core).  Additional shim level objects are created through the interface of the device base class (xrt_core::device).  

#### How problem was solved, alternative solutions (if any) and why they were rejected
https://confluence.amd.com/pages/viewpage.action?spaceKey=AIE&title=XRT+ISHIM+Upgrade#XRTISHIMUpgrade-Systemclass
#### What has been tested and how, request additional testing if necessary
Validated below flows
PCIE - HW (testcase verification, plp_programming, xbutil and xbmgmt tools, all xbutil reports validation) , XMA
PCIE , EDGE - SW Emulation & HW Emulation
Windows - Compilation
#### Documentation impact (if any)
n/a

PR #7758, opened 2023-10-19
get type information from device

vboggara-xilinx:
get type information from the device.
requirement came from recent xbutil changes where some help menu options needs to be displayed based on the device type. 
added new api called get_type() in all device classes
Low
local cases.
n/a


stsoe (commented):
I would like to know how this will be used?  And why whatever information xbutil needs isn't implemented as a query request.

A string based coupling is not cool if it is used in comparisons.

dbenusov-xilinx:
@vboggara-xilinx Hey Naresh, I did not expect to see this change going into XRT separately from your ishim upgrade. I also agree with @stsoe that this should be a device query. At the moment I have a temporary function to identify device "class" being merged as part of https://github.com/Xilinx/XRT/pull/7743.

We would not need this change until everything is ready with https://github.com/Xilinx/XRT/pull/7536.

vboggara-xilinx:
closing this PR as per comments.

PR #7761, opened 2023-10-20
Fix for CR-1178694: Kernel Panic - Parallel XRT Kernels Run

saumyag-xilinx:
https://jira.xilinx.com/browse/CR-1178694
Partition fd from AIE driver is same in multithread case which leads to kernel crash as a thread tries to close the fd which is already closed by another thread, added lock before registering aie to handle this case.
Low
tested with a multithread application on vck190 
NA

### DIFF
--- a/src/runtime_src/core/edge/user/shim.cpp
+++ b/src/runtime_src/core/edge/user/shim.cpp
@@ -1739,7 +1739,7 @@ void
 shim::
 registerAieArray()
 {
-  std::lock_guard<std::mutex> guard(registerAieLock);
+  std::lock_guard guard(registerAieLock);
   delete aieArray.release();
   aieArray = std::make_unique<zynqaie::Aie>(mCoreDevice);
   aied = std::make_unique<zynqaie::Aied>(mCoreDevice.get());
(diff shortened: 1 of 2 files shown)


chvamshi-xilinx (approved):

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -1739,6 +1739,7 @@ void
 shim::
 registerAieArray()
 {
+  std::lock_guard<std::mutex> guard(registerAieLock);
   delete aieArray.release();
   aieArray = std::make_unique<zynqaie::Aie>(mCoreDevice);
```
stsoe, line 1744:
```suggestion
  std::lock_guard guard(registerAieLock);
  aieArray = std::make_unique<zynqaie::Aie>(mCoreDevice);
```

stsoe (changes requested):

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -1739,6 +1739,7 @@ void
 shim::
 registerAieArray()
 {
+  std::lock_guard guard(registerAieLock);
   delete aieArray.release();
```
stsoe, line 1743: Why?   The point of smart pointers if that we don't write new and delete.

chvamshi-xilinx:
Hi @saumyag-xilinx , 
Looks like this is still not working for multiple devices. Lets have a generic fix. Closing this for now

PR #7763, opened 2023-10-20
Fix to remove observed execution stall when aie_trace & host_trace is enabled.

vipangul:
[Debug]
  aie_trace=true
  host_trace=true

With the above plugins enabled, the generated native trace (native_trace.csv) is ~17MB. Also the generated events stored in unsortedEvents vector were also in the order of 500K.
Hence, existing filtering algorithm doesn't scale for the large no of events.

Modified the algorithm to move the the event contents of the vector to the end, so that it can be resized later at once. Since this method avoids the individual event erase and shifting the other events post erase; gains the needed performance.

Execution time before the change:
real    3m1.812s
user    2m55.907s
sys     0m3.915s

Execution time after the change:
real    0m13.913s
user    0m9.121s
sys     0m3.783s
Verified that behavior is not seen and have passed all the tests.

### DIFF
--- a/src/runtime_src/xdp/profile/database/dynamic_info/host_db.cpp
+++ b/src/runtime_src/xdp/profile/database/dynamic_info/host_db.cpp
@@ -18,6 +18,7 @@
 
 #include "xdp/profile/database/dynamic_info/host_db.h"
 #include "xdp/profile/database/events/vtf_event.h"
+#include <algorithm>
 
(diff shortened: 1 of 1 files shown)


jvillarre:
ok to test

stsoe:
ok to test

jvillarre (approved):

PR #7772, opened 2023-10-25
Updates to AIE Profile Client Plugin

nishraptor:
Refactoring the AIE Profile client plugin. 

- Updated metadata parsing to use parsing in aie_util.cpp and removed functions from aie_profile_metadata class
- Fixed issue with channel numbers not being configured unless both channels were specified in xrt.ini. The channel numbers are now parsed correctly when one channel is specified.
- Fixed issue with update device so the device information is updated properly from the metadata. The CSV now has the correct header.

Tested on client devices


stsoe (approved):
Approved as far as hw_context_int.h is concerned.

PR #7787, opened 2023-11-03
Modified CMake files to build xclbinutil seperately

ManojTakasi:
[CR-1179805](https://jira.xilinx.com/browse/CR-1179805) xclbinutil has to be build seperately from XRT.

NA

Introduced a new check in the CMake files called XCLBINUTIL_ONLY_BUILD. This determines whether to build xclbinutil only. If check is not specified, build process will proceed as usual including all components. Provided an option in build.sh script

NA
Conducted multiple builds with different checks to ensure thorough verification.


chvamshi-xilinx:
Hi @ManojTakasi ,
Lets use this opportunity to clean up the cmakefiles in XRT.  Closing this PR.

PR #7788, opened 2023-11-07
VITIS-10034 Add Usage metrics for all XRT objects

rbramand-xilinx:
Added new Usage metrics logger that logs info about XRT objects like device, buffer, hw context, kernel etc
This information is collected in thread safe manner and the info is dumped in JSON format to a file after application run finishes
The info collected and spec for this solution is captured in [confluence page](https://confluence.amd.com/display/~rbramand/XRT+Usage+metrics+Logger)

NA

The proposed solution is explained in above confluence page

Low

Tested with a simple multi threaded bandwidth application and was able to generate JSON file properly with necessary info
The following is the JSON dump generated with application that has 5 threads :
```
{
    "thread 0": {
        "device": {
            "device_index": "0",
            "bdf": "0000:3b:00.1",
            "bos_peak_count": "3",
            "global_bos": {
                "total_count": "1",
                "size": "67108864 bytes",
                "avg_size": "67108864 bytes",
                "peak_size": "67108864 bytes",
                "bytes_synced_to_device": "0 bytes",
                "bytes_synced_from_device": "0 bytes"
            },
            "hw_context": [
                {
                    "id": "0",
                    "xclbin_uuid": "b3b30b4a-5ded-3ee7-88a9-f9d0859459ea",
                    "bos": {
                        "total_count": "2",
                        "size": "134217728 bytes",
                        "avg_size": "67108864 bytes",
                        "peak_size": "67108864 bytes",
                        "bytes_synced_to_device": "67108864 bytes",
                        "bytes_synced_from_device": "67108864 bytes"
                    },
                    "kernels": [
                        {
                            "name": "bandwidth",
                            "num_of_args": "4",
                            "num_total_runs": "1",
                            "avg_run_time": "172652 us"
                        }
                    ]
                }
            ]
        }
    },
    "thread 1": {
        "device": {
            "device_index": "0",
            "bdf": "0000:3b:00.1",
            "bos_peak_count": "3",
            "global_bos": {
                "total_count": "1",
                "size": "67108864 bytes",
                "avg_size": "67108864 bytes",
                "peak_size": "67108864 bytes",
                "bytes_synced_to_device": "0 bytes",
                "bytes_synced_from_device": "0 bytes"
            },
            "hw_context": [
                {
                    "id": "0",
                    "xclbin_uuid": "b3b30b4a-5ded-3ee7-88a9-f9d0859459ea",
                    "bos": {
                        "total_count": "2",
                        "size": "134217728 bytes",
                        "avg_size": "67108864 bytes",
                        "peak_size": "67108864 bytes",
                        "bytes_synced_to_device": "67108864 bytes",
                        "bytes_synced_from_device": "67108864 bytes"
                    },
                    "kernels": [
                        {
                            "name": "bandwidth",
                            "num_of_args": "4",
                            "num_total_runs": "1",
                            "avg_run_time": "84010 us"
                        }
                    ]
                }
            ]
        }
    },
    "thread 2": {
        "device": {
            "device_index": "0",
            "bdf": "0000:3b:00.1",
            "bos_peak_count": "3",
            "global_bos": {
                "total_count": "1",
                "size": "67108864 bytes",
                "avg_size": "67108864 bytes",
                "peak_size": "67108864 bytes",
                "bytes_synced_to_device": "0 bytes",
                "bytes_synced_from_device": "0 bytes"
            },
            "hw_context": [
                {
                    "id": "0",
                    "xclbin_uuid": "b3b30b4a-5ded-3ee7-88a9-f9d0859459ea",
                    "bos": {
                        "total_count": "2",
                        "size": "134217728 bytes",
                        "avg_size": "67108864 bytes",
                        "peak_size": "67108864 bytes",
                        "bytes_synced_to_device": "67108864 bytes",
                        "bytes_synced_from_device": "67108864 bytes"
                    },
                    "kernels": [
                        {
                            "name": "bandwidth",
                            "num_of_args": "4",
                            "num_total_runs": "1",
                            "avg_run_time": "342594 us"
                        }
                    ]
                }
            ]
        }
    },
    "thread 3": {
        "device": {
            "device_index": "0",
            "bdf": "0000:3b:00.1",
            "bos_peak_count": "3",
            "global_bos": {
                "total_count": "1",
                "size": "67108864 bytes",
                "avg_size": "67108864 bytes",
                "peak_size": "67108864 bytes",
                "bytes_synced_to_device": "0 bytes",
                "bytes_synced_from_device": "0 bytes"
            },
            "hw_context": [
                {
                    "id": "0",
                    "xclbin_uuid": "b3b30b4a-5ded-3ee7-88a9-f9d0859459ea",
                    "bos": {
                        "total_count": "2",
                        "size": "134217728 bytes",
                        "avg_size": "67108864 bytes",
                        "peak_size": "67108864 bytes",
                        "bytes_synced_to_device": "67108864 bytes",
                        "bytes_synced_from_device": "67108864 bytes"
                    },
                    "kernels": [
                        {
                            "name": "bandwidth",
                            "num_of_args": "4",
                            "num_total_runs": "1",
                            "avg_run_time": "261316 us"
                        }
                    ]
                }
            ]
        }
    },
    "thread 4": {
        "device": {
            "device_index": "0",
            "bdf": "0000:3b:00.1",
            "bos_peak_count": "3",
            "global_bos": {
                "total_count": "1",
                "size": "67108864 bytes",
                "avg_size": "67108864 bytes",
                "peak_size": "67108864 bytes",
                "bytes_synced_to_device": "0 bytes",
                "bytes_synced_from_device": "0 bytes"
            },
            "hw_context": [
                {
                    "id": "0",
                    "xclbin_uuid": "b3b30b4a-5ded-3ee7-88a9-f9d0859459ea",
                    "bos": {
                        "total_count": "2",
                        "size": "134217728 bytes",
                        "avg_size": "67108864 bytes",
                        "peak_size": "67108864 bytes",
                        "bytes_synced_to_device": "67108864 bytes",
                        "bytes_synced_from_device": "67108864 bytes"
                    },
                    "kernels": [
                        {
                            "name": "bandwidth",
                            "num_of_args": "4",
                            "num_total_runs": "1",
                            "avg_run_time": "84019 us"
                        }
                    ]
                }
            ]
        }
    }
}
```

NA

### DIFF
--- a/src/runtime_src/core/common/api/hw_context_int.h
+++ b/src/runtime_src/core/common/api/hw_context_int.h
@@ -32,15 +32,6 @@ get_core_device_raw(const xrt::hw_context& ctx);
 void
 set_exclusive(xrt::hw_context& ctx);
 
-// Get xclbin uuid associated with this context
-xrt::uuid
-get_xclbin_uuid(const xrt::hw_context& hwctx);
-
-// Get underlying hw ctx handle, this is used to uniquely identify
-// hw ctx when logging info related to it
-xrt_core::hwctx_handle*
-get_hwctx_handle(const xrt::hw_context& hwctx);
-
 // Allows the creation of the hardware context from a void pointer
 // to the hardware context implementation. We use a void pointer
 // because we need to dynamically link to the callbacks that exist in 
--- a/src/runtime_src/core/common/api/hw_context_int.h
+++ b/src/runtime_src/core/common/api/hw_context_int.h
@@ -32,9 +32,12 @@ get_core_device_raw(const xrt::hw_context& ctx);
 void
 set_exclusive(xrt::hw_context& ctx);
 
+// Get xclbin uuid associated with this context
 xrt::uuid
 get_xclbin_uuid(const xrt::hw_context& hwctx);
 
+// Get underlying hw ctx handle, this is used to uniquely identify
+// hw ctx when logging info related to it
 xrt_core::hwctx_handle*
 get_hwctx_handle(const xrt::hw_context& hwctx);
 
--- a/src/runtime_src/core/common/api/hw_context_int.h
+++ b/src/runtime_src/core/common/api/hw_context_int.h
@@ -32,6 +32,12 @@ get_core_device_raw(const xrt::hw_context& ctx);
 void
 set_exclusive(xrt::hw_context& ctx);
 
+xrt::uuid
+get_xclbin_uuid(const xrt::hw_context& hwctx);
+
+xrt_core::hwctx_handle*
+get_hwctx_handle(const xrt::hw_context& hwctx);
+
 // Allows the creation of the hardware context from a void pointer
 // to the hardware context implementation. We use a void pointer
 // because we need to dynamically link to the callbacks that exist in 
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -118,10 +118,10 @@ public:
     return m_hdl.get();
   }
 
-  std::shared_ptr<xrt_core::usage_metrics::base_logger>
+  xrt_core::usage_metrics::base_logger*
   get_usage_logger()
   {
-    return m_usage_logger;
+    return m_usage_logger.get();
   }
 };
 
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -12,6 +12,7 @@
 #include "core/common/device.h"
 #include "core/common/trace.h"
 #include "core/common/shim/hwctx_handle.h"
+#include "core/common/usage_metrics.h"
 #include "core/common/xdp/profile.h"
 
 #include <limits>
@@ -32,6 +33,7 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
   cfg_param_type m_cfg_param;
   access_mode m_mode;
   std::unique_ptr<xrt_core::hwctx_handle> m_hdl;
+  std::shared_ptr<xrt_core::usage_metrics::base_logger> m_usage_logger;
 
 public:
   hw_context_impl(std::shared_ptr<xrt_core::device> device, const xrt::uuid& xclbin_id, const cfg_param_type& cfg_param)
@@ -40,6 +42,7 @@ public:
     , m_cfg_param(cfg_param)
     , m_mode(xrt::hw_context::access_mode::shared)
     , m_hdl{m_core_device->create_hw_context(xclbin_id, m_cfg_param, m_mode)}
+    , m_usage_logger(xrt_core::usage_metrics::get_usage_metrics_logger())
   {
   }
 
@@ -48,6 +51,7 @@ public:
     , m_xclbin{m_core_device->get_xclbin(xclbin_id)}
     , m_mode{mode}
     , m_hdl{m_core_device->create_hw_context(xclbin_id, m_cfg_param, m_mode)}
+    , m_usage_logger(xrt_core::usage_metrics::get_usage_metrics_logger())
   {}
 
   std::shared_ptr<hw_context_impl>
@@ -114,6 +118,12 @@ public:
   {
     return m_hdl.get();
   }
+
+  std::shared_ptr<xrt_core::usage_metrics::base_logger>
+  get_usage_logger()
+  {
+    return m_usage_logger;
+  }
 };
 
 } // xrt
@@ -170,6 +180,11 @@ alloc_hwctx_from_cfg(const xrt::device& device, const xrt::uuid& xclbin_id, cons
   // called in XDP create a hw_context to the underlying implementation
   xrt_core::xdp::update_device(handle.get());
 
+  handle->get_usage_logger()->log_hw_ctx_info(handle->get_core_device()->get_device_id(),
+      reinterpret_cast<uintptr_t>(handle->get_hwctx_handle()), xclbin_id);
+  
+  std::cout << "hw ctx hdl ptr - " << reinterpret_cast<uintptr_t>(handle->get_hwctx_handle()) << std::endl;
+
   return handle;
 }
 
@@ -184,6 +199,9 @@ alloc_hwctx_from_mode(const xrt::device& device, const xrt::uuid& xclbin_id, xrt
   // The create_hw_context_from_implementation function is then 
   // called in XDP create a hw_context to the underlying implementation
   xrt_core::xdp::update_device(handle.get());
+
+  handle->get_usage_logger()->log_hw_ctx_info(handle->get_core_device()->get_device_id(),
+      reinterpret_cast<uintptr_t>(handle->get_hwctx_handle()), xclbin_id);
 
   return handle;
 }
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -33,7 +33,8 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
   cfg_param_type m_cfg_param;
   access_mode m_mode;
   std::unique_ptr<xrt_core::hwctx_handle> m_hdl;
-  std::shared_ptr<xrt_core::usage_metrics::base_logger> m_usage_logger;
+  std::shared_ptr<xrt_core::usage_metrics::base_logger> m_usage_logger =
+      xrt_core::usage_metrics::get_usage_metrics_logger();
 
 public:
   hw_context_impl(std::shared_ptr<xrt_core::device> device, const xrt::uuid& xclbin_id, const cfg_param_type& cfg_param)
@@ -42,7 +43,6 @@ public:
     , m_cfg_param(cfg_param)
     , m_mode(xrt::hw_context::access_mode::shared)
     , m_hdl{m_core_device->create_hw_context(xclbin_id, m_cfg_param, m_mode)}
-    , m_usage_logger(xrt_core::usage_metrics::get_usage_metrics_logger())
   {
   }
 
@@ -51,7 +51,6 @@ public:
     , m_xclbin{m_core_device->get_xclbin(xclbin_id)}
     , m_mode{mode}
     , m_hdl{m_core_device->create_hw_context(xclbin_id, m_cfg_param, m_mode)}
-    , m_usage_logger(xrt_core::usage_metrics::get_usage_metrics_logger())
   {}
 
   std::shared_ptr<hw_context_impl>
@@ -149,18 +148,6 @@ void
 set_exclusive(xrt::hw_context& hwctx)
 {
   hwctx.get_handle()->set_exclusive();
-}
-
-xrt::uuid
-get_xclbin_uuid(const xrt::hw_context& hwctx)
-{
-  return hwctx.get_handle()->get_uuid();
-}
-
-xrt_core::hwctx_handle*
-get_hwctx_handle(const xrt::hw_context& hwctx)
-{
-  return hwctx.get_handle()->get_hwctx_handle();
 }
 
 xrt::hw_context
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -119,7 +119,7 @@ public:
     return m_hdl.get();
   }
 
-  inline std::shared_ptr<xrt_core::usage_metrics::base_logger>
+  std::shared_ptr<xrt_core::usage_metrics::base_logger>
   get_usage_logger()
   {
     return m_usage_logger;
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -119,7 +119,7 @@ public:
     return m_hdl.get();
   }
 
-  std::shared_ptr<xrt_core::usage_metrics::base_logger>
+  inline std::shared_ptr<xrt_core::usage_metrics::base_logger>
   get_usage_logger()
   {
     return m_usage_logger;
@@ -151,6 +151,18 @@ set_exclusive(xrt::hw_context& hwctx)
   hwctx.get_handle()->set_exclusive();
 }
 
+xrt::uuid
+get_xclbin_uuid(const xrt::hw_context& hwctx)
+{
+  return hwctx.get_handle()->get_uuid();
+}
+
+xrt_core::hwctx_handle*
+get_hwctx_handle(const xrt::hw_context& hwctx)
+{
+  return hwctx.get_handle()->get_hwctx_handle();
+}
+
 xrt::hw_context
 create_hw_context_from_implementation(void* hwctx_impl)
 {
@@ -180,10 +192,7 @@ alloc_hwctx_from_cfg(const xrt::device& device, const xrt::uuid& xclbin_id, cons
   // called in XDP create a hw_context to the underlying implementation
   xrt_core::xdp::update_device(handle.get());
 
-  handle->get_usage_logger()->log_hw_ctx_info(handle->get_core_device()->get_device_id(),
-      reinterpret_cast<uintptr_t>(handle->get_hwctx_handle()), xclbin_id);
-  
-  std::cout << "hw ctx hdl ptr - " << reinterpret_cast<uintptr_t>(handle->get_hwctx_handle()) << std::endl;
+  handle->get_usage_logger()->log_hw_ctx_info(handle.get());
 
   return handle;
 }
@@ -200,8 +209,7 @@ alloc_hwctx_from_mode(const xrt::device& device, const xrt::uuid& xclbin_id, xrt
   // called in XDP create a hw_context to the underlying implementation
   xrt_core::xdp::update_device(handle.get());
 
-  handle->get_usage_logger()->log_hw_ctx_info(handle->get_core_device()->get_device_id(),
-      reinterpret_cast<uintptr_t>(handle->get_hwctx_handle()), xclbin_id);
+  handle->get_usage_logger()->log_hw_ctx_info(handle.get());
 
   return handle;
 }
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -2348,7 +2348,7 @@ public:
     // This is in critical path, we need to reduce log overhead 
     // as much as possible, passing kernel impl pointer instead of 
     // constructing args in place
-    m_usage_logger->log_kernel_start_info(kernel.get(), this);
+    m_usage_logger->log_kernel_run_info(kernel.get(), this, true, ERT_CMD_STATE_NEW);
     cmd->run();
   }
 
@@ -2429,14 +2429,21 @@ public:
   ert_cmd_state
   wait(const std::chrono::milliseconds& timeout_ms) const
   {
+    ert_cmd_state state {ERT_CMD_STATE_NEW}; // initial value doesn't matter
     if (timeout_ms.count()) {
       auto [ert_state, cv_status] = cmd->wait(timeout_ms);
-      return (cv_status == std::cv_status::timeout)
-        ? ERT_CMD_STATE_TIMEOUT
-        : ert_state;
+      if (cv_status == std::cv_status::timeout)
+        return ERT_CMD_STATE_TIMEOUT;
+
+      state = ert_state;
+    }
+    else {
+      state = cmd->wait();
     }
 
-    return cmd->wait();
+    m_usage_logger->log_kernel_run_info(kernel.get(), this, false, state);
+
+    return state;
   }
 
 
@@ -2459,8 +2466,10 @@ public:
       state = cmd->wait();
     }
 
-    if (state == ERT_CMD_STATE_COMPLETED)
+    if (state == ERT_CMD_STATE_COMPLETED) {
+      m_usage_logger->log_kernel_run_info(kernel.get(), this, false, state);
       return std::cv_status::no_timeout;
+    }
 
     std::string msg = "Command failed to complete successfully (" + cmd_state_to_string(state) + ")";
     throw xrt::run::command_error(state, msg);
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -1297,7 +1297,8 @@ private:
   size_t num_cumasks = 1;              // Required number of command cu masks
   control_type protocol = control_type::none; // Default opcode
   uint32_t uid;                        // Internal unique id for debug
-  std::shared_ptr<xrt_core::usage_metrics::base_logger> m_usage_logger;
+  std::shared_ptr<xrt_core::usage_metrics::base_logger> m_usage_logger =
+      xrt_core::usage_metrics::get_usage_metrics_logger();
 
   // Open context of a specific compute unit.
   //
@@ -1521,7 +1522,6 @@ public:
     , xkernel(get_kernel_or_error(xclbin, name))               // kernel meta data managed by xclbin
     , properties(xrt_core::xclbin_int::get_properties(xkernel))// cache kernel properties
     , uid(create_uid())
-    , m_usage_logger(xrt_core::usage_metrics::get_usage_metrics_logger())
   {
     XRT_DEBUGF("kernel_impl::kernel_impl(%d)\n" , uid);
 
@@ -2080,7 +2080,8 @@ class run_impl
   uint32_t uid;                           // internal unique id for debug
   std::unique_ptr<arg_setter> asetter;    // helper to populate payload data
   bool encode_cumasks = false;            // indicate if cmd cumasks must be re-encoded
-  std::shared_ptr<xrt_core::usage_metrics::base_logger> m_usage_logger;
+  std::shared_ptr<xrt_core::usage_metrics::base_logger> m_usage_logger =
+      xrt_core::usage_metrics::get_usage_metrics_logger();
 
 public:
   uint32_t
@@ -2125,7 +2126,6 @@ public:
     , data(initialize_command(cmd.get()))
     , m_header(0)
     , uid(create_uid())
-    , m_usage_logger(xrt_core::usage_metrics::get_usage_metrics_logger())
   {
     XRT_DEBUGF("run_impl::run_impl(%d)\n" , uid);
   }
@@ -2145,7 +2145,6 @@ public:
     , m_header(rhs->m_header)
     , uid(create_uid())
     , encode_cumasks(rhs->encode_cumasks)
-    , m_usage_logger(xrt_core::usage_metrics::get_usage_metrics_logger())
   {
     XRT_DEBUGF("run_impl::run_impl(%d)\n" , uid);
   }
@@ -3199,12 +3198,6 @@ get_regmap_size(const xrt::kernel& kernel)
     return kernel.get_handle()->get_regmap_size();
 }
 
-std::string
-get_kernel_name(const xrt::kernel& kernel)
-{
-  return kernel.get_handle()->get_name();
-}
-
 xrt::hw_context
 get_hw_ctx(const xrt::kernel& kernel)
 {
@@ -3212,13 +3205,12 @@ get_hw_ctx(const xrt::kernel& kernel)
 }
 
 xrt::kernel
-create_kernel_from_implementation(void* kernel_impl)
+create_kernel_from_implementation(const xrt::kernel_impl* kernel_impl)
 {
   if (!kernel_impl)
     throw std::runtime_error("Invalid kernel context implementation."); 
 
-  xrt::kernel_impl* impl_ptr = static_cast<xrt::kernel_impl*>(kernel_impl);
-  return xrt::kernel(impl_ptr->get_shared_ptr());
+  return xrt::kernel(const_cast<xrt::kernel_impl*>(kernel_impl)->get_shared_ptr());
 }
 
 }} // kernel_int, xrt_core
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -1557,7 +1557,7 @@ public:
     // amend args with computed data based on kernel protocol
     amend_args();
 
-    m_usage_logger->log_kernel_info(device->core_device, hwctx, name, args.size());
+    m_usage_logger->log_kernel_info(device->core_device.get(), hwctx, name, args.size());
   }
 
   // Delegating constructor with no module
@@ -2347,7 +2347,8 @@ public:
     // This is in critical path, we need to reduce log overhead 
     // as much as possible, passing kernel impl pointer instead of 
     // constructing args in place
-    m_usage_logger->log_kernel_run_info(kernel.get(), this, true, ERT_CMD_STATE_NEW);
+    // sending state as ERT_CMD_STATE_NEW for kernel start
+    m_usage_logger->log_kernel_run_info(kernel.get(), this, ERT_CMD_STATE_NEW);
     cmd->run();
   }
 
@@ -2440,7 +2441,7 @@ public:
       state = cmd->wait();
     }
 
-    m_usage_logger->log_kernel_run_info(kernel.get(), this, false, state);
+    m_usage_logger->log_kernel_run_info(kernel.get(), this, state);
 
     return state;
   }
@@ -2466,7 +2467,7 @@ public:
     }
 
     if (state == ERT_CMD_STATE_COMPLETED) {
-      m_usage_logger->log_kernel_run_info(kernel.get(), this, false, state);
+      m_usage_logger->log_kernel_run_info(kernel.get(), this, state);
       return std::cv_status::no_timeout;
     }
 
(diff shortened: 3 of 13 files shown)


src/runtime_src/core/common/usage_metrics.h
```diff
@@ -0,0 +1,167 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2023 Advanced Micro Devices, Inc. All rights reserved.
+#ifndef XRT_CORE_USAGE_METRICS_H
+#define XRT_CORE_USAGE_METRICS_H
+
+#include <chrono>
+#include <map>
+#include <memory>
+#include <stdint.h>
```
stsoe, line 9:
```suggestion
#include <cstdint>
```

src/runtime_src/core/common/usage_metrics.h
```diff
@@ -0,0 +1,167 @@
...
+struct device_metrics
+{
+  std::string bdf = "";
+  struct bo_metrics global_bos_met;
+  uint32_t bo_active_count = 0;
+  uint32_t bo_peak_count = 0;
+  std::vector<struct hw_ctx_metrics> hw_ctx_vec;
+};
+
+using metrics_map = std::map<std::thread::id, std::map<unsigned int, device_metrics>>;
```
stsoe, line 86: I don't think any of these structs are needed in the header file?
Can they be moved to the .cpp file?
This goes for the concrete `class usage_metrics_logge` also, it can be moved to .cpp.
We really don't want to expose the implementation of the metrics to  compilation units that include this header.

src/runtime_src/core/common/usage_metrics.h
```diff
@@ -0,0 +1,167 @@
...
+  log_buffer_info_destruct(unsigned int) {}
+
+  virtual void
+  log_buffer_sync(unsigned int, void*, size_t, xclBOSyncDirection) {}
+
+  virtual void
+  log_kernel_info(std::shared_ptr<xrt_core::device>, const xrt::hw_context&, const std::string&, size_t) {}
+
+  virtual void
+  log_kernel_run_info(void*, const void*, bool, ert_cmd_state) {}
```
stsoe, line 113: I am not sure it worthwhile to go to void*, pass actual objects by const reference is probably better.

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -40,6 +42,7 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
     , m_cfg_param(cfg_param)
     , m_mode(xrt::hw_context::access_mode::shared)
     , m_hdl{m_core_device->create_hw_context(xclbin_id, m_cfg_param, m_mode)}
+    , m_usage_logger(xrt_core::usage_metrics::get_usage_metrics_logger())
```
stsoe, line 45: You could use default member initialization since no constructor arguments are relied on.

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -48,6 +51,7 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
     , m_xclbin{m_core_device->get_xclbin(xclbin_id)}
     , m_mode{mode}
     , m_hdl{m_core_device->create_hw_context(xclbin_id, m_cfg_param, m_mode)}
+    , m_usage_logger(xrt_core::usage_metrics::get_usage_metrics_logger())
```
stsoe, line 54: ditto

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -141,6 +151,18 @@ set_exclusive(xrt::hw_context& hwctx)
...
+get_xclbin_uuid(const xrt::hw_context& hwctx)
+{
+  return hwctx.get_handle()->get_uuid();
+}
+
+xrt_core::hwctx_handle*
+get_hwctx_handle(const xrt::hw_context& hwctx)
+{
+  return hwctx.get_handle()->get_hwctx_handle();
+}
```
stsoe, line 164: Probably not needed see header file comment

src/runtime_src/core/common/usage_metrics.h
```diff
@@ -0,0 +1,167 @@
...
+  std::vector<uint32_t> cu_index_vec;
+  uint32_t total_runs = 0;
+  std::chrono::microseconds total_time = {};
+  std::unordered_map<uintptr_t, timestamp> exec_times; // run handle ptr is used for indexing
+  size_t num_args;
+};
+
+struct hw_ctx_metrics
+{
+  uintptr_t id;  // using hw_ctx handle ptr as unique identifier for logging
```
stsoe, line 71: You might as well use xrt::hw_context if possible.
At a minimum use hwctx_handle*

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -3171,6 +3199,28 @@ get_regmap_size(const xrt::kernel& kernel)
     return kernel.get_handle()->get_regmap_size();
 }
 
+std::string
+get_kernel_name(const xrt::kernel& kernel)
+{
+  return kernel.get_handle()->get_name();
+}
```
stsoe, line 3206: Not needed

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -3171,6 +3199,28 @@ get_regmap_size(const xrt::kernel& kernel)
...
+
+xrt::hw_context
+get_hw_ctx(const xrt::kernel& kernel)
+{
+  return kernel.get_handle()->get_hw_context();
+}
+
+xrt::kernel
+create_kernel_from_implementation(void* kernel_impl)
```
stsoe, line 3215:
```suggestion
create_kernel_from_implementation(const hwctx_impl* kernel_impl)
```
At least use the hwctx_impl pointer.  I am not sure about the `const` in context of getting shared from this and `const` would require the shared_ptr<base_logger> to be mutable for sure.

stsoe (changes requested):
Lots of comments.   Let me know what you think, not all may be possible.  We can meet and discuss.

rbramand-xilinx:
Thanks Soren :)

rbramand-xilinx:
> This is very good Rahul. Nice clean code. Lots of comments. Let me know what you think, not all may be possible. We can meet and discuss. Thank you for making this a manageable pull request. The changes are about what I can comprehend in one sitting :-)

Thanks Soren :)

src/runtime_src/core/common/shim/buffer_handle.h
```diff
@@ -39,6 +40,9 @@ class buffer_handle
   };
 
 public:
+  XCL_DRIVER_DLLESPEC
```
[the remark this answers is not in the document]
rbramand-xilinx: As buffer_handle class is inherited by shim buffer_object classes and when shim objects are constructed constructor of buffer_handle is called and if we don't export it throws linker error in windows flow. That said this goes away as I will move logging to bo_impl class instead of buffer_handle

src/runtime_src/core/common/shim/buffer_handle.h
```diff
@@ -4,6 +4,7 @@
 #define XRT_CORE_BUFFER_HANDLE_H
 
 #include "core/common/shim/shared_handle.h"
+#include "core/common/usage_metrics.h"
```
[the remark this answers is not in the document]
rbramand-xilinx: sure Soren I will move logging to bo_impl class which would be more flexible. Thanks

src/runtime_src/core/common/api/hw_context_int.h
```diff
@@ -32,6 +32,15 @@ get_core_device_raw(const xrt::hw_context& ctx);
 void
 set_exclusive(xrt::hw_context& ctx);
 
+// Get xclbin uuid associated with this context
+xrt::uuid
+get_xclbin_uuid(const xrt::hw_context& hwctx);
```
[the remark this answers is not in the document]
rbramand-xilinx: I forgot to define XRT_API_SOURCE macro and was getting linker warnings in windows flow so defined new apis but now after defining the macro these warnings went away. I will use existing apis for other calls as well

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -114,6 +117,12 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
   {
     return m_hdl.get();
   }
+
+  std::shared_ptr<xrt_core::usage_metrics::base_logger>
+  get_usage_logger()
+  {
+    return m_usage_logger;
+  }
```
stsoe, line 125: I understand the point of keeping the user_logger as a shared_ptr data member, but this function should return `xrt_core::usage_metrics::base_logger*`, because callers will not share ownership.  Please change all such functions in other compilation units also if there are any same functions.
```suggestion
  xrt_core::usage_metrics::base_logger*
  get_usage_logger()
  {
    return m_usage_logger.get();
  }
```

src/runtime_src/core/common/usage_metrics.h
```diff
@@ -0,0 +1,81 @@
...
+  log_buffer_info_destruct(unsigned int) {}
+
+  virtual void
+  log_buffer_sync(unsigned int, const xrt_core::hwctx_handle*, size_t, xclBOSyncDirection) {}
+
+  virtual void
+  log_kernel_info(std::shared_ptr<xrt_core::device>, const xrt::hw_context&, const std::string&, size_t) {}
+
+  virtual void
+  log_kernel_run_info(const xrt::kernel_impl*, const xrt::run_impl*, bool, ert_cmd_state) {}
```
stsoe, line 66: What is the bool.  Probably document these APIs.

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,467 @@
...
+                kernel_it->total_time += std::chrono::duration_cast<std::chrono::microseconds>(ts_now - kernel_it->exec_times[run_hdl].start_time);
+              }
+              // invalidate start time, run may be finished, aborted or timed out
+              kernel_it->exec_times[run_hdl].is_valid = false;
+            }
+          }
+        }
+      }
+    }
+  }
```
stsoe, line 461: Much too deeply nested.   Use early exit, and/or use automatic binding to [first, second], and/or refactor into more functions.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -2330,6 +2343,11 @@ class run_impl
       xrt_core::module_int::sync(m_module);
 
     prep_start();
+    // log kernel start info
+    // This is in critical path, we need to reduce log overhead 
+    // as much as possible, passing kernel impl pointer instead of 
+    // constructing args in place
+    m_usage_logger->log_kernel_run_info(kernel.get(), this, true, ERT_CMD_STATE_NEW);
```
stsoe, line 2350: I wonder if the bool argument to `log_kernel_run_info` is needed to could simply be inferred from `state != ERT_CMD_STATE_NEW` ?

src/runtime_src/core/common/device.h
```diff
@@ -463,6 +464,15 @@ class device : public ishim
...
   }
 
+  /**
+   * get_usage_logger() - get usage metrics logger
+   */
+  std::shared_ptr<usage_metrics::base_logger>
+  get_usage_logger()
+  {
+    return m_usage_logger;
+  }
```
stsoe, line 474: return `usage_metrics::base_logger*`

stsoe (changes requested):
Some more changes.  Please review.

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,467 @@
...
+  std::string bdf = "";
+  struct bo_metrics global_bos_met;
+  uint32_t bo_active_count = 0;
+  uint32_t bo_peak_count = 0;
+  std::vector<struct hw_ctx_metrics> hw_ctx_vec;
+};
+
+using metrics_map = std::map<std::thread::id, std::map<unsigned int, device_metrics>>;
+// Global map to store usage metrics of all threads
+std::shared_ptr<metrics_map> usage_metrics_map = std::make_shared<metrics_map>();
```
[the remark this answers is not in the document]
rbramand-xilinx: As this map is global and we want to make sure this map gets destroyed after all thread objects are destroyed I am maintaining a reference of it in usage_metrics class. So made it a shared_ptr.

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,467 @@
...
+print_usage_metrics()
+{
+  bpt::ptree thread_array;
+
+  uint32_t t_count = 0;
+  // iterate over all threads
+  for (const auto& thread_it : *usage_metrics_map) {
+    bpt::ptree dev_array;
+    // iterate over all devices
+    for (auto dev_it : thread_it.second) {
```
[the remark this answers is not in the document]
rbramand-xilinx: sure Soren, I will make changes accordingly

src/runtime_src/core/common/usage_metrics.cpp
rbramand-xilinx: Sure Soren

stsoe (changes requested):
I am sorry to drag this on.  But it is making good progress.  Just need to get everything perfect :-)

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,525 @@
...
+#include <thread>
+
+#ifdef _WIN32
+# pragma warning ( disable : 4996 )
+#endif
+
+namespace bpt = boost::property_tree;
+namespace {
+// global variables
+std::mutex m;
```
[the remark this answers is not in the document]
keryell: No need since the preferred modern coding style using anonymous namespace is used here, isn't it?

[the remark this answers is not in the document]
stsoe: Right, this has always confused me.   static was deprecated in unnamed namespace at some point, then undeprecated.   I don't think it hurts anything. And I was merely looking at the inconsistency between line 35 and 36.

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,525 @@
...
+
+  auto map_it = m_dev_map.find(dev_id);
+  // dont log if device didn't match
+  if (map_it == m_dev_map.end())
+    return;
+  
+  auto& [device_id, dev_metrics] = *map_it;
+  if (!handle) {
+    // global bo
+    bo_met = &dev_metrics.global_bos_met;
```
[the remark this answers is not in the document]
rbramand-xilinx: BO's that are created without using hw_context

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,525 @@
...
+  std::string bdf = "";
+  struct bo_metrics global_bos_met;
+  uint32_t bo_active_count = 0;
+  uint32_t bo_peak_count = 0;
+  std::vector<struct hw_ctx_metrics> hw_ctx_vec;
+};
+
+using metrics_map = std::map<std::thread::id, std::map<device_id, device_metrics>>;
+// Global map to store usage metrics of all threads
+std::shared_ptr<metrics_map> usage_metrics_map = std::make_shared<metrics_map>();
```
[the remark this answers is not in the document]
rbramand-xilinx: This map is used for storing metrics of all threads and each thread when going out of scope pushes its data to map.
As the threads data is static thread_local (also static) we need to make sure this map should be destroyed after all static thread_local objects are destroyed so using a reference of it in each thread_local object created.

I have tried simple test case as below and found that if we dont hold reference of global object the order of destruction is random (atleast in Windows) but if we hold reference then order of destruction is as expected.

```
#include <iostream>
#include <thread>
#include <memory>

using namespace std;

// Global object 
struct GlobalObject {
    GlobalObject() {
        std::cout << "Global object constructed." << std::endl;
    }

    ~GlobalObject() {
        std::cout << "Global object destroyed." << std::endl;
    }
};

static GlobalObject globalObj;
//std::shared_ptr<GlobalObject> globalObj = std::make_shared<GlobalObject>();

void foo() {
    // Static object within a function
    struct StaticObject {
        //std::shared_ptr<GlobalObject> ref = globalObj;
        StaticObject() {
            std::cout << "Static object constructed." << std::endl;
        }

        ~StaticObject() {
            std::cout << "Static object destroyed." << std::endl;
        }
    };
    static thread_local StaticObject staticObj;
    //static StaticObject staticObj;
}

int main() {
    foo();

    thread t(foo);
    t.join();

    foo();

    return 0;
}
```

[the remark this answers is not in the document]
keryell: On my Linux laptop I get:
```
Global object constructed.
Static object constructed.
Static object constructed.
Static object destroyed.
Static object destroyed.
Global object destroyed.
```
That sounds good.
Do you have something different on Windows? If so, buggy old Windows compiler or runtime?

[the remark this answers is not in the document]
stsoe: Order of destruction is well-defined within a compilation unit.  Globals are constructed in order defined and destructed in reverse order.  It does not matter if the globals are user defined objects or a shared_ptr to a user defined objects.  If you see random behavior, it's a compiler bug :-)  I am expecting exactly the output Ronan shows, the global object is controlled outside the scope of main, constructed during static global construction and destructed during static global destruction.  The static local objects are within the scope of main.  Anything else is crazy :-)

Show me the buggy output, and where I can duplicate that behavior; I'd be very interested in seeing that.   If it is fixed by a shared_ptr it sounds like an unreliable work-around.

src/runtime_src/core/common/usage_metrics.cpp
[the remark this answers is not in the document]
stsoe: Got it.  I think it wouldn't hurt having two log_buffer_info_construct overloads, one for device and one for hwctx.  This assumes some refactoring for the shared pieces is possible.

src/runtime_src/core/common/usage_metrics.cpp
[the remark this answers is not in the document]
stsoe: Hey @rbramand-xilinx I missed the fact that you actually share owner ship of the metrics map with the thread's usage_metrics_logger.  I think this makes sense and certainly controls the ownership of the global until the last thread is gone.    I rest my case, not wanting to dig into the order of destruction of namespace globals vs statically scoped thread local variables within the get_usage_metrics_logger.   What you have definitely doesn't hurt.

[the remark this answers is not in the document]
rbramand-xilinx: Hi @stsoe and @keryell, I too feel its because of old windows compiler bug, in Linux I observed the proper output for the example I shared.

src/runtime_src/core/common/usage_metrics.cpp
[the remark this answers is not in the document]
rbramand-xilinx: I will remove the static from other global under unnamed namespace as well for uniformity.

src/runtime_src/core/common/usage_metrics.cpp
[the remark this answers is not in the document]
rbramand-xilinx: Soren if we implement overload function for global and hw ctx bo we will have 2 functions each for construct, destruct and sync bo. So using single function to avoid this. But I have refactored code to push common functionality into different function

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,515 @@
...
+  auto it = std::find_if(metrics_vec.begin(), metrics_vec.end(), 
+                        [finder](met_type& met) 
+                        {
+                          return met.handle == finder;
+                        }
+                        );
+  return it == metrics_vec.end()
+    ? nullptr
+    : &(*it);
+}
```
stsoe, line 54: The convention in at least most of our code is to use CamelCase (sorry, yes I hate CamelCase in all other cases) for template parameters so as not to confuse with variables and user defined types.
```suggestion
template <typename MetType, typename FindType>
static MetType*
get_metrics(std::vector<MetType>& metrics_vec, const FindType& finder)
{
  auto it = std::find_if(metrics_vec.begin(), metrics_vec.end(), 
                        [finder](const auto& met) 
                        {
                          return met.handle == finder;
                        }
                        );
  return it == metrics_vec.end()
    ? nullptr
    : &(*it);
}
```

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,515 @@
...
+    auto dev_id = xrt_core::hw_context_int::get_core_device(hw_ctx)->get_device_id();
+    auto uuid = hw_ctx.get_xclbin_uuid();
+
+    // dont log if device didn't match
+    auto dev_metrics = get_device_metrics(m_dev_map, dev_id);
+    if (!dev_metrics)
+      return;
+
+    // log if this entry is not logged before
+    if (!get_metrics<hw_ctx_metrics, xrt_core::hwctx_handle*>(dev_metrics->hw_ctx_vec, hwctx_handle))
```
stsoe, line 384: Template types are automatically inferred when you are passing all arguments.
```suggestion
    if (!get_metrics(dev_metrics->hw_ctx_vec, hwctx_handle))
```

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,515 @@
...
+
+static bo_metrics*
+get_buffer_metrics(device_metrics* dev_metrics, const xrt_core::hwctx_handle* handle)
+{
+  if (!handle) {
+    // global bo
+    return &dev_metrics->global_bos_met;
+  }
+  else {
+    auto hw_ctx_met = get_metrics<hw_ctx_metrics, const xrt_core::hwctx_handle*>(dev_metrics->hw_ctx_vec, handle);
```
stsoe, line 400:
```suggestion
    auto hw_ctx_met = get_metrics(dev_metrics->hw_ctx_vec, handle);
```

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,515 @@
...
+log_kernel_info(const xrt_core::device* dev, const xrt::hw_context& ctx, const std::string& name, size_t args)
+{
+  auto dev_id = dev->get_device_id();
+  auto hwctx_handle = static_cast<xrt_core::hwctx_handle*>(ctx);
+
+  auto dev_metrics = get_device_metrics(m_dev_map, dev_id);
+  if (!dev_metrics)
+    return;
+
+  auto hw_ctx_met = get_metrics<hw_ctx_metrics, xrt_core::hwctx_handle*>(dev_metrics->hw_ctx_vec, hwctx_handle);
```
stsoe, line 468:
```suggestion
  auto hw_ctx_met = get_metrics(dev_metrics->hw_ctx_vec, hwctx_handle);
```

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,515 @@
...
+  // dont log if hw ctx didn't match existing ones
+  if (!hw_ctx_met)
+    return;
+  
+  // log if this entry is not logged before
+  if (!get_metrics<kernel_metrics, std::string>(hw_ctx_met->kernel_metrics_vec, name)) {
```
stsoe, line 474:
```suggestion
  if (!get_metrics(hw_ctx_met->kernel_metrics_vec, name)) {
```

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,515 @@
...
+    auto hwctx_handle = static_cast<xrt_core::hwctx_handle*>(hw_ctx);
+
+    auto dev_id = xrt_core::hw_context_int::get_core_device(hw_ctx)->get_device_id();
+    auto name = kernel.get_name();
+
+    auto dev_metrics = get_device_metrics(m_dev_map, dev_id);
+    if (!dev_metrics)
+      return;
+
+    auto hw_ctx_met = get_metrics<hw_ctx_metrics, xrt_core::hwctx_handle*>(dev_metrics->hw_ctx_vec, hwctx_handle);
```
stsoe, line 499:
```suggestion
    auto hw_ctx_met = get_metrics(dev_metrics->hw_ctx_vec, hwctx_handle);
```

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,515 @@
...
+    // dont log if hw ctx didn't match existing ones
+    if (!hw_ctx_met)
+      return;
+  
+    auto kernel_met = get_metrics<kernel_metrics, std::string>(hw_ctx_met->kernel_metrics_vec, name);
```
stsoe, line 504:
```suggestion
    auto kernel_met = get_metrics(hw_ctx_met->kernel_metrics_vec, name);
```

stsoe (changes requested):

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,515 @@
...
+  uint32_t total_count = 0;
+  size_t   total_size_in_bytes = 0;
+  size_t   peak_size_in_bytes = 0;
+  size_t   bytes_synced_to_device = 0;
+  size_t   bytes_synced_from_device = 0;
+};
+
+using tp = std::chrono::time_point<std::chrono::high_resolution_clock>;
+struct kernel_metrics
+{
```
stsoe, line 67: Scope to where it is used
```suggestion
struct kernel_metrics
{
  using tp = std::chrono::time_point<std::chrono::high_resolution_clock>;
```

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,515 @@
...
+    }
+  }
+};
+
+struct hw_ctx_metrics
+{
+  const xrt_core::hwctx_handle* handle;  // using hw_ctx handle ptr as unique identifier for logging
+  xrt::uuid xclbin_uuid;
+  struct bo_metrics bos_met;
+  std::vector<struct kernel_metrics> kernel_metrics_vec;
```
stsoe, line 109: In C++ the use of `struct` where types are used is not commonly used.   I would recommend just
```suggestion
  std::vector<kernel_metrics> kernel_metrics_vec;
```

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,515 @@
...
+
+  void
+  log_kernel(const std::string& name, size_t args)
+  {
+    struct kernel_metrics k;
```
stsoe, line 114:
```suggestion
    kernel_metrics k;
```

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,515 @@
...
+  log_hw_ctx(const xrt_core::hwctx_handle* handle, const xrt::uuid& uuid)
+  {
+    hw_ctx_vec.emplace_back(hw_ctx_metrics{handle, uuid, {}, {}});
+  }
+};
+
+using metrics_map = std::map<std::thread::id, std::map<device_id, device_metrics>>;
+using device_metrics_map = std::map<device_id, device_metrics>;
+// Global map to store usage metrics of all threads
+std::shared_ptr<metrics_map> usage_metrics_map = std::make_shared<metrics_map>();
```
stsoe, line 139: Definitely external linkage of variables in a named namespace.  @keryell this is why I always stick with static even in the unnamed namespace.  Here we for sure need static to avoid external linkage.
```suggestion
static std::shared_ptr<metrics_map> usage_metrics_map = std::make_shared<metrics_map>();
```

keryell: DRY
```c++
static auto usage_metrics_map = std::make_shared<metrics_map>();
```
Could be an anonymous namespace instead. :-)

stsoe: Of course `auto`.    Yes, it could be in anonymous but to what advantage?

keryell: Some people think this is good for some reasons https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnon-exported-entities

keryell: I have not followed the discussion behind this in the ISO C++ committee but I guess `static` is a very overloaded C keyword with a lot of different things.
In C++, anonymous namespace is here to make very clear the meaning of the programmer instead of writing `static` where a reader has to think about what is *this* `static` means here.

stsoe: I agree with that.  I always use unnamed namespace for implementation details and only the named one for exported (per header) content.  I kind of missed that in this PR.  @rbramand-xilinx I think we should move everything in usage_metrics.cpp into the unamed namespace with the exception of the `get_usage_metrics_logger()` of course which the only external API.

stsoe: > In C++, anonymous namespace is here to make very clear the meaning of the programmer instead of writing `static` where a reader has to think about what is _this_ `static` means here.

Completely disagree.   You may have to search long and hard to realize that the definition is within an unnamed namespace and indeed has desired internal linkage.   Personally I will continue to use static even within unnamed, unless you tell me it has side effects beyond not needed.

keryell: I understand you point here.
At the end this is just a coding guideline among others and why not using yours in your project. :-)

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,522 @@
...
+    thread_array.add_child("thread " + std::to_string(t_count), dev_array);
+    t_count++;
+  }
+
+  print_json(thread_array);
+}
+
+} // namespace
+
+namespace xrt_core::usage_metrics {
```
stsoe, line 285: Everything below can stay in unnamed namespace except for the entry function `get_usage_metrics_logger()` at line 514.

stsoe (approved):
Thank you for working through all the changes. Looks good now.  Reads well IMO.

stsoe:
retest this please

chvamshi-xilinx:
retest this please

chvamshi-xilinx (approved):

manikandan-xilinx:
retest this please. (PACT board test is failing repeatedly. So witching back from it )

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,516 @@
...
+};
+
+usage_metrics_logger::
+usage_metrics_logger() : map_ptr(usage_metrics_map)
+{
+  thread_count++;
+}
+
+usage_metrics_logger::
+~usage_metrics_logger()
```
stsoe, line 329: We need try / catch to resolve 
- https://scan9.scan.coverity.com/reports.htm#v36614/p13446/fileInstanceId=123592600&defectInstanceId=9463167&mergedDefectId=330809
- https://scan9.scan.coverity.com/reports.htm#v36614/p13446/fileInstanceId=123592600&defectInstanceId=9463171&mergedDefectId=330815
- https://scan9.scan.coverity.com/reports.htm#v36614/p13446/fileInstanceId=123592600&defectInstanceId=9463146&mergedDefectId=330816

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,516 @@
...
+usage_metrics_logger::
+log_device_info(const xrt_core::device* dev)
+{
+  auto dev_id = dev->get_device_id();
+  if (!get_device_metrics(m_dev_map, dev_id)) {
+    // initialize map with this device index
+    m_dev_map[dev_id] = {};
+    try {
+      auto bdf = xrt_core::query::pcie_bdf::to_string(xrt_core::device_query<xrt_core::query::pcie_bdf>(dev));
+      m_dev_map[dev_id].bdf = bdf;
```
stsoe, line 356: std::move(bdf)
https://scan9.scan.coverity.com/reports.htm#v36614/p13446/fileInstanceId=123592600&defectInstanceId=9463145&mergedDefectId=330811

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,516 @@
...
+#ifdef _WIN32
+# pragma warning ( disable : 4996 )
+#endif
+
+namespace bpt = boost::property_tree;
+
+namespace {
+// global variables
+static std::mutex m;
+static uint32_t thread_count;
```
stsoe, line 37: static std::atomic<uint32_t> thread_count {0};
https://scan9.scan.coverity.com/reports.htm#v36614/p13446/fileInstanceId=123592600&defectInstanceId=9463130&mergedDefectId=330812

src/runtime_src/core/common/usage_metrics.cpp
```diff
@@ -0,0 +1,516 @@
...
+usage_metrics_logger() : map_ptr(usage_metrics_map)
+{
+  thread_count++;
+}
+
+usage_metrics_logger::
+~usage_metrics_logger()
+{
+  {
+    std::lock_guard<std::mutex> lk(m);
```
stsoe, line 332: Move before 337 now that thread_count is atomic

stsoe (commented):
Hi @rbramand-xilinx 
Please check https://scan9.scan.coverity.com/reports.htm#v36614/p13446/fileInstanceId=123592600&defectInstanceId=9463167&mergedDefectId=330809
Comments added

PR #7789, opened 2023-11-08
Add inline specifier to xrt::ini::set(const string&, unsigned int)

melko:
Defining a function in a header file causes a multiple definition linker error when the header in included in multiple files which are to be linked together.

Mark the function as inline to avoid this issue.

### DIFF
--- a/src/runtime_src/core/include/experimental/xrt_ini.h
+++ b/src/runtime_src/core/include/experimental/xrt_ini.h
@@ -56,7 +56,7 @@ set(const std::string& key, const std::string& value);
  *
  * Throws if key value cannot be changed.
  */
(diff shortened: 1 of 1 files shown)


stsoe:
Yes, indeed.  Thank you @melko

stsoe:
ok to test

stsoe (approved):

melko:
Is it possible to see the build failure log? The details link points to a private url I guess

stsoe:
@melko This is an internal CI error.  I will make sure to shepherd this PR through the processes and get it merged.  Thanks again!

stsoe:
retest this please

PR #7791, opened 2023-11-09
Adding a flag o and m

manikandan-xilinx:
Adding a flag o and m for unzipping https://github.com/Xilinx/ERT-BSP/raw/main/BSPs/sched_bsp.tar.bz2 files inside container during the build stages

The above flag is needed to perform cleanup stage in the pipeline once the build is completed. without "o" and "m" flag the ownership of the files and time are preserved which is "jpocock" and  time of file created is 2 years back and when unzipping it's user is same as jpocock and hence we could not delete those files in the pipeline after the build facing permission issue


manikandan-xilinx:
@stsoe If possible could you pls review this

stsoe (approved):

maxzhen (approved):

Issue #10, opened 2018-06-21
Add AWS to CMake Build Lists


rradjabi:
Build AWS HAL by adding to the CMake Lists.

PR #7795, opened 2023-11-10
VITIS-9991 - Removing boost::any

haghilin:
Problem solved by the commit:
The boost::any removed from the XRT code.

N/A

None

System Configuration
  OS Name              : Linux
  Release              : 6.5.0-rc1+
  Version              : #10 SMP PREEMPT_DYNAMIC Wed Aug 16 22:28:12 PDT 2023
  Machine              : x86_64
  CPU Cores            : 16
  Memory               : 63990 MB
  Distribution         : Ubuntu 22.04 LTS
  GLIBC                : 2.35
  Model                : Precision 5820 Tower
  BIOS vendor          : Dell Inc.
  BIOS version         : 2.0.2

XRT
  Version              : 2.17.0
  Branch               : master4
  Hash                 : cc9246d122580380d58c5f9c3c2a0f35758c4bb8
  Hash Date            : 2023-11-10 10:17:40
  XOCL                 : 2.16.0, 893580509d3f7b0f1fab5a0da6210c709529a740
  XCLMGMT              : 2.16.0, 893580509d3f7b0f1fab5a0da6210c709529a740

Devices present
BDF             :  Shell                            Logic UUID                            Device ID       Device Ready* 
-------------------------------------------------------------------------------------------------------------------------
[0000:04:00.1]  :  xilinx_u55c_gen3x16_xdma_base_3  97088961-FEAE-DA91-52A2-1D9DFD63CCEF  user(inst=129)  Yes

### DIFF
--- a/src/runtime_src/core/common/device.h
+++ b/src/runtime_src/core/common/device.h
@@ -17,12 +17,12 @@
 #include "core/include/xrt.h"
 #include "core/include/experimental/xrt_xclbin.h"
 
+#include <any>
 #include <cstdint>
 #include <vector>
 #include <string>
 #include <map>
 #include <memory>
-#include <boost/any.hpp>
 #include <boost/property_tree/ptree.hpp>
 #include <boost/optional/optional.hpp>
 
@@ -204,10 +204,10 @@ public:
    * query() - Query the device for specific property
    *
    * @QueryRequestType: Template parameter identifying a specific query request
-   * Return: QueryRequestType::result_type value wrapped as boost::any.
+   * Return: QueryRequestType::result_type value wrapped as std::any.
    */
   template <typename QueryRequestType>
-  boost::any
+  std::any
   query() const
   {
     auto& qr = lookup_query(QueryRequestType::key);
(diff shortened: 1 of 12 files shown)


stsoe (changes requested):
Hi @haghilin Unfortunately we cannot change xrt::device and remove the boost::any deprecated interface.   We need to continue to support boost::any at the xrt::device interface level else we break existing binaries in the field.   If you check the comments carefully and trace the code, you will see that the std::any is used for all new code but a few functions are left to preserve binary compatibility (ABI).  So leave detail/any.h, xrt_device.h, xrt_device.cpp as is.   For all new compiled host code it will end up using std::any.   

Only change the boost::any -> std::any in the internal implementation code.

haghilin:
> Hi @haghilin Unfortunately we cannot change xrt::device and remove the boost::any deprecated interface. We need to continue to support boost::any at the xrt::device interface level else we break existing binaries in the field. If you check the comments carefully and trace the code, you will see that the std::any is used for all new code but a few functions are left to preserve binary compatibility (ABI). So leave detail/any.h, xrt_device.h, xrt_device.cpp as is. For all new compiled host code it will end up using std::any.
> 
> Only change the boost::any -> std::any in the internal implementation code.

haghilin:
Thanks @stsoe, I removed the changes for xrt::device.

stsoe:
retest this please

stsoe (approved):
@haghilin This looks good.  Thanks.  The changes in XRT will break any project that use XRT as submodule and at the same time implements the query infrastructure.   I think we can handle that, but immediately after merging this PR we should update the internal projects.

PR #7796, opened 2023-11-11
Add hybrid access to xrt::ext:bo for cross-adapter allocations

stsoe:
Support allocating buffer objects that are shared between hybrid devices. On Windows / MCDM this is a cross-adapter allocation.

Add new access mode `xrt::ext::bo::access_mode::hybrid` and encode in bo flags passed to shim for allocatng BO.

This PR changes the encoded BO flags for access extension removing old XRT_BO_ACCESS_EXPORTED and instead keeps a 1:1 mapping from xrt::ext::bo:access::mode to these access flags.

It is an error to bitwise or the access mode local, shared, process, and hybrid and an exception is thrown if the access mode is not exactly one of these 4 values.

This PR is a pre-requisite for MCDM / UMD changes to support creating cross adapter allocations.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -1506,41 +1506,43 @@ copy(const bo& src, size_t sz, size_t src_offset, size_t dst_offset)
 ////////////////////////////////////////////////////////////////
 namespace xrt::ext {
 
-static bool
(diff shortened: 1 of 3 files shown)


stsoe:
retest this please - apu

manikandan-xilinx:
retest this please. (PACT board test is failing repeatedly. So witching back from it )

PR #7798, opened 2023-11-13
Changes to XDP Plugin metadata parsing

nishraptor:
Created a new structure to allow support for both compiler_report and aie_control_config.json file types.

Implementations are now under a base filetype class. The correct filetype is created at runtime and then used to extract the correct fields in the metadata.

Changes are also on x86 and edge devices

Tested on client devices


jvillarre (approved):

IshitaGhosh (commented):
Changes look good.

PR #7800, opened 2023-11-14
Add hybrid access to xrt::fence for cross-adapter sharing

stsoe:
Support allocating fence objects that are shared between hybrid devices. On Windows / MCDM this a cross-adapter fence.

Add new access mode `xrt::fence::access_mode::hybrid`.


stsoe:
retest this please - changed code is not used by example

PR #7802, opened 2023-11-15
AIE Trace implementation for client devices

nishraptor:
Added support for AIE Trace on Client Devices.

Created a new implementation specific to client devices. Added hooks in xdp/profile.cpp for the aie_trace plugin to be loaded when the correct settings are in the xrt.ini

Could break aie_trace on edge platforms due to changes in the plugin

Tested on client devices for core, shim, and memtile trace.


stsoe (approved):

jvillarre (approved):

pgschuey (approved):
Changes look good. The impact on edge was previously discussed and appropriately reduced.

PR #7811, opened 2023-11-21
Added temporary fix to write ap_start twice on trace S2MM cores (workaround for CR-1181692)

pgschuey:
Trace S2MM core does not start after application of a single ap_start

Apply two ap_start calls to AP control register. Eventually, the HW flow will be fixed, but trace needs to be up and running.

Old xclbins will receive a second unnecessary ap_start but should still work

Tested old and new xclbins on vek280

N/A


pgschuey:
Retest this please

jvillarre (approved):

Issue #6783, opened 2022-06-19
CMAKE_INSTALL_PREFIX is not honored


bryanloz-xilinx:
The XRT package is not relocatable.

https://github.com/Xilinx/XRT/blob/master/src/CMakeLists.txt#L69

/opt/xilinx is hardcoded in 56 different files within XRT.

bryanloz-xilinx:
@gbuildx @NaAl-Xlnx 
Can you please take a look at the build pipelines and make sure that `/opt/xilinx` or `/opt/xilinx/xrt` is not hardcoded anywhere in the build pipeline or in any automated tests?

In the near future we would like to install XRT to `/usr/` by default.

qberthet:
@stsoe what is the status of this issue? In the discussion under #6779 you mention that you want to address this issue but "will not promise a quick fix".

I would like to weight in that XRT not being relocatable is a problem for the large project I work on. Knowing when we could expect a fix for this could be of great help. Thanks!

qberthet:
Now that  #7835 is merged, shouldn't this issue be closed ? Or is there still something else to it?
Thanks!

stsoe:
Thanks @qberthet.  The PR was missing meta comment for closing the issue.

PR #7825, opened 2023-11-30
Enabling setting XRT install prefix to custom location on Linux build…

zhangchiming:
Problem solved by the commit
Enabled setting XRT install prefix to custom location on Linux build.

https://github.com/Xilinx/XRT/issues/6783

Modified corresponding CMake files, Makefiles and build scripts.

Low

Built XRT on Ubuntu 22.04.

Documentation impact (if any)
NA

### DIFF
--- a/build/board.sh
+++ b/build/board.sh
@@ -10,9 +10,6 @@ XRTBUILD=$(readlink -f $(dirname ${BASH_SOURCE[0]}))
 # Environment
 ################################################################
 xrt=$XRTBUILD/Release/opt/xilinx/xrt
-if [ x"${XRT_INSTALL_PREFIX}" != "x" ]; then
-    xrt="$XRTBUILD/Release"$XRT_INSTALL_PREFIX"/xrt"
-fi
 board=vcu1525
 keep=1
 sync=0
--- a/build/board.sh
+++ b/build/board.sh
@@ -10,7 +10,9 @@ XRTBUILD=$(readlink -f $(dirname ${BASH_SOURCE[0]}))
 # Environment
 ################################################################
 xrt=$XRTBUILD/Release/opt/xilinx/xrt
-
+if [ x"${XRT_INSTALL_PREFIX}" != "x" ]; then
+    xrt="$XRTBUILD/Release"$XRT_INSTALL_PREFIX"/xrt"
+fi
 board=vcu1525
 keep=1
 sync=0
--- a/build/build.sh
+++ b/build/build.sh
@@ -240,8 +240,7 @@ edge_dir=${EDGE_DIR:-Edge}
 cmake_flags+=" -DXRT_ENABLE_WERROR=$werror"
 
 # set CMAKE_INSTALL_PREFIX
-export XRT_INSTALL_PREFIX=$xrt_install_prefix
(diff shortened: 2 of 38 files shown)


build/board.sh
```diff
@@ -10,7 +10,9 @@ XRTBUILD=$(readlink -f $(dirname ${BASH_SOURCE[0]}))
 # Environment
 ################################################################
 xrt=$XRTBUILD/Release/opt/xilinx/xrt
-
+if [ x"${XRT_INSTALL_PREFIX}" != "x" ]; then
+    xrt="$XRTBUILD/Release"$XRT_INSTALL_PREFIX"/xrt"
+fi
```
stsoe, line 15: Where does XRT_INSTALL_PREFIX come from?  I don't see it is this script?  
board.sh is a very old legacy script, I don't think it needs to support anything but the default install prefix, I say we revert the changes in this one file.

build/build.sh
```diff
@@ -232,6 +239,10 @@ edge_dir=${EDGE_DIR:-Edge}
 # Disable with '-disable-werror' option.
 cmake_flags+=" -DXRT_ENABLE_WERROR=$werror"
 
+# set CMAKE_INSTALL_PREFIX
+export XRT_INSTALL_PREFIX=$xrt_install_prefix
```
stsoe, line 243: Why export?  This script is not sourced, it is used only for invoking the build.  Isn't it sufficient to set CMAKE_INSTALL_PREFIX to xrt_install_prefix?  Do we need the XRT_INSTALL_PREFIX variable at all?

stsoe (commented):
Thanks for working on relocation, this is long overdue.
I have a fundamental question regarding env XRT_INSTALL_PREFIX.  I didn't get too far reviewing all changes, I need to understand the reason for this external ENV var first.

build/board.sh
zhangchiming: Ok. I will remove it.

build/build.sh
zhangchiming: I thought it would be simpler to have a separate variable XRT_INSTALL_PREFIX for XRT install location. I will modify it to use CMAKE_INSTALL_PREFIX only.

abidihaider:
Hi @zhangchiming, 

I just wanted to reach out and ask about this pull request. We have been trying to get XRT in one of research framework at CERN and this hardcoded path has been a huge blocker on our end. So we were pretty excited to see that you were able to figure this out? However, I see this request being closed. Are there plans from your side to continue and still push this fix in?

abidihaider:
> > Hi @zhangchiming,
> > I just wanted to reach out and ask about this pull request. We have been trying to get XRT in one of research framework at CERN and this hardcoded path has been a huge blocker on our end. So we were pretty excited to see that you were able to figure this out? However, I see this request being closed. Are there plans from your side to continue and still push this fix in?
> 
> I run into some problem with pull request DCO verification because my email was originally set to private in github profile. "[144274849+zhangchiming@users.noreply.github.com](mailto:144274849+zhangchiming@users.noreply.github.com)" is expected sign off email even after I set the email to public. I will submit another pull request soon.

Awesome! Thank you so so much :)

PR #7831, opened 2023-12-05
Create file with unique name to dump usage metrics

rbramand-xilinx:
When dumping usage metrics into JSON file, the file name is not unique. Including pid in the file name to create unique file as the log is dumped per process.

PR - https://github.com/Xilinx/XRT/pull/7788 introduced the bug
This bug was detected when debugging CR - https://jira.xilinx.com/browse/CR-1181804. Files names matched for processes exiting at same time and there was issue of files being overwritten.

Problem was solved by adding pid in file name as log is dumped per process and it will be unique

Low

Tested with the application that is attached with CR and things are working as expected

NA

### DIFF
--- a/src/runtime_src/core/common/utils.cpp
+++ b/src/runtime_src/core/common/utils.cpp
@@ -28,6 +28,13 @@
 #include <string>
 #include <boost/algorithm/string.hpp>
 
+#ifdef __linux__
+#include <unistd.h>
+#endif
+#ifdef _WIN32
+# include <process.h>
+#endif
+
 namespace {
 
 inline unsigned int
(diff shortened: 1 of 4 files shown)


chvamshi-xilinx (approved):

src/runtime_src/core/common/utils.cpp
```diff
@@ -28,6 +28,13 @@
 #include <string>
 #include <boost/algorithm/string.hpp>
 
+#ifdef __linux__
+#include <unistd.h>
```
stsoe, line 32:
```suggestion
# include <unistd.h>
```

stsoe (approved):
Minor

PR #7836, opened 2023-12-06
Add missing apt-get update to have an up-to-date package list

keryell:
Also remove random TAB and spaces.
This file would require a serious beautifier.

### DIFF
--- a/src/runtime_src/tools/scripts/xrtdeps.sh
+++ b/src/runtime_src/tools/scripts/xrtdeps.sh
@@ -475,6 +475,8 @@ validate()
 prep_ubuntu()
 {
     echo "Preparing ubuntu ..."
+    # Update the list of available packages
+    apt-get update
 }
 
 prep_centos7()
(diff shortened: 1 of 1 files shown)


src/runtime_src/tools/scripts/xrtdeps.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -vx
```
stsoe, line 1: Really?

stsoe (changes requested):

stsoe:
@keryell Please `git commit -s --amend --no-edit`

src/runtime_src/tools/scripts/xrtdeps.sh
keryell: Ouch, I pushed my debug code. :-(

stsoe:
> Actually in Emacs/magit I just need to press 'r' and 'm'. I tend to forget about all the git CLI...

I was implying you need to sign your PR.

keryell:
> > @keryell Please `git commit -s --amend --no-edit`
> 
> Actually in Emacs/magit I just need to press 'r' and 'm'. I tend to forget about all the git CLI...

Oh, this is a new policy? By signing I guess this is just adding my e-mail, no cryptographic signature?

stsoe:
ok to test

stsoe:
add to whitelist

PR #7844, opened 2023-12-11
Add warning exception for early return

dbenusov-xilinx:
XRT-IPU does not compile with Visual Studio 2022 with the the changes from https://github.com/Xilinx/XRT/pull/7802. We really need to add VS2022 as a compilation check to pipeline.

Introduced in https://github.com/Xilinx/XRT/pull/7802. I attempted to compile with Visual Studio 2022 and encountered an early return warning.

Added a warning exception for early returns with a TODO comment.

None.

With this change XRT-IPU will compile.

None.


jvillarre (approved):
The right fix will be to change the finalFlushAIEDevice function so the return only appears in the right context.  That will come in a future pull request, but I'll merge this to temporarily unblock any pipelines.

PR #7855, opened 2023-12-18
EDGEML-6881 Random test failure when looking for driver store path

stsoe:
Adjust copying of wchar[] array into std::wstring accounting for bytes vs wchar.

This bug was introduced in #7679 and is present in MCDM drops starting M6.

Code queries windows adpater for driver store path.  The driver store path is returned as array of wchar but was copied as bytes, resulting in copying from unallocated memory.


stsoe:
Merging without waiting for CI to finish.  The change passes all build steps, but is not exercised by any testing.

PR #7857, opened 2023-12-19
Skeleton directory and build infrastructure for HIP bindings

stsoe:
Very preliminary infra for HIP bindings.
Build on linux with `build.sh -hip [other options]`

### DIFF
--- a/src/runtime_src/hip/core/device.cpp
+++ b/src/runtime_src/hip/core/device.cpp
@@ -1,0 +1,9 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2023 Advanced Micro Device, Inc. All rights reserved.
+#include "device.h"
+
+namespace xrthip_core {
+
+// Implementation  
+  
+}
--- a/src/runtime_src/hip/core/device.cpp
+++ b/src/runtime_src/hip/core/device.cpp
@@ -2,7 +2,7 @@
 // Copyright (C) 2023 Advanced Micro Device, Inc. All rights reserved.
 #include "device.h"
 
-namespace xrthip_core {
+namespace xrt::core::hip {
 
 // Implementation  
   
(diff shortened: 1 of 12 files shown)


src/runtime_src/hip/core/device.cpp
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2023 Advanced Micro Device, Inc. All rights reserved.
+#include "device.h"
+
+namespace xrthip_core {
```
keryell, line 5: Something more C++17 friendly and clearer like `xrt::hip::core`?

stsoe: @keryell .  I thought hard and long about this.  I don't want xrt::  because that namespace is reserved for anything publicly visible from XRT, e.g. our public APIs.   This is an internal namespace, the only thing that will eventually be exported are the APIs declarations, if any of these are C++ they will be in namespace xrt::hip.   FYI,  in xrt_coreutil we use namespace xrt_core for exactly the same reason.  So, believe or not, like it or not, this is actually thought through, :-)

keryell: > @stsoe So, believe or not, like it or not, this is actually thought through, :-)

Of course I think this is not some random writing. ;-) Thanks for the context.

Cf Doxygen `\namespace to document a namespace. ` so this is not lost? :-)

stsoe: @keryell Maybe I should relax the xrt:: constraint and clean up this nonsensical use of namespaces.  Replace xrt_core with xrt::core, replace xrthip_core with xrt::hip::core.  Continue to use xrt:: for publicly exported XRT native C++ APIs, use xrt::hip for publicly exported XRT C++ HIP APIs if any.   At the same time I could use xrt::opencl::core for the opencl bindings which for historical reasons are in yet another obscure namespace `xocl` . Is that a more intuitive convention?

stsoe: Actually, maybe it should be xrt::core::hip, xrt::core::opencl which would make it much easier for the bindings implementation to reference xrt::core which will be used heavily?

PR #7841, opened 2023-12-08
Fixing XRT issue for Ubuntu 23.10

haghilin:
Problem solved by the commit:
There was compilation issues on Ubuntu 23.10 which was reported in issue https://github.com/Xilinx/XRT/issues/7806.

Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered
issue https://github.com/Xilinx/XRT/issues/7806

How problem was solved, alternative solutions (if any) and why they were rejected
Risks (if any) associated the changes in the commit
What has been tested and how, request additional testing if necessary

PR #7858, opened 2023-12-19
Fix missing include

qberthet:
Probably in the trail of #7841 , it seems that there is still a missing `#include <cstdint>`.

While compiling XRT master on Alma9 with GCC 13.1.0 and CMake 3.27.5 (I know, unsupported os), I encountered the following error:

```
[ 56%] Building CXX object runtime_src/core/tools/xbmgmt2/CMakeFiles/xbmgmt2.dir/__/common/reports/ReportAiePartitions.cpp.o
In file included from /home/qberthet/devel/atlasexternals/XRT/src/runtime_src/core/tools/common/Process.cpp:37:
/home/qberthet/devel/atlasexternals/XRT/src/runtime_src/core/tools/common/EscapeCodes.h:26:22: error: expected ')' before '_color'
   26 |       fgcolor(uint8_t _color) : m_color(_color) {};
      |              ~       ^~~~~~~
      |                      )
/home/qberthet/devel/atlasexternals/XRT/src/runtime_src/core/tools/common/EscapeCodes.h:32:6: error: 'uint8_t' does not name a type
   32 |      uint8_t m_color;
      |      ^~~~~~~
```

Added the missing include in `EscapeCodes.h` and compilation succeed.

### DIFF
--- a/src/runtime_src/core/tools/common/EscapeCodes.h
+++ b/src/runtime_src/core/tools/common/EscapeCodes.h
@@ -16,6 +16,7 @@
 
 // Include files
 // Please keep these to the bare minimum
+#include <cstdint>
 #include <string>
 
 namespace EscapeCodes {


chvamshi-xilinx:
ok to test

stsoe:
Thank you @qberthet , please sign off the commit.

PR #7861, opened 2023-12-20
Update dependency installation, build and packaging scripts for AlmaLinux9

qberthet:
While not a supported OS, XRT can be built and used on AlmaLinux 9, but there is some checks for specific linux flavors which forbid to run dependency installation script and packaging (Build is mostly ok, except for location of pybind11).

With the proposed modifications, dependency installation script, build and packaging works on AlmaLinux9

Tested on almalinux:latest docker and a fresh almalinux-minimal native install.

Caveat: on docker image, the `curl-minimal` clash with the installation of `curl` until `--allowerasing` is used during yum install. As this change would be more invasive and probably controversial, it have been left out and using `yum install curl --allowerasing` before running the `xrtdeps.sh` script is an easy workaround.

This is not an attempt to add AlmaLinux 9 to the supported OS lists, thus no documentation modification is required.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -12,7 +12,7 @@ CMAKE_MAJOR_VERSION=`cmake --version | head -n 1 | awk '{print $3}' |awk -F. '{p
 CPU=`uname -m`
 
 if [[ $CMAKE_MAJOR_VERSION != 3 ]]; then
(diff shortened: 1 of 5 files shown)


stsoe (approved):

stsoe:
ok to test

PR #7875, opened 2024-01-05
CR-1181687 Fix hwctx_handle::create_hw_queue() to return shared queue

stsoe:
Share the ownership of xrt_core::hwqueue_handle with shim.

A hwqueue is managed by hwctx (today 1:1), and used by coreutil XRT for submitting commands for execution and more. XRT shim may itself use the hwqueue for miscellaneous commands and as such the underlying queue is actually owned by shim (hwctx_handle) but shared with coreutil.

This PR makes it explicit that the hwqueue ownership is with the hwctx_handle. XRT coreutil no longer creates a hwqueue but instead gets a hwqueue_handle from the hwctx_handle (`hwctx::handle::get_hw_queue()`).   XRT coreutil does not participate in the lifetime of the hwqueue_handle obtained from a hwctx_handle it is the responsibility of XRT coreutil implementation to ensure that lifetime of a hwctx exceeds the use of the hwqueue_handle that was obtained from the hwctx_handle.

Prior to this change there were strange hacks at shim level to return a unique_ptr to something that is actually shared.  This strange logic can now be changed.

### DIFF
--- a/src/runtime_src/core/common/api/hw_queue.cpp
+++ b/src/runtime_src/core/common/api/hw_queue.cpp
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright (C) 2021-2022 Xilinx, Inc. All rights reserved.
-// Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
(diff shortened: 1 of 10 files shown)


maxzhen (approved):

stsoe:
Merge based on Jenkins passing.

PR #7870, opened 2024-01-04
VITIS-11043 Refactor XRT to cache created devices

dbenusov:
#### Problem solved by the commit
https://jira.xilinx.com/browse/VITIS-11043
Each time XRT creates a device reference a new device wrapper is created. This is not correct, only one device should be created and that referenced cached.

#### Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered
Prevent XRT from constantly creating new xrt_core::device handles.

#### How problem was solved, alternative solutions (if any) and why they were rejected
Fixed by intercepting the create device call and storing the returned shared pointer. If a request is made for the same device id, the original referenced is returned.

#### Risks (if any) associated the changes in the commit
None. Should be an improvement in performance hopefully :) 

#### What has been tested and how, request additional testing if necessary
Ubuntu 22.04 U55c
```
dbenusov@xsjdbenusov50:/proj/rdi/staff/dbenusov$ xbutil examine -d -r electrical
WARNING: Unexpected xocl version (2.16.0) was found. Expected 2.17.0, to match XRT tools.

-------------------------------------------------
[0000:04:00.1] : xilinx_u55c_gen3x16_xdma_base_3
-------------------------------------------------
Electrical
  Max Power              : 225 Watts
  Power                  : 21.445936 Watts
  Power Warning          : false

  Power Rails            :   Voltage  Current
  ---------------------------------------------
  12 Volts Auxillary     :  12.256 V  0.481 A
  12 Volts PCI Express   :  12.240 V  0.985 A
  3.3 Volts PCI Express  :   3.360 V  1.040 A
  3.3 Volts Auxillary    :   0.000 V  0.000 A
  Internal FPGA Vcc      :   0.852 V  6.100 A
  Internal FPGA Vcc IO   :   0.854 V  3.800 A
  DDR Vpp Bottom         :   0.000 V  0.000 A
  DDR Vpp Top            :   0.000 V  0.000 A
  5.5 Volts System       :   5.006 V  0.000 A
  Vcc 1.2 Volts Top      :   0.000 V  0.000 A
  Vcc 1.2 Volts Bottom   :   0.000 V  0.000 A
  1.8 Volts Top          :   1.813 V  0.000 A
  0.9 Volts Vcc          :   0.898 V  0.000 A
  12 Volts SW            :   0.000 V  0.000 A
  Mgt Vtt                :   1.200 V  0.000 A
  3.3 Volts Vcc          :   3.354 V  0.000 A
[...]

PR #7890, opened 2024-01-17
fixing crash in xbmgmt/xbutil examine

chvamshi-xilinx:
xbutil/xbmgmt examine is crashing with latest XRT. This happens only for golden images case.

https://github.com/Xilinx/XRT/pull/7870 introduced this bug. Sprite caught this issue

Fixed the newly added code. Throwing proper error message

None

Verified the overall flow

None


dbenusov-xilinx:
Thanks for fixing this. It looks good to me, though  I would now probably like to replace the vector with a map so it is easier to handle these types of cases.

PR #7893, opened 2024-01-18
Changed all names from XDP_MINIMAL_BUILD to XDP_CLIENT_BUILD

nishraptor:
Renamed XDP_MINIMAL_BUILD to XDP_CLIENT_BUILD
Naming update for XDP plugins

Tested when building for client to verify dll's were generated correctly


stsoe (approved):
This is a breaking change for XRT-MCDM and XRT-IPU.  I suggest that immediately after this PR is merged, you create a PR to update XRT on both XRT-MCDM and XRT-IPU.  Thanks!

dayeh-xilinx:
retest this please.

nishraptor:
@stsoe There were two XDP_MINIMAL_BUILD variables. The first is a CMake Variable, set by XRT-IPU, and the second is a compile-time variable only set if the CMake variable is on. I've only changed the  compile-time variable to XDP_CLIENT_BUILD, so the CMake variable should be still called XDP_MINIMAL_BUILD. I don't believe anything should break in the IPU branches from this change.

nishraptor:
@stsoe After conversation with @jvillarre, we have decided to rename the CMake Variable as well. I will create pull requests in XRT-IPU and XRT-MCDM as well to change the CMake variable name and update the xrt commit. This will have to be done after this PR is merged.

dayeh-xilinx:
retest this please.

jvillarre (approved):

PR #7895, opened 2024-01-19
Updated Struct Data types in XDP codebase

nishraptor:
Updated struct data types to use proper integer sizes in XDP, and updated all methods using those structs and struct parameters to accept the proper data types. This mainly involves changing uint16_t values to uint8_t, which is smaller than all row/column information.

Since AIE drivers use uint8_t for row and column information, we were static_casting everywhere in order to convert the row and column values from uint16_t to uint8_t. 

Changed relevant struct sizes to uint8_t from uint16_t in tile_type, TraceGMIO, and AIECounter.

Verified compilation and execution of the xdp plugins on client, and compilation on x86.


jvillarre (approved):

dayeh-xilinx:
retest this please.

pgschuey (approved):
Everything looks so much cleaner

PR #7901, opened 2024-01-23
Removed support for u50lv, u55n and u280 platforms in XRT

ManojTakasi:
[VITIS-10325](https://jira.xilinx.com/browse/VITIS-10325) u50lv, u55n and u280 are not supported in 2024.1. So removed support for these platorms in XRT.

NA

Updated devices.h file

NA

NA

Yes

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/devices.h
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/devices.h
@@ -2041,14 +2041,6 @@ struct xocl_subdev_map {
 		.subdev_num = ARRAY_SIZE(USER_RES_DSA52),		\
 	}
 
-#define	XOCL_BOARD_USER_DSA52_U280					\
-	(struct xocl_board_private){					\
-		.flags		= 0,					\
-		.subdev_info	= USER_RES_DSA52,			\
-		.subdev_num = ARRAY_SIZE(USER_RES_DSA52),		\
-		.p2p_bar_sz = 64,					\
-	}
-
 #define	XOCL_BOARD_USER_SMARTN						\
 	(struct xocl_board_private){					\
 		.flags		= XOCL_DSAFLAG_SMARTN,			\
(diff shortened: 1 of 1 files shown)


AShivangi:
@ManojTakasi, does this mean we can remove support for these platforms for utilities as well?
I believe we have some platform specific code for these platforms

saifuddin-xilinx (approved):

ManojTakasi:
Hi @AShivangi, yes we can remove support for these platforms in utilities also.

chvamshi-xilinx (approved):

houlz0507 (approved):

keryell:
I am puzzled by such a PR.
Imagine I have a machine with both supported and unsupported cards. I need a modern XRT to run with the modern card but it would break running with the old card and its old `xlcbin`???
Not being able to compile new applications is a thing, no longer able to run old application is another thing.
Are we going along the AMD GPU support rant like https://github.com/ROCm/ROCm/issues/2308 and others?

goekce:
So much hardware turn to junk because due to missing updates. I would expect from an industrial product more than couple of years of support. Sad to see this. Would be great if pros/cons of support for these cards are openly discussed.

Background: I bought a U280 for research exactly two years ago.

maxzhen (approved):

chvamshi-xilinx:
Hi @keryell / @goekce ,
These cards are still supported by XRT 2023.2 branch.  If there are any issues reported, we are going to publish fixes to 2023.2 branch.

keryell:
It seems unlikely for example that XRT 2023.2 will ever support Linux kernels and so on.
But the problem is to support *both* old and latest cards at the same time on the same machine, as GPUs work.
@chvamshi-xilinx how do you think it will work?

chvamshi-xilinx:
> It seems unlikely for example that XRT 2023.2 will ever support Linux kernels and so on. But the problem is to support _both_ old and latest cards at the same time on the same machine, as GPUs work. @chvamshi-xilinx how do you think it will work?

Yeah. Mixing of old and latest cards on same machine is not going to work (you can run them on two separate VMs by passing thru, but that is not ideal).

Let me discuss these cases with our marketing folks and get back to you. adding "Do Not Merge" tag to this PR.

chvamshi-xilinx:
> > It seems unlikely for example that XRT 2023.2 will ever support Linux kernels and so on. But the problem is to support _both_ old and latest cards at the same time on the same machine, as GPUs work. @chvamshi-xilinx how do you think it will work?
> 
> Yeah. Mixing of old and latest cards on same machine is not going to work (you can run them on two separate VMs by passing thru, but that is not ideal).
> 
> Let me discuss these cases with our marketing folks and get back to you. adding "Do Not Merge" tag to this PR.

Merging this PR based on offline discussions.

- For continuous support of these platforms, we will backport changes to 2023.2
- Mixing of old and latest cards on same machine is not going to work. We need to run them on two separate VMs by passing thru.

Issue #7911, opened 2024-01-26
Potential buffer overflow in error report


stellaraccident:
Getting pages of these for different source files:

```
[481/496] Building CXX object src/runtime_src/core/pcie/linux/CMakeFiles/core_pcielinux_objects.dir/device_linux.cpp.o
In file included from /home/stella/src/iree-amd-aie/third_party/XRT/src/runtime_src/core/pcie/linux/device_linux.cpp:8:
/home/stella/src/iree-amd-aie/third_party/XRT/src/runtime_src/core/common/query_requests.h:942:72: warning: adding 'const xrt_core::query::device_status::result_type' (aka 'const unsigned int') to a string does not append to the string [-Wstring-plus-int]
        throw xrt_core::system_error(EINVAL, "Invalid device status: " + status);
                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
/home/stella/src/iree-amd-aie/third_party/XRT/src/runtime_src/core/common/query_requests.h:942:72: note: use array indexing to silence this warning
        throw xrt_core::system_error(EINVAL, "Invalid device status: " + status);
                                                                       ^
                                             &                         [       ]
/home/stella/src/iree-amd-aie/third_party/XRT/src/runtime_src/core/common/query_requests.h:3418:77: warning: adding 'const xrt_core::query::performance_mode::result_type' (aka 'const unsigned int') to a string does not append to the string [-Wstring-plus-int]
        throw xrt_core::system_error(EINVAL, "Invalid performance status: " + status);
                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~
/home/stella/src/iree-amd-aie/third_party/XRT/src/runtime_src/core/common/query_requests.h:3418:77: note: use array indexing to silence this warning
        throw xrt_core::system_error(EINVAL, "Invalid performance status: " + status);
                                                                            ^
                                             &                              [       ]
2 warnings generated.
```

I think this is a legit error and should be fixed.

I'm getting something like it on recent gcc and clang. The clang version is 14.0.0. Please don't just silence the warning -- this is a legit bug.

dbenusov-xilinx:
@stellaraccident Thanks for pointing this out! I will update the code to use the parse_status function.

dbenusov-xilinx:
#7916

dbenusov-xilinx:
@stellaraccident Please let me know if the warnings stop with the update in XRT.

PR #7682, opened 2023-08-29
VITIS-9224 Xbutil Enhancements for Power Management

AShivangi:
#### Problem solved by the commit
Plumbing for
- [VITIS-9561](https://jira.xilinx.com/browse/VITIS-9561) Set high, medium and low performance mode
- [VITIS-9562](https://jira.xilinx.com/browse/VITIS-9562) Report current frequencies and performance mode

#### Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered
N/A
#### How problem was solved, alternative solutions (if any) and why they were rejected
Created a new subcommand to set performance and display which mode it is in platform report

#### Risks (if any) associated the changes in the commit
None

#### What has been tested and how, request additional testing if necessary
Tested with simulated values:
```
$ xbutil examine -d b3:00 -r platform

---------------------------------------------------
[0000:b3:00.1] : xilinx_vck5000_gen4x8_qdma_base_2
---------------------------------------------------
Platform
  XSA Name               : xilinx_vck5000_gen4x8_qdma_base_2 
  Logic UUID             : 05DCA096-76CB-730B-8D19-EC1192FBAE3F 
  FPGA Name              :  
  JTAG ID Code           : 0x0 
  DDR Size               : 0 Bytes
  DDR Count              : 0 
  Revision               : 1 
  MFG Date               : 0xcee875 
  Mig Calibrated         : false 
  P2P Status             : not supported 
  Performance Mode       : Default

Clocks
  DATA_CLK (Data)        : 300 MHz
  KERNEL_CLK (Kernel)    : 500 MHz

Mac Addresses            : 00:0A:35:0D:55:8E
                         : 00:0A:35:0D:55:8F
```
```
$ xbutil configure --performance -d b3:00 low

Performance mode is set to low 

```

Negative test:
```
$ xbutil configure --performance -d b3:00 lo
ERROR: Invalid action value: 'lo'

COMMAND: configure

DESCRIPTION: Change performance mode of the device

USAGE:  configure --performance [--help] [-d arg] action

OPTIONS:
  --performance      - Change performance mode of the device
  -d, --device       - The Bus:Device.Function (e.g., 0000:d8:00.0) device of interest
  --help             - Help to use this sub-command


ERROR: : Operation canceled
```

#### Documentation impact (if any)
Added the new command usage to github.io

PR #7916, opened 2024-01-29
Issue-7911 Fix clang issues with query requests

dbenusov-xilinx:
https://github.com/Xilinx/XRT/issues/7911

Warnings are clogging up the compilation window for the Node.Ai team.

Introduced as two separate PRs: #7507 and #7682 

Rather than appending number to a string (which does not work), convert it to a string. 

None,

Compilation success on Ubuntu 22.04.

None.


stsoe (approved):

keryell:
The name of the PR is misleading, since it is not a Clang issue, it is a bug in XRT code.

PR #7917, opened 2024-01-29
CR-1182298: Fixing seg fault on edge when aie_status is enabled

jvillarre:
A race condition involving the ordering of the AIE status polling thread and the destruction of the AIE status object when it held the last local copy of a shared pointer was causing a seg fault in some instances on Edge platforms.

The AIE status plugin creates a thread to periodically poll the AIE as the application runs.  This provides us the capability of detecting when the AIE is hung as well as keeping track of execution over time.

The polling thread was creating an instance of a shared pointer to xrt_core::device every time it dumped to disk.  This lead to a condition at the end of execution where every other shared pointer to the xrt_core::device had already been reclaimed and the thread had the last pointer to the xrt_core::device.  When we joined this thread in our plugin destructor, it triggered the Shim destructor as no other shared pointer existed.  However, this led to problems as we recursively try to flush out any remaining data when the Shim destructor is called and this causes issues if we try to flush as we're in the process of finishing a flush operation.

The issue was discovered through regression testing.

The aie_status plugin module now has a local copy of the shared_ptr to the xrt_core::device.  This way, the Shim will only be destroyed after the aie_status plugin has been destroyed instead of in the middle of the aie_status plugin destructor.  This guarantees that all the threads have already joined and there are no longer any race conditions.

As with all profiling, this plugin instance only exists if the user has actively enabled the profiling feature and it will not affect the execution when aie_status is turned off.

Low risk as this replaces an already existing instance of the shared pointer with a different instance that is destroyed at a time when all the threads have already been joined.

The failing test case has been tested on vck190.

No documentation impact

### DIFF
--- a/src/runtime_src/xdp/profile/plugin/aie_status/aie_status_plugin.cpp
+++ b/src/runtime_src/xdp/profile/plugin/aie_status/aie_status_plugin.cpp
@@ -393,6 +393,8 @@ namespace xdp {
     if (!xrt_core::config::get_aie_status())
       return;
 
+    localDevice = xrt_core::get_userpf_device(handle);
+
     std::array<char, sysfs_max_path_length> pathBuf = {0};
     xclGetDebugIPlayoutPath(handle, pathBuf.data(), (sysfs_max_path_length-1) ) ;
     std::string sysfspath(pathBuf.data());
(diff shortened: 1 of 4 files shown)


pgschuey (approved):
looks good

vipangul (approved):

PR #7919, opened 2024-01-30
Update xbutil validate paths for xclbins

AShivangi:
For MCDM, we can't rename the xclbins back to validate.xclbin if the files are being placed in driverstore. So the <device-id>/validate.xclbin solution doesn't work. For the upcoming release, we need to revert back to doing the device check in xbutil

with current changes, xbutil validate skips saying "xclbin not found"

Checking device id, and based on that, we set xclbin name

none

tested on MCDM phoenix setup:
```
Test 1 [00c3:00:01.1]     : verify
    Description           : Run 'Hello World' test on IPU
    Xclbin                : C:\Windows\System32\DriverStore\FileRepository\ipukmddrv.inf_amd64_27240dcece074c72\validate_phx.xclbin
                            Total duration: '1.4's
                            Average throughput: '7241.1' ops/s
                            Average latency: '138.1' us
```

N/A

### DIFF
--- a/src/runtime_src/core/tools/common/tests/TestDF_bandwidth.cpp
+++ b/src/runtime_src/core/tools/common/tests/TestDF_bandwidth.cpp
@@ -80,10 +80,16 @@ TestDF_bandwidth::run(std::shared_ptr<xrt_core::device> dev)
 {
   boost::property_tree::ptree ptree = get_test_header();
 
+  // workaround: can't rename files when copying to driver store
+  // so need to name the files as _phx and _stx
+  // will revisit this after the current release
   auto device_id = xrt_core::query::pcie_device::to_string(xrt_core::device_query<xrt_core::query::pcie_device>(dev));
-  std::filesystem::path xpath{device_id};
-  xpath /= m_xclbin;
-  ptree.put("xclbin", xpath.string());
+  if (device_id.compare("0x1502") == 0)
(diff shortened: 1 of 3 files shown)


maxzhen:
Merging per request by @AShivangi

stsoe:
I don't think this should have been merged.  It should be possible to rename files during driver install AFAIK.   
The files can be renamed as part of driver install only not in the driver package.  Alas I don't think XRT needs to change.
If true, then this PR should be reverted.

maxzhen:
I was about to ask what should we do on Linux since we have already provided validate.xclbin for both phx and ipu2 platforms. According to @stsoe , it seems that we should be able to make validate.xclbin work and I don't need to change anything on Linux?

stsoe:
> I don't think this should have been merged. It should be possible to rename files during driver install AFAIK.
> The files can be renamed as part of driver install only not in the driver package. Alas I don't think XRT needs to change.
> If true, then this PR should be reverted, @AShivangi

Apparently files cannot be renamed in the Windows driver store during install.  Only files copied out of the driver store to e.g. system32, can be renamed.  Given that the xclbins must be in the driver store on windows, this PR is correct from a windows usage point of view.   Whether the same convention needs to be used on Linux is a different question, I guess for simplicity it should be the same on both Linux and Windows.  @maxzhen

PR #7934, opened 2024-02-08
add bind_at() SHIM API

maxzhen:
When set_arg(BO) is called upon run object, we can call bind_at() SHIM API to notify driver that a BO is now is used by (bond to) the exec buf command BO. This is useful for NPU Linux driver to keep track of BOs used by a particular cmd. Then, when the cmd is submitted to the driver, it can better prepare the BOs (e.g., pin the pages on demand) before it push the cmd to hardware.
N/A
Alternatively, we could send the list of BO handles together with the exec buf cmd to driver as other drivers are doing. But, it seems to be:
1. more complicated than this solution.
2. less efficient than this solution since the BO handle to DRM BO pointer conversion needs to happen in critical path (v.s. in set_arg code path, which is non-critical).
3. less efficient than this solution since the BO handle to DRM BO pointer conversion needs to happen every time when the command is submitted (v.s. happens only once when set_arg is called).
N/A
Done full build of XRT and installed on a system with U55n. Made sure xbutil validate works.
Also added some debug messages to make sure the newly added function is called properly.
N/A

### DIFF
--- a/src/runtime_src/core/common/api/bo_int.h
+++ b/src/runtime_src/core/common/api/bo_int.h
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2024 Advanced Micro Devices, Inc. All rights reserved.
+//
+#ifndef _XRT_COMMON_BO_INT_H_
+#define _XRT_COMMON_BO_INT_H_
+
+// This file defines implementation extensions to the XRT BO APIs.
+#include "core/include/experimental/xrt_bo.h"
+#include "core/common/shim/buffer_handle.h"
+
+namespace xrt_core::bo_int {
+
+xrt_core::buffer_handle*
+get_buffer_handle(const xrt::bo& bo);
+
+} // bo_int, xrt_core
+
+#endif
--- a/src/runtime_src/core/common/api/bo_int.h
+++ b/src/runtime_src/core/common/api/bo_int.h
@@ -10,8 +10,11 @@
 
 namespace xrt_core::bo_int {
 
-xrt_core::buffer_handle*
+const xrt_core::buffer_handle*
 get_buffer_handle(const xrt::bo& bo);
+
+size_t
+get_offset(const xrt::bo& bo);
 
 } // bo_int, xrt_core
 
--- a/src/runtime_src/core/common/api/bo_int.h
+++ b/src/runtime_src/core/common/api/bo_int.h
@@ -5,7 +5,7 @@
 #define _XRT_COMMON_BO_INT_H_
 
 // This file defines implementation extensions to the XRT BO APIs.
-#include "core/include/experimental/xrt_bo.h"
+#include "core/include/xrt/xrt_bo.h"
 #include "core/common/shim/buffer_handle.h"
 
 namespace xrt_core::bo_int {
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -1451,15 +1451,6 @@ size() const
   }) ;
 }
 
-size_t
-bo::
-offset() const
-{
-  return xdp::native::profiling_wrapper("xrt::bo::offset", [this]{
-    return handle->get_offset();
-  }) ;
-}
-
 uint64_t
 bo::
 address() const
@@ -1973,11 +1964,18 @@ xrtBOAddress(xrtBufferHandle bhdl)
 
 namespace xrt_core::bo_int {
 
-xrt_core::buffer_handle*
+const xrt_core::buffer_handle*
 get_buffer_handle(const xrt::bo& bo)
 {
   auto handle = bo.get_handle();
   return handle->get_handle();
+}
+
+size_t
+get_offset(const xrt::bo& bo)
+{
+  auto handle = bo.get_handle();
+  return handle->get_offset();
 }
 
 } // xrt_core::bo_int
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright (C) 2020-2022 Xilinx, Inc
-// Copyright (C) 2022-2023 Advanced Micro Devices, Inc. All rights reserved.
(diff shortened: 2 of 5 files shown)


src/runtime_src/core/common/api/bo_int.h
```diff
@@ -0,0 +1,18 @@
...
+#ifndef _XRT_COMMON_BO_INT_H_
+#define _XRT_COMMON_BO_INT_H_
+
+// This file defines implementation extensions to the XRT BO APIs.
+#include "core/include/experimental/xrt_bo.h"
+#include "core/common/shim/buffer_handle.h"
+
+namespace xrt_core::bo_int {
+
+xrt_core::buffer_handle*
```
stsoe, line 13: Maybe return `const xrt_core::buffer_handle*` ?

src/runtime_src/core/common/shim/buffer_handle.h
```diff
@@ -77,6 +77,20 @@ class buffer_handle
   {
     return XRT_NULL_BO;
   }
+
+  // Indicates to SHIM/driver that bh is going to be used by this BO. With
+  // offset and zie, it can also support using portion of bh (sub-BO).
+  // For now, this is only used when set_arg() is called upon an exec buf
+  // BO where pos is the arg index.
+  virtual void
+  bind_at(size_t pos, buffer_handle* bh, size_t offset, size_t size)
```
stsoe, line 86:
```suggestion
  bind_at(size_t pos, const buffer_handle* bh, size_t offset, size_t size) const
```
I think const will work in these two places.  At a minimum the bh argument should be const because you will not be modifying the buffer handle or what it points to, right?

stsoe (changes requested):
Looks good Max.  Few suggestions.

src/runtime_src/core/include/xrt/xrt_bo.h
```diff
@@ -469,6 +469,16 @@ class bo
...
 
+  /**
+   * offset() - Get the offset of this buffer
+   *
+   * @return
+   *  Offset of buffer in bytes
+   */
+  XCL_DRIVER_DLLESPEC
+  size_t
+  offset() const;
```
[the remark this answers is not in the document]
maxzhen: OK.

src/runtime_src/core/common/shim/buffer_handle.h
```diff
@@ -77,6 +77,20 @@ class buffer_handle
...
+  {
+    bool is_exec_buf = !!(get_properties().flags & XCL_BO_FLAGS_EXECBUF);
+
+    if (!is_exec_buf)
+      throw std::runtime_error("not supported");
+    // By default, nothing needs to be done for exec buf
```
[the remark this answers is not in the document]
maxzhen: This is the default behavior and should be good for everybody only except for Linux NPU. In this default implementation, we make sure:
1. you can only call bind_at() on exec buf BO
2. nothing needs to be done if you properly call it on exec buf BO
If I don't provide this default implementation, I'll have to implement it for all other platforms and the code will be the same.

src/runtime_src/core/common/api/xrt_bo.cpp
```diff
@@ -1443,6 +1451,15 @@ size() const
...
 
+size_t
+bo::
+offset() const
+{
+  return xdp::native::profiling_wrapper("xrt::bo::offset", [this]{
+    return handle->get_offset();
+  }) ;
+}
+
```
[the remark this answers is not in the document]
maxzhen: OK.

src/runtime_src/core/common/api/bo_int.h
maxzhen: OK.

src/runtime_src/core/common/shim/buffer_handle.h
maxzhen: I might have a std::map to keep track of the BO and index. I will need to modify them. bh can be const, sure.

src/runtime_src/core/common/shim/buffer_handle.h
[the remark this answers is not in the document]
stsoe: Let keep the body empty, aka a noop.   Calling get_properties at shim level is an ioctl call for at least some shims, the properties are cached in xrt_core::bo_impl but not in shim.  No need to go driver here.

src/runtime_src/core/common/api/bo_int.h
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2024 Advanced Micro Devices, Inc. All rights reserved.
+//
+#ifndef _XRT_COMMON_BO_INT_H_
+#define _XRT_COMMON_BO_INT_H_
+
+// This file defines implementation extensions to the XRT BO APIs.
+#include "core/include/experimental/xrt_bo.h"
```
stsoe, line 8:
```suggestion
#include "core/include/xrt/xrt_bo.h"
```

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -2225,6 +2226,11 @@ class run_impl
   {
     get_arg_setter()->set_arg_value(arg, bo);
 
+    auto bh = xrt_core::bo_int::get_buffer_handle(bo);
+    auto off = xrt_core::bo_int::get_offset(bo);
+    auto sz = bo.size();
+    cmd->get_exec_bo()->bind_at(arg.index(), bh, off, sz);
```
stsoe, line 2232: Hmm, maybe this should be a kernel_command member function?   Keeps this function simpler and moves the responsibility to the object that owns the exec bo.
`cmd->bind_arg_at_index(arg.index(), bo);`

stsoe (changes requested):

src/runtime_src/core/common/api/bo_int.h
maxzhen: Good to know that the flags are not cached. OK, I can make it a no-op one.

src/runtime_src/core/common/api/xrt_kernel.cpp
maxzhen: Sounds good. I'll make the change.

stsoe (approved):

PR #7940, opened 2024-02-10
Initial implementation of HIP apis and objects

rbramand-xilinx:
Added changes for hip objects - device, context, stream, module
Added stubs for apis related to hip objects
Implemented device enumeration

Device enumeration is done at hip library load.

Low

Tested using sw_emu and device enumeration works as expected

NA


src/runtime_src/hip/core/device.cpp
```diff
@@ -1,10 +1,16 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright (C) 2023 Advanced Micro Device, Inc. All rights reserved.
+// Copyright (C) 2023-204 Advanced Micro Device, Inc. All rights reserved.
```
keryell, line 2:
```suggestion
// Copyright (C) 2023-2024 Advanced Micro Device, Inc. All rights reserved.
```

src/runtime_src/hip/core/stream.h
```diff
@@ -0,0 +1,25 @@
...
+#ifndef xrthip_stream_h
+#define xrthip_stream_h
+
+#include "context.h"
+
+namespace xrt::core::hip {
+/**
+ * typedef stream_handle - opaque stream handle
+ */
+typedef void* stream_handle;
```
keryell, line 12: `using`?

src/runtime_src/hip/api/hip_module.cpp
```diff
@@ -0,0 +1,166 @@
...
+#include "core/common/error.h"
+
+#include "hip/config.h"
+#include "hip/hip_runtime_api.h"
+
+#include "hip/core/module.h"
+
+namespace xrt::core::hip {
+static void
+hipModuleLaunchKernel(hipFunction_t f, uint32_t gridDimX, uint32_t gridDimY,
```
keryell, line 13: I am curious about which device will be targeted. FPGA with HLS or OpenCL?

src/runtime_src/hip/core/context.cpp
```diff
@@ -0,0 +1,34 @@
...
+// Copyright (C) 2024 Advanced Micro Device, Inc. All rights reserved.
+
+#include "common.h"
+#include "context.h"
+#include "device.h"
+
+namespace xrt::core::hip {
+context::
+context(std::shared_ptr<device> device)
+  : m_device(device)
```
stsoe, line 11:
```suggestion
  : m_device(std::move(device))
```

src/runtime_src/hip/core/device.cpp
```diff
@@ -1,10 +1,16 @@
...
-
 // Implementation
+xrt_core::handle_map<device_handle, std::shared_ptr<device>> device_cache;
 
+device::
+device(uint32_t _device_id)
+{
+  m_xrt_device = xrt::device{_device_id};
+  m_device_id = _device_id;
+}
```
stsoe, line 15: Prefer initializer list.
And please no leading or trailing underscore in parameters or variables.
```suggestion
device(uint32_t device_id)
  : m_device_id{device_id}
  , m_xrt_device{m_device_id}
{}
```

src/runtime_src/hip/core/device.h
```diff
@@ -3,12 +3,51 @@
...
 
 class device
 {
+  xrt::device m_xrt_device;
+  uint32_t m_device_id = UINT32_MAX;
+  unsigned int m_flags;
+
+public:
+  explicit
+  device(uint32_t _device_id);
```
stsoe, line 28: No leaning or trailing underscore, please.
```suggestion
  device(uint32_t device_id);
```

src/runtime_src/hip/core/module.cpp
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2024 Advanced Micro Device, Inc. All rights reserved.
+
+#include "module.h"
+
+namespace xrt::core::hip {
+module::
+module(std::shared_ptr<context> ctx)
+  : m_ctx(ctx)
```
stsoe, line 9:
```suggestion
  : m_ctx{std::move(ctx)}
```

src/runtime_src/hip/core/device.h
```diff
@@ -3,12 +3,51 @@
...
+ */
+typedef uint32_t device_handle;
+
+// forward declaration
+class context;
 
 class device
 {
+  xrt::device m_xrt_device;
+  uint32_t m_device_id = UINT32_MAX;
```
stsoe, line 23: Looks like this needs <cstdint> to be included.
It is custom to do default initialization as `var {value}`  that that it makes a difference.

src/runtime_src/hip/core/common.h
```diff
@@ -0,0 +1,23 @@
...
+#include <stack>
+
+namespace xrt::core::hip {
+// thread local hip objects
+class hip_tls_objs
+{
+public:
+  std::stack<std::shared_ptr<context>> ctx_stack;
+  bool pri_ctx_active = false;
+  std::shared_ptr<context> pri_ctx = nullptr;
```
stsoe, line 18: std::shared_ptr default initializes to nullptr.

src/runtime_src/hip/core/common.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include <stack>
+
+namespace xrt::core::hip {
+// thread local hip objects
+class hip_tls_objs
+{
+public:
+  std::stack<std::shared_ptr<context>> ctx_stack;
+  bool pri_ctx_active = false;
```
stsoe, line 17:
```suggestion
  bool pri_ctx_active {false};
```

src/runtime_src/hip/api/hip_stream.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include "core/common/error.h"
+
+#include "hip/config.h"
+#include "hip/hip_runtime_api.h"
+
+#include "hip/core/stream.h"
+
+namespace xrt::core::hip {
+static void
+hipStreamCreateWithFlags(hipStream_t *stream, unsigned int flags)
```
stsoe, line 13: Suggest `type* var` over `type *var`, which has been done in other place (but maybe not in all :-) )
```suggestion
hipStreamCreateWithFlags(hipStream_t* stream, unsigned int flags)
```

src/runtime_src/hip/api/hip_device.cpp
```diff
@@ -0,0 +1,263 @@
...
+  if (uuid == nullptr)
+    throw xrt_core::system_error(hipErrorInvalidValue, "arg passed is nullptr");
+
+  throw std::runtime_error("Not implemented");
+}
+
+static void
+hipDeviceGetAttribute(int* pi, hipDeviceAttribute_t attr, int device)
+{
+  if (pi == nullptr)
```
stsoe, line 133:
```suggestion
  if (pi)
```

src/runtime_src/hip/api/hip_device.cpp
```diff
@@ -0,0 +1,263 @@
...
+  throw std::runtime_error("Not implemented");
+}
+
+static void
+hipDeviceGetUuid(hipUUID* uuid, hipDevice_t device)
+{
+  if (device < 0 || xrt::core::hip::device_cache.count(static_cast<xrt::core::hip::device_handle>(device)) == 0)
+    throw xrt_core::system_error(hipErrorInvalidDevice, "device requested is not available");
+
+  if (uuid == nullptr)
```
stsoe, line 124:
```suggestion
  if (uuid)
```

src/runtime_src/hip/api/hip_context.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include "hip/core/device.h"
+
+namespace xrt::core::hip {
+// Returns handle to context
+// Throws on error
+static void
+hipCtxCreate(hipCtx_t* ctx, unsigned int flags, hipDevice_t device)
+{
+  auto hip_dev = device_cache.get(static_cast<device_handle>(device));
+  if (hip_dev == nullptr) {
```
stsoe, line 18: Change all occurrences please, I started in some other file, but too many.  
```suggestion
  if (hip_dev) {
```
As a general comment, my preference would be to use lower_case function names and more C++'ish for the functions we define (not the hip APIs of course).  In C++  this function really should return what it creates rather than use an out-variable.   It any way throws on error, which is good.

stsoe (changes requested):
This is nice clean code, which reads very well. 
Few coding suggestions.

src/runtime_src/hip/api/hip_module.cpp
rbramand-xilinx: it can be used for different kind of devices - alveo, ryzen ai etc

src/runtime_src/hip/api/hip_context.cpp
rbramand-xilinx: Hi @stsoe, I have changed internal static functions implementation as suggested. Please review

src/runtime_src/hip/core/context.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+// Global map of contexts
+xrt_core::handle_map<context_handle, std::shared_ptr<context>> context_cache;
+
+// thread local hip objects
+thread_local hip_tls_objs tls_objs;
+
+// returns current context
+// if primary context is active it is current
+// else returns top of ctx stackk
```
keryell, line 22:
```suggestion
// else returns top of ctx stack
```
It is unclear to me what is this stack about.

src/runtime_src/hip/core/device.h
```diff
@@ -3,12 +3,63 @@
...
 
+// device_handle - opaque device handle
+using device_handle = uint32_t;
+
+// forward declaration
+class context;
+
 class device
 {
+  uint32_t m_device_id{UINT32_MAX};
```
keryell, line 21: Do not use macros when you can in C++, use `std::numeric_limits` instead.

src/runtime_src/hip/api/hip_context.cpp
```diff
@@ -0,0 +1,241 @@
...
+  auto handle = reinterpret_cast<context_handle>(ctx);
+  if (!handle) {
+    throw xrt_core::system_error(hipErrorInvalidValue, "device requested is not available");
+  }
+
+  auto hip_ctx = context_cache.get(handle);
+  if (!hip_ctx)
+    throw xrt_core::system_error(hipErrorInvalidValue, "context handle not found");
+
+  // Need to remove the ctx of calling thread if its the top one
```
keryell, line 40:
```suggestion
  // Need to remove the ctx of calling thread if it is the top one
```
Could you explain the logic of the stack and the cache somewhere?
So it is possible to destroy the current context. But what happens then? And if there is no more context?
Is there a `shared_ptr` still alive for it?

keryell (commented):
Quite interesting!

src/runtime_src/hip/api/hip_context.cpp
```diff
@@ -0,0 +1,241 @@
...
+  if (!tls_objs.ctx_stack.empty() && tls_objs.ctx_stack.top().lock() == hip_ctx) {
+    tls_objs.ctx_stack.pop();  // remove it is as current
```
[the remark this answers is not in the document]
rbramand-xilinx: stack holds weak_ptrs, so even though stack still contains entry its not an issue as when we call get_current_context() we get nullptr if the weak_ptr entry that is on top of stack is pointing to ptr that is already removed from cache

src/runtime_src/hip/api/hip_context.cpp
rbramand-xilinx: In HIP, application doesn't explicitly specify context in api calls so stack is used for getting active context. Application can call hipCtxCreate multiple times and when there is any api which needs context we check active context from stack(top one) and use it. This follows hip and cuda documentation.

Cache is map for getting actual object from handles (HIP operates on handles). We followed XRT native c++/c approach here. 
Even though application destroys Ctx using hipCtxDestroy that removed entry from map, ctx can be still alive because some other object is holding its reference and will be destroyed gracefully afterwards. This way we avoid application crashes and undefined behavior. As per cuda documentation it is undefined behavior in their case which we are avoiding in our implementation

rbramand-xilinx: sure @keryell, I will add it. Thanks

stsoe (approved):
Looks good.  Thank you.

chvamshi-xilinx (approved):

PR #7944, opened 2024-02-13
CR-1123282 - Use get_size sdr request to receive sensor

karthdmg-xilinx:
Use get_size sdr request to receive sensor data size

This is an enhancement to allocate the required sensor buffer size before processing sensor request.

Get sensor buffer size before and allocate buffer accordingly before processing sensor request.
None
1. Verify the sensor buffer information with changes in both XRT and VMR.
2. Verify the sensor buffer information with changes in VMR only to verify backward compatibility. In this case, the buffer size is allocated with 4096 default value.
3. Verify the sensor buffer information with changes in XRT only to verify backward compatibility. In this case, the buffer size is allocated with 4096 default value.

None


src/runtime_src/core/include/xgq_impl.h
```diff
@@ -533,6 +533,7 @@ static inline int xgq_attach(struct xgq *xgq, uint64_t flags, uint64_t io_hdl, u
 		// Write 1 to GCQ interrupt enable register to always enable interrupt
 		xgq_reg_write32(xgq->xq_io_hdl, sqprod + XGQ_INTR_ENABLE_OFFSET, 1);
 	}
+
```
xdavidz, line 536: please don't change the xgq_impl.h unless you have to. The vmr build.sh will check if the xgq_impl.h in vmr repo is matching with XRT version.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
```diff
@@ -361,20 +361,28 @@ static int show_sensors_raw(struct xocl_hwmon_sdm *sdm, char *buf,
...
+	size_t resp_len = 0;
 	char* sdr_buf;
 	int repo_type;
 	uint64_t data_args = 0;
 
 	mutex_lock(&sdm->sdm_lock);
+	/* Collect the SDR size from VMR */
+	xocl_xgq_collect_sensors_get_size(xdev, &resp_len, repo_id);
+
+	if (resp_len <= 0) {
```
xdavidz, line 373: I would think there should be some "defensive" coding like
if ((resp_len <= 0) || (rep_len > RESP_LEN))
  goto fail

xdavidz (approved):
LGTM, but please fix some nits.

prapulkrishnamurthy (approved):

src/runtime_src/core/include/xgq_impl.h
karthdmg-xilinx: noted thank you

src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
```diff
@@ -361,20 +361,28 @@ static int show_sensors_raw(struct xocl_hwmon_sdm *sdm, char *buf,
...
 	int ret = 0, kind;
-	size_t resp_len = RESP_LEN;
+	size_t resp_len = 0;
 	char* sdr_buf;
 	int repo_type;
 	uint64_t data_args = 0;
 
 	mutex_lock(&sdm->sdm_lock);
+	/* Collect the SDR size from VMR */
+	xocl_xgq_collect_sensors_get_size(xdev, &resp_len, repo_id);
```
rajkumar-xilinx, line 371: Hi @karthdmg-xilinx ,
IMO, call the function xocl_xgq_collect_sensors_get_size() only once in hwmon_sdm probe() and save the resp_size data in the struct xocl_hwmon_sdm for future usage in the driver.

rajkumar-xilinx (changes requested):
Hi @karthdmg-xilinx ,
IMO, call the function xocl_xgq_collect_sensors_get_size() only once in hwmon_sdm probe() and save the resp_size data in the struct xocl_hwmon_sdm for future usage in the driver.

src/runtime_src/core/pcie/driver/linux/xocl/subdev/hwmon_sdm.c
karthdmg-xilinx: during hot reset (xclmgmt_hot_reset()), I believe re-probe wont happen so for this we may be able to read the new size information from VMR. So I added code for every request. overall sensor size may not change though even after hot reset, I'm not sure if it works correctly under various cases. correct me if I'm wrong.

xdavidz: I think either approach is fine: pulling size every time or cache size. But please consider the corner case like if the cached size could be over flow the allocated memory buffer. I do not check the detail logics, but this case should be avoided in any combination.

Performance should be fine. Please test "xbutil examine" "xbutil query". I recall there was a CR for the xbutil runs over 1 second.

karthdmg-xilinx: buffer length is read per repo_id, we need array sensor_size params  to store in struct xocl_hwmon_sdm. if we consolidate all the repo_ids, then I see it is crossing 4k size

rajkumar-xilinx: yes, you need to define array of sensor_size[] and update it in probe(), reset() paths

karthdmg-xilinx: secondly, if not privileged (sdm->privileged), is it okay to send xgq command to calculate size? or the get size request should go through hwmon_sdm_read_from_peer() ?

rajkumar-xilinx (approved):

PR #7946, opened 2024-02-14
Clock Frequencies for Compute Units and Monitors

predutta:

Updated our profiling internal databases.
Added specific clock frequency for each compute unit instance.
Added specific clock frequency for each monitor.
When executing update_device, parsed the specific clock frequencies from the SYSTEM_METADATA section.
Updated our trace alignment.
When adjusting device trace timestamps, used the trace clock for calculations.
The trace clock is determined by choosing the fastest compute unit clock.
Updated our profiling tables.
For each table in the summary writer that uses the PL clock frequency, adjusted to use the correct Compute Unit Instance clock.


stsoe (changes requested):
Would it be possible to split this PR into multiple ones?
I suggest one PR for changing mode to at least get that noise out of the way.

There are bunch of .foo.o.cmd files  that do not belong in XRT.

stsoe:
It looks like this PR is a mistake, it's definitely a mess.   I am closing it.

PR #7949, opened 2024-02-14
Move metadata Reader to static database

vipangul:
- AIE Metadata reader is now moved to static database. This change is needed for AIE Relaod/Reset feature as it expects all the required AIE Metadata is parsed and updated in static database as part of updateDevice().
- This also includes a change to compile xdp_core for client.

NA

Test Performed:
Verified on VCK190 and Phoenix.

### DIFF
--- a/src/runtime_src/xdp/profile/plugin/aie_trace/aie_trace_metadata.cpp
+++ b/src/runtime_src/xdp/profile/plugin/aie_trace/aie_trace_metadata.cpp
@@ -71,21 +71,13 @@ namespace xdp {
       xrt_core::message::send(severity_level::warning, "XRT", AIE_TRACE_DUMP_INTERVAL_WARN_MSG);
     }
 
-    #ifdef XDP_CLIENT_BUILD
-
-    metadataReader = aie::readAIEMetadata("aie_control_config.json", aie_meta);
-    
-    #else
-
-    auto device = xrt_core::get_userpf_device(handle);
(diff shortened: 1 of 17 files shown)


src/runtime_src/xdp/profile/plugin/aie_trace/aie_trace_metadata.cpp
```diff
@@ -71,21 +71,13 @@ namespace xdp {
...
-
-    metadataReader = aie::readAIEMetadata(data.first, data.second, aie_meta);
-
-    #endif
-
-    if (metadataReader == nullptr)
+    metadataReader = (VPDatabase::Instance()->getStaticInfo()).getAIEmetadataReader();
+    if (!metadataReader) {
+      std::stringstream msg;
+      msg << "Metadata reader reader couldn't be created for AIE trace";
```
pgschuey, line 77: This message reads funny. I'd suggest rewording as "Metadata reader could not be created for AIE trace". We also don't have to create a stringstream here. It would be simpler to just put the string as the third argument to send().

pgschuey (changes requested):

src/runtime_src/xdp/profile/plugin/aie_trace/aie_trace_metadata.cpp
vipangul: Thanks, pushed a commit addressing this at applicable plugins.

jvillarre (approved):

PR #7955, opened 2024-02-19
Add support for debug buffer object in XRT for internal use

stsoe:
Allocate a “special” type of BO in a virtual region shared with instruction buffers within a hardware context. The driver will share this buffer object with firmware where the buffer will be populated with data to be shared with user space XRT.

The implementation makes use of a new XRT_BO_USE_DEBUG flag which is ignored for all other use cases.

The allocation of the debug buffer object is through an internal API which takes as argument the xrt::hw_context in which the buffer is to be allocated, plus the size of the debug buffer.

None, the new API is currently not used.

### DIFF
--- a/src/runtime_src/core/common/api/bo_int.h
+++ b/src/runtime_src/core/common/api/bo_int.h
@@ -5,6 +5,7 @@
 #define _XRT_COMMON_BO_INT_H_
 
 // This file defines implementation extensions to the XRT BO APIs.
+#include "core/common/config.h"
 #include "core/include/xrt/xrt_bo.h"
 #include "core/common/shim/buffer_handle.h"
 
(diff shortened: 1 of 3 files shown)


maxzhen (approved):

dezhiAmd:
> #### Problem solved by the commit
> Allocate a “special” type of BO in a virtual region shared with instruction buffers within a hardware context. The driver will share this buffer object with firmware where the buffer will be populated with data to be shared with user space XRT.
> 
> #### How problem was solved, alternative solutions (if any) and why they were rejected
> The implementation makes use of a new XRT_BO_USE_DEBUG flag which is ignored for all other use cases.
> 
> The allocation of the debug buffer object is through an internal API which takes as argument the xrt::hw_context in which the buffer is to be allocated, plus the size of the debug buffer.
> 
> #### Risks (if any) associated the changes in the commit
> None, the new API is currently not used.

Anyone happen to know which firmware feature is related to this debug buffer obejct?

PR #7957, opened 2024-02-20
CR-1190692 Add method to set read range on xrt::ip

stsoe:
Add internal API to set the read range of an xrt::ip object.

Access to xclAPIs is being removed and everyone including internal XRT must use native objects for access shim functionality.  For special deadlock detection, IPs have registers outside the advertised address range. In order to read these registers, the address range must be expanded.  

This PR adds support for expanding the read range of an xrt::ip.

Internal API, currently unused.

No testing


jvillarre (approved):

IshitaGhosh (approved):
The change looks good. I could use the set_read_range API for deadlock diagnosis in XDP.

PR #7961, opened 2024-02-21
using branch for test summary instead of version

SubhedarV:
- using branch name instead of version for test summary job


manikandan-xilinx (approved):

manikandan-xilinx:
merging as one line change in workflow file

PR #7963, opened 2024-02-22
add xrt::elf raw elf data constructor

HimanshuChoudhary-Xilinx:
Added new xrt::elf constructor to take in-memory buffer as input

NA

added new constructor to take memory elf buffer

None

taken multiple efl and check we are able to get correct xrt::elf object
```
 // Sample usage
  aiebu::aiebu_assembler as(aiebu::aiebu_assembler::buffer_type::blob_instr_dpu, v1, v2, patch_data);
  auto e = as.get_elf();

  std::istringstream elf_stream;
  elf_stream.rdbuf()->pubsetbuf(e.data(), e.size());
  xrt::elf elf{elf_stream};
  xrt::module mod{elf};
```

Yes, we have to update XRT doc

### DIFF
--- a/src/runtime_src/core/common/api/xrt_elf.cpp
+++ b/src/runtime_src/core/common/api/xrt_elf.cpp
@@ -14,6 +14,7 @@
 #include <stdexcept>
 #include <string>
 #include <vector>
+#include <sstream>
 
 namespace xrt {
 
@@ -26,6 +27,14 @@ public:
   {
     if (!m_elf.load(fnm))
       throw std::runtime_error(fnm + " is not found or is not a valid ELF file");
+  }
+
+  elf_impl(const std::vector<char>& data)
+  {
+    std::string str(data.begin(), data.end());
+    std::istringstream isstr(str);
+    if (!m_elf.load(isstr))
+      throw std::runtime_error("not a valid ELF data");
   }
 
   const ELFIO::elfio&
@@ -83,6 +92,11 @@ namespace xrt {
 elf::
 elf(const std::string& fnm)
   : detail::pimpl<elf_impl>{std::make_shared<elf_impl>(fnm)}
+{}
+
+elf::
+elf(const std::vector<char>& data)
+  : detail::pimpl<elf_impl>{std::make_shared<elf_impl>(data)}
 {}
 
 xrt::uuid
--- a/src/runtime_src/core/common/api/xrt_elf.cpp
+++ b/src/runtime_src/core/common/api/xrt_elf.cpp
@@ -31,8 +31,8 @@ public:
 
   elf_impl(const std::vector<char>& data)
   {
-    const std::string elf_string(data.begin(), data.end());
-    std::istringstream elf_stream(elf_string);
+    std::istringstream elf_stream;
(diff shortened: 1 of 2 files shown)


chvamshi-xilinx:
ok to test/retest is no longer working. We need to go to actions and run all jobs again. Started run just now

chvamshi-xilinx (approved):

src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -28,6 +29,14 @@ class elf_impl
       throw std::runtime_error(fnm + " is not found or is not a valid ELF file");
   }
 
+  elf_impl(const std::vector<char>& data)
+  {
+    const std::string elf_string(data.begin(), data.end());
+    std::istringstream elf_stream(elf_string);
+    if (!m_elf.load(elf_stream))
```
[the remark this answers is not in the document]
dezhiAmd: Someone did this:
https://stackoverflow.com/questions/16875321/stdistringstream-from-stdstring-without-copying

src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -28,6 +29,14 @@ class elf_impl
       throw std::runtime_error(fnm + " is not found or is not a valid ELF file");
   }
 
+  elf_impl(const std::vector<char>& data)
+  {
+    std::istringstream elf_stream;
+    elf_stream.rdbuf()->pubsetbuf(const_cast<char*>(data.data()), data.size());
```
sonals, line 35: The const_cast looks evil here. Can you add a comment why we are doing it and the fact that elf_stream is input stream (read only) so it cannot accidentally change **data**.

HimanshuChoudhary-Xilinx: since data is const type so data.data() return const char* 
https://cplusplus.com/reference/vector/vector/data/
`If the vector object is const-qualified, the function returns a pointer to const value_type. Otherwise, it returns a pointer to value_type.`
while 
`streambuf* pubsetbuf (char* s, streamsize n);`
https://cplusplus.com/reference/streambuf/streambuf/pubsetbuf/

src/runtime_src/core/include/experimental/xrt_elf.h
```diff
@@ -32,7 +33,17 @@ class elf : public detail::pimpl<elf_impl>
...
+
+  /**
+   * elf() - Constructor from raw ELF data
+   *
+   * @param data
+   *  Raw data of elf
+   *
+   */
+  XRT_API_EXPORT
+  elf(const std::vector<char>& data);
```
stsoe, line 45: A constructor from std::vector isn't great IMO.   What if data is not held in a vector?   A std::span would be better, but that's c++20.  I would opt for `const char*, size_t` but you are still faced with the problem of how to create the istream without copying the in-memory buffer.

HimanshuChoudhary-Xilinx: @stsoe 
Updated that PR with changes discussed offline.

stsoe (approved):

sonals (approved):

PR #7964, opened 2024-02-22
Initial implementation of Hip module and function apis

rbramand-xilinx:
Implemented hip module and function classes, added stubs for apis related to them
added a function to throw exception for reusability

low

Tested on pcie sw_emu by creating test case with multiple contexts

NA

### DIFF
--- a/src/runtime_src/hip/api/hip_context.cpp
+++ b/src/runtime_src/hip/api/hip_context.cpp
@@ -29,8 +29,10 @@ hip_ctx_create(unsigned int flags, hipDevice_t device)
   hip_dev->set_flags(flags);
   auto hip_ctx = std::make_shared<context>(hip_dev);
   tls_objs.ctx_stack.push(hip_ctx);  // make it current
-  context_cache.add(hip_ctx.get(), std::move(hip_ctx));
-  return hip_ctx.get();
+  tls_objs.dev_hdl = device;
+  auto handle = hip_ctx.get();
+  context_cache.add(handle, std::move(hip_ctx));
+  return handle;
 }
 
 static void
@@ -74,8 +76,10 @@ hip_ctx_set_current(hipCtx_t ctx)
 
   auto handle = reinterpret_cast<context_handle>(ctx);
   auto hip_ctx = context_cache.get(handle);
-  if (hip_ctx)
+  if (hip_ctx) {
     tls_objs.ctx_stack.push(hip_ctx);
+    tls_objs.dev_hdl = hip_ctx->get_dev_id();
+  }
 }
 
 // remove primary ctx as active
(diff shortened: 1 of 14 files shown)


src/runtime_src/hip/core/common.h
```diff
@@ -13,17 +18,26 @@ namespace xrt::core::hip {
...
 };
 extern thread_local hip_tls_objs tls_objs;
 } // xrt::core::hip
 
+namespace {
+// common function that throws hip error
+inline void
+throw_if(bool check, hipError_t err, const std::string& err_msg)
+{
+  check ? throw xrt_core::system_error(err, err_msg) : void();
```
keryell, line 39: Nice! But that looks a little bit obfuscated
```suggestion
  if (check)
    throw xrt_core::system_error(err, err_msg);
```
Perhaps you could introduce some `throw_invalid_value_if` and so on for common cases?

rbramand-xilinx: @keryell I have added changes as suggested, please review

src/runtime_src/hip/api/hip_module.cpp
```diff
@@ -1,64 +1,81 @@
...
+static module_handle
+hip_module_load_data_ex(const void* image, unsigned int numOptions,
+                        hipJitOption* options, void** optionsValues)
 {
-  if (!module)
-    throw xrt_core::system_error(hipErrorInvalidResourceHandle, "module is nullptr");
+  // Jit options are ignored for now
+  (void)numOptions;
+  (void)options;
+  (void)optionsValues;
```
stsoe, line 60: I have never liked this syntax.  Newer C++ has std::ignore which you can assign to, but personally I always either comment out the named parameters in the function signature or remove it e.g. `void foo(Type /*type*/)` or `void foo(Type)`.  You don't have to make this change, it's up to you.

stsoe (approved):
Clean, simple, and reads well.  Looks good to me.  
I didn't go into deep details.

src/runtime_src/hip/api/hip_module.cpp
keryell: Yes, just removing the parameter names is a good way to go with C++ in this case.

src/runtime_src/hip/api/hip_module.cpp
```diff
@@ -1,64 +1,81 @@
...
+create_module(const void* image)
 {
-  if (!module)
-    throw xrt_core::system_error(hipErrorInvalidResourceHandle, "module is nullptr");
-
-  throw std::runtime_error("Not implemented");
+  auto ctx = get_current_context();
+  auto hip_module = std::make_shared<module>(ctx, const_cast<void*>(image));
+  auto module_hdl = hip_module.get();
+  module_cache.add(module_hdl, std::move(hip_module));
```
keryell, line 49: There is a lot of boiling plate here and at the end you pass twice your module in a different way, which make me think that `module_cache.add(module_hdl, std::move(hip_module))` might not have the perfect API.

src/runtime_src/hip/api/hip_context.cpp
```diff
@@ -23,27 +20,25 @@ static context_handle
...
+  throw_invalid_value_if(!hip_dev, "device requested is not available");
 
   hip_dev->set_flags(flags);
   auto hip_ctx = std::make_shared<context>(hip_dev);
   tls_objs.ctx_stack.push(hip_ctx);  // make it current
-  context_cache.add(hip_ctx.get(), std::move(hip_ctx));
-  return hip_ctx.get();
+  tls_objs.dev_hdl = device;
+  auto handle = hip_ctx.get();
+  context_cache.add(handle, std::move(hip_ctx));
```
keryell, line 30: API to be improved?

src/runtime_src/hip/api/hip_module.cpp
```diff
@@ -1,64 +1,81 @@
...
 
-  throw std::runtime_error("Not implemented");
+  auto mod_hdl = reinterpret_cast<module_handle>(hmod);
+  auto hip_mod = module_cache.get(mod_hdl);
+  throw_invalid_resource_if(!hip_mod, "module not available");
+
+  auto hip_func = std::make_shared<function>(mod_hdl, std::string(name));
+  auto func_hdl = hip_func.get();
+  hip_mod->add_function(func_hdl, std::move(hip_func));
+  return func_hdl;
```
keryell, line 40: Perhaps a terser API could be
```suggestion
  return  hip_mod->add_function(std::make_shared<function>(mod_hdl, std::string(name));
```

keryell (commented):
Great clarification!

src/runtime_src/hip/api/hip_module.cpp
rbramand-xilinx: @keryell thanks for the suggestion, I have made changes. Please review

keryell (approved):
Thanks.
Quite higher level!

PR #7965, opened 2024-02-22
Clean up hip tests to please clang-tidy

sonals:
Clean clang-tidy report of hip tests

NA

Addressed clang-tidy reported issues.

None

Compile only test

NA

### DIFF
--- a/tests/hip/vadd-stream/main.cpp
+++ b/tests/hip/vadd-stream/main.cpp
@@ -7,6 +7,8 @@
 #include <iostream>
 #include <memory>
 #include <thread>
+#include <array>
+#include <vector>
 
 #include "hip/hip_runtime_api.h"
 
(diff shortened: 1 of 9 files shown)


tests/hip/vadd-stream/main.cpp
```diff
@@ -72,14 +72,14 @@ mainworkerthread(hipFunction_t function, hipStream_t stream, bool validate = tru
 
   std::cout << "*********************************************************************************\n";
 
-  std::unique_ptr<float[]> host_a(new float[vector_length]);
-  std::unique_ptr<float[]> host_b(new float[vector_length]);
-  std::unique_ptr<float[]> host_c(new float[vector_length]);
+  std::array<float, vector_length> host_a{};
+  std::array<float, vector_length> host_b{};
+  std::array<float, vector_length> host_c{};
```
stsoe, line 77: Array may be problematic depending on size.  This is stack memory.  I think std::vector will be better / safer.

stsoe (commented):
LGTM.  Since this is C++ you should probably use static_cast in place of c-style cast.

sonals:
> LGTM. Since this is C++ you should probably use static_cast in place of c-style cast.

Updated to use static_cast

PR #7968, opened 2024-02-24
Add pcie revision id support

vengutta18:
pcie revision id support is needed for new platforms.

It's not a bug. I'm adding the support to query the PCI device revision id. 

There was no provision to query the revision id

N/A

Validate the existing platform functionality and the shim_tests

N/A

### DIFF
--- a/src/runtime_src/core/common/query_requests.h
+++ b/src/runtime_src/core/common/query_requests.h
@@ -43,6 +43,7 @@ enum class key_type
 {
   pcie_vendor,
   pcie_device,
+  pcie_revision,
   pcie_subsystem_vendor,
   pcie_subsystem_id,
   pcie_link_speed,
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/common/query_requests.h
```diff
@@ -352,6 +353,23 @@ struct pcie_device : request
   }
 };
 
+struct pcie_revision : request
```
dbenusov-xilinx, line 356: There are a few places where XRT depends on `pcie_device` as being the only identifier for the system. By adding in an additional value we throw a lot of that logic away. We need a better way to get the device ID that includes the revision number...

Instead of creating a new query for the revision, perhaps we can create a new query that creates a pcie identifier? Whatever change we make in XRT would need to update all usages of `pcie_device` so this is not a little update.

dbenusov-xilinx (changes requested):

src/runtime_src/core/common/query_requests.h
vengutta18: @maxzhen  I just did that PCIe identifier key: "pci device id and revision id combined" in our driver. 
I added the pci_revision here in XRT module to let shim_tests run without any issues.

As per Daniel comments looks like if we want to include revision id as the PCIE identifier it impacts multiple places in XRT module. So can this be taken care of by the XRT team?
I don't have much exposure to this XRT module and don't wanna cause lot of regressions.

dbenusov-xilinx:
Superceded by https://github.com/Xilinx/XRT/pull/7971

PR #7971, opened 2024-02-27
CR-1191367 Use PCI revision ID and deivce ID to identify device

dbenusov-xilinx:
Package revision ID with device ID for better device categorization.

https://jira.xilinx.com/browse/CR-1191367

This is an enhancement to the original PR: https://github.com/Xilinx/XRT/pull/7968

Various files may be different for devices that share a device ID but have different PCI Revision IDs.

The "xbutil validate" command is looking for validate.xclbin based on the device ID. This CR is to add support of revision id in XRT.

After the change, the validate.xclbin path on Linux system will be,
/lib/firmware/amdnpu/<device_id>_<revision_id>/validate.xclbin

The device_id is a 4 char length hex number, such as, 17f0.
The revision_id is a 2 char length hex number, such as, 00.

Below is the final validate xclbin path for Phoenix (device id 1502, revision id 00)
/lib/firmware/amdnpu/1502_00/validate.xclbin

Found internally.

Solved by adding a new query request that packages the device id and revision id into a single structure. Additionally various helper functions are included to make our lives a bit easier.

This should not impact windows NPU or alveo. This change is focused on the Linux implementation.

Tested on Ubuntu 22.04 U55c
```
dbenusov@xsjdbenusov50:/proj/rdi/staff/dbenusov$ xbutil examine -r pcie-info -o test.json --force
WARNING: Unexpected xocl version (2.16.0) was found. Expected 2.17.0, to match XRT tools.

-------------------------------------------------
[0000:04:00.1] : xilinx_u55c_gen3x16_xdma_base_3
-------------------------------------------------
Pcie Info
  Vendor                 : 0x10ee
  Device                 : 505d
  Sub Device             : 0x000e
  Sub Vendor             : 0x10ee
  PCIe                   : Gen3x4
  DMA Thread Count       : 2
  CPU Affinity           : 0-15
  Shared Host Memory     : 0 Byte
  Max Shared Host Memory : 0 Byte
  Enabled Host Memory    : 0 Byte

Successfully wrote the json file: test.json
dbenusov@xsjdbenusov50:/proj/rdi/staff/dbenusov$ cat test.json
{
    "schema_version": {
        "schema": "JSON",
        "creation_date": "Tue Feb 27 00:33:31 2024 GMT"
    },
    "devices": [
        {
            "interface_type": "pcie",
            "device_id": "0000:04:00.1",
            "device_status": "HEALTHY",
            "pcie_info": {
                "vendor": "0x10ee",
                "device": "505d",
                "revision": "00",
                "sub_device": "0x000e",
                "sub_vendor": "0x10ee",
                "link_speed_gbit_sec": "3",
                "expected_link_speed_gbit_sec": "3",
                "express_lane_width_count": "4",
                "expected_express_lane_width_count": "16",
                "dma_thread_count": "2",
                "cpu_affinity": "0-15",
                "max_shared_host_mem_aperture_bytes": "0 Byte",
                "shared_host_mem_size_bytes": "0 Byte",
                "enabled_host_mem_size_bytes": "0 Byte"
            }
        }
    ]
}
```

Ubuntu 22.04 U55C (Change to look like a NPU device)
The output of the firmware files are in
```
/lib/firmware/amdnpu/505d_00/
```
Where 505d is the device ID for a u55c and 00 is the revision number.

None

### DIFF
--- a/src/runtime_src/core/common/query_requests.h
+++ b/src/runtime_src/core/common/query_requests.h
@@ -50,6 +50,7 @@ enum class key_type
   pcie_express_lane_width,
   pcie_express_lane_width_max,
   pcie_bdf,
+  pcie_id,
 
   instance,
   edge_vendor,
@@ -463,6 +464,44 @@ struct pcie_bdf : request
     return boost::str
       (boost::format("%04x:%02x:%02x.%01x") % std::get<0>(value) %
        std::get<1>(value) % std::get<2>(value) % std::get<3>(value));
+  }
+};
+
+/**
+ *  Useful for identifying devices that utilize revision numbers. Prefer this request over pcie_device.
+ */
+struct pcie_id : request
+{
+  struct data {
+    uint16_t device_id;
+    uint8_t revision_id;
+  };
+
+  using result_type = data;
+  static const key_type key = key_type::pcie_id;
+  static const char* name() { return "pcie_id"; }
+
+  virtual std::any
+  get(const device*) const = 0;
+
+  static std::string
+  device_to_string(const result_type& value)
+  {
+    return boost::str(boost::format("%04x") % value.device_id);
+  }
+
+  static std::string
+  revision_to_string(const result_type& value)
+  {
+    // The cast is required. This is a boost bug. https://github.com/boostorg/format/issues/60
+    return boost::str(boost::format("%02x") % (uint16_t) value.revision_id);
+  }
+
+  static std::string
+  to_path(const result_type& value)
+  {
+    // The cast is required. This is a boost bug. https://github.com/boostorg/format/issues/60
+    return boost::str(boost::format("%04x_%02x") % value.device_id % (uint16_t) value.revision_id);
   }
 };
 
--- a/src/runtime_src/core/common/query_requests.h
+++ b/src/runtime_src/core/common/query_requests.h
@@ -494,14 +494,14 @@ struct pcie_id : request
   revision_to_string(const result_type& value)
   {
     // The cast is required. This is a boost bug. https://github.com/boostorg/format/issues/60
-    return boost::str(boost::format("%02x") % (uint16_t) value.revision_id);
+    return boost::str(boost::format("%02x") % static_cast<uint16_t>(value.revision_id));
   }
 
   static std::string
   to_path(const result_type& value)
   {
     // The cast is required. This is a boost bug. https://github.com/boostorg/format/issues/60
-    return boost::str(boost::format("%04x_%02x") % value.device_id % (uint16_t) value.revision_id);
+    return boost::str(boost::format("%04x_%02x") % value.device_id % static_cast<uint16_t>(value.revision_id));
   }
 };
 
--- a/src/runtime_src/core/pcie/linux/device_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/device_linux.cpp
@@ -208,7 +208,7 @@ struct bdf
   }
 };
 
-struct id
+struct pcie_id
 {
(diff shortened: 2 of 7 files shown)


src/runtime_src/core/tools/common/tests/TestTCTAllColumn.cpp
```diff
@@ -78,15 +78,20 @@ TestTCTAllColumn::run(std::shared_ptr<xrt_core::device> dev)
...
-  ptree.put("xclbin", m_xclbin);
+  auto device_id = xrt_core::device_query<xrt_core::query::pcie_device>(dev);
+  switch (device_id) {
+  case 5378: // 0x1502
+    ptree.put("xclbin", "validate_phx.xclbin");
+    break;
+  case 6128: // 0x17f0
+    ptree.put("xclbin", "validate_stx.xclbin");
+    break;
+  }
```
maxzhen, line 93: Don't you need revision ID here as well? The same xclbin does not work for both A0 and B0.

src/runtime_src/core/common/query_requests.h
```diff
@@ -466,6 +467,44 @@ struct pcie_bdf : request
...
+  device_to_string(const result_type& value)
+  {
+    return boost::str(boost::format("%04x") % value.device_id);
+  }
+
+  static std::string
+  revision_to_string(const result_type& value)
+  {
+    // The cast is required. This is a boost bug. https://github.com/boostorg/format/issues/60
+    return boost::str(boost::format("%02x") % (uint16_t) value.revision_id);
```
stsoe, line 497: prefer static_cast for narrowing casts

src/runtime_src/core/common/query_requests.h
```diff
@@ -466,6 +467,44 @@ struct pcie_bdf : request
...
+  }
+
+  static std::string
+  to_path(const result_type& value)
+  {
+    // The cast is required. This is a boost bug. https://github.com/boostorg/format/issues/60
+    return boost::str(boost::format("%04x_%02x") % value.device_id % (uint16_t) value.revision_id);
```
stsoe, line 504: ditto

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -127,6 +208,24 @@ struct bdf
   }
 };
 
+struct id
```
stsoe, line 211: Uhh, `id` is very generic, what is it in this context?  Maybe be a little more descriptive like `pcie_id` ?

src/runtime_src/core/pcie/linux/device_linux.cpp
```diff
@@ -1426,6 +1444,7 @@ initialize_query_table()
   emplace_sysfs_get<query::xocl_errors>                        ("", "xocl_errors");
 
   emplace_func0_request<query::pcie_bdf,                       bdf>();
+  emplace_func0_request<query::pcie_id,                        id>();
```
stsoe, line 1447: I think `pcie_id` would be better?
```suggestion
  emplace_func0_request<query::pcie_id,                        pcie_id>();
```

stsoe (changes requested):

src/runtime_src/core/tools/common/tests/TestTCTAllColumn.cpp
dbenusov-xilinx: This bit of code runs only on Windows, and the testcase state is unknown to me @AShivangi This will definitely be fixed as part of a later PR.

stsoe (approved):

PR #7972, opened 2024-02-27
Fix seg fault when device not available is provided

rbramand-xilinx:
when device id not available is given to xrt::device object creation we get a seg fault, this PR fixes that issue

It was discovered while refactoring XDP code to remove xcl apis

This issue can also be reproduced with simple application :
If a system has say 1 device and we try to create 2nd device (index >= 1) we get seg fault
`auto dev = xrt::device(1);`

Fixed the if else condition, now code will not access mgmt list when is_user is true
Also throwing an exception if device with given index is not available as top functions which use get_pcidev directly access ptr returned by this function. 
Checked all the functions whose code flow uses this function and all of them are either 
1. catching exception  (or)
2. expected to throw exception (eg: xrt classes object creation)

Low but needs further testing

Tested on multi device machine for pcie hw flow

NA

### DIFF
--- a/src/runtime_src/core/pcie/linux/system_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/system_linux.cpp
@@ -178,9 +178,8 @@ get_device_id(const std::string& bdf) const
   if (bdf.find_first_not_of("0123456789") == std::string::npos)
     return system::get_device_id(bdf);
 
-  unsigned int i = 0;
   try {
-    while (1) {
+    for (unsigned int i = 0;; i++) {
       auto dev = get_pcidev(i);
       // [dddd:bb:dd.f]
       auto dev_bdf = boost::str(boost::format("%04x:%02x:%02x.%01x") % dev->m_domain % dev->m_bus % dev->m_dev % dev->m_func);
--- a/src/runtime_src/core/pcie/linux/system_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/system_linux.cpp
@@ -96,23 +96,23 @@ std::shared_ptr<pci::dev>
 system_linux::
 get_pcidev(unsigned index, bool is_user) const
 {
-  try {
-    if (is_user) {
-      if (index < user_ready_list.size())
-        return user_ready_list[index];
+  if (is_user) {
+    if (index < user_ready_list.size())
+      return user_ready_list[index];
 
-      if ((index - user_ready_list.size()) < user_nonready_list.size())
-        return user_nonready_list.at(index - user_ready_list.size());
-    }
-
+    if ((index - user_ready_list.size()) < user_nonready_list.size())
+      return user_nonready_list.at(index - user_ready_list.size());
+  }
+  else {
     if (index < mgmt_ready_list.size())
       return mgmt_ready_list[index];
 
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -179,20 +179,24 @@ get_device_id(const std::string& bdf) const
     return system::get_device_id(bdf);
 
   unsigned int i = 0;
-  for (auto dev = get_pcidev(0); dev; dev = get_pcidev(++i)) {
+  try {
+    while (1) {
+      auto dev = get_pcidev(i);
```
stsoe, line 184: Hmm,  I get the try/catch, but not the `while (1)`,   Seems like `i` is never incremented so this would be an infinite loop?   Isn't the for loop better?

stsoe (changes requested):

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -96,23 +96,23 @@ std::shared_ptr<pci::dev>
...
-    return mgmt_nonready_list.at(index - mgmt_ready_list.size());
-  }
-  catch (const std::exception&) {
-    return nullptr;
+    if ((index - mgmt_ready_list.size()) < mgmt_nonready_list.size())
+      return mgmt_nonready_list.at(index - mgmt_ready_list.size());
   }
+
+  // given index is not present in list
+  throw std::runtime_error(" No such device with index '"+ std::to_string(index) + "'");
```
stsoe, line 115: Does it make sense to throw?  Or is nullptr better?

src/runtime_src/core/pcie/linux/system_linux.cpp
```diff
@@ -178,21 +178,24 @@ get_device_id(const std::string& bdf) const
   if (bdf.find_first_not_of("0123456789") == std::string::npos)
     return system::get_device_id(bdf);
 
-  unsigned int i = 0;
-  for (auto dev = get_pcidev(0); dev; dev = get_pcidev(++i)) {
+  try {
+    for (unsigned int i = 0;; i++) {
+      auto dev = get_pcidev(i);
```
stsoe, line 183: The reason I asked about throw or nullptr is this code.  It is not cool to rely on exceptions in success path, which is  kind of what this code is doing in order to break the loop.    It's nitpicking, but still.

rbramand-xilinx: I agree soren but the reason why I have implemented this way is because there are lot of functions which directly use pointer returned from get_pcidev() function without checking it for non null. In order to eliminate lot of changes I used this way.

stsoe: I see and in these cases the emitted exception is eventually caught?

stsoe (approved):

src/runtime_src/core/pcie/linux/system_linux.cpp
rbramand-xilinx: Yes Soren, upper layer catches the exception thrown

chvamshi-xilinx (approved):

PR #7979, opened 2024-02-28
Workaround to limit Aperture Size Limit as part of QDMA Queues for SOFT QDMA IP. 

karthdmg-xilinx:
The SOFT IP on RAVE Platform is unable to handle DMA requests if the block size exceeds 64k.
HW team is working towards the fix in IP.
Adding a temporary WA as part of XRT QDMA subdev until we have a fix in SOFT IP. 
The SOFT IP on RAVE Platform is unable to handle DMA requests if the block size exceeds 64k.
Added a temporary WA as part of XRT QDMA subdev until we have a fix in SOFT IP. 
NA
-Tested the aperture size is being set to 32k size for SOFT IPs.
- Tested on HARD IP Platform (V70) to verify this WA has no effect.
None.


src/runtime_src/core/pcie/driver/linux/xocl/subdev/qdma.c
```diff
@@ -541,6 +555,11 @@ static int set_max_chan(struct xocl_qdma *qdma, u32 count)
 		qconf->qidx = qidx;
 		qconf->irq_en = (qdma->dev_conf.qdma_drv_mode == POLL_MODE) ?
 					0 : 1;
+		if (is_soft_qdma)
+		{
```
chvamshi-xilinx, line 559: Please use "{" along with if condition.

PR #7984, opened 2024-02-29
Use xrt::ip to read registers for XDP PL Deadlock Diagnosis

IshitaGhosh:
HW PL Deadlock Diagnosis still used xclRegRead and other xcl APIs to read deadlock diagnosis status from device. Now, xrt::ip and  its methods are used to read register from device.

Instantiation of xrt::ip in user space XDP plugin required HW Context in shared mode. So, update device for Deadlock Plugin is now called when a new HW Context is created. We keep track whether the current HW Context is for a device which is already updated in XDP Static DB. Deadlock poll thread is only created for a new device load.

PCIe HW, HW Emu, Client, Edge Device


jvillarre (approved):

jvillarre:
Changes are localized to profiling code and profiling hooks, no expected affect on normal operation.

PR #7989, opened 2024-03-05
UE-66864: HEAD CustTA IVT's are failing from XRT:202410.2.17.216

saumyag-xilinx:
UE-66864
Low
Testcase compilation on linkage with libadf.a and libxrt_coreutil.so both
NA


chvamshi-xilinx:
TA is blocked due to this, Merging this change.

chvamshi-xilinx:
We need to find a mechanism to move these functions back to extern "C"

PR #7995, opened 2024-03-08
VITIS-8728 Update skd files to remove xcl apis

rbramand-xilinx:
Removed xcl apis from skd flow

Conversion of xclBufferHandle to xrtBufferHandle is deprecated
So added internal shim function which converts buffer fd to buffer_handle
This way use of xcl buffer shim apis can be eliminated.

Low

Tested pcie and edge  builds by hiding symbols exported by xrt.h header file. Needs additional testing with ps kernels test cases

NA

### DIFF
--- a/src/runtime_src/core/edge/ps_kernels/profiling/aie_trace_gmio/gmio_config_kernel.cpp
+++ b/src/runtime_src/core/edge/ps_kernels/profiling/aie_trace_gmio/gmio_config_kernel.cpp
@@ -19,6 +19,7 @@
 #include "xaiefal/xaiefal.hpp"
 #include "core/edge/include/pscontext.h"
 #include "core/edge/user/shim.h"
+#include "core/include/shim_int.h"
 #include "xdp/profile/database/static_info/aie_constructs.h"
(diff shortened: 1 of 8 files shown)


stsoe (approved):

chvamshi-xilinx (changes requested):
Looks good. 
I dont see a change to link with xrt_coreutil instead of xrt_core.

rbramand-xilinx:
> Looks good. I dont see a change to link with xrt_coreutil instead of xrt_core.
We still have functions like xclGetHostBO, xclGetSKCmd which are from xrt_core so I haven't changed the linking part, I will create a CR which can be used for rewriting SKD code. At the time time of handling the CR I will create apis at xrt_coreutil level to replace these, then we can remove xrt_core linking.

chvamshi-xilinx:
> We still have functions like xclGetHostBO, xclGetSKCmd which are from xrt_core so I haven't changed the linking part, I will create a CR which can be used for rewriting SKD code. At the time time of handling the CR I will create apis at xrt_coreutil level to replace these, then we can remove xrt_core linking.

Ok. Thanks Rahul. Approving.

chvamshi-xilinx (approved):

PR #7997, opened 2024-03-08
VITIS-11369 Improve get xclbin name method

dbenusov-xilinx:
https://jira.xilinx.com/browse/VITIS-11369

Currently, there is a division on where and how paths/files are named between Windows and Linux. To solve this issue we utilize the existing `xclbin_repo_paths` function. However, that is no longer sufficient and additional upgrades are necessary.

Not a bug, more of an inconvenience brought upon by the Windows driver store and Linux driver store rules.

There are a few important changes to note:

1. `xclbin_repo_paths` now returns a vector of viable paths. Rather than a single path. On Linux device specific files can be found in multiple locations.
2. Added a device query to allow for the user to query a device for the appropriate xclbin to load. This allows us to dynamically select the xclbin that will do onto the device, rather than forcing the caller to know what xclbin is needed. As a side effect, tests for new devices will now specify an xclbin type they wish to load instead of the xclbin name. The device shim layer can then construct a repository appropriate solution rather than relying on XRT.

Hopefully none.

Tested with Linux NPU
`????` Has replaced information that is not relevant to the change.
```
Test 1 [0000:c3:00.1]     : df-bw
    Description           : Run bandwidth test on data fabric
    Xclbin                : /lib/firmware/amdnpu/??????/validate.xclbin
    DPU-Sequence          : /opt/xilinx/xrt/amdxdna/DPU_Sequence/df_bw.txt
    Details               : Total duration: '??????'s
                            Average bandwidth per shim DMA: '????' GB/s
Validation completed
```
None

### DIFF
--- a/src/runtime_src/core/common/module_loader.cpp
+++ b/src/runtime_src/core/common/module_loader.cpp
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright (C) 2016-2020 Xilinx, Inc
 // Copyright (C) 2024 Advanced Micro Devices, Inc. All rights reserved.
+
 #define XRT_CORE_COMMON_SOURCE
 #include "core/common/module_loader.h"
 
@@ -101,15 +102,15 @@ xilinx_xrt()
   return xrt;
 }
 
-// Get list of xclbin repository paths from ini file and append
+// Get list of platform repository paths from ini file and append
 // default repository paths
 std::vector<sfs::path>
-get_xclbin_repo_paths()
+get_platform_repo_paths()
 {
   std::vector<sfs::path> paths;
   
-  // Get repo path from ini file if anya
-  auto repo = xrt_core::config::get_xclbin_repo();
+  // Get repo path from ini file if any
+  auto repo = xrt_core::config::get_platform_repo();
   auto token = std::strtok(repo.data(), ":;");
   while (token) {
     paths.push_back(token);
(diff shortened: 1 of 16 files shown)


stsoe (approved):
Looks good to me.

src/runtime_src/core/common/module_loader.cpp
```diff
@@ -113,65 +102,63 @@ xilinx_xrt()
...
   auto token = std::strtok(repo.data(), ":;");
   while (token) {
     paths.push_back(token);
     token = std::strtok(nullptr, ":;");
   }
 
   // Append default path(s)
-  paths.emplace_back(xrt_core::detail::xclbin_repo_path());
+  const auto default_paths = xrt_core::detail::platform_repo_path();
+  paths.insert(paths.end(), default_paths.begin(), default_paths.end());
```
AShivangi, line 122: Just for my knowledge, what is this change doing?

AShivangi (approved):
Looks good. Thanks for the change

chvamshi-xilinx (approved):
Looks good

src/runtime_src/core/common/module_loader.cpp
dbenusov-xilinx: https://en.cppreference.com/w/cpp/container/vector/insert

This inserts the elements between default_paths.begin() and default_paths.end() at paths.end(). Pretty much appends default paths onto paths in a single line.

PR #7998, opened 2024-03-11
Hide symbols from xrt.h header

rbramand-xilinx:
Symbols in xrt.h are made hidden. This way applications can't use these symbols anymore and in future we can stop shipping xrt.h.

Added new macro XCL_DRIVER_DLLHIDDEN and marked all functions in xrt.h with this macro. Also removed functions from opencl that were using Hal operations and made them to use function from xrt_core::device
Also added new functions to xrt_binding.py to make bindings test work
CR - https://jira.xilinx.com/browse/CR-1193149 can be used for further cleanup like SKD

Removed Hal operations completely

Low

Tested pcie and edge build flows


chvamshi-xilinx (approved):
Looks good. Long pending work :-)

stsoe (approved):
Wonderful.  Thanks!

rbramand-xilinx:
@stsoe , @chvamshi-xilinx after pipeline tests failed I realized that OpenCl implementation uses dlopen and dlsym for xclProbe, xclVersion and xclGetDeviceInfo2 symbols. 
Locally I ran xrt native apis tests so dint catch this issue.
For now I have made these symbols visible and OpenCl tests passed locally.

chvamshi-xilinx:
> @stsoe , @chvamshi-xilinx after pipeline tests failed I realized that OpenCl implementation uses dlopen and dlsym for xclProbe, xclVersion and xclGetDeviceInfo2 symbols. Locally I ran xrt native apis tests so dint catch this issue. For now I have made these symbols visible and OpenCl tests passed locally.

We should be able to remove first 2 APIs dependency. 
1) Calling enumerate devices API
2) Remove calling Version as we dont support old versions anymore. You can assume, by default, HAL version is 2.
3) what is the information we are using from device info ?

src/runtime_src/xrt/device/hal2.cpp
```diff
@@ -50,9 +51,16 @@ send_exception_message(const std::string& msg)
...
-  : m_ops(std::move(ops)), m_idx(idx), m_devinfo{}
+device(std::shared_ptr<operations> ops, unsigned int idx, const std::string& dll)
+  : m_ops{std::move(ops)}
+  , m_idx(idx)
+  , mFileName{std::move(dll)}
+  , m_devinfo{}
 {
+  // creating xrt device ahead so that other functions can
+  // be called using the same
+  open();
```
stsoe, line 62: This fundamentally changes when a device is opened.
I am hoping this is okay, but there must be some reason why it was deferred earlier?

I am hesitant with this change.  Can we wrap getting the xrt_core::device* in a function that returns xrt_core::device*.  Something like:

```
xrt_core::device*
get_core_device_raw() 
{
  if (auto cdev = m_handle.get_handle().get())  // xrt_core::device*
    return cdev;

  // if xrt::device has not be created it is wrong to call this function
  throw std::runtime_error("Internal error: device has not been opened");
}
```
All instances of m_core_dev_hdl->zzz() to be replaced with get_core_device_raw()->zzz()

src/runtime_src/xrt/device/hal2.h
```diff
@@ -58,15 +62,17 @@ class device : public xrt_xocl::hal::device
 
   std::shared_ptr<hal2::operations> m_ops;
   unsigned int m_idx;
+  std::string mFileName;
```
stsoe, line 65:
```suggestion
  std::string m_filename;
```

src/runtime_src/xrt/device/hal2.h
```diff
@@ -58,15 +62,17 @@ class device : public xrt_xocl::hal::device
...
 
   xrt::device m_handle;
+  std::shared_ptr<xrt_core::device> m_core_dev_hdl;
```
stsoe, line 68: No sure we need this one, see other comment.

src/runtime_src/xrt/device/hal.cpp
```diff
@@ -133,6 +136,9 @@ is_noop_emulation()
 static void
 createHalDevices(hal::device_list& devices, const std::string& dll, unsigned int count=0)
 {
+  if (!count)
+    count = xrt::system::enumerate_devices();
+
```
stsoe, line 141: Do we need the below dlopen/dlclose?  Do we need handle?

src/runtime_src/xrt/device/hal.cpp
```diff
@@ -143,33 +149,9 @@ createHalDevices(hal::device_list& devices, const std::string& dll, unsigned int
...
-    if (version==1) {
-      throw std::runtime_error("Legacy HAL version " + std::to_string(version) + " not supported");
-    }
-    else if (version==2) {
-      hal2::createDevices(devices,dll,handle.release(),count);
-    }
-    else
-      throw std::runtime_error("HAL version " + std::to_string(version) + " not supported");
-  }
+  hal2::createDevices(devices, dll, handle.release(), count);
```
stsoe, line 152: Confused about the handle here?
I think there should be no need to dlopen shim after everything goes through xrt_core::device?

src/runtime_src/xrt/device/halops2.cpp
```diff
@@ -25,121 +25,18 @@ operations(const std::string &fileName, void *fileHandle, unsigned int count)
   : mFileName(fileName)
   ,mDriverHandle(fileHandle)
   ,mDeviceCount(count)
-  ,mProbe(0)
```
stsoe, line 28: Can we get rid of halop2.{h,cpp} and halops2_static.cpp?

stsoe (changes requested):
Some questions.   Regardless of what we do here, we need thorough testing.  Let me try resurrect my OpenCL test suite for Alveo U200.  I will run it now without your changes so we have a base line.

rbramand-xilinx:
> Some questions. Regardless of what we do here, we need thorough testing. Let me try resurrect my OpenCL test suite for Alveo U200. I will run it now without your changes so we have a base line.

That would be great. Thanks Soren

src/runtime_src/xrt/device/halops2.cpp
rbramand-xilinx: Yes Soren, I will completely remove hal ops in my next PR.
For some 5 functions there is no 1-1 mapping functions which I can call as an alternate so did it this way.

src/runtime_src/xrt/device/hal2.cpp
```diff
@@ -231,27 +238,37 @@ void
...
-      throw std::runtime_error(std::string("failed to release CU(")
-                               + std::to_string(cuidx)
-                               + ") context '"
-                               + std::strerror(errno)
-                               + "'");
+  try {
+    get_core_device()->close_context(uuid, cuidx);
+  }
+  catch (const std::runtime_error& e) {
+    throw std::runtime_error(e.what());
```
stsoe, line 264:
```suggestion
  catch (const std::exception&) {
    throw; // rethrow
```

src/runtime_src/xrt/device/hal2.cpp
```diff
@@ -275,20 +292,25 @@ allocExecBuffer(size_t sz)
...
 
   auto ubo = std::make_unique<ExecBufferObject>();
-  ubo->handle = m_ops->mAllocBO(m_handle,sz, 0, XCL_BO_FLAGS_EXECBUF);  // xrt_mem.h
-  if (ubo->handle == NULLBO)
+  try {
+    ubo->handle = get_core_device()->alloc_bo(sz, XCL_BO_FLAGS_EXECBUF);
+  }
+  catch (const std::runtime_error& e) {
+    throw std::runtime_error(e.what());
+  }
```
stsoe, line 305:
```suggestion
  catch (const std::exception&) {
    throw; // rethrow;
  }
```

src/runtime_src/xrt/device/hal2.cpp
```diff
@@ -386,14 +408,26 @@ size_t
...
 write_register(size_t offset, const void* buffer, size_t size)
 {
-  return m_ops->mWrite(m_handle, XCL_ADDR_KERNEL_CTRL, offset, buffer, size);
+  try {
+    get_core_device()->xwrite(XCL_ADDR_KERNEL_CTRL, offset, buffer, size);
+    return size;
+  }
+  catch (const xrt_core::system_error& e) {
+    return e.get_code();
+  }
```
stsoe, line 430:
```suggestion
   get_core_device()->xwrite(XCL_ADDR_KERNEL_CTRL, offset, buffer, size);
   return size;
  }
```

stsoe (changes requested):
Few changes. Please review if suggestions are applicable in other places also.

keryell (commented):
Interesting.
You could remove a lot of boiling plate code by using functions.

src/runtime_src/xrt/device/hal2.h
```diff
@@ -316,216 +321,208 @@ class device : public xrt_xocl::hal::device
...
-    if (!m_ops->mRead)
-      return hal::operations_result<void>();
-    m_ops->mRead(m_handle, space, offset, hostBuf, size);
-    return hal::operations_result<void>(0);
+    try {
+      get_core_device()->xread(space, offset, hostBuf, size);
+      return static_cast<int>(size);
+    }
+    catch (...) {
+      return hal::operations_result<void>(0);
```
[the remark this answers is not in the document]
rbramand-xilinx: Thanks @keryell for the suggestion I have added template function to remove boiler plate code. Please review

src/runtime_src/xrt/device/hal2.cpp
```diff
@@ -216,7 +216,13 @@ std::shared_ptr<xrt_core::device>
 device::
 get_core_device() const
 {
-  return m_handle.get_handle();
+  if (m_handle) {
+    if (auto cdev = m_handle.get_handle())
+      return cdev;
+  }
```
stsoe, line 222:
```suggestion
    if (auto cdev = m_handle.get_handle())
      return cdev;
```
m_handle is an xrt::device, you can safely get the pimpl (get_handle()) from it, but the pimpl can be nullptr if the device object is not initialized (opened).

src/runtime_src/xrt/device/hal2.cpp
```diff
@@ -231,27 +237,37 @@ void
...
-    if (m_ops->mOpenContext(m_handle,uuid.get(),cuidx,shared))
-      throw std::runtime_error(std::string("failed to acquire CU(")
-                               + std::to_string(cuidx)
-                               + ") context '"
-                               + std::strerror(errno)
-                               + "'");
+  try {
+    get_core_device()->open_context(uuid, cuidx, shared);
+  }
+  catch (const xrt_core::system_error&) {
```
stsoe, line 243: I think you should catch `const std::exception&` to be just a bit broader before catching `...`, The std::exception may carry more information compared to default `...`.

stsoe (changes requested):
Awesome.  Thank you.   One recommended change.
I did run suite of OCL tests and they passed no problem.

chvamshi-xilinx (approved):
Great !!!

PR #8003, opened 2024-03-11
Do not move an r-value

keryell:

This triggered a warning about pessimistic behavior preventing copy-elision.
It was treated as an error by the normal compilation flow.

Just trying to compile with a recent compiler.


stsoe:
ok to test

stsoe (approved):
Thank you.

keryell:
I have signed-off the commit now.

PR #8005, opened 2024-03-12
Port kernel driver to Linux 6.8

keryell:
To be rebased on `main` once https://github.com/Xilinx/XRT/pull/8003 has landed.
I think the code would benefit from some review by DRM experts because I have 0 expertise. @superm1 @airlied
This is to have https://github.com/amd/xdna-driver working along 
https://github.com/amd/xdna-driver/issues/3

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
@@ -1583,7 +1583,7 @@ static int xocl_cma_mem_alloc(struct xocl_dev *xdev, uint64_t size)
 		return -EINVAL;
 	}
 
-	if (page_sz > (PAGE_SIZE << (MAX_ORDER-1))) {
+	if (page_sz > (PAGE_SIZE*MAX_ORDER_NR_PAGES)) {
 		DRM_WARN("Unable to allocate with page size 0x%llx", page_sz);
 		return -EINVAL;
 	}
--- a/src/runtime_src/core/edge/drm/zocl/common/zocl_drv.c
+++ b/src/runtime_src/core/edge/drm/zocl/common/zocl_drv.c
@@ -952,6 +952,11 @@ const struct vm_operations_struct zocl_bo_vm_ops = {
 	.close = drm_gem_vm_close,
 };
 
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 8, 0)
+/* This was removed in 6.8 */
+#define DRM_UNLOCKED 0
+#endif
+
 static const struct drm_ioctl_desc zocl_ioctls[] = {
 	DRM_IOCTL_DEF_DRV(ZOCL_CREATE_BO, zocl_create_bo_ioctl,
 			DRM_AUTH|DRM_UNLOCKED|DRM_RENDER_ALLOW),
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/feature_rom.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/feature_rom.c
@@ -389,18 +389,15 @@ static bool is_valid_firmware(struct platform_device *pdev,
 	return true;
 }
 
+/* Return the length of the string or -E2BIG in case of error */
 static int get_vendor_firmware_dir(u16 vendor, char *buf, size_t len) {
-	size_t ret;
 	switch (vendor) {
 		case XOCL_ARISTA_VEN:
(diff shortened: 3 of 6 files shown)


keryell:
I have signed-off the commit now.
This PR is not supposed to be squashed as its commits and their messages are independent.

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
```diff
@@ -1583,8 +1583,14 @@ static int xocl_cma_mem_alloc(struct xocl_dev *xdev, uint64_t size)
...
 
+/* TODO: check that MAX_PAGE_ORDER is actually well used and this code is not
+ * too pessimistic */
+#if KERNEL_VERSION(6, 8, 0) <= LINUX_VERSION_CODE
+	if (page_sz > (PAGE_SIZE << (MAX_PAGE_ORDER-1))) {
+#else
 	if (page_sz > (PAGE_SIZE << (MAX_ORDER-1))) {
-		DRM_WARN("Unable to allocate with page size 0x%llx", page_sz);
+#endif
+	DRM_WARN("Unable to allocate with page size 0x%llx", page_sz);
```
houlz0507, line 1593: indent?

keryell:
I do not know if removing `DRM_UNLOCKED` which is no longer in use since Linux 6.3 is OK, since the driver is supposed to be compiled also on some antique kernels where it might be required.
I have no idea about the big picture of this driver.

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
houlz0507: Oh. Thanks for mention this. NR_PAGE_ORDERS should be used instead of MAX_PAGE_ORDER.

houlz0507:
> I do not know if removing `DRM_UNLOCKED` which is no longer in use since Linux 6.3 is OK, since the driver is supposed to be compiled also on some antique kernels where it might be required. I have no idea about the big picture of this driver.

It should not be unconditionally removed. Maybe just add a Macro like
#if KERNEL_VERSION(6, 8, 0) <= LINUX_VERSION_CODE
    #define DRM_UNLOCKED 0
#endif

keryell:
> It should not be unconditionally removed. Maybe just add a Macro

Yes I can do that.

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
keryell: I put some links to some discussion in each commit message for reference.
So, after rereading the literature and some changes like https://lwn.net/ml/linux-kernel/20231228144704.14033-2-kirill.shutemov@linux.intel.com/
I think the XRT code was not pessimistic about the allocation (and at least avoided the floppy-disk bughttps://lwn.net/ml/linux-mm/20230315113133.11326-4-kirill.shutemov@linux.intel.com/  ;-) )
So now the modern way to write it is just
```c
if (page_sz > (PAGE_SIZE << MAX_ORDER_NR_PAGES)) {
```
by using definitions from https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/include/linux/mmzone.h?id=e8f897f4afef0031fe618a8e94127a0934896aba#n34

But actually this is the definitive answer that seems to have been used for quite some Linux releases, as it it is in 5.4 and 4.18.

houlz0507 (approved):

keryell:
This is not clear why the test on `pcie-hw-tests (xilinx_u250_gen3x16_xdma_4_1_202210_1)` is not working.
How to investigate this?

keryell:
XOAH shows *ERROR: Not enough host mem. Please check grub settings.: Cannot allocate memory* on `pcie-hw-tests (xilinx_u250_gen3x16_xdma_4_1_202210_1)`.

keryell:
OK, the problem was that in [`af7a239` (#8005)](https://github.com/Xilinx/XRT/pull/8005/commits/af7a2393297f92246b33623c8bd19c7997b58adc) I forgot to replace the `<<` by `*`, so there was a UB operation due to the overflow, probably leading 0, so the allocation was always too big...
Ahem.

keryell:
This PR is not supposed to be squashed as its commits and their messages are independent.

chvamshi-xilinx (approved):

mamin506 (approved):

keryell:
So, what is the problem on this PR?

keryell:
@stsoe Thanks!

PR #8010, opened 2024-03-13
Dump instruction and control packet (Do not merge)

dezhiAmd:
This is for discussion.
It used to compare the dumped binary file from this request: 
https://gitenterprise.xilinx.com/XRT/testcases/pull/380/

Compare 
bo_instr.bin with m_instr_buf.bin
bo_mc.bin with m_ctrlpkt_buf.bin

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -808,6 +808,19 @@ class module_sram : public module_impl
     // copy instruction into bo
     fill_instr_buf(m_instr_buf, data);
 
+    {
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -834,6 +847,19 @@ class module_sram : public module_impl
 
       // copy instruction into bo
       fill_ctrlpkt_buf(m_ctrlpkt_buf, data);
+      {
+          std::string filename = "m_ctrlpkt_buf.bin";
```
dezhiAmd, line 851: control packet from elf file

PR #8021, opened 2024-03-18
Incorporate partition shifts and support ML metadata

pgschuey:
* AIE partitions on client devices are reported in profile/trace with relative columns
* Compiler metadata for ML designs needs expanded support

N/A

* Device is queried for AIE partition information and reported when found
* Added support for proposed ML metadata

* New ML metadata must contain all required information since only one metadata reader is supported

* Tested on vek280 and Ryzen AI

Mention partition shifting support in AIE profile/trace on client devices


jvillarre (approved):
Approving on condition that the stray debug cout statement in client/aie_trace.cpp is removed in a future pull request.

PR #8029, opened 2024-03-20
Hip test tidy up

sonals:

HIP test clang-tidy fixes and consolidation of common code 

NA

Code cleanups

None

1. XRT compile tested 
2. Validated on GPU

NA


tests/hip/common/common.h
```diff
@@ -147,14 +161,13 @@ hip_test_device {
 
   hipFunction_t
   get_function(const char *fileName, const char *funcName) {
-    std::map<std::string, hipModule_t>::iterator it = mModuleTable.find(fileName);
-    hipModule_t hmodule;
-    hmodule = it->second;
+    auto it = mModuleTable.find(fileName);
+    hipModule_t hmodule = it->second;
     if (it == mModuleTable.end()) {
```
keryell, line 166: There is a bug here probably. How it is possible to use `it->second` and only later testing the validity of `it`?

keryell (commented):
Thanks for contributing to a cleaner planet!

tests/hip/common/common.h
sonals: I have validated that reading `it->second` is fine provided we do not use that value if the iterator, `it` is pointing to `end()`.  Logically it does look wrong however, as if we are trying to read the value past the end of the map. I will update the PR.

keryell: The problem is that then the compiler can do time-travel optimization. \o/
Since we write perfect_™ software_™ which does not have undefined behavior, `hipModule_t hmodule = it->second;` is perfect, so `it == mModuleTable.end()` can never be true, so the test is dead code and removed. :-|

sonals: Updated :-)

chvamshi-xilinx (approved):

PR #8033, opened 2024-03-22
Fix on client to correct partition info device query request

vipangul:
On client, evice query was running into errors while fetching the partition info

Corrected the way to get hardware context on client instead of device handle.

No risk as it is verified on client MCDM.

Verified on client MCDM XRT.

### DIFF
--- a/src/runtime_src/xdp/profile/database/static_info/aie_util.cpp
+++ b/src/runtime_src/xdp/profile/database/static_info/aie_util.cpp
@@ -371,27 +371,26 @@ namespace xdp::aie {
     try {
       xrt::hw_context context = xrt_core::hw_context_int::create_hw_context_from_implementation(handle);
       auto device = xrt_core::hw_context_int::get_core_device(context);
-      if(device==nullptr)
+      if (device==nullptr)
         return std::vector<uint8_t>{0};
 
       auto infoVector = xrt_core::device_query<xrt_core::query::aie_partition_info>(device.get());
       if (infoVector.empty()) {
         xrt_core::message::send(severity_level::info, "XRT", "No AIE partition information found.");
-        startCols.push_back(0);
+        return std::vector<uint8_t>{0};
       }
-      else {
-        for (auto& info : infoVector) {
-          auto startCol = static_cast<uint8_t>(info.start_col);
-          xrt_core::message::send(severity_level::info, "XRT",
-              "Partition shift of " + std::to_string(startCol) + " was found.");
-          startCols.push_back(startCol);
-        }
+
+      for (auto& info : infoVector) {
+        auto startCol = static_cast<uint8_t>(info.start_col);
+        xrt_core::message::send(severity_level::info, "XRT",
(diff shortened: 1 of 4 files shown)


src/runtime_src/xdp/profile/database/static_info/aie_util.cpp
```diff
@@ -363,13 +364,15 @@ namespace xdp::aie {
...
    ***************************************************************************/
   std::vector<uint8_t>
-  getPartitionStartColumns(void* handle)
+  getPartitionStartColumnsClient(void* handle)
   {
     std::vector<uint8_t> startCols;
 
     try {
-      std::shared_ptr<xrt_core::device> dev = xrt_core::get_userpf_device(handle);
-      auto infoVector = xrt_core::device_query<xrt_core::query::aie_partition_info>(dev);
```
vipangul, line 372: Sorry this code is missed in diff.  it shouldn't delete device_query

jvillarre (approved):

pgschuey (approved):

src/runtime_src/xdp/profile/plugin/aie_profile/client/aie_profile.cpp
```diff
@@ -112,8 +112,11 @@ namespace xdp {
 
     // Get partition columns
     // NOTE: for now, assume a single partition
-    auto partitionCols = xdp::aie::getPartitionStartColumns(metadata->getHandle());
-    auto startCol = partitionCols.at(0);
+    uint8_t startCol = 0;
+#ifdef XDP_CLIENT_BUILD
```
IshitaGhosh, line 116: We don't need this macro dependency check here. The current file is only compiled for Client flow with XDP_CLIENT_BUILD defined.

src/runtime_src/xdp/profile/plugin/aie_trace/client/aie_trace.cpp
```diff
@@ -808,9 +808,11 @@ namespace xdp {
 
     // Get partition columns
     // NOTE: for now, assume a single partition
-    auto partitionCols = xdp::aie::getPartitionStartColumns(handle);
-    auto startCol = partitionCols.at(0);
-
+    uint8_t startCol = 0;
+#ifdef XDP_CLIENT_BUILD
```
IshitaGhosh, line 812: We don't need this macro dependency check here. The current file is only compiled for Client flow with XDP_CLIENT_BUILD defined.

IshitaGhosh (approved):
The change should work fine. 
I have shared some feedback on unnecessary macro checks and calling create_hw_context_from_omplementation when AIE Trace/Profile plugin already has the hw_context in Client flow. But those can be improved later. The current code should  work fine.

src/runtime_src/xdp/profile/plugin/aie_profile/client/aie_profile.cpp
vipangul: Agree, removed the usage.

pgschuey (approved):

PR #8041, opened 2024-04-01
Completed HIP Command Related Functions

haghilin:
Completed HIP Command Related Functions

Problem solved by the commit
Completed HIP Command Related Functions

NA

NA

Documentation impact (if any)
None.

### DIFF
--- a/src/runtime_src/hip/api/hip_event.cpp
+++ b/src/runtime_src/hip/api/hip_event.cpp
@@ -28,11 +28,9 @@ static void hip_event_record(hipEvent_t eve, hipStream_t stream)
 {
   throw_invalid_value_if(!eve, "event passed is nullptr");
   throw_invalid_value_if(!stream, "stream passed is nullptr");
-  /* TODO
   auto hip_stream = get_stream(stream);
   auto hip_ev = std::dynamic_pointer_cast<event>(command_cache.get(eve));
   hip_ev->record(hip_stream);
-  */
 }
 
 static void hip_event_synchronize(hipEvent_t eve)
@@ -51,7 +49,9 @@ static float hip_event_elapsed_time(hipEvent_t start, hipEvent_t stop)
   throw_invalid_value_if(!hip_ev_start, "dynamic_pointer_cast failed");
   auto hip_ev_stop = std::dynamic_pointer_cast<event>(command_cache.get(stop));
   throw_invalid_value_if(!hip_ev_stop, "dynamic_pointer_cast failed");
-  return hip_ev_start->elapsed_time(hip_ev_stop);
+  auto duration = hip_ev_stop->get_time() - hip_ev_start->get_time();
+  auto millis = std::chrono::duration_cast<std::chrono::milliseconds>(duration).count();
+  return static_cast<float>(millis);
 }
 
 static bool hip_event_query(hipEvent_t eve)
--- a/src/runtime_src/hip/core/event.cpp
+++ b/src/runtime_src/hip/core/event.cpp
@@ -15,26 +15,25 @@ void event::record(std::shared_ptr<stream> s)
   cstream = std::move(s);
   auto ev = std::dynamic_pointer_cast<event>(command_cache.get(static_cast<command_handle>(this)));
   throw_invalid_handle_if(!ev, "event passed is invalid");
-  /*
-  TODO
   if (is_recorded()) {
     // already recorded
-    cstream->erase_cmd(ev); It is commented will be uncomment after stream PR is pushed
+    cstream->erase_cmd(ev);
   }
   // update recorded commands list
-  cstream->enqueue_event(std::move(ev));It is commented will be uncomment after stream PR is pushed
-  */
+  cstream->enqueue_event(ev);
   set_state(state::recorded);
 }
 
 bool event::is_recorded() const
 {
+  //the event is recorded only if the state is not init
   return get_state() >= command::state::recorded;
 }
 
 bool event::query()
 {
-  for (auto rec_com : recorded_commands){
+  //This function will return true if all commands in the appropriate stream which specified to hipEventRecord() have completed.
+  for (auto& rec_com : m_recorded_commands){
     state command_state = rec_com->get_state();
     if (command_state != state::completed){
       return false;
@@ -45,11 +44,17 @@ bool event::query()
 
 bool event::synchronize()
 {
-  for (auto rec_com : recorded_commands) {
+  //wait for commands in recorded list of the event to be completed
+  std::lock_guard rec_lock(m_mutex_rec_coms);
+  for (auto& rec_com : m_recorded_commands) {
     rec_com->wait();
   }
+  //then the event is considered as completed
   set_state(state::completed);
-  for (auto coms_ch :chain_of_commands){
+
+  //all commands depend on the event start running
+  std::lock_guard ch_lock(m_mutex_chain_coms);
+  for (auto& coms_ch :m_chain_of_commands){
     coms_ch->submit();
   }
   return true;
@@ -80,19 +85,14 @@ std::shared_ptr<stream> event::get_stream()
 
 void event::add_to_chain(std::shared_ptr<command> cmd)
 {
-  // lock and add
+  std::lock_guard lock(m_mutex_chain_coms);
+  m_chain_of_commands.push_back(std::move(cmd));
 }
 
 void event::add_dependency(std::shared_ptr<command> cmd)
 {
-  // lock and add
-}
-
-float event::elapsed_time (const std::shared_ptr<command>& end)
-{
-  auto duration = end->get_time() - get_time();
-  auto millis = std::chrono::duration_cast<std::chrono::milliseconds>(duration).count();
-  return millis;
+  std::lock_guard lock(m_mutex_rec_coms);
+  m_recorded_commands.push_back(std::move(cmd));
 }
 
 kernel_start::kernel_start(std::shared_ptr<stream> s, std::shared_ptr<function> f, void** args)
@@ -196,6 +196,7 @@ bool copy_buffer::submit()
 bool copy_buffer::wait()
 {
   handle.wait();
+  set_state(state::completed);
   return true;
 }
 
--- a/src/runtime_src/hip/core/event.cpp
+++ b/src/runtime_src/hip/core/event.cpp
@@ -33,6 +33,7 @@ bool event::is_recorded() const
 bool event::query()
 {
   //This function will return true if all commands in the appropriate stream which specified to hipEventRecord() have completed.
+  std::lock_guard lock(m_mutex_rec_coms);
   for (auto& rec_com : m_recorded_commands){
     state command_state = rec_com->get_state();
     if (command_state != state::completed){
--- a/src/runtime_src/hip/core/event.h
+++ b/src/runtime_src/hip/core/event.h
@@ -60,23 +60,42 @@ public:
 
   virtual bool submit() = 0;
   virtual bool wait() = 0;
-  state get_state() const { return cstate; }
-  std::chrono::time_point<std::chrono::system_clock> get_time() { return ctime; }
-  void set_state(state newstate) { cstate = newstate; };
-  type
-  get_type() const { return ctype; }
+  
+  state
+  get_state() const 
+  {
+    return cstate;
+  }
+
+  std::chrono::time_point<std::chrono::system_clock>
+  get_time() 
+  {
+    return ctime;
+  }
+
+  void 
+  set_state(state newstate) 
+  {
+    cstate = newstate;
+  }
+
+  type 
+  get_type() const 
+  {
+    return ctype;
+  }
 };
 
 class event : public command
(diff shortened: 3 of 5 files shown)


src/runtime_src/hip/core/event.cpp
```diff
@@ -80,19 +77,14 @@ std::shared_ptr<stream> event::get_stream()
...
-  // lock and add
-}
-
-float event::elapsed_time (const std::shared_ptr<command>& end)
-{
-  auto duration = end->get_time() - get_time();
-  auto millis = std::chrono::duration_cast<std::chrono::milliseconds>(duration).count();
-  return millis;
+  std::lock_guard<std::mutex> lock(m_mutex);
+  recorded_commands.push_back(cmd);
```
stsoe, line 87: pass by value means you should should preferably std::move(cmd) here to turn cmd into an rvalue.

src/runtime_src/hip/core/event.cpp
```diff
@@ -172,13 +164,15 @@ copy_buffer::copy_buffer(std::shared_ptr<stream> s)
 
 bool copy_buffer::submit()
 {
-  //handle = std::async(&xrt::bo::sync, &cbo, cdirection);
+  handle = std::async(static_cast<void (xrt::bo::*)(xclBOSyncDirection)>(&xrt::bo::sync), &cbo, cdirection);
```
stsoe, line 167: I think this should work and is clearer:
`std::async([&cbo, cdirection] { cbo.sync(cdirection); });`

src/runtime_src/hip/core/event.h
```diff
@@ -62,8 +62,7 @@ class command
   state get_state() const { return cstate; }
   std::chrono::time_point<std::chrono::system_clock> get_time() { return ctime; }
   void set_state(state newstate) { cstate = newstate; };
-  type
-  get_type() const { return ctype; }
+  type get_type() const { return ctype; }
```
stsoe, line 65: This is where I complete disagree with @keryell, I have to scan really really hard to find the functions far to the right, e.g. get_time().  It is much more preferable to format as
```
ReturnType
function() 
{}
```
where everything is on the left margin.

stsoe (changes requested):

src/runtime_src/hip/core/event.cpp
haghilin: Hi Soren, Thanks for the comments.
Since cbo and cdirection are member variables of the copy_buffer class, I think I need to capture "this" for the Lambda expression and then use cbo and cdirection through this, otherwise I've been getting compiler error. Please let me know if it is not a good replace.

src/runtime_src/hip/core/event.h
haghilin: Thanks, I modified the functions thoroughly.

src/runtime_src/hip/core/event.cpp
stsoe: yes, you need to capture "this".  I changed the comment accordingly.

haghilin: Thanks

src/runtime_src/hip/api/hip_event.cpp
```diff
@@ -51,7 +49,9 @@ static float hip_event_elapsed_time(hipEvent_t start, hipEvent_t stop)
   throw_invalid_value_if(!hip_ev_start, "dynamic_pointer_cast failed");
   auto hip_ev_stop = std::dynamic_pointer_cast<event>(command_cache.get(stop));
   throw_invalid_value_if(!hip_ev_stop, "dynamic_pointer_cast failed");
-  return hip_ev_start->elapsed_time(hip_ev_stop);
+  auto duration = hip_ev_stop->get_time() - hip_ev_start->get_time();
+  auto millis = std::chrono::duration_cast<std::chrono::milliseconds>(duration).count();
+  return millis;
```
sonals, line 54: I believe there is an implicit type conversion from long to float here. Is this code clang-tidy clean?

haghilin: Thanks for the comment. I fixed it.

src/runtime_src/hip/core/event.h
```diff
@@ -59,11 +59,22 @@ class command
...
   virtual bool wait() = 0;
-  state get_state() const { return cstate; }
-  std::chrono::time_point<std::chrono::system_clock> get_time() { return ctime; }
-  void set_state(state newstate) { cstate = newstate; };
-  type
-  get_type() const { return ctype; }
+  
+  state
+  get_state() const 
+  { return cstate; }
```
chvamshi-xilinx, line 65: Please move implementation to new line. Fix all the functions
example:
state 
get_state() const
{
  return cstate;
}

chvamshi-xilinx (commented):
Looks good. Changes requested only w.r.t indentation

src/runtime_src/hip/core/event.h
haghilin: Thanks Vamshi, I modified the code.

src/runtime_src/hip/core/event.cpp
```diff
@@ -15,15 +15,12 @@ void event::record(std::shared_ptr<stream> s)
...
-  TODO
   if (is_recorded()) {
     // already recorded
-    cstream->erase_cmd(ev); It is commented will be uncomment after stream PR is pushed
+    cstream->erase_cmd(ev);
   }
   // update recorded commands list
-  cstream->enqueue_event(std::move(ev));It is commented will be uncomment after stream PR is pushed
-  */
+  cstream->enqueue_event(std::move(ev));
```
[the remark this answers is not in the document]
stsoe: Echo the question.  Why is stream signature of enqueue and enqueue_event take rvalue of shared_ptr?  It's not that expensive to change these to shared_ptr and then move the shared ptr in these functions.

I just looked at some of the classes under hip/   There is a serious lack of comments.

src/runtime_src/hip/core/event.cpp
```diff
@@ -80,19 +83,14 @@ std::shared_ptr<stream> event::get_stream()
 
 void event::add_to_chain(std::shared_ptr<command> cmd)
 {
-  // lock and add
+  std::lock_guard<std::mutex> lock(m_mutex);
```
chvamshi-xilinx, line 86: Use different lock for chain_of_commands and recorded commands?

src/runtime_src/hip/core/event.cpp
```diff
@@ -45,10 +44,14 @@ bool event::query()
 
 bool event::synchronize()
 {
+  //wait for commands in recorded list of the event to be completed
```
chvamshi-xilinx, line 47: You need to acquire the lock before accessing these commands.

src/runtime_src/hip/core/event.cpp
```diff
@@ -45,10 +44,14 @@ bool event::query()
...
   for (auto rec_com : recorded_commands) {
     rec_com->wait();
   }
+  //then the event is considered as completed
   set_state(state::completed);
+
+  //all commands depend on the event start running
```
chvamshi-xilinx, line 54: DItto. You need to acquire the  lock

chvamshi-xilinx (changes requested):

src/runtime_src/hip/core/event.cpp
```diff
@@ -15,25 +15,24 @@ void event::record(std::shared_ptr<stream> s)
...
 bool event::is_recorded() const
 {
+  //the event is recorded only if the state is not init
   return get_state() >= command::state::recorded;
 }
 
 bool event::query()
 {
+  //This function will return true if all commands in the appropriate stream which specified to hipEventRecord() have completed.
   for (auto rec_com : recorded_commands){
```
stsoe, line 36: May want to double check and change all range-based for loops to use `auto&` when not a fundamental type.  In this case rec_com is a shared_ptr, copying is not for free.

src/runtime_src/hip/core/event.cpp
```diff
@@ -45,11 +44,17 @@ bool event::query()
...
   for (auto rec_com : recorded_commands) {
+    std::lock_guard<std::mutex> lock(m_mutex_rec_coms);
     rec_com->wait();
   }
+  //then the event is considered as completed
   set_state(state::completed);
+
+  //all commands depend on the event start running
   for (auto coms_ch :chain_of_commands){
+    std::lock_guard<std::mutex> lock(m_mutex_chain_coms);
```
stsoe, line 57:
```suggestion
    std::lock_guard lock(m_mutex_chain_coms);
```

src/runtime_src/hip/core/event.cpp
```diff
@@ -172,13 +172,15 @@ copy_buffer::copy_buffer(std::shared_ptr<stream> s)
 
 bool copy_buffer::submit()
 {
-  //handle = std::async(&xrt::bo::sync, &cbo, cdirection);
+  handle = std::async([this] { this->cbo.sync(this->cdirection); });
+  set_state(state::running);
```
stsoe, line 176: Not necessarily running unless a policy is specified.

src/runtime_src/hip/core/event.cpp
```diff
@@ -172,13 +172,15 @@ copy_buffer::copy_buffer(std::shared_ptr<stream> s)
...
   return true;
 }
 
 bool copy_buffer::wait()
 {
   handle.wait();
```
stsoe, line 182: This is cool, but to above comment, running may be deferred until exactly this point unless std::launch::async is speciied when calling std::async.

src/runtime_src/hip/core/stream.cpp
```diff
@@ -27,7 +27,7 @@ stream::
 
 void
 stream::
-enqueue(std::shared_ptr<command>&& cmd)
+enqueue(std::shared_ptr<command> cmd)
```
stsoe, line 30: Yes, by-value here makes sense since you move in m_cmd_queue, but up the chain it should be `const std::shared_ptr<command>& cmd` to avoid reference count logic.

Incidentally, the std::move on line 51 below should be removed.

stsoe (changes requested):

src/runtime_src/hip/core/event.cpp
```diff
@@ -45,11 +44,17 @@ bool event::query()
 
 bool event::synchronize()
 {
-  for (auto rec_com : recorded_commands) {
+  //wait for commands in recorded list of the event to be completed
+  for (auto& rec_com : m_recorded_commands) {
+    std::lock_guard lock(m_mutex_rec_coms);
```
chvamshi-xilinx, line 49: you should take the lock outside for loop

src/runtime_src/hip/core/event.cpp
```diff
@@ -45,11 +44,17 @@ bool event::query()
...
     rec_com->wait();
   }
+  //then the event is considered as completed
   set_state(state::completed);
-  for (auto coms_ch :chain_of_commands){
+
+  //all commands depend on the event start running
+  for (auto& coms_ch :m_chain_of_commands){
+    std::lock_guard lock(m_mutex_chain_coms);
```
chvamshi-xilinx, line 57: ditto

chvamshi-xilinx (changes requested):

chvamshi-xilinx:
Please rebase with Chiming changes. It has changes related to Async buffer sync.

stsoe (changes requested):
Few tweaks on rvalue/lvalue.

A more general comment/question.  When using smart pointers (shared and unique) one should think clearly about ownership.  Typically you just don't blindly use shared_ptr, if there is one clear owner.   You should not be afraid of raw pointers when ownership is managed by enclosed scope.  E.g. say a class owns a resource (unique_ptr) then it can freely pass a raw pointer (uptr.get()) to functions with this resource provided these functions don't take copies that outlive the actual owner.

I am not suggesting you change anything here.

src/runtime_src/hip/core/stream.cpp
```diff
@@ -27,7 +27,7 @@ stream::
 
 void
 stream::
-enqueue(std::shared_ptr<command>&& cmd)
+enqueue(const std::shared_ptr<command>& cmd)
```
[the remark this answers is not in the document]
haghilin: Thanks Soren, now it makes more sense for me. I applied your suggestion.

src/runtime_src/hip/core/event.cpp
```diff
@@ -15,26 +15,25 @@ void event::record(std::shared_ptr<stream> s)
...
 bool event::is_recorded() const
 {
+  //the event is recorded only if the state is not init
   return get_state() >= command::state::recorded;
 }
 
 bool event::query()
 {
-  for (auto rec_com : recorded_commands){
+  //This function will return true if all commands in the appropriate stream which specified to hipEventRecord() have completed.
```
chvamshi-xilinx, line 35: missed holding the lock

stsoe (approved):
LGTM.  Thank you!

chvamshi-xilinx (approved):

PR #8043, opened 2024-04-01
VITIS-11503 - Dump instruction bo created from Elf so verification can be done

dezhiAmd:
Create a mechanism to dump instruction-buffer BO(s) constructed from elf. This is critical to verify the elf's correctness. 
Only in debug mode, Use the xrt_core::config::get_feature_toggle which reading xrt.ini to decide whether or not dumping instruction-buffer bo. This can avoid any potential performance impact to real applications.
None

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -201,7 +201,7 @@ struct patcher
   }
 };
 
-  void
+  XRT_CORE_UNUSED void
   dump_bo(xrt::bo& bo, const std::string& filename)
   {
     if (!xrt_core::config::get_feature_toggle(Debug_Bo_From_Elf_Feature))
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -3,6 +3,7 @@
 #define XCL_DRIVER_DLL_EXPORT  // exporting xrt_module.h
 #define XRT_API_SOURCE         // exporting xrt_module.h
 #define XRT_CORE_COMMON_SOURCE // in same dll as core_common
+#include "core/common/config_reader.h"
 #include "experimental/xrt_module.h"
 #include "experimental/xrt_elf.h"
 #include "experimental/xrt_ext.h"
@@ -40,6 +41,9 @@ namespace
 static constexpr size_t column_page_size = AIE_COLUMN_PAGE_SIZE;
 static constexpr uint8_t Elf_Amd_Aie2p  = 69;
 static constexpr uint8_t Elf_Amd_Aie2ps = 64;
+
+// When Debug.dump_bo_from_elf is true in xrt.ini, instruction bo(s) from elf will be dumped
+static std::string Debug_Bo_From_Elf_Feature = "Debug.dump_bo_from_elf";
 
 struct buf
 {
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -808,11 +812,42 @@ class module_sram : public module_impl
...
+#ifdef _DEBUG
+      if (xrt_core::config::get_feature_toggle(Debug_Bo_From_Elf_Feature)) {
+          std::string filename = "instrBoPatchedByCtrlPacket.bin";
+          std::ofstream ofs(filename, std::ios::out | std::ios::binary);
+
+          if (!ofs.is_open()) {
+              std::cout << "Failure opening file " + filename + " for writing!!" << std::endl;
+              abort();
+          }
+          char* buf = static_cast<char*>(m_instr_buf.map());
```
stsoe, line 845:
```suggestion
          auto buf = m_instr_buf.map<char*>();
```

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -808,11 +812,42 @@ class module_sram : public module_impl
...
+          ofs.write((char*)buf, m_instr_buf.size());
```
stsoe, line 846:
```suggestion
          ofs.write(buf, m_instr_buf.size());
```

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -808,11 +812,42 @@ class module_sram : public module_impl
...
+          ofs.close();
```
stsoe, line 847:
```suggestion
```

stsoe (changes requested):

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -834,6 +865,20 @@ class module_sram : public module_impl
...
+      if (xrt_core::config::get_feature_toggle(Debug_Bo_From_Elf_Feature)) {
+          std::string filename = "ctrlpktBo.bin";
+          std::ofstream ofs(filename, std::ios::out | std::ios::binary);
+
+          if (!ofs.is_open())
+            throw std::runtime_error("Failure opening file " + filename + " for writing!");
+
+          auto buf = m_ctrlpkt_buf.map<char*>();
+          ofs.write(buf, m_ctrlpkt_buf.size());
+      }
```
stsoe, line 879: This appears to be a repeated code block from above; I didn't notice that before.  If so, please refactor in a separate function that take output filename and xrt::bo to dump.  E.g.
dump_bo(m_ctrl_buf, "ctrlpktbo.bin");

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -790,6 +794,22 @@ class module_sram : public module_impl
...
+      std::ofstream ofs(filename, std::ios::out | std::ios::binary);
+
+      if (!ofs.is_open())
+        throw std::runtime_error("Failure opening file " + filename + " for writing!");
+
+      auto buf = bo.map<char*>();
+      ofs.write(buf, bo.size());
+    }
+  }
+#endif
```
stsoe, line 811: Move to global function under unnamed namespace within this compilation unit as there no need for state access.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c4-make-a-function-a-member-only-if-it-needs-direct-access-to-the-representation-of-a-class

Remove the surrounding `#ifdef _DEBUG` from the function.  It is okay and good to protect the call from create_instr_buf as you are doing.

stsoe (approved):

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -197,6 +201,19 @@ struct patcher
   }
 };
 
+  void
+  dump_bo(xrt::bo& bo, const std::string& filename)
+  {
+    if (xrt_core::config::get_feature_toggle(Debug_Bo_From_Elf_Feature)) {
```
sonals, line 207: Can you invert the check and do an early exit?

stsoe (approved):

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -197,6 +201,19 @@ struct patcher
   }
 };
 
+  void
```
stsoe, line 204: Hi @dezhiAmd .  The build is failing because the function is unused in release builds.
Please do this:
```suggestion
  XRT_CORE_UNUSED void
```

stsoe (changes requested):

PR #8047, opened 2024-04-03
Providing a work-around to allocate instruction buffer in HIP

chvamshi-xilinx:
Providing a work-around to allocate instruction buffer in HIP

Allocating buffer in Bank1 for instruction buffer sothat we can run simple xrt run flow test 

Provided a work-around

None

Verified a simple test

None

### DIFF
--- a/src/runtime_src/hip/core/memory.cpp
+++ b/src/runtime_src/hip/core/memory.cpp
@@ -21,8 +21,7 @@ namespace xrt::core::hip
       : m_device(std::move(dev)),
 	m_size(sz),
 	m_type(memory_type::device),
-	m_flags(0),
-	m_bo(nullptr)
+	m_flags(0)
   {
     assert(m_device);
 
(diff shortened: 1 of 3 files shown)


src/runtime_src/hip/core/memory.cpp
```diff
@@ -57,11 +57,18 @@ namespace xrt::core::hip
...
         init_xrt_bo();
         break;
-
+      }
+      case hipHostMallocWriteCombined: {
+        // This is a workaround to create a buffer with cacheable flag if WriteComined flag is provided. 
+        // This gets used to create instruction buffer on NPU
+        // TODO This would go away once xrt::elf flow is enabled
+        auto xrt_device = m_device->get_xrt_device();
+        m_bo = std::make_shared<xrt::bo>(xrt_device, m_size, xrt::bo::flags::cacheable, 1);
```
stsoe, line 69: I am not following the need to create xrt::bo as a shared_ptr?  xrt::bo is a reference counted impl already.

chvamshi-xilinx: Yes Soren. We dont require it, I thought of changing in my last PR itself, let me change it in this PR.

chvamshi-xilinx: Fixed the issue. Fixed function signatures of some of hip memory class aswell due to this change.
Tried making hip::memory::read as const, but it is calling xrt::bo::read which is non-const, thatswhy i have to remove the const-ness of this function.

PR #8048, opened 2024-04-03
Fix on xocl driver crash during hwmon init on RAVE platform -  CR-1195545

karthdmg-xilinx:
RAVE Platform does not has VMC and hence management driver does not create subdev for hwmon. This is done by verifyingis the paltform has SC. This validation is not handled in xocl driver and with recent changes if xocl driver tries to init hwmon sensor indormation from xclmgmt dev through mailbox, crash occurs.

Fixed to skip hwmon subdev creation is SC is not present.
Fixed to skip hwmon subdev creation is SC is not present.
Facing a permission denied issue while running xbutil examine command. working on it.
Loaded xclmgmt and xocl drivers to verify crash does not occur. verify examine commands and sysfs nodes on both RAVE and v70 platform.
N/A


karthdmg-xilinx:
Reverting XRT changes, so this change is not necessary

PR #8050, opened 2024-04-03

zhangchiming:
Initial implementation for some HIP memory APIs.

Problem solved by the commit
Added Initial implementation for hipMemcpyHtoDAsync().

NA

NA

Compiled  on Ubuntu 22.04.

Documentation impact (if any)
None.


src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -215,6 +217,25 @@ namespace xrt::core::hip
...
+
+    auto hip_mem_dst = memory_database::instance().get_hip_mem_from_addr(dst);
+    if (!hip_mem_dst)
+      throw xrt_core::system_error(hipErrorInvalidValue, "Invalid destination handle in hipMemCpyH2DAsync");
+
+    auto hip_stream = get_stream(stream);
+    if (!hip_stream)
+      throw xrt_core::system_error(hipErrorInvalidValue, "Invalid stream handle in hipMemCpyH2DAsync");
+   
+    copy_buffer cp_buf(hip_stream, XCL_BO_SYNC_BO_TO_DEVICE, hip_mem_dst, src);
```
chvamshi-xilinx, line 235: You should need to create a shared pointer of command and push it to the stream . You should not submit the command in this function directly.
Please check hip_module_launch_kernel function for reference.

zhangchiming: fixed.

chvamshi-xilinx (approved):
Looks good. Please add comments to the functions you have added.

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -307,3 +330,8 @@ hipMemset(void* dst, int value, size_t size)
   return handle_hip_memory_error([&] { xrt::core::hip::hip_memset(dst, value, size); });
 }
 
+hipError_t
+hipMemcpyHtoDAsync(hipDeviceptr_t dst, void* src, size_t size, hipStream_t stream)
+{
+  return handle_hip_memory_error([&] { xrt::core::hip::hip_memcpy_host2device_async(dst, src, size, stream); });
+}
```
chvamshi-xilinx, line 337: Looks like eof is missing for this file

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -215,6 +217,27 @@ namespace xrt::core::hip
...
   }
 
+
+  static void
+  hip_memcpy_host2device_async(hipDeviceptr_t dst, void* src, size_t size, hipStream_t stream)  
+  {
+    if (!src)
+      throw xrt_core::system_error(hipErrorInvalidValue, "Invalid src pointer in hipMemCpyH2DAsync");
+
+    auto hip_mem_dst = memory_database::instance().get_hip_mem_from_addr(dst);
```
sonals, line 227: `dst` may point to somewhere in the middle of `him_mem_dst` BO? Should we not keep track of that offset when we issue the actual copy command?

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -215,6 +217,27 @@ namespace xrt::core::hip
...
+    if (!hip_mem_dst)
+      throw xrt_core::system_error(hipErrorInvalidValue, "Invalid destination handle in hipMemCpyH2DAsync");
+
+    auto hip_stream = get_stream(stream);
+    if (!hip_stream)
+      throw xrt_core::system_error(hipErrorInvalidValue, "Invalid stream handle in hipMemCpyH2DAsync");
```
sonals, line 233: It would be best to have a helper inline function for these nullptr checks:
```
template <typename T>
inline void
throw_if_null(const T *ptr, const char *msg) {
    if (ptr != nullptr)
        return;
    throw xrt_core::system_error(hipErrorInvalidValue, msg);
} 

```
and call it like this for all the three cases here...

`throw_if_null(hip_mem_dst, "Invalid destination handle in hipMemCpyH2DAsync");`

`throw_if_null(hip_stream, "Invalid stream handle in hipMemCpyH2DAsync");`

...

sonals (changes requested):
Please make the suggested changes.

src/runtime_src/hip/api/hip_memory.cpp
chvamshi-xilinx: That is a good catch. @zhangchiming , I am already handling this in other memcpy functions, Please use same logic. 
ideal way is to handle this to change the signature of get_hip_mem_from_addr function. It should return pair of hip_memory and offset.

src/runtime_src/hip/api/hip_memory.cpp
chvamshi-xilinx: @zhangchiming , we already have throw_if* functions in hip/core/common.h . Please use those. I also didn't use them in my previous hip_memory PR :-( . Can you please replace them aswell.

src/runtime_src/hip/api/hip_memory.cpp
zhangchiming: fixed.

src/runtime_src/hip/api/hip_memory.cpp
zhangchiming: fixed

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -215,6 +205,27 @@ namespace xrt::core::hip
     free(host_src);
```
stsoe, line 205: This one caught my attention.  What happens if `write` throws?   Never use new/delete, alloc/free.
Consider using xrt_core::aligned_alloc (core/common/memalign.h) which manages the memory (RAII).

stsoe (changes requested):

src/runtime_src/hip/api/hip_memory.cpp
zhangchiming: fixed.

src/runtime_src/hip/core/event.cpp
```diff
@@ -164,15 +164,28 @@ bool kernel_start::wait()
...
   ctype = type::buffer_copy;
 }
 
 bool copy_buffer::submit()
 {
-  //handle = std::async(&xrt::bo::sync, &cbo, cdirection);
+  switch(cdirection)
+  {
+    case XCL_BO_SYNC_BO_TO_DEVICE:
+      handle = std::async(&memory::write, buffer, host_ptr, buffer->get_size(), 0, 0);
```
stsoe, line 178: By not specifying a policy, execution can be deferred and may not start the thread until `handle.get()` or `handle.wait()` is called.  Is that what you intend?

src/runtime_src/hip/core/event.cpp
```diff
@@ -164,15 +164,28 @@ bool kernel_start::wait()
   return false;
 }
 
-copy_buffer::copy_buffer(std::shared_ptr<stream> s)
-  : command(std::move(s))
+copy_buffer::copy_buffer(std::shared_ptr<stream> s, xclBOSyncDirection direction, std::shared_ptr<memory> buf, void* ptr)
+  : command(std::move(s)), cdirection(direction), buffer(buf), host_ptr(ptr)
```
stsoe, line 168:
```suggestion
  : command(std::move(s)), cdirection(direction), buffer(std::move(buf)), host_ptr(ptr)
```

zhangchiming: Fixed.

src/runtime_src/hip/core/event.cpp
zhangchiming: Fixed.

stsoe (approved):
LGTM.  Thanks

src/runtime_src/hip/api/hip_memory.cpp
chvamshi-xilinx: Looks like you are still not handling offset. Please check the hipMemCpy functions to understand the handling of offset,

zhangchiming: fixed.

chvamshi-xilinx: I still see the issue. I am essentially talking about handling of offset https://github.com/Xilinx/XRT/blob/master/src/runtime_src/hip/api/hip_memory.cpp#L142

zhangchiming: Fixed.

src/runtime_src/hip/core/memory.h
```diff
@@ -145,10 +145,10 @@ namespace xrt::core::hip
     remove(uint64_t addr);
   
     std::shared_ptr<xrt::core::hip::memory>
-    get_hip_mem_from_addr(void* addr);
+    get_hip_mem_from_addr(void* addr, size_t* offset=nullptr);
```
chvamshi-xilinx, line 148: Overall looks good, 
One last comment. Returning values in arguments is kind of a C style coding.
Please change the return value to std::pair<std::shared_ptr<memory, uint64_t> . send hip::memory and offset in that pair. Use it all the places where ever we are using

zhangchiming: fixed.

chvamshi-xilinx (approved):
Thanks for making changes. LGTM.

PR #8059, opened 2024-04-10
VITIS-11832 - Support transaction buffer patching in XRT

dezhiAmd:
Transaction buffer has new patching schemes. This PR aims to support them.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -109,9 +109,11 @@ struct patcher
     uc_dma_remote_ptr_symbol_kind = 1,
     shim_dma_base_addr_symbol_kind = 2, // patching scheme needed by AIE2PS firmware
     scalar_32bit_kind = 3,
-    control_packet_48 = 4,              // patching scheme needed by IPU firmware to patch control packet
-    shim_dma_48 = 5,                    // patching scheme needed by IPU firmware to patch instruction buffer
-    unknown_symbol_kind = 6
+    control_packet_48 = 4,              // patching scheme needed by firmware to patch dpu-sequence control packet
+    shim_dma_48 = 5,                    // patching scheme needed by firmware to patch dpu-seuqnece instruction buffer
+    tansaction_ctrlpkt_48 = 6,          // patching scheme needed by firmware to patch transaction buffer control packet
+    tansaction_48 = 7,                  // patching scheme needed by firmware to patch transaction buffer
+    unknown_symbol_kind = 8
   };
 
   symbol_type m_symbol_type;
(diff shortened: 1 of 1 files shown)


stsoe (changes requested):
No c-style casts please. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es49-if-you-must-use-a-cast-use-a-named-cast
Please add `// NOLINT` as comment to each line with magic constant.

stsoe (approved):

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -176,6 +178,19 @@ struct patcher
...
+    uint64_t val = *reinterpret_cast<uint64_t*>(bd_data_ptr);
+    val &= 0x0000FFFFFFFFFFFF;                                                        // NOLINT
+    val += patch;
+    auto lower32BitVal = static_cast<uint32_t>(val);
+    auto higher16BitVal = static_cast<uint16_t>((val & 0x0000FFFF00000000) >> 32);    // NOLINT
+    *bd_data_ptr = lower32BitVal;
+
+    auto pHigher16Bit = reinterpret_cast<uint16_t*>(bd_data_ptr + 1);
+    *pHigher16Bit = higher16BitVal;
+  }
```
larry9523, line 192: I think the key here is by doing this patching, we are actually define the bd_data_ptr is pointing to a BD instead of pointing to the beginning of a control packet. My suggestion is the patch schema is based on the hardware architecture but not dpu sequence and transaction buffer. 

My thought on this is we should be able to reuse patch_ctrl48 and patch_shim48 for both DPU sequence and transaction buffer

larry9523 (commented):
I think for aie2 and aie2p, we should use same patching scheme regardless DPU sequence or transaction buffer.

src/runtime_src/core/common/api/xrt_module.cpp
dezhiAmd: "Reuse patch_ctrl48 and patch_shim48 for both DPU sequence and transaction buffer" is an optimization.
In order to achieve that, we need to modify assembler.
Refer to this PR: https://gitenterprise.xilinx.com/XRT/aiebu/pull/16
It would be nice if "Reusing patch_ctrl48 and patch_shim48 for both DPU sequence and transaction buffer" was brought up there. And we do not need this new enum in this [line](https://gitenterprise.xilinx.com/XRT/aiebu/blob/ad0d57f9b6b66bdd83ed9be0835e2e2207ee7c70/src/python/aiebu/scripts/symbol.py#L23).

With that all being said, can we use what we have, optimize in another PR?

larry9523: I would not say it is an optimization because it is an design option. With this PR the enum is not clean. We should have 
dpu_patch_ctrl48
transaction_patch_ctrl48
if we want to two set of patching schema.

But if we add some comment explaining this is a temporary solution before we fix the assembler, and the code extra enum will be removed, I am fine to merge it for now.

dezhiAmd: Comment added

larry9523 (approved):
Nit: higher16BitVal and bd_data_ptr are two different style to define a variable. It would be nice if we can make it consistent in a file. But since this code will be removed, I approved it.

PR #8063, opened 2024-04-10
VITIS-11806 Command-chaining: XRT C++ Command List

stsoe:
Add first class experimental API for xrt::runlist.

class runlist - A class to manage a list of xrt::run objects This class is used to manage a list of xrt::run objects such that they can be executed atomically in the order they are added to the list.

Run objects are added to the list using the add() method. A runlist is submitted for execution using the execute() method. Once the list is executing, no more run objects can be added until execution of the current last run object has completed. The list can be reset using the reset() method, which will clear all run objects, or the list can be reused by calling execute() again maybe with additional run objects.

There is no support for removing individual run objects from the list.

Check experimental/xrt_kernel.h for more documentation.

Some minor changes to existing classes for code sharing, but runlist should not affect existing flows.

OpenCL regression suite.
Clang-tidy.

New exposed first class API should be documented.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -8,6 +8,7 @@
 #define XCL_DRIVER_DLL_EXPORT  // exporting xrt_kernel.h
 #define XRT_CORE_COMMON_SOURCE // in same dll as core_common
 #include "core/include/xrt/xrt_kernel.h"
+#include "core/include/experimental/xrt_kernel.h"
 
 #include "core/common/shim/buffer_handle.h"
 #include "core/common/shim/hwctx_handle.h"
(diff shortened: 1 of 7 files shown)


maxzhen (approved):

larry9523 (approved):

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -2879,6 +2910,168 @@ class run::command_error_impl
...
+  }
+};
+
+
+class runlist::command_error_impl
+{
+public:
+  xrt::run m_run;
+  ert_cmd_state m_state;
+  std::string m_message;
```
sonals, line 3066: Should these be `const` members?

stsoe: Why?  This are data members are part of a pimpl, opaque to exposed API.  Access is limited to API implementation within this source file, which does return by-value copies.

PR #8076, opened 2024-04-16
 VITIS-11623 - use same patching scheme for same hardware regardless of firmware code path

dezhiAmd:
Currently transaction-buffer and dpu-sequence uses different patching scheme. But same hardware should use same patching scheme.
Merge different patching scheme by adjusting offset in assembler. Refer to assembler PR (https://gitenterprise.xilinx.com/XRT/aiebu/pull/22/files)
1. Dump file ctrlpktBoPatched.bin before adjusting offset and after adjusting offset, compare the two files and they are identical
2. With this change, now all 4 test cases from xcompiler-team are working.
Without this change, case-conv1 does not work, This is documented in https://jira.xilinx.com/browse/VITIS-11623

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -178,20 +178,6 @@ struct patcher
     bd_data_ptr[2] = (bd_data_ptr[2] & 0xFFFF0000) | (base_address >> 32);            // NOLINT
   }
 
-  // TODO reuse patch_ctrl48 for both DPU sequence and transaction buffer
-  // Need change assembler first
-  void patch_tansaction_ctrlpkt_48(uint32_t* bd_data_ptr, uint64_t patch)
-  {
-    uint64_t val = *reinterpret_cast<uint64_t*>(bd_data_ptr);
-    val &= 0x0000FFFFFFFFFFFF;                                                        // NOLINT
-    val += patch;
-    *bd_data_ptr = static_cast<uint32_t>(val);
-
-    auto higher16BitVal = static_cast<uint16_t>((val & 0x0000FFFF00000000) >> 32);    // NOLINT
-    auto pHigher16Bit = reinterpret_cast<uint16_t*>(bd_data_ptr + 1);
-    *pHigher16Bit = higher16BitVal;
-  }
-
   void
   patch(uint8_t* base, uint64_t patch)
   {
@@ -211,7 +197,8 @@ struct patcher
         patch_shim48(bd_data_ptr, patch);
         break;
       case symbol_type::tansaction_ctrlpkt_48:
-        patch_tansaction_ctrlpkt_48(bd_data_ptr, patch);
+          // Will change "bd_data_ptr - 8" to "bd_data_ptr" once assembler is fixed
+          patch_ctrl48(bd_data_ptr - 8, patch);
         break;
       case symbol_type::tansaction_48:
         // No patching since transaction buffer firmware does not support
@@ -858,12 +845,8 @@ class module_sram : public module_impl
     // patch_instr is a virtual method, what is actually called here ???
     if (m_ctrlpkt_buf) {
       patch_instr("control-packet", m_ctrlpkt_buf);
-
-#ifdef _DEBUG
-      dump_bo(m_instr_buf, "instrBoPatchedByCtrlPacket.bin");
-#endif
-      XRT_PRINTF("<- module_sram::create_instr_buf()\n");
     }
+    XRT_PRINTF("<- module_sram::create_instr_buf()\n");
   }
 
   void
@@ -990,8 +973,14 @@ class module_sram : public module_impl
     }
     else if (os_abi == Elf_Amd_Aie2p) {
       m_instr_buf.sync(XCL_BO_SYNC_BO_TO_DEVICE);
+#ifdef _DEBUG
+      dump_bo(m_instr_buf, "instrBoPatched.bin");
+#endif
       if (m_ctrlpkt_buf)
         m_ctrlpkt_buf.sync(XCL_BO_SYNC_BO_TO_DEVICE);
+#ifdef _DEBUG
+        dump_bo(m_ctrlpkt_buf, "ctrlpktBoPatched.bin");
+#endif
     }
 
     m_dirty = false;
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -314,6 +314,7 @@ public:
   //
   // @param symbol - symbol name
   // @param bo - global argument to patch into ctrlcode
+  // @param buf_type - whether it is control-code, control-packet, preempt-save or preempt-restore
   virtual void
   patch_instr(const std::string&, const xrt::bo&, patcher::buf_type)
   {
@@ -346,6 +347,7 @@ public:
   // @param base - base address of control code buffer object
   // @param symbol - symbol name
   // @param patch - patch value
+  // @param buf_type - whether it is control-code, control-packet, preempt-save or preempt-restore
   // @Return true if symbol was patched, false otherwise  //
   virtual bool
   patch(uint8_t*, const std::string&, uint64_t, patcher::buf_type)
@@ -984,7 +986,7 @@ class module_sram : public module_impl
     auto os_abi = m_parent.get()->get_os_abi();
     if (os_abi == Elf_Amd_Aie2ps) {
       if (m_patched_args.size() != m_parent->number_of_arg_patchers()) {
-          boost::format fmt = boost::format("ctrlcode requires %d patched arguments, but only %d are patched")
+        auto fmt = boost::format("ctrlcode requires %d patched arguments, but only %d are patched")
             % m_parent->number_of_arg_patchers() % m_patched_args.size();
         throw std::runtime_error{ fmt.str() };
       }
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -211,7 +197,8 @@ struct patcher
         patch_shim48(bd_data_ptr, patch);
         break;
       case symbol_type::tansaction_ctrlpkt_48:
-        patch_tansaction_ctrlpkt_48(bd_data_ptr, patch);
+          // Will change "bd_data_ptr - 8" to "bd_data_ptr" once assembler is fixed
```
dezhiAmd, line 200: We need make change in assembler accordingly for the offset value

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -178,23 +187,12 @@ struct patcher
...
-    auto higher16BitVal = static_cast<uint16_t>((val & 0x0000FFFF00000000) >> 32);    // NOLINT
-    auto pHigher16Bit = reinterpret_cast<uint16_t*>(bd_data_ptr + 1);
-    *pHigher16Bit = higher16BitVal;
-  }
-
   void
-  patch(uint8_t* base, uint64_t patch)
+  patch(uint8_t* base, uint64_t patch, buf_type type)
   {
+    if (type != m_buf_type)
```
dezhiAmd, line 193: With this line of change, now all 4 test cases from xcompiler-team are working.
Without this change, case-conv1 does not work, This is documented in https://jira.xilinx.com/browse/VITIS-11623

larry9523 (approved):

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -81,7 +81,7 @@ struct buf
     if (!column_page_size)
       return;
 
-    auto pad = (page + 1) * column_page_size;
+    auto pad = static_cast<size_t>(page + 1) * column_page_size;
```
stsoe, line 84: Is this cast really necessary?  Was it added because of compiler complaining about possible roll over?
If so the cast does nothing to prevent that because it is applied after the roll over.

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -116,15 +116,24 @@ struct patcher
...
-  symbol_type m_symbol_type;
+   enum class buf_type {
+       ctrltext = 0,   // control code
+       ctrldata,       // control packet
+       preempt_save,   // preempt_save
+       preempt_restore // preempt_restore
+  };
+
+  buf_type m_buf_type = buf_type::ctrltext;
+  symbol_type m_symbol_type = symbol_type::shim_dma_48;
```
stsoe, line 127:
```suggestion
  auto m_buf_type = buf_type::ctrltext;
  auto m_symbol_type = symbol_type::shim_dma_48;
```

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -312,10 +310,14 @@ class module_impl
     throw std::runtime_error("Not supported");
   }
 
+  // Patch ctrlcode buffer object for global argument
+  //
+  // @param symbol - symbol name
+  // @param bo - global argument to patch into ctrlcode
   virtual void
-  patch_instr(const std::string&, const xrt::bo&)
+  patch_instr(const std::string&, const xrt::bo&, const patcher::buf_type)
```
stsoe, line 318: No point adding const on integral parameter by value.

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -346,7 +348,7 @@ class module_impl
   // @param patch - patch value
   // @Return true if symbol was patched, false otherwise  //
   virtual bool
-  patch(uint8_t*, const std::string&, uint64_t)
+  patch(uint8_t*, const std::string&, uint64_t, patcher::buf_type)
```
stsoe, line 351: Yes, this more common, compared to earlier const.

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -570,7 +577,7 @@ class module_elf : public module_impl
           auto symbol_type = static_cast<patcher::symbol_type>(rela->r_addend);
           std::vector<uint64_t> offsets;
           offsets.push_back(offset);
-          arg2patchers.emplace(std::move(argnm), patcher{ symbol_type, std::move(offsets) });
+          arg2patchers.emplace(std::move(argnm), patcher{ symbol_type, std::move(offsets), buf_type });
```
stsoe, line 580: Maybe this will work without the local offsets variable and push back above?
```suggestion
          arg2patchers.emplace(std::move(argnm), patcher{ symbol_type, {offset}, buf_type });
```

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -634,23 +641,24 @@ class module_elf : public module_impl
         std::string argnm{ symname, symname + std::min(strlen(symname), dynstr->get_size()) };
         auto symbol_type = static_cast<patcher::symbol_type>(rela->r_addend);
 
+        patcher::buf_type buf_type = patcher::buf_type::ctrltext;
         std::vector<uint64_t> offsets;
         offsets.push_back(ctrlcode_offset);
-        arg2patcher.emplace(std::move(argnm), patcher{ symbol_type, offsets });
+        arg2patcher.emplace(std::move(argnm), patcher{ symbol_type, offsets, buf_type});
```
stsoe, line 647: Ditto ??
```suggestion
        arg2patcher.emplace(std::move(argnm), patcher{ symbol_type, {ctrlcode_offset}, buf_type});
```

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -836,7 +844,7 @@ class module_sram : public module_impl
   create_instr_buf(const module_impl* parent)
   {
     XRT_PRINTF("-> module_sram::create_instr_buf()\n");
-    auto data = parent->get_instr();
+    const instr_buf& data = parent->get_instr();
```
stsoe, line 847: Plus and minus here.  Is it clearer with the type in your face, maybe?
```suggestion
    const auto& data = parent->get_instr();
```

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -854,23 +862,17 @@ class module_sram : public module_impl
...
     }
+    XRT_PRINTF("<- module_sram::create_instr_buf()\n");
   }
 
   void
   create_ctrlpkt_buf(const module_impl* parent)
   {
     XRT_PRINTF("-> module_sram::create_ctrlpkt_buf()\n");
-    auto data = parent->get_ctrlpkt();
+    const control_packet& data = parent->get_ctrlpkt();
```
stsoe, line 875: Same comment.
```suggestion
    const auto& data = parent->get_ctrlpkt();
```

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -898,7 +900,7 @@ class module_sram : public module_impl
   create_instruction_buffer(const module_impl* parent)
   {
     XRT_PRINTF("-> module_sram::create_instruction_buffer()\n");
-    auto data = parent->get_data();
+    const std::vector<ctrlcode>& data = parent->get_data();
```
stsoe, line 903: And here:
```suggestion
    const auto& data = parent->get_data();
```

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -917,10 +919,10 @@ class module_sram : public module_impl
     XRT_PRINTF("<- module_sram::create_instruction_buffer()\n");
   }
 
-  void
-  patch_instr(const std::string& argnm, const xrt::bo& bo) override
+  virtual void
+  patch_instr(const std::string& argnm, const xrt::bo& bo, const patcher::buf_type type) override
```
stsoe, line 923: No typical adding const on by value paramters.  
```suggestion
  patch_instr(const std::string& argnm, const xrt::bo& bo, patcher::buf_type type) override
```

stsoe (changes requested):
Nothing major.

src/runtime_src/core/common/api/xrt_module.cpp
dezhiAmd: These two lines declare two member variables in struct patcher and initialize them.
Keyword "auto " cannot be used there.

src/runtime_src/core/common/api/xrt_module.cpp
dezhiAmd: Maybe this will work. But it has been this way for a while, And I do not see significant improvement using the suggested change either.

src/runtime_src/core/common/api/xrt_module.cpp
dezhiAmd: Maybe this will work. But it has been this way for a while, And I do not see significant improvement using the suggested change either.

src/runtime_src/core/common/api/xrt_module.cpp
stsoe: Ah, of course, I missed the context.

stsoe (approved):

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -982,16 +984,22 @@ class module_sram : public module_impl
     auto os_abi = m_parent.get()->get_os_abi();
     if (os_abi == Elf_Amd_Aie2ps) {
       if (m_patched_args.size() != m_parent->number_of_arg_patchers()) {
-        auto fmt = boost::format("ctrlcode requires %d patched arguments, but only %d are patched")
+          boost::format fmt = boost::format("ctrlcode requires %d patched arguments, but only %d are patched")
```
HimanshuChoudhary-Xilinx, line 987: "auto" was good

src/runtime_src/core/common/api/xrt_module.cpp
HimanshuChoudhary-Xilinx, line 351: document above not updated, same goes multiple places

HimanshuChoudhary-Xilinx (approved):
minor comments.

src/runtime_src/core/common/api/xrt_module.cpp
dezhiAmd: Thanks for catch that. Fixed.

PR #8080, opened 2024-04-17
CR-1197359: Modifying XRT codebase to send absolute tile column value to AIE driver

saumyag-xilinx:
The tile gating error of CR-1197359 will be fixed with this change
The information in the xclbin denotes relative values of the tile column, so the change is made to send absolute values to aie driver.
Low
Test case provided in the CR has been verified for this portion of flow.
NA


chvamshi-xilinx (approved):

chvamshi-xilinx:
@saumyag-xilinx , Just now observed that, there are some more places where we  need to add  start column. Please make those changes aswell.

chvamshi-xilinx (approved):

PR #8011, opened 2024-03-13
integrate aie-pdi-transform into xclbinutil

xfreid:
#### Problem solved by the commit
Integrate aie-pdi-transform to xclbinutil per Larry's request

The request was from Larry Liu and Sonal Santan
"All current shipping xclbins need to be updated to use the new format.
We need to integrate [aie-pdi-transform] (https://gitenterprise.xilinx.com/ai-engine/aie-pdi-transform) into xclbinutil so we can upgrade an existing xclbin into the new format using xclbinutil. The details of the tools are here—
https://confluence.xilinx.com/pages/viewpage.action?spaceKey=~huaj&title=Transform+PDI+In+IPU-PHX"

The code is from aie-pdi-transform git repo - https://gitenterprise.xilinx.com/ai-engine/aie-pdi-transform
This code was initially developed and maintained by Hua, but I think he is no longer responsible for maintaining this repo. 

Per Sonal's guidance, we copied the source code from the repo above to XRT so that transform_static can be built and shipped along with xclbinutil

This repo is only used for IPU, once we copy the code to XRT repo under xclbinutil, the source code in XRT becomes the single source of truth, so there won’t be an out-of-sync concern.

The source code in aie-pdi-transform is going through legal sweep right now. [@Varughese, Irene](mailto:Irene.Varughese@amd.com) will help us to get legal approval once all the flagged items are addressed.

The details about “what does it do, why, and how” can be found in https://confluence.xilinx.com/pages/viewpage.action?spaceKey=~huaj&title=Transform+PDI+In+IPU-PHX

#### Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered
N/A

#### How problem was solved, alternative solutions (if any) and why they were rejected
N/A

#### Risks (if any) associated the changes in the commit
N/A

#### What has been tested and how, request additional testing if necessary
Manual test
Ran clang-tidy to fix the reported warning

#### Documentation impact (if any)
None

PR #8086, opened 2024-04-19
More clang-tidy cleanups of HIP code and fix a major bug in new PDI trans…

sonals:
…form error handling path

1. Compilation failure on Ubuntu 23.10 due to a bug in the new pdi-transform code
2. More clang-tidy cleanups for HIP code

https://github.com/Xilinx/XRT/pull/8011 introduced the compilation failure bug. clang-tidy changes are ongoing.

Fix the code

None

Compile tested

NA

### DIFF
--- a/src/.clang-tidy
+++ b/src/.clang-tidy
@@ -97,6 +97,8 @@ CheckOptions:
     value:           '16'
   - key:             bugprone-easily-swappable-parameters.NamePrefixSuffixSilenceDissimilarityTreshold
     value:           '1'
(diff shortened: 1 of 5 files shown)


stsoe (approved):
I too cluttered my code with [[nodiscard]], I personally think we should remove this warning.   There is only ever one case where this is a good directive and that is when a function returns a variable for RAII that you must hold on to in current scope;  this is what [[nodiscard]] is meant for.   Could you disable [[nodiscard]]  warning?

chvamshi-xilinx (approved):

PR #8087, opened 2024-04-19
Enabling pipeline for 2024.1

manikandan-xilinx:
Enabling pipeline for 2024.1 as per request @chvamshi-xilinx

### DIFF
--- a/.github/workflows/xrt_master.yml
+++ b/.github/workflows/xrt_master.yml
@@ -1,4 +1,4 @@
-name: XRT Master
+name: XRT 2024.1
 
 env:  
   RELEASE: '2024.1'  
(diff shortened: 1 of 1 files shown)


.github/workflows/xrt_master.yml
```diff
@@ -1,14 +1,15 @@
-name: XRT Master
+name: XRT 2024.1
```
keryell, line 1: Do you mean that `master` is no longer tested?
Should the file be renamed?

manikandan-xilinx: Hi @keryell This would be similar to master for 2024.1 branch and current master is 2024.2 so there will be another file for that

manikandan-xilinx:
Build pipeline check is coming from jenkins which we are not using now and all the github actions checks have passed.

PR #8089, opened 2024-04-19
Add 1ms delay in ML Timeline, AIE Profile/Debug Plugin to avoid BO sync issue

IshitaGhosh:
Issue with BO sync (CR-1194803) is still under investigation. This commit adds 1 ms delay between bo sync and use of its contents as a workaround.

This is a workaround. Correct fix should be in sync_bo implementation.

The XRT run time for any application will increase by 1ms with each of ML Timeline, AIE Profile and AIE debug plugin enabled.

### DIFF
--- a/src/runtime_src/xdp/profile/device/common/client_transaction.cpp
+++ b/src/runtime_src/xdp/profile/device/common/client_transaction.cpp
@@ -16,7 +16,9 @@
 
 #define XDP_PLUGIN_SOURCE
 
+#include <chrono>
 #include <sstream>
+#include <thread>
 
 #include "client_transaction.h"
 #include "core/common/message.h"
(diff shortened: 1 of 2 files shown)


dezhiAmd:
It is not clear how to reproduce the "Issue with BO sync " from the description.
For instance, what test/application did we use? What the steps should we use?

Down the road, say another developer want the remove the workaround, it would be difficult to do so without the information above.

With that being said, can we associate this PR with a CR report?

IshitaGhosh:
@dezhiAmd I just added the CR number in the description.

IshitaGhosh:
This PR is a workaround and the actual fix is still being investigated. So, closing this PR for now.

PR #8090, opened 2024-04-19
Silence C++ [[nodiscard]] warning

sonals:
Add a note about new aie-pdi-transform component

Tune clang-tidy behavior to not generate warnings about [[nodiscard]]

1. Silence the [[nodiscard]] clang-tidy messages
2. Add a note about aie-pdi-transform

Updated .clang-tidy rules

None

Compile tested

None


src/runtime_src/hip/core/memory.cpp
```diff
@@ -103,7 +103,7 @@ namespace xrt::core::hip
         assert(src_hip_mem->get_flags() == hipHostMallocDefault || src_hip_mem->get_flags() == hipHostMallocPortable);
     }
 
-    const unsigned char *src_ptr = reinterpret_cast<const unsigned char *>(src);
+    const auto *src_ptr = reinterpret_cast<const unsigned char *>(src);
```
stsoe, line 106:
```suggestion
    auto src_ptr = reinterpret_cast<const unsigned char *>(src);
```
Already const and ptr per cast.

src/runtime_src/hip/core/memory.cpp
```diff
@@ -119,21 +119,21 @@ namespace xrt::core::hip
         // pinned hip mem
         assert(dst_hip_mem->get_flags() == hipHostMallocDefault || dst_hip_mem->get_flags() == hipHostMallocPortable);
     }
-    unsigned char *dst_ptr = reinterpret_cast<unsigned char *>(dst);
+    auto *dst_ptr = reinterpret_cast<unsigned char *>(dst);
```
stsoe, line 122:
```suggestion
    auto dst_ptr = reinterpret_cast<unsigned char *>(dst);
```

stsoe (changes requested):

src/runtime_src/hip/core/memory.cpp
sonals: Updated

src/runtime_src/hip/core/memory.cpp
sonals: Updated

PR #8099, opened 2024-04-22
fix for centos78 issue

manikandan-xilinx:
Using checkout version v4 resulting into "fatal: Expected git repo version <= 0, found 1" for centos builds so going back to v3


SubhedarV (approved):

manikandan-xilinx:
This is a pipeline change so CI check failure is fine.

PR #8100, opened 2024-04-23
Fixing extraction logic

manikandan-xilinx:
Extraction logic was using feeder_latest but should use qualification instead. v0.0.2 has qualification


manikandan-xilinx:
This has no dependency on build or pcie_hw tests so merging. In future will remove xrt_ci check for workflow changes as it's unnecessary and add different checks

PR #8104, opened 2024-04-24
Fix GeMM coverity

AShivangi:
Resolve coverity's divide-by-zero 

N/A

If the clock freq is 0 (which will never be the case), divide by 1

N/A

Compiled locally

N/A

### DIFF
--- a/src/runtime_src/core/tools/common/tests/TestGemm.cpp
+++ b/src/runtime_src/core/tools/common/tests/TestGemm.cpp
@@ -202,7 +202,7 @@ TestGemm::run(std::shared_ptr<xrt_core::device> dev)
     ptree.put("status", test_token_failed);
     return ptree;
   }
-  double ipu_hclck_period= 1000000000.0/((ipu_hclock)*1000000); //MHz to ns
+  double ipu_hclck_period = 1000000000.0 / (ipu_hclock * 1000000); // MHz to ns
   uint32_t* core_ptr = reinterpret_cast<uint32_t*>(bo_result_map+offset_3K);
   double TOPS = 0.0;
   double total_cycle_count = 0.0;
--- a/src/runtime_src/core/tools/common/tests/TestGemm.cpp
+++ b/src/runtime_src/core/tools/common/tests/TestGemm.cpp
@@ -163,7 +163,8 @@ TestGemm::run(std::shared_ptr<xrt_core::device> dev)
 
   //set to performance mode
   xrt_core::device_update<xrt_core::query::performance_mode>(dev.get(), xrt_core::query::performance_mode::power_type::high);
-  //5 second delay gives the clocks time to reach the targeted frequency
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/tools/common/tests/TestGemm.cpp
```diff
@@ -196,7 +197,7 @@ TestGemm::run(std::shared_ptr<xrt_core::device> dev)
   std::this_thread::sleep_for(std::chrono::milliseconds(30));
 
   //Calculate TOPS
-  double ipu_hclck_period= 1000000000.0/(ipu_hclock*1000000); //MHz to ns
+  double ipu_hclck_period= 1000000000.0/((ipu_hclock != 0 ? ipu_hclock : 1)*1000000); //MHz to ns
```
sonals, line 200: If `(ipu_hclock == 0) `should we fail the test?

AShivangi: Yes, I've made a change to do so

src/runtime_src/core/tools/common/tests/TestGemm.cpp
```diff
@@ -196,7 +197,12 @@ TestGemm::run(std::shared_ptr<xrt_core::device> dev)
   std::this_thread::sleep_for(std::chrono::milliseconds(30));
 
   //Calculate TOPS
-  double ipu_hclck_period= 1000000000.0/(ipu_hclock*1000000); //MHz to ns
+  if(ipu_hclock == 0) {
```
stsoe, line 200:
```suggestion
  if (ipu_hclock == 0) {
```

src/runtime_src/core/tools/common/tests/TestGemm.cpp
```diff
@@ -196,7 +197,12 @@ TestGemm::run(std::shared_ptr<xrt_core::device> dev)
...
+    logger(ptree, "Error", "IPU H-clock is 0");
+    ptree.put("status", test_token_failed);
+    return ptree;
+  }
+  double ipu_hclck_period= 1000000000.0/((ipu_hclock)*1000000); //MHz to ns
```
stsoe, line 205: These are magic constants, are you sure you have the right number of zeros in the various places? :-)
Not overly important, just fix the formatting.

```suggestion
  double ipu_hclck_period = 1000000000.0 / (ipu_hclock * 1000000); // MHz to ns
```

stsoe (changes requested):

src/runtime_src/core/tools/common/tests/TestGemm.cpp
AShivangi: Yes! I made sure that the numbers of zeroes is right. I am not sure how else to specify them. Hence I added the comment of what this line is doing

PR #8105, opened 2024-04-24
Passing as a separte module of mgmtpf & XOCL to avoid execution confl…

raghkris:
…ict.

Used separate module string name  for both MGMTPF & XOCL  to avoid macro and  execution at runtime.


houlz0507:
Can you explain more on why this change is needed?

raghkris:
> 
As we were working with customer kernel & as part of custom kernel changes, **xocl_debug_init()** function is being called by both xclmgmt_init() and xocl_init()  fucntions, so we differentiated with by passing parameters as module like "mgmtpf" & "xocl", to avoid execution conflict of  xocl_debug_init() fucntions.

PR #8110, opened 2024-04-26
Enable HIP build in windows

rbramand-xilinx:
make changes to cmake files to enable HIP build on windows
Fix compilation errors

TODO: send submodule update with this changes to XRT-MCDM repo and enable hip build option in build-win.bat of MCDM repo

In windows HIP SDK can be installed to get HIP include files and it install cmake config files of HIP to fixed path
For example latest SDK available on windows is of 5.7 version and it installs it to - C:/Program Files/AMD/ROCm/5.7/lib/cmake/hip
So when searching for HIP config files I have added this hardcode path but also gave an option for user to set environment variable HIP_DIR to pick config files from that location.

Low

Tested xrt_flow test using HIP apis on phoenix using MCDM flow

NA


src/CMake/nativeWin.cmake
```diff
@@ -69,6 +69,18 @@ endif()
...
+if (XRT_ENABLE_HIP)
+  message("-- Looking for HIP include files...")
+  set(HIP_PLATFORM "amd")
+  # HIP SDK installs hip files to C:/Program Files/AMD/ROCm
+  # Latest version available on windows is 5.7
+  # Users can set HIP_DIR to location of HIP installation or default path is used
+  set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} $ENV{HIP_DIR} "C:/Program Files/AMD/ROCm/5.7/lib/cmake/hip")
+  include(hip-config)
+  message("-- Found at ${HIP_INCLUDE_DIR}")
+endif()
```
stsoe, line 82: I am no longer keen on distributed make snippets.  Is it possible to have this piece inside src/runtime_src/hip CMakeLists.txt or config.cmake instead?

src/runtime_src/hip/core/device.h
```diff
@@ -3,6 +3,10 @@
 #ifndef xrthip_device_h
 #define xrthip_device_h
 
+#ifdef _WIN32
+#undef max
+#endif
```
stsoe, line 8: This is annoying yes.   But I think the better fix is to #define NOMINMAX before inclusion of windows.h in all units that include this header.  I have seen this in Microsoft's own code:
```
#define NOMINMAX
#include <windows.h>
```

stsoe (changes requested):
Few comments, take it or leave it.

src/runtime_src/hip/core/device.h
rbramand-xilinx: Hi @stsoe, when I remove this I get some different errors from MSVC internal code. So I haven't removed this. Hope that is okay

src/runtime_src/CMakeLists.txt
```diff
@@ -40,7 +40,9 @@ if (XRT_DKMS_ALVEO STREQUAL "ON")
   xrt_add_subdirectory(ert)
 endif()
 
+# --- Optional HIP bindings ---
 if (XRT_ENABLE_HIP)
+  include(${CMAKE_CURRENT_SOURCE_DIR}/../CMake/hip.cmake)
```
stsoe, line 45: Can you move hip.cmake to src/runtime_src/hip/ some place and include it in src/runtime_src/hip/CMakeLists.txt?  We need to move away from global configuration.

src/runtime_src/CMakeLists.txt
```diff
@@ -81,4 +83,10 @@ else()
   xrt_add_subdirectory(xdp)
   xrt_add_subdirectory(tools/scripts)
   xrt_add_subdirectory(core)
+
+  # --- Optional HIP bindings ---
+  if (XRT_ENABLE_HIP)
+    include(${CMAKE_CURRENT_SOURCE_DIR}/../CMake/hip.cmake)
```
stsoe, line 89: Ditto

stsoe (changes requested):

src/runtime_src/CMakeLists.txt
rbramand-xilinx: Done

stsoe (approved):

chvamshi-xilinx (approved):

PR #8111, opened 2024-04-26
CR-1186952 xbutil examine -r all -d BDF failed while another terminal is running program at the same time

AShivangi:
When a context is destroyed and xbutil is trying to get information about it, we get an escape call exception. 

DSV testing

wrap the call in a try/catch block

N/A

Compiled locally


chvamshi-xilinx:
Isnt this to be taken care by the driver when two processes are trying to access the information? I am not quite sure whether this fix is correct or not. Can you please elaboarate more.

chvamshi-xilinx:
Had an offline discussion with @rchane , This is not a complete fix. This is work-around which needs for release.
Ideal fix should go here https://gitenterprise.xilinx.com/XRT/XRT-MCDM/blob/02250b9d71c40eaff169799a5024b86e349aff17/src/umd/mcdm/query.cpp#L376

PR #8112, opened 2024-04-26
fix clang-tidy warnings for hip

haghilin:
Fixed some clang-tidy warnings for HIP 

Problem solved by the commit
Fixed some clang-tidy warnings for HIP

NA

NA

Documentation impact (if any)
None.


chvamshi-xilinx (approved):

sonals (commented):
@haghilin , can you add comments on why we should override clang-tidy in all places where we have added // NO LINT

haghilin:
> @haghilin , can you add comments on why we should override clang-tidy in all places where we have added // NO LINT

Hi @sonals Thanks for the comment, it has been applied.

chvamshi-xilinx (approved):

PR #8118, opened 2024-04-29
[2024.1] Cover all stream switch events and fix the logic to check if event is stream switch port

vipangul:
**This is double commit of issue mentioned in https://github.com/Xilinx/XRT/pull/8117** 

For one of design, when both aie_trace & aie_profile plugins were enabled, there is discrepency observed for aie_profile stream switch port configuration vs what is reported in generated report.

it is observed that Running and Stalled events are coming from ports , PORT0 & PORT1 respectively. These ports are incorrectly reported as it configures stream port 4 for profile as trace takes up first few ports.
This is not expected behavior, Ideally these events should come from the same ports.

We have now added all 8 ports with different states supported for each port ( Running, Stalled, Idle )
Also Fix the logic to check if event is stream switch port type.

We have fix the logic to correctly mark the event is stream switch port which is a bigger set now. This could generate more report data.

Verified on reported design for AIE1.

Before fix:

[XRT] DEBUG: Configuring interface tile stream switch to monitor slave stream port 4
1933.09,26,0,332,332,0,0,342998101906,4,
1933.11,26,0,339,339,0,0,342998101906,4,
After Fix:

2195.09,26,0,340,340,0,0,243977582107,4,
2195.1,26,0,339,339,0,0,243977582107,4,


manikandan-xilinx:
please ignore (Build pipeline!) failure that's coming from Jenkins which we are not using and all the GitHub Actions checks have passed. That check will only be visible for branches other than master as per branch protection rules

jvillarre (approved):

PR #8122, opened 2024-04-30
Passing as a separate module of mgmtpf & xocl to avoid execution confl…

raghkris:
…ict in a custom kernel

As the 2 drivers xocl/ & xclmgmt/ are being compiled as part of kernel image with static compilation, there was an issue of conflict with two driver registrations.
Resolution is to add the string as new parameter in mgmt driver initialization.
As part of custom kernel changes, xocl_debug_init() function is being called by both xclmgmt_init() and xocl_init() fucntions, so we differentiated with by passing parameters as module like "mgmtpf" & "xocl", to avoid execution conflict of xocl_debug_init() fucntions.
Used separate module string name for both MGMTPF & XOCL to avoid macro and execution at runtime.

### DIFF
--- a/.github/workflows/xrt_master.yml
+++ b/.github/workflows/xrt_master.yml
@@ -1,4 +1,4 @@
-name: XRT Master
+name: XRT 2024.1
 
 env:  
   RELEASE: '2024.1'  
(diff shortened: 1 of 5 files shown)


chvamshi-xilinx:
Closing as these are available in 2022.2

PR #8127, opened 2024-05-01
fix hipMalloc() issue when running in multithreads.

zhangchiming:
fix hipMalloc() issue when running in multithreads.

Problem solved by the commit
crash on calling hipMalloc() from multiple threads.

Issue was found during developing unit testing case.

Add lock for memory_database::remove() and  memory_database::insert(). Allow device_init() be called once per thread.

Compiled and tested on Ubuntu 22.04 running on Ryzen 7840.

Documentation impact (if any)
None.

### DIFF
--- a/src/runtime_src/hip/core/memory.cpp
+++ b/src/runtime_src/hip/core/memory.cpp
@@ -171,12 +171,14 @@ namespace xrt::core::hip
   void
   memory_database::insert(uint64_t addr, size_t size, std::shared_ptr<xrt::core::hip::memory> hip_mem)
   {
+    std::lock_guard<std::mutex> lock(m_mutex);
     m_addr_map.insert({address_range_key(addr, size), hip_mem});
   }
 
   void
   memory_database::remove(uint64_t addr)
   {
+    std::lock_guard<std::mutex> lock(m_mutex);
     m_addr_map.erase(address_range_key(addr, 0));
   }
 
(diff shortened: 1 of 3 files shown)


src/runtime_src/hip/core/memory.cpp
```diff
@@ -171,12 +171,14 @@ namespace xrt::core::hip
   void
   memory_database::insert(uint64_t addr, size_t size, std::shared_ptr<xrt::core::hip::memory> hip_mem)
   {
+    std::lock_guard<std::mutex> lock(m_mutex);
```
stsoe, line 174:
```suggestion
    std::lock_guard lock(m_mutex);
```

src/runtime_src/hip/core/memory.cpp
```diff
@@ -171,12 +171,14 @@ namespace xrt::core::hip
...
     m_addr_map.insert({address_range_key(addr, size), hip_mem});
   }
 
   void
   memory_database::remove(uint64_t addr)
   {
+    std::lock_guard<std::mutex> lock(m_mutex);
```
stsoe, line 181:
```suggestion
    std::lock_guard lock(m_mutex);
```

stsoe (changes requested):
Minor

src/runtime_src/hip/core/memory.cpp
zhangchiming: fixed

src/runtime_src/hip/core/memory.cpp
zhangchiming: fixed.

stsoe (approved):

chvamshi-xilinx (approved):

chvamshi-xilinx (commented):
I missed one point. You need acquire lock in get_hip_mem_from_addr function also.
Please send a PR for that.

PR #8132, opened 2024-05-02
CR-1199042: Interface Tile Input Channel is not producing event ID for input_stalls (2024.1)

pgschuey:
PLIO are sometimes not getting profiled correctly

Replace DMA events in metric set when profiling PLIO

Results may be confusing since they're different for PLIO and GMIO
 
Tested on vek280

Document different results for PLIO and GMIO


jvillarre (approved):

jvillarre:
Merging the fix for the approved CR 1199042.  Ignoring the build failure as for the 2024.1 branch it is a false failure as described by the pipeline team.

PR #8138, opened 2024-05-03

zhangchiming:
Problem solved by the commit
1) returning wrong hip mem object when input address is 0
2) crash when trying to set xrt::bo object from 0 address as kernel argument

crash was discovered in developing unit test cases

1) return correct result in hip mem object look up from 0 address.
2) allow host app feed 0 as kernel argument.

Compiled and tested on Ubuntu 22.04 (running on Ryzen 7840).

Documentation impact (if any)
None.

### DIFF
--- a/src/runtime_src/hip/core/event.cpp
+++ b/src/runtime_src/hip/core/event.cpp
@@ -118,14 +118,16 @@ kernel_start::kernel_start(std::shared_ptr<stream> s, std::shared_ptr<function> 
         xrt_core::kernel_int::set_arg_at_index(r, arg->index, args[idx], arg->size);
         break;
       case karg::argtype::global : {
-        auto hip_mem = memory_database::instance().get_hip_mem_from_addr(args[idx]).first;
-        if (!hip_mem)
-          throw std::runtime_error("failed to get memory from arg at index - " + std::to_string(idx));
+        if (args[idx] != 0) {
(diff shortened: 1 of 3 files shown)


src/runtime_src/hip/core/event.cpp
```diff
@@ -118,14 +118,16 @@ kernel_start::kernel_start(std::shared_ptr<stream> s, std::shared_ptr<function>
...
       case karg::argtype::global : {
-        auto hip_mem = memory_database::instance().get_hip_mem_from_addr(args[idx]).first;
-        if (!hip_mem)
-          throw std::runtime_error("failed to get memory from arg at index - " + std::to_string(idx));
-
-        // NPU device is not coherent. We need to sync the buffer objects before launching kernel
-        if (hip_mem->get_type() != memory_type::device)
-          hip_mem->sync(xclBOSyncDirection::XCL_BO_SYNC_BO_TO_DEVICE);
-        r.set_arg(idx, hip_mem->get_xrt_bo());
+        if (args[idx] != 0) {
```
chvamshi-xilinx, line 121: Early exit/break
if ( !arg[idx] )
  break;

chvamshi-xilinx (approved):
Small suggestion.

src/runtime_src/hip/core/event.cpp
zhangchiming: fixed.

chvamshi-xilinx (approved):

PR #8150, opened 2024-05-08
Fix for CR-1178294: XRT AIE Status .json file does not include tile information for a tile where only memory is used (#8147)

saumyag-xilinx:
(cherry picked from commit e5960ad4e8049de8e8519d606fddeaf31285f466)


manikandan-xilinx:
Tests failure for centos 7.8 can be ignored. I have created a PR to remove that for 2024.1 https://github.com/Xilinx/XRT/pull/8151. once that is merged we will not see centos 7.8 tests running for 2024.1

PR #8151, opened 2024-05-08
Run on centos 8.1 tests

manikandan-xilinx:
Run only centos 8.1 tests as centos 7.8 is not supported


manikandan-xilinx:
It's a pipeline change. So no impact on tests

PR #8152, opened 2024-05-08
Added global platform structures for both USERPF & MGMTPF for all subdev files.

mnaguban:
Added 3 commits in this PR for customkernel.
commmit-1 -> Added global platform structures for both USERPF & MGMTPF for all subdev files.
commmit-2 ->Macro names added to subdev/ device files to differentiate between userpf and mgmtpf functions.
commmit-3 ->To separate userpf & mgmtpf modules, we used one flag instead of macro in the xocl_fdt blob functions.


mnaguban:
Created new pull request for the same. closing this duplicate pull request.

PR #8157, opened 2024-05-09
VITIS-11384 - [Preemption] XRT to support patching save/restore scratch pad address

dezhiAmd:

 preempt_save/restore bo are patched by gpu-va address of scrach-pad-mem bo

. Create scrach-pad-mem bo and use its address to patch preempt_save/restore bo 

1. Run xrt_elf_test to test elf-flow using dpu sequence with control packet, tests pass
3. Run xrt_elf_test to test elf-flow using dpu sequence without control packet, tests pass


src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -969,6 +972,8 @@ class module_sram : public module_impl
     if (m_ctrlpkt_bo) {
       m_column_bo_address.push_back({ m_ctrlpkt_bo.address(), m_ctrlpkt_bo.size() }); // NOLINT
     }
+    m_column_bo_address.push_back({ m_preempt_save_bo.address(), m_preempt_save_bo.size() }); // NOLINT
```
dezhiAmd, line 975: I am hoping firmware have a way to figure out whether there is control packet or not.
Hopefully, firmware will not think m_preempt_save_bo is m_ctrlpkt_bo

larry9523 (changes requested):
The save/restore control code should come from ELF but not hard coded in XRT.

dezhiAmd:
> The save/restore control code should come from ELF but not hard coded in XRT.

**The save/restore control code does come from elf. preempt_restore_1col is only used when the elf file itself does not have a preempt section.**

Quote the original wors form https://confluence.amd.com/display/AIE/Ryzen+AI+preemption+End-to-End:
https://gitenterprise.xilinx.com/XRT/aiebu/pull/14 provides a common library for assembler. It can be overwritten by customer's own same/restore method. In another word, the actual method used for save/restore context can be determined by compiler. In case compiler is not providing a method to save/restore context, assembler will add a default one.

larry9523:
> > The save/restore control code should come from ELF but not hard coded in XRT.
> 
> **The save/restore control code does come from elf. preempt_restore_1col is only used when the elf file itself does not have a preempt section.**
> 
> Quote the original wors form https://confluence.amd.com/display/AIE/Ryzen+AI+preemption+End-to-End: https://gitenterprise.xilinx.com/XRT/aiebu/pull/14 provides a common library for assembler. It can be overwritten by customer's own same/restore method. In another word, the actual method used for save/restore context can be determined by compiler. In case compiler is not providing a method to save/restore context, assembler will add a default one.

If ELF does not have save/restore, XRT should not add it. That means this control code does not support preemption

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -825,12 +853,16 @@ class module_sram : public module_impl
   std::shared_ptr<module_impl> m_parent;
   xrt::hw_context m_hwctx;
 
+  static constexpr size_t m_scratchmem_size = 512 * 1024;
```
larry9523, line 856: we should get the size from elf instead of hard code it in the case we want to save different size of l2 memory

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -328,11 +342,12 @@ class module_impl
 
   // Patch ctrlcode buffer object for global argument
   //
+  // @param bo_ctrlcode - bo conaining ctrlcode
```
satishra-xilinx, line 345: typo in the comment.

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -740,7 +755,8 @@ class module_elf : public module_impl
       m_restore_buf_exist = initialize_restore_buf(xrt_core::elf_int::get_elfio(m_elf), m_restore_buf);
       if (m_save_buf_exist != m_restore_buf_exist)
           throw std::runtime_error{ "Invalid elf because preempt save and restore is not paired" };
-      m_arg2patcher = initialize_arg_patchers(xrt_core::elf_int::get_elfio(m_elf));
+
+     m_arg2patcher = initialize_arg_patchers(xrt_core::elf_int::get_elfio(m_elf));
```
satishra-xilinx, line 759: indent issue, seems to be one space off

larry9523 (changes requested):
Looks good to me. The only change I suggest is to get scratch-pad-mem size from elf. If our assembler does not put it in elf, we should enhance the assembler

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -905,24 +942,41 @@ class module_sram : public module_impl
...
+
+    const auto& preempt_restore_data = parent->get_preempt_restore();
+    auto preempt_restore_data_size = preempt_restore_data.size();
+    if (preempt_restore_data_size > 0) {
+      m_preempt_restore_bo = xrt::bo{ m_hwctx, preempt_restore_data_size, xrt::bo::flags::cacheable, 1 /* fix me */ };
+      fill_bo_with_data(m_preempt_restore_bo, preempt_restore_data);
+    }
+
+    if ((preempt_save_data_size > 0) && (preempt_restore_data_size > 0)) {
+      m_scratchmem = xrt::ext::bo{ m_hwctx, m_scratchmem_size };
```
satishra-xilinx, line 973: how are we handling sync_bo of this buffer?

dezhiAmd:
> The save/restore control code should come from ELF but not hard coded in XRT.

@larry9523 , Currently assembler does not have this feature. Can this be done in a follow-up PR once assembler have this feature?

src/runtime_src/core/common/api/xrt_module.cpp
dezhiAmd: Elf currently does not have the size information. This should be done in a follow up PR.

larry9523:
> > The save/restore control code should come from ELF but not hard coded in XRT.
> 
> @larry9523 , Currently assembler does not have this feature. Can this be done in a follow-up PR once assembler have this feature?

I am OK if you add a TODO comment. cc @Himanshu-xilinx

src/runtime_src/core/common/api/xrt_module.cpp
dezhiAmd: Interface function get_scratchmem is added, once the result come back to this bo, use the function to sync bo.

larry9523 (approved):

satishra-xilinx (approved):

stsoe (approved):

PR #8158, opened 2024-05-10
Fixed rw_shared flag for user managed kernel

saifuddin-xilinx:
-- There is a issue for rw_shared flag for user managed kernel. 
-- Placed this configuration before coming out from this function for user managed kernel


saifuddin-xilinx:
retest this please

PR #8160, opened 2024-05-10
elf flow for transaction buffer

dezhiAmd:
Transaction buffer does not use CU argument name as patch symbol. Instead it uses argument index as patch symbol. This is different from dpu-sequence. This PR is trying to solve this issue.
1) Transaction buffer uses different host app code from dpu-sequence. The opcode for transaction buffer is always set when index is 0. This is the first assumption used to solve this issue.
2) Once a transaction buffer opcode is detected, Patcher uses string generated by index to search the patch method.
Transaction_buffer_Test is used to test this PR.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -323,10 +323,11 @@ public:
 
   // Patch ctrlcode buffer object for global argument
   //
-  // @param symbol - symbol name
+  // @param argname - argument name
+  // @param index - argument index
   // @param bo - global argument to patch into ctrlcode
   virtual void
-  patch(const std::string&, const xrt::bo&)
+  patch(const std::string&, size_t, const xrt::bo&)
   {
     throw std::runtime_error("Not supported");
   }
@@ -337,7 +338,7 @@ public:
   // @param value - patch value
   // @param size - size of patch value
   virtual void
-  patch(const std::string&, const void*, size_t)
+  patch(const std::string&, size_t, const void*, size_t)
   {
     throw std::runtime_error("Not supported");
   }
@@ -765,6 +766,15 @@ public:
 // where buffer object address of offset for each column.
 class module_sram : public module_impl
 {
+  // opcode from host app
+  // 1 is dpu_sequence
+  // 2 is transaction_buffer
+  static constexpr uint64_t Transaction_Buffer_Opcode = 2;
+  uint64_t m_opcode = 0;
+
+  // Transaction buffer actual argument start from index 3 in host code
+  static constexpr uint64_t Transaction_Buffer_Arg_Offset = 3;
+
   std::shared_ptr<module_impl> m_parent;
   xrt::hw_context m_hwctx;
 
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -961,18 +971,42 @@ class module_sram : public module_impl
...
-  patch(const std::string& argnm, const void* value, size_t size) override
+  patch(const std::string& argnm, size_t index, const void* value, size_t size) override
   {
     if (size > 8) // NOLINT
       throw std::runtime_error{ "patch_value() only supports 64-bit values or less" };
+    
+    auto arg_value = *static_cast<const uint64_t*>(value);
+    if (index == 0) {
+       // First argument is always the opcode, no need to patch
+       m_opcode = arg_value;
```
larry9523, line 996: Conceptually, we should not assume this. If we made this change, this piece of code is tightly bound to IPU DPU/TXN flow. I don't think it is a good idea to do so. I would rather search twice.

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -765,6 +766,15 @@ class module_userptr : public module_impl
 // where buffer object address of offset for each column.
 class module_sram : public module_impl
 {
+  // opcode from host app
+  // 1 is dpu_sequence
+  // 2 is transaction_buffer
+  static constexpr uint64_t Transaction_Buffer_Opcode = 2;
```
dezhiAmd, line 772: Check this line

src/runtime_src/core/common/api/xrt_module.cpp
dezhiAmd: It already tightly bound to IPU/DPU/TXN flow without this change.

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -961,18 +971,42 @@ class module_sram : public module_impl
...
+       m_opcode = arg_value;
+       return;
+    }
+    else {
+      if (Transaction_Buffer_Opcode == m_opcode) {
+        if (index < Transaction_Buffer_Arg_Offset)
+          return;
 
-    patch_value(argnm, *static_cast<const uint64_t*>(value));
+        const std::string index_string = std::to_string(index - Transaction_Buffer_Arg_Offset);
```
dezhiAmd, line 1004: We cannot avoid this

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -627,7 +646,10 @@ class module_elf : public module_impl
           throw std::runtime_error("Invalid offset " + std::to_string(offset));
 
         std::string argnm{ symname, symname + std::min(strlen(symname), dynstr->get_size()) };
-        std::string key_string = generate_key_string(argnm, buf_type);
+
+        std::string key_string = (is_pure_digit(argnm)) ? // If true then argnm is argument index
+                                 std::to_string(generate_unique_number(static_cast<uint64_t>(std::stoi(argnm)), static_cast<uint64_t>(buf_type))) :
+                                 generate_key_string(argnm, buf_type);
```
larry9523, line 652: Can you please add some comments to explain why are we appending buftype (digital version) to argname? Seems a bit more complicated to me. So want to understand what benefit do we get from here.

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -706,12 +728,17 @@ class module_elf : public module_impl
...
   bool
-  patch(uint8_t* base, const std::string& argnm, uint64_t patch, patcher::buf_type type) override
+  patch(uint8_t* base, const std::string& argnm, size_t index, uint64_t patch, patcher::buf_type type) override
   {
     const std::string key_string = generate_key_string(argnm, type);
     auto it = m_arg2patcher.find(key_string);
-    if (it == m_arg2patcher.end())
-      return false;
+    auto found_use_argument_name = (it == m_arg2patcher.end());
+    if (false == found_use_argument_name) {// Search using index
```
larry9523, line 736: this should be 
true ==

After change this, the test passed

src/runtime_src/core/common/api/xrt_module.cpp
larry9523: What benefit do we get by generating a Cantor Pairing key instead of using pure argidx?

src/runtime_src/core/common/api/xrt_module.cpp
stsoe: Huh :-)  No, it should be `if (cond)`

src/runtime_src/core/common/api/xrt_module.cpp
dezhiAmd: Use case 1:
    argname is "control-packet", without appending buftype, control-packet bo address will be used to patch instruction-buffer and patch control-packet itself. This is the root cause of VITIS-11623

Use case 2:
argname is "1", without appending buftype, ifm bo address will be used to patch instruction-buffer and control-packet regardless. But in a test case we may not need patch both

larry9523 (commented):
Overall looks good to me. But I am still not very clear why do we use the complicated Cantor pairing function to generate the key. What benefit do we get from it comparing to purely use argidx (or + buftype if you prefer) as a key?

dezhiAmd:
> Overall looks good to me. But I am still not very clear why do we use the complicated Cantor pairing function to generate the key. What benefit do we get from it comparing to purely use argidx (or + buftype if you prefer) as a key?

It is a mistake in my code leading to the use of Cantor pairing function. There is no need to convert string to int when building patch-map. Commit e8b22f8 fixed this issue

larry9523 (approved):

stsoe (approved):

PR #8163, opened 2024-05-11
Fixing reserved memory logic in zocl

chvamshi-xilinx:
Fixing memory allocation algo using reserved memory logic in zocl

Issue is there from long time

Checking whether start address and size are exactly available in dt or not 

None

Verified multiple edge cases

None


chvamshi-xilinx:
Will cherry-pick from master as master pipeline is now working

PR #8166, opened 2024-05-14
Implementation of read/write mem and reg of xrt::aie::device 

rbramand-xilinx:
Added new apis for reading/writing AIE memory and registers in xrt::aie::device class

Added these new apis in xrt::aie::device class as the functions are related to aie device
These apis can be used only in admin mode
Logic for checking admin/root user will be added in corresponding shim layer that support these apis

Low as apis work only in admin mode

Tested on phoenix class of devices using MCDM flow

Documented the apis in the function signature

### DIFF
--- a/src/runtime_src/core/common/api/xrt_device.cpp
+++ b/src/runtime_src/core/common/api/xrt_device.cpp
@@ -396,6 +396,87 @@ open_context(xrt::aie::device::access_mode am)
   core_device->open_aie_context(am);
 }
 
+static void
+get_abs_col(const xrt_core::device* device, uint16_t context_id, uint16_t& col)
+{
+  auto data = xrt_core::device_query_default<xrt_core::query::aie_partition_info>(device, {});
+  for (const auto& entry : data) {
+    if (std::stoi(entry.metadata.id) == context_id) {
+      col += entry.start_col;
+      if (col >= entry.num_cols)
+        throw std::out_of_range("col index out of range");
+      return;
+    }
+  }
+  throw std::runtime_error("requested context_id not found");
+}
+
+std::vector<char>
+device::
+read_aie_mem(uint16_t context_id, uint16_t col, uint16_t row, uint32_t offset, uint32_t size) const
+{
+  return xdp::native::profiling_wrapper("xrt::aie::device::read_aie_mem",
+    [this, &col, row, offset, size, context_id] {
+      try {
+        // calculate absolute col index
+        get_abs_col(get_handle().get(), context_id, col);
+        return get_handle()->read_aie_mem(col, row, offset, size);
+      }
+      catch (const xrt_core::query::no_such_key&) {
+        throw std::runtime_error("read_aie_mem is not supported on this platform");
+      }
+    });
+}
+
+size_t
+device::
+write_aie_mem(uint16_t context_id, uint16_t col, uint16_t row, uint32_t offset, std::vector<char>& data)
+{
+  return xdp::native::profiling_wrapper("xrt::aie::device::write_aie_mem",
+    [this, &col, row, offset, &data, context_id] {
+      try {
+        // calculate absolute col index
+        get_abs_col(get_handle().get(), context_id, col);
+        return get_handle()->write_aie_mem(col, row, offset, data);
+      }
+      catch (const xrt_core::query::no_such_key&) {
+        throw std::runtime_error("write_aie_mem is not supported on this platform");
+      }
+    });
+}
+uint32_t
+device::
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/common/api/xrt_device.cpp
```diff
@@ -396,6 +396,87 @@ open_context(xrt::aie::device::access_mode am)
...
+get_abs_col(const xrt_core::device* device, uint16_t context_id, uint16_t& col)
+{
+  auto data = xrt_core::device_query_default<xrt_core::query::aie_partition_info>(device, {});
+  for (const auto& entry : data) {
+    if (std::stoi(entry.metadata.id) == context_id) {
+      col += entry.start_col;
+      if (col >= entry.num_cols)
+        throw std::out_of_range("col index out of range");
+      return;
+    }
```
stsoe, line 409: While I am okay with this as is (it reads well), you may want to reverse/continue and use early-exit thereby saving one level of indentation.

src/runtime_src/core/common/api/xrt_device.cpp
```diff
@@ -396,6 +396,87 @@ open_context(xrt::aie::device::access_mode am)
   core_device->open_aie_context(am);
 }
 
+static void
+get_abs_col(const xrt_core::device* device, uint16_t context_id, uint16_t& col)
```
stsoe, line 400: For the purpose of ordering, let's move this function before the member function definitions within this namespace, e.g. move to line 383.

This function should return the result.  It is confusing without comments that `col` is both used as input and adjusted.  The function should probably return the new adjusted value rather than modify a reference argument; so pass by value, adjust, return adjusted, and maybe comment to explain what it means to get abs col.

src/runtime_src/core/include/xrt/xrt_aie.h
```diff
@@ -88,7 +88,98 @@ class device : public xrt::device
...
+   *  vector of bytes to be written
+   * @return
+   *  number of bytes written
+   *
+   * This function writes data to L1/L2 memory of AIE tile within a context
+   * ** This function works only in Admin mode **
+   */
+  XCL_DRIVER_DLLESPEC
+  size_t
+  write_aie_mem(uint16_t context_id, uint16_t col, uint16_t row, uint32_t offset, std::vector<char>& data);
```
stsoe, line 135: Is `data` changed?

src/runtime_src/core/common/api/xrt_device.cpp
```diff
@@ -396,6 +396,87 @@ open_context(xrt::aie::device::access_mode am)
...
+      }
+      catch (const xrt_core::query::no_such_key&) {
+        throw std::runtime_error("read_aie_mem is not supported on this platform");
+      }
+    });
+}
+
+size_t
+device::
+write_aie_mem(uint16_t context_id, uint16_t col, uint16_t row, uint32_t offset, std::vector<char>& data)
```
stsoe, line 433: const ref data?

stsoe (changes requested):

dezhiAmd:
Quote: "Tested on phoenix class of devices using MCDM flow"
What host code do we use for the test?

src/runtime_src/core/common/api/xrt_device.cpp
```diff
@@ -380,6 +380,25 @@ operator== (const device& d1, const device& d2)
...
+{
+  auto data = xrt_core::device_query_default<xrt_core::query::aie_partition_info>(device, {});
+  for (const auto& entry : data) {
+    if (std::stoi(entry.metadata.id) != context_id)
+      continue;
+
+    auto abs_col = col + entry.start_col;
+    if (abs_col >= entry.num_cols)
+      throw std::out_of_range("col index out of range");
+    return abs_col;
```
stsoe, line 397: This is fine, but I meant to say that you can reuse `col` now that is passed by value.  The increment you did was fine just not as reference.

stsoe (approved):

src/runtime_src/core/common/api/xrt_device.cpp
rbramand-xilinx: Yes Soren understood what you meant but I thought this way reads well

chvamshi-xilinx (approved):
Looks good. Do we also need to provide an API to end user to get the hw_context_id from hw_context?

rbramand-xilinx:
> Quote: "Tested on phoenix class of devices using MCDM flow" What host code do we use for the test?

This feature is for profiling team. They created application with the apis implemented in this PR to debug things when there is hang in application running on phoenix.

PR #8170, opened 2024-05-15

zhangchiming:
Problem solved by the commit
add hipMemsetAsync()/hipMemsetD32Async()/hipMemsetD16Async()/hipMemsetD8Async().

N/A

Compiled and tested on Ubuntu 22.04 (running on Ryzen 7840).

Documentation impact (if any)
None.


src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -229,6 +229,32 @@ namespace xrt::core::hip
     s_hdl->enqueue(command_cache.get(cmd_hdl));
   }
 
+  // fill data to dst async.
+  template<typename element_type> static void
+  hip_memset_async(void* dst, int value, size_t size, hipStream_t stream)
```
sonals, line 234: Second argument should be `element_type value`

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -229,6 +229,32 @@ namespace xrt::core::hip
...
+  {
+    auto hip_mem_info = memory_database::instance().get_hip_mem_from_addr(dst);
+    auto hip_mem_dst = hip_mem_info.first;
+    auto offset = hip_mem_info.second;
+    assert(hip_mem_dst->get_type() != xrt::core::hip::memory_type::invalid);
+    throw_invalid_value_if(offset + size > hip_mem_dst->get_size(), "dst out of bound in hip_memset!");
```
sonals, line 240: Should we print `dst` pointer also in the error message?

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -326,3 +352,31 @@ hipMemcpyHtoDAsync(hipDeviceptr_t dst, void* src, size_t size, hipStream_t strea
...
+// Fills the memory area pointed to by dev with the constant integer value for specified number of times.
+hipError_t
+hipMemsetD32Async(void* dst, int value, size_t count, hipStream_t stream)
+{
+  return handle_hip_memory_error([&] { xrt::core::hip::hip_memset_async<std::uint32_t>(dst, value, count*sizeof(std::uint32_t), stream); });
+}
+
+// Fills the memory area pointed to by dev with the constant integer value for specified number of times.
+hipError_t
+hipMemsetD16Async(void* dst, int value, size_t count, hipStream_t stream)
```
sonals, line 372: According to the spec the second argument should be `unsigned short value`

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -326,3 +352,31 @@ hipMemcpyHtoDAsync(hipDeviceptr_t dst, void* src, size_t size, hipStream_t strea
...
+{
+  return handle_hip_memory_error([&] { xrt::core::hip::hip_memset_async<std::uint16_t>(dst, value, count*sizeof(std::uint16_t), stream); });
+}
+
+// Fills the memory area pointed to by dev with the constant integer value for specified number of times.
+hipError_t
+hipMemsetD8Async(void* dst, int value, size_t count, hipStream_t stream)
```
sonals, line 379: According to the HIP spec the second argument should be` unsigned char value`
https://rocm.docs.amd.com/projects/HIP/en/latest/doxygen/html/group___memory.html

src/runtime_src/hip/core/event.cpp
```diff
@@ -118,19 +118,19 @@ kernel_start::kernel_start(std::shared_ptr<stream> s, std::shared_ptr<function>
...
-        r.set_arg(idx, hip_mem->get_xrt_bo());
+          auto hip_mem = memory_database::instance().get_hip_mem_from_addr(args[idx]).first;
+          if (!hip_mem)
+            throw std::runtime_error("failed to get memory from arg at index - " + std::to_string(idx));
+
+          // NPU device is not coherent. We need to sync the buffer objects before launching kernel
+          if (hip_mem->get_type() != memory_type::device)
+            hip_mem->sync(xclBOSyncDirection::XCL_BO_SYNC_BO_TO_DEVICE);
+          r.set_arg(idx, hip_mem->get_xrt_bo());
+        }
```
sonals, line 132: White space formatting is not consistent.

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -212,13 +212,13 @@ namespace xrt::core::hip
   static void
   hip_memcpy_host2device_async(hipDeviceptr_t dst, void* src, size_t size, hipStream_t stream)
   {
-    throw_invalid_value_if(!src, "Invalid src pointer in hipMemCpyH2DAsync!");
+    throw_invalid_value_if(!src, "Invalid src pointer in hip_memcpy_host2device_async!");
```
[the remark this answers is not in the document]
zhangchiming: fixed.

src/runtime_src/hip/api/hip_memory.cpp
zhangchiming: fixed.

src/runtime_src/hip/api/hip_memory.cpp
zhangchiming: fixed.

src/runtime_src/hip/api/hip_memory.cpp
zhangchiming: fixed

src/runtime_src/hip/api/hip_memory.cpp
zhangchiming: fixed

src/runtime_src/hip/core/event.cpp
zhangchiming: fixed.

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -212,23 +219,49 @@ namespace xrt::core::hip
...
+    auto element_count = size / element_size;
+    std::vector<element_type> host_src(element_count, value);
+
+    auto hip_stream = get_stream(stream);
+    throw_invalid_value_if(!hip_stream, "invalid stream handle "  + to_hex(stream) + " in " + __func__);
+
+    auto s_hdl = hip_stream.get();
+    auto cmd_hdl = insert_in_map(command_cache,
+                                std::make_shared<copy_buffer>(hip_stream, XCL_BO_SYNC_BO_TO_DEVICE, hip_mem_dst, host_src.data(), size, offset));
+    s_hdl->enqueue(command_cache.get(cmd_hdl));
```
sonals, line 263: We are capturing the buffer pointer behind the `host_src` vector. But `host_src` will go out of scope after this function returns, so will have a stale pointer in `copy_buffer`?

sonals (changes requested):
Please review the use of host_ptr vector.

src/runtime_src/hip/api/hip_memory.cpp
zhangchiming: fixed.

sonals (changes requested):
Please review the memset implementation so we do not allocate buffers twice and drop the extra copy operation.

src/runtime_src/hip/api/hip_memory.cpp
zhangchiming: fixed.

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -212,23 +219,52 @@ namespace xrt::core::hip
...
+    auto element_size = sizeof(element_type);
+
+    throw_invalid_value_if((element_size != 1 && element_size != 2 && element_size != 4), std::string("invalid element_type in ") + __func__);
+    throw_invalid_value_if(size % element_size != 0, "invalid size " + std::to_string(size) + " in " + __func__);
+
+    auto element_count = size / element_size;
+    std::vector<element_type> vec(element_count, value);
+    auto host_src = std::shared_ptr<uint8_t>(new uint8_t[size]);
+    memcpy(host_src.get(), vec.data(), size);
+
```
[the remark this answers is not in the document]
zhangchiming: fixed.

stsoe (changes requested):

src/runtime_src/hip/core/event.cpp
```diff
@@ -171,6 +171,12 @@ bool kernel_start::wait()
   return false;
 }
 
+copy_buffer::copy_buffer(std::shared_ptr<stream> s, xclBOSyncDirection direction, std::shared_ptr<memory> buf, std::shared_ptr<uint8_t> ptr, size_t size, size_t offset)
+  : command(std::move(s)), cdirection(direction), buffer(std::move(buf)), host_ptr(std::move(ptr).get()), copy_size(size), dev_offset(offset)
+{
+  ctype = type::buffer_copy;
```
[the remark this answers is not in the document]
zhangchiming: switched to std::vector.

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -212,23 +219,54 @@ namespace xrt::core::hip
...
+    assert(hip_mem_dst->get_type() != xrt::core::hip::memory_type::invalid);
+    throw_invalid_value_if(offset + size > hip_mem_dst->get_size(), "dst: " + to_hex(dst) + " out of bound in " + __func__);
+
+    auto element_size = sizeof(element_type);
+
+    throw_invalid_value_if((element_size != 1 && element_size != 2 && element_size != 4), std::string("invalid element_type in ") + __func__);
+    throw_invalid_value_if(size % element_size != 0, "invalid size " + std::to_string(size) + " in " + __func__);
+
+    auto element_count = size / element_size;
+    auto host_src = std::shared_ptr<uint8_t>(new uint8_t[size]);
```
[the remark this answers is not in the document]
zhangchiming: switched to std::vector.

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -212,23 +219,49 @@ namespace xrt::core::hip
...
+    throw_invalid_value_if(offset + size > hip_mem_dst->get_size(), "dst: " + to_hex(dst) + " out of bound in " + __func__);
 
     auto hip_stream = get_stream(stream);
-    throw_invalid_value_if(!hip_stream, "Invalid stream handle in hipMemCpyH2DAsync!");
+    throw_invalid_value_if(!hip_stream, "invalid stream handle " + to_hex(stream) + " in " + __func__);
    
     auto s_hdl = hip_stream.get();
     auto cmd_hdl = insert_in_map(command_cache,
-                                std::make_shared<copy_buffer>(hip_stream, XCL_BO_SYNC_BO_TO_DEVICE, hip_mem_dst, src, size, offset));
+                                std::make_shared<copy_buffer<uint8_t>>(hip_stream, XCL_BO_SYNC_BO_TO_DEVICE, hip_mem_dst, src, size, offset));
```
sonals, line 235: copy_buffer<char> instead of copy_buffer<uint8_t>

zhangchiming: > copy_buffer instead of copy_buffer<uint8_t>

This will generate compile error: "error: no matching function for call to 'make_shared'"

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -15,6 +15,14 @@
...
 namespace xrt::core::hip
 {
+  static std::string
+  to_hex(const void* addr)
+  {
+    std::stringstream str;
+    str << std::hex << addr;
+    return str.str();
+  }
+
```
sonals, line 25: Please move this help code to core/common/utils.h

src/runtime_src/hip/api/hip_memory.cpp
sonals: Sorry, meant `copy_buffer<char>`

zhangchiming: done.

src/runtime_src/hip/core/event.cpp
```diff
@@ -171,21 +169,29 @@ bool kernel_start::wait()
...
+}
+
+template<typename T>
+bool copy_buffer<T>::submit()
 {
   switch(cdirection)
   {
     case XCL_BO_SYNC_BO_TO_DEVICE:
-      handle = std::async(std::launch::async, &memory::write, buffer, host_ptr, copy_size, 0, dev_offset);
+      handle = std::async(std::launch::async, &memory::write, buffer, host_ptr?host_ptr:host_vec->data(), copy_size, 0, dev_offset);
```
sonals, line 190: white space for ternary operator.

sonals (changes requested):

src/runtime_src/hip/core/event.h
```diff
@@ -135,6 +141,7 @@ class copy_buffer : public command
   xclBOSyncDirection cdirection;
   std::shared_ptr<memory> buffer;
   void* host_ptr;
+  std::shared_ptr<std::vector<T>> host_vec;
```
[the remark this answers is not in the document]
zhangchiming: removed host_vec.

src/runtime_src/hip/core/event.cpp
zhangchiming: fixed.

src/runtime_src/xocl/core/device.cpp
```diff
@@ -43,14 +44,7 @@ namespace {
...
 
-static
-std::string
-to_hex(void* addr)
-{
-  std::stringstream str;
-  str << std::hex << addr;
-  return str.str();
-}
+using namespace xrt_core::utils;
```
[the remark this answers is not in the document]
zhangchiming: fixed.

sonals:
Please rebase and resolve the conflicts

zhangchiming:
> Please rebase and resolve the conflicts

Done.

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -75,7 +76,7 @@ namespace xrt::core::hip
     assert(device_ptr);
 
     auto hip_mem = memory_database::instance().get_hip_mem_from_addr(host_ptr).first;
-    throw_invalid_value_if(!hip_mem, "Error getting device pointer from host_malloced memory!");
+    throw_invalid_value_if(!hip_mem, "Error getting device pointer from host pointer " + xrt_core::utils::to_hex(host_ptr) + " in " + __func__);
```
stsoe, line 79: Too expensive to call `to_hex()` and string append on success path.  
I don't know the signature of `throw_invalid_value_if`, but it should take no more than condition and `const char*`.

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -92,7 +93,7 @@ namespace xrt::core::hip
   hip_free(void* ptr)
   {
     auto hip_mem = memory_database::instance().get_hip_mem_from_addr(ptr).first;
-    throw_invalid_handle_if(!hip_mem || hip_mem->get_type() != memory_type::device, "Invalid handle passed to hipFree!");
+    throw_invalid_handle_if(!hip_mem || hip_mem->get_type() != memory_type::device, "invalid handle " + xrt_core::utils::to_hex(ptr) + " was passed to " + __func__);
```
stsoe, line 96: ditto here and all following instances.

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -212,23 +212,49 @@ namespace xrt::core::hip
...
+    auto element_count = size / element_size;
+    auto vec = std::make_shared<std::vector<T>>(element_count, value);
+
+    auto hip_stream = get_stream(stream);
+    throw_invalid_value_if(!hip_stream, "invalid stream handle "  + xrt_core::utils::to_hex(stream) + " in " + __func__);
+
+    auto s_hdl = hip_stream.get();
+    auto cmd_hdl = insert_in_map(command_cache,
+                                std::make_shared<copy_buffer<std::shared_ptr<std::vector<T>>>>(hip_stream, XCL_BO_SYNC_BO_TO_DEVICE, hip_mem_dst, vec, size, offset));
+    s_hdl->enqueue(command_cache.get(cmd_hdl));
```
stsoe, line 256: I am not following what is going on here and there are no comments to help me.  The shared std::vector does not look right and a templated copy_buffer instantiated from a shared_ptr to std::vector does not seem right either.  Please carefully comment non obvious code.

src/runtime_src/hip/core/event.cpp
```diff
@@ -171,22 +169,23 @@ bool kernel_start::wait()
   return false;
 }
 
-copy_buffer::copy_buffer(std::shared_ptr<stream> s, xclBOSyncDirection direction, std::shared_ptr<memory> buf, void* ptr, size_t size, size_t offset)
-  : command(std::move(s)), cdirection(direction), buffer(std::move(buf)), host_ptr(ptr), copy_size(size), dev_offset(offset)
+template<class T>
+copy_buffer<T>::copy_buffer(std::shared_ptr<stream> s, xclBOSyncDirection direction, std::shared_ptr<memory> buf, T host_buf, size_t size, size_t offset)
```
stsoe, line 173: This looks very complicated without understanding what is happening and why?

src/runtime_src/hip/core/event.h
```diff
@@ -124,17 +128,22 @@ class kernel_start : public command
   bool wait() override;
 };
 
+template<class T>
```
stsoe, line 131: This class and others and entire file is lacking comments.   What is the responsibility of `copy_buffer` and why is the class templated.  I am not saying this is wrong, I am just trying to understand.

stsoe (changes requested):

src/runtime_src/hip/api/hip_memory.cpp
stsoe: The ternary case you show in principle looks cheap.   

Calling throw_invalid_value_if(bool, const std::string&) is not kosher on success path.  You will incur at least one string ctor for the empty string; cheap, albeit annoying, especially when debugging.   I vaguer that the signature should be throw_invalid_value_if(bool, const char* msg).  If you need any fancy error message, I think it ought to be a special case.

zhangchiming: type of err_msg has been changed to const char*.

src/runtime_src/hip/api/hip_memory.cpp
zhangchiming: Ok, template use has been removed from class copy_buffer.

src/runtime_src/hip/core/event.cpp
zhangchiming: The original template use is just attempt for incorporate two use case: 1. source/destination host buffer is an opaque handle. 2. host buffer is source only and its storage need to be released by class copy buffer after async copy operation.

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -212,23 +212,49 @@ namespace xrt::core::hip
...
+    assert(hip_mem_dst->get_type() != xrt::core::hip::memory_type::invalid);
+    throw_invalid_value_if(offset + size > hip_mem_dst->get_size(), "dst out of bound.");
+
+    auto element_size = sizeof(T);
+
+    throw_invalid_value_if((element_size != 1 && element_size != 2 && element_size != 4), "Invalid element type.");
+    throw_invalid_value_if(size % element_size != 0, "Invalid size.");
+
+    auto element_count = size / element_size;
+    auto vec = std::make_shared<std::vector<T>>(element_count, value);
```
stsoe, line 248: Who owns this vec?   When I look at this code, I see the shared_ptr being passed into `copy_shared_host_buffer` only.   Why is the vector then a shared_ptr not just a plain vector moved into `copy_shared_host_buffer`, which then assumes ownership?

Basically, I am not able to understand what is going on.  It's possible everything is fine, but if so the comments are still lacking to explain what is done and why?

zhangchiming: You are correct. Using shared_ptr there is wrong. I made a mistake when trying to use template combine two type of copy source buffer into one copy_buffer class.

I have  removed shared_ptr usage in above code.

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -212,23 +212,49 @@ namespace xrt::core::hip
...
+
+    auto element_count = size / element_size;
+    std::vector<T> host_vec(element_count, value);
+
+    auto hip_stream = get_stream(stream);
+    throw_invalid_value_if(!hip_stream, "Invalid stream handle.");
+
+    auto s_hdl = hip_stream.get();
+    auto cmd_hdl = insert_in_map(command_cache,
+                                std::make_shared<copy_from_host_buffer_command<T>>(hip_stream, XCL_BO_SYNC_BO_TO_DEVICE, hip_mem_dst, host_vec, size, offset));
```
stsoe, line 255:
```suggestion
    auto cmd_hdl = insert_in_map(command_cache,
                                std::make_shared<copy_from_host_buffer_command<T>>(hip_stream, XCL_BO_SYNC_BO_TO_DEVICE, hip_mem_dst, std::move(host_vec), size, offset));
```

The vector is moved into the `copy_from_host_buffer_command`.  By using move, not only do we inform compiler of intent, we also make the code explicit that `host_vec` is no longer useable in this scope.

src/runtime_src/hip/core/event.h
```diff
@@ -124,22 +128,58 @@ class kernel_start : public command
...
   size_t dev_offset; // offset for device memory
   std::future<void> handle;
 };
 
+// copy command for copying data from a source only bost buffer of type std::vector<uint8|uint16|uint32>
+template<class T>
+class copy_from_host_buffer_command : public copy_buffer
+{
+public:
+  copy_from_host_buffer_command(std::shared_ptr<stream> s, xclBOSyncDirection direction, std::shared_ptr<memory> buf, std::vector<T>& host_vec, size_t size, size_t offset)
```
stsoe, line 155:
```suggestion
  copy_from_host_buffer_command(std::shared_ptr<stream> s, xclBOSyncDirection direction, std::shared_ptr<memory> buf, std::vector<T>&& host_vec, size_t size, size_t offset)
```
This must take an rvalue reference that can then be moved into the data member without incurring copy.

stsoe (commented):
Two minor but important changes.

src/runtime_src/hip/api/hip_memory.cpp
zhangchiming: Done.

src/runtime_src/hip/core/event.h
zhangchiming: Done.

src/runtime_src/hip/core/event.h
```diff
@@ -124,22 +128,58 @@ class kernel_start : public command
...
   std::future<void> handle;
 };
 
+// copy command for copying data from a source only bost buffer of type std::vector<uint8|uint16|uint32>
+template<class T>
+class copy_from_host_buffer_command : public copy_buffer
+{
+public:
+  copy_from_host_buffer_command(std::shared_ptr<stream> s, xclBOSyncDirection direction, std::shared_ptr<memory> buf, std::vector<T>&& vec, size_t size, size_t offset)
+    : copy_buffer(s, direction, buf, nullptr, size, offset), host_vec(vec)
```
stsoe, line 156: Ahh, you now removed std::move, I swear it was there before, and it is still needed.  Yeah, C++ is obscure.
```suggestion
    : copy_buffer(s, direction, buf, nullptr, size, offset), host_vec(std::move(vec))
```

stsoe (commented):
Improvement and one regression if I remember right :-)

src/runtime_src/hip/core/event.h
zhangchiming: Fixed.

stsoe (approved):
This is good now.   Thanks!

One thing I am vary about is overuse of std::shared_ptr, it communicates shared ownership, but I wonder if for example the commands in command_cache are really sharing ownership or are in fact the ultimate unique owner?  

I know the command_cache is storing std::shared_ptr and get() returns std::shared_ptr, this fine if and only if the shared_ptr returned by get() outlives the cache itself.  Is that the case?

chvamshi-xilinx (approved):

PR #8171, opened 2024-05-16
VITIS-11806 Break runlist into multiple submissions if necessary

stsoe:
Break an xrt::runlist into multiple submissions of max fixed size.

Device supports only fixed number of chained commands, so logic is
moved to XRT coreutil to break a runlist into multiple chunks.

The internal hwqueue API for runlist submission is changed to take a
std::span of buffer_handles rather than a vector.  This allows for
easier / cheaper submission as it avoids creating a new vector
every time a chunk is submitted.

To leverage std::span on compilers that do not yet support this, XRT
now includes Microsoft GSL (Guidelines Support Library) as a
submodule.  This library comes with gsl::span which we use until
std::span is supported by all compilers (c++20).

xrt::runlist is not yet used, so no immediate risk.
This code is more complicated than it should be.  To be revisited.

Tested by debugging simple example with multiple run objects in a runlist.  
Induced error via debugger.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -2957,7 +2957,7 @@ class runlist_impl
       if (status == std::cv_status::timeout)
         return status;
     }
-    catch (const xrt::run::command_error& ex) {
+    catch (const xrt::run::command_error&) {
       // errors are picked up by caller, here just change state
       m_state = state::error;
     }
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -2936,6 +2936,23 @@ class runlist_impl
     return st2str.at(st);
   }
 
+  // Pre-condition is that all commands in runlist have completed
+  // (error or not) or they have not been successfully submitted
+  // in the first place.  This function uses xrt::run::wait() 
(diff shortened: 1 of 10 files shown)


src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -2913,12 +2913,16 @@ class run::command_error_impl
 // class runlist_impl - The internals of a runlist
 class runlist_impl
 {
+  static constexpr size_t submit_size = 25;
```
maxzhen, line 2916: Let's use 24. We mis-calculated it. @mamin506 , please confirm.

maxzhen (changes requested):

src/runtime_src/core/common/api/xrt_kernel.cpp
stsoe: Sure, done.

maxzhen (approved):

sonals (commented):
Should we mention these third party libraries: ELFIO and GSL, in the NOTICE file?

stsoe:
> Should we mention these third party libraries: ELFIO and GSL, in the NOTICE file?

Yes, we already do for ELFIO, and I wanted to to do it for GSL also, but updates to NOTICE  breaks edge build which does a checksum check on LICENSE file.   Need to get help from @chvamshi-xilinx; I forget what to fix?

PR #8178, opened 2024-05-17
Fix for CR-1182821: AIE Status was unable to get good data and was outputting empty files

jvillarre:
Recently, AIE status regressed and is reporting that no valid information can be read from the AIE, and requests users to use xbutil instead to check the status.  This regression also possibly meant that the runtime AIE deadlock detection was also non-functional.  This pull request addresses the issue and re-enables proper functionality of AIE status.

The issue was due to race conditions in the destructors of global static objects inside XRT.  AIE status has a hook to catch when the global static objects are being destroyed, but there were several additional static objects that were necessary for proper functionality of the device query APIs that could be destroyed before the hooks were hit.

The hook for AIE status was moved to a newly created destructor of the zynq device and additional static objects were refactored.

Low risk as the order of destructors on edge affects only AIE status which has been tested.

The original failing test cases were verified, and a full regression test suite was run.

No documentation impact

### DIFF
--- a/src/runtime_src/core/edge/user/device_linux.cpp
+++ b/src/runtime_src/core/edge/user/device_linux.cpp
@@ -215,7 +215,7 @@ struct aie_core_info_sysfs
     const aie_sys_parser asp(aiepart);
 
     /* Loop each all aie core tiles and collect core, dma, events, errors, locks status. */
-    for (int i = 0; i < aie_meta .num_cols; ++i)
+    for (int i = 0; i < aie_meta.num_cols; ++i)
       for (int j = 0; j < (aie_meta.num_rows-1); ++j)
         ptarray.push_back(std::make_pair(std::to_string(i) + "_" + std::to_string(j),
                           asp.aie_sys_read(i,(j + aie_meta.core_row))));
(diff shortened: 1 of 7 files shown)


src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -212,12 +212,13 @@ struct aie_core_info_sysfs
     boost::property_tree::ptree ptarray;
     aie_metadata_info aie_meta = get_aie_metadata_info(device);
     const std::string aiepart = std::to_string(aie_meta.shim_row) + "_" + std::to_string(aie_meta.num_cols);
+    const aie_sys_parser asp(aiepart);
 
     /* Loop each all aie core tiles and collect core, dma, events, errors, locks status. */
-    for (int i = 0; i < aie_meta.num_cols; ++i)
+    for (int i = 0; i < aie_meta .num_cols; ++i)
```
pgschuey, line 218: Did you mean to have the space before the '.'?

pgschuey (approved):

src/runtime_src/core/edge/user/device_linux.cpp
jvillarre: No, I've pushed a commit to remove that unintentional space.

chvamshi-xilinx (approved):
Looks good. Merging and backporting to 2024.1 as this is must fix for release.

PR #8180, opened 2024-05-18
Support add_end in elf

dezhiAmd:
Current code base use all 32 bit  of the "rela->r_addend" as patching scheme. However, According to spec, Lower 4 bit means patching scheme. Higher 28 bit means add-end.
Change member variable holding an array of "offset" to an array of "Offset-to-patch-buffer + Offset-to-base-bo-Addr". Save add_end as Offset-to-base-bo-Addr.
Test result is good using:
    1. elf file generated from [Assembler PR 51](https://gitenterprise.xilinx.com/XRT/aiebu/pull/51) 
    2. SNL
 
Without this PR, test with the above conditions, result fails


src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -637,8 +640,10 @@ class module_elf : public module_impl
         if (auto search = arg2patchers.find(key_string); search != arg2patchers.end())
           search->second.m_ctrlcode_offset.emplace_back(offset);
         else {
-          auto symbol_type = static_cast<patcher::symbol_type>(rela->r_addend);
-          arg2patchers.emplace(std::move(key_string), patcher{ symbol_type, {offset}, buf_type });
+          uint64_t add_end_higher_28bit = (rela->r_addend & 0xFFFFFFF0) >> 4;
+          uint64_t patch_scheme_lower_4bit = rela->r_addend & 0xF; // NOLINT
+          auto symbol_type = static_cast<patcher::symbol_type>(patch_scheme_lower_4bit);
+          arg2patchers.emplace(std::move(key_string), patcher{ symbol_type, {offset}, buf_type, add_end_higher_28bit });
```
larry9523, line 646: addend (high 28 bits) do not belong to a symbol, they are distributed into each offset.

Relocation section '.rela.dyn' at offset 0x47e8 contains 12 entries:
 Offset     Info    Type            Sym.Value  Sym. Name + Addend
00001c80  0000010a unrecognized: a       00000000   4 + 785
00001ed8  0000020a unrecognized: a       00000000   3 + 5
00002018  0000030a unrecognized: a       00000000   4 + 505
000021b0  0000040a unrecognized: a       00000000   3 + 5
000022f0  0000050a unrecognized: a       00000000   4 + 285
00002488  0000060a unrecognized: a       00000000   3 + 5
000025c8  0000070a unrecognized: a       00000000   4 + 5
00002760  0000080a unrecognized: a       00000000   3 + 5
000042a0  0000090a unrecognized: a       00000000   5 + 785
00004398  00000a0a unrecognized: a       00000000   5 + 505
00004490  00000b0a unrecognized: a       00000000   5 + 285
00004588  00000c0a unrecognized: a       00000000   5 + 5

So looks to me, we need to expand m_ctrlcode_offset from std::vector<uint64_t> to std::vector<std::pair<uint64_t, uint32_t>>
First element is offset and second is addend

larry9523 (changes requested):
Tested on SNL and it does not work. I think the reason is addend is added to wrong place.

src/runtime_src/core/common/api/xrt_module.cpp
larry9523: Also, instead of naming it add_end or add_end_higher_28bit, I think a better more informative name should be offset_to_bobase. The other offset could be offset_to_patchbuffer.

larry9523:
@dezhiAmd , I have something working. It is a prototype. You can take a look at https://github.com/larry9523/XRT/commit/da96b7fa3929f52adbdd13df3fcb71bd460856d2

larry9523 (commented):
The changes look good. But we need to hold this PR a bit until we solve a conflict when handling patching at host and FW

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -140,14 +140,17 @@ struct patcher
   buf_type m_buf_type = buf_type::ctrltext;
   symbol_type m_symbol_type = symbol_type::shim_dma_48;
 
-  // Offsets from base address of control code buffer object
-  // The base address is passed in as a parameter to patch()
-  std::vector<uint64_t> m_ctrlcode_offset;
+  struct patch_info {
+      uint64_t offset_to_patch_buffer;
+      uint32_t offset_to_base_bo_addr;
```
larry9523, line 145: is this 4 space ident?

larry9523 (approved):
looks good except the ident nit.'

A couple of questions
1. did we pass the text with latest assembler PR where we clear the offset in BD?
2. what is the patching schema value for ctrl-packet in TXN flow? do we still need two patching schemas? (control_packet_48 and tansaction_ctrlrpkt_48, sim_dma_48 and tansaction_48)? If not, can we clean it up in next PR?

stsoe (approved):

PR #8186, opened 2024-05-20
Apply mask to rela->r_addend to get the patching scheme

dezhiAmd:
Current code base uses all 32 bit of the "rela->r_addend" as patching scheme. However, according to spec, only lower 4 bit means patching scheme
Apply mask to "rela->r_addend", then store it as patching scheme

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -408,6 +408,11 @@ public:
 // construct patcher objects for each argument.
 class module_elf : public module_impl
 {
+  // rela->addend have offset to base-bo-addr info along with schema
+  // [0:3] bit are used for patching schema, [4:31] used for base-bo-addr
+  constexpr static uint32_t Addend_Shift = 4;
+  constexpr static uint32_t Addend_Mask = ~((uint32_t)0) << Addend_Shift;
+  constexpr static uint32_t Schema_Mask = ~Addend_Mask;
   xrt::elf m_elf;
   uint8_t m_os_abi = Elf_Amd_Aie2p;
   std::vector<ctrlcode> m_ctrlcodes;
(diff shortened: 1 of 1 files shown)


larry9523 (approved):
SNL test passed

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -408,6 +408,11 @@ class module_impl
 // construct patcher objects for each argument.
 class module_elf : public module_impl
 {
+  // rela->addend have offset to base-bo-addr info along with schema
+  // [0:3] bit are used for patching schema, [4:31] used for base-bo-addr
+  constexpr static uint32_t Addend_Shift = 4;
+  constexpr static uint32_t Addend_Mask = ~((uint32_t)0) << Addend_Shift;
+  constexpr static uint32_t Schema_Mask = ~Addend_Mask;
```
stsoe, line 415: The coding style for better or worse in `core/common` is no `Camel_Case`, so could you change to `no_camel_case`  :-)

stsoe (changes requested):

src/runtime_src/core/common/api/xrt_module.cpp
dezhiAmd: Please refer to line 678: const ELFIO::Elf32_Rela
The code style is already there.
Please...

dezhiAmd: There are more:

      auto begin = reinterpret_cast<const ELFIO::Elf32_Rela*>(sec->get_data());
      auto end = begin + sec->get_size() / sizeof(const ELFIO::**Elf32_Rela**);
      for (auto rela = begin; rela != end; ++rela) {
        auto symidx = ELFIO::get_sym_and_type<ELFIO::Elf32_Rela>::get_r_sym(rela->r_info);

        auto dynsym_offset = symidx * sizeof(ELFIO::**Elf32_Sym**);

stsoe: We can hardly do anything about code that is not our code.
Please just follow the existing coding style of the code you edit.  And it seems like the 3 lines I pointed to are indeed our code lines.

stsoe:
When you get a chance. please run clang-tidy on xrt_module.cpp.   There are many errors we need to address.

PR #8190, opened 2024-05-21
Fixed ERT firmware packaging issue

fquinner:
When proving the user's own `sched*.bin` binaries, the `ERTFW_FILES` variable contains the required ERT firmware files but it is only used in one target firmware location in the target install directory structure. This fix ensure it is also added to the directory expected by the xrt base packages.

RPM package created does not contain the sched bin files in /opt/xilinx/xrt/share/fw. This causes the xrt base RPM to fail to build the required XSA shells so they in turn silently fail.

Discovered while trying to set up a Xilinx card with XRT for the first time.


chvamshi-xilinx:
\build

PR #8193, opened 2024-05-22
Fix boost library link issue when using static boost

rbramand-xilinx:
When using boost static build the package generated still has dependency on boost dynamic library (eg: libboost_system.so)
added changes to fix this issue

Issue raised by customer

Added boost static libraries path to the search path for libraries using link_directories
This way required static boost library is found when target_link_libraries is used and wrong dependency is not added

Ideal fix would be to remove boost_system dependency as we have moved to std::filesystem and this lib dependency is not needed any more. Same fix has been sent to master. But to keep things simpler going through this route

Low

Tested dependencies on xrt packages generated and dependencies are as expected

NA

### DIFF
--- a/src/CMake/nativeLnx.cmake
+++ b/src/CMake/nativeLnx.cmake
@@ -108,6 +108,11 @@ if (DEFINED ENV{XRT_BOOST_INSTALL})
     set (Boost_LIBRARY_DIRS $ENV{XRT_BOOST_INSTALL}/lib)
   endif()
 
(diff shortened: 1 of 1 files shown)


rbramand-xilinx:
Added do not merge label as this PR needs to be merged post 2024.1 release

stsoe (approved):

rbramand-xilinx:
Hi @chvamshi-xilinx, as 2024.1 branched out I have removed do not merge label. Please review and merge this PR

chvamshi-xilinx (approved):

PR #8194, opened 2024-05-22
VITIS-11806 Support runlist submission as chained command objects

stsoe:
Add new ERT_CMD_CHAIN opcode for chained command submssion of a runlist.  A runlist is broken into multiple chained commands if necessary.

This PR reworks #8171 and replaces disjoint set submission of individual command buffers with submission of a single ert_packet command buffer that chains individual commands.

The chained command is represented as an `ert_packet` where the payload is interpreted as `ert_cmd_chain_data` per ert.h.  The pay holds an array of buffer_handles that are to be submitted atomically for execution.

This chained command submission is somewhat easier to manage and is a step towards supporting command fusing where multiple run objects are stiched together into a single command without managing individual run objects.

This PR also parameterizes bo_cache with the size of bos needed. This is done because runlist will not currently need a 4K buffer for the chained ert packet.  This may change in future when runlist fuses the individual commands.  It is understood that 4K may in fact be the minimum underlying allocation size, but that is unbeknowst to the code that uses the bo cache.

No immediate risk as runlist use is a new feature

Extensive testing through debugger induced error conditions.  Tested on MCDM.

### DIFF
--- a/src/runtime_src/core/include/ert.h
+++ b/src/runtime_src/core/include/ert.h
@@ -60,7 +60,7 @@
 
 #ifdef _WIN32
 # pragma warning( push )
-# pragma warning( disable : 4201 )
+# pragma warning( disable : 4200 4201 )
 #endif
 
 #if defined(__GNUC__)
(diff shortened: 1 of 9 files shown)


src/runtime_src/core/include/ert.h
```diff
@@ -172,6 +172,24 @@ struct ert_dpu_data {
...
+ * struct ert_cmd_chain_data - interpretation of data payload for ERT_CMD_CHAIN
+ *
+ * @command_count: number of commands in chain
+ * @submit_index:  index of last successfully submitted command in chain
+ * @error_index:   index of failing command if cmd status is not completed
+ * @data[]:        address of each command in chain
+ *
+ * This is the payload of an *ert_packet* when the opcode is ERT_CMD_CHAIN
+ */
+struct ert_cmd_chain_data {
```
maxzhen, line 185: Can we also have the error status defined in this data structure so that driver does not need to update the error status for each chained command, when the list is processed by device.

stsoe: That's already there is the ert_packet itself.  It is indeed the goal that KMD will no longer update status of individual commands.

maxzhen (approved):

maxzhen (approved):

PR #8200, opened 2024-05-29
Bug fix in abs column calculation

rbramand-xilinx:
Fixed bug in validating abs column calculated
Added check in apis to throw when address provided is not 4 bytes aligned

pr - https://github.com/Xilinx/XRT/pull/8166 added wrong check

Low

Tested by running local application with aie read/write apis and test works as expected

NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_device.cpp
+++ b/src/runtime_src/core/common/api/xrt_device.cpp
@@ -392,7 +392,7 @@ get_abs_col(const xrt_core::device* device, uint16_t context_id, uint16_t col)
       continue;
 
     auto abs_col = col + entry.start_col;
-    if (abs_col >= entry.num_cols)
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/xrt_device.cpp
```diff
@@ -424,6 +431,7 @@ read_aie_mem(uint16_t context_id, uint16_t col, uint16_t row, uint32_t offset, u
       try {
         // calculate absolute col index
         auto abs_col = get_abs_col(get_handle().get(), context_id, col);
+        is_4byte_aligned(offset); // DRC check
```
stsoe, line 434: This is nit, but I'd call it `is_4byte_aligned_or_throw`

stsoe (approved):

chvamshi-xilinx (approved):

PR #8204, opened 2024-05-30
Always set chained in struct ert_dpu_data as 0 for aie2p

dezhiAmd:
Linux driver has an issue with chained field in ert_dpu_data for aie2p.
Control-packet address and size in ert_dpu_data is redundant in Windows KMD driver
Always set chained in struct ert_dpu_data as 0 for aie2p
Test using DPU sequence passed on PHX

### DIFF
--- a/src/runtime_src/core/common/api/module_int.h
+++ b/src/runtime_src/core/common/api/module_int.h
@@ -14,7 +14,7 @@
 
 namespace xrt_core::module_int {
 
-void
+uint32_t*
 fill_ert_dpu_data(const xrt::module& module, uint32_t* payload);
 
 
--- a/src/runtime_src/core/common/api/module_int.h
+++ b/src/runtime_src/core/common/api/module_int.h
@@ -19,6 +19,12 @@ namespace xrt_core::module_int {
 const std::vector<std::pair<uint64_t, uint64_t>>&
 get_ctrlcode_addr_and_size(const xrt::module& module);
 
+static constexpr uint8_t elf_amd_aie2p = 69;
+static constexpr uint8_t elf_amd_aie2ps = 64;
+
+// Return control code type such as aie2p or aie2ps
+const uint8_t
+get_os_abi(const xrt::module& module);
 
 // Patch buffer object into control code at given argument
 void
--- a/src/runtime_src/core/common/api/module_int.h
+++ b/src/runtime_src/core/common/api/module_int.h
@@ -14,17 +14,9 @@
 
 namespace xrt_core::module_int {
 
-// Provide access to underlying xrt::bo representing the instruction
-// buffer
-const std::vector<std::pair<uint64_t, uint64_t>>&
-get_ctrlcode_addr_and_size(const xrt::module& module);
+void
+get_ctrlcode_addr_and_size(const xrt::module& module, uint32_t* payload);
 
-static constexpr uint8_t elf_amd_aie2p = 69;
-static constexpr uint8_t elf_amd_aie2ps = 64;
-
-// Return control code type such as aie2p or aie2ps
-const uint8_t
-get_os_abi(const xrt::module& module);
 
 // Patch buffer object into control code at given argument
 void
--- a/src/runtime_src/core/common/api/module_int.h
+++ b/src/runtime_src/core/common/api/module_int.h
@@ -15,7 +15,7 @@
 namespace xrt_core::module_int {
 
 void
-get_ctrlcode_addr_and_size(const xrt::module& module, uint32_t* payload);
+fill_ert_dpu_data(const xrt::module& module, uint32_t* payload);
 
 
 // Patch buffer object into control code at given argument
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -2056,7 +2056,9 @@ class run_impl
       auto dpu = reinterpret_cast<ert_dpu_data*>(payload);
       dpu->instruction_buffer = addr;
       dpu->instruction_buffer_size = size;
-      dpu->chained = --ert_dpu_data_count;
+      dpu->chained = xrt_core::module_int::get_os_abi(m_module) == xrt_core::module_int::elf_amd_aie2p ?
+                     0 : // Always set chained as 0 for aie2p
+                     --ert_dpu_data_count;
       payload += sizeof(ert_dpu_data) / sizeof(uint32_t);
     }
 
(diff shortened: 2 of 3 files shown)


src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -2056,7 +2056,9 @@ class run_impl
       auto dpu = reinterpret_cast<ert_dpu_data*>(payload);
       dpu->instruction_buffer = addr;
       dpu->instruction_buffer_size = size;
-      dpu->chained = --ert_dpu_data_count;
+      dpu->chained = xrt_core::module_int::get_os_abi(m_module) == xrt_core::module_int::elf_amd_aie2p ?
```
dezhiAmd, line 2059: This is the main change for this PR

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -915,12 +913,10 @@ class module_sram : public module_impl
   }
 
   void
-  fill_bo_addresses()
+  aie2p_fill_bo_addresses()
   {
     m_column_bo_address.clear();
     m_column_bo_address.push_back({ m_instr_bo.address(), m_instr_bo.size() }); // NOLINT
-    if (m_ctrlpkt_bo)
```
dezhiAmd, line 922: These two lines are removed

src/runtime_src/core/common/api/xrt_kernel.cpp
stsoe: I would prefer to have this check within xrt_module implementation.   Can we avoid exposing the constants in module_int.h and instead add an API that fills up the entire payload or minimum returns a bool for either of the decisions taken here.

But based on what I see here, it may actually make more sense to encapsulate everything related to `addr_and_size` and `ert_dpu_data` within xrt_module.cpp.   Simply pass the full payload into an new internal module API, let the implementation collect and walk the `addr_and_size` and fill up the payload.

src/runtime_src/core/common/api/module_int.h
```diff
@@ -19,6 +19,12 @@ namespace xrt_core::module_int {
 const std::vector<std::pair<uint64_t, uint64_t>>&
 get_ctrlcode_addr_and_size(const xrt::module& module);
 
+static constexpr uint8_t elf_amd_aie2p = 69;
+static constexpr uint8_t elf_amd_aie2ps = 64;
+
+// Return control code type such as aie2p or aie2ps
+const uint8_t
+get_os_abi(const xrt::module& module);
```
stsoe, line 27: I think not needed, see next comment.

stsoe (commented):
Yeah, what's dangerous here is the raw uint32_t* payload which is now detached from caller in a non obvious fashion.

This is still internal API so assumption is that we know what we are doing, but please add comments regarding pre-condition assumptions about the payload size.

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1169,10 +1168,29 @@ class module_sram : public module_impl
...
+  fill_ert_dpu_data(uint32_t* payload) override
   {
-    return m_column_bo_address;
+    auto os_abi = m_parent.get()->get_os_abi();
+    if (os_abi == Elf_Amd_Aie2p) {
+      for (auto [addr, size] : m_column_bo_address) {
+        auto dpu = reinterpret_cast<ert_dpu_data*>(payload);
+        dpu->instruction_buffer = addr;
+        dpu->instruction_buffer_size = static_cast<uint32_t>(size);
+        dpu->chained = 0; // Always set chained as 0 for aie2p
```
maxzhen, line 1180: Since you are doing a loop here, is it really OK to set chained to 0? The definition of "chained" is to indicate that there are more "ert_dpu_data" followed and the loop does mean that there might be more than 1 ert_dpu_data?

maxzhen (commented):
I do not believe we need the change because Linux driver is not happy about chained not being 0. This is an API between user space and driver. If chained is required to be non-zero, we should fix the driver, not changing the user space code here. Do we have a new definition of the interface?

larry9523:
> I do not believe we need the change because Linux driver is not happy about chained not being 0. This is an API between user space and driver. If chained is required to be non-zero, we should fix the driver, not changing the user space code here. Do we have a new definition of the interface?

I am working on a new ERT command interface. Will raise a new PR merging changes from this one too.

src/runtime_src/core/common/api/xrt_module.cpp
dezhiAmd: According to @larry9523 , 
**The definition of "chained" is to indicate that there are more "ert_dpu_data" followed**
only applies to aie2ps.

This is why I start this PR.

dezhiAmd:
> > I do not believe we need the change because Linux driver is not happy about chained not being 0. This is an API between user space and driver. If chained is required to be non-zero, we should fix the driver, not changing the user space code here. Do we have a new definition of the interface?
> 
> I am working on a new ERT command interface. Will raise a new PR merging changes from this one too.

What @maxzhen is saying is this change is not needed, then why should we merge changes from this one?

larry9523:
> > > I do not believe we need the change because Linux driver is not happy about chained not being 0. This is an API between user space and driver. If chained is required to be non-zero, we should fix the driver, not changing the user space code here. Do we have a new definition of the interface?
> > 
> > 
> > I am working on a new ERT command interface. Will raise a new PR merging changes from this one too.
> 
> What @maxzhen is saying is this change is not needed, then why should we merge changes from this one?

Please refer to https://github.com/Xilinx/XRT/pull/8211

dezhiAmd:
> > > > I do not believe we need the change because Linux driver is not happy about chained not being 0. This is an API between user space and driver. If chained is required to be non-zero, we should fix the driver, not changing the user space code here. Do we have a new definition of the interface?
> > > 
> > > 
> > > I am working on a new ERT command interface. Will raise a new PR merging changes from this one too.
> > 
> > 
> > What @maxzhen is saying is this change is not needed, then why should we merge changes from this one?
> 
> Please refer to #8211

It appears to me that #8211 is a very big change. It will break all KMD driver.
With that being said, I think we should test this PR carefully and merge this PR. Then we should start testing #8211 (possible split #8211 in smaller changes)

stsoe (commented):
How does this compare to #8211 ?

larry9523:
> How does this compare to #8211 ?

I merged change in this PR into 8211. I think we can close this PR and review from there

larry9523:
https://github.com/Xilinx/XRT/pull/8211 was merged which include the fix from this one

PR #8213, opened 2024-06-04
added module name as parameter to function to avoid run time conflicts

rmeridoddi:
Added a module name as a parameter to functions in xocl driver and xclmgmt drivers as to avoid conflicts b/w xocl and xclmgmt drivers at run time .


chvamshi-xilinx:
Closing as we have changes in 2022.2 branch anyway.

PR #8223, opened 2024-06-05
removing edited from workflow file

manikandan-xilinx:
remove edited from workflow file as we do not want to trigger PR for editing names.


manikandan-xilinx:
Checking If comment triggers workflow

PR #8224, opened 2024-06-07
Add cstdint as necessary

ManojTakasi:
In GCC 13.1 usage of uint64 and similar will result in an error without #include <cstdint>


chvamshi-xilinx:
@manikandan-xilinx , can you please check why build is not triggerred?

stsoe (approved):

stsoe (commented):
Sort the includes cstdint would go first.

ManojTakasi:
Made changes as suggested. Thank you Soren.

PR #8226, opened 2024-06-10
add api to patch address directly

maxzhen:
Exposing API for patching address directly. This API will only be called by internal test cases in the future.
N/A
Exposing existing API.
No risk. The new API is not used at this moment.
Build XRT.
N/A


stsoe (approved):
Maybe add a comment in module_int.h to why this API is necessary compared to the xrt::bo one.  I know your PR comment says for testing and that makes sense, but please note this.

keryell:
It would be nice to have a clearer PR description. "Patch" and "control" are pretty vague.

larry9523 (approved):

PR #8231, opened 2024-06-11
Larger Debug Buffer Support

IshitaGhosh:
Enable use of larger debug buffer in ML Timeline, AIE Profile and AIE Debug on Client devices.

Due to limited buffer size on device, we could only get few hundred timestamps for ML Timeline feature which is far short of requirement for most ML designs.
With help of support in driver and ert, we can now sync from device to host when device buffer is full. This facilitates larger debug buffer sufficient to hold data from ML designs.

Unit and large models on device

### DIFF
--- a/src/runtime_src/core/tools/common/tests/TestGemm.cpp
+++ b/src/runtime_src/core/tools/common/tests/TestGemm.cpp
@@ -8,6 +8,9 @@
 #include "xrt/xrt_bo.h"
 #include "xrt/xrt_device.h"
 #include "xrt/xrt_kernel.h"
+
+#include "core/common/api/bo_int.h"
+
 namespace XBU = XBUtilities;
 
 // System - Include Files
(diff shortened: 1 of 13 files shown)


src/runtime_src/core/tools/common/tests/TestGemm.cpp
```diff
@@ -17,14 +20,7 @@ namespace XBU = XBUtilities;
...
 static constexpr size_t host_app = 1; //opcode
 static constexpr uint32_t num_of_cores = 32;
-static constexpr uint32_t size_4K   = 0x1000;
-static constexpr uint32_t offset_3K = 0x0C00;
-/* Each record timer entry has 32bit ID and 32bit AIE Timer low value.
-* The first 32 bit in the buffer is used to store total number 
-* of record timer entries written so far. So, max_count_in_size_3K is 1 less 
-* than total number of entries possible in 3K buffer section.
-*/ 
-static constexpr uint32_t max_count_in_size_3K = (offset_3K / (2 * sizeof(uint32_t))) - 1;
```
AShivangi, line 27: Thanks for removing this!

AShivangi (approved):

jvillarre (approved):

PR #8241, opened 2024-06-14
CR-1201062: Fixing issue where enabling both AIE trace and PL trace resulted in PL trace not being generated

jvillarre:
During application shutdown, if trace is active then all of the trace information from the device needs a final flush from memory to disk.  The order in which AIE trace and PL trace is offloaded is not defined, so either order has to be supported.  The AIE trace was removing the connection to the PL hardware after it had finished flushing AIE trace, which resulted in the PL trace to be lost.

This pull request fixes that issue by moving the ownership of the DeviceIntf object into a common location so each individual profiling/trace plugin doesn't have to be responsible for managing the shared memory.

A future planned pull request will change this implementation from raw pointers to shared_ptrs to further clarify ownership and make sure we catch all possible memory issues.

Additionally, this pull request renames several objects in profiling to make it clear which ones are associated with PL constructs as opposed to AIE constructs, as the generic names of "device" and "trace" were ambiguous.

The bug was due to the aie_trace plugin removing the object used to communicate with the PL portion of the device after AIE trace had been flushed.  It may have been introduced in recent refactorings, but was discovered with regression testing on multiple combinations of profiling options.

The problem was solved by no longer deleting the PL device interface at the end of AIE trace flushing.  This necessitated moving code to a common location and altering all of the profiling plugins that communicate with the PL side of the device.

Low risk of affecting profiling execution in some configuration.  Low risk of uncaught memory leaks to be addressed in future pull request.

The original failing test case has been tested on the board.  Full regression with all combinations of profiling options enabled needs to be run to fully verify.

No documentation impact.


vipangul (approved):

IshitaGhosh (commented):
As the PR renames DeviceIntf to PLDeviceIntf, we can drop "device" and just call it PLIntf. It can be done in a later PR too.

PR #8244, opened 2024-06-18
Use IO type and module type enums

vipangul:
Defines IO type enums and uses it to define PLIO or GMIO IO type.
Also module_type  has been used at applicable places.

NA

code readability enhancement.

Minimal as it adds enums in place of int variable types.

Verified on edge vck190


src/runtime_src/xdp/profile/database/static_info/device_info.cpp
```diff
@@ -318,13 +319,13 @@ namespace xdp {
       return ;
 
     switch (moduleType) {
-    case 0:
+    case static_cast<int>(module_type::core):
```
IshitaGhosh, line 322: Why do we need the static_cast?

IshitaGhosh (commented):
Thanks for the PR. This improves readability a lot.
Only one comment, why do we need the static_cast in the switch -case I pointed out?

src/runtime_src/xdp/profile/database/static_info/device_info.cpp
vipangul: This is because type safety provided by `enum class` over traditional enum. I have changed enum definition to traditional type for implicit conversion of enumerators.

jvillarre (approved):

PR #8245, opened 2024-06-20
CR-1203991: On Windows, don't issue incorrect profiling warning

jvillarre:
On Windows, profiling was issuing a warning stating that a profiling option is invalid, even if the user didn't specify that particular option.  This pull request makes it so that on Windows we don't use that option by default so the warning message won't be output unless the user specifically turns it on.  The option is still not supported on Windows.

On Windows, change the default of the option to false so the user must explicitly try to create a profiling writer instance with the option enabled before they see the warning message.

Low risk as the default behavior on Linux remains the same (where the option is supported).

Testing needed on client.

No documentation impact


IshitaGhosh (approved):
Change looks good.

PR #8248, opened 2024-06-24
CR-1203967: uninit value and unlocked mutex in icap.c

hlaccabu:
for CR-1203967: uninit value and unlocked mutex in icap.c

icap.c file had a function with an uninitialized int variable and also had an early exit without unlocking a mutex. 
Bugs / issues from above fixed, they were discovered by a Coverity scan. 
initialized int variable to 0 and included a line to unlock the mutex if early exit occurred.

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
@@ -697,7 +697,7 @@ static int icap_xclbin_validate_clock_req(struct platform_device *pdev,
 	struct icap *icap = platform_get_drvdata(pdev);
 	uint32_t slot_id = 0;
 	struct islot_info *islot = NULL;
-	int err;
+	int err = 0;
 
 	mutex_lock(&icap->icap_lock);
 	for (slot_id = 0; slot_id < MAX_SLOT_SUPPORT; slot_id++) {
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
```diff
@@ -697,7 +697,7 @@ static int icap_xclbin_validate_clock_req(struct platform_device *pdev,
 	struct icap *icap = platform_get_drvdata(pdev);
 	uint32_t slot_id = 0;
 	struct islot_info *islot = NULL;
-	int err;
```
xdavidz, line 700: @saifuddin-xilinx do you think the err=0 is good or err=-EINVAL should be set by default for this special case?

xdavidz:
> Can one of the admins verify this patch?

ok to test

xdavidz:
add to whitelist

src/runtime_src/core/pcie/driver/linux/xocl/subdev/icap.c
saifuddin-xilinx: err=0 is fine.

saifuddin-xilinx (approved):

xdavidz (approved):
looks good to me.

maxzhen (approved):

PR #8256, opened 2024-06-27
Fixing a few Coverity issues found in a couple files in /src/runtime_src/core/pcie/driver/linux/xocl/userpf

hlaccabu:
A few issues found by the Coverity scan that could cause problems. 
Copy into fixed size buffer, uninitialized scalar variable, and logically dead code in xocl_hwctx.c, xocl_ioctl.c, and xocl_sysfs.c, respectively.  
Use strlcpy() instead of strcpy(), initialized uninit variable to 0 (as is consistent with rest of code), and removed conditional that was logically dead.


hlaccabu:
Signed-off-by: Hayden Laccabue hlaccabu@amd.com

saifuddin-xilinx (approved):
Looks good to me.

stsoe:
Hi @hlaccabu .   Changes look good.   Can we change the title of the PR to include the CID numbers you have fixed?
We try to do "CID-number, number, number  ... " as PR title then include a brief summary in the description if necessary.
The CID-number is the Coverity ID of the defect you are fixing.

PR #8261, opened 2024-06-28
Fix issue where AIE profiling and PL device trace were not working together

jvillarre:
AIE profiling and PL device trace were working independently, but not when both were enabled.  Additionally, change the recently introduced raw pointers into unique_ptrs to prevent possible memory leaks.

When multiple plugins are active, each will attempt to register profiling metadata when an xclbin is loaded.  If the second plugin tries to register the same metadata, it will return early and not update any data structures.  The bug happens when the first plugin to register metadata doesn't need a connection to PL but the second plugin does.  The early return had to be adjusted to make sure cases where the PL interface was needed are covered correctly

PR 8241 introduced the bug.  This was discovered while debugging a separate issue in hardware emulation.

The bug was resolved by changing the function that determines if we need to refresh the profiling metadata to take into account the case where a PL interface is necessary but doesn't currently exist.

The test case on hardware emulation has been verified.  Full regression testing is necessary.

No documentation impact

### DIFF
--- a/src/runtime_src/xdp/profile/device/pl_device_intf.cpp
+++ b/src/runtime_src/xdp/profile/device/pl_device_intf.cpp
@@ -218,8 +218,6 @@ PLDeviceIntf::~PLDeviceIntf() {
   delete mFifoCtrl;
   delete mFifoRead;
   delete mDeadlockDetector;
-
-  //  delete mDevice;
 }
 
 void PLDeviceIntf::setDevice(std::unique_ptr<xdp::Device> devHandle) {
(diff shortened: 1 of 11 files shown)


src/runtime_src/xdp/profile/device/pl_device_intf.cpp
```diff
@@ -604,28 +602,28 @@ void PLDeviceIntf::readDebugIPlayout() {
       for (uint64_t i = 0; i < map->m_count; i++) {
         switch (map->m_debug_ip_data[i].m_type) {
         case AXI_MM_MONITOR:
-          mAimList.push_back(new AIM(mDevice, i, &(map->m_debug_ip_data[i])));
+          mAimList.push_back(new AIM(mDevice.get(), i, &(map->m_debug_ip_data[i])));
```
hackwa, line 605: Maybe these can also be converted into smart pointers using std::make_unique in future

hackwa (approved):

IshitaGhosh (approved):
Change looks good.

PR #8266, opened 2024-07-02
Infra: adding additional build params for windows - 2024.1

manikandan-xilinx:
This PR address the SR-954254. Added additional parameters for windows builds which can help support providing vs version and build command from the user itself


manikandan-xilinx:
This PR is a change to workflow file which would not impact tests. I have made some changes to 2024.2 branch for addressing the issue mentioned here SR-956183 in the workflow file and this PR is getting those changes.

PR #8242, opened 2024-06-17
VITIS-11024 enable hw_context support for xrt::graph objects

SravanKumarAllu-xilinx:
#### Problem solved by the commit

https://jira.xilinx.com/browse/VITIS-11024
currently we can load only one xclbin in graph supported devices , we cannot load multiple xclbins in multiple partitions in device..

#### Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered

N/A

#### How problem was solved, alternative solutions (if any) and why they were rejected

These first set of changes are enabling interface to support multiple xclbins in multiple partitions in device by creating "hardware context" associated with device id and xclbin. Interfaces are created for application developers and driver component owners to load mulitple xclbins in device by creating hardware context with particular device and particualr xclbin. New shim changes (driver changes) are required to fully utilize hardware context interfaces and fullfill actual requirements. currently used old shim by extracting device and xclbin from hardware context (loaded single device id with single xclbin)

#### Risks (if any) associated the changes in the commit

Documentation for this change might be required, should not break old flow where it still supports to load xclbin with device id. And as mentioned in above field new shim changes (driver changes) are required.

#### What has been tested and how, request additional testing if necessary

Used Graph test use case where input array and input factor are provided, ouput array would be generated by muliplying input array with input factor.

Initializing ADF API...
[ 611.693563] zocl-drm axi:zyxclmm_drm: zocl_create_client: created KDS client for pid(598), ret: 0
[ 611.699710] zocl-drm axi:zyxclmm_drm: zocl_destroy_client: client exits pid(598)
[ 611.774761] zocl-drm axi:zyxclmm_drm: zocl_create_client: created KDS client for pid(598), ret: 0
[ 620.614085] [drm] Loading xclbin 1482d52e-6e09-f433-e081-3e5e84b5c99f to slot 0
[ 620.614352] [drm] found kind 29(AIE_RESOURCES)
[ 620.616651] [drm] skip kind 8(IP_LAYOUT) return code: -22
[ 620.617634] [drm] found kind 9(DEBUG_IP_LAYOUT)
[ 620.618947] [drm] found kind 25(AIE_METADATA)
[ 620.619933] [drm] skip kind 7(CONNECTIVITY) return code: -22
[ 620.621097] [drm] found kind 6(MEM_TOPOLOGY)
[ 620.623921] [drm] Memory 0 is not reserved in device tree. Will allocate memory from CMA
[ 620.884871] [drm] AIE create successfully finished.
[...]

PR #8267, opened 2024-07-03
VITIS-11024 enable hw_context support for xrt::graph objects 2nd commit

SravanKumarAllu-xilinx:

https://jira.xilinx.com/browse/VITIS-11024
currently we can load only one xclbin in graph supported devices , we cannot load multiple xclbins in multiple partitions in device..

This is second commit for VITIS-11024
First commit: https://github.com/Xilinx/XRT/pull/8242

N/A

These first set of changes are enabling interface to support multiple xclbins in multiple partitions in device by creating "hardware context" associated with device id and xclbin. Interfaces are created for application developers and driver component owners to load mulitple xclbins in device by creating hardware context with particular device and particualr xclbin. New shim changes (driver changes) are required to fully utilize hardware context interfaces and fullfill actual requirements. currently used old shim by extracting device and xclbin from hardware context (loaded single device id with single xclbin)

Documentation for this change might be required, should not break old flow where it still supports to load xclbin with device id. And as mentioned in above field new shim changes (driver changes) are required.

Used Graph test use case where input array and input factor are provided, ouput array would be generated by muliplying input array with input factor.

Initializing ADF API...
[   99.707868] zocl-drm axi:zyxclmm_drm: zocl_create_client: created KDS client for pid(589), ret: 0
[   99.716043] zocl-drm axi:zyxclmm_drm: zocl_destroy_client: client exits pid(589)
[   99.778415] zocl-drm axi:zyxclmm_drm: zocl_create_client: created KDS client for pid(589), ret: 0
[  108.057897] [drm] Loading xclbin 63723cb0-4f53-4aa2-5ab0-60152b9af36b to slot 0
[  108.058281] [drm] found kind 29(AIE_RESOURCES)
[  108.062967] [drm] skip kind 8(IP_LAYOUT) return code: -22
[  108.065298] [drm] found kind 9(DEBUG_IP_LAYOUT)
[  108.067367] [drm] found kind 25(AIE_METADATA)
[  108.069022] [drm] skip kind 7(CONNECTIVITY) return code: -22
[  108.070547] [drm] found kind 6(MEM_TOPOLOGY)
[  108.073657] [drm] Memory 0 is not reserved in device tree. Will allocate memory from CMA
[  108.403376] [drm] AIE create successfully finished.
XAIEFAL: INFO: Resource group Avail is created.
XAIEFAL: INFO: Resource group Static is created.
XAIEFAL: INFO: Resource group Generic is created.
[  108.404487] [drm] zocl_xclbin_read_axlf 63723cb0-4f53-4aa2-5ab0-60152b9af36b ret: 0
[  108.585794] [drm] bitstream 63723cb0-4f53-4aa2-5ab0-60152b9af36b locked, ref=1
[  108.594516] zocl-drm axi:zyxclmm_drm:  ffff000001f5d410 kds_add_context: Client pid(589) add context Domain(65535) CU(0xffff) shared(true)
[  108.603146] zocl-drm axi:zyxclmm_drm:  ffff000001f5d410 kds_del_context: Client pid(589) del context Domain(65535) CU(0xffff)
graph run completed .....
trying alias 
graph rtp read completed .....
graph end completed .....
output Vector 1 value 3
output Vector 1 value 6
output Vector 1 value 9
output Vector 1 value 12
output Vector 1 value 15
output Vector 1 value 18
output Vector 1 value 21
output Vector 1 value 24
TEST PASSED

Graph doumentation may needs to be updated with "hwctx" things. And currently both kernel and graph support only c++ API's with "hardware context"

### DIFF
--- a/src/runtime_src/core/edge/user/aie/aied.cpp
+++ b/src/runtime_src/core/edge/user/aie/aied.cpp
@@ -30,14 +30,14 @@
 
 namespace zynqaie {
 
-Aied::Aied(xrt_core::device* device): mCoreDevice(device)
+aied::aied(xrt_core::device* device): m_device(device)
 {
   done = false;
-  pthread_create(&ptid, NULL, &Aied::pollAIE, this);
+  pthread_create(&ptid, NULL, &aied::poll_aie, this);
   pthread_setname_np(ptid, "Graph Status");
 }
 
-Aied::~Aied()
+aied::~aied()
 {
   done = true;
   pthread_kill(ptid, SIGUSR1);
@@ -45,16 +45,16 @@ Aied::~Aied()
 }
 
 /* Dummy signal handler for SIGTERM */
-static void signalHandler(int signum) {}
+static void signal_handler(int signum) {}
 
-void* 
-Aied::pollAIE(void* arg)
+void*
+aied::poll_aie(void* arg)
 {
-  Aied* ai = (Aied*)arg;
-  ZYNQ::shim *drv = ZYNQ::shim::handleCheck(ai->mCoreDevice->get_device_handle());
+  aied* ai = (aied*)arg;
+  ZYNQ::shim *drv = ZYNQ::shim::handleCheck(ai->m_device->get_device_handle());
   xclAIECmd cmd;
 
-  signal(SIGUSR1, signalHandler);
+  signal(SIGUSR1, signal_handler);
 
   if (!xrt_core::config::get_enable_aied())
     return NULL;
@@ -65,7 +65,7 @@ Aied::pollAIE(void* arg)
      * infinite for loop */
     sleep(1);
     /* Calling XRT interface to wait for commands */
-    if (ai->mGraphs.empty() || drv->xclAIEGetCmd(&cmd) != 0) {
+    if (ai->m_graphs.empty() || drv->xclAIEGetCmd(&cmd) != 0) {
       /* break if destructor called */
       if (ai->done)
         return NULL;
@@ -77,7 +77,7 @@ Aied::pollAIE(void* arg)
       boost::property_tree::ptree pt;
       boost::property_tree::ptree pt_status;
 
-      for (auto graph : ai->mGraphs) {
+      for (auto graph : ai->m_graphs) {
         pt.put(graph->getname(), graph->getstatus());
       }
 
@@ -98,14 +98,14 @@ Aied::pollAIE(void* arg)
 }
 
 void
-Aied::registerGraph(const graph_type *graph)
+aied::register_graph(const graph_instance *graph)
 {
-  mGraphs.push_back(graph);
+  m_graphs.push_back(graph);
 }
 
 void
-Aied::deregisterGraph(const graph_type *graph)
+aied::deregister_graph(const graph_instance *graph)
 {
-  mGraphs.erase(std::remove(mGraphs.begin(), mGraphs.end(), graph), mGraphs.end());
+  m_graphs.erase(std::remove(m_graphs.begin(), m_graphs.end(), graph), m_graphs.end());
 }
 }
--- a/src/runtime_src/core/edge/user/aie/graph.cpp
+++ b/src/runtime_src/core/edge/user/aie/graph.cpp
@@ -48,10 +48,12 @@ zynqaie::Aie* getAieArray()
 
 namespace zynqaie {
 
-graph_type::
-graph_type(std::shared_ptr<xrt_core::device> dev, const uuid_t uuid, const std::string& graph_name, xrt::graph::access_mode am)
-  : device(std::move(dev)), name(graph_name)
+graph_instance::
+graph_instance(std::shared_ptr<xrt_core::device> dev, std::string graph_name,
+              xrt::graph::access_mode am, const zynqaie::hwctx_object* hwctx, const xrt_core::uuid uuid)
+  : device{std::move(dev)}, name{std::move(graph_name)}, m_hwctxHandle{hwctx}
 {
+    auto xclbin_uuid = m_hwctxHandle ? m_hwctxHandle->get_xclbin_uuid() : uuid;
 #ifndef __AIESIM__
     auto drv = ZYNQ::shim::handleCheck(device->get_device_handle());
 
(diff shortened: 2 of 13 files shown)


chvamshi-xilinx (changes requested):
Hi @SravanKumarAllu-xilinx ,
Good cleanup again.
1) you should be able to remove all xclGraph* APIS from shim now.
2) please use class names "_" separated small letters. We dont use CamelCase
3) Some formatting changes in try catch blocks.

src/runtime_src/core/common/shim/hwctx_handle.h
```diff
@@ -53,6 +53,9 @@ class hwctx_handle
   virtual slot_id
   get_slotidx() const = 0;
 
+  virtual xrt::uuid
+  get_xclbin_uuid() const = 0;
```
stsoe, line 57: This will break all existing shims, and may in fact not be supported by all shims.   It would be better to default the implementation to throw an exception (not implemented or not supported).

Currently xrt::hw_context::get_xclbin_uuid() is implemented in the opaque impl class, not in shim.  I am not objecting to this change, just mentioning that it will not work without many changes.

I am also looking for where this function is even implemented in edge in this PR?

I want to mention that from a code architecture point of view `xrt_core -- depends on --> xrt_coreutil`, which implies that xrt_core can use xrt:: first class objects, e.g. xrt::hw_context.   From a run-time point of view, xrt_coreutil loads xrt_core, so at run time xrt_core is lower than xrt_coreutil, but coding wise it is reversed.  I like xrt_core using the handle classes only, but I have found this very difficult to do  without duplicating much code to each shim.  For example for the xclbin , yes, a hw context is associated with an xclbin, but where should this xclbin be managed?  In every shim?  Or in shared xrt_coreutil?  I prefer the latter.

stsoe (changes requested):

src/runtime_src/core/common/shim/hwctx_handle.h
chvamshi-xilinx: Hi @stsoe , I too have same point, xrt_core should just use the handles. Can you please point me the code where this PR using the xrt:: first class objects in xrt_core.  I see one include "xrt/xrt_hw_context.h" which needs to be fixed.
All other places are using handles only.

stsoe: @chvamshi-xilinx My point was opposite.  xrt_core can and does use xrt:: first class objects.  I am saying that in this PR you are forcing xclbin  management upon every shim, which is not cool.

One could argue that the shim handle classes should contain the shared data, and maybe that would be better, but that's now how it is designed today; e.g. all shim handles are pure interfaces with no data members.

chvamshi-xilinx: Thanks @stsoe  for the offline discussion.
Posting  the conclusion from our discussion here.
1) remove the xclbin_uuid function from base hwctx class. 
2) Store uuid/xclbin in shim level hwctx class if needed. 
3) Update all the shim level APIs to use shim level handles.

chvamshi-xilinx (approved):
Looks good.  Please club graph_object and graph_instance to single class in next PR.

src/runtime_src/core/edge/user/aie/aied.cpp
```diff
@@ -30,31 +30,31 @@
...
 }
 
 /* Dummy signal handler for SIGTERM */
-static void signalHandler(int signum) {}
+static void signal_handler(int signum) {}
 
-void* 
-Aied::pollAIE(void* arg)
+void*
+aied::poll_aie(void* arg)
```
stsoe, line 51: Can this thread routine communicate with a condition variable and avoid the sleep, along with signal handler?

src/runtime_src/core/edge/user/aie/graph.cpp
```diff
@@ -48,10 +48,13 @@ zynqaie::Aie* getAieArray()
 
 namespace zynqaie {
 
-graph_type::
-graph_type(std::shared_ptr<xrt_core::device> dev, const uuid_t uuid, const std::string& graph_name, xrt::graph::access_mode am)
-  : device(std::move(dev)), name(graph_name)
+graph_instance::
+graph_instance(std::shared_ptr<xrt_core::device> dev, const std::string& graph_name,
+              xrt::graph::access_mode am, const zynqaie::hwctx_object* hwctx, const xrt_core::uuid uuid)
+  : device{std::move(dev)}, name{graph_name}, m_hwctxHandle{hwctx}
```
stsoe, line 54: Since I believe you are copying the string (`graph_name`) clang-tidy will complain that it should have been passed by value and moved.

src/runtime_src/core/edge/user/aie/graph.cpp
```diff
@@ -48,10 +48,13 @@ zynqaie::Aie* getAieArray()
...
 {
+    auto xclbin_uuid = m_hwctxHandle ? m_hwctxHandle->get_xclbin_uuid() : uuid;
+    std::cout<<"xclbin_uuid is "<<xclbin_uuid<<std::endl;
```
stsoe, line 57: stray?

src/runtime_src/core/edge/user/graph_object.cpp
```diff
@@ -0,0 +1,199 @@
...
+#include "core/common/system.h"
+#include <memory>
+
+namespace zynqaie {
+  graph_object::graph_object(ZYNQ::shim* shim, const xrt::uuid& uuid , const char* name,
+                  xrt::graph::access_mode am, const zynqaie::hwctx_object* hwctx)
+    : m_shim{shim}
+  {
+    auto device{xrt_core::get_userpf_device(m_shim)};
+    std::string gname{name};
```
stsoe, line 14: Don't think this variable is needed.

src/runtime_src/core/edge/user/graph_object.cpp
```diff
@@ -0,0 +1,199 @@
...
+      m_graphInstance->reset();
+    }
+    catch (const std::exception& e)
+    {
+      xrt_core::send_exception_message(std::string("fail to reset graph: ") + e.what());
+    }
+    catch (...)
+    {
+      xrt_core::send_exception_message(std::string("fail to reset graph"));
+    }
```
stsoe, line 32: As for formatting.  Not sure about what convention is used throughout this file, but I prefer the braces on same line as condition or try/catch.
```suggestion
    try {
      m_graphInstance->reset();
    }
    catch (const std::exception& e)  {
      xrt_core::send_exception_message(std::string("fail to reset graph: ") + e.what());
    }
    catch (...) {
      xrt_core::send_exception_message(std::string("fail to reset graph"));
    }
```

src/runtime_src/core/edge/user/hwctx_object.cpp
```diff
@@ -0,0 +1,57 @@
...
+  // support hardware context, it still must implement a shim
+  // hardware context handle representing the default slot
+  //
+  //
+  hwctx_object::hwctx_object(ZYNQ::shim* shim, slot_id slotidx, xrt::uuid uuid, xrt::hw_context::access_mode mode)
+	  : m_shim(shim)
+	    , m_uuid(std::move(uuid))
+	    , m_slotidx(slotidx)
+	    , m_mode(mode)
+  {}
```
stsoe, line 18: Nice, this is how I prefer to initialized member variables in initializer list one line per variable.
Nit, align `,` with `:`

stsoe (commented):
This is very good.  Many style changes for the better --- thank you!.
My only real concern is the use of pthread, the thread routine, and the dummy signal handler.

src/runtime_src/core/edge/user/aie/aied.cpp
```diff
@@ -30,31 +30,31 @@
 
 namespace zynqaie {
 
-Aied::Aied(xrt_core::device* device): mCoreDevice(device)
+aied::aied(xrt_core::device* device): m_device(device)
 {
   done = false;
-  pthread_create(&ptid, NULL, &Aied::pollAIE, this);
+  pthread_create(&ptid, NULL, &aied::poll_aie, this);
   pthread_setname_np(ptid, "Graph Status");
```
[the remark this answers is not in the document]
SravanKumarAllu-xilinx: This is old code, just changed the function names and class names to proper ones. Regarding functionality we can work with another PR

src/runtime_src/core/edge/user/aie/aied.cpp
SravanKumarAllu-xilinx: This is old code, just changed the function names and class names to proper ones. Regarding functionality we can work with another PR

src/runtime_src/core/edge/user/graph_object.cpp
SravanKumarAllu-xilinx: "gname" is required here as "graph_instance"  constructor is expecting "std:string" , but we have only " char* "

src/runtime_src/core/edge/user/aie/aied.cpp
stsoe: Nah, std::string constructs from const char* so you don't need the variable.

src/runtime_src/core/edge/user/graph_object.cpp
stsoe: Nah, std::string constructs from const char* so you don't need the variable.

src/runtime_src/core/edge/user/hwctx_object.cpp
SravanKumarAllu-xilinx: done

chvamshi-xilinx (approved):

SravanKumarAllu-xilinx:
If there are no more feedback comments , can the review / approval be completed ?

stsoe (approved):

PR #8269, opened 2024-07-03
Add proper declspec define.  Amend #8242.

stsoe:
Amend #8242 to ensure hwctx APIs are exported rather than imported.

Slightly confusing here.  But all sources files in core/common must ensure that they export symbols implemented in core/common.    There are different macros used to export symbols, in particular the definition of XRT_API_EXPORT is controlled by XRT_API_SOURCE for export.   We should change the code to only use XRT_API_EXPORT and remove XRT_DRIVER_DLLSPEC which is legacy and not very clear.


chvamshi-xilinx (approved):
Thanks Soren.

PR #8271, opened 2024-07-03
VITIS-12786 - Fetch scratch-pad-mem size from elf

dezhiAmd:
scratch-pad-mem is a bo used to store L2 memory for the purpose of preemption. Currently scratch-pad-mem size is hardcoded instead of read from elf file
scratch-pad-mem size is encoded into the elf file by assembler. Refer to https://gitenterprise.xilinx.com/XRT/aiebu/pull/65/.
This PR parse the scratch-pad-mem size from the elf file.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -309,6 +309,12 @@ public:
       throw std::runtime_error("Not supported");
   }
 
+  [[nodiscard]] virtual size_t
+      get_scratch_pad_mem_size() const
+  {
+      throw std::runtime_error("Not supported");
+  }
+
   [[nodiscard]] virtual const control_packet&
   get_ctrlpkt() const
   {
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -658,6 +665,10 @@ class module_elf : public module_impl
           throw std::runtime_error("Invalid symbol name offset " + std::to_string(dynstr_offset));
         auto symname = dynstr->get_data() + dynstr_offset;
 
+        if (!m_scratch_pad_mem_size && (strcmp(symname, Scratch_Pad_Mem_Symbol) == 0)) {
```
dezhiAmd, line 668: Since m_scratch_pad_mem_size is initialized as 0, if there are two symbols with "scratch-pad-mem-symbol" in .dynsym section, only the first symbol will be used.

larry9523 (approved):
Looks good. Please work with @NishadSaraf to get it tested on preemption test case with small save/restore size.

NishadSaraf:
Hi @larry9523, and @dezhiAmd, I verified this change locally on Strix and it looks good. Thanks.

PR #8272, opened 2024-07-03
revise API to patch address directly

maxzhen:
Revise API introduced in PR#8226 for patching address directly. The original API requires module_sram, which, in turn, requires xrt::hwctx, which is not available for SHIM level test cases. Revising the API to only require module_elf, which is more friendly to SHIM test cases.

Run through a SHIM test case which uses this API and make sure the control buffer is patched properly.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -366,17 +366,6 @@ public:
     throw std::runtime_error("Not supported");
   }
 
-  // Patch ctrlcode buffer address for global argument
-  //
-  // @param argname - argument name
-  // @param index - argument index
-  // @param address - global argument to patch into ctrlcode
-  virtual void
-  patch(const std::string&, size_t, uint64_t)
-  {
-    throw std::runtime_error("Not supported");
-  }
-
   // Patch ctrlcode buffer object for scalar argument
   //
   // @param symbol - symbol name
@@ -1125,12 +1114,6 @@ class module_sram : public module_impl
   }
 
   void
-  patch(const std::string& argnm, size_t index, uint64_t address) override
-  {
-    patch_value(argnm, index, address);
-  }
-
-  void
   patch(const std::string& argnm, size_t index, const void* value, size_t size) override
   {
     if (size > 8) // NOLINT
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1285,9 +1268,35 @@ patch(const xrt::module& module, const std::string& argnm, size_t index, const x
 }
 
 void
-patch(const xrt::module& module, const std::string& argnm, size_t index, uint64_t address)
+patch(const xrt::module& module, uint8_t *buf, size_t *sz, const std::vector< std::pair<std::string, uint64_t> > *args)
```
stsoe, line 1271:
```suggestion
patch(const xrt::module& module, uint8_t* buf, size_t* sz, const std::vector<std::pair<std::string, uint64_t>>& args)
```
Please document this function.   It's not obvious that with `buf==nullptr`, the function returns the size the caller should allocate for `buf`.   What is `args` representing?

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1285,9 +1268,35 @@ patch(const xrt::module& module, const std::string& argnm, size_t index, const x
...
 {
-  module.get_handle()->patch(argnm, index, address);
+  auto hdl = module.get_handle();
+  size_t orig_sz = *sz;
+  const struct buf *inst = nullptr;
```
stsoe, line 1275:
```suggestion
  const buf* inst = nullptr;
```

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1285,9 +1268,35 @@ patch(const xrt::module& module, const std::string& argnm, size_t index, const x
...
+
+  *sz = inst->size();
+  if (orig_sz == 0)
+    return; // Caller is discovering real size of control code buffer.
+
+  if (orig_sz < *sz)
+    throw std::runtime_error{"Control code buffer passed in is too small"}; // Need a bigger buffer.
+  std::memcpy(buf, inst->data(), *sz);
+
+  for (size_t index = 0; index < args->size(); index++) {
```
stsoe, line 1295: Use structured binding.  Please pick descriptive names of what ever an element in args represents.

```suggestion
  for (auto& [str, size] : args) {
   ... 
  }
```

stsoe (changes requested):

src/runtime_src/core/common/api/xrt_module.cpp
maxzhen: The args here representing a list of arguments that needs to be patched to the control code buffer. I have added comments to make it clear that how caller can discover the size for buffer allocation. I still need to keep args as a pointer (v.s. a reference) since when caller tries to discover the real size, it does not pass in a real arg list (nullptr would do).

src/runtime_src/core/common/api/xrt_module.cpp
maxzhen: Fixed.

src/runtime_src/core/common/api/xrt_module.cpp
maxzhen: I added a loop counter and used more descriptive names in the binding.

stsoe (approved):

PR #8273, opened 2024-07-04
Fix for CR-1186978

aktondak:
This fixes the issue that an "abort" was thrown by windows when spawning more than 8 processes of xrt-smi simultaneously.

CR-1186978

The problem was solved by guarding the opening of hardware context and kernel creation in a try catch block. We do not support more than 8 contexts of hardware and the tool should throw an exception in such cases, instead of an abort from OS.

Tested with the test-case used in the CR. The test-case now fails with tool erroring the following instead of abort: 
Test 1 [00c5:00:01.1]     : df-bw
    Description           : Run bandwidth test on data fabric
    Xclbin                : C:\Windows\System32\DriverStore\FileRepository\ipukmddrv.inf_amd64_d029eeb7affac76f\validate_17f0_10.xclbin
    DPU-Sequence          : C:\Windows\System32\DriverStore\FileRepository\ipukmddrv.inf_amd64_d029eeb7affac76f\DPU_Sequence/df_bw.txt
    Details               : Buffer size: '1'GB
                            No. of iterations: '600'
    Error(s)              : Command failed to complete successfully
                            (ERT_CMD_STATE_ERROR)
    Test Status           : [FAILED]


src/runtime_src/core/tools/common/tests/TestCmdChainLatency.cpp
```diff
@@ -67,8 +67,17 @@ TestCmdChainLatency::run(std::shared_ptr<xrt_core::device> dev)
...
-  xrt::hw_context hwctx{working_dev, xclbin.get_uuid()};
-  xrt::kernel testker{hwctx, kernelName};
+
+  xrt::hw_context hwctx;
+  xrt::kernel testker;
+  try {
+    hwctx = xrt::hw_context(working_dev, xclbin.get_uuid());
+    testker = xrt::kernel(hwctx, kernelName);
+  }
+  catch (const std::runtime_error& ex){
```
AShivangi, line 77: please change this to std::exception

src/runtime_src/core/tools/common/tests/TestNPUThroughput.cpp
```diff
@@ -65,8 +65,17 @@ TestNPUThroughput::run(std::shared_ptr<xrt_core::device> dev)
...
-  xrt::hw_context hwctx{working_dev, xclbin.get_uuid()};
-  xrt::kernel testker{hwctx, kernelName};
+
+  xrt::hw_context hwctx;
+  xrt::kernel testker;
+  try {
+    hwctx = xrt::hw_context(working_dev, xclbin.get_uuid());
+    testker = xrt::kernel(hwctx, kernelName);
+  }
+  catch (const std::runtime_error& ex){
```
AShivangi, line 75: same

AShivangi (changes requested):
Let's add this catch for tct one and all column tests as well!

AShivangi (approved):
Looks good!

PR #8275, opened 2024-07-08
Add PID argument to aie mem and reg read/write apis

rbramand-xilinx:
Added PID argument to aie debug apis to uniquely identify hw ctx.

Context ID alone doesn't uniquely identify a hw context as per driver implementation, so added extra PID arg to debug APIs.

Low

Tested the flow on both Linux and windows and the apis work as expected.

Added doxygen comment for the arg added

### DIFF
--- a/src/runtime_src/core/common/api/xrt_device.cpp
+++ b/src/runtime_src/core/common/api/xrt_device.cpp
@@ -387,15 +387,12 @@ namespace xrt::aie {
 // of that partition. Get patition info using the context id passed and
 // convert relative column index to absolute using this info
 static uint16_t
-get_abs_col(const xrt_core::device* device, int32_t pid, uint16_t context_id, uint16_t col)
+get_abs_col(const xrt_core::device* device, pid_t pid, uint16_t context_id, uint16_t col)
 {
   auto data = xrt_core::device_query_default<xrt_core::query::aie_partition_info>(device, {});
   for (const auto& entry : data) {
-    // if PID info is present compare PID's else skip
-    if (entry.pid != -1 && (entry.pid != pid))
-      continue;
-
-    if (std::stoi(entry.metadata.id) != context_id)
+    // compare PID and context id
+    if (entry.pid != pid || std::stoi(entry.metadata.id) != context_id)
       continue;
 
     auto abs_col = col + entry.start_col;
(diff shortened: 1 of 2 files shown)


hackwa (approved):
looks good! If our tool can grab it from xrt-smi, that will work for us.

src/runtime_src/core/tools/common/reports/ReportAiePartitions.cpp
```diff
@@ -114,6 +116,7 @@ writeReport(const xrt_core::device* /*_pDevice*/,
       const auto& hw_context = pt_hw_context.second;
 
       const std::vector<std::string> entry_data = {
+        hw_context.get<std::string>("pid"),
```
AShivangi, line 119: We don't want to print -1 for Ryzen devices. Please do something like:
```
const auto device_class = xrt_core::device_query_default<xrt_core::query::device_class>(dev, xrt_core::query::device_class::type::alveo);
  switch (device_class) {
  case xrt_core::query::device_class::type::alveo:
  {
    hw_context.get<std::string>("pid"),
    break;
  }
```
and print this field only if it is populated

AShivangi (changes requested):
Please use [this ticket](https://jira.xilinx.com/browse/VITIS-12237 ) to track this change

src/runtime_src/core/tools/common/reports/ReportAiePartitions.cpp
rbramand-xilinx: Hi @AShivangi , I have already filed a CR to get this info on windows - https://jira.xilinx.com/browse/CR-1204588
If we dont print PID in case of Ryzen devices then users can't provide PID arg and the debug apis behavior will be different on different platforms. So added changes this way.
But that said I am holding off this PR till PID support is added in windows, there is discussion going on this and if it is difficult to provide PID in MCDM flow they will provide some unique identifier in its place.

AShivangi: Ok, sounds good!

rbramand-xilinx:
In both Linux and Windows we print PID now using xrt-smi, the same can be provided to debug apis.
Rebased the changes with latest XRT, please review. Thanks

stsoe (commented):
So are these APIs are called from a different process compared to the process that opened the hw_context (the pid argument)?

AShivangi:
@rbramand-xilinx, can you please update the description of this PR? It looks old

rbramand-xilinx:
> So are these APIs are called from a different process compared to the process that opened the hw_context (the pid argument)?

Yes Soren, it will be called from debugger process. Pair <context id, pid> uniquely identifies a hw_context

stsoe (approved):

AShivangi (approved):

chvamshi-xilinx (approved):

PR #8278, opened 2024-07-09
fix coverity issue - CID 395489:  Null pointer dereferences

maxzhen:
Fixing coverity issue: CID 395489:  Null pointer dereferences (FORWARD_NULL)
Introduced by PR#8272
Throw in case inst is nullptr, instead of dereferencing it.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -1277,12 +1277,14 @@ patch(const xrt::module& module, uint8_t* ibuf, size_t* sz, const std::vector<st
   if (hdl->get_os_abi() == Elf_Amd_Aie2p) {
     const auto& instr_buf = hdl->get_instr();
     inst = &instr_buf;
-  } else if(hdl->get_os_abi() == Elf_Amd_Aie2ps) {
+  }
+  else if(hdl->get_os_abi() == Elf_Amd_Aie2ps) {
     const auto& instr_buf = hdl->get_data();
     if (instr_buf.size() != 1)
       throw std::runtime_error{"Patch failed: only support patching single column"};
     inst = &instr_buf[0];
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1280,8 +1280,10 @@ patch(const xrt::module& module, uint8_t* ibuf, size_t* sz, const std::vector<st
   } else if(hdl->get_os_abi() == Elf_Amd_Aie2ps) {
```
stsoe, line 1280:
```suggestion
  } 
  else if (hdl->get_os_abi() == Elf_Amd_Aie2ps) {
```

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1280,8 +1280,10 @@ patch(const xrt::module& module, uint8_t* ibuf, size_t* sz, const std::vector<st
...
     const auto& instr_buf = hdl->get_data();
     if (instr_buf.size() != 1)
-      throw std::runtime_error{"Only support patching single column"};
+      throw std::runtime_error{"Patch failed: only support patching single column"};
     inst = &instr_buf[0];
+  } else {
```
stsoe, line 1285:
```suggestion
  } 
  else {
```

stsoe (approved):
Unrelated style comments

src/runtime_src/core/common/api/xrt_module.cpp
maxzhen: Ah, sure. Fixed.

src/runtime_src/core/common/api/xrt_module.cpp
maxzhen: Fixed.

PR #8296, opened 2024-07-18
Corrected the logic in the npu preempt data function

dkotipal-xilinx:
Signe-off-by: Durga Prasad <durgaprasad.kotipalli@amd.com>

Function returns properly as the logic corrected because the previous logic was incorrect

Tested and verified the functionality, all unit test cases are passing

### DIFF
--- a/src/runtime_src/core/include/ert.h
+++ b/src/runtime_src/core/include/ert.h
@@ -1087,7 +1087,7 @@ get_ert_npu_data(struct ert_start_kernel_cmd* pkt)
 static inline struct ert_npu_preempt_data*
 get_ert_npu_preempt_data(struct ert_start_kernel_cmd* pkt)
 {
-  if (pkt->opcode != ERT_START_NPU)
+  if (pkt->opcode != ERT_START_NPU_PREEMPT)
     return NULL;
 
   // past extra cu_masks embedded in the packet data


src/runtime_src/core/include/ert.h
```diff
@@ -1087,7 +1087,7 @@ get_ert_npu_data(struct ert_start_kernel_cmd* pkt)
 static inline struct ert_npu_preempt_data*
 get_ert_npu_preempt_data(struct ert_start_kernel_cmd* pkt)
 {
-  if (pkt->opcode != ERT_START_NPU)
+  if (pkt->opcode != ERT_START_NPU_PREEMPT)
```
dezhiAmd, line 1090: In my humble opinion, this check is not needed at all since this function is an internal function.
In another word, the person who call "get_ert_npu_preempt_data" should ensure that
pkt->opcode == ERT_START_NPU_PREEMPT
is true.

With that being said, we should add 
assert(pkt->opcode == ERT_START_NPU_PREEMPT)

dkotipal-xilinx: All functions in this file have similar checks, so maintaining symmetry is beneficial. Additionally, since xrt is a submodule in various code bases (e.g., driver, etc.), it's important to include this check and ensure that callers handle it properly.

chvamshi-xilinx: Hi @dkotipal-xilinx / @dezhiAmd ,
What do you want to do with this PR? Shall i merge this change as there are other functions using similar check and check in this function is incorrect? Or u remove checks in all functions.
For me, adding check is good foe sanity.

dezhiAmd: Merging this change is better than waiting. I realize our pipelines does not have the capability to catch assert under debug mode anyway.

We could do assert in another PR.

chvamshi-xilinx: Thanks @dezhiAmd for the confirmation. Merging this change.

PR #8300, opened 2024-07-22
Remove Hardcoded register space 64KB in shim and zocl

ManojTakasi:
[CR-1185445](https://jira.xilinx.com/browse/CR-1185445) XRT has hardcoded the register space for read/write to 64KB.

In the zocl side, we are hardcoding CU size as 64KB size, mmap is failing for IP's which has more than 64KB size .

Removed Hardcoded check 0x1000 from the user side and zocl side.

NA

Tested on vck190

NA

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/common/zocl_ioctl.c
+++ b/src/runtime_src/core/edge/drm/zocl/common/zocl_ioctl.c
@@ -86,6 +86,8 @@ out:
 	args->paddr = addr;
 	args->apt_idx = apt_idx;
 	args->cu_idx = cu_idx;
+        //update cu size based on the apt_index
+        args->cu_size = apts[apt_idx].size;
 	return 0;
 }
 
--- a/src/runtime_src/core/edge/drm/zocl/common/zocl_xclbin.c
+++ b/src/runtime_src/core/edge/drm/zocl/common/zocl_xclbin.c
@@ -303,6 +303,9 @@ zocl_update_apertures(struct drm_zocl_dev *zdev, struct drm_zocl_slot *slot)
 	int total = 0;
 	int apt_idx = 0;
 	int i = 0;
+        char kname[64] = {0};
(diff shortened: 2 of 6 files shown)


src/runtime_src/core/common/xclbin_parser.cpp
```diff
@@ -258,22 +259,30 @@ get_kernel_id(const pt::ptree& xml_kernel, const std::string& element)
...
+        continue;
+      for (auto& xml_remap : xml_inst.second) {
+        if (xml_remap.first != "addrRemap")
+          continue;
+        address_range = convert(xml_remap.second.get<std::string>("<xmlattr>.range"));
+        if (address_range != 0)
+          return address_range;
+        else
+          return default_address_range;
+      }
```
stsoe, line 280: This is not the same behavior.  The value returned is not the same after this change.   There are xclbins with kernel attribute "range=0xValue"  but same xclbins do not have range attribute.   I will vaguer that you cannot make this change.

stsoe (changes requested):

src/runtime_src/core/common/xclbin_parser.cpp
ManojTakasi: Thank you Soren for your inputs and for pointing that out. I haven't considered the case where xclbins might not have the range attribbute. I will make the necessary adjustments to ensure the behaviour remains consistent.

chvamshi-xilinx: Hi Manoj,
As discussed offline, Please remove this change and ask users to have their xclbin's updated with proper range information of slave,

ManojTakasi: Hi Vamshi, Soren,

The Changes have been made as suggest. could you pelase review them once again? Please let me know if any further changes are required.

chvamshi-xilinx (approved):

Issue #7909, opened 2024-01-25
Build failure(s) with pybind11


spaceotter:
First, the xrtdeps.sh script install pybind11 with pip as sudo, which is not recommended, the xrt build script also then fails to detect it. I installed pybind11 with a virtualenv and this was found by cmake.

Next, there is a build error:
```
In file included from /usr/include/pybind11/cast.h:16,
                 from /usr/include/pybind11/attr.h:13,
                 from /usr/include/pybind11/pybind11.h:13,
                 from /home/eric/src/xdna-driver/xrt/src/python/pybind11/src/pyxrt.cpp:20:                              
/usr/include/pybind11/detail/type_caster_base.h: In function ‘std::string pybind11::detail::error_string()’:            
/usr/include/pybind11/detail/type_caster_base.h:482:26: error: invalid use of incomplete type ‘PyFrameObject’ {aka ‘struct _frame’}
  482 |             frame = frame->f_back;                                                                              
      |                          ^~
```
I'm guessing it's related to this: https://docs.python.org/3/whatsnew/3.11.html#pyframeobject-3-11-hiding
Even though I'm using python 3.10, maybe pybind updated for 3.11 and this affected 3.10?

Edit: I might have been mistaken, it could be picking up an older pybind11-dev package in the system.

spaceotter:
The problems seem to originate from having python3.11 and/or pybind11-dev installed. In this case, for whatever reason the XRT build scripts always pick up the system python 3.11 even you activate a virtualenv. The xrtdeps.sh script installs the pybind11 dependency with the default python, which in ubuntu 22 is 3.10. And the pybind11-dev package is old and not compatible with 3.11 due to previously linked reason.

Quick fix is to use `sudo python3.11 -m pip install pybind11` - but that's not a great solution. I'd rather like to know why the XRT build script completely ignores the user environment...

ManojTakasi:
Hi @spaceotter , we have added a check in our CMakeLists.txt to determine if user is in a virtual environment. Based on that, we are setting the python3 exectable accordingly. With this PR i hpoe your issue is resolved. If you think it remains unresolved, please free to reopen this issue.

Thank you.

PR #8331, opened 2024-08-05
XRT OS Support for AlmaLinux 8 and pick correct python3 env if user is in virtualenv

ManojTakasi:
https://github.com/Xilinx/XRT/issues/7909 xrtdeps.sh script installs the pybind11 dependency with the default python.  XRT build scripts always pick up the system python3 even if there is activate a virtualenv. In alma8.10 python3.11 comes as dependency while installing libdrm-devel package but still /usr/bin/python3 points to default python3. findpackage(Python3) is directly taking the newest version alothough few packages are not installed like python3.11-pip python3.11-devel, pybind11.

setting python3 executable to /usr/bin/python3 as pybind11 is installed with this version. By adding appropiate checks in the build.sh file and modified xrtdeps.sh to identify the version for almalinux.

low

Tested with u250 in ubuntu 22.04  and almalinux 8.10.

Yes. We have to provide installatoin steps for almalinux

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -36,9 +36,9 @@ if [[ $CPU == "aarch64" ]] && [[ $OSDIST == "ubuntu" ]]; then
     fi
 fi
 
-# Use GCC 9 on CentOS 8 and RHEL 8 for std::filesystem
+# Use GCC 9 on CentOS 8, RHEL 8, AlmaLinux 8 for std::filesystem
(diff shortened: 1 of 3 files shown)


chvamshi-xilinx (approved):
Looks good. I am not sure whether there is any better solution other than keeping the python executable path to /usr/bin/python3

PR #8334, opened 2024-08-05
Starting Event trace from specific layer

jyothees99:
Overrun: While running an ML design, if there is any overrun of trace events in between, then there won't be any trace after the overrun. Instead of starting trace when the design starts, user can specify from which ML layer they want to start the trace. So that they can get the trace for layers which they were not able to before.
External Buffer size: For large ML designs, we need more space to store trace data. By only tracing range of layers, external buffer size for trace can be decreased.

1.  A user event will be generated in shim tile (0,0) by controller after each layer end and start of the run.
2. Performance counter at shim (0,0) is configured to count the above event.
3. Through xrt.ini, user passes start layer.
4. An event broadcast network is used to broadcast an event to all tiles when performance counter reaches the start layer.
5. All the trace modules are configured to start when the broadcast event reaches them.

No. Code will be executed only if user passes start_type=layer in xrt.ini
Tested resnet50 design on strix windows and partial resnet50 design on phoenix windows.

### DIFF
--- a/src/runtime_src/core/common/config_reader.h
+++ b/src/runtime_src/core/common/config_reader.h
@@ -863,6 +863,13 @@ get_aie_trace_settings_start_iteration()
   return value;
 }
 
+inline unsigned int
+get_aie_trace_settings_start_layer()
+{
+  static unsigned int value = detail::get_uint_value("AIE_trace_settings.start_layer", UINT_MAX);
+  return value;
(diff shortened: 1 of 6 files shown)


jvillarre (approved):
Approved if the name of the utility function is updated to getPartitionNumColumnsClient

PR #8338, opened 2024-08-06
VITIS-13074 - Dump common control codes before and after being patched

dezhiAmd:
The overall goal is to provide a debugging method when there is a hang caused by control-codes.
Add one flag **dump-control-codes** in file xrt.ini Debug section. When this flags is true, control-codes before and after patching are dumped for users to further identify the issue.

Refer to this confluence page for spec: https://confluence.amd.com/x/8avhV
**Problem1**: If application has many run-object. How to tell which set of pre- and post- patch bin file is for which run?
Ultimate solution is to add an id in elf file
Currently the id is a static member of class xrt::module_sram. It is incremented every time when constructor is called.

**Problem2**: minimize performance impact brought by debug flag
Selected solution:
At the phase of constructing xrt::module_sram, Check whether Dump-control-codes is true. If yes, record it in xrt::module_sram::m_debug_mode.

Alternative solution (rejected because concept of control-codes is also in xrt-device)
At the phase of creating xrt::device, Check whether Dump-control-codes is true. If yes, record it in xrt::device::m_debug_mode. The reason is to avoid checking Dump-control-codes in a use case with many xrt::module.

Test on strixb0 has been done using one of the standard transaction-buffer test cases from Larry.
The new entry in xrt.ini needs to be documented

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -4,6 +4,7 @@
 #define XRT_API_SOURCE         // exporting xrt_module.h
 #define XRT_CORE_COMMON_SOURCE // in same dll as core_common
 #include "core/common/config_reader.h"
+#include "core/common/message.h"
 #include "experimental/xrt_module.h"
 #include "experimental/xrt_elf.h"
 #include "experimental/xrt_ext.h"
@@ -28,6 +29,7 @@
 #include <map>
 #include <set>
 #include <string>
+#include <sstream>
 
 #ifndef AIE_COLUMN_PAGE_SIZE
 # define AIE_COLUMN_PAGE_SIZE 8192  // NOLINT
@@ -778,6 +780,18 @@ class module_elf : public module_impl
     }
 
     it->second.patch(base, patch);
+    if (xrt_core::config::get_xrt_debug()) {
+      if (not_found_use_argument_name) {
+        std::stringstream ss;
+        ss << "Patched " << patcher::section_name_to_string(type) << " use argument index " << index << " with value " << std::hex << patch;
+        xrt_core::message::send( xrt_core::message::severity_level::debug, "xrt_module", ss.str());
+      }
+      else {
+        std::stringstream ss;
+        ss << "Patched " << patcher::section_name_to_string(type) << "use argument name " << argnm << " with value " << std::hex << patch;
+        xrt_core::message::send( xrt_core::message::severity_level::debug, "xrt_module", ss.str());
+      }
+    }
     return true;
   }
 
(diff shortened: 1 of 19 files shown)


maxzhen (approved):

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -955,6 +949,36 @@ class module_sram : public module_impl
...
+  union debug_flag_union {
+    struct debug_mode_struct {
+      uint32_t dump_control_codes     : 1;
+      uint32_t dump_control_packet    : 1;
+      uint32_t dump_preemption_codes  : 1;
+      uint32_t reserved : 29;
+    } debug_flags;
+    uint32_t all;
+  }m_debug_mode = {};
+  static uint32_t s_id; //TODO: it needs come from the elf file
```
stsoe, line 961: Don't you need this as a regular non static data member.   The value is associated with the object (module_sram) once created.   If you use the static it is not specific to this object hence all the dump file names could be wrong.

```suggestion
  uint32_t m_id {0}; //TODO: it needs come from the elf file
```

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1280,6 +1329,7 @@ class module_sram : public module_impl
       return m_scratch_pad_mem;
   }
 };
+uint32_t module_sram::s_id = 0;
```
stsoe, line 1332: Remove

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1248,6 +1292,11 @@ class module_sram : public module_impl
     , m_parent{ std::move(parent) }
     , m_hwctx{ std::move(hwctx) }
   {
+    m_debug_mode.debug_flags.dump_control_codes = xrt_core::config::get_feature_toggle("Debug.dump_control_codes");
+    m_debug_mode.debug_flags.dump_control_packet = xrt_core::config::get_feature_toggle("Debug.dump_control_packet");
+    m_debug_mode.debug_flags.dump_preemption_codes = xrt_core::config::get_feature_toggle("Debug.dump_preemption_codes");
+    s_id++;
```
stsoe, line 1298:
```suggestion
    static std::atomic<uint32_t> s_id {0}
    m_id = s_id++;
    
    m_debug_mode.debug_flags.dump_control_codes = xrt_core::config::get_feature_toggle("Debug.dump_control_codes");
    m_debug_mode.debug_flags.dump_control_packet = xrt_core::config::get_feature_toggle("Debug.dump_control_packet");
    m_debug_mode.debug_flags.dump_preemption_codes = xrt_core::config::get_feature_toggle("Debug.dump_preemption_codes");
```

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -955,6 +949,36 @@ class module_sram : public module_impl
...
+    return m_debug_mode.debug_flags.dump_control_packet != 0;
+  }
+
+  bool
+  inline is_dump_preemption_codes() {
+    return m_debug_mode.debug_flags.dump_preemption_codes != 0;
+  }
+
+  uint32_t get_id() const {
+    return s_id;
```
stsoe, line 979:
```suggestion
    return m_id;
```

stsoe (changes requested):

larry9523 (approved):

PR #8344, opened 2024-08-11
XDP AIE Halt Plugin to set Debug Halt on AIE Cores for debugging on Client

IshitaGhosh:
Debugging tools for client require an initial setup before txn starts and so AIE cores need to be paused. This XDP AIE Halt plugin provides user-space solution for the setup requirement of debugging tool.
It sets debug halts on AIE cores of current design via txn.

Earlier, design control code needed to be recompiled with halts so that it can be debugging. This recompilation step is not needed now and saves time.

Unit tests. This has been used successfully with Debugger too.


jvillarre (approved):
Approved on the understanding that this should only be used on debugging cases where more there is only one hardware context created.

PR #8345, opened 2024-08-12
VITIS-11630 : P1 - Make CMake system in XRT modular

rbramand-xilinx:
This PR has changes to make XRT CMake build flow modular, as part of this attempt added initial changes targeting **xclbinutil**  as xclbinutil standalone build option is immediately needed for yocto flows.
Users can build xclbinutil only by using build.sh script option -xclbinutil

Created new CMake Util scripts under <XRT>/src/CMake folder that can be included into different modules
As xclbinutil doesn't depend on any other component of XRT adding just new CMake files is sufficient.

Low

Tested build for x86 linux, windows and embedded flows

May be needed once we finish working on all components


chvamshi-xilinx (approved):
Looks good and clean. We should be able to extend this to other folders easily after this.

rbramand-xilinx:
> This looks fine to me. Lots of changes. Have you verified a build from scratch comparing file content of {Debug,Release}/opt/xilinx/xrt before and after your changes? When making these sort of changes, I think it would be worthwhile to use `find <dir> -type f |sort` on the install dirs before and after changes.
> 
> Make sure to verify both Linux and Windows.
> 
> Second, please try build XRT-MCDM also with these changes and do the same validation.

Hi @stsoe, I have verified the contents of Debug and Release folders with and without my changes in both Linux and Windows and files are as expected and also XRT-MCDM build passed with my changes in xrt submodule.

PR #8347, opened 2024-08-13
Changing default slot for PS kernels to Zero

karthdmg-xilinx:
The existing zocl driver assumes the default slot for ps kernels to be zero but this is not aligned with the logic
implemented in host XRT driver code.
Changing host XRT logic such that both PS and PL kernel slots are downloaded to slot 0 by default.
This solution persist until we have a clear implementation for multi slot functionality.

Changing host XRT logic such that both PS and PL kernel slots are downloaded to slot 0 by default.
This solution persist until we have a clear implementation for multi slot functionality.

Changing host XRT logic such that both PS and PL kernel slots are downloaded to slot 0 by default.
This solution persist until we have a clear implementation for multi slot functionality.

NA
1. Program and Run PL and PL-AIE Kernels.
2. Program and Run  PS-AIE Kernels.

xocl_resolver implementation document

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-bifurcation-reset.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/mgmtpf/mgmt-bifurcation-reset.c
@@ -157,7 +157,7 @@ done:
 static long xclmgmt_hot_reset_post(struct xclmgmt_dev *lro, bool force)
 {
 	long err = 0;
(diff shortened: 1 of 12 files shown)


saifuddin-xilinx (approved):

karthdmg-xilinx:
[ps_pl_logs.txt](https://github.com/user-attachments/files/16608855/ps_pl_logs.txt)
Test pass logs attached for ps and pl kernels

chvamshi-xilinx (approved):

PR #8351, opened 2024-08-15
add new action mask to indicate the driver to program the PDI.

xvijaysri:
In this PR, added support for segmented configuration dynamic reload flows via a new --key-value option.
xclbinutil --key-value SYS:action_mask:LOAD_PDI
When XRT encounters that this action mask is set, it will program the PDI.

This is considered low risk change. Implementation similar to the existing LOAD_AIE action mask. 

I tested my changes locally by creating xclbin with the SYS:action_mask:LOAD_PDI option and verified that the output of the --info option on the xclbin contains LOAD_PDI action mask. This confirms that m_actionMask in axlf header is correctly set when SYS:action_mask:LOAD_PDI option is used and reads this LOAD_PDI bit mask from m_actionMask correctly when --info option is used. Further testing would be carried out by XRT to program the PDI.

### DIFF
--- a/src/runtime_src/tools/xclbinutil/XclBinClass.cxx
+++ b/src/runtime_src/tools/xclbinutil/XclBinClass.cxx
@@ -94,6 +94,7 @@ XclBin::initializeHeader(axlf& _xclBinHeader)
   memset(_xclBinHeader.m_keyBlock, 0xFF, sizeof(_xclBinHeader.m_keyBlock));
   _xclBinHeader.m_uniqueId = time(nullptr);
   _xclBinHeader.m_header.m_timeStamp = time(nullptr);
+  _xclBinHeader.m_header.m_actionMask = 0;
 
   // Now populate the version information
   getVersionMajorMinorPath(xrt_build_version,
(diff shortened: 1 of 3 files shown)


src/runtime_src/tools/xclbinutil/XclBinClass.cxx
```diff
@@ -1548,16 +1549,15 @@ XclBin::setKeyValue(const std::string& _keyValue)
...
-        }
+      if (sValue == "LOAD_AIE") {
+        m_xclBinHeader.m_header.m_actionMask |= AM_LOAD_AIE;
+      }
+      else if (sValue == "LOAD_PDI") {
+        m_xclBinHeader.m_header.m_actionMask |= AM_LOAD_PDI; 	
+      }
+      else {
+        auto errMsg = boost::format("ERROR: Unknown bit mask '%s' for the key '%s'. Key-value pair: '%s'.") % sValue % sKey % _keyValue;
+        throw std::runtime_error(errMsg.str());
```
xfreid, line 1560: The old code seems to support "mask1|mask2|...", but the new code assumes LOAD_AIE and LOAD_PDI are exclusive, have you confirmed that it is the expected behavior?

xvijaysri: The old code doesn't recognize command line option like SYS:action_mask:LOAD_AIE|LOAD_PDI. The developer didn't have any processing logic to recognize multiple masks in one key value pair.  The error when I got when I tried this  SYS:action_mask:LOAD_AIE|LOAD_PDI option was like "LOAD_PDI command not found" which makes sense as it considers '|' char as pipe symbol in linux command prompt. At the same time the new code does support multiple masks like LOAD_AIE and LOAD_PDI  together as this if-else logic iteratively sets the action mask for each key-value pair i,e. once for SYS:action_mask:LOAD_AIE and another time for SYS:action_mask:LOAD_PDI.

xfreid: what would the command line look like if user wants to specify both masks?

xvijaysri: Something like this "xclbinutil --key-value SYS:action_mask:LOAD_AIE SYS:action_mask:LOAD_PDI --output final.xclbin"

Output of --info on final.xclbin:

xclbin Information
------------------
   Generated by:           <unknown>
   Version:                2.18.0
   Kernels:                <unknown>
   Signature:              
   **Action Mask(s):**         LOAD_AIE LOAD_PDI 
   Content:                
   UUID (xclbin):          94b5e2a1-a0bc-bf32-cdb7-82b3fe9ddc2b
   Sections:

xfreid (approved):

stsoe (approved):
LGTM, but I am not qualified to review the semantics of the change.

PR #8355, opened 2024-08-16
elfloader changes for shim patching for aie4

HimanshuChoudhary-Xilinx:
Host address patching on shimBD for aie4.

aie4 have different shimbd configuration, Patching is done on 0th and 1st word of BD.

added new patching schema, shim_dma_aie4_base_addr_symbol_kind

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -45,7 +45,6 @@ namespace
 static constexpr size_t column_page_size = AIE_COLUMN_PAGE_SIZE;
 static constexpr uint8_t Elf_Amd_Aie2p  = 69;
 static constexpr uint8_t Elf_Amd_Aie2ps = 64;
-static constexpr uint8_t Elf_Amd_npu3   = 65;
 
 static const char* Scratch_Pad_Mem_Symbol = "scratch-pad-mem";
 static const char* Control_Packet_Symbol = "control-packet";
@@ -116,6 +115,7 @@ struct patcher
     scalar_32bit_kind = 3,
     control_packet_48 = 4,              // patching scheme needed by firmware to patch control packet
     shim_dma_48 = 5,                    // patching scheme needed by firmware to patch instruction buffer
+    shim_dma_aie4_base_addr_symbol_kind = 6, // patching scheme needed by AIE4 firmware
     unknown_symbol_kind = 8
   };
 
@@ -140,7 +140,6 @@ struct patcher
 
   buf_type m_buf_type = buf_type::ctrltext;
   symbol_type m_symbol_type = symbol_type::shim_dma_48;
-  uint8_t m_os_abi = Elf_Amd_Aie2p;
 
   struct patch_info {
     uint64_t offset_to_patch_buffer;
@@ -150,9 +149,8 @@ struct patcher
 
   std::vector<patch_info> m_ctrlcode_patchinfo;
 
-  patcher(uint8_t os_abi, symbol_type type, std::vector<patch_info> ctrlcode_offset, buf_type t)
-    : m_os_abi(os_abi)
-    , m_buf_type(t)
+  patcher(symbol_type type, std::vector<patch_info> ctrlcode_offset, buf_type t)
+    : m_buf_type(t)
     , m_symbol_type(type)
     , m_ctrlcode_patchinfo(std::move(ctrlcode_offset))
   {}
@@ -186,7 +184,7 @@ struct patcher
   }
 
   void
-  patch57_npu3(uint32_t* bd_data_ptr, uint64_t patch)
+  patch57_aie4(uint32_t* bd_data_ptr, uint64_t patch)
   {
     uint64_t base_address =
       ((static_cast<uint64_t>(bd_data_ptr[0]) & 0x1FFFFFF) << 32) |                   // NOLINT
@@ -239,10 +237,11 @@ struct patcher
         break;
       case symbol_type::shim_dma_base_addr_symbol_kind:
         // new_value is a bo address
-        if (m_os_abi == Elf_Amd_npu3)
-          patch57_npu3(bd_data_ptr, new_value + item.offset_to_base_bo_addr);
-        else
-          patch57(bd_data_ptr, new_value + item.offset_to_base_bo_addr);
+        patch57(bd_data_ptr, new_value + item.offset_to_base_bo_addr);
+        break;
+      case symbol_type::shim_dma_aie4_base_addr_symbol_kind:
+        // new_value is a bo address
+        patch57_aie4(bd_data_ptr, new_value + item.offset_to_base_bo_addr);
         break;
       case symbol_type::control_packet_48:
         // new_value is a bo address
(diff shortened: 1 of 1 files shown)


stsoe (changes requested):
Please fill out the PR description.

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -796,7 +814,7 @@ class module_elf : public module_impl
   ert_cmd_opcode
   get_ert_opcode() const override
   {
-    if (m_os_abi == Elf_Amd_Aie2ps)
+    if (m_os_abi == Elf_Amd_Aie2ps || m_os_abi == Elf_Amd_npu3)
```
xdavidz, line 817: I am not an expert of this area of the code, but it looks to me that there are lots of code like
"if aie2ps or aie4(npu3) then ..." can we make it more clear if you don't want to use more comment to explain what we are doing here.
1) can we use aie2ps and aie4 instead of aie2ps and npu3, those seems not a pair in same level
2) can we use a inline function or macro to make those dup code less?

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -223,7 +238,10 @@ struct patcher
         break;
       case symbol_type::shim_dma_base_addr_symbol_kind:
         // new_value is a bo address
-        patch57(bd_data_ptr, new_value + item.offset_to_base_bo_addr);
+        if (m_os_abi == Elf_Amd_npu3)
+          patch57_npu3(bd_data_ptr, new_value + item.offset_to_base_bo_addr);
+        else
+          patch57(bd_data_ptr, new_value + item.offset_to_base_bo_addr);
```
larry9523, line 244: This OK. But I think differentiating this by a new symbol type is cleaner than os_abi. Conceptly, one symobl type should map to one patching schema

dezhiAmd: I think @HimanshuChoudhary-Xilinx is trying to avoid adding more patching schema because currently we only use 3bits  for encoding schema-type.

larry9523: > I think @HimanshuChoudhary-Xilinx is trying to avoid adding more patching schema because currently we only use 3bits for encoding schema-type.

not 4 bits (0-3)?

dezhiAmd:
> Please fill out the PR description.

I keep saying this in almost all PRs from the link below:
https://gitenterprise.xilinx.com/XRT/aiebu/pulls?q=is%3Apr+author%3Achoudhar+

Please, please fill out PR description especially
1) Problem solved by this commit
2) How problem is solved

src/runtime_src/core/common/api/xrt_module.cpp
HimanshuChoudhary-Xilinx: its not a different patching... shimbd in aie2ps and npu3 have different configuration
in aie2ps, host address get patched on 1st,2nd, 8th word of BD while in npu3 its on 0th and 1st word of BD.
it the same 57bit patching

larry9523: > its not a different patching... shimbd in aie2ps and npu3 have different configuration in aie2ps, host address get patched on 1st,2nd, 8th word of BD while in npu3 its on 0th and 1st word of BD. its the same 57bit patching

I would argue shim_dma_48 is not a different patching either. It just patch to second and third of 8 BD WORDs.

OS_ABI is used to determine ELF's architecture. What ELF spec we should follow.

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -42,6 +42,7 @@ namespace
 static constexpr size_t column_page_size = AIE_COLUMN_PAGE_SIZE;
 static constexpr uint8_t Elf_Amd_Aie2p  = 69;
 static constexpr uint8_t Elf_Amd_Aie2ps = 64;
+static constexpr uint8_t Elf_Amd_npu3   = 65;
```
dezhiAmd, line 45: There is no good rationale to add a new os_abi since both aie2ps and npu3 are using 57 bits address internally. But we might consider renaming Elf_Amd_Aie2ps so it can include npu3

dezhiAmd (changes requested):

src/runtime_src/core/common/api/xrt_module.cpp
xuhz: do we have enough symbol type to hold all potential patching method?
if so, then we need something like,
patch57_aie2ps
patch57_aie4
if not, we may need combination of symbol and os_abi, like what Himanshu did.

dezhiAmd: Yes. We do have enough symbol type to hold all potential patching method for the moment.

HimanshuChoudhary-Xilinx: A Open question, if os_abi is about ELF spec, then how we differentiate between a elf with aie2ps controlcode and elf with aie4 controlcode because an elf with aie2ps controlcode will not work on aie4 device or vice-versa.
XRT should have a check on this.

larry9523: > A Open question, if os_abi is about ELF spec, then how we differentiate between a elf with aie2ps controlcode and elf with aie4 controlcode because an elf with aie2ps controlcode will not work on aie4 device or vice-versa. XRT should have a check on this.

I am totally fine if we define a different os_abi between aie2ps and aie4 if we need to. I am just saying we should use a new patching schema at this place instead of using os_abi.

HimanshuChoudhary-Xilinx:
added new patching schema

xuhz (commented):
i think the os_abi is still required, but not to be used to choose patching method
e.g to check an elf compiled with -t aie4 loading to aie4 hw

HimanshuChoudhary-Xilinx:
> i think the os_abi is still required, but not to be used to choose patching method e.g to check an elf compiled with -t aie4 loading to aie4 hw

we dont need os_abi because as @larry9523 said it represent elf spec and elf spec for aie2ps and aie4 is same

HimanshuChoudhary-Xilinx:
> > i think the os_abi is still required, but not to be used to choose patching method e.g to check an elf compiled with -t aie4 loading to aie4 hw
> 
> Let's clean this up by using e_machine

lets clean that in new pr as it will need aiebu changes also

xuhz (approved):
i didn't read the previous comments carefully
we have e_machine to tell aie2ps or aie4
os_abi is not for that purpose. 
but elf_amd_aie2ps is not a good name which caused the confusion. It should be changed in future PR with a better name, something like elf_amd_aie_uc

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -115,6 +115,7 @@ struct patcher
     scalar_32bit_kind = 3,
     control_packet_48 = 4,              // patching scheme needed by firmware to patch control packet
     shim_dma_48 = 5,                    // patching scheme needed by firmware to patch instruction buffer
+    shim_dma_aie4_base_addr_symbol_kind = 6, // patching scheme needed by AIE4 firmware
```
larry9523, line 118: Please update the ELF spce accordingly.

larry9523 (approved):

dezhiAmd (approved):

maxzhen (approved):

xdavidz (approved):

PR #8362, opened 2024-08-22
Update HIP Module Load APIs

rbramand-xilinx:
Updated hipModuleLoadData API according to spec mentioned in page - https://confluence.amd.com/display/AIE/HIP+Architecture#HIPArchitecture-Elfbasedflow
Earlier in HIP there was only support for loading xclbin and elf using file path passed as string. This change extends the load of xclbin/elf using buffer also.

Also added support for latest HIP release (6.1) in windows

This was ask from IREE team to load xclbin and elf data using buffer.

hipModuleLoadData argument void* image is used to pass object to structure hipModuleData which has the required fields to take xclbin/elf data as buffer along with its size. New header file xrt_hip.h is created and shipped with xrt package that has above mentioned structure. This header file can also be used in future to have more structures and APIs according to usecase.

Low

Tested build of XRT along with hip library on both Linux and Windows
Tested HIP Elf flow test on NPU

Updated confluence page and added example to demonstrate the use case. XRT doc update is NA.

### DIFF
--- a/src/runtime_src/hip/api/hip_device.cpp
+++ b/src/runtime_src/hip/api/hip_device.cpp
@@ -207,8 +207,29 @@ hipDeviceGetName(char* name, int len, hipDevice_t device)
   return hipErrorUnknown;
 }
 
+#if HIP_VERSION >= 60000000
+#undef hipGetDeviceProperties
+
 hipError_t
 hipGetDeviceProperties(hipDeviceProp_t* props, hipDevice_t device)
+{
+  return hipErrorNotSupported;
+}
+
+hipError_t
+hipGetDevicePropertiesR0600(hipDeviceProp_tR0600* props, int device)
+#else
+#define hipDeviceProp_tR0600 hipDeviceProp_t
+
+hipError_t
+hipGetDevicePropertiesR0600(hipDeviceProp_tR0600* props, int device)
+{
+  return hipErrorNotSupported;
+}
+
+hipError_t
+hipGetDeviceProperties(hipDeviceProp_t* props, hipDevice_t device)
+#endif
 {
   try {
     throw_invalid_value_if(!props, "arg passed is nullptr");
--- a/src/runtime_src/hip/api/hip_device.cpp
+++ b/src/runtime_src/hip/api/hip_device.cpp
@@ -219,7 +219,7 @@ hipGetDeviceProperties(hipDeviceProp_t* props, hipDevice_t device)
 hipError_t
 hipGetDevicePropertiesR0600(hipDeviceProp_tR0600* props, int device)
 #else
-#define hipDeviceProp_tR0600 hipDeviceProp_t
+using hipDeviceProp_tR0600 = hipDeviceProp_t;
 
 hipError_t
 hipGetDevicePropertiesR0600(hipDeviceProp_tR0600* props, int device)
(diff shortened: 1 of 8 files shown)


src/runtime_src/hip/core/module.cpp
```diff
@@ -24,6 +26,15 @@ module_xclbin(std::shared_ptr<context> ctx, const std::string& file_name)
   create_hw_context();
 }
 
+module_xclbin::
+module_xclbin(std::shared_ptr<context> ctx, void* data, size_t size)
+  : module{std::move(ctx), true}
+{
+  std::vector<char> buf(static_cast<char*>(data), static_cast<char*>(data) + size);
+  m_xrt_xclbin = xrt::xclbin{buf};
+  create_hw_context();
```
stsoe, line 35: Prefer initializer list, avoid ctor body.
Initialize the data members in the order needed, something like this:
```
: module(...)
, m_xrt_xclbin{create_xclbin(data, size)}
, m_hw_context{get_context()->get_xrt_device, m_xrt_xclbin.get_uuid()}
{}
```
where create_xclbin() is a static free namespaced function.

src/runtime_src/hip/core/module.cpp
```diff
@@ -33,6 +44,18 @@ module_elf(module_xclbin* xclbin_module, const std::string& file_name)
...
+module_elf::
+module_elf(module_xclbin* xclbin_module, void* data, size_t size)
+  : module{xclbin_module->get_context(), false}
+  , m_xclbin_module{xclbin_module}
+{
+  std::istringstream stream;
+  stream.rdbuf()->pubsetbuf(static_cast<char*>(data), size);
+
+  m_xrt_elf = xrt::elf{stream};
+  m_xrt_module = xrt::module{m_xrt_elf};
```
stsoe, line 56: ditto, no ctor body, use static helper functions as necessary. Why?  Because the stream created here is for the purpose of ctoring the elf only, it has nothing to do with m_xrt_module.

src/runtime_src/hip/xrt_hip.h
```diff
@@ -0,0 +1,25 @@
...
+// Copyright (C) 2024 Advanced Micro Devices, Inc. All rights reserved.
+#ifndef xrthip_h
+#define xrthip_h
+
+#include "hip/hip_runtime_api.h"
+
+enum hipModuleDataType {
+  hipModuleDataFilePath = 0,
+  hipModuleDataBuffer
+};
```
stsoe, line 11: class enum?

src/runtime_src/hip/api/hip_device.cpp
```diff
@@ -207,8 +207,29 @@ hipDeviceGetName(char* name, int len, hipDevice_t device)
...
+hipError_t
+hipGetDeviceProperties(hipDeviceProp_t* props, hipDevice_t device)
+{
+  return hipErrorNotSupported;
+}
+
+hipError_t
+hipGetDevicePropertiesR0600(hipDeviceProp_tR0600* props, int device)
+#else
+#define hipDeviceProp_tR0600 hipDeviceProp_t
```
keryell, line 222: Why using a macro instead of `using` in a C++ program?

src/runtime_src/hip/xrt_hip.h
rbramand-xilinx: this header file is shipped to compile with HIP applications and all the other HIP headers are C based, so using C style for the headers that we ship. Also IREE wants to compile their applications using C compiler

stsoe (approved):

chvamshi-xilinx (approved):

PR #8373, opened 2024-08-27
CR-1185333: Update pts data type to 64-bit signed integer

sureshreddyavula:
This change fixes the PTS overflow issue if the bitstream has 64-bit PTS value.

User reported video and audio getting out of sync after transcoding a bitstream that has >32-bit PTS. The application stack has 64-bit PTS and the XMA has 32-bit PTS value.

Made PTS parameter as 64-bit in XmaDataBuffer.

None.

It has been tested with the bitstream where the sync issue happened and multiple other transcoder commands.


sureshreddyavula:
@chvamshi-xilinx Can you please review this PR

PR #8375, opened 2024-08-27
CR-1191227: Release xma session resources in session destroy

sureshreddyavula:
Released all the xma session resources in session destroy.

Observed small memory leak when we create and destroy session multiple times without exiting the application/process. In a streaming applications, the process will be live always. During an incoming stream, session creation will be done and session destroy will be called at the end of stream. If we run create and destroy session in a loop, memory gradually increases.

During session destroy all the resources(including init time) are freed.

Tested with sample application looping session create and destroy for multiple hours, didn't observed memory increase.


sureshreddyavula:
@chvamshi-xilinx Can you please review this PR

PR #8378, opened 2024-08-28
add myself as code owner for xclbinutil

xfreid:
Add myself as code owner for xclbinutil

None

N/A

None

None

None

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -37,6 +37,7 @@ src/runtime_src/xocl/api/plugin/xdp/ @jvillarre
 src/runtime_src/xdp/ @jvillarre
 src/runtime_src/doc/ @uday610
 src/runtime_src/tools/ @rozumx
+src/runtime_src/tools/xclbinutil/ @feirui
 src/runtime_src/tools/scripts/ @rozumx @stsoe
 src/runtime_src/hip/ @sonals @chvamshi-xilinx @stsoe
 src/CMake/ @rozumx @stsoe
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -37,6 +37,7 @@ src/runtime_src/xocl/api/plugin/xdp/ @jvillarre
 src/runtime_src/xdp/ @jvillarre
 src/runtime_src/doc/ @uday610
 src/runtime_src/tools/ @rozumx
+src/runtime_src/tools/xclbinutil/ @feirui
```
chvamshi-xilinx, line 40: Hi Fei , i see your username is @xfreid .  you added @feirui here. i am not sure whether this works or not.

stsoe: This will not work.  See the underlined error.

stsoe (changes requested):

.github/CODEOWNERS
xfreid: My bad, just fixed it.

stsoe (changes requested):
When you specialize a subdirectory you change the rule to what you write for that subdir.  E.g. previously src/runtime_src/tools/xclbinutil/ was covered by src/runtime_src/tools/  which had @rozumx as a reviewer.  Now with your addition xclbintuil/ is speciailized and only notifies you.  You probably should add @rozumx as well.  See how tools/scripts/ is specialized .

stsoe (approved):

PR #8379, opened 2024-08-28
External Buffer support for Edge devices

chvamshi-xilinx:
 XRT doesnt have external buffer support. Enabled it with this commit.

This is a new feature.

Plugged the external buffers also into existing sync and async bo APIs. 

None as this new feature.

Verified single/multi external buffers with blocking/non-blocking cases.

Yes, We need to add documentation of external buffer support

### DIFF
--- a/src/runtime_src/core/edge/user/aie/common_layer/adf_runtime_api.cpp
+++ b/src/runtime_src/core/edge/user/aie/common_layer/adf_runtime_api.cpp
@@ -873,6 +873,19 @@ err_code dma_api::waitDMAChannelDone(int tileType, uint8_t column, uint8_t row, 
     return err_code::ok;
 }
 
+err_code dma_api::updateBDAddress(int tileType, uint8_t column, uint8_t row, uint8_t bdId, uint64_t address)
+{
+  int driverStatus = XAIE_OK; //0
+  XAie_LocType tileLoc = XAie_TileLoc(column, relativeToAbsoluteRow(tileType, row));
+
+  driverStatus |= XAie_DmaUpdateBdAddr(config_manager::s_pDevInst, tileLoc, address, bdId);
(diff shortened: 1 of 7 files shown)


src/runtime_src/core/edge/user/aie/common_layer/adf_runtime_api.cpp
```diff
@@ -873,6 +873,19 @@ err_code dma_api::waitDMAChannelDone(int tileType, uint8_t column, uint8_t row,
     return err_code::ok;
 }
 
+err_code dma_api::updateBDAddress(int tileType, uint8_t column, uint8_t row, uint8_t bdId, uint64_t address)
```
chvamshi-xilinx, line 876: This code is a copy of aiecompiler code. Thatswhy didnt use XRT naming convention here,

larry9523 (approved):
The changes look good.

One minor question. Is there an implicit API change to xrt::aie::bo.sync()? We documented the interace, "port", as GMIO port name https://github.com/Xilinx/XRT/blob/master/src/runtime_src/core/include/xrt/xrt_aie.h#L243. 
Is it still valid for external buffer? Or we should update it to be either GMIO port name or external buffer config name?

chvamshi-xilinx:
> The changes look good.
> 
> One minor question. Is there an implicit API change to xrt::aie::bo.sync()? We documented the interace, "port", as GMIO port name https://github.com/Xilinx/XRT/blob/master/src/runtime_src/core/include/xrt/xrt_aie.h#L243. Is it still valid for external buffer? Or we should update it to be either GMIO port name or external buffer config name?

Thanks @larry9523  for the quick review. I am going to change the names and documentation everywhere in my next PR.
There is no change in the API for now, but we may need to change the API interface for the ping pong buffer support which will come later in this release. I will come up with new interface on how to support ping pong buffer.

PR #8380, opened 2024-08-28
[XDP] Fix the incorrect xclbin type deduction & add safe check

vipangul:
CR-1208980 : PLIO designs mentioned in the CR are running into segfaults.

code flow incorrectly deduce the xclbin type is of AIE Only type and doesn't create PLIO interface.

To correctly check if xclbin type specially PL part of xclbin , now we rely on both `IP_LAYOUT` and `DEBUG_IP_LAYOUT`.
These both sections are not packaged in AIE only xclbin.

minimal

Testing pending on vck190. Local changes are not being packaged along with latest petalinux image.
Update:
Verified on VCK190 by falling back to older label and petalinux image. Confirmed that  mentioned crash is not seen & trace is generated.
However, we do see lof messages emitted from driver like below during execution which needs to be triaged separately.
`xrtSyncBOAIENB: name gr.gm_out0, bo 0xaaaaaab3da40, XCL_BO_SYNC_BO_AIE_TO_GMIO, transaction size 128, offset 262016`

### DIFF
--- a/src/runtime_src/xdp/profile/plugin/aie_trace/aie_trace_plugin.cpp
+++ b/src/runtime_src/xdp/profile/plugin/aie_trace/aie_trace_plugin.cpp
@@ -238,7 +238,10 @@ void AieTracePluginUnified::updateAIEDevice(void *handle) {
 
   // First, check against memory bank size
   // NOTE: Check first buffer for PLIO; assume bank 0 for GMIO
-  uint8_t memIndex = (isPLIO && (deviceIntf != nullptr))? deviceIntf->getAIETs2mmMemIndex(0) : 0;
+  uint8_t memIndex = 0;
+  if (isPLIO && (deviceIntf != nullptr))
+   memIndex = deviceIntf->getAIETs2mmMemIndex(0);
+
   Memory *memory = (db->getStaticInfo()).getMemory(deviceID, memIndex);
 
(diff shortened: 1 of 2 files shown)


jvillarre (approved):

src/runtime_src/xdp/profile/plugin/aie_trace/aie_trace_plugin.cpp
```diff
@@ -238,7 +238,7 @@ void AieTracePluginUnified::updateAIEDevice(void *handle) {
 
   // First, check against memory bank size
   // NOTE: Check first buffer for PLIO; assume bank 0 for GMIO
-  uint8_t memIndex = isPLIO ? deviceIntf->getAIETs2mmMemIndex(0) : 0;
+  uint8_t memIndex = (isPLIO && (deviceIntf != nullptr))? deviceIntf->getAIETs2mmMemIndex(0) : 0;
```
stsoe, line 241: This is difficult to debug with a break point.  I recommend ternaries that call functions to be split as
```
(cond) 
? true_path()
: false_path();

```

vipangul: Let me fix this.

PR #8381, opened 2024-08-29
XRT OS Support for Ubuntu 24.04. Removed one element arrays in the driver code

ManojTakasi:
https://jira.xilinx.com/browse/VITIS-12053 XRT OS Support for Ubuntu 24.04

This is an intial PR to move to flexible arrays instead of one length array in structures. Reference: https://www.kernel.org/doc/html/latest/process/deprecated.html#zero-length-and-one-element-arrays

Removed mmap xocl_gem_prime_mmap function as drm_gem_mmap() function  is capible of setting flags before mmap.

Removed the xocl_gem_prime_mmap and changed some structures into flexible arrays as one-legth arrays are throwing call trace while using memcpy() function.

low

Tested with u250 in ubuntu 22.04 and 24.04

Yes

### DIFF
--- a/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
+++ b/src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
@@ -674,7 +674,7 @@ int xocl_reclock(struct xocl_dev *xdev, void *data)
 	req = kzalloc(reqlen, GFP_KERNEL);
 	req->req = XCL_MAILBOX_REQ_RECLOCK;
 	memcpy(req->data, data, data_len);
- 
+
 	if (get_live_clients(xdev, NULL)) {
 		userpf_err(xdev, "device is in use, can't reset");
 		err = -EBUSY;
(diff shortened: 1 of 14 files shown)


chvamshi-xilinx (changes requested):
As discussed, please try to create  a macro instead of calling sizeof in lot of places.

ManojTakasi:
Hi Vamshi, as suggested used struct_size in all the places.

keryell:
I do not understand the reference to VLA here.
Where are the VLA in the code?

keryell:
Just to clarify, the "flexible array" is a coding convention, like VLA () are a C99 features made optional starting with C11
https://en.wikipedia.org/wiki/Variable-length_array
```c
int n = 42;
float a[n];
```
It is not in C++ because you cannot have `sizeof(a)` not being `constexpr`.

ManojTakasi:
Hi Keryell, Thanks for your clarification. Sorry for referencing wrong link. Please check this link: https://www.kernel.org/doc/html/latest/process/deprecated.html#zero-length-and-one-element-arrays

src/runtime_src/core/pcie/driver/linux/xocl/userpf/xocl_drv.c
```diff
@@ -327,7 +327,7 @@ int xocl_program_shell(struct xocl_dev *xdev, bool force)
 		goto failed;
 
 	userpf_info(xdev, "request mgmtpf to program prp");
-	mbret = xocl_peer_request(xdev, &mbreq, sizeof(struct xcl_mailbox_req),
+	mbret = xocl_peer_request(xdev, &mbreq, sizeof(mbreq) + sizeof(int32_t),
```
chvamshi-xilinx, line 330: struct_size here aswell?

chvamshi-xilinx (approved):

PR #8387, opened 2024-09-02
Improve compatibility with other Compilers like MinGW

Mellich:
I am working on an integration of [XRT into Julia](https://github.com/JuliaPackaging/Yggdrasil/pull/9290). However, the current Windows build does only support MSVC and the packages for Julia are build using MinGW, so I XRT to improve the compatibility with MinGW. I ported back some of the patches because they should not affect the currently supported toolchains and could be useful to support further toolchains in the future. No features are added or removed to XRT with this PR - it only affects the build process.

Currently, some preprocessor defines are used in non-standard ways, increasing the difficulty to compile the code on Windows with GCC. Also, some library dependencies seem to be missing for some of the Windows binaries.

 I changed some occurrences of `__GNU__` to `__linux__` because the contained codes rather depend on the used OS than on the compiler.
The missing libraries were added as link dependency.

I was not able to test the build on a native Windows machine with MSVC, because I have no access to one.
The proposed changes should only affect the build and linking of binaries, and not the functionality.

Build on Red Head Enterprise Linux and Alpine Linux using GCC 9.1 as well as MinGW on Alpine Linux with GCC 9.1 and some [additional minor patches](https://github.com/JuliaPackaging/Yggdrasil/pull/9290).

Should not affect documentation, because this PR only applies changes to the build process.

### DIFF
--- a/src/runtime_src/core/pcie/windows/alveo/CMakeLists.txt
+++ b/src/runtime_src/core/pcie/windows/alveo/CMakeLists.txt
@@ -36,6 +36,13 @@ target_link_libraries(xrt_core
   xrt_coreutil
   )
 
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  target_link_libraries(xrt_core
+    PRIVATE
+    setupapi
+  )
+endif()
+
 # For DLL platforms the DLL part of a shared library is treated as a
 # RUNTIME target and the corresponding import library is treated as an
 # ARCHIVE target. All Windows-based systems including Cygwin are DLL
--- a/src/runtime_src/core/tools/xbutil2/CMakeLists.txt
+++ b/src/runtime_src/core/tools/xbutil2/CMakeLists.txt
@@ -100,6 +100,10 @@ target_link_libraries(${XBUTIL2_NAME}
 
 if (NOT WIN32)
   target_link_libraries(${XBUTIL2_NAME} PRIVATE pthread uuid dl)
+else()
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    target_link_libraries(${XBUTIL2_NAME} PRIVATE ws2_32)
+  endif()
 endif()
 
 # Package xrt sub commands json file for embedded builds
--- a/src/runtime_src/tools/xclbinutil/CMakeLists.txt
+++ b/src/runtime_src/tools/xclbinutil/CMakeLists.txt
@@ -136,6 +136,11 @@ if(NOT WIN32)
    target_link_libraries(${XCLBINUTIL_NAME} PRIVATE transformcdo)
 endif()
 
+# Link required windows specific libraries with MinGW
+if(WIN32 AND (CMAKE_CXX_COMPILER_ID STREQUAL "GNU"))
+  target_link_libraries(${XCLBINUTIL_NAME} PRIVATE wsock32 ws2_32)
+endif()
+
 install (TARGETS ${XCLBINUTIL_NAME} RUNTIME DESTINATION ${XRT_INSTALL_UNWRAPPED_DIR})
 install (PROGRAMS ${XRT_LOADER_SCRIPTS} DESTINATION ${XRT_INSTALL_BIN_DIR})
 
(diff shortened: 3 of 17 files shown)


src/runtime_src/core/tools/xbutil2/CMakeLists.txt
```diff
@@ -100,6 +100,8 @@ target_link_libraries(${XBUTIL2_NAME}
 
 if (NOT WIN32)
   target_link_libraries(${XBUTIL2_NAME} PRIVATE pthread uuid dl)
+else()
+  target_link_libraries(${XBUTIL2_NAME} PRIVATE ws2_32)
```
stsoe, line 104: Why is  ws2_32 neeeded?

src/runtime_src/tools/xclbinutil/CMakeLists.txt
```diff
@@ -136,6 +136,11 @@ if(NOT WIN32)
    target_link_libraries(${XCLBINUTIL_NAME} PRIVATE transformcdo)
 endif()
 
+# Link required windows specific libraries
+if(WIN32)
+  target_link_libraries(${XCLBINUTIL_NAME} PRIVATE wsock32 ws2_32)
```
stsoe, line 141: Not sure why these link libraries are needed?

src/runtime_src/core/pcie/windows/alveo/CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ add_library(xrt_core_static STATIC
 target_link_libraries(xrt_core
   PRIVATE
   xrt_coreutil
+  setupapi
```
stsoe, line 37: This link library is included as a pragma already.

stsoe (changes requested):
In general this looks okay.  Few question around the added link libraries.
Are we certain that `__attribute__ ((...))` converted from `__GNUC__` to `__linux__` are in fact not GNU specific?

src/runtime_src/core/tools/xbutil2/CMakeLists.txt
Mellich: ws2_32 is included here and used by some of the binaries: https://github.com/Xilinx/XRT/blob/c34c5afb4c92cb68460349185da7e8fb24f9d957/src/runtime_src/core/tools/common/XBUtilities.cpp#L32

src/runtime_src/tools/xclbinutil/CMakeLists.txt
Mellich: winsock is included i.e. here and on other places in xclbinutil: https://github.com/Xilinx/XRT/blob/c34c5afb4c92cb68460349185da7e8fb24f9d957/src/runtime_src/tools/xclbinutil/DTC.cxx#L26
ws2_32 is included here and used by some of the binaries: https://github.com/Xilinx/XRT/blob/c34c5afb4c92cb68460349185da7e8fb24f9d957/src/runtime_src/core/tools/common/XBUtilities.cpp#L32

src/runtime_src/core/pcie/windows/alveo/CMakeLists.txt
Mellich: This is true. The pragmas are not used by MinGW, so I have to link them differently. However, these lines should depend on the used compiler. I updated the files accordingly to only use them if the compiler is GNU.

Mellich:
> Are we certain that `__attribute__ ((...))` converted from `__GNUC__` to `__linux__` are in fact not GNU specific?

At least for Clang, this seems also to be available: https://clang.llvm.org/docs/LTOVisibility.html, so it is not GNU specific.
Regarding the additional libraries: I made them depend on the compiler ID to prevent issues with MSVC. As said in another comment, these are required because linking via pragmas is not supported by the GNU compilers and the libraries should be provided via link time instead.

stsoe (changes requested):
This looks good.  @Mellich please signoff your commits.

@manikandan-xilinx, we will need to get CI to kick in and validate this PR, what's the trick to run CI for outside contributors?

stsoe (approved):

stsoe:
@Mellich .  Thank you very much for this effort.

PR #8388, opened 2024-09-03
Removing rhel9 and apu_package for 2021.1 as per request

manikandan-xilinx:
Removing rhel9 and apu_package for 2021.1 as per request and adding a master pipeline for 2021.1


manikandan-xilinx:
Merging it as it's a pipeline change needed for 2021.1 PR's

PR #8392, opened 2024-09-04
Segmented Config changes

chvamshi-xilinx:
We were relying on pr_isolation address to program the PDI earlier, With segmented config support, we need to relax that check. We will get information from new bitmap entry
None
Relaxed all the pr_isolation checks in zocl
None
Verified zynqmp/versal aie/non-aie cases
None

### DIFF
--- a/src/runtime_src/core/edge/user/shim.cpp
+++ b/src/runtime_src/core/edge/user/shim.cpp
@@ -684,12 +684,10 @@ xclLoadAxlf(const axlf *buffer)
   std::string dtbo_path("");
 
 #ifndef __HWEM__
-  auto is_pr_platform = (buffer->m_header.m_mode == XCLBIN_PR ) ? true : false;
+  auto is_pr_platform = (buffer->m_header.m_mode == XCLBIN_PR || buffer->m_header.m_actionMask & AM_LOAD_PDI) ? true : false;
   auto is_flat_enabled = xrt_core::config::get_enable_flat(); //default value is false
   auto force_program = xrt_core::config::get_force_program_xclbin(); //default value is false
   auto overlay_header = xclbin::get_axlf_section(buffer, axlf_section_kind::OVERLAY);
-  //force load PDI if below action mask is set
-  is_pr_platform = (buffer->m_header.m_actionMask & AM_LOAD_PDI) ? true : false;
 
   if (is_pr_platform)
     flags = DRM_ZOCL_PLATFORM_PR;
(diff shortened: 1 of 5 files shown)


src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -688,6 +688,8 @@ xclLoadAxlf(const axlf *buffer)
   auto is_flat_enabled = xrt_core::config::get_enable_flat(); //default value is false
   auto force_program = xrt_core::config::get_force_program_xclbin(); //default value is false
   auto overlay_header = xclbin::get_axlf_section(buffer, axlf_section_kind::OVERLAY);
+  //force load PDI if below action mask is set
+  is_pr_platform = (buffer->m_header.m_actionMask & AM_LOAD_PDI) ? true : false;
```
stsoe, line 692: huh?  Why not change line 687?

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -1181,6 +1183,8 @@ int shim::prepare_hw_axlf(const axlf *buffer, struct drm_zocl_axlf *axlf_obj)
...
```
stsoe, line 1187: ditto?

stsoe (changes requested):

src/runtime_src/core/edge/user/shim.cpp
chvamshi-xilinx: Thanks @stsoe . Fixed it. It was a bug indeed :-(

src/runtime_src/core/edge/user/shim.cpp
chvamshi-xilinx: Fixed.

src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -684,7 +684,7 @@ xclLoadAxlf(const axlf *buffer)
   std::string dtbo_path("");
 
 #ifndef __HWEM__
-  auto is_pr_platform = (buffer->m_header.m_mode == XCLBIN_PR ) ? true : false;
+  auto is_pr_platform = (buffer->m_header.m_mode == XCLBIN_PR || buffer->m_header.m_actionMask & AM_LOAD_PDI) ? true : false;
```
stsoe, line 687: Kind of strange to do `if (true) true; else false`;  :-)  Maybe drop the ternary?

chvamshi-xilinx: yeah.. Fixed it Soren. Thank you.

PR #8394, opened 2024-09-04
VITIS-13337 Add pmode turbo to xrt-smi

AShivangi:
[VITIS-13337](https://jira.xilinx.com/browse/VITIS-13337) Add pmode turbo to xrt-smi

N/A

New feature

N/A

Windows/Strix setup (driver changes yet to come)

N/A


src/runtime_src/core/common/query_requests.h
```diff
@@ -3685,9 +3685,10 @@ struct performance_mode : request
   enum class power_type
   {
     basic, // deafult
```
sonals, line 3687: Since we want to use the enum value as integer scalar, we should explicitly initialize them to integer values:
`basic = 0`

src/runtime_src/core/common/query_requests.h
```diff
@@ -3707,11 +3708,13 @@ struct performance_mode : request
       case 0:
```
sonals, line 3708: Please use all lowercase as we are treating them as keywords.

AShivangi: The parse_status function is like a to_string_pretty, we print this out directly in reports so we need the first letter capitalized

PR #8396, opened 2024-09-05
VITIS-11710 AIE Utilization

rchane:
https://jira.xilinx.com/browse/VITIS-11710
N/A
Added AIE column utilization to the aie-partitions report. This change requires Divyajyoti's previously merged changes from MCDM. An additional PR will be opened in MCDM once this PR is merged to update the xrt submodule there and add the getter function.
N/A
Tested on STX using df-bw test (with the MCDM changes).
```
C:\Users\rchane\Desktop\xrt>xrt-smi examine -r aie-partitions

---------------------
[00c5:00:01.1] : NPU
---------------------
AIE Partitions
Total Column Utilization: 100%
  Partition Index: 0
    Columns: [0, 1, 2, 3]
    HW Contexts:
      |Context ID  ||Submissions  ||Completions  ||Migrations  ||Errors  |
      |------------||-------------||-------------||------------||--------|
      |9           ||133          ||132          ||0           ||0       |

  Partition Index: 1
    Columns: [4, 5, 6, 7]
    HW Contexts:
      |Context ID  ||Submissions  ||Completions  ||Migrations  ||Errors  |
      |------------||-------------||-------------||------------||--------|
      |10          ||34           ||33           ||0           ||0       |

C:\Users\rchane\Desktop\xrt>xrt-smi examine -r aie-partitions -o tmp.txt --force

---------------------
[00c5:00:01.1] : NPU
---------------------
AIE Partitions
Total Column Utilization: 50%
  Partition Index: 0
    Columns: [0, 1, 2, 3]
    HW Contexts:
      |Context ID  ||Submissions  ||Completions  ||Migrations  ||Errors  |
      |------------||-------------||-------------||------------||--------|
      |12          ||161          ||160          ||0           ||0       |

```
Sample JSON output:
```
{
    "schema_version": {
        "schema": "JSON",
        "creation_date": "Thu Sep  5 18:38:31 2024 GMT"
    },
    "devices": [
        {
            "interface_type": "pcie",
            "device_id": "00c5:00:01.1",
            "aie_partitions": {
                "description": "AIE Partition Information",
                "partitions": [
                    {
                        "start_col": "0",
                        "num_cols": "4",
                        "partition_index": "0",
                        "hw_contexts": [
                            {
                                "context_id": "12",
                                "command_submissions": "161",
                                "command_completions": "160",
                                "migrations": "0",
                                "errors": "0"
                            }
                        ]
                    }
                ],
                "total_col_utilization": "50%"
            }
        }
    ]
}
```
N/A


stsoe (approved):

src/runtime_src/core/tools/common/reports/ReportAiePartitions.cpp
```diff
@@ -61,6 +61,41 @@ populate_aie_partition(const xrt_core::device* device)
...
 
+std::string
+calculate_col_utilization(const xrt_core::device* device, boost::property_tree::ptree &_pt)
+{
+  xrt_core::query::total_cols::result_type total_device_cols;
+  try {
+    total_device_cols = xrt_core::device_query_default<xrt_core::query::total_cols>(device, 0);
+  }
+  catch (...) {
+    return "N/A"; //Error getting total device cols
```
AShivangi, line 72: You don't need to do try/catch with device_query_default, check the function definition
```
auto total_device_cols = xrt_core::device_query_default<xrt_core::query::total_cols>(device, 0);
```
This is sufficient

src/runtime_src/core/tools/common/reports/ReportAiePartitions.cpp
```diff
@@ -79,6 +114,7 @@ getPropertyTree20202(const xrt_core::device* _pDevice,
   boost::property_tree::ptree pt;
   pt.put("description", "AIE Partition Information");
   pt.add_child("partitions", populate_aie_partition(_pDevice));
+  pt.put("total_col_utilization", calculate_col_utilization(_pDevice, pt));
```
AShivangi, line 117: Can you move this under description so it shows up before the partition info in the json?

src/runtime_src/core/tools/common/reports/ReportAiePartitions.cpp
```diff
@@ -97,6 +133,12 @@ writeReport(const xrt_core::device* /*_pDevice*/,
     return;
   }
 
+  const boost::property_tree::ptree pt_info = _pt.get_child("aie_partitions", empty_ptree);
+  if (pt_info.empty())
+    _output << "Total Column Utilization: N/A\n";
+  else
+    _output << boost::str(boost::format("Total Column Utilization: %s\n") % pt_info.get<std::string>("total_col_utilization"));
```
AShivangi, line 140: Why don't you place N/A or an actual value in calculate_col_utilization()?
In L96 you can do something like:
`return (total_col_utilization==0) ? "N/A" : boost::str(boost::format("%.0f%%") % total_col_utilization);`

AShivangi (changes requested):

src/runtime_src/core/tools/common/reports/ReportAiePartitions.cpp
rchane: I agree that it would look better if it showed up beforehand; the main reason I put it after, however, is so that all the partition info is properly organized after the call to populate_aie_partition and I can cleanly use the separate calculate_col_utilization function. I think one possible alternative with minimal change could be to temporarily store the ptree result of populate_aie_partition before adding it to `pt` after calculating utilization? Please let me know what you think!

src/runtime_src/core/tools/common/reports/ReportAiePartitions.cpp
rchane: I simplified this if suite to `_output << boost::str(boost::format("Total Column Utilization: %s\n") % _pt.get<std::string>("aie_partitions.total_col_utilization"));`, as total_col_utilization should always be there with "N/A" or a "%" string (I originally had that suite like that because I was imitating the empty ptree checks for partition data, my mistake!). Do we want to show "N/A" for "0%" utilization, though? This would only be if there are HW contexts on the device that are somehow all idle.

src/runtime_src/core/tools/common/reports/ReportAiePartitions.cpp
```diff
@@ -61,6 +61,41 @@ populate_aie_partition(const xrt_core::device* device)
   return pt;
 }
 
+std::string
```
AShivangi, line 64: static

AShivangi (approved):

PR #8305, opened 2024-07-24
Added hw ctx support for edge

bisingha-xilinx:
#### Problem solved by the commit
create/destroy hw_context enablement for edge
#### Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered
Fixed issues: register xclbin support, hw ctx support
#### How problem was solved, alternative solutions (if any) and why they were rejected
Programming the xclbin while creating the hw ctx and added zocl driver support for create/destroy hw_context
#### Risks (if any) associated the changes in the commit
n/a
#### What has been tested and how, request additional testing if necessary
Please find the table below. Tested on hw emu with vck190 xpfm with 2024.2 TA
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>

<div style='direction:ltr'>

testname | legacy | hwctx
-- | -- | --
complex_graph | passed | passed
independent_graph | passed | passed
simple_gmio | passed | passed
simple_graph | passed | passed
simple_rtp | passed | passed
simple_vadd(opencl) | passed | testcase is n/a
simple_vadd(xrt   APIs) | passed | passed
zynqmp_HelloWorld(opencl) | passed | testcase is n/a

</div>

</body>

</html>

#### Documentation impact (if any)
Please note: by default, hw context is enabled on the edge shim. hw_context=false has to be provided on xrt.ini if someone wants to go with the legacy (non hwctx) flow.

PR #8402, opened 2024-09-06
Revert Edge hw context creation hook to XDP plugin

vipangul:
Recent  hw context support on Edge had also enable all XDP plugins via https://github.com/Xilinx/XRT/pull/8305 

This resulted in calling of plugins twice once during xclbin load and later during hw_context creation. This further resulted in driver related errors because of unavailability of resources.

This will need to be enabled hw_context and disable from xclbin load if needed.

https://github.com/Xilinx/XRT/pull/8305 

Removed profiling wrapper from newly added hw_context as we already have for xclbin load.

Low

Verified on edge to see no double load of XDP plugins.
No more driver related issues seen.
However, plugins are still running into separate issue i.e. CR-1209732

### DIFF
--- a/src/runtime_src/core/edge/user/shim.cpp
+++ b/src/runtime_src/core/edge/user/shim.cpp
@@ -1312,10 +1312,6 @@ int shim::load_hw_axlf(xclDeviceHandle handle, const xclBin *buffer, drm_zocl_cr
     drv->registerAieArray();
   #endif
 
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx:
Hi @vipangul ,
This is already fixed as part of https://github.com/Xilinx/XRT/pull/8400
Please check whether this change is still required.

PR #8405, opened 2024-09-09
CR-1211324 Hw_Emulation failing with XRT error for design with event …

SravanKumarAllu-xilinx:
…api usage for latency check

[CR-1211324] Hw_Emulation failing with XRT error for design with event api usage for latency check

Profile handle related changes

1) Now using shared_ptr instead of unique_ptr and passing shared_ptr across to avoid using raw pointer
2) Duplicate "stop profiling" is getting called from "profile_impl" class destructor, removed such call now.

N/A

Profile hw-emu use case

N/A

### DIFF
--- a/src/runtime_src/core/common/api/aie/xrt_graph.cpp
+++ b/src/runtime_src/core/common/api/aie/xrt_graph.cpp
@@ -126,6 +126,7 @@ class profiling_impl
 {
 private:
   std::unique_ptr<xrt_core::profile_handle> m_profile_handle{nullptr};
+  bool is_stopped{false};
 
 public:
   static constexpr int invalid_handle = -1;
@@ -144,7 +145,6 @@ public:
   {
     stop();
   }
-
   profiling_impl() = delete;
   profiling_impl(const profiling_impl&) = delete;
   profiling_impl(profiling_impl&&) = delete;
@@ -166,7 +166,10 @@ public:
   void
   stop()
   {
-    return m_profile_handle->stop();
+    if(is_stopped == false)
+      m_profile_handle->stop();
+
+    is_stopped = true;
   }
 };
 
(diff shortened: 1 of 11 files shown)


src/runtime_src/core/common/api/aie/xrt_graph.cpp
```diff
@@ -140,11 +140,6 @@ class profiling_impl
     m_profile_handle = hwctx_handle->open_profile_handle();
   }
 
-  ~profiling_impl()
-  {
-    stop();
```
chvamshi-xilinx, line 145: You still require this, but add it under a check if stop is called early , then call stop.

src/runtime_src/core/edge/user/aie/graph.cpp
```diff
@@ -110,7 +110,7 @@ xclGMIOWait(xclDeviceHandle handle, const char *gmioName)
 
   if (!drv->isAieRegistered())
     throw xrt_core::error(-EINVAL, "No AIE presented");
-  auto aieArray = drv->getAieArray();
+  auto aieArray = drv->get_aie_array_shared();
```
chvamshi-xilinx, line 113: We are mixing Camel case and "_" seperated names in lot of places, Can you please fix all of them in next PR.
I know this is not from your change, but just to be consistent.

src/runtime_src/core/edge/user/hwctx_object.h
```diff
@@ -85,7 +85,7 @@ namespace zynqaie {
     open_profile_handle() override;
 
 #ifdef XRT_ENABLE_AIE
-    Aie*
+    std::shared_ptr<Aie>
     get_aie_array_from_hwctx();
```
chvamshi-xilinx, line 89: rename this to get_aie_array as this is a member function of hw_context

chvamshi-xilinx (changes requested):

src/runtime_src/core/edge/user/aie/graph.cpp
SravanKumarAllu-xilinx: sure, I could cleanup in next PR

chvamshi-xilinx (approved):

PR #8407, opened 2024-09-10
Add xrt::runlist::state() to return runlist ERT command state

stsoe:
The xrt::runlist::state() API mirrors xrt::run::state() and is the preferred method for polling a runlist for completion.

Except for return value, the xrt::runlist::poll() API is functionally the same, but is now deprecated.

The state() function returns ERT_CMD_STATE_COMPLETED for an empty runlist.  It otherwise reflects the 
ert command state of the runlist itself.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -3264,9 +3264,21 @@ public:
   }
 
   ert_cmd_state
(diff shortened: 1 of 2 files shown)


maxzhen (approved):

stsoe:
@maxzhen.  Amened state return value to reflect potential error state of the first submitted chained command.

maxzhen (approved):

PR #8409, opened 2024-09-10
Update Xilinx copyright

srioguri:


stsoe (commented):
I think that when you remove 2023 from Xilinx Inc copyright, you should add 2023 Advanced Micro Devices, Inc.  We were originally told to use Xilinx copyright for 2023, so you will find that in many files.  Now that we change this we really should add the AMD line instead.

PR #8416, opened 2024-09-13
Use 64 bit timestamps in ML Timeline Record timer data

IshitaGhosh:
ML Timeline Record Timer was using 32 bit timestamps. But large long running design can overflow 32 bit timer, so we need to use 64bit timestamps. This also helps match with AIE Event trace 64bit timestamped data.

Large long running designs showed overflow in 32 bit timer
 
Long running designs with and without AIE Event trace


jvillarre (approved):

IshitaGhosh:
@stsoe Thanks for merging it

PR #8419, opened 2024-09-13
Creating initial support for NPU3 no-op test latency

hlaccabu:
Introducing xrt-smi npu3 no-op validate commands

xrt-smi validate latency fully functional with npu3 driver, no additional testing

### DIFF
--- a/src/runtime_src/core/tools/xbreplay/CMakeLists.txt
+++ b/src/runtime_src/core/tools/xbreplay/CMakeLists.txt
@@ -22,6 +22,8 @@ add_library(xbreplay_objects OBJECT
   src/replay_xrt/replay_xrt_kernel.cpp
   src/replay_xrt/replay_xrt_run.cpp
   src/replay_xrt/replay_xrt_xclbin.cpp
+  src/replay_xrt/replay_xrt_module.cpp
+  src/replay_xrt/replay_xrt_elf.cpp
 )
 
 target_include_directories(xbreplay_objects PRIVATE
--- a/src/runtime_src/core/tools/xbreplay/src/replay_xrt/replay_xrt.hpp
+++ b/src/runtime_src/core/tools/xbreplay/src/replay_xrt/replay_xrt.hpp
@@ -69,6 +69,12 @@ class replay_xrt
 
   std::map <std::string, std::function < void (std::shared_ptr<utils::message>)>> m_api_map;
 
+  /*Map between handle from log and xrt::module */
+  std::unordered_map<uint64_t, std::shared_ptr<xrt::module>> m_module_hndle_map;
+
+  /*Map between handle from log and xrt::elf */
+  std::unordered_map<uint64_t, std::shared_ptr<xrt::elf>> m_elf_hndle_map;
+
   /* Registers device class API's */
   void register_device_class_func();
 
@@ -86,6 +92,12 @@ class replay_xrt
 
   /*Register hw_ctx class API's*/
   void register_hwctxt_class_func();
+
+  /*Register module class API s */
+  void register_module_class_func();
+
+  /*Register elf class API s */
+  void register_elf_class_func();
 
   /* Validate's file path */
   bool is_file_exist(const std::string& fileName)
--- a/src/runtime_src/core/tools/xbreplay/src/replay_xrt/replay_xrt_device.cpp
+++ b/src/runtime_src/core/tools/xbreplay/src/replay_xrt/replay_xrt_device.cpp
@@ -62,7 +62,7 @@ void replay_xrt::register_device_class_func()
   [this] (std::shared_ptr<utils::message>msg)
   {
     const std::vector <std::pair<std::string, std::string>> &args = msg->m_args;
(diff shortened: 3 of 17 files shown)


src/runtime_src/core/common/query_requests.h
```diff
@@ -532,7 +532,8 @@ struct xclbin_name : request
 {
   enum class type {
     validate,
-    gemm
+    gemm,
+    nop
```
AShivangi, line 536: let's reuse validate and not add nop

src/runtime_src/core/tools/common/tests/TestNPU3Latency.cpp
```diff
@@ -0,0 +1,152 @@
...
+  const auto elf_name = xrt_core::device_query<xrt_core::query::sequence_name>(dev, xrt_core::query::sequence_name::type::nop);
+  auto elf_path = findPlatformFile(elf_name, ptree);
+  if (!std::filesystem::exists(elf_path))
+    return ptree;
+
+  logger(ptree, "Elf", elf_path);
+
+  xrt::elf elf;
+  try {
+    elf = xrt::elf(elf_path);
```
AShivangi, line 62: we need elf for noop? We didn't need dpu seq on prev devices so wondering what has changed.

src/runtime_src/core/tools/xbutil2/xbutil.cpp
```diff
@@ -66,7 +66,7 @@ R"(
     }]
   },{
     "validate": [{
-      "test": ["latency", "throughput", "cmd-chain-latency", "cmd-chain-throughput", "df-bw", "tct-one-col", "tct-all-col", "gemm", "aie-reconfig-overhead"]
+      "test": ["latency", "throughput", "npu3latency", "npu3throughput", "cmd-chain-latency", "cmd-chain-throughput", "df-bw", "tct-one-col", "tct-all-col", "gemm", "aie-reconfig-overhead"]
```
AShivangi, line 69: we don't want to add "npu3latency" the user should be indifferent to npu*. The new test should work with "latency"

src/runtime_src/core/tools/common/tests/TestNPU3Throughput.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+  xrt::elf elf;
+  try {
+    elf = xrt::elf(elf_path);
+  }
+  catch (const std::runtime_error& ex) {
+    logger(ptree, "Error", ex.what());
+    ptree.put("status", test_token_failed);
+    return ptree;
+  }
```
AShivangi, line 67: From what I can tell, this is the only part which is different from previous devices. Is there a way we can use the older host code and trigger this part only when npu3 is detected? You can probably query the device name and see if it has "npu3" in it. 
```
if(xrt_core::device_query<xq::rom_vbnv>(device).find("npu3") != std::string::npos)
    //run elf stuff
```
I'm not fond of the device name search thing, but we'll have better way of managing this in the future. For now this should be ok

AShivangi (changes requested):
@hlaccabu, You should be able to test your code out by manually making these changes in the xrt submodule of the driver

src/runtime_src/core/tools/common/tests/TestNPU3Throughput.cpp
xdavidz: @AShivangi the npu3 has different device type as UMQ.
https://github.com/amd/xdna-driver/blob/f158f2e5ab27ce21ed7f488823d4fe33dc8597b3/src/include/uapi/drm_local/amdxdna_accel.h#L60

However, I think the KMQ might be switched to new flow as well in the future.
We can temporary use the device type UMQ for NPU3, and eventually all xmi should switch to elf flow?

xdavidz (approved):

src/runtime_src/core/tools/common/tests/TestNPU3Latency.cpp
```diff
@@ -0,0 +1,152 @@
...
+  xrt::bo bo_ofm(working_dev, buffer_size, XRT_BO_FLAGS_HOST_ONLY, testker.group_id(argno++));
+  xrt::bo bo_inter(working_dev, buffer_size, XRT_BO_FLAGS_HOST_ONLY, testker.group_id(argno++));
+  xrt::bo bo_instr(working_dev, buffer_size, XCL_BO_FLAGS_CACHEABLE, testker.group_id(argno++));
+  argno++;
+  xrt::bo bo_mc(working_dev, buffer_size, XRT_BO_FLAGS_HOST_ONLY, testker.group_id(argno++));
+  //Create ctrlcode with NOPs
+  std::memset(bo_instr.map<char*>(), 0, buffer_size);
+
+  //Sync BOs
+  bo_instr.sync(XCL_BO_SYNC_BO_TO_DEVICE);
```
xdavidz, line 118: is this sync inherited from older driver? NPU3 doesn't need sync, but it is ok, because sync will just return 0 without any operation done.

xdavidz (approved):

PR #8420, opened 2024-09-14
External Buffer implementation with new AIE driver API

chvamshi-xilinx:
 AIE driver team provided a new API to update the bd addresses. Using that new API now.

None

There are no alternative solution. Using new AIE driver APIs now

None

Verified external buffer tests

Need to add external buffer documentation.


chvamshi-xilinx:
This change cannot be merged until AIE driver APIs are available in Petalinux. Marking this PR as "donot merge"

PR #8422, opened 2024-09-16
CI/CD: Adding workflow_dispatch and workflow_call for manual trigger

manikandan-xilinx:
1. Ability to do manual trigger for workflow_dispatch and workflow_call.
2. Continue on error  to workaround build failures during cleanup due to legacy files getting generated in the build with random user


manikandan-xilinx:
@sureshreddyavula @rbramand-xilinx  as per our discussion added a logic to continue on error in this PR with even though it fails during cleanup as this is a less frequent request

sureshreddyavula (approved):

PR #8423, opened 2024-09-16
Enhance no-op test logics

AShivangi:
Creating run objects inside the loop that is being timed increases the latency by 1-2 us

Developer testing by Max

latency: I create an object and before starting the timer and call start/wait inside the loop
throughput: I enqueue 9 objects and as soon as an object finishes, I enqueue another object

N/A

Tested on stx/mcdm

N/A


maxzhen (approved):

AShivangi:
@rchane is testing it out on linux to make sure there's no regression. Will update the PR accordingly

AShivangi:
birman29 wasn't giving correct numbers, but Max tested it out on his Strix/linux and it works as expected. This PR is ready to be merged

PR #8424, opened 2024-09-17
[XDP] Fix for  profiling plugins to work on independent compilation designs

vipangul:
[CR-1201066][CR-1211412] Independent compilation designs were failing for AIE Trace/Profile plugin

Update: AIE Status Plugin is also supported on Edge.

NA
Metadata has partition column start shift available. This is now accounted to get the absolute column location from relative locations in metadata.

Minimal . by default shift is zero if partition shift is not populated or older metadata is used 

Verified on VCK190 Edge.
Update: AIE Status Plugin is also verified on Edge.


src/runtime_src/xdp/profile/database/static_info/filetypes/aie_trace_config_filetype.cpp
```diff
@@ -31,6 +31,22 @@ using severity_level = xrt_core::message::severity_level;
...
 : AIEControlConfigFiletype(aie_project) {}
 
+std::vector<uint8_t>
+AIETraceConfigFiletype::getPartitionOverlayStartCols() const {
+    auto partitionOverlays = aie_meta.get_child_optional("aie_metadata.driver_config.partition_overlay_start_cols");
+    if (!partitionOverlays) {
+        return std::vector<uint8_t>{0};
+    }
+
+    std::vector<uint8_t> allStartColShits;
```
pgschuey, line 41: I'm assuming you meant to add an 'f' here towards the end of the name? :-)

pgschuey (changes requested):
Looks good as we discussed. Please address variable name change before merging.

jvillarre (approved):

PR #8430, opened 2024-09-18
Added checks to xrt::version api's to throw exception when information is not available

ManojTakasi:
In canonical/debian build for embedded platforms, we create achieve file of our source code and remove all .git files (this is standard process and .git files will not be archieved in their flow). xrt_version.cpp file we are returning typed values like uint from #def 's in xrt_version.h file which would be empty in this case because of lack of .git files.

Modified cmake to set the defualt value as 0 and added checks to version api's to throw exception when information is not available.

low
tested with u250 and build_edge_deb flow

NA

### DIFF
--- a/build/debian/control
+++ b/build/debian/control
@@ -29,6 +29,7 @@ Build-Depends: cmake,
                pkg-config,
                protobuf-compiler,
                rapidjson-dev,
+               systemtap-sdt-dev,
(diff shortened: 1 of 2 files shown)


chvamshi-xilinx (approved):

stsoe (approved):
LGTM

PR #8431, opened 2024-09-18
graph_object constructor with aieArray instance

SravanKumarAllu-xilinx:

proper aieArray instance (hwctx / device) will be set

N/A

proper aieArray instance (hwctx / device) is set

N/A

graph test cases with device / hwctx

N/A

### DIFF
--- a/src/runtime_src/core/edge/user/aie/graph_object.cpp
+++ b/src/runtime_src/core/edge/user/aie/graph_object.cpp
@@ -17,10 +17,12 @@ namespace zynqaie {
       auto device{xrt_core::get_userpf_device(m_shim)};
       auto drv = ZYNQ::shim::handleCheck(device->get_device_handle());
 
-      if (!drv->isAieRegistered())
-        throw xrt_core::error(-EINVAL, "No AIE presented");
-
-      aieArray = drv->get_aie_array_shared();
+#ifdef XRT_ENABLE_AIE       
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/edge/user/aie/graph_object.cpp
```diff
@@ -17,10 +17,15 @@ namespace zynqaie {
...
-
-      aieArray = drv->get_aie_array_shared();
+#ifdef XRT_ENABLE_AIE       
+      if (nullptr != m_hwctx) {
+        aieArray = m_hwctx->get_aie_array_shared(); 
+      }
+      else {
+        if (drv->isAieRegistered())
+          aieArray = drv->get_aie_array_shared();
+      }
```
stsoe, line 27:
```suggestion
      if (m_hwctx)
        aieArray = m_hwctx->get_aie_array_shared(); 
      else if (drv->isAieRegistered())
        aieArray = drv->get_aie_array_shared();
```

And just to confirm, it is okay for `m_hwctx` to be nullptr in line 30?

stsoe (changes requested):

src/runtime_src/core/edge/user/aie/graph_object.cpp
SravanKumarAllu-xilinx: it is still ok, as null check is taken care in called function

PR #8006, opened 2024-03-12
XDP support for AIE Reload Reset feature on Edge platform.

vipangul:
#### Problem solved by the commit
This is planned feature support of XDP plugins for AIE reload reset tracked via Vitis-8456

#### Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered
Planned enhancement

#### How problem was solved, alternative solutions (if any) and why they were rejected
This feature considers a new concept of config , where each config could contain multiple xclbins
that are needed for the successful run. E.g. PL only and AIE only xclbins.

#### Risks (if any) associated the changes in the commit
Moderate.

#### What has been tested and how, request additional testing if necessary
Verified on edge and client.

PR #8434, opened 2024-09-19
[XDP] Fix Freq store & fetch from database

vipangul:
Currently XRT runtime plugins wasn't storing the frequency from metadata correctly. This resulted in misalignment of timelines in post-processing like system timeline.
Also added try-catch block of device query to fetch xclbins UUIDs.

[https://github.com/Xilinx/XRT/pull/8006](https://github.com/Xilinx/XRT/pull/8006/files#diff-ab0236d72c813d03912fcb3166ce5474055f569d19978e2a436267820a07db2e)

Method to  set clock-frequency in database is now moved after configInfo creation.

Low

Verified on edge that correct frequency is reported in aie profile plugin report.

### DIFF
--- a/src/runtime_src/xdp/profile/database/static_info/aie_util.cpp
+++ b/src/runtime_src/xdp/profile/database/static_info/aie_util.cpp
@@ -133,6 +133,18 @@ namespace xdp::aie {
     return *hwGen;
   }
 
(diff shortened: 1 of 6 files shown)


jvillarre (approved):

vipangul:
@gbuildx Please retest

PR #8437, opened 2024-09-19
Vitis 13170 - Allow setting thread CPU affinity for XRT applications on Windows

diukirde:
We want to set cpu affinity for threads on Windows like we do currently for Linux. 
#### Issue
https://jira.xilinx.com/browse/VITIS-13170
#### How problem was solved
Using the APIs provided by windows - SetThreadAffinityMask function (winbase.h)

#### What has been tested and how
cpu_affinity is being honored from xrt.ini. Verified affinity mask calculation seems to be accurate. No error in the API call observed.

cpu_affinity

None


AShivangi:
Do we want to show this in host report? We used to have it, but it has been removed since

src/runtime_src/core/common/thread.cpp
```diff
@@ -140,8 +140,47 @@ set_thread_policy(std::thread&)
 }
 
 static void
-set_cpu_affinity(std::thread&)
+set_cpu_affinity(std::thread& thread)
 {
+  static bool initialized = false;
+  static DWORD_PTR affinityMask = 0;
+  static bool all = false;
```
stsoe, line 147: Let's fix this here and for Linux as well.  Remove this `all` variable and replace all occurrences of `all = true;` with `return;`.  Drop the outer `else` and reduce indentation with early-exit.

src/runtime_src/core/common/thread.cpp
```diff
@@ -140,8 +140,47 @@ set_thread_policy(std::thread&)
...
+
+  if (!initialized) {
+    initialized = true;
+
+    std::string cpus = xrt_core::config::detail::get_string_value("Runtime.cpu_affinity", "default");
+    if (cpus == "default") {
+      all = true;
+    }
```
stsoe, line 155:
```suggestion
    if (cpus == "default")
      return;

```

src/runtime_src/core/common/thread.cpp
```diff
@@ -140,8 +140,47 @@ set_thread_policy(std::thread&)
...
+    else {
```
stsoe, line 156: Remove and reduce identation.
```suggestion
```

src/runtime_src/core/common/thread.cpp
```diff
@@ -140,8 +140,47 @@ set_thread_policy(std::thread&)
...
+        else {
+          xrt_core::message::send(xrt_core::message::severity_level::warning, "XRT", "Ignoring cpu affinity since cpu #" + tok + " is out of range\n");
+          all = true;
+        }
+      }
+    }
+  }
+
+  if (all)
+    return;
```
stsoe, line 178: Not needed.
```suggestion
```

src/runtime_src/core/common/thread.cpp
```diff
@@ -140,8 +140,47 @@ set_thread_policy(std::thread&)
...
+        auto cpu = std::stoul(tok);
+        cpu = static_cast<DWORD_PTR>(cpu);
+        auto one = static_cast<DWORD_PTR>(1U);
+        if (cpu < max_cpus) {
+          XRT_DEBUG(std::cout, "adding cpu #", cpu, " to affinity mask\n");
+          affinityMask |= (one << cpu);  // Set the respective bit for the CPU
+        }
+        else {
+          xrt_core::message::send(xrt_core::message::severity_level::warning, "XRT", "Ignoring cpu affinity since cpu #" + tok + " is out of range\n");
+          all = true;
```
stsoe, line 171:
```suggestion
          return;
```

src/runtime_src/core/common/thread.cpp
```diff
@@ -140,8 +140,47 @@ set_thread_policy(std::thread&)
 }
 
 static void
-set_cpu_affinity(std::thread&)
+set_cpu_affinity(std::thread& thread)
 {
+  static bool initialized = false;
+  static DWORD_PTR affinityMask = 0;
```
stsoe, line 146: user space code never uses camelCase; of course windows functions are CamelCase, but variables we create should not be.

src/runtime_src/core/common/thread.cpp
```diff
@@ -140,8 +140,47 @@ set_thread_policy(std::thread&)
...
+    }
+  }
+
+  if (all)
+    return;
+
+  HANDLE threadHandle = thread.native_handle();
+  if (!SetThreadAffinityMask(threadHandle, affinityMask)) {
+    throw std::runtime_error("error calling SetThreadAffinityMask");
+  }
```
stsoe, line 183:
```suggestion
  if (!SetThreadAffinityMask(threadHandle, affinityMask))
    throw std::runtime_error("error calling SetThreadAffinityMask");
```

src/runtime_src/core/common/thread.cpp
```diff
@@ -140,8 +140,47 @@ set_thread_policy(std::thread&)
...
+
+    std::string cpus = xrt_core::config::detail::get_string_value("Runtime.cpu_affinity", "default");
+    if (cpus == "default") {
+      all = true;
+    }
+    else {
+      boost::trim_if(cpus, boost::is_any_of("{}"));
+      using tokenizer = boost::tokenizer<boost::char_separator<char>>;
+      boost::char_separator<char> sep(", ");
+      auto max_cpus = GetMaximumProcessorCount(ALL_PROCESSOR_GROUPS);
```
stsoe, line 160: Does `std::thread::hardware_concurrency` not work?

stsoe (changes requested):

stsoe (approved):
LGTM.  Some opportunity for refactoring of code shared between Linux and Windows, but no big deal.

stsoe (approved):
Looks good :-)  So sorry for the confusion.

PR #8441, opened 2024-09-23
VITIS-12882 Add limits for benchmarks to specify pass criteria

AShivangi:
[VITIS-12882](https://jira.xilinx.com/browse/VITIS-12882) Add limits for benchmarks to specify pass criteria
Spec page can be found [here](https://confluence.amd.com/x/4QcRW)

New feature
Currently, we pass all the benchmarks in validate. This PR compares the benchmark value to the threshold/expected value for a given device and based on that decides if the test has passed or failed. 

A json will be added per platform to VTD (benchmark_devid_rev_id.json) which will be used for comparison

Low: might print out some unexpected messages, but the functionality will not be compromised

Tested all possible scenarios on Strix/MCDM. Please refer to the spec page for expected behavior

N/A

### DIFF
--- a/src/runtime_src/core/tools/common/tests/TestTCTOneColumn.cpp
+++ b/src/runtime_src/core/tools/common/tests/TestTCTOneColumn.cpp
@@ -30,11 +30,10 @@ TestTCTOneColumn::run(std::shared_ptr<xrt_core::device> dev)
   boost::property_tree::ptree ptree = get_test_header();
   ptree.erase("xclbin");
 
-  double m_threshold = 0.0;
   try {
-    m_threshold = find_threshold(dev, ptree);
+    set_threshold(dev, ptree);
     if(XBU::getVerbose())
-      logger(ptree, "Deatils", boost::str(boost::format("Threshold is %.1f TCT/s") % m_threshold));
+      logger(ptree, "Deatils", boost::str(boost::format("Threshold is %.1f TCT/s") % get_threshold()));
   }
   catch (const std::runtime_error& ex) {
     logger(ptree, "Details", ex.what());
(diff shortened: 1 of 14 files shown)


rchane (approved):
Seems good to me! Thank you for doing additional code cleanup, too 👍

aktondak:
If every benchmark has a threshold, is it possible to move the m_threshold to TestRunner class and let it set the value in its ctor/initializer instead of each test doing it separately?

aktondak:
> > If every benchmark has a threshold, is it possible to move the m_threshold to TestRunner class and let it set the value in its ctor/initializer instead of each test doing it separately?
> 
> I think that needs to come as a part of cleanup, which we can probably do with run recipe implementation. So currently, `find_threshold(...)` takes ptree and populates it with a few messages. This ptree is initialized in `run(...)` and not the ctor.
> 
> A lot of the functions in these tests rely on passing in ptree as a parameter and I do see an opportunity for refactor in the future.

I agree this needs some level of refactoring. But to start, we can still move m_threshold to TestRunner class and change "find_threshold" to "set_threshold" (from getter to setter type). This forces any new benchmark to forcefully take care of "threshold" setting as well. 
"m_" should be reserved for a class member variable.

AShivangi:
Makes sense! I've made m_threshold a member variable :)

src/runtime_src/core/tools/common/tests/TestTCTOneColumn.cpp
```diff
@@ -30,11 +30,10 @@ TestTCTOneColumn::run(std::shared_ptr<xrt_core::device> dev)
   boost::property_tree::ptree ptree = get_test_header();
   ptree.erase("xclbin");
 
-  double m_threshold = 0.0;
   try {
-    m_threshold = find_threshold(dev, ptree);
+    set_threshold(dev, ptree);
     if(XBU::getVerbose())
-      logger(ptree, "Deatils", boost::str(boost::format("Threshold is %.1f TCT/s") % m_threshold));
+      logger(ptree, "Deatils", boost::str(boost::format("Threshold is %.1f TCT/s") % get_threshold()));
```
aktondak, line 36: Typo.

aktondak (approved):

PR #8461, opened 2024-09-26
[Xbutil] Add support to in xbutil for the newly added device utilization queries

MichaelHuang-Amd:
Add Max TOPs value to example report for platform

Not a bug. It's an enhancement to xrt-smi.exe. It was brought up in a code review.
Ticket to remove this workaround for GA: [https://ontrack-internal.amd.com/browse/EDGEML-8265](https://ontrack-internal.amd.com/browse/EDGEML-8265)

Add Maxed Tops value to the section of platform report

Low

Tested on Strix

```shell
>xrt-smi.exe examine -r platform

---------------------
[00c5:00:01.1] : NPU
---------------------
Platform
  Name                   : NPU
  Power Mode             : Default

Clocks
  H Clock                :   0 MHz
  MP-NPU Clock           :   0 MHz

TOPs
  Max Supported TOPs     : 56 Terabyte ops/second

Power                    : 0.000 Watts
```


src/runtime_src/core/include/xclbin.h
```diff
@@ -455,6 +455,30 @@ extern "C" {
...
+            double data_double;            /* Number for double value */
+        };
+    };
+    XCLBIN_STATIC_ASSERT(sizeof(struct res_entry) == 16, "res_entry structure no longer is 16 bytes in size");
+
+    struct res_info {                      /* Resource Infor section */
+        uint16_t m_count;                  /* Number of entries */
+        struct res_entry m_res[1];         /* resource info data */
+    };
+    XCLBIN_STATIC_ASSERT(sizeof(struct res_info) == 24, "res_info structure no longer is 24 bytes in size");
```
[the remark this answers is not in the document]
MichaelHuang-Amd: I just followed the example of clock topology. This is to add TOPs number to the platform report in xrt-smi tool. Maybe xclbin.h is not a good place for these structures. could you suggest the proper header file to host this so that both xrt-smi and kmd driver can reference them? thanks.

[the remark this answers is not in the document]
stsoe: Right, xclbin is not the proper place.  @AShivangi can suggest.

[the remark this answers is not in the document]
MichaelHuang-Amd: fixed.

src/runtime_src/core/tools/common/reports/platform/ReportRyzenPlatform.cpp
```diff
@@ -62,6 +62,16 @@ ReportRyzenPlatform::writeReport(const xrt_core::device* /*_pDevice*/,
       }
     }
 
+    const boost::property_tree::ptree& tops = pt_platform.get_child("tops", empty_ptree);
+    if (!tops.empty()) {
+      _output << std::endl << "TOPs" << std::endl;
```
stsoe, line 67: Please forget that `std::endl` ever existed.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#slio50-avoid-endl

src/runtime_src/core/common/query_requests.h
```diff
@@ -1184,6 +1185,50 @@ struct clock_freq_topology_raw : request
   get(const device*) const = 0;
 };
 
+struct xrt_resource_raw : request
+{
+  enum resource_type {
```
stsoe, line 1190: No plain enums.  Use enum class instead.
xrt_resource_raw is a rather generic name, its fine, but if it is ipu specific, which the enum suggests, then please document this.  In general the structs in this file should be documented, e.g. what exactly is an xrt_resource_raw query?

src/runtime_src/core/common/info_platform.cpp
```diff
@@ -352,6 +352,32 @@ add_clock_info(const xrt_core::device* device, ptree_type& pt)
...
+        for (int i = 0; i < count; i++) {
+            if(resInfo[i].type == xrt_core::query::xrt_resource_raw::resource_type::ipu_tops_max) {
+                ptree_type pt_tops;
+                pt_tops.add("id", xq::xrt_resource_raw::get_name(resInfo[i].type));
+                pt_tops.add("value", resInfo[i].data_double);
+                pt_tops_array.push_back(std::make_pair("", pt_tops));
+            }
+        }
+        pt.put_child("tops", pt_tops_array);
+    }
```
stsoe, line 375: Indentation is not following the convention used in the file.
Do not use `camelCase`, instead use `under_score`.
The query return type is a vector, please use range based for-loop

stsoe (changes requested):

src/runtime_src/core/common/query_requests.h
```diff
@@ -1184,6 +1185,50 @@ struct clock_freq_topology_raw : request
...
+  get_name(xrt_core::query::xrt_resource_raw::resource_type type) {
+      switch (type) {
+      case resource_type::ipu_clk_max:
+          return "Max H-Clocks";
+      case resource_type::ipu_tops_max:
+          return "Max TOPs";
+      case resource_type::ipu_task_max:
+          return "Max Tasks";
+      case resource_type::ipu_tops_curr:
+          return "Current TOPs";
```
AShivangi, line 1220: Do we have this value? I vaguely remember we discussed that we don't have this readily available and that we should start with showing just the Max TOPS. Did something change after that?

src/runtime_src/core/common/query_requests.h
```diff
@@ -1184,6 +1185,50 @@ struct clock_freq_topology_raw : request
...
+    ipu_tops_curr,
+    ipu_task_curr
+  };
+
+  struct xrt_resource_query {
+      resource_type type;
+      union
+      {
+          uint64_t data_uint64;
+          double data_double;
```
AShivangi, line 1203: what are these values?

src/runtime_src/core/common/info_platform.cpp
```diff
@@ -352,6 +352,32 @@ add_clock_info(const xrt_core::device* device, ptree_type& pt)
   }
 }
 
+void
+add_tops_info(const xrt_core::device* device, ptree_type& pt)
+{
+    ptree_type pt_tops_array;
+
+    try {
+        auto resInfo = xrt_core::device_query<xq::xrt_resource_raw>(device);
```
AShivangi, line 361: instead of try/catch, please use `device_query_default`

AShivangi (changes requested):

src/runtime_src/core/common/query_requests.h
MichaelHuang-Amd: fixed.

src/runtime_src/core/common/info_platform.cpp
MichaelHuang-Amd: camelCase is fixed. 
For now, we only support displaying max TOPs value. In the future, we will expand the supporting list.  We need to loop the result to find out the one value we need.

src/runtime_src/core/tools/common/reports/platform/ReportRyzenPlatform.cpp
MichaelHuang-Amd: fixed.

src/runtime_src/core/common/query_requests.h
MichaelHuang-Amd: Yes, the decision still holds. We only return MAX TOPs in add_tops_info@info_platform.cpp. But these resources are usually queried in bunch. the code is written in such that we don't have to come back to refactoring the query resources code to support more resources when they are ready. later we just need to add them to the returned data buffer, that will be a limited change.

src/runtime_src/core/common/info_platform.cpp
MichaelHuang-Amd: but the return value is an array of multiple values. we don't have default values for them.

src/runtime_src/core/common/info_platform.cpp
stsoe: Understood, but please loop using range-based for-loop.
```
for (auto& res : res_info) {
  if (res.type != ...::ipu_tops_max)
    continue;
  ...;
}
```

MichaelHuang-Amd: fixed. thanks.

src/runtime_src/core/common/query_requests.h
MichaelHuang-Amd: Added comments, thanks.

src/runtime_src/core/common/info_platform.cpp
```diff
@@ -352,6 +352,32 @@ add_clock_info(const xrt_core::device* device, ptree_type& pt)
...
+
+        for (auto& res: res_info) {
+            if(res.type == xrt_core::query::xrt_resource_raw::resource_type::ipu_tops_max) {
+                ptree_type pt_tops;
+                pt_tops.add("id", xq::xrt_resource_raw::get_name(res.type));
+                pt_tops.add("value", res.data_double);
+                pt_tops_array.push_back(std::make_pair("", pt_tops));
+                break;
+            }
+        }
```
stsoe, line 373: XRT uses early-exit to reduce the level of indentation, please change to:

```suggestion
        for (auto& res: res_info) {
          if (res.type != xrt_core::query::xrt_resource_raw::resource_type::ipu_tops_max) 
            continue;
            
          ptree_type pt_tops;
          pt_tops.add("id", xq::xrt_resource_raw::get_name(res.type));
          pt_tops.add("value", res.data_double);
          pt_tops_array.push_back(std::make_pair("", pt_tops));
          return;
        }
```

stsoe (changes requested):

src/runtime_src/core/common/info_platform.cpp
MichaelHuang-Amd: fixed.

src/runtime_src/core/common/info_platform.cpp
```diff
@@ -352,6 +352,32 @@ add_clock_info(const xrt_core::device* device, ptree_type& pt)
...
+            pt_tops.add("id", xq::xrt_resource_raw::get_name(res.type));
+            pt_tops.add("value", res.data_double);
+            pt_tops_array.push_back(std::make_pair("", pt_tops));
+        }
+        pt.put_child("tops", pt_tops_array);
+    }
+    catch (const xq::no_such_key&) {
+        // ignoring if not available: Edge Case
+    }
+}
```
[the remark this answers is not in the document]
MichaelHuang-Amd: no worries, Soren, it's good to keep it consistent. 
fixed.

stsoe (approved):
Thank you!

src/runtime_src/core/tools/common/reports/platform/ReportRyzenPlatform.cpp
```diff
@@ -62,6 +62,16 @@ ReportRyzenPlatform::writeReport(const xrt_core::device* /*_pDevice*/,
       }
     }
 
+    const boost::property_tree::ptree& tops = pt_platform.get_child("tops", empty_ptree);
+    if (!tops.empty()) {
+      _output << "\nTOPs\n";
```
AShivangi, line 67: Should we say something like "Max Supported TOPs" or something?

AShivangi:
Can you please add some outputs so we can make sure that the output look consistent?
The rest looks good to me :)

src/runtime_src/core/tools/common/reports/platform/ReportRyzenPlatform.cpp
MichaelHuang-Amd: fixed.

AShivangi (approved):

PR #8462, opened 2024-09-27
CR-1214311 Fix xrt-smi validate crash on alveo due to query

rchane:
https://jira.xilinx.com/browse/CR-1214311
Crash of xrt-smi validate due to query; discovered while regression testing.
Changed the performance query to use device_query_default to prevent the crash, as asked in CR.
N/A
Tested on my local Linux machine, but as I don't have all the alveo cards in the CR, please test on those cards.
N/A

### DIFF
--- a/src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
+++ b/src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
@@ -691,7 +691,7 @@ SubCmdValidate::execute(const SubCmdOptions& _options) const
   }
 
   //get current performance mode
-  const auto curr_mode = xrt_core::device_query<xrt_core::query::performance_mode>(device);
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -691,7 +691,7 @@ SubCmdValidate::execute(const SubCmdOptions& _options) const
   }
 
   //get current performance mode
-  const auto curr_mode = xrt_core::device_query<xrt_core::query::performance_mode>(device);
+  const auto curr_mode = xrt_core::device_query_default<xrt_core::query::performance_mode>(device, 0);
```
AShivangi, line 694: So if it is not supported, it will show default?

rchane: Yes, it's how it would show up as default in the ptree as well in the previous query.

AShivangi (approved):

PR #8463, opened 2024-09-27
Add DRC for dpu_kernel_id of AIE_PARTITION section and m_kernel_id of IP_LAYOUT section. 

xvijaysri:
Need a DRC to check if dpu_kernel_ids of AIE_PARTITION section matches with the m_kernel_ids of IP_LAYOUT section because the dpu_kernel_id is the key in mapping between CU and PDI.

For more info, refer to Tom's comments in [CR-1176455](https://jira.xilinx.com/browse/CR-1176455).

Testcases attached in [CR-1176455](https://jira.xilinx.com/browse/CR-1176455) were used to test if the DRC is triggered when the dpu_kernel_id of AIE_PARTITION metadata and m_kernel_id of IP_LAYOUT metadata are not compliant.

### DIFF
--- a/src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx
+++ b/src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx
@@ -32,6 +32,7 @@
 #include <iostream>
 #include <memory>
 #include <vector>
+#include <set>
 
 #if (BOOST_VERSION >= 106400)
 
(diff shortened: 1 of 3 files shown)


src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx
```diff
@@ -909,7 +910,44 @@ transformMemoryBankGroupingCollections(const std::vector<boost::property_tree::p
...
+  boost::property_tree::ptree ipDatas = ptIPlayout.get_child("m_ip_data");
+  for (const auto& element : ipDatas) {
+    boost::property_tree::ptree ptIPData = element.second;
+    std::string sm_type = ptIPData.get<std::string>("m_type");
+    std::string sSubType = ptIPData.get<std::string>("m_subtype", "");
+    // Check if each m_kernel_id is present in the set of all dpu_kernel_ids
+    if(sm_type == "IP_PS_KERNEL" && sSubType == "DPU"){
+      std::string sKernelId = ptIPData.get<std::string>("m_kernel_id", "");
+      if(allDpuKernelIDs.find(sKernelId) == allDpuKernelIDs.end()){
+        return false;
```
xfreid, line 945: Add a message to indicate which sKernelId is not found in aie_partition

xfreid: Adding a unit test might be worthy, you can do that later

src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx
```diff
@@ -909,7 +910,44 @@ transformMemoryBankGroupingCollections(const std::vector<boost::property_tree::p
...
+      allDpuKernelIDs.insert(dpuKernelIDs.begin(), dpuKernelIDs.end());
+    }
+  }
+                                             
+  // Get IP_LAYOUT metadata
+  Section *pIPLayout = xclbin.findSection(IP_LAYOUT);
+  boost::property_tree::ptree ptIPLayout;
+  pIPLayout->getPayload(ptIPLayout);
+  boost::property_tree::ptree ptIPlayout = ptIPLayout.get_child("ip_layout");
+  boost::property_tree::ptree ipDatas = ptIPlayout.get_child("m_ip_data");
```
xfreid, line 936: I thought we should use XUtil::as_vector here since m_ip_data is an array

xfreid (commented):
In https://jira.xilinx.com/browse/CR-1176455
Two DRCs were requested, I only saw the code change for first DRC
You might need to add the implementation for the second DRC - "2.  This is to check the "type" of cdo_groups in aie_partition.json. The "PRIMARY" pdi type is not allowed anymore. "

src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx
xvijaysri: I am iterating over all the elements in ipDatas below in line 938 which is equivalent to your approach of using XUtil::as_vector (refer def of as_vector). But would need to iterate this vector to check the elements in this vector.

larry9523:
> In https://jira.xilinx.com/browse/CR-1176455 Two DRCs were requested, I only saw the code change for first DRC You might need to add the implementation for the second DRC - "2. This is to check the "type" of cdo_groups in aie_partition.json. The "PRIMARY" pdi type is not allowed anymore. "

Let's leave "PRIMARY" out of the DRC for now. The CR was created when the FW was handling "PRIMARY" and "PRE_POST" differently. But now, the FW treats them equaly. I am worried if we add the DRC, it will break the existing cases where they are still using "PRIMARY" as a PDI type.

src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx
```diff
@@ -909,7 +910,45 @@ transformMemoryBankGroupingCollections(const std::vector<boost::property_tree::p
...
+  std::vector<boost::property_tree::ptree> ptPDIs = XUtil::as_vector<boost::property_tree::ptree>(ptAIEPartition, "PDIs");
+  for (const auto& pdi : ptPDIs) {
+    std::vector<boost::property_tree::ptree> ptCDOs = XUtil::as_vector<boost::property_tree::ptree>(pdi, "cdo_groups");
+    for (const auto& cdo_group : ptCDOs) {
+      std::vector<std::string> dpuKernelIDs = XUtil::as_vector_simple<std::string>(cdo_group, "dpu_kernel_ids");
+      allDpuKernelIDs.insert(dpuKernelIDs.begin(), dpuKernelIDs.end());
+    }
+  }
+                                             
+  // Get IP_LAYOUT metadata
```
xfreid, line 931: Please add a comment here to briefly describe what the DRC check is

xfreid (approved):

src/runtime_src/tools/xclbinutil/XclBinUtilities.cxx
xvijaysri: Added at the top.

PR #8481, opened 2024-10-01
[XDP] Multiple CRs - Correct parsing of buffer names and channel IDs

pgschuey:
* Buffer names were not verified correctly in graph-based memory tile settings
* Channel IDs in single-tile settings were not parsed correctly

* Gather and use valid set of buffers (instead of kernels)
* Correctly parse channel IDs from single tile settings

* None

* Tested on vck190 and vek280

* Verify that docs already describe buffer name usage

### DIFF
--- a/src/runtime_src/xdp/profile/database/static_info/filetypes/aie_control_config_filetype.cpp
+++ b/src/runtime_src/xdp/profile/database/static_info/filetypes/aie_control_config_filetype.cpp
@@ -126,6 +126,32 @@ AIEControlConfigFiletype::getValidKernels() const
     return kernels;
 }
 
+std::vector<std::string>
+AIEControlConfigFiletype::getValidBuffers() const
+{
+    if (getHardwareGeneration() == 1) 
+        return {};
+        
+    std::vector<std::string> buffers;
+
+    // Grab all shared buffers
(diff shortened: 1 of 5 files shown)


jvillarre (approved):

src/runtime_src/xdp/profile/database/static_info/filetypes/aie_control_config_filetype.cpp
```diff
@@ -126,6 +126,32 @@ AIEControlConfigFiletype::getValidKernels() const
...
+    if (!sharedBufferTree) {
+        xrt_core::message::send(severity_level::info, "XRT", 
+            getMessage("TileMapping.SharedBufferToTileMapping"));
+        return {};
+    }
+
+    // Now parse all shared buffers
+    for (auto const &shared_buffer : sharedBufferTree.get()) {
+        std::string bufferStr = shared_buffer.second.get<std::string>("bufferName");
+        auto nameStr = bufferStr.substr(bufferStr.find_last_of(".") + 1);
```
vipangul, line 149: Do we need this way checking specified buffer name for memory Tiles as well ?  `getMemoryTiles()`

vipangul (approved):

PR #8482, opened 2024-10-01
CR-1214559 `xrt-smi validate --verbose` cannot find benchmark_1502_00.json and prints "Details: Threshold is 0.0 us"

AShivangi:
[CR-1214559](https://jira.xilinx.com/browse/CR-1214559) `xrt-smi validate --verbose` cannot find benchmark_1502_00.json and prints "Details: Threshold is 0.0 us"

DSV testing

We don't support benchmark.json for phoenix, so we check if it's that. If so, we don't look for benchmark.json

N/A

Tested on phx/strix

N/A

### DIFF
--- a/src/runtime_src/core/tools/common/TestRunner.cpp
+++ b/src/runtime_src/core/tools/common/TestRunner.cpp
@@ -530,7 +530,7 @@ TestRunner::set_threshold(const std::shared_ptr<xrt_core::device>& dev,
   const auto pcie_id = xrt_core::device_query<xq::pcie_id>(dev);
 
   //phoenix is not supported
-  if(xq::pcie_id::device_to_string(pcie_id).find("1500") != std::string::npos)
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/tools/common/TestRunner.cpp
```diff
@@ -528,6 +528,11 @@ TestRunner::set_threshold(const std::shared_ptr<xrt_core::device>& dev,
 {
   //find the benchmark.json
   const auto pcie_id = xrt_core::device_query<xq::pcie_id>(dev);
+
+  //phoenix is not supported
+  if(xq::pcie_id::device_to_string(pcie_id).find("1502") != std::string::npos)
```
stsoe, line 533:
```suggestion
  if (xq::pcie_id::device_to_string(pcie_id).find("1502") != std::string::npos)
```

PR #8503, opened 2024-10-08
VITIS-13530 : Add support to dump scratch pad memory

rbramand-xilinx:
Added support to dump scratch pad memory when Debug ini option dump_scratchpad_mem is enabled.

Added support as part of VITIS-13530

After kernel run is finished dumping scratch pad memory to a file when required ini option is set. No extra or minimal overhead is added as checking of ini option is done at object creation only.
Dumping scratchpad memory after kernel run as that will be the content useful for debugging.

Low

Tested by running application with preemption enabled using RyzenAI strix on windows

Not needed as the ini option is for debug purpose(mostly internal). I will update jira ticket with ini option details.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -2087,7 +2087,6 @@ class run_impl
   bool encode_cumasks = false;            // indicate if cmd cumasks must be re-encoded
   std::shared_ptr<xrt_core::usage_metrics::base_logger> m_usage_logger =
       xrt_core::usage_metrics::get_usage_metrics_logger();
-  bool m_dump_scratchpad_mem = xrt_core::config::get_feature_toggle("Debug.dump_scratchpad_mem");
 
   const runlist_impl* m_runlist = nullptr;// runlist that owns this run (optional)
   std::mutex m_mutex;                     // mutex synchronization
@@ -2479,7 +2478,8 @@ public:
     }
 
     m_usage_logger->log_kernel_run_info(kernel.get(), this, state);
-    if (m_dump_scratchpad_mem)
+    static bool dump = xrt_core::config::get_feature_toggle("Debug.dump_scratchpad_mem");
+    if (dump)
       xrt_core::module_int::dump_scratchpad_mem(m_module);
 
     return state;
@@ -2507,8 +2507,10 @@ public:
 
     if (state == ERT_CMD_STATE_COMPLETED) {
       m_usage_logger->log_kernel_run_info(kernel.get(), this, state);
-      if (m_dump_scratchpad_mem)
+      static bool dump = xrt_core::config::get_feature_toggle("Debug.dump_scratchpad_mem");
+      if (dump)
         xrt_core::module_int::dump_scratchpad_mem(m_module);
+
       return std::cv_status::no_timeout;
     }
 
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -1404,18 +1404,15 @@ public:
       return;
     }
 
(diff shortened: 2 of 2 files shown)


src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1394,6 +1394,29 @@ class module_sram : public module_impl
...
+    if (!ofs.is_open())
+      throw std::runtime_error("Failure opening file " + dump_file_name + " for writing!");
+
+    m_scratch_pad_mem.sync(XCL_BO_SYNC_BO_FROM_DEVICE);
+    auto buf = m_scratch_pad_mem.map<char*>();
+    ofs.write(buf, m_scratch_pad_mem.size());
+
+    std::stringstream ss;
+    ss << "dumped file " << dump_file_name;
+    xrt_core::message::send(xrt_core::message::severity_level::debug, "xrt_module", ss.str());
```
stsoe, line 1418: A stream is a bit of an overkill here.
```suggestion
    std::string msg {"dumped file"};
    msg.append(dump_file_name);
    xrt_core::message::send(xrt_core::message::severity_level::debug, "xrt_module",msg);
```

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1394,6 +1394,29 @@ class module_sram : public module_impl
...
+  {
+    if (m_scratch_pad_mem.size() == 0) {
+      xrt_core::message::send(xrt_core::message::severity_level::debug, "xrt_module",
+                              "preemption scratchpad memory is not available");
+      return;
+    }
+
+    std::string dump_file_name = "preemption_scratchpad_mem" + std::to_string(get_id()) + ".bin";
+    std::ofstream ofs(dump_file_name, std::ios::out | std::ios::binary);
+    if (!ofs.is_open())
```
stsoe, line 1409:
```suggestion
    if (!ofs)
```

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -2087,6 +2087,7 @@ class run_impl
   bool encode_cumasks = false;            // indicate if cmd cumasks must be re-encoded
   std::shared_ptr<xrt_core::usage_metrics::base_logger> m_usage_logger =
       xrt_core::usage_metrics::get_usage_metrics_logger();
+  bool m_dump_scratchpad_mem = xrt_core::config::get_feature_toggle("Debug.dump_scratchpad_mem");
```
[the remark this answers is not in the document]
rbramand-xilinx: Hi @stsoe, I have made it data member as it is used in two functions and wanted to check the value at every run object creation as the memory is dumped after kernel run.
Do you prefer adding static variable in two functions?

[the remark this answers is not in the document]
stsoe: I missed that, but yes, I'd prefer static variable in each.

stsoe (approved):

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1394,6 +1394,29 @@ class module_sram : public module_impl
...
+
+  void
+  dump_scratchpad_mem()
+  {
+    if (m_scratch_pad_mem.size() == 0) {
+      xrt_core::message::send(xrt_core::message::severity_level::debug, "xrt_module",
+                              "preemption scratchpad memory is not available");
+      return;
+    }
+
```
larry9523, line 1406: There is already a dump_bo function in this file, can we reuse that one?

rbramand-xilinx: changed code to use existing function

larry9523 (approved):

chvamshi-xilinx (approved):

PR #8506, opened 2024-10-08
Spatial Sharing test update for low variance

aktondak:
The variance seen in the overhead for spatial-sharing-overhead test was high. This Change solves that.
There is some additional code cleanup which moves 2 APIs out of the TestRunner class and creates and API for waiting for clock speed logic for free usage within other code.
The file TestHelper has also been renamed appropriately to be used as a placeholder for future xrt-smi validate APIs.

https://jira.xilinx.com/browse/CR-1215567
https://jira.xilinx.com/browse/CR-1214574

The problem was solved by running the df-bw DPU instruction instead of the no-op kernel being run previously. The test also now waits for the clock speed to reach a threshold value, similar to the Gemm test-case
We are now running 8 kernels simultaneously for 10,000 iterations
The overhead now seen in 10 subsequent runs is seen between 900ms - 1100ms (~10% variance).

None

Tested on Kracken board. Needs further testing from DSV

https://confluence.amd.com/display/AIE/User+Guide%3A+xrt-smi+on+Ryzen+AI will be updated once pulled


rchane (approved):

AShivangi (approved):
Please remove the XrtSmi namespace. Rest looks good to me!

PR #8514, opened 2024-10-11
Capture and Replay: Add support for xrt::runlist APIs

theharshwardhan:
Issue: Caertain applications are calling runlist APIs for which trace and replay support is not there.

Fix: Added support for runlist APIs.


stsoe:
NO longer applicable

PR #8523, opened 2024-10-14
Dont throw an exception if there are no PL Kernels in xclbin

chvamshi-xilinx:
Dont throw an exception if there are no PL Kernels. There are xclbins with only AIE information

None

Catching an exception within the code to move forward.

None

Verified multiple tests with AIE only sections

None


chvamshi-xilinx:
Fixing in better way :-)

PR #8534, opened 2024-10-16
Added hw_context reset array support

chvamshi-xilinx:
We have enabled multi-partition support in edge in this release, There is one functionality missing in this support is to reset the partition instead of whole device. Added hw_context reset support now

None. This is new feature

Enabled hw_context reset array support. Provided a public API on xrt::aie::hw_context class

None

Verified hw_context reset support

Yes. Updated the API Documentation. Need to add an example


stsoe (approved):
Core xrt changes LGTM.

PR #8538, opened 2024-10-16
Vitis 13612 : Allow for gemm test to wait for the clock ramp up to its maximum value

aktondak:
Allow for the gemm test to wait for the clock ramp up to it maximum value.

https://jira.xilinx.com/browse/VITIS-13612
Discovered through a bug seen earlier in the Gemm test-case where metric numbers varied because of the clock not being at threshold level.

The problem was solved through querying the driver for max clock value and allowing validate tests to reach that clock value before beginning time measurement.

Needs more testing.

Tested by running validate tests on kracken and strix boards.

No

### DIFF
--- a/src/runtime_src/core/tools/common/tests/TestValidateUtilities.cpp
+++ b/src/runtime_src/core/tools/common/tests/TestValidateUtilities.cpp
@@ -148,7 +148,7 @@ uint64_t
 wait_for_max_clock(std::shared_ptr<xrt_core::device> dev) {
   uint64_t target_h_clock_freq = 0;
   uint64_t ipu_hclock = 0;
-  auto res_info = xrt_core::device_query<xrt_core::query::xrt_resource_raw>(dev);
+  auto res_info = xrt_core::device_query_default<xrt_core::query::xrt_resource_raw>(dev, {});
   if (res_info.empty())
     return ipu_hclock;
 
--- a/src/runtime_src/core/tools/common/tests/TestValidateUtilities.cpp
+++ b/src/runtime_src/core/tools/common/tests/TestValidateUtilities.cpp
@@ -136,7 +136,8 @@ get_instr_size(const std::string& dpu_file) {
 }
 
 void
-wait_for_max_clock(int& ipu_hclock, std::shared_ptr<xrt_core::device> dev, double& target_h_clock_freq) {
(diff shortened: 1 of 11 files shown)


src/runtime_src/core/tools/common/tests/TestValidateUtilities.cpp
```diff
@@ -137,39 +137,26 @@ get_instr_size(const std::string& dpu_file) {
 
 void
 wait_for_max_clock(int& ipu_hclock, std::shared_ptr<xrt_core::device> dev) {
-  int ipu_hclock_pre = 0;
-  auto hclock_steady_counter = 0;
-  auto first_steady_state = -1, second_steady_state = -1;;
-
-  for(int i=0; i<100;i++){
+  uint64_t target_h_clock_freq = 0;
+  auto res_info = xrt_core::device_query<xrt_core::query::xrt_resource_raw>(dev);
```
AShivangi, line 141: always use `device_query`. You never know which device doesn't support which query!

src/runtime_src/core/tools/common/tests/TestValidateUtilities.cpp
```diff
@@ -137,39 +137,26 @@ get_instr_size(const std::string& dpu_file) {
 
 void
 wait_for_max_clock(int& ipu_hclock, std::shared_ptr<xrt_core::device> dev) {
```
AShivangi, line 139: You don't need to pass `ipu_hclock` in the function

AShivangi, line 139: Add a comment for this function explaining why we are doing this. We won't remember the reasoning behind this later.

AShivangi (changes requested):

src/runtime_src/core/tools/common/tests/TestValidateUtilities.cpp
aktondak: Can you elaborate on this point ? I am using device_query only

src/runtime_src/core/tools/common/tests/TestValidateUtilities.cpp
aktondak: Done.

aktondak: I have changed the API to a returning type

AShivangi (approved):

PR #8555, opened 2024-10-22
Build xrt drivers on Linux 6.12-rc3

dima-anet:
xrt doesn't build on linux kernels newer than v6.10

Add more ifdefs, one patch generated with Сoccinelle.

So far, I only build tested on v6.12-rc3.
And local sources yet use 2023.1 release, so it might not fix all build issues (though, the biggest diff I generated on the latest master).

For older kernels this should be no-op as guarded by ifdefs.

### DIFF
--- a/src/runtime_src/core/common/drv/include/xrt_cu.h
+++ b/src/runtime_src/core/common/drv/include/xrt_cu.h
@@ -413,19 +413,19 @@ static inline char *prot2str(enum CU_PROTOCOL prot)
 	}
 }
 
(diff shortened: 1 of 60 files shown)


dima-anet:
Also just to note: I fixed only build-errors, but there are also quite some warnings. Mostly about inline declarations and missed static keywords or extern declarations.

houlz0507 (approved):

stsoe:
Thank you @dima-anet.  Rebased in #8928.  Sorry for the long hold, we finally ran into the same problems you encountered.

PR #8557, opened 2024-10-22
Change default ml timeline buffer size to 192K

Xlnx78115:
Since the size of cycle value in record_timer_ts.json is changed from from 32bit to 64bit, which will cause the alignment issue when the buffer is overwritten. This commit could resolve this alignment issue.
PR8416 introduces this bug. In the test case, generated record_timer_ts.json contains the wrong records.
Set size value of a multiple of 12K would make the buffer is aligned when it's full. In this case, the issue disappears.
Run test and make sure the buffer is overwritten, then the alignment caused wrong result in record_timer_ts.json has gone.  It means that this commit resolve this issue.

### DIFF
--- a/src/runtime_src/core/common/config_reader.h
+++ b/src/runtime_src/core/common/config_reader.h
@@ -250,7 +250,7 @@ get_ml_timeline()
 inline std::string
 get_ml_timeline_buffer_size()
 {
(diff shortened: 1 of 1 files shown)


stsoe (approved):

stsoe (approved):

IshitaGhosh (approved):
The change looks good. Thanks Jason Fu!

jvillarre (approved):

PR #8567, opened 2024-10-24
running authorize job on - [self-hosted, Ubuntu-22.04]

SubhedarV:
- job is landing on cloud runners hence using [self-hosted, Ubuntu-22.04]
 - https://jira.xilinx.com/browse/SR-960180

### DIFF
--- a/.github/workflows/xrt_ci.yml
+++ b/.github/workflows/xrt_ci.yml
@@ -15,7 +15,7 @@ concurrency:
   
 jobs:
   authorize:
-    runs-on: Ubuntu-22.04
+    runs-on: [self-hosted, Ubuntu-22.04]
     steps:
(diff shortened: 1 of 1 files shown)


sunilvenkatraman-amd (approved):

.github/workflows/xrt_ci.yml
```diff
@@ -15,7 +15,7 @@ concurrency:
   
 jobs:
   authorize:
-    runs-on: Ubuntu-22.04
+    runs-on: [self-hosted, Ubuntu-22.04]
```
sunilvenkatraman-amd, line 18: @venkatp-xilinx @MattSnow-amd Could we not use Ubuntu-22.04 on the self hosted machines, since this is conflicting with the public Github cloud runnners

PR #8568, opened 2024-10-24
Add -npu switch to compile for Lin Client. 

IshitaGhosh:
Compiling with -npu will build xrt package named "npu*" targeted for Linux Client. The Alveo drivers will be disabled and ERT is not included. This will include XDP Plugins targeted only for Client devices. Only host side profiling plugins are included for now. Other features like ML Timeline, AIE profile/trace will be enabled in later PRs.
 
We need to update compilation instructions in amd-xdna repo with this change.


maxzhen (approved):

jvillarre (approved):

stsoe (approved):
LGTM.

PR #8570, opened 2024-10-25
ml_timeline record json file add more info

Xlnx78115:
Because the size of cycle value in record_timer_ts.json is changed from from 32bit to 64bit, when try to parse it, just depends on the information provided by the header, it's not known that the cycle value is 64bit or 32bit, which means it's difficult to be compatible to the previous version. This commit resolves compatible problem. 
Providing id/cycle size would make record_timer_ts.json parser easily distinguish the last version of previous version.
Run test and generate record_timer_ts.json, then check the header which would contain the id/cycle size information.


jvillarre (approved):

IshitaGhosh (approved):
The changes look good.

Xlnx78115:
@jvillarre Thanks for merging it

PR #8583, opened 2024-10-29
VITIS-13786 Update aie partition report to show active and suspended contexts

rchane:
https://jira.xilinx.com/browse/VITIS-13786
Users had concerns over column occupancy/utilization
We will revisit column occupancy in the future and have removed it for now. Removed the double dividers from various tables and made them single dividers as asked. Added status column ("Active" or "Idle") to each context.
Double-check documentation for GA RAI_1.3 to ensure they don't expect column occupancy if this is backported.
Tested on STX/MCDM using Divyajyoti's changes in MCDM for status values. 
Sample Output

Possibly, see comment in risks


AShivangi (approved):
Looks good! Thank you

PR #8593, opened 2024-11-08
Add changes to patch instruction buffer with ctrlpkt preemption buffer addresses

rbramand-xilinx:
We need a way to provide control packet preemption codes packed into ELF to AIE hw.
This PR adds a way to create xrt::bo's  from content extracted from ELF .ctrlpkt.pm.* sections and patch instruction buffer with these addresses.

This is a new feature

Patched instruction buffer with xrt::bo addresses with ctrlpkt-pm content for symbols present in .rela.dyn section

Low

Tested ELF parsing and patching part
TODO : test end to end flow once we have proper application

NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -120,11 +120,13 @@ struct patcher
   };
 
   enum class buf_type {
-    ctrltext = 0,   // control code
-    ctrldata = 1,       // control packet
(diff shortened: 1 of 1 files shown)


larry9523 (approved):
Looks good to me

larry9523:
We have an end 2 end test case working. I just removed the "do not merge" label.

stsoe (approved):

chvamshi-xilinx (approved):

PR #8613, opened 2024-11-22
xrt::bo flags for for carveout memory

haghilin:
Created the xrt::bo flags for carveout memory

Problem solved by the commit
xrt::bo for carveout memory

I reused one of the flags which already been defined and only used in xocl files in linux. xocl driver is not used on npu platform, hence I could reuse one of them for both windows and linux.

None.

Documentation impact (if any)
None.

### DIFF
--- a/src/runtime_src/core/include/xrt/xrt_bo.h
+++ b/src/runtime_src/core/include/xrt/xrt_bo.h
@@ -117,12 +117,13 @@ public:
    */
   enum class flags : uint32_t
   {
-    normal      = 0,
-    cacheable   = XRT_BO_FLAGS_CACHEABLE,
-    device_only = XRT_BO_FLAGS_DEV_ONLY,
-    host_only   = XRT_BO_FLAGS_HOST_ONLY,
-    p2p         = XRT_BO_FLAGS_P2P,
-    svm         = XRT_BO_FLAGS_SVM,
+    normal        = 0,
+    cacheable     = XRT_BO_FLAGS_CACHEABLE,
+    device_only   = XRT_BO_FLAGS_DEV_ONLY,
+    host_only     = XRT_BO_FLAGS_HOST_ONLY,
+    p2p           = XRT_BO_FLAGS_P2P,
+    svm           = XRT_BO_FLAGS_SVM,
+    device_memory = XRT_BO_FLAGS_CARVEOUT,
   };
 
 #ifdef _WIN32
--- a/src/runtime_src/core/include/xrt/xrt_bo.h
+++ b/src/runtime_src/core/include/xrt/xrt_bo.h
@@ -112,7 +112,8 @@ public:
    * @var svm
    *  Create a BO for SVM (supported on specific platforms only)
    * @var device_memory
-   *  Create a BO with a carveout memory buffer. It is an extra separate pool of memory that is assigned for NPU use. 
+   *  Create a BO from device memory, which is a separate pool of
+   *  memory that is supported on specific platforms only.
    * 
    * The flags used by xrt::bo are compatible with XCL style
    * flags as define in ``xrt_mem.h``
--- a/src/runtime_src/core/include/xrt/xrt_bo.h
+++ b/src/runtime_src/core/include/xrt/xrt_bo.h
@@ -112,8 +112,9 @@ public:
    * @var svm
    *  Create a BO for SVM (supported on specific platforms only)
    * @var carveout
-   *  Create a BO from a special memory pool dedicated to the device.
-   *  For AMD Ryzen NPU this memory is allocated from the host memory carveout pool.
+   *  Create a BO from a reserved memory pool. Supported for specific
+   *  platforms only. For AMD Ryzen NPU this memory is allocated from
+   *  a host memory carveout pool.
    * 
    * The flags used by xrt::bo are compatible with XCL style
    * flags as define in ``xrt_mem.h``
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/include/xrt/xrt_bo.h
```diff
@@ -117,12 +117,13 @@ class bo
...
-    host_only   = XRT_BO_FLAGS_HOST_ONLY,
-    p2p         = XRT_BO_FLAGS_P2P,
-    svm         = XRT_BO_FLAGS_SVM,
+    normal        = 0,
+    cacheable     = XRT_BO_FLAGS_CACHEABLE,
+    device_only   = XRT_BO_FLAGS_DEV_ONLY,
+    host_only     = XRT_BO_FLAGS_HOST_ONLY,
+    p2p           = XRT_BO_FLAGS_P2P,
+    svm           = XRT_BO_FLAGS_SVM,
+    device_memory = XRT_BO_FLAGS_CARVEOUT,
```
stsoe, line 126: Since `device_memory` is a public flag, you have to document it along with the other flags above.  Explain what it means, and why it is different from `device_only`.

src/runtime_src/core/include/xrt_mem.h
```diff
@@ -118,6 +118,7 @@ struct xcl_bo_flags
  */
 #define XRT_BO_USE_NORMAL 0
 #define XRT_BO_USE_DEBUG  1
+#define XRT_BO_USE_KMD    2
```
stsoe, line 121: Please add brief comment as to the purpose of this new use flag.  I am asking because I already forgot why we need the flag?  I think it is something like this:  For the table BO we allocate a normal host only BO but we set the USE flag to this new `XRT_BO_USE_KMD` so that shim can take proper action when creating the allocation for the BO?

So amend the comment to something like:
```
The flag `XRT_BO_USE_KMD`  indicates that the buffer content can be shared with the kernel mode driver.  
This flag controls how the shim level constructs the allocation for the buffer.
```

stsoe (changes requested):
Looks good.  Just add some comments.

src/runtime_src/core/include/xrt/xrt_bo.h
haghilin: Thanks Soren, for the comment. I added the comment.

stsoe (approved):

src/runtime_src/core/include/xrt/xrt_bo.h
```diff
@@ -111,18 +111,22 @@ class bo
    *  Create a BO for peer-to-peer use
    * @var svm
    *  Create a BO for SVM (supported on specific platforms only)
-   *
+   * @var device_memory
+   *  Create a BO from device memory, which is a separate pool of
```
sonals, line 115: We should qualify this crisply so end users can understand this.
Create a BO from a special memory pool dedicated to the device.  For AMD Ryzen NPU
this memory is allocated from the host memory carveout pool.

src/runtime_src/core/include/xrt/xrt_bo.h
```diff
@@ -111,18 +111,22 @@ class bo
...
-    host_only   = XRT_BO_FLAGS_HOST_ONLY,
-    p2p         = XRT_BO_FLAGS_P2P,
-    svm         = XRT_BO_FLAGS_SVM,
+    normal        = 0,
+    cacheable     = XRT_BO_FLAGS_CACHEABLE,
+    device_only   = XRT_BO_FLAGS_DEV_ONLY,
+    host_only     = XRT_BO_FLAGS_HOST_ONLY,
+    p2p           = XRT_BO_FLAGS_P2P,
+    svm           = XRT_BO_FLAGS_SVM,
+    device_memory = XRT_BO_FLAGS_CARVEOUT,
```
sonals, line 129: device_memory is misleading. Can we call it what it really is -- carveout?

sonals (changes requested):
Please update as suggested

haghilin:
> Please update as suggested

Thanks @sonals for the comments. I updated the code/comments accordingly.

src/runtime_src/core/include/xrt/xrt_bo.h
stsoe: @sonals carveout is even worse., please don't use that.

src/runtime_src/core/include/xrt/xrt_bo.h
```diff
@@ -111,18 +111,22 @@ class bo
...
-    host_only   = XRT_BO_FLAGS_HOST_ONLY,
-    p2p         = XRT_BO_FLAGS_P2P,
-    svm         = XRT_BO_FLAGS_SVM,
+    normal       = 0,
+    cacheable    = XRT_BO_FLAGS_CACHEABLE,
+    device_only  = XRT_BO_FLAGS_DEV_ONLY,
+    host_only    = XRT_BO_FLAGS_HOST_ONLY,
+    p2p          = XRT_BO_FLAGS_P2P,
+    svm          = XRT_BO_FLAGS_SVM,
+    carveout     = XRT_BO_FLAGS_CARVEOUT,
```
stsoe, line 129: No, carveout is horrible as an API facing symbolic name.   I will discuss with Sonal.

src/runtime_src/core/include/xrt/xrt_bo.h
stsoe: Please don't use carveout.  How  about "allocated from a reserved host memory pool"

src/runtime_src/core/include/xrt/xrt_bo.h
stsoe: I did a bit of search, and carveout is used by android to indicated special heap memory pool created a boot time.  I backtrack and say we can use `carveout` after all.

PR #8616, opened 2024-11-22
For NPU Linux, build only XRT component renamed with suffix "npu"

IshitaGhosh:
xrt package component on Linux Client, built with "-npu", contains runtime and XDP support for NPU. So, rename the xrt component package to "npu". Remove the "npu" prefix. Also, other packages like xbflash, azure, aws, container, runtime, Runtime are not targeted for NPU. 
More specifically, xbflash, azure, aws, container should be built only for Alveo. So, skip these components when Alveo driver build is disabled.
AIEBU creates the “Runtime” component and aiert creates the “runtime” package. To disable these packages for NPU Lin, changes need to be added in their respective submodules. AIE BU PR https://github.com/Xilinx/aiebu/pull/33

As only 1 component is now built under "-npu", CMake does not add the component name in the package. So, this PR hardcodes "npu" suffix to indicate the xrt component is targeted for NPU.
Example, xrt_<xrt_version>....-npu.deb
In future, if more components are enabled for NPU, they will all have a "-npu" substring right before component name. It will look like :
xrt_<xrt_version>....-npu-<component_name>.deb

Installation and removal of xrt package. Unit tests, and "validate" test with ML Timeline feature on Lin Client.


maxzhen (approved):

stsoe:
@IshitaGhosh @maxzhen Can I ask why we are not simply using -DNPU to control the XRT component name and to disable building of the pieces that should not be included?  Why the renaming?

stsoe (commented):
I am not sure I fancy this change.   Would it be possible to use the -DNPU already passed to CMake to switch the component name from `xrt` to `npu` and make use the same CMake variable to disable the pieces that should not be included in the NPU build?  This way there would be no surgery to cpackLin.cmake.

Also, when / if XRT builds submodules that should be included in the -npu package, then these submodules should install themselves to the component name passed down from XRT, which in this case would be `npu`.

IshitaGhosh:
@stsoe Sonal mentioned the aws, azure components should not be even built when Alveo driver is diabled. So, I excluded those components based on "XRT_DKMS_ALVEO", instead of "NPU".
For your 2nd question, CMake does not add the component name in the package when only 1 component is built. So, I had to hardcode the suffix "npu".

stsoe:
Replace by #8619

PR #8619, opened 2024-11-23
Amend npu build option

stsoe:
Create CMake variables for XRT_NPU and XRT_ALVEO in place of NPU and XRT_DKMS_ALVEO.  Modify CMake files accordingly.

Fix CMake file for CPack to create component specific package for single component project.  This will be used for NPU builds where ultimately all but the default npu package will be disabled.

The top level build.sh script defines one of  two CMake variables depending on whether `-npu` is specified or not.
The variables control the default CMake install component., which for `-npu` will be `npu` and otherwise legacy `xrt` component.  The install components created by CMake result in CPack creating packages with component names appended. 

Changes will have to be made to submodules built by XRT to honor any install component specified by XRT prior to including the submodule in CMake.   Basically the installable pieces of a submodule can be added to any XRT component as desired.  Supposedly this will be either the `npu` or default `xrt` component.   The XRT default component is defined in `src/CMakeLists.txt` in the `XRT_COMPONENT` CMake variable which is either `npu` or `xrt` depending on build option. 

This PR renders #8616 obsolete.


src/CMakeLists.txt
```diff
@@ -27,13 +27,21 @@ include(CMake/settings.cmake)
...
+  
 # Enable development package by specifying development component name
 # If XRT_DEV_COMPONENT is same DEFAULT_COMPONENT then only that package
 # is created with both development and run-time content.
 #set (XRT_DEV_COMPONENT "xrt-dev")
+set (XRT_COMPONENT "xrt")
 set (XRT_DEV_COMPONENT "xrt")
 
+# For NPU builds the defalt component is "npu".
+if (XRT_NPU)
```
IshitaGhosh, line 39: This changes install package name to "xrt-npu". Please review whether we want to pursue it.

IshitaGhosh (commented):
Please consider my feedback shared in email.

maxzhen (approved):

AShivangi (approved):

sonals (approved):

src/CMakeLists.txt
stsoe: Agreement that we will pursue -npu build switch with separate -npu.{deb,rpm} knowing that -npu and -xrt canot be installed side-by-side.

PR #8624, opened 2024-11-26
Clang tidy and other code cleanups

sonals:
Renamed patch() to patch_it() to disambiguate between method 'patch' and method argument 'patch'

clang-tidy fixes, minor code reorganization

None

Code changes

None

Compile tested

NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -29,6 +29,7 @@
 #include <map>
 #include <set>
 #include <string>
+#include <string_view>
 #include <sstream>
 
 #ifndef AIE_COLUMN_PAGE_SIZE
@@ -139,10 +140,10 @@ struct patcher
 
   std::vector<patch_info> m_ctrlcode_patchinfo;
 
-  inline static const std::string&
+  inline static const std::string_view
   section_name_to_string(buf_type bt)
   {
-    static const std::array<std::string, static_cast<int>(buf_type::buf_type_count)> Section_Name_Array =
+    static constexpr std::array<std::string_view, static_cast<int>(buf_type::buf_type_count)> Section_Name_Array =
       { ".ctrltext",
         ".ctrldata",
         ".preempt_save",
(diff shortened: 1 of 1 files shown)


stsoe (approved):

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -153,6 +139,20 @@ struct patcher
...
+  {
+    static const std::array<std::string, static_cast<int>(buf_type::buf_type_count)> Section_Name_Array =
+      { ".ctrltext",
+        ".ctrldata",
+        ".preempt_save",
+        ".preempt_restore",
+        ".pdi",
+        ".ctrlpkt.pm"};
+
+    return Section_Name_Array[static_cast<int>(bt)];
```
stsoe, line 153: I think std::string_view is the more modern way of doing this these days.   It's efficient and supported by std::string::find.  Needless to say, need to include `<string_view>`.

```suggestion
    static const std::array<std::string_view, static_cast<int>(buf_type::buf_type_count)> section_name_array =
      { ".ctrltext",
        ".ctrldata",
        ".preempt_save",
        ".preempt_restore",
        ".pdi",
        ".ctrlpkt.pm"};

    return section_name_array[static_cast<int>(bt)];
```

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -817,7 +817,8 @@ class module_elf : public module_impl
         patcher::buf_type buf_type = patcher::buf_type::ctrltext;
 
         auto symbol_type = static_cast<patcher::symbol_type>(rela->r_addend);
-        std::string key_string = generate_key_string(argnm, buf_type);
+
+        const std::string key_string = generate_key_string(argnm, buf_type);
```
stsoe, line 821:
```suggestion
        auto key_string = generate_key_string(argnm, buf_type);
```

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -830,15 +831,15 @@ class module_elf : public module_impl
         if (auto search = arg2patcher.find(key_string); search != arg2patcher.end())
           search->second.m_ctrlcode_patchinfo.emplace_back(patcher::patch_info{ctrlcode_offset, 0, 0});
         else
-          arg2patcher.emplace(std::move(key_string), patcher{ symbol_type, {{ctrlcode_offset, 0}}, buf_type});
+          arg2patcher.emplace(key_string, patcher{symbol_type, {{ctrlcode_offset, 0}}, buf_type});
```
stsoe, line 834: key_string should be moved., why was this changed?, maybe because key_string was made `const`  which was not quite right.
```suggestion
          arg2patcher.emplace(std::move(key_string), patcher{symbol_type, {{ctrlcode_offset, 0}}, buf_type});
```

stsoe (changes requested):

stsoe (approved):

PR #8626, opened 2024-11-27
Made aie resources 0 for hw_gen greater than 1

bisingha-xilinx:
While requesting for AIE partition the meta_data of aie_partition_req should be "zero" for hw_gen greater than 1 since AIE resource loading only supported on hw_gen 1 (AIE1). Confirmed with the AIE driver team.
Customer faced issue with 2023.2 XRT (https://jira.xilinx.com/browse/CR-1220658). To avoid future similar issues fixing it on XRT master. 
making meta_data of aie_partition_req "zero" for hw_gen greater than 1.
None
Tested on vck190(AIE1) and vek280(AIE2)
None

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/edge/zocl_aie.c
+++ b/src/runtime_src/core/edge/drm/zocl/edge/zocl_aie.c
@@ -524,12 +524,12 @@ zocl_create_aie(struct drm_zocl_slot *slot, struct axlf *axlf, char __user *xclb
 		}
 	}
 
-	/* TODO figure out the partition id and uid from xclbin or PDI */
+	/* TODO figure out the uid from xclbin or PDI */
 	req.partition_id = partition_id;
 	req.uid = 0;
 	req.meta_data = 0;
 
-	if (aie_res)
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/edge/drm/zocl/edge/zocl_aie.c
```diff
@@ -524,12 +524,12 @@ zocl_create_aie(struct drm_zocl_slot *slot, struct axlf *axlf, char __user *xclb
...
 	}
 
-	/* TODO figure out the partition id and uid from xclbin or PDI */
+	/* TODO figure out the uid from xclbin or PDI */
 	req.partition_id = partition_id;
 	req.uid = 0;
 	req.meta_data = 0;
 
-	if (aie_res)
+	if (hw_gen == 1)
```
chvamshi-xilinx, line 532: this is not the right change , remove the condition completely

chvamshi-xilinx: Thanks for pointing out. it is already set at 530

PR #8634, opened 2024-11-30
Enabled control packet flow for next gen AIE devices

chvamshi-xilinx:
Enabled Control packet flow for next gen AIE devices. Currently this flow is supported only for AIE2 devices. 

This is a new feature. Enabled control packet flow for next gen AIE devices.

Patching control packet address in Instruction buffer, 
Created a new patching schema for new class of AIE devices,
Patching control packet with other buffers.

Low

Verified multiple cases with and without control packets

None


chvamshi-xilinx:
Thanks @rbramand-xilinx for providing initial changes.

stsoe (approved):
The functions in this file are getting much too long.
It may be time for a re-write.

chvamshi-xilinx:
There is small spec change , need to update this PR after aiebu is ready.

chvamshi-xilinx:
Infra issue on apu package build. It is not starting. Verified it locally.

PR #8639, opened 2024-12-03
Fix xrt-smi throughput tests to measure throughput

stsoe:
Fix throughput tests for xrt-smi to actually measure throughput.

NPU test for chained command throughput was measuring latency but computing throughput.   Additionally the test failed to hide the ramp-up clock scaling cost because iteration count was too low.

NPU test for individual command throughput was missing wait on last batch of commands submitted.

Fix chained command throughput to ensure NPU is always busy while measuring throughput.  This was done by using two xrt::runlists.

Additional problem for chained command throughput test was the iteration count, which was too small to hide the ramp-up cost of clock scaling, so the test was fixed to increase the iteration count and lowering the runlist command count.

Added wait on last batch of submitted commands in NPU throughput test for individual commands.

xrt-smi validate


AShivangi (approved):
Thanks for fixing the tests :)

PR #8642, opened 2024-12-05
Added AIE flag to build XRT edge debian packages on ubuntu 24.04

ManojTakasi:
We are encountering a build error while creating XRT edge debian packages using Ubuntu 24.04, even with XRT_AIE_BUILD=true. The AIE flag has been added properly to build XRT edge Debian packages on Ubuntu 24.04

We are using a same xrt package for both zynqmp and versal aie devices which requires AIE flag to be enabled by default.

Added XRT_AIE_BUILD flag to debian/rules file

Testd on kria kv260

### DIFF
--- a/build/debian/rules
+++ b/build/debian/rules
@@ -6,7 +6,7 @@
 override_dh_auto_configure:
 	set -ex; \
 	if [ "$$(dpkg-architecture -q DEB_TARGET_ARCH)" = arm64 ]; then \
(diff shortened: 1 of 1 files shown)


stsoe:
The PR description is missing the why.

ManojTakasi:
Hi Soren, I have added the information. Could you please review once again.

stsoe (approved):

PR #8648, opened 2024-12-09
Reorganize XRT header files under core/include

stsoe:
Under runtime_src/core/include we have a bunch of more or less random header files.  Some are completely internal to XRT and are not exported, but several are exported in an install target to XRT_INSTALL_INCLUDE_DIR.

This PR, while work-in-progress, moves all exported header files under `core/include/xrt`.   For legacy XRT installs, header wrappers are added in the old location(s) such that applications can continue to compile after the move..  It is strongly advised that application change to include headers from the new location(s), pragma messages will advise.

We cannot have the content of the top-level include directory exported into e.g. /usr/include,  so all the files have to be (re)moved or not exported. 

Move exported headers from -> to:

- `include/*.*  -> include/xrt/detail`
- `include/deprecated -> include/xrt/deprecated`
- `include/experimental -> include/xrt/experimental`

Create wrappers to preserve compilation of existing applications until they are changed.

Update driver sources for DKMS to copy the actual headers rather than the new wrappers.

Update all XRT source code to include from new location with the exception of code that is shared between
userspace and driver.  For these latter situations, a few header wrappers remain in `core/include` and are included from this location.  When DKMS builds sources the wrappers have been replaced with the actual files.

Risks are primarily edge builds, which I couldn't validate sufficiently.  This PR is bound to fail CI as is. and will be updated with help from others.

Manual build of legacy XRT application.


AShivangi (approved):
Looks good. We'll work on updating the includes in xrt-smi

mamin506 (approved):

PR #8649, opened 2024-12-10
Add changes to support 32 bit addend address in ELF patching

rbramand-xilinx:
Earlier rela dynamic sections addend fields higher 28 bits were used for holding base buffers addend address but 28 bits are not sufficient in all cases.
This change adds support for 32 bit addend were higher 4 bits (28 - 31) bits are picked form rela dynamic sections info field higher 4 bits.

Added support for 32 bit addend

Ideally moving to 64 bit ELF should solve the problem but it requires lot of changes from both aiebu and XRT side so this is planned for upcoming releases. Also other solutions which breaks existing ELFs are rejected.
This change doesn't cause any issue as an unused fields bits are used.

Low

Tested failing psq2 application and test passes with this fix

NA


chvamshi-xilinx (approved):

stsoe (approved):
This is getting out of hand, shifting, bit specific, long long functions that appear to be generic and not specific to a particular bit pattern.

larry9523 (changes requested):
Let's go with OSABI Version change and use the whole 32 bits addend as argplus.

chvamshi-xilinx:
> This is getting out of hand, shifting, bit specific, long long functions that appear to be generic and not specific to a particular bit pattern.

Yeah Soren. Discussed with Sonal aswell. He suggested to move all of this code to aiebu and XRT uses functions from it. It is going to take some time :-)

rbramand-xilinx:
> Let's go with OSABI Version change and use the whole 32 bits addend as argplus.

Hi @larry9523 made changes as discussed, please review

sonals (approved):

PR #8655, opened 2024-12-11
Disable install of aiebu and aie-rt

stsoe:
Improper install targets pollutes the XRT packages with xaiengine header and static libraries.  This PR disables the install target for aie-rt both via aiebu and directly as submodule used by XDP on NPU.

PR updates src/runtime_src/core/common/aiebu to include 
- https://github.com/Xilinx/aiebu/pull/35
- https://github.com/Xilinx/aiebu/pull/34

PR updates src/runtime_src/aie-rt to be in synch with what aiebu submodule expects after it was updated.  It also removes the [binary_dir] used in add_subdirectory for aie-rt. The binary_dir destination is not needed and not commonly used, it also interferes with the override to remove install artifacts, which was added in earlier commits in this PR.

Strenuous files in /opt/xilinx and in xrt packages

Add CMake function in src/CMake/utilities.cmake to disable the install target of added subdirectory.  The function leverages `CMAKE_SKIP_INSTALL_RULES` but additional works around a missing feature in CMake.

That someone relied on the exported content; very unlikely

Builds comparing install content. Both Linux and Windows and for MCDM.


sonals:
We do need to release aiebu files (the statically linked aiebu-asm and the header files) as part of XRT. We do **not need aie-rt** as it is not used by end users or model developers. However, aiebu artifacts are required by end users and model developers.

stsoe:
> We **do need to release aiebu** artifacts (the statically linked aiebu-asm and the header files) as part of XRT. We do **not need aie-rt** as it is not used by end users or model developers.
> 
> However, aiebu artifacts are required by end users and model developers.

Alright.  Thanks for this note. @sonals 
Could you please clarify what aiebu artifacts you are talking about?  And which package you want to release what.  Below files are contained in a component Runtime package (xrt_202510.2.19.0_22.04-amd64-Runtime.tar.gz), please suggest a better component name or if  what files should be contained in which component, I am thinking deployment and development here.  

We will need surgery to aiebu in order to prevent it from releasing aie-rt content.  Would have been nice had this been addressed when submodules were added; things don't just magically clean up by themselves :-(

opt/xilinx/aiebu/lib/libaiebu.so
opt/xilinx/aiebu/lib/libaiebu_static.a
opt/xilinx/aiebu/lib/aie2/preempt_restore_stx_4x1.bin
opt/xilinx/aiebu/lib/aie2/preempt_restore_stx_4x2.bin
opt/xilinx/aiebu/lib/aie2/preempt_restore_stx_4x4.bin
opt/xilinx/aiebu/lib/aie2/preempt_save_stx_4x1.bin
opt/xilinx/aiebu/lib/aie2/preempt_save_stx_4x2.bin
opt/xilinx/aiebu/lib/aie2/preempt_save_stx_4x4.bin
opt/xilinx/aiebu/lib/aie2/preempt_save_phx_4x1.bin
opt/xilinx/aiebu/lib/aie2/preempt_save_phx_4x2.bin
opt/xilinx/aiebu/lib/aie2/preempt_save_phx_4x4.bin
opt/xilinx/aiebu/lib/aie2/preempt_restore_phx_4x1.bin
opt/xilinx/aiebu/lib/aie2/preempt_restore_phx_4x2.bin
opt/xilinx/aiebu/lib/aie2/preempt_restore_phx_4x4.bin
opt/xilinx/aiebu/lib/aie2/copy_Mem_Tile_to_DDR.bin
opt/xilinx/aiebu/lib/aie2/copy_DDR_to_Mem_Tile.bin
opt/xilinx/aiebu/include/aiebu.h
opt/xilinx/aiebu/include/aiebu_assembler.h
opt/xilinx/aiebu/include/aiebu_error.h
opt/xilinx/aiebu/bin/aiebu-asm

sonals:
We need to weed out all the bin files for aiebu. But I agree with you that the CMake for aiebu should be updated so aiebu does not blindly release all the artifacts. Let me take a stab at it later this week. 

Maybe you can update this PR to disable aie-rt release and leave out the aiebu for now?

stsoe:
> Maybe you can update this PR to disable aie-rt release and leave out the aiebu for now?

I have to update aiebu to stop the release of aie-rt artifacts that come from aiebu.  Are we okay bumping aiebu submodule in XRT so I can include this change in aiebu?

jvillarre (approved):

PR #8657, opened 2024-12-13
[XDP] Handle incorrect range of tiles usage for memory tiles settings

vipangul:
[CR-1222763] Incorrect usage of range of memory tiles was resulting into segmentation faults.

CR-1222763 - UG1076 has documentation error and needs an update for  tile based settings of memory tiles.

Parsing of user settings are in try-catch block. However, this was skipped as invalid string to int conversion was not thrown and re-caught.

Low.

Verified on VEK280.
Verified that user now doesn't see crash and appropriate warning is emitted.
Verified that correct usage of memory tile settings are parsed correctly by plugin.

Yes. UG1076 needs an update.

### DIFF
--- a/src/runtime_src/xdp/profile/database/static_info/aie_util.cpp
+++ b/src/runtime_src/xdp/profile/database/static_info/aie_util.cpp
@@ -385,7 +385,11 @@ namespace xdp::aie {
    ***************************************************************************/
   uint8_t
   convertStringToUint8(const std::string& input) {
-    return static_cast<uint8_t>(std::stoi(input));
+    try {
+      return static_cast<uint8_t>(std::stoi(input));
+    } catch (const std::exception&) {
+      throw;
+    }
   }
 
   std::string


src/runtime_src/xdp/profile/database/static_info/aie_util.cpp
```diff
@@ -385,7 +385,11 @@ namespace xdp::aie {
    ***************************************************************************/
   uint8_t
   convertStringToUint8(const std::string& input) {
-    return static_cast<uint8_t>(std::stoi(input));
+    try {
+      return static_cast<uint8_t>(std::stoi(input));
+    } catch (const std::exception&) {
+      throw;
+    }
```
stsoe, line 392: This change is a no-op.  It has no effect.

vipangul:
This seem to have resolved the issue in my last testing. I'll re-investigate and re-open this PR with more findings for reported CR.

PR #8658, opened 2024-12-14
Added 6.12 kernel support to the zocl

ManojTakasi:
https://jira.xilinx.com/browse/CR-1222615 Added 6.12 kernel support to the zocl. In 6.12 kernel remove funciton pointer is changed from to void ([Reference](https://lxr.bootlin.com/linux/v6.12.4/source/include/linux/platform_device.h#L236)). Fixed compiler warnings.
changed the function type by adding #if #endif directives.
low
Tested by building with both 6.6 and 6.12 kernels

### DIFF
--- a/src/runtime_src/core/edge/drm/zocl/common/cu.c
+++ b/src/runtime_src/core/edge/drm/zocl/common/cu.c
@@ -9,7 +9,6 @@
  * This file is dual-licensed; you may select either the GNU General Public
  * License version 2 or Apache License, Version 2.0.
  */
-
 #include "zocl_drv.h"
 #include "xrt_cu.h"
 #include "zocl_ert_intc.h"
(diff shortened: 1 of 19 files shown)


src/runtime_src/core/edge/drm/zocl/common/cu.c
```diff
@@ -445,18 +444,22 @@ static int cu_probe(struct platform_device *pdev)
 	kfree(zcu);
 	return err;
 }
-
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 12, 0)
+static void cu_remove(struct platform_device *pdev)
```
chvamshi-xilinx, line 448: why? Lets not change the return type. Instead of that, return proper result in this function.
Am i missing anything?

ManojTakasi: Hi Vamshi, In kernel version 6.12 and newer, the return type is void, whereas in the older kernels like 6.6 it's int. To maintain backward compatibility, I have added a macro.

chvamshi-xilinx (approved):

src/runtime_src/core/edge/drm/zocl/common/cu.c
chvamshi-xilinx: Got it. Thanks Manoj

PR #8660, opened 2024-12-15
Fix bug in patching args in multiple runs

rbramand-xilinx:
When kernel is run multiple times and in some of the runs if arg is updated using set_arg the present code patches the arg address on top of existing already patched value this will result in test failures. This PR fixes this issue

Storing BD data ptr entries in patch info and marking it dirty after first patching.
When second or subsequent run is made with set_arg the original BD data ptr values stored are restored and we patch on top of it which is proper.

Moderate as this code changes xrt code critical path

Tested xrt elf flow test on aie2p by calling set_arg in multiple kernel runs and test passes
Also tested vadd application on aie2ps by modifying host application accordingly.
Thorough testing is needed to make sure this change doesn't break all the existing flows

NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -137,10 +137,13 @@ struct patcher
   buf_type m_buf_type = buf_type::ctrltext;
   symbol_type m_symbol_type = symbol_type::shim_dma_48;
 
+  static constexpr size_t num_bd_data_ptrs = 8; // max number of bd ptrs accessed while patching
   struct patch_info {
     uint64_t offset_to_patch_buffer;
     uint32_t offset_to_base_bo_addr;
     uint32_t mask; // This field is valid only when patching scheme is scalar_32bit_kind
+    bool dirty = false; // Tells whether this entry is already patched or not
+    uint32_t bd_data_ptrs[num_bd_data_ptrs]; // array to store bd ptrs original values
   };
 
   std::vector<patch_info> m_ctrlcode_patchinfo;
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -137,10 +137,13 @@ struct patcher
   buf_type m_buf_type = buf_type::ctrltext;
   symbol_type m_symbol_type = symbol_type::shim_dma_48;
 
+  static constexpr size_t num_bd_data_ptrs = 8; // max number of bd ptrs accessed while patching
   struct patch_info {
```
sonals, line 141: Can we add a comment on why **8** and make this a file constant?

rbramand-xilinx: Made changes accordingly Sonal, Thanks

chvamshi-xilinx (approved):
I know, this file is not following other XRT user space style of coding, Please update name of variable to lower case. "max_bd_words"

larry9523 (commented):
This fix introduces a new dependency between xrt_core and AIE architecture. Ideally, all patching knowledges should come from ELF.

But given the plan that we will move all the patching service out of XRT to aiebu, I think this temporal solution is OK.

sonals (approved):

larry9523:
@rbramand-xilinx  please provide a list of tests we have done before we are confident enough to merge this PR

rbramand-xilinx:
> @rbramand-xilinx please provide a list of tests we have done before we are confident enough to merge this PR

Hi @larry9523 , I have tested xrt flow test on aie2p and vadd test on aie2ps by modifying host application and test passes on both of them.

rbramand-xilinx:
> > @rbramand-xilinx please provide a list of tests we have done before we are confident enough to merge this PR
> 
> Hi @larry9523 , I have tested xrt flow test on aie2p and vadd test on aie2ps by modifying host application and test passes on both of them.

@larry9523 I have verified non control packet , control packet elf flow tests and also full elf (xclbin2elf) flows and patching args in multiple runs tests passes only after applying this change.
@sayyanna  is verifying preemption test, if that test passes I guess we are good to go.

sayyanna:
> > > @rbramand-xilinx please provide a list of tests we have done before we are confident enough to merge this PR
> > 
> > 
> > Hi @larry9523 , I have tested xrt flow test on aie2p and vadd test on aie2ps by modifying host application and test passes on both of them.
> 
> @larry9523 I have verified non control packet , control packet elf flow tests and also full elf (xclbin2elf) flows and patching args in multiple runs tests passes only after applying this change. @sayyanna is verifying preemption test, if that test passes I guess we are good to go.

Verified preemption test with patching args multiple times and tests passes only after applying this change.

larry9523 (approved):

PR #8669, opened 2024-12-20
CR-1214907 Modified strcture data element from fixed size to flexible array.

ManojTakasi:
https://jira.xilinx.com/browse/CR-1214907 We are getting UBSAN: array-index-out-of-bounds call trace while compiling xocl xclmgmt drivers in ubuntu 24.04

Reference: https://www.kernel.org/doc/html/latest/process/deprecated.html#zero-length-and-one-element-arrays

Added macros to pick flexible arrays when comipiling kernel code. 

Modifications are in common files and will affect all platforms.

Tested on u250 by running xrt-smi examine. Ran vadd and graph testcases on vck190 .

NA


stsoe (approved):
The changes to ert.h are exactly what flexible arrays should be all over the place. Iit's so confusing to have `data[1]`, but unfortunately much code in user space accounts for the one default element; so the guard to enable for kernel space only is good.

chvamshi-xilinx (approved):

PR #8671, opened 2024-12-20
Addition of query for xrt_smi_configuration.json

aktondak:
This change adds a new query object to shim layer for drivers to provide the path for xrt_smi_configuration.json.
This structure can/should be augmented for future jsons we plan to add.

https://jira.xilinx.com/browse/VITIS-14245

The problem was solved by adding a new structure to query objects

N/A

Tested on Linux xdna

N/A

### DIFF
--- a/src/runtime_src/core/common/query_requests.h
+++ b/src/runtime_src/core/common/query_requests.h
@@ -55,6 +55,7 @@ enum class key_type
   instance,
   edge_vendor,
   device_class,
+  xrt_smi_config,
   xclbin_name,
   sequence_name,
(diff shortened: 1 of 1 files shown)


stsoe (approved):
Looks good to me.  Not really sure what the enum class type enumerators signify?

aktondak:
> Looks good to me. Not really sure what the enum class type enumerators signify?

The idea is to keep this query struct extendible to other xrt-smi specific JSONs that might come in future. Though there's just one today. We discussed on possibly making validate tests json based as well.

PR #8680, opened 2025-01-06
CR-1224774 "xbutil examine -r aie" command does not work for AIE partitions

ManojTakasi:
https://jira.xilinx.com/browse/CR-1224774   We are getting a 'node is not present' exception when we run "xrt-smi examine -r aie " command.

We are trying to read the whole AIE partition array (0_50) even when the directory itself is not present, which is not correct when multiple partitions are created with hw_ctx.

Hardcoded the AIE paritions path and read all the AIE partitions present in that directory, which helps to read all the partitions when multiple AIE xclbins are present.

low
Tested on vck190 with both device specific implementation(which uses whole array) and  hw_ctx (which uses partition of an array) implementation.

NA

### DIFF
--- a/src/runtime_src/core/edge/user/device_linux.cpp
+++ b/src/runtime_src/core/edge/user/device_linux.cpp
@@ -15,9 +15,11 @@
 #include "core/edge/user/aie/aie_buffer_object.h"
 #endif
 #include "core/edge/user/aie/profile_object.h"
+#include <filesystem>
 #include <map>
 #include <memory>
 #include <poll.h>
+#include <regex>
 #include <string>
 
 #include <fcntl.h>
@@ -216,14 +218,30 @@ struct aie_core_info_sysfs
   {
     boost::property_tree::ptree ptarray;
     aie_metadata_info aie_meta = get_aie_metadata_info(device);
-    const std::string aiepart = std::to_string(aie_meta.shim_row) + "_" + std::to_string(aie_meta.num_cols);
-    const aie_sys_parser asp(aiepart);
+    auto base_path = "/sys/class/aie/";
+    std::regex pattern(R"(aiepart_(\d+)_(\d+))");
 
-    /* Loop each all aie core tiles and collect core, dma, events, errors, locks status. */
-    for (int i = 0; i < aie_meta.num_cols; ++i)
-      for (int j = 0; j < (aie_meta.num_rows-1); ++j)
-        ptarray.push_back(std::make_pair(std::to_string(i) + "_" + std::to_string(j),
-                          asp.aie_sys_read(i,(j + aie_meta.core_row))));
+    for (const auto& entry : std::filesystem::directory_iterator(base_path)) {
+        if (!entry.is_directory())
+            continue;
+
+        std::string dir_name = entry.path().filename().string();
+        std::smatch matches;
+
+        if (!std::regex_match(dir_name, matches, pattern))
+            continue;
+
+        auto start_col = std::stoi(matches[1].str());
+        auto num_col = std::stoi(matches[2].str());
+        auto aiepart = std::to_string(start_col) + "_" + std::to_string(num_col);
+        const aie_sys_parser asp(aiepart);
+
+        /* Loop each all aie core tiles and collect core, dma, events, errors, locks status. */
+        for (int i = start_col; i < (start_col + num_col); ++i)
+          for (int j = 0; j < (aie_meta.num_rows - 1); ++j)
+            ptarray.push_back(std::make_pair(std::to_string(i) + "_" + std::to_string(j),
+                              asp.aie_sys_read(i,(j + aie_meta.core_row))));
+    }
 
     boost::property_tree::ptree pt;
     pt.add_child("aie_core",ptarray);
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -216,14 +218,28 @@ struct aie_core_info_sysfs
...
-    for (int i = 0; i < aie_meta.num_cols; ++i)
-      for (int j = 0; j < (aie_meta.num_rows-1); ++j)
-        ptarray.push_back(std::make_pair(std::to_string(i) + "_" + std::to_string(j),
-                          asp.aie_sys_read(i,(j + aie_meta.core_row))));
+    std::string base_path = "/sys/class/aie/";
+    std::regex pattern(R"(aiepart_(\d+)_(\d+))");
+    int start_col=0, num_col=0;
+    for (const auto& entry : std::filesystem::directory_iterator(base_path)) {
+        if (entry.is_directory()) {
+            std::string dir_name = entry.path().filename().string();
```
stsoe, line 226: Please use early exit to avoid deep indentation.
```suggestion
        if (!entry.is_directory())  
          continue;
```

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -216,14 +218,28 @@ struct aie_core_info_sysfs
...
+            std::smatch matches;
+            if (std::regex_match(dir_name, matches, pattern)) {
+                start_col = std::stoi(matches[1].str());
+                num_col = std::stoi(matches[2].str());
+
+                const std::string aiepart = std::to_string(start_col) + "_" + std::to_string(num_col);
```
stsoe, line 232:
```suggestion
                auto aiepart = std::to_string(start_col) + "_" + std::to_string(num_col);
```

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -216,14 +218,28 @@ struct aie_core_info_sysfs
...
+                const aie_sys_parser asp(aiepart);
+
+                /* Loop each all aie core tiles and collect core, dma, events, errors, locks status. */
+                for (int i = start_col; i < (start_col + num_col); ++i)
+                  for (int j = 0; j < (aie_meta.num_rows-1); ++j)
```
stsoe, line 237:
```suggestion
                  for (int j = 0; j < (aie_meta.num_rows -1 ); ++j)
```

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -216,14 +218,28 @@ struct aie_core_info_sysfs
...
-    const aie_sys_parser asp(aiepart);
-
-    /* Loop each all aie core tiles and collect core, dma, events, errors, locks status. */
-    for (int i = 0; i < aie_meta.num_cols; ++i)
-      for (int j = 0; j < (aie_meta.num_rows-1); ++j)
-        ptarray.push_back(std::make_pair(std::to_string(i) + "_" + std::to_string(j),
-                          asp.aie_sys_read(i,(j + aie_meta.core_row))));
+    std::string base_path = "/sys/class/aie/";
+    std::regex pattern(R"(aiepart_(\d+)_(\d+))");
+    int start_col=0, num_col=0;
```
stsoe, line 223: I can't quite tell, but these look like local variables?
```suggestion
```

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -216,14 +218,28 @@ struct aie_core_info_sysfs
...
+    for (const auto& entry : std::filesystem::directory_iterator(base_path)) {
+        if (entry.is_directory()) {
+            std::string dir_name = entry.path().filename().string();
+            std::smatch matches;
+            if (std::regex_match(dir_name, matches, pattern)) {
+                start_col = std::stoi(matches[1].str());
+                num_col = std::stoi(matches[2].str());
```
stsoe, line 230: I think these are local variables:
```suggestion
                auto start_col = std::stoi(matches[1].str());
                auto num_col = std::stoi(matches[2].str());
```

stsoe (changes requested):

ManojTakasi:
Hi Soren,

Thank you for your suggestions. I have updated the code. Please let me know if I need to make any further changes.

chvamshi-xilinx (approved):

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -216,14 +218,28 @@ struct aie_core_info_sysfs
...
+    auto base_path = "/sys/class/aie/";
+    std::regex pattern(R"(aiepart_(\d+)_(\d+))");
+
+    for (const auto& entry : std::filesystem::directory_iterator(base_path)) {
+        if (!entry.is_directory())
+            continue;
+
+        std::string dir_name = entry.path().filename().string();
+        std::smatch matches;
+        if (std::regex_match(dir_name, matches, pattern)) {
```
stsoe, line 230:
```suggestion
        if (!std::regex_match(dir_name, matches, pattern))
          continue;
          
```

stsoe (changes requested):
One more :-)

src/runtime_src/core/edge/user/device_linux.cpp
ManojTakasi: Missed it. I have updated the code. Thank you

stsoe (approved):
Thank you!

PR #8686, opened 2025-01-09
VITIS-13807 : Use opencl icd dispatcher instead of ocl icd for embedded flows

rbramand-xilinx:
Moved to Opencl icd dispatcher from ocl icd dispatcher for embedded platforms as yocto is deprecating the use of ocl icd flows.

Yocto team filed story VITIS-13807

The best fix is to move all the flows to opencl icd way but the problem is all the linux distributions (RHEL, CentOS) doesn't ship packages to support opencl icd flow, so only made changes to embedded flow.
Also this change is made under temporary flag OPENCL_ICD_LOADER so that XRT builds doesn't break till changes in XRT recipe related to opencl icd loader in yocto/petalinux get into TA. This flag will be removed soon.

Low as both ocl icd and opencl icd are almost similar and the only change is structure cl_icd_dispatch members differ a little which we already dont use (we already fill zeros as those APIs are not supported in our flow).

Tested Embedded flows build and OpenCL test cases.
Tested x86 build.

Most likely no as user doesn't need to understand about the dispatcher we use.

### DIFF
--- a/src/runtime_src/xocl/api/icd/ocl_icd_bindings.h
+++ b/src/runtime_src/xocl/api/icd/ocl_icd_bindings.h
@@ -33,7 +33,7 @@ using cl_icd_dispatch = KHRicdVendorDispatchRec;
 #else
 // All x86 linux distros doesn't have opencl icd dispatcher
 // support so using ocl icd dispatcher for x86 flows.
-# include "ocl_icd.h"
+# include <ocl_icd.h>
 using cl_icd_dispatch = _cl_icd_dispatch;
 #endif
 #endif
--- a/src/runtime_src/xocl/api/icd/ocl_icd_bindings.h
+++ b/src/runtime_src/xocl/api/icd/ocl_icd_bindings.h
@@ -23,10 +23,20 @@
 #ifdef _WIN32
 # define NOMINMAX
 # include "windows/icd_dispatch.h"
-using _cl_icd_dispatch = KHRicdVendorDispatchRec;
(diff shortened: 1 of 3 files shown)


src/runtime_src/xocl/api/icd/ocl_icd_bindings.h
```diff
@@ -23,10 +23,20 @@
...
-# include <ocl_icd.h>
+#if (defined (__aarch64__) || defined (__arm__)) && defined (OPENCL_ICD_LOADER)
+// In Yocto ocl icd dispatcher is deprecated and
+// opencl icd dispatcher is the recommended one.
+// Using opencl icd dispatcher for embedded flows.
+# include <CL/cl_icd.h>
+#else
+// All x86 linux distros doesn't have opencl icd dispatcher
+// support so using ocl icd dispatcher for x86 flows.
+# include "ocl_icd.h"
```
stsoe, line 36: Can we indicate system header with angle brackets as it used to be?
```suggestion
# include <ocl_icd.h>
```

stsoe (approved):
LGTM.  One change request.

src/runtime_src/xocl/api/icd/ocl_icd_bindings.h
rbramand-xilinx: sure Soren

chvamshi-xilinx (approved):

PR #8688, opened 2025-01-11
Support AIE columns with multiple controllers

stsoe:
When extracting MicroBlaze control code from ELF for a multi-column partition, account for each column having multiple controllers and that a partition may not use all of these.

ELF sections identify which column controllers are used.  Currently this done through an ELF section naming convention.

Before this change, each column had exactly one controller and it was assumed that either the lead controller in a partition was used alone or with all remaining controllers in subsequent columns. With this PR, support is added for multiple controllers in each column and for only a subset of these controllers being used.

The ERT packet payload interpretation for ert_dpu_data is changed to indicate which column processor the data is associated with.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -705,7 +705,7 @@ class module_elf : public module_impl
     // is asserted)
     std::vector<ctrlcode> ctrlcodes;  // indexed by uC index
     static_assert(std::is_same_v<decltype(uc_sections), std::map<uc_index, elf_sections>>, "fix std::map assumption");
-    ctrlcodes.resize(uc_sections.empty() ? 0 : uc_sections.rbegin()->first);
+    ctrlcodes.resize(uc_sections.empty() ? 0 : uc_sections.rbegin()->first + 1);
     pad_offsets.resize(ctrlcodes.size());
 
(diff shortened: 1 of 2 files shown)


stsoe:
@xuhz please review

maxzhen (approved):

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -643,78 +653,83 @@ class module_elf : public module_impl
...
+    // requirement, then pad to page size for page of a column so that
+    // embedded processor can load a page at a time.  Note, that not
+    // all column uC need be used, so account for holes in
+    // uc_sections.  Leverage that uc_sections is a std::map and that
+    // std::map stores its elements in ascending order of keys (this
+    // is asserted)
+    std::vector<ctrlcode> ctrlcodes;  // indexed by uC index
+    static_assert(std::is_same_v<decltype(uc_sections), std::map<uc_index, elf_sections>>, "fix std::map assumption");
+    ctrlcodes.resize(uc_sections.empty() ? 0 : uc_sections.rbegin()->first);
+    pad_offsets.resize(ctrlcodes.size());
```
xuhz, line 709: i believe 'pad_offsets' should also be resized to the same size to ctrlcodes

xuhz (approved):

PR #8692, opened 2025-01-14
Modified the aie-rt function APIs and updated the datatype of its parameter from u8 to u16

ManojTakasi:
This PR is the same as https://github.com/Xilinx/XRT/pull/8684. Reverted the changes on the client side. 

No functionality changes.

By modifying the mentioned APIs and its parameter from u8 to u16.

low

Ran graph test case by building and installing xrt,czocl, aiefal, libaiengine packages on vck190.


pgschuey (changes requested):
This looks like a much better set of changes as it avoids client-based code which already supports 16 bits. My only concern is this PR conflicts with PR 8689. Given that this is now the third set of changes which the other PR has had to merge/accommodate, I would suggest waiting for PR 8689 to merge, then make the appropriate adjustments to this PR and merge accordingly.

ManojTakasi:
Hi, the required changes have already been made in the mentioned PR-8689. Therefore, I will be closing this PR. Thank you

PR #8695, opened 2025-01-15
Build XDP depending on XRT component being built

stsoe:
Change how XDP is built in XRT.  Infer XDP build options based on which component is being built.

XDP is included in the npu and alveo components and the in the legacy xrt component, but it is being built differently depending on the component. XDP is excluded from XRT's base component.

This PR builds on #8691.  It includes XDP in `npu`, `alveo`, and legacy `xrt` components, but excludes XDP from `base`. Compile flags for XDP are based on the component. XDP_CLIENT_BUILD_CMAKE is now an internal  configuration variable for XDP.

There are no visible changes in this PR.  However, anything invoking CMake with XDP_CLIENT_BUILD_CMAKE will be affected and must change to ensure that XRT is building the NPU component by adding `-DXRT_NPU` when invoking CMake.

### DIFF
--- a/build/build-win22.sh
+++ b/build/build-win22.sh
@@ -95,7 +95,6 @@ while [ $# -gt 0 ]; do
 	-npu)
             shift
 	    npu_build=1
-	    cmake_flags+=" -DXDP_CLIENT_BUILD_CMAKE=yes"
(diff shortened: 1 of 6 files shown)


IshitaGhosh (approved):
Changes look good.

PR #8699, opened 2025-01-16
Remove some wrong header files included in three files.

zhangchiming:
Problem solved by the commit
removed wrong header files included.

N/A

Compiled and tested on both Windows and Ubuntu 22.04.

Documentation impact (if any)
None.


stsoe (approved):
Thanks @zhangchiming

PR #8708, opened 2025-01-22
VITIS-13806 Fix return codes for xrt-smi

rchane:
https://jira.xilinx.com/browse/VITIS-13806
Error codes returned by xrt-smi were being masked by loader and helper scripts such that return code was always zero.
After discussion, we decided we no longer even needed the helper scripts. Similar to when the xbutil wrapper was removed on Windows for MCDM, the relevant helper scripts for xrt-smi were removed from Windows and Linux. In addition, xrt-smi has been moved up one directory to the bin directory.
I will need to contact other teams regarding the removal of the scripts/the movement of xrt-smi for when they package xrt-smi with MCDM driver.
Tested on Linux, as desired, one can see the correct error code again.
```
$ xrt-smi configure --pmode performance
ERROR: Please specify a device using --device option
 Available devices:

$ echo $?
1
```
N/A


AShivangi:
Please test this on MCDM setup. We will have to probably change the inf file as well. This probably needs to be propogated to mcdm_sw_stack as well

rchane:
I have reached out to Akshata and confirmed the build on Windows MCDM. Akshata will take care of things on mcdm sw stack side!

AShivangi (approved):

stsoe:
Hey @rchane @AShivangi.  Can we really get rid of the loader scripts?  At least on Linux, xrt-smi may be invoked directly as `/opt/xilinx/xrt/bin/xrt-smi`.  This will not work unless LD_LIBRARY_PATH, XILINX_XRT are set.  The point of the loader was not only to intelligently decide between xbutil and xbutil2, but more importantly to not require sourcing a setup script.  Of course of windows we do not need a loader as xrt-smi.exe will be able to find xrt_coreutil in system32/.

rchane:
Hi @stsoe , if I understand correctly, we currently ask people to source the setup script on Linux in xdna-driver. Is this fine?

stsoe:
> Hi @stsoe , if I understand correctly, we currently ask people to source the setup script on Linux in xdna-driver. Is this fine?

Maybe, but AFAI can see,  the change in this PR breaks the use-case I gave.   I think it is reasonable to require that any tools we install, must be able to execute without any required setup.

rchane:
> > Hi @stsoe , if I understand correctly, we currently ask people to source the setup script on Linux in xdna-driver. Is this fine?
> 
> Maybe, but AFAI can see, the change in this PR breaks the use-case I gave. I think it is reasonable to require that any tools we install, must be able to execute without any required setup.

I see, thank you for the insight, Soren! I'll look into possible alternate solutions.

rchane:
Apologies for the extra review requests; I messed up my rebase. Please ignore!!!

PR #8714, opened 2025-01-23
xrt-smi rearchitecture Patch 4

aktondak:
This PR does the following :
1. Replaces embedded json configuration to a generated configuration. This generation code is common to all shims and each shim will derive from it to define whatever custom xrt-smi behavior they want to have.
2. Adds support for allowing multi-configuration Description field for xrt-smi options (through description_array). With this, the test and report list are simple vector stores and formatting is taken care of by XRT correctly.
3. Builds SubCmdJsonObjects only with xbutil and xbmgmt objects only as it should be.
4. Also fixes CR-1226870. This issue was caused due to incorrect configuration in default device cases.

Problem was discovered when we wanted to scale to multiple different tests separately in each shim.
https://jira.xilinx.com/browse/CR-1226870

Solved through addition of a new field.

xrt-smi should be tested for alveo devices.

Tested with MCDM and XDNA drivers.

None


AShivangi (approved):

src/runtime_src/core/common/smi.cpp
```diff
@@ -0,0 +1,263 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+#include "smi.h"
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/json_parser.hpp>
+#include <vector>
+#include <tuple>
+#include <string>
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/json_parser.hpp>
```
stsoe, line 10: Sort in groups.  Keep smi.h first, then boost includes and finally system headers

src/runtime_src/core/common/smi.cpp
```diff
@@ -0,0 +1,263 @@
...
+         const std::string& default_value, 
+         const std::string& value_type, 
+         const std::vector<basic_option>& description_array = {})
+      : basic_option{name, description, type}, 
+        alias(alias), 
+        default_value(default_value), 
+        value_type(value_type), 
+        description_array(description_array) {}
+
+  ptree to_ptree() const {
```
stsoe, line 111: Like it or not.  The convention I prefer to keep under core/common is
```
return-type
function() [const] 
{
}
```

src/runtime_src/core/common/smi.h
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+
+#pragma once
+#define XRT_CORE_COMMON_SOURCE
```
stsoe, line 5: Move to smi.cpp

stsoe (changes requested):

src/runtime_src/core/common/smi.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+// 3rd Party Library - Include Files
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/json_parser.hpp>
+
+#include <vector>
+#include <tuple>
+#include <string>
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/json_parser.hpp>
```
stsoe, line 17:
```suggestion
```

src/runtime_src/core/common/smi.cpp
```diff
@@ -0,0 +1,234 @@
...
+// Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+
+#define XRT_CORE_COMMON_SOURCE
+
+// Local - Include Files
+#include "smi.h"
+
+// 3rd Party Library - Include Files
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/json_parser.hpp>
```
stsoe, line 11: sort

src/runtime_src/core/common/smi.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+#include <vector>
+#include <tuple>
+#include <string>
```
stsoe, line 15: sort

src/runtime_src/core/common/smi.cpp
```diff
@@ -0,0 +1,234 @@
...
+#include <boost/property_tree/ptree.hpp>
+#include <boost/property_tree/json_parser.hpp>
+#include <iostream>
```
stsoe, line 18: I don't think iostream is needed?  But maybe <sstream>

src/runtime_src/core/common/smi.h
```diff
@@ -0,0 +1,72 @@
...
+
+#include <boost/property_tree/ptree.hpp>
+
+#include <string>
+#include <vector>
+#include <tuple>
+
+namespace xrt_core::smi {
+
+using TupleVector = std::vector<std::tuple<std::string, std::string, std::string>>;
```
stsoe, line 16: No camel case
```suggestion
using tuple_vector_type = std::vector<std::tuple<std::string, std::string, std::string>>; 
```

src/runtime_src/core/common/smi.h
```diff
@@ -0,0 +1,72 @@
...
+         const std::string& description,
+         const std::string& type, 
+         const std::string& default_value, 
+         const std::string& value_type, 
+         const std::vector<basic_option>& description_array = {})
+      : basic_option{name, description, type}, 
+        alias(alias), 
+        default_value(default_value), 
+        value_type(value_type), 
+        description_array(description_array) {}
```
stsoe, line 41: Any values copied into class data members should be passed by value and `std::move` d to the data members.  In this case it goes for all parameters to option constructor

src/runtime_src/core/common/smi.h
```diff
@@ -0,0 +1,72 @@
...
+protected:
+  virtual const TupleVector& get_validate_test_desc() const;
+  virtual const TupleVector& get_examine_report_desc() const;
+
+  std::vector<basic_option> construct_run_option_description() const;
+  std::vector<basic_option> construct_report_option_description() const;
+
+  boost::property_tree::ptree construct_validate_subcommand() const;
+  boost::property_tree::ptree construct_examine_subcommand() const;
+  boost::property_tree::ptree construct_configure_subcommand() const;
```
stsoe, line 64: I would prefer to see this as
```
virtual return_type
function() const;
```
I say this because code reading is easier when the important stuff is aligned to the left.

```
virtual const tuple_vector_type&
get_validate_test_desc() const;

std::vector<basic_option>
cosntruct_run_option_description() const;
```

src/runtime_src/core/common/smi.h
```diff
@@ -0,0 +1,72 @@
...
+  boost::property_tree::ptree to_ptree() const;
+};
+
+/*
+* Each shim's smi class derives from this class
+* and adds its custom functionalities. Currently only validate tests and examine
+* reports differ between each shim but going forward, each shim can define its 
+* custom behavior for xrt-smi as required. This also gives us the flexibility
+* to add device specific xrt-smi behavior.
+*/
```
stsoe, line 52: Good comment, but for C++ comments prefer `//`

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -678,6 +679,29 @@ struct am_counter
...
+struct xrt_smi_config
+{
+  using result_type = std::any;
+
+  static result_type
+  get(const xrt_core::device* device, key_type key, const std::any& reqType)
+  {
+    if (key != key_type::xrt_smi_config)
+      throw xrt_core::query::no_such_key(key, "Not implemented");
+    std::string xrt_smi_config;
```
stsoe, line 691: Blank line between single statement expression and following code.
```suggestion
    if (key != key_type::xrt_smi_config)
      throw xrt_core::query::no_such_key(key, "Not implemented");
      
    std::string xrt_smi_config;
```

src/runtime_src/core/edge/user/smi.cpp
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+#include "smi.h"
+
+namespace shim_edge::smi {
+
+const std::vector<std::tuple<std::string, std::string, std::string>>& 
+smi_edge::get_validate_test_desc() const {
+  static const std::vector<std::tuple<std::string, std::string, std::string>> validate_test_desc = {
```
stsoe, line 9: Generally all our code uses a new line between function and curly brace.
```
return_type
class::
fcn() const
{
}
```

src/runtime_src/core/edge/user/smi.h
```diff
@@ -0,0 +1,17 @@
...
+
+#include <string>
+#include "core/common/smi.h"
+
+namespace shim_edge::smi {
+
+class smi_edge : public xrt_core::smi::smi_base {
+protected:
+  const std::vector<std::tuple<std::string, std::string, std::string>>& get_validate_test_desc() const override;
+  const std::vector<std::tuple<std::string, std::string, std::string>>& get_examine_report_desc() const override;
```
stsoe, line 12:
```suggestion
  const std::vector<std::tuple<std::string, std::string, std::string>>& 
  get_validate_test_desc() const override;
  
  const std::vector<std::tuple<std::string, std::string, std::string>>& 
  get_examine_report_desc() const override;
```

stsoe (changes requested):
Some style change requests, else looks good.

src/runtime_src/core/common/smi.h
aktondak: Updated with latest patch. Dont know why this comment doesn't show up on the main PR screen :/

stsoe (approved):
Looks good.  Thanks for making the changes.

PR #8715, opened 2025-01-24
[XDP] fixed compile bug in MDM latency calculation

pgschuey:
Due to potential overflows, latency values from MDM need to be 64 bits

Compilation bug on clients introduced in PR-8702

Make latency values use 64 bits

None

Compiling on clients

None


vipangul (approved):
MCDM Build : PASS

PR #8716, opened 2025-01-24
VITIS-13806 Fix helper scripts and return codes for xrt-smi

rchane:
https://jira.xilinx.com/browse/VITIS-13806
The xrt-smi and xbutil helper scripts weren't trying to get the loader and returned 0 due to a faulty if check.
(e.g. `bin` is still in the path on Windows MCDM and Linux amdxdna, so we should still be finding the loader)
Fixed the if check. This should also preserve the original return behavior for edge platforms.
N/A
Tested/built on amdxdna and MCDM.
```
$ xrt-smi configure --pmode performance
ERROR: Please specify a device using --device option
 Available devices:

$ echo $?
1
```
N/A

### DIFF
--- a/src/runtime_src/core/tools/xbutil2/xrt-smi
+++ b/src/runtime_src/core/tools/xbutil2/xrt-smi
@@ -35,7 +35,7 @@ XRT_LOADER_DIR="`dirname \"$0\"`"
 
 # For edge platforms loader is not required as tools are in standard location(/usr).
 # So calling unwrapped tool from this script itself.
-if [[ $XRT_LOADER_DIR =~ "/usr" || $XRT_LOADER_DIR =~ "/bin" ]]; then
+if [[ $XRT_LOADER_DIR =~ ^(/usr|/bin) ]]; then
     "${XRT_LOADER_DIR}/unwrapped/${XRT_PROG}" "${XRTWRAP_PROG_ARGS[@]}"
     exit 0
 fi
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/tools/xbutil2/xrt-smi
```diff
@@ -35,7 +35,7 @@ XRT_LOADER_DIR="`dirname \"$0\"`"
 
 # For edge platforms loader is not required as tools are in standard location(/usr).
 # So calling unwrapped tool from this script itself.
-if [[ $XRT_LOADER_DIR =~ "/usr" || $XRT_LOADER_DIR =~ "/bin" ]]; then
+if [[ $XRT_LOADER_DIR =~ "/usr" ]]; then
```
stsoe, line 38: Hmm, doesn't this break what you fixed in #8426 ?

rchane: Ah, yes, I'll need to discuss with Rahul on how to fix this for them.

AShivangi (approved):

rbramand-xilinx (approved):
Thanks for making the changes, checking if the string starts with /usr or /bin works with embedded platforms.

chvamshi-xilinx (changes requested):
You cannot remove xbutil as it is still being used in edge platforms. There were concerns in Ops review last week, We still need to keep xbutil for one more release.

AShivangi (approved):

rchane:
> You cannot remove xbutil as it is still being used in edge platforms. There were concerns in Ops review last week, We still need to keep xbutil for one more release.

Thank you, Vamshi; updated the PR and description accordingly.

chvamshi-xilinx (approved):

PR #8721, opened 2025-01-28
Dynamic trace support in ve2

rbramand-xilinx:
Enabled support for dumping dynamic trace. This feature is supported in Telluride only for now.
Trace is dumped only when ini option is enabled.

Input is a control script based on which trace info will be filled. Spec for same can be found at - https://confluence.amd.com/pages/viewpage.action?pageId=1665323347

NA

Trace is dumped into a file only when required ini options are enabled.

Also we are calling APIs from dtrace library and the library path is taken as an ini option, ideally the library code should be part of AIEBU repo but as Telluride based AIEBU is still not made public we are taking library path as ini option and calling dlopen / dlsym to use APIs from the library. This should go away once AIEBU has the APIs for dtrace integrated into it.

Low as dtrace is dumped only when ini option is enabled.

Tested dtrace dump on Telluride qemu flow and the output was as expected.

NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -15,9 +15,11 @@
 
 #include "xrt/detail/ert.h"
 
+#include "bo_int.h"
 #include "elf_int.h"
 #include "module_int.h"
 #include "core/common/debug.h"
+#include "core/common/dlfcn.h"
 
 #include <boost/format.hpp>
 #include <elfio/elfio.hpp>
@@ -26,6 +28,7 @@
 #include <atomic>
 #include <cstdint>
 #include <cstring>
+#include <filesystem>
 #include <numeric>
 #include <map>
 #include <regex>
@@ -154,7 +157,8 @@ struct patcher
     pdi = 4,             // pdi
     ctrlpkt_pm = 5,      // preemption ctrl pkt
     pad = 6,             // scratchpad/control packet section name for next gen aie devices
-    buf_type_count = 7   // total number of buf types
+    dump = 7,            // dump section containing debug info for trace etc
+    buf_type_count = 8   // total number of buf types
   };
 
   buf_type m_buf_type = buf_type::ctrltext;
@@ -180,7 +184,8 @@ struct patcher
       ".preempt_restore",
       ".pdi",
       ".ctrlpkt.pm",
-      ".pad"
+      ".pad",
+      ".dump"
     };
 
     return section_name_array[static_cast<int>(bt)];
@@ -473,6 +478,12 @@ public:
 
   virtual xrt::bo&
   get_scratch_pad_mem()
+  {
+    throw std::runtime_error("Not supported");
+  }
+
+  virtual const buf&
+  get_dump_buf() const
   {
     throw std::runtime_error("Not supported");
   }
@@ -1196,6 +1207,7 @@ public:
 class module_elf_aie2ps : public module_elf
 {
   std::vector<ctrlcode> m_ctrlcodes;
+  buf m_dump_buf; // buffer to hold .dump section used for debug/trace
 
   // The ELF sections embed column and page information in their
   // names.  Extract the column and page information from the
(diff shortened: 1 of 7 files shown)


xuhz (approved):

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1885,6 +1926,140 @@ class module_sram : public module_impl
...
+    {
+      handle = xrt_core::dlopen(path.c_str(), RTLD_LAZY);
+    }
+
+    ~dlib_handle()
+    {
+      if (handle)
+        xrt_core::dlclose(handle);
+    }
+  };
```
stsoe, line 1943: This is marginally silly, but you may want to consider hiding the unique_ptr within struct dlib_handle.   Take a look at runtime_src/core/common/runner/cpu.cpp:dllwrap.

If this is temporary code based on AIEBU comment below, then please ignore this comment.

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1885,6 +1926,140 @@ class module_sram : public module_impl
...
+      return false;
+    }
+
+    dtrace_obj.map_data = std::string{reinterpret_cast<const char*>(data.data()), data.size()};
+    return true;
+  }
+
+  void
+  initialize_dtrace_buf(const module_impl* parent)
+  {
```
stsoe, line 1994: Since this function is rather long, can we insert some comments in between the various blocks as to what exactly is being initialized?

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1885,6 +1926,140 @@ class module_sram : public module_impl
...
+      // sync dtrace control buffer
+      m_dtrace_ctrl_bo.sync(XCL_BO_SYNC_BO_TO_DEVICE);
+      msg = "[dtrace] : dtrace buffers initialized successfully";
+    }
+    catch (const std::exception &e) {
+      m_dtrace_ctrl_bo = {};
+      msg = std::string{"[dtrace] : dtrace buffers initialization failed, "} + e.what();
+    }
+
+    xrt_core::message::send(xrt_core::message::severity_level::debug, "xrt_module", msg);
```
stsoe, line 2060: Remove and put on line 2053 and 2057.  Move std::string from line 2030 to local variable if at all needed.

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1977,6 +2153,50 @@ class module_sram : public module_impl
...
+        create_result_file(dtrace.ctrl_file_path.c_str(), dtrace.map_data.c_str(),
+                           (m_dtrace_ctrl_bo.size() / sizeof(uint32_t)), m_dtrace_ctrl_bo.map<uint64_t*>(),
+                           0, nullptr, result_file_path.c_str());
+      msg = std::string{"[dtrace] : dtrace buffer dumped successfully to - "} + result_file_path;
+    }
+    catch (const std::exception &e) {
+      msg = std::string{"[dtrace] : dtrace buffer dump failed, "} + e.what();
+    }
+
+    xrt_core::message::send(xrt_core::message::severity_level::debug, "xrt_module", msg);
```
stsoe, line 2198: Ditto has previous.

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1885,6 +1926,140 @@ class module_sram : public module_impl
...
+
+  struct dtrace_util
+  {
+    std::unique_ptr<dlib_handle> lib_hdl;
+    std::string ctrl_file_path;
+    std::string map_data;
+  };
+
+  static bool
+  init_dtrace_helper(const module_impl* parent, dtrace_util& dtrace_obj)
```
stsoe, line 1953: Return codes are not usually desired, but in this case make sense after reading how it is used.  This deserves a comment.  Basically, the exception alternative is not good when the success path allows init_dtrace_helper to be a no-op.

stsoe (commented):
Some comments.

src/runtime_src/core/common/api/xrt_module.cpp
rbramand-xilinx: It is temporary code but I will refactor it as code of dllwrap makes more sense, Thanks Soren.

stsoe (approved):

chvamshi-xilinx (approved):

PR #8736, opened 2025-01-31
VITIS-13434 Provide more error details for AIE async errors via xrt::error

zhangchiming:
Problem solved by the commit
Initial check in for adding more error info (NPU error id, col and row) from driver to xrt::error.

N/A

Add another 64 bit error code to existing xrt::error_impl and it will be populated by device driver.

Low, since the new code path will not be used until the NPU device driver code is updated with new error reporting.

Compile and tested on Windows 11 with both existing driver and the upcoming new driver with extra error code reporting.

Documentation impact (if any)
N/A


src/runtime_src/core/common/api/xrt_error.cpp
```diff
@@ -213,12 +214,23 @@ class error_impl
...
+      else {
+        auto ect = xrt_core::query::xocl_errors::to_value(buf, ecl);
+        std::tie(m_errcode, m_timestamp) = ect;
+        m_ex_error_str = "";
+      }
     } catch (const xrt_core::query::no_such_key&) {
       // Ignoring for now. Check below for edge if not available
       // query table of zocl doesn't have xocl_errors key
+    } catch (...) {
+      return;
```
stsoe, line 233: Catch all is not generally a good idea, should we not let unhandled errors propagate to caller?

src/runtime_src/core/common/api/xrt_error.cpp
```diff
@@ -205,6 +205,7 @@ class error_impl
 {
   xrtErrorCode m_errcode = 0;
   xrtErrorTime m_timestamp = 0;
+  std::string  m_ex_error_str;
```
stsoe, line 208: I don't seem to be able to find any code that uses this member variable after it is initialized in constructor?  What purpose does it serve?

src/runtime_src/core/common/api/xrt_error.cpp
```diff
@@ -213,12 +214,23 @@ class error_impl
...
         return;
-      auto ect = xrt_core::query::xocl_errors::to_value(buf, ecl);
-      std::tie(m_errcode, m_timestamp) = ect;
-      return;
+      if (device->get_ex_error_support() == true) {
+        auto ect = xrt_core::query::xocl_errors::to_ex_value(buf, ecl);
+        m_errcode = std::get<0>(ect);
+        m_timestamp = std::get<1>(ect);
+        uint64_t ex_error_code = std::get<2>(ect);
+        m_ex_error_str = xrt_core::device_query<xrt_core::query::xocl_ex_error_code2string>(device, ex_error_code);
```
stsoe, line 222: Initialized here, but is the variable and its value ever used?

src/runtime_src/core/common/query_requests.h
```diff
@@ -1645,6 +1648,39 @@ struct error : request
...
+
+  static bool
+    to_bool(const result_type& value)
+  {
+    return (value == std::numeric_limits<uint32_t>::max())
+      ? false : value;
+  }
+};
+
+// Retrieve support for extended asynchronous xocl errors from xocl driver
```
stsoe, line 1668: I don't think this comment matches what the query is doing?  With reference to the comments in xrt_error.cpp, I am not sure of the purpose of this query request?

stsoe (commented):
The implementation here is clean and good, but I have few clarifying questions.

src/runtime_src/core/common/api/xrt_error.cpp
zhangchiming: > I don't seem to be able to find any code that uses this member variable after it is initialized in constructor? What purpose does it serve?

Sorry, there is a line missing in previous commit. "m_ex_error_str" was added to capture the extra error string (aie error id/col/row) to be returned by the device driver.

src/runtime_src/core/common/api/xrt_error.cpp
zhangchiming: Just added the missing line that actually use this variable.

src/runtime_src/core/common/api/xrt_error.cpp
zhangchiming: Removed unnecessary catch all.

src/runtime_src/core/common/query_requests.h
zhangchiming: Sorry, wrong comment has been corrected.

PR #8751, opened 2025-02-10
Added aie, aiemem, aieshim reports to the ryzen class of devices

ManojTakasi:
Added aie, aiemem, aieshim reports to the ryzen class of devices

We are using ryzen device type on the Telluride side, and it is essential to enable these reports in order to get the status of AIE.

Added requied reports on the Ryzen device in order to generate them.

low

Tested by building and running multi-layer, vadd, eff_net test cases on the  telluride.

### DIFF
--- a/src/runtime_src/core/tools/xbutil2/xbutil.cpp
+++ b/src/runtime_src/core/tools/xbutil2/xbutil.cpp
@@ -56,7 +56,7 @@ R"(
 },{
   "aie": [{
     "examine": [{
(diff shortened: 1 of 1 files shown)


AShivangi (changes requested):
With the latest XRT revamp, this change won't work. Please close this PR and file a new one with the correct changes!

ManojTakasi:
Sure Shivangi, Thank you..

PR #8759, opened 2025-02-13
Bash CR fixes

aktondak:
Fixes 2 CRs :
Changes code to query validate test names from the newly architected structure. This fixes CR-1228351.
Adds a new option --advanced (similar to --show-hidden) to print and run a detailed xrt-smi

https://jira.xilinx.com/browse/CR-1228351
https://jira.xilinx.com/browse/CR-1228349

Solved through altering the handling of validate test list

None
Tested on Ryzen platform

None


AShivangi (changes requested):
Let's keep only one option "--advanced" for simplicity

aktondak:
> Let's keep only one option "--advanced" for simplicity

The API XBU::getShowHidden() is being used at some places in HelpMenu printing logics. I'm not sure if replacing them would be a good idea ?

AShivangi:
> > Let's keep only one option "--advanced" for simplicity
> 
> The API XBU::getShowHidden() is being used at some places in HelpMenu printing logics. I'm not sure if replacing them would be a good idea ?

we're just renaming it to getAdvanced(). The logic remains the same, we just change the word from -show-hidden to advanced

PR #8773, opened 2025-02-19
Remove comparison magical -1

AShivangi:
Not all parameters of telemetry are available for all devices. To distinguish between what is available on a given device, firmware either populates a field w/ an appropriate value or initializes it to 'f's

Internal

Compare against a defined sentinel value, i.e., `na_value`

N/A

Tested with MCDM

N/A

### DIFF
--- a/src/runtime_src/core/common/info_telemetry.cpp
+++ b/src/runtime_src/core/common/info_telemetry.cpp
@@ -8,7 +8,12 @@
 #include <vector>
 
 namespace {
-constexpr uint64_t na_value = std::numeric_limits<uint64_t>::max();
+template <typename IntegralType>
+bool 
+is_value_na(IntegralType value) 
+{
+  return value == std::numeric_limits<IntegralType>::max();
+}
 
 static void
 add_rtos_tasks(const xrt_core::device* device, boost::property_tree::ptree& pt)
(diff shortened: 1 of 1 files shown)


stsoe (commented):
Will this really work? Many values are compared against `no_value` which is uint64_t max, but are the values themselves uint64_t?

I think you need to encapsulate this better if you want to consolidate all compares.
I would think you can do something like this:
```
template <typename IntegralType>
bool 
no_value(IntegralType value) 
{
   return value == std::numeric_limts<IntegralType>::max();
}
```
And use like this
```
if (no_value(stream_buf.tokens))
  return; // not supported
```

If you prefer the operator syntax (`==`) then templatize `operator ==(lhs, rhs)` rather than `novalue(value)`

AShivangi:
> Will this really work? Many values are compared against `no_value` which is uint64_t max, but are the values themselves uint64_t?
> 
> I think you need to encapsulate this better if you want to consolidate all compares. I would think you can do something like this:
> 
> ```
> template <typename IntegralType>
> bool 
> no_value(IntegralType value) 
> {
>    return value == std::numeric_limts<IntegralType>::max();
> }
> ```
> 
> And use like this
> 
> ```
> if (no_value(stream_buf.tokens))
>   return; // not supported
> ```
> 
> If you prefer the operator syntax (`==`) then templatize `operator ==(lhs, rhs)` rather than `novalue(value)`

Yes! All the values are uint64_t so what I have works. However, I like what you have proposed. I'll make that change and test it out!

stsoe (approved):
Looks good.  Thanks!

PR #8776, opened 2025-02-20
CR-1229258 seeing "[XRT] ERROR: Can't Reset AIE: AIE is not initializ…

SravanKumarAllu-xilinx:
…ed" error

Properly resetting aie array upon resetAieArray() is called , so that for next loadXclbin, aie Array can be initialized properly

N/A

Resetting aie array shared pointer

N/A

simple_graph

N/A


src/runtime_src/core/edge/user/aie/aie.cpp
```diff
@@ -418,6 +418,8 @@ reset(const xrt_core::device* device, uint32_t hw_context_id, uint32_t partition
   int ret = drv->resetAIEArray(reset);
   if (ret)
     throw xrt_core::error(ret, "Fail to reset AIE Array");
+  
+  drv->reset_aie_array();
```
chvamshi-xilinx, line 422: Looks like you are already calling  resetAieArray function in  418 line. Why dont you update that function itself to do this extra functionality aswell
  int ret = drv->resetAIEArray(reset);

PR #8787, opened 2025-02-24
[XDP] Core/Memory module trace for the tiles in same column

vipangul:
CR-1229145: As reported in this CR, core and memory module trace metric sets were not working in combination.

It is observed that tiles in same column were causing broadcast event conflict because of missing broadcast masking.
This commit enables required masking for DMA metric sets as well.

Low

MCDM build - Verified on Client


jvillarre (approved):

pgschuey (approved):
Changes look good as discussed

PR #8789, opened 2025-02-25
[xrt-smi] Add power check for amdxdna

rchane:
We need to display "N/A" for xrt-smi platform report power reading for amdxdna for the time being.
Noticed during internal testing.
amdxdna will return max integer value for now until we get a real value. If reading is max, we will not add to ptree.
N/A
Tested on STX amdxdna.
N/A

### DIFF
--- a/src/runtime_src/core/common/sensor.cpp
+++ b/src/runtime_src/core/common/sensor.cpp
@@ -215,7 +215,7 @@ read_data_driven_electrical(const std::vector<xq::sdm_sensor_info::data_type>& c
   // iterate over power data, store to ptree by converting to watts.
   for (const auto& tmp : power) {
     if (boost::iequals(tmp.label, "Total Power")) {
-      if (tmp.input != UINT32_MAX)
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/sensor.cpp
```diff
@@ -214,7 +215,8 @@ read_data_driven_electrical(const std::vector<xq::sdm_sensor_info::data_type>& c
   // iterate over power data, store to ptree by converting to watts.
   for (const auto& tmp : power) {
     if (boost::iequals(tmp.label, "Total Power")) {
-      bd_power = xrt_core::utils::format_base10_shiftdown(tmp.input, tmp.unitm, 3);
+      if (tmp.input != UINT32_MAX)
```
stsoe, line 218: I am not sure if this is important, but this code will break if someone changes the type of `xq::sdm_sensor_info::data_type::input`.  
```suggestion
      if (tmp.input != std::numeric_limits<decltype(xq::sdm_sensor_info::data_type::input)>::max())`
```

rchane: Thank you! I edited the check.

AShivangi (approved):

PR #8791, opened 2025-02-25
CR-1230538 xrt-smi validate for amdxdna supported devices requires root priviledges

AShivangi:
[CR-1230538](https://jira.xilinx.com/browse/CR-1230538) xrt-smi validate for amdxdna supported devices requires root priviledges

behavior change requested

- If pmode is set to performance before running validate, don't try to set it again
- If "permission denied" (this is because sudo was not used) - so print a warning and run validate

N/A

TBD on linux-xdna

N/A

### DIFF
--- a/src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
+++ b/src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
@@ -607,8 +607,14 @@ SubCmdValidate::execute(const SubCmdOptions& _options) const
   }
 
   //get current performance mode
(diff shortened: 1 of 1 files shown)


rchane (approved):

aktondak:
How is this code change paltform agnostic ? Can we stop pushing code with actual string checks:
`catch (const std::exception & ex) { //check if permission was denied, i.e., no sudo access
    if(!boost::icontains(ex.what(), "Permission denied")) {
      std::cerr << boost::format("\nERROR: %s\n") % ex.what();
    }`

PR #8581, opened 2024-10-29
Xclbin to ELF flow migration

rbramand-xilinx:
#### Problem solved by the commit
This PR enables new flow where ELF file is input instead of xclbin.
spec - https://confluence.amd.com/display/AIE/AIE+Compiler+Artifacts#AIECompilerArtifacts-Flow2

spec of new config elf - https://confluence.amd.com/pages/viewpage.action?pageId=1485002156#Profile/Config/CtrlcodeElfSpec-Config.elf

This new Elf  has partition size and kernel signature.
It has multiple .ctrltext and .ctrldata sections and multiple .pdi sections.
Each ctrltext section represents one control code and pdi addresses along with kernel args needs to be patched into the control code buffer.
Control code that needs to be run is decided while creating xrt::kernel object using kernel name  -  <kernel_name>:<ctrl_code_id>   eg: "DPU:0"  runs with control code from section .ctrltext0  and  "DPU:1" runs with control code from section .ctrlcode1

Sample test case :
```
#include "xrt/xrt_hw_context.h"
#include "xrt/xrt_kernel.h"
#include "xrt/xrt_elf.h"

int main(int argc, char** argv)
{
    std::string elf_path {argv[1]};   // elf path
    auto elf = xrt::elf(elf_path);

    auto device = xrt::device(0);
    auto ctx = xrt::hw_context(device, elf);

    std::string kernel_name = "DPU:1";
    auto kernel = xrt::ext::kernel(ctx, kernel_name);
    
    //  create args
    auto run = kernel(arg1, arg2 ...);
    run.wait();

    return 0;
}
```

XRT first class object changes :
xrt::hw_context  -  Added new APIs to create context using Elf instead of xclbin
xrt::kernel  -  Added new ext API to create kernel object using kernel name, added changes to construct kernel args from kernel signature.
xrt::elf  -  can now take new Elf with os abi 70 as input
xrt::module  -  added changes to parse this new Elf, also refactored code

**TODO** :  Refactor code to parse ELF file using xrt::elf class and make it similar to xrt::xclbin. Created CR - https://jira.xilinx.com/browse/CR-1219757 for tracking this.

#### Bug / issue (if any) fixed, which PR introduced the bug, how it was discovered
This is a new feature

#### How problem was solved, alternative solutions (if any) and why they were rejected
Hw context can be created either with xclbin (traditional flow) or with xrt::elf. A new constructor is provided for same.
A new xrt::ext::kernel constructor is provided that takes ctx and kernel name as input.
[...]

PR #8794, opened 2025-02-26
Fix bug when using static ref and std::make_pair

rbramand-xilinx:
Fixed bug when using std::make_pair to return const reference to a static object.

std::make_pair doesn't work when either first or second is a reference, it makes a local copy of what the reference refers to and but as the local copy goes out of scope we get junk values

PR - https://github.com/Xilinx/XRT/pull/8581 introduced the bug and it was discovered when doing a local testing

Using initializer list to create std::pair instead of std::make_pair

Low

Tested the failing test case (elf_txn_no_cp) on phoenix hw and test passes after the fix

NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -914,7 +914,7 @@ class module_elf_aie2p : public module_elf
       
       uint32_t index = get_section_name_index(name);
       buf.append_section_data(sec.get());
-      map.emplace(std::make_pair(index, std::make_pair(sec_index, buf)));
+      map.emplace(index, std::pair{sec_index, buf});
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -914,7 +914,7 @@ class module_elf_aie2p : public module_elf
       
       uint32_t index = get_section_name_index(name);
       buf.append_section_data(sec.get());
-      map.emplace(std::make_pair(index, std::make_pair(sec_index, buf)));
+      map.emplace(index, std::pair{sec_index, buf});
```
stsoe, line 917: Ah yes,  std::pair is necessary here.  std::make_pair would have worked in this case, because the first and second of the pair are values, but emplacing as a tuple is good.

stsoe (approved):
Looks good.  Thanks!

PR #8797, opened 2025-02-28
Fix build against boost 1.87.0

FireBurn:
The io_service has been renamed to io_context


stsoe:
Thank you for this fix @FireBurn

PR #8799, opened 2025-03-04
CR-1228350 fix : Virtualizing configure options for alveo/pcie 

aktondak:
This PR changes the xrt-smi configure options for alveo/edge case vs MCDM/XDNA case. MCDM and XDNA will have a minimal list of configure options

https://jira.xilinx.com/browse/CR-1228350

The problem was solved by creating a default structure configure options for the default, alveo and edge cases which consists of a list of all options. This structure will be overridden by MCDM and XDNA shims to only show a minimal list of configure options.

None

Tested on windows platform

None

### DIFF
--- a/src/runtime_src/core/common/smi.cpp
+++ b/src/runtime_src/core/common/smi.cpp
@@ -49,6 +49,20 @@ smi_base()
   examine_report_desc = {
     {"host", "Host information", "common"}
   };
+
+  configure_options = {
+    {"device", "d", "The Bus:Device.Function (e.g., 0000:d8:00.0) device of interest", "common", "", "string"},
+    {"help", "h", "Help to use this sub-command", "common", "", "none"},
+    {"daemon", "", "Update the device daemon configuration", "hidden", "", "none"},
+    {"purge", "", "Remove the daemon configuration file", "hidden", "", "string"},
+    {"host", "", "IP or hostname for device peer", "hidden", "", "string"},
+    {"security", "", "Update the security level for the device", "hidden", "", "string"},
+    {"clk_throttle", "", "Enable/disable the device clock throttling", "hidden", "", "string"},
+    {"ct_threshold_power_override", "", "Update the power threshold in watts", "hidden", "", "string"},
+    {"ct_threshold_temp_override", "", "Update the temperature threshold in celsius", "hidden", "", "string"},
+    {"ct_reset", "", "Reset all throttling options", "hidden", "", "string"},
+    {"showx", "", "Display the device configuration settings", "hidden", "", "string"}
+  };
 }
 
 std::vector<basic_option> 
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/common/smi.cpp
```diff
@@ -49,6 +49,20 @@ smi_base()
   examine_report_desc = {
     {"host", "Host information", "common"}
   };
+
+  configure_options = {
```
AShivangi, line 53: shouldn't this go in alveo/smi.cpp?

aktondak: I have updated the change.
Now the default behavior is a minimal list of 2 options which are going to be used by MCDM and XDNA shims.
edge and pcie shims have this exhaustive list

src/runtime_src/core/edge/user/smi.cpp
```diff
@@ -37,6 +37,20 @@ smi_edge() : smi_base()
     {"qspi-status", "QSPI write protection status", "common"},
     {"thermal", "Thermal sensors present on the device", "common"}
   };
+
+  configure_options = {
+    {"device", "d", "The Bus:Device.Function (e.g., 0000:d8:00.0) device of interest", "common", "", "string"},
+    {"help", "h", "Help to use this sub-command", "common", "", "none"},
```
AShivangi, line 43: Is it ok if these are repeated?

aktondak: configure_options structure is used as a whole. These are not repeated since the whole vector is constructed for smi_edge

AShivangi (approved):

stsoe:
Before this gets out of hand, I thought I would mention that clang-tidy is unhappy about protected data members.  A cpp core guideline is that we should avoid protected data members. 
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c133-avoid-protected-data

For the validata_test_desc and examine_report_desc, it is better to make these private and have constructors that take tuple vector by value and moves the content into the private data members.   Why?  Because if you tomorrow need to change the content of tuple vector (new element or reordered element), then you can add tuple_vector_extended as data members while still preserving compatibility for old tuple_vector constructors, which now of course would have to transform into tuple_vector_extended.

This may not seem overly important as we manage all uses of smi across our projects, but imagine a true exported API, where ABI compatibility is required.  It would sure be a disaster to use protected data members in such a case.

aktondak:
> Before this gets out of hand, I thought I would mention that clang-tidy is unhappy about protected data members. A cpp core guideline is that we should avoid protected data members. https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c133-avoid-protected-data
> 
> For the validata_test_desc and examine_report_desc, it is better to make these private and have constructors that take tuple vector by value and moves the content into the private data members. Why? Because if you tomorrow need to change the content of tuple vector (new element or reordered element), then you can add tuple_vector_extended as data members while still preserving compatibility for old tuple_vector constructors, which now of course would have to transform into tuple_vector_extended.
> 
> This may not seem overly important as we manage all uses of smi across our projects, but imagine a true exported API, where ABI compatibility is required. It would sure be a disaster to use protected data members in such a case.

I agree with this. Thanks for the feedback.
I'll take care of this in a future change since this is already merged.

PR #8800, opened 2025-03-04
Fix build against boost 1.87.0

stsoe:
The io_service has been renamed to io_context

This PR is to enable CI on #8797


stsoe:
#8797 is now merged

PR #8807, opened 2025-03-07
[XDP] Prevent segfault by checking correct usage for range of tiles for aie_profile

snigdha-gupta:
[CR-1223004](https://jira.xilinx.com/browse/CR-1223004) Using range of columns for memory tile causes segmentation fault

[UG-1076](https://docs.amd.com/r/en-US/ug1076-ai-engine-environment/XRT-Flow?tocId=UBk2qtIX2tunIJmjD4G4pw) has a documentation error which specifies the wrong format for range of tiles for tile-based settings of memory tiles.

Ensured that there were two parameters (col, row) entered. If they weren't, we skip it.

Low.

Tested on VEK280. Verified that there is no segmentation fault and instead the user sees the appropriate warning message. Also ensured that for the correct format there is no crash.

Yes - [UG-1076](https://docs.amd.com/r/en-US/ug1076-ai-engine-environment/XRT-Flow?tocId=UBk2qtIX2tunIJmjD4G4pw) needs to be updated to have the correct format for for range of tiles for tile-based settings of memory tiles.

### DIFF
--- a/src/runtime_src/xdp/profile/plugin/aie_profile/aie_profile_metadata.cpp
+++ b/src/runtime_src/xdp/profile/plugin/aie_profile/aie_profile_metadata.cpp
@@ -501,11 +501,21 @@ namespace xdp {
 
         std::vector<std::string> minTile;
         boost::split(minTile, metrics[i][0], boost::is_any_of(","));
-        minCol = aie::convertStringToUint8(minTile[0]);
-        minRow = aie::convertStringToUint8(minTile[1]) + rowOffset;
 
         std::vector<std::string> maxTile;
         boost::split(maxTile, metrics[i][1], boost::is_any_of(","));
+        
+        if (minTile.size() != 2 || maxTile.size() != 2) {
+          std::stringstream msg;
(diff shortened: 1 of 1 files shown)


src/runtime_src/xdp/profile/plugin/aie_profile/aie_profile_metadata.cpp
```diff
@@ -501,11 +501,21 @@ namespace xdp {
 
         std::vector<std::string> minTile;
         boost::split(minTile, metrics[i][0], boost::is_any_of(","));
-        minCol = aie::convertStringToUint8(minTile[0]);
-        minRow = aie::convertStringToUint8(minTile[1]) + rowOffset;
 
         std::vector<std::string> maxTile;
         boost::split(maxTile, metrics[i][1], boost::is_any_of(","));
+        
+        if (minTile.size() != 2 || maxTile.size() != 2) {
```
pgschuey, line 508: Please put parenthesis around each separate check

pgschuey (approved):

PR #8815, opened 2025-03-13
Add changes to partial sync xrt::module buffers after they are patched

rbramand-xilinx:
Added changes to sync xrt::module buffers like instruction, control pkt, save, restore etc partially after they are patched. This is to improve performance.

The current code syncs entire buffer even when we patch single arg which is an overhead and increases the overall time taken for test to run.
Modified the code to sync only parts of buffer(partial sync) based on the locations of patch. This improved the performance.

Moderate and may cause regression

Tested by running application multiple times with big instruction buffer size and with the changes in PR the performance improved.

NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -42,6 +42,7 @@
 #include <tuple>
 #include <type_traits>
 #include <unordered_set>
+#include <variant>
 
 #ifdef _WIN32
 #include <dbghelp.h>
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -293,10 +293,36 @@ struct patcher
...
+  patch_it(xrt::bo bo, uint64_t value)
+  {
+    patch_it_impl(bo, value);
+  }
+
+private:
+  template<typename T>
+  void
+  patch_it_impl(T base_or_bo, uint64_t new_value)
+  {
```
stsoe, line 311: I would recommend to keep just the uint8_t version, then do the map & cast where you are calling patch_it_impl.
What you have here, is not the right use of a template.
```
void
patch_it_impl(uint8_t* base, uint64_t new_value)
{
   ...
}
```

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -710,6 +781,21 @@ class module_elf : public module_impl
...
+    return patch_it_impl(base, argnm, index, patch, type, sec_index);
+  }
+
+  bool
+  patch_it(xrt::bo bo, const std::string& argnm, size_t index, uint64_t patch,
+           patcher::buf_type type, uint32_t sec_index) override
+  {
+    return patch_it_impl(bo, argnm, index, patch, type, sec_index);
+  }
```
stsoe, line 797:
```suggestion
  bool
  patch_it(const xrt::bo& bo, const std::string& argnm, size_t index, uint64_t patch,
           patcher::buf_type type, uint32_t sec_index) override
  {
    return patch_it_impl(bo.map<uint8_t*>(), argnm, index, patch, type, sec_index);
  }
```

stsoe (changes requested):

src/runtime_src/core/common/api/xrt_module.cpp
rbramand-xilinx: Hi @stsoe, we need to call sync on the bo for the args that are patched.
My initial implementation was to have 2 separate functions so that I can do bo specific code but refactored it this way to avoid code duplication. I can have 2 functions if code duplication is okay.

The other way to solve this is :

bool
patch_it(xrt::bo bo, ....)
{
  if (!patch_it_impl(bo.map<>....))
     return false;

  for (...) {
     switch (m_symbol_type) {
        case a :
          bo.sync();    // call sync here
        case b :
           bo.sync();
        ....
     }
  }
}

Let me know your thoughts, Thanks

stsoe: I see.  The template just isn't right in this case, but neither is code duplication.  Two options (1) different kind of refactoring, (2) std::variant with std::visit for the special action depending on the variant type.

rbramand-xilinx: Thanks Soren for the suggestion, refactored the code to use std::variant and std::visit.
Please review

stsoe (commented):
I am not entirely following this code.   I don't think it is clearer now with std::variant.  I will leave it to you deciding on this version or the previous templated one, whichever you are most comfortable with.  We can rewrite another day if necessary.

rbramand-xilinx:
> I am not entirely following this code. I don't think it is clearer now with std::variant. I will leave it to you deciding on this version or the previous templated one, whichever you are most comfortable with. We can rewrite another day if necessary.

I reverted std::variant changes and kept the original templated version for now, we can rewrite the code if necessary like you said. Thanks Soren

stsoe (approved):

PR #8820, opened 2025-03-17
fix the race condition between device::get_xclbin() and device::record_xclbin().

zhangchiming:
Problem solved by the commit
Application crash because of the race condition between device::get_xclbin() and device::record_xclbin().

CR-1233298 - Back-to-back hw context creation and deletion caused crash of application
https://jira.xilinx.com/browse/CR-1233298

Move the lock in device::get_xclbin() to fix the race condition.

Low.

Tested on RyzenAI+Windows.

Documentation impact (if any)
N/A

### DIFF
--- a/src/runtime_src/core/common/device.cpp
+++ b/src/runtime_src/core/common/device.cpp
@@ -189,9 +189,9 @@ get_xclbin(const uuid& xclbin_id) const
   // Allow access to xclbin in process of loading via device::load_xclbin
   if (xclbin_id && xclbin_id == m_xclbin.get_uuid())
     return m_xclbin;
-  if (xclbin_id) {
+
+  if (xclbin_id)
     return m_xclbins.get(xclbin_id);
-  }
 
   // Single xclbin case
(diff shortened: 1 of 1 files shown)


stsoe (commented):
I am a little worried about this change.  The xclbin loading is convoluted because of a legacy single xclbin support for Alveo and Emulation.   I am worried, but not sure, that for the single xclbin case, get_xclbin(uuid) may be called while m_mutex is locked.   If emulation test cases pass in CI, then supposedly we are okay. 

Based on https://jira.xilinx.com/browse/CR-1233298, I do understand the need for the lock.  But it is possible we need to revisit xclbin loading for the single xclbin case.

stsoe (commented):
I don't understand what the last change is doing?  I think your original one was good and correct, I just wasn't 100% sure if the mutex could be locked already.  I think if all tests are passing, your first change was good and should be merged.

We have problems with sw_emu right now that I don't see related to your change.

PR #8824, opened 2025-03-19
Add xrt::device exception class hierarchy

stsoe:
Define xrt::device specific exceptions for handling by application.


stsoe:
Failing sw_emu because sw_emu is no longer supported by Vitis.

PR #8825, opened 2025-03-19
XRT smi code cleanup to remove device specific help printing logic

aktondak:
This PR cleans up the xrt-smi code which depended on the configuration embedded in xbutil.cpp. 
With all the xrt-smi re-architecture we did previously, there is no longer a need for xrt-smi subcommands to book keep device/platform specific behavior within XRT. Instead its the responsibility of each shim to provide correct configuration. This clean-up aims at simplifying the xrt-smi help printing and option adding logic. There are following broad changes in this PR :
1. Some improved naming conventions for smi specific files and data structures in each shim (smi.cpp --> smi_pcie.cpp etc). Also moved the base smi class members to private as pointed out in a previous review. 
2. Removal of m_commandConfig setting in two subcommands : validate and configure. Configure is common between xbmgmt thus cleaning it up will require pcie/edge shim changes. I'll plan that in subsequent patches.
3. Simplifies the help printing by removing all print_help_internal calls/definitions which were device dependent and simply using the base printHelp() for all help printing. Note for future, there should be no device specific help printing logic added to XRT code.

https://jira.xilinx.com/browse/VITIS-14551
Discovered through internal testing

Solved through xrt-smi re-architecture done in last release. This is mostly subsequent code cleanup.

This change has a potential of breaking some of the help printing possibly in non xrt-smi code. Will be testing it further

Tested for all 3 commands on windows.

None


AShivangi (approved):
Thanks for the cleanup!!
Please double check if the copyrights are up to date

rchane (approved):

stsoe (approved):

PR #8829, opened 2025-03-20
[XDP] Enable shim tiles matching using both channelId and streamId.

vipangul:
[CR-1218875] - streamIDs are shown in Vitis analyzer for Shim GMIO interface channels. However, by default only support profiling of channel0 or channel1 (Master/Slave)
In this defect, it tries to profile for particular stream ID instead of channel ID.
`tile_based_interface_tile_metrics=all:mm2s_throughputs:7`

[CR-1218875]

For such case of GMIOs, code now checks for matching channelID and streamID.
This implicit understanding of channel/streamId is currently overloaded and will be  streamlined with json settings using separate keys.

Low

vck190 DUT


jvillarre (approved):

pgschuey (commented):
As discussed, this is a temporary fix, since there is the possibility that channel number and stream IDs overlap. Since it's only applicable when that occurs for GMIO designs and the user has specified a desired ID, it's a rare case when that matters. In the future, json can more aptly cover these different cases.

PR #8832, opened 2025-03-24
CR-1232385 : Fix seg fault when xrt::device is created in a loop in pcie hw_emu flow

rbramand-xilinx:
Fixed seg fault when xrt::device is created in a loop
The seg fault happens in second run as core device held by shim is recreated and existing one is destroyed and underlying handles are cleaned but in the recreation only cached object ptr is returned which is a dangling ptr.
This issue is only in pcie hw_emu flow.

Creating core device in shim xclOpen call only when it is null (first time)

Low

Tested with test case attached in CR (https://github.com/Xilinx/Vitis_Accel_Examples/tree/main/performance/axi_burst_performance) and it passes

NA


stsoe (approved):
Wow.  Surprising that this is surfacing now and not long ago.  This is not new code.

chvamshi-xilinx (approved):

PR #8836, opened 2025-03-26
CR-1235101 : Fix error when including xrt_device.h in C file

rbramand-xilinx:
Fixed compilation error when including xrt_device.h file in .c file.

Moved inclusion of xrt_exception.h in xrt_device.h under __cplusplus gaurd

Low

Tested vvas build flow and it passes after applying this fix

NA


stsoe (changes requested):
I think we need to fix xrt_exception.h instead to be includable from C code.

PR #8841, opened 2025-03-28
Gemm ELF test fix for Linux

Json288:
Fixed invalid arg error when running Gemm test with ELF on Linux
Added a empty bo_ifm to the kernel
Tested both on Windows and Linux with MCDM and Linux and passed. 
Tested command: 
xrt-smi validate -r gemm
xrt-smi validate -r gemm --elf --advance
xrt-smi validate -r all --advance
xrt-smi validate -r all --elf --advance

### DIFF
--- a/src/runtime_src/core/tools/common/tests/TestGemm.cpp
+++ b/src/runtime_src/core/tools/common/tests/TestGemm.cpp
@@ -20,6 +20,8 @@ namespace XBU = XBUtilities;
 #include <thread>
 
 static constexpr uint32_t num_of_cores = 32;
+static constexpr size_t buffer_size_gb = 1;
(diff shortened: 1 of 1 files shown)


AShivangi (approved):

maxzhen:
How does it work on Windows? Why the ifm arg is not required on Windows?

Json288:
> How does it work on Windows? Why the ifm arg is not required on Windows?

XDNA does not allow passing 0 arguments (all NULL). Seems that Windows does not have this requirement.

maxzhen (changes requested):
This PR has to be further evaluated before it can be merged.

Json288:
Closing this PR after an off-line discussion. 
This is caused by an XDNA driver bug and should be fixed in XDNA, not in XRT.

PR #8844, opened 2025-03-31
Temporal sharing overhead test enhancement

aktondak:
This change enhances the spatial/temporal sharing overhead tests (under --advance) to use the mobilenet test representing actual workload to calculate the overhead of temporally shared hardware contexts.

https://jira.xilinx.com/browse/CR-1225327

The problem was solved using input feature map, parameters and xclbin from a real test mobilenet. The mentioned 3 binaries are installed in the driver store under Mobilenet which are queried through a newly added device query. These are then used to run 1000 kernels to saturate the hardware and calculate the overhead of running 2 temporally shared hardware contexts

None. Test is still under advance

Tested on windows kracken platform
```
Z:\Repos\XRT-MCDM-FORK\XRT-MCDM\build\WRelease\xilinx\xrt>.\xrt-smi validate --run temporal-sharing-overhead --advanced
    Platform              : NPU Kracken
    Power Mode            : Performance
Test 1 [00c5:00:01.1]     : temporal-sharing-overhead
    Details               : Overhead: 0.4 ms
```

None


src/runtime_src/core/tools/common/tests/TestSpatialSharingOvd.cpp
```diff
@@ -144,15 +172,15 @@ boost::property_tree::ptree TestSpatialSharingOvd::run(std::shared_ptr<xrt_core:
...
 
   // Log the latencies and the overhead
   if(XBU::getVerbose()){
-    XBValidateUtils::logger(ptree, "Details", boost::str(boost::format("Single context latency: %.1f ms") % (latencySingle * 1000)));
-    XBValidateUtils::logger(ptree, "Details", boost::str(boost::format("Spatially shared multiple context latency: %.1f ms") % (latencyShared * 1000)));
+    XBValidateUtils::logger(ptree, "Details", boost::str(boost::format("Single context latency: %.1f ms") % (latencySingle)));
+    XBValidateUtils::logger(ptree, "Details", boost::str(boost::format("Spatially shared multiple context latency: %.1f ms") % (latencyShared)));
   }
-  auto overhead = (latencyShared - latencySingle) * 1000;
+  auto overhead =  (latencyShared - 2 * latencySingle)/ num_kernels;
```
AShivangi, line 183: what's -2?
`num_kernels` is iterarions?

src/runtime_src/core/tools/common/tests/TestSpatialSharingOvd.h
```diff
@@ -36,7 +36,10 @@ class TestSpatialSharingOvd : public TestRunner {
 
   // Constructor to initialize the test runner with a name and description
   TestSpatialSharingOvd()
-    : TestRunner("spatial-sharing-overhead", "Run Spatial Sharing Overhead Test"), ptree(get_test_header()){}
+  //For the time, the driver mandates even 4 column hardware contexts to 
+  //Occupy all 8 columns. Thus the logic for spatial sharing is implementing temporal sharing.
+  //This should be renamed back once the MCDM driver switches to spatial sharing.
+    : TestRunner("temporal-sharing-overhead", "Run Spatial Sharing Overhead Test"), ptree(get_test_header()){}
```
AShivangi, line 42:
```suggestion
    : TestRunner("temporal-sharing-overhead", "Run Temporal Sharing Overhead Test"), ptree(get_test_header()){}
```

AShivangi (approved):

src/runtime_src/core/common/query_requests.h
```diff
@@ -715,6 +722,36 @@ struct elf_name : request
   get(const device*, const std::any& req_type) const override = 0;
 };
 
+struct mobilenet : request 
+{
+  enum class type {
```
sonals, line 727: _type_ is a very generic term. May be _buffer_type_?

src/runtime_src/core/common/query_requests.h
```diff
@@ -715,6 +722,36 @@ struct elf_name : request
...
+      case type::buffer_sizes:
+        return "buffer_sizes";
+    }
+    return "unknown";
+  }
+
+  using result_type = std::string;
+  static const key_type key = key_type::mobilenet;
+  static const char* name() { return "mobilenet"; }
+
```
sonals, line 750: Who are using these? Should these be virtual functions in the request class for each derived class to override?

stsoe: @sonals These are integrals pieces of the query infrastructure.  The infrastructure is templated, these cannot be virtual functions.

src/runtime_src/core/common/query_requests.h
stsoe: This is a local type used purely in context of the query struct.  While "type" is generic, it surely isn't a buffer_type.  If renaming is desired then please make the name reflect what this really is.

aktondak: Since the usage of type "type" is scoped (xrt_core::query::mobilenet::type), I'll keep it as is for it to be concurrent with other query structures, which keep the name to be "type" as well.

src/runtime_src/core/tools/common/tests/TestSpatialSharingOvd.cpp
```diff
@@ -15,7 +15,7 @@
 namespace XBU = XBUtilities;
 
 static constexpr size_t host_app = 1; //opcode
-static constexpr size_t buffer_size = 1024 * 1024 * 1024; //1 GB
+static constexpr size_t num_kernels = 1000;
```
sonals, line 18: Please rename to _num_kernel_iterations_

sonals (commented):
What is buffer sizes file? Can we not determine the buffer sizes by reading the buffers stored on disk?

sonals (commented):
There is a lot code duplication between src/runtime_src/core/tools/common/tests/TestSpatialSharingOvd.cpp and src/runtime_src/core/tools/common/tests/TestTemporalSharingOvd.cpp. Either we consolidate the common code or we should just get rid of src/runtime_src/core/tools/common/tests/TestSpatialSharingOvd.cpp for now and resurrect it in future when spatial is re-enabled in the execution model.

aktondak:
> What is buffer sizes file? Can we not determine the buffer sizes just by reading the buffers stored on disk?

Following is the content of buffer_sizes.json, which lives along with other Mobilenet binaries in driver store :
    "ifm_size": 201736,
    "param_size": 7233612,
    "inter_size": 7192640,
    "mc_size": 16,
    "ofm_size": 8224,
    "instr_word_size": 122575
    
    Out of these, only ifm and param word have a binary associated with them, rest are constant buffer sizes which I didn't want to hardcode in the test itself. The buffer sizes query object and structures can be re-used for other tests if they want to use any AMD model with same structure. This is similar to DSV's tests which use a ddr_range.txt to query buffer sizes from bigger designs.

aktondak:
> There is a lot code duplication between src/runtime_src/core/tools/common/tests/TestSpatialSharingOvd.cpp and src/runtime_src/core/tools/common/tests/TestTemporalSharingOvd.cpp. Either we consolidate the common code or we should just get rid of src/runtime_src/core/tools/common/tests/TestSpatialSharingOvd.cpp for now and resurrect it in future when spatial is re-enabled in the execution model.

The files for TestSpatialSharing have been removed till we enable spatial sharing from driver.

sonals (approved):

PR #8846, opened 2025-04-01
Add API in xrt::run to get control scratchpad memory

rbramand-xilinx:
Added new API to get control scratchpad memory.
This memory is used by NPU to store control state information of a model. ELF is parsed to look for this section and a xrt::bo object is created corresponding to it.
This PR provides API  to get this buffer object so that users can read/write to this memory.

Provided required function with xrt::run class as each running instance of a model will have its own control scratchpad memory

Low

Tested with Preemption application on strix linux and things work as expected

Added comments with the API declaration.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -2593,6 +2593,24 @@ public:
   {
     return cmd->get_ert_packet();
   }
+
+  [[nodiscard]] std::vector<char>
+  read_ctrl_scratchpad(uint32_t offset, size_t size)
+  {
+    if (!m_module)
+      throw xrt_core::error("No module associated with run object");
+  
+    return xrt_core::module_int::read_ctrl_scratchpad(m_module, offset, size);
+  }
+
+  void
+  write_ctrl_scratchpad(uint32_t offset, const std::vector<char>& buf)
+  {
+    if (!m_module)
+      throw xrt_core::error("No module associated with run object");
+  
+    xrt_core::module_int::write_ctrl_scratchpad(m_module, offset, buf);
+  }
 };
 
 // class mailbox_impl - Extension of run_impl for mailbox support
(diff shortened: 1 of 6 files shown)


src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -2593,6 +2593,24 @@ class run_impl
   {
     return cmd->get_ert_packet();
   }
+
+  [[nodiscard]] std::vector<char>
```
stsoe, line 2597: There is no harm in discarding this return value.   I suggest you remove [[nodiscard]].

src/runtime_src/core/include/xrt/xrt_kernel.h
```diff
@@ -614,6 +614,43 @@ class run
...
+   * @return
+   *  Data read from control scratchpad memory
+   * 
+   * This API is valid only for run objects that are associated with an ELF.
+   * Use this API to read data from control scratchpad memory.
+   * This memory is created by XRT based on ELF used to create xrt::kernel.
+   * 
+   * This function throws if the read fails.
+   */
+  std::vector<char>
```
stsoe, line 633: Should we make the type `uint8_t` ?

I am not excited about std::vector as return type.   Reason is that doesn't handle a case where application controls the data storage, say a user pointer.  Using a std::vector could lead to unnecessary memcpy.

Thinking out loud here, I wonder if the exchange should be through an xrt::bo?  Obviously not a BO that is created by driver, but a type of BO that holds just system memory, either allocated when the BO is created or as a userptr passed by application?

stsoe: I don't think a BO makes sense.  The data is not managed by a BO.  Once written the connection to the BO is lost, I think.

rbramand-xilinx: Yes Soren, I agree this approach is simple and let user manage bo however he wants. Thanks for the suggestion.
I will rewrite the code accordingly.

stsoe (approved):

larry9523 (approved):

PR #8848, opened 2025-04-01
Added config reader support to get mdm privilege mode

bisingha-xilinx:
https://jira.xilinx.com/browse/CR-1229912
Need to have a mdm privilege mode at runtime as a requirement to https://jira.xilinx.com/browse/CR-1229912. So, added a config reader function to get the mode.
Getting the mdm privilege mode with xrt.ini under Runtime. And based on the mode initialize the aie partition. By default, It'll be true, means AIE partitions will be initialized with Privilege mode.
n/a
Tested on XDNA branch of the Telluride driver
n/a


src/runtime_src/core/common/config_reader.h
```diff
@@ -747,6 +747,13 @@ get_hardware_context_type()
   return value;
 }
 
+inline bool
```
chvamshi-xilinx, line 750: Please add some comment/description of this flag

bisingha-xilinx: Hi Vamshi, added a comment and also renamed the function to a generic name.

stsoe: It would be good if the comment explained what a privileged context is.

bisingha-xilinx: Hi @stsoe I've updated the comment with more info. 
Thanks

stsoe (approved):
Thank you.

chvamshi-xilinx (approved):

PR #8854, opened 2025-04-03
Update build of xdp_core to differentiate when building for client

jvillarre:
Enable the utility function "isClient" in the profiling core library.

Issue was discovered when testing new changes in the profiling code that require the isClient() function.

The need for the isClient() function is due to some query_device functionality not being implemented on certain devices.  In the future the XRT APIs we're calling must be implemented in multiple targets.

Low risk as the compile time guard being defined is only used in a single utility function in the profiling core library.

No documentation change necessary.


IshitaGhosh (approved):
Changes look good and works on Win Client.

IshitaGhosh:
This PR should be merged before https://github.com/Xilinx/XRT/pull/8853

PR #8856, opened 2025-04-03
Adjusting the reading of AIE metadata for profiling

jvillarre:
In profiling, we were checking for AIE metadata inside the xclbin and then only if we didn't find it checking the disk.  Instead, if the parameter asks us to check the disk for metadata, we should look at the disk only and ignore anything in the xclbin.  This is because some flows may have a known aie_control_config in the xclbin that does not contain the information we need for profiling and we must avoid processing it.

This disk check will be disabled once all xclbin generation flows have the aie_trace_config added to the xclbin.

PR 8853 added this issue and it was found by testing on client.

Low risk as this changes the metadata reading back to what it was before PR 8853.

The reading of metadata on client.

No documentation impact.


IshitaGhosh (approved):
Changes look good. Now, Client XDP picks up aie_trace_config.json from run dir.

PR #8858, opened 2025-04-03
Elf handling enhancements

sonals:
1. Add  an xrt::elf constructor from void * buffer
2. Add python plumbing for xrt::elf and xrt::program

NA

See above

None

Compile tested

None

### DIFF
--- a/src/runtime_src/core/common/api/xrt_elf.cpp
+++ b/src/runtime_src/core/common/api/xrt_elf.cpp
@@ -49,6 +49,9 @@ public:
 
   explicit elf_impl(const void *data, size_t size)
   {
+    // Uses the same approach as in aiebu reporter.cpp
+    // ibufferstream allows reading from data without first copying over
+    // the data to create the stream
     boost::interprocess::ibufferstream istr(static_cast<const char *>(data), size);
     if (!m_elf.load(istr))
       throw std::runtime_error("not valid ELF data");
--- a/src/runtime_src/core/common/api/xrt_elf.cpp
+++ b/src/runtime_src/core/common/api/xrt_elf.cpp
@@ -18,6 +18,8 @@
 #include <string>
 #include <vector>
 
+#include <boost/interprocess/streams/bufferstream.hpp>
+
 namespace xrt {
 
 // class elf_impl - Implementation
@@ -43,6 +45,13 @@ public:
   {
     if (!m_elf.load(stream))
       throw std::runtime_error("not a valid ELF stream");
+  }
+
+  explicit elf_impl(const void *data, size_t size)
+  {
+    boost::interprocess::ibufferstream istr(static_cast<const char *>(data), size);
+    if (!m_elf.load(istr))
+      throw std::runtime_error("not valid ELF data");
   }
 
   const ELFIO::elfio&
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -45,6 +47,16 @@ class elf_impl
...
 
+  explicit elf_impl(const void *data, size_t size)
+  {
+    // Uses the same approach as in aiebu reporter.cpp
+    // ibufferstream allows reading from data without first copying over
+    // the data to create the stream
+    boost::interprocess::ibufferstream istr(static_cast<const char *>(data), size);
+    if (!m_elf.load(istr))
+      throw std::runtime_error("not valid ELF data");
+  }
```
stsoe, line 58: I think this should work with delegating constructor.
```suggestion
  // Uses the same approach as in aiebu reporter.cpp
  // ibufferstream allows reading from data without first copying over
  // the data to create the stream
  explicit 
  elf_impl(const void *data, size_t size)
    :  elf_impl{boost::interprocess::ibufferstream{static_cast<const_char*>(data), size}}
  {}
```

src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -137,6 +149,11 @@ elf(std::istream& stream)
   : detail::pimpl<elf_impl>{std::make_shared<elf_impl>(stream)}
 {}
 
+elf::
+elf(const void *data, size_t size)
+    : detail::pimpl<elf_impl>{std::make_shared<elf_impl>(data, size)}
+{}
```
stsoe, line 155: Actually, just do it here and remove the code in elf_impl.
```suggestion
elf::
elf(const void *data, size_t size)
    : elf{boost::interprocess::ibufferstream{static_cast<const_char*>(data), size}}
{}
```

stsoe (changes requested):

src/runtime_src/core/common/api/xrt_elf.cpp
sonals: This does not compile, because the elf(std::istream &strm) constructor takes a non const reference? If we make it const then we cannot read() from it :-(

```
/home/sonals/git/XRT/src/runtime_src/core/common/api/xrt_elf.cpp: In constructor ‘xrt::elf::elf(const void*, size_t)’:
/home/sonals/git/XRT/src/runtime_src/core/common/api/xrt_elf.cpp:153:33: error: cannot bind non-const lvalue reference of type ‘std::istream&’ {aka ‘std::basic_istream<char>&’} to an rvalue of type ‘std::basic_istream<char>’
  153 |     :  elf{boost::interprocess::ibufferstream{static_cast<const char*>(data), size}}
      |                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/sonals/git/XRT/src/runtime_src/core/common/api/xrt_elf.cpp:147:19: note:   initializing argument 1 of ‘xrt::elf::elf(std::istream&)’
  147 | elf(std::istream& stream)
      |     ~~~~~~~~~~~~~~^~~~~~
```

stsoe: Oh well, makes sense.

stsoe (approved):

PR #8869, opened 2025-04-10
Fix Compiler Warnings for GCC 14 Compatibility (Re-open)

KerwinTsaiii:
This commit addresses two compiler errors that arise with GCC 14's stricter warning settings:

1. A dangling reference warning when using `Boost::format`
2. An unused variable warning in the `concat_args` template function when instantiated with an empty argument list

The issues were discovered during compilation with GCC 14 which has enhanced warning detection compared to previous GCC versions. These aren't new bugs but rather existing code patterns that GCC 14 now identifies as potentially problematic. These issues were discovered while compiling on 

- MINISFOURM AI370 mini PC (Ryzen™ AI 9 HX 370)
- OS: Ubuntu 25.04 with kernel 6.14.0-13.
- Compiler: GCC 14.2.0 (Ubuntu 14.2.0-19ubuntu2)

Error messages:

1. `error: possibly dangling reference to a temporary [-Werror=dangling-reference]`
2. `error: variable 'first' set but not used [-Werror=unused-but-set-variable]`

1. For the dangling reference issue:

    - Changed const auto& `errMsg` to const auto `errMsg` to avoid referencing a temporary `Boost::format object`.
    - Alternative: Convert to string immediately with `.str()`, but this was rejected as unnecessary when simply removing the reference solves the issue more cleanly.

 2. For the unused variable issue:
     - Added `if constexpr (sizeof...(args) > 0)` condition to only declare and use the first variable when arguments are present.
     - Alternative: Create a template specialization for empty arguments, but this would duplicate more code than necessary (?).

1. Dangling reference fix: No risks? as this corrects a potential undefined behavior.
2. Unused variable fix: No risks? as the function behavior with empty arguments remains the same.

- Compiled with GCC 14 with `-Werror=dangling-reference` and `-Werror=unused-but-set-variable` flags enabled.
- Verified that both issues no longer trigger compiler errors.
- Functionality remains unchanged as these are correctness fixes rather than behavioral changes.

None. These are compiler warning fixes that don't change API behavior or usage.

### DIFF
--- a/src/runtime_src/core/pcie/linux/device_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/device_linux.cpp
@@ -674,7 +674,7 @@ struct clk_scaling_info
     std::vector<std::string> stat;
     boost::split(stat, keyValue, boost::is_any_of(":"));
     if (stat.size() != 2) {
-      const auto& errMsg = boost::format("Error: KeyValue pair doesn't meet expected format '<key>:<value>': '%s'") % keyValue;
+      const auto errMsg = boost::format("Error: KeyValue pair doesn't meet expected format '<key>:<value>': '%s'") % keyValue;
       throw std::runtime_error(errMsg.str());
     }
     return std::stoi(std::string(stat.at(1)));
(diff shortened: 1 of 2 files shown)


KerwinTsaiii:
Hi @stsoe, I have fixed the DCO auth issue.

AShivangi (approved):

PR #8870, opened 2025-04-10
AIESW-1793 : Add changes to patch function to support patching different buffer types

rbramand-xilinx:
Added changes to patch function  of module_int to patch different buffer types (ctrltext, ctrldata, save, restore etc)
Also added a function to get buffer size based on patch buffer type passed. These two functions are for internal testing at shim level.

Discovered when running preemption shim test case where patching of save/restore buffer is needed and the existing function only handles patching instruction buffer (ctrltext)

Added extra patch_buf_type arg that tells xrt about the buffer to be patched.
Also moved the enum to header file so that it can be accessed from shim level test application

Low as it is for internal testing and doesnt effect the application flow

NA

### DIFF
--- a/src/runtime_src/core/common/api/module_int.h
+++ b/src/runtime_src/core/common/api/module_int.h
@@ -15,6 +15,24 @@
 
 #include <string>
 
+namespace {
+// enum with different buffer types that supports patching
+// Some of the internal shim tests use this enum, so moving it to header
+// Ideal place for this enum would be in a header that has patching logic
+// TODO: Move this when patching logic is added to aiebu 
+enum class patch_buf_type {
+  ctrltext = 0,        // control code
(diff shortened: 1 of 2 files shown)


maxzhen (approved):
Thanks for adding the support!

One question: in the new APIs, what does 'index' argument mean?

rbramand-xilinx:
> Thanks for adding the support!
> 
> One question: in the new APIs, what does 'index' argument mean?

Hi Max, with the new ELF format there can be multiple .ctrltext/.ctrldata sections packed into single ELF, this index indicates which section we want to use. It takes a default value as 0 for older ELFs but when we move to new ELF we need to pass the proper index value. At present there are no ELFs with multiple sections of same type so it should not matter.

chvamshi-xilinx (approved):

PR #8872, opened 2025-04-10
Fix for CR-1235776 and CR-1236948

bisingha-xilinx:
https://jira.xilinx.com/browse/CR-1235776
https://jira.xilinx.com/browse/CR-1236948
CR-1236948: We dont used to validate the path before accessing it. 
CR-1235776: We used to register err callback for AIE. Somehow there is an error callback is coming for passing case also. For now, we're just ignoring the errors coming from hw.
CR-1236948: Added a proper check before accessing the path for the accel device. And printing proper message.
CR-1235776: As a temporary fix we're just ignoring the errors coming from the hw.
low
Tested on vck190 and vek280 with some testcases.
n/a


chvamshi-xilinx:
Merging this as this is EA blocker.

PR #8874, opened 2025-04-11
Revert "Add changes to partial sync xrt::module buffers after they ar…

AShivangi:
…e patched (#8815)"

This reverts commit e0b211154c52f9c313417f498e60ef72dcaf69e9.

This is causing performance degradation in SD

ATG testing

Low risk

N/A


rbramand-xilinx (approved):
The PR improved performance when we do multiple runs but added overhead with single run. We can revert this PR.
I will send a new one that doesn't degrade performance in single run. Thanks

PR #8875, opened 2025-04-11
Add changes to send only opcode kernel arg size in ert pkt payload data

rbramand-xilinx:
Added changes to send only info about first arg (opcode) in payload data of ert packet for which size of payload is updated.
This is done as elf patching is done at host side and we need not send all args info in ert packet in ELF based flows.

Adjusted the count in header to send only size of first arg in payload

Moderate need to test ELF flow on all the devices that support the flow

Tested on Strix Linux
Additional testing needs to be on Strix windows, Telluride
Marking the PR with do not merge label till this testing is done.

NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -2106,8 +2106,13 @@ class run_impl
         kcmd->opcode == ERT_START_NPU_PREEMPT_ELF) {
       auto payload_past_dpu = initialize_dpu(payload);
 
-      // adjust count to include the prepended ert_dpu_data structures
-      kcmd->count += payload_past_dpu - payload;
+      // Adjust count to include the prepended ert_dpu_data structures
+      // Also, for ELF flow we dont need kernel args info in cmd payload
+      // as args are patched at host side, only the first arg(opcode)
+      // info is sent in this case.
+      // opcode is uint64_t so (2 * uint32_t) is the size in payload
+      kcmd->count += payload_past_dpu - payload - kernel->get_regmap_size()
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx (approved):

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -2106,8 +2106,13 @@ class run_impl
         kcmd->opcode == ERT_START_NPU_PREEMPT_ELF) {
       auto payload_past_dpu = initialize_dpu(payload);
 
-      // adjust count to include the prepended ert_dpu_data structures
-      kcmd->count += payload_past_dpu - payload;
+      // Adjust count to include the prepended ert_dpu_data structures
+      // Also, for ELF flow we dont need kernel args info in cmd payload
+      // as args are patched at host side, only the first arg(opcode)
+      // info is sent in this case.
+      // opcode is uint64_t so (2 * uint32_t) is the size in payload
```
stsoe, line 2113:
```suggestion
      // Adjust count to include the prepended ert_dpu_data structures
      // Also, for ELF flow we dont need kernel args info in cmd payload
      // as args are patched at host side, only the first arg(opcode)
      // info is sent in this case and it is written into the command register
      // map at offset 0x0.   The command count is initialized earlier to 
      // the size the command register map plus cu masks.
      // opcode is uint64_t so (2 * uint32_t) is the size in payload so
      // subtract register map size and add 2.
```

larry9523 (approved):
Let's do a completed test on Linux (@NishadSaraf ) and Windows (@pranjalv-xilinx ) before merging.

stsoe (commented):
Minor edits to tweak the magic explanation :-)

larry9523:
Merge it and start testing.

PR #8890, opened 2025-04-21
Add changes to partial sync xrt::module buffers after they are patched

rbramand-xilinx:
This PR is improvement over PR - https://github.com/Xilinx/XRT/pull/8815
xrt::module buffers like instruction, control pkt are synced only once for the first time and in subsequent runs only parts of buffer where patching happens is synced to reduce the overhead of buffer sync. This improves performance.

A boolean is used to indicate if its first run where entire buffer is synced and in following runs only part of buffer is synced

Moderate and needs proper testing as all the ELF flows can be affected.

Testing is pending, so marking the PR with do not merge label

NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -189,8 +189,9 @@ struct patcher
     , m_ctrlcode_patchinfo(std::move(ctrlcode_offset))
   {}
 
+private:
   void
-  patch64(uint32_t* data_to_patch, uint64_t addr)
+  patch64(uint32_t* data_to_patch, uint64_t addr) const
   {
     *data_to_patch = static_cast<uint32_t>(addr & 0xffffffff);
     *(data_to_patch + 1) = static_cast<uint32_t>((addr >> 32) & 0xffffffff);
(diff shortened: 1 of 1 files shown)


stsoe (approved):
I assume this doesn't change much over the previous PR.  I am not excited about the conditional code in the templated function, but don't have a better suggestion right now.

larry9523 (changes requested):
I think we can make this simpler by
 1. mark the buffer dirty right after we allocate and fill in data 
 2. when set_arg is called, where we patch a single argument, if the buffer is dirty, don't do partial sync. if the buffer is clean, then do a partial sync and keep the buffer in clean state
 3. right before we send the command xrt::module_int::sync, if the buffer is dirty, sync the whole buffer. skipping sync if the buffer is clean.

rbramand-xilinx:
> I think we can make this simpler by
> 
> 1. mark the buffer dirty right after we allocate and fill in data
> 2. when set_arg is called, where we patch a single argument, if the buffer is dirty, don't do partial sync. if the buffer is clean, then do a partial sync and keep the buffer in clean state
> 3. right before we send the command xrt::module_int::sync, if the buffer is dirty, sync the whole buffer. skipping sync if the buffer is clean.

Hi @larry9523, in the existing design the dirty bit is for all the buffers and it is not specific to single buffer. So when set_arg is called if dirty bit is set we are syncing all the buffers. But with my changes it simplifies things and doesnt sync all the buffers and sync specific buffer in patcher patch calls in subsequent calls.
That said we can have wrapper bo class that can hold dirty bit to do it as you are suggesting. That can be done during refactoring of code.

larry9523:
> > I think we can make this simpler by
> > 
> > 1. mark the buffer dirty right after we allocate and fill in data
> > 2. when set_arg is called, where we patch a single argument, if the buffer is dirty, don't do partial sync. if the buffer is clean, then do a partial sync and keep the buffer in clean state
> > 3. right before we send the command xrt::module_int::sync, if the buffer is dirty, sync the whole buffer. skipping sync if the buffer is clean.
> 
> Hi @larry9523, in the existing design the dirty bit is for all the buffers and it is not specific to single buffer. So when set_arg is called if dirty bit is set we are syncing all the buffers. But with my changes it simplifies things and doesnt sync all the buffers and sync specific buffer in patcher patch calls in subsequent calls. That said we can have wrapper bo class that can hold dirty bit to do it as you are suggesting. That can be done during refactoring of code.

OK. If that is the case, let's get this merged and refactor it in next PR.

larry9523 (approved):

PR #8906, opened 2025-04-23
XRT-SMI Configure and Examine migration

aktondak:
This change migrates xrt-smi configure to the new shim based option setting flow. This change also separates out the flow between xbmgmt and xrt-smi since currently, they were both using same code (SubCmd****Internal[.cpp][.h]). Now all xrt-smi subcommands are moved to new flow with subcommand files present under xbutil2 directory. Ideally we should change the subdirectory names as well to xrt-smi. 
If need be in future, we should migrate the xbmgmt subcommands as well. 

https://jira.xilinx.com/browse/AIESW-1737

An alternate to separating out xbmgmt and xrt-smi was to keep them together and migrate only xrt-smi which proved to be untidy and cumbersome.

None

Tested all xrt-smi examine and xrt-smi configure commands on krck01

none


stsoe (approved):

rchane (approved):

AShivangi (approved):
Looks good!
Can you please check the copyright dates and if they need to be updated?

I believe only SubCmdAdvanced is left rn. Once we move that to the shim we can get rid of the json file in xbutil.cpp?

aktondak:
> Looks good! Can you please check the copyright dates and if they need to be updated?
> 
> I believe only SubCmdAdvanced is left rn. Once we move that to the shim we can get rid of the json file in xbutil.cpp?

Once SubCmdAdvanced is done. We'll need to make another change where we get the list of subcommands from shim config itself, and then we can remove the xbutil.cpp embedded json. I'm thinking of separating out xrt-smi and xbmgmt there as well (writing a different main_ for xrt-smi) instead of relying on the common main in XBMain.cpp. Will make this change in the next PR.

AShivangi:
> > Looks good! Can you please check the copyright dates and if they need to be updated?
> > I believe only SubCmdAdvanced is left rn. Once we move that to the shim we can get rid of the json file in xbutil.cpp?
> 
> Once SubCmdAdvanced is done. We'll need to make another change where we get the list of subcommands from shim config itself, and then we can remove the xbutil.cpp embedded json. I'm thinking of separating out xrt-smi and xbmgmt there as well (writing a different main_ for xrt-smi) instead of relying on the common main in XBMain.cpp. Will make this change in the next PR.

Sounds good! 
Once this is done, can you please create a confluence page w/ a UML diagram explaining the new architecture? It might be helpful in the future.

chvamshi-xilinx (approved):

PR #8919, opened 2025-04-29
CMake: Win: use find_package(Boost CONFIG) lastest cmake

wendyliang25:
Since > CMake 3.29, FindBoost module is removed, it looks for BoostConfig.cmake instead. And thus, we need to change our CMakefile to find boost with find_package to search for BoostConfig.cmake instead. For details, please see https://cmake.org/cmake/help/latest/policy/CMP0167.html

This patch is to support CMake 3.30 and above to find Boost package.
`FindBoost` module is removed from CMake 3.30 and above. I installed CMake 4.0 on my Windows11 system, it complains:
```
CMake Warning (dev) at src/xrt/src/CMake/nativeWin.cmake:29 (find_package):
  Policy CMP0167 is not set: The FindBoost module is removed.
```
We need to change CMakeFil to use `find_package(Boost CONFIG)` instead of 
```
include(FindBoost)
find_package(Boost)
```
to find Boost package.

Installed CMake 4.0 on Windows11 when this issue was discovered.

Check if the CMake version is > `3.29`, if yes, use `find_package(Boost CONFIG)` instead of `find_package(Boost)` and removed `include(FindBoost)`.
If you just removed `include(FindBoost)`, CMake can find the Boost module, however, it will reports this warning:
```
CMake Warning (dev) at src/xrt/src/CMake/boostUtil.cmake:25 (find_package):
  Policy CMP0167 is not set: The FindBoost module is removed.  Run "cmake
  --help-policy CMP0167" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.
```
To remove the above warning, will need to change to use `find_package(Boost CONFIG)`.

Alternatives, make our project to only support CMake < = 3.29.

Built on Windows11 with CMake 4.0 installed. requires additional testing with older CMake version which CI is supposed to be used.

### DIFF
--- a/src/CMake/boostUtil.cmake
+++ b/src/CMake/boostUtil.cmake
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
-# Copyright (C) 2024 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
 
 # Support building XRT or its components with local build of Boost libraries. 
 # In particular the script runtime_src/tools/script/boost.sh downloads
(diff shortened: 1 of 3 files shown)


stsoe (commented):
Would it be possible to include boostUtil.cmake in nativeWin.cmake just like nativeLnx.cmake does?

wendyliang25:
> Would it be possible to include boostUtil.cmake in nativeWin.cmake just like nativeLnx.cmake does?

Hi @stsoe , i have updated the PR to change `nativeWin.cmake` to use `boostUtil.cmake`, could you check the updated patches set? thanks

stsoe (approved):
LGTM

rozumx (approved):
Approved.

Please note that the copyright dates need to be updated.

PR #8923, opened 2025-05-01
tools: Add windows helper functions header

wendyliang25:
This patch is to introduce a header to hold Windows helper functions. It is supposed to define commonly used helper functions or macros, such as function to get the error message from last error calling Windows API.

None

If not using the header, then every file will need to implement duplicated code to achieve some common function when it is required such as getting the error message from last Windowns API call.

None

Tested with `xbtracer` with non-exist executable:
```
PS C:\Users\wendlian\test\WDebug\xrt> .\xbtracer .\unwrapped\xrt

C:\Users\wendlian\test\WDebug\xrt>REM SPDX-License-Identifier: Apache-2.0

C:\Users\wendlian\test\WDebug\xrt>REM Copyright (C) 2024-2025 Advanced Micro Devices, Inc. All rights reserved.
Failed to launch - Reason [xbtracer.exe] F: Child process creation failed. Error: The system cannot find the file specified.
. Aborted!
```

### DIFF
--- a/src/runtime_src/core/tools/xbtracer/src/app/launcher.cpp
+++ b/src/runtime_src/core/tools/xbtracer/src/app/launcher.cpp
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright (C) 2024 Advanced Micro Devices, Inc. All rights reserved.
+// Copyright (C) 2024 - 2025 Advanced Micro Devices, Inc. All rights reserved.
 
 #include <array>
 #include <chrono>
@@ -16,10 +16,12 @@
 # include "getopt.h"
 # include <shlwapi.h>
 # include <windows.h>
+# include "core/common/windows/win_utils.h"
 #else
 # include <sys/stat.h>
 # include <unistd.h>
 # include <cstring>
+# include "core/common/linux/linux_utils.h"
 #endif /* #ifdef _WIN32 */
 
 #ifdef _WIN32
@@ -529,7 +531,8 @@ int create_child_proc_as_suspended(launcher& app)
       &app.m_si,                    // Pointer to STARTUPINFO structure
       &app.m_pi))          // Pointer to PROCESS_INFORMATION structure
   {
-    log_f("Child process creation failed");
+    log_f("Child process creation failed. Error: ",
+          sys_dep_get_last_err_msg().c_str());
     return -1;
   }
 


AShivangi (approved):

src/runtime_src/core/common/linux/linux_utils.h
```diff
@@ -0,0 +1,50 @@
...
+ * @note This file is specific to Linux platforms and relies on the Linux
+ *       API. It is not portable to other operating systems.
+ *
+ * @copyright
+ * SPDX-License-Identifier: Apache-2.0  
+ * Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+ */
+
+#ifndef _linux_utils_h_
+#define _linux_utils_h_
```
stsoe, line 25: No leading underscore in guards and variables

src/runtime_src/core/common/linux/linux_utils.h
```diff
@@ -0,0 +1,50 @@
...
+ * This function uses the Linux API to obtain a human-readable error message
+ * corresponding to the last error code set by the system. It retrieves the error
+ * code using `GetLastError()` and formats the message using `FormatMessage()`.
+ *
+ * @return A string containing the error message associated with the last error code.
+ *         If no error has occurred, the returned string may be empty.
+ */
+static inline std::string sys_dep_get_last_err_msg() {
+    return  std::string(strerror(errno));
+}
```
stsoe, line 45:
```suggestion
inline std::string 
sys_dep_get_last_err_msg() 
{
  return  std::string(strerror(errno));
}
```

I am trying to keep everything under core/common left centered.  To me it is easier to scan code when mostly lined up to the left.  This is just my personal style, which I try to enforce under core/common.

```
[qualifier] return_type
[class_name]::
function_name()
{
  ...;
}
```

src/runtime_src/core/common/utils.h
```diff
@@ -142,6 +147,20 @@ XRT_CORE_COMMON_EXPORT
...
+/**
+ * @brief Retrieve the last system error message.
+ *
+ * This function fetches the last error message from the operating system,
+ * which could originate from a Linux or Windows system call. It provides
+ * a human-readable string describing the error.
+ *
+ * @return A string containing the last system error message.
+ */
+inline std::string get_sys_last_err_msg()
```
stsoe, line 159: Let's make non-inline.  Move the definition to utils.cpp

```suggestion
XRT_CORE_COMMON_EXPORT
std::string 
get_sys_last_err_msg();
```

src/runtime_src/core/common/utils.h
```diff
@@ -20,6 +20,11 @@
 
 #include "config.h"
 #include "scope_guard.h"
+#ifdef _WIN32
+#include "core/common/windows/win_utils.h"
+#else
+#include "core/common/linux/linux_utils.h"
+#endif
```
stsoe, line 27: Let's move to cpp, see next comment

src/runtime_src/core/common/windows/win_utils.h
```diff
@@ -0,0 +1,62 @@
...
+ * @note This file is specific to Windows platforms and relies on the Windows
+ *       API. It is not portable to other operating systems.
+ *
+ * @copyright
+ * SPDX-License-Identifier: Apache-2.0  
+ * Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+ */
+
+#ifndef _win_utils_h_
+#define _win_utils_h_
```
stsoe, line 26:
```suggestion
#ifndef win_utils_h
#define win_utils_h
```

src/runtime_src/core/tools/xbtracer/src/app/launcher.cpp
```diff
@@ -16,6 +16,7 @@
 # include "getopt.h"
 # include <shlwapi.h>
 # include <windows.h>
+# include "core/common/utils.h"
```
stsoe, line 19: Shouldn't have to be guarded.

src/runtime_src/core/common/windows/win_utils.h
```diff
@@ -0,0 +1,62 @@
+
```
stsoe, line 1: // SPDX-License-Identifier: Apache-2.0
// Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.

src/runtime_src/core/common/linux/linux_utils.h
```diff
@@ -0,0 +1,50 @@
+
```
stsoe, line 1: // SPDX-License-Identifier: Apache-2.0
// Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.

src/runtime_src/core/common/linux/linux_utils.h
```diff
@@ -0,0 +1,50 @@
...
+ * SPDX-License-Identifier: Apache-2.0  
+ * Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+ */
+
+#ifndef _linux_utils_h_
+#define _linux_utils_h_
+
+#ifdef __linux__
+
+#include <cerrno>
```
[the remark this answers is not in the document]
stsoe: Strange

src/runtime_src/core/common/windows/win_utils.h
```diff
@@ -0,0 +1,62 @@
...
+        NULL,
+        error_code,
+        MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+        (LPTSTR)&error_msg,
+        0,
+        NULL);
+    std::string error_message = static_cast<char*>(error_msg);
+    LocalFree(error_msg);
+    return error_message;
+}
```
stsoe, line 57:
```suggestion
inline std::string 
sys_dep_get_last_err_msg() 
{
  DWORD error_code = GetLastError();
  LPVOID error_msg;
  FormatMessage(
        FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
      NULL,
      error_code,
      MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
      (LPTSTR)&error_msg,
      0,
      NULL);
  std::string error_message = static_cast<char*>(error_msg);
  LocalFree(error_msg);
  return error_message;
}
```

stsoe (changes requested):
Few suggestions.

wendyliang25:
> Few suggestions.

Thanks @stsoe , i have updated the code based on your comments, could you check it again? thanks

src/runtime_src/core/tools/xbtracer/CMakeLists.txt
```diff
@@ -39,6 +39,11 @@ if (XRT_STATIC_BUILD)
     RUNTIME DESTINATION ${XRT_INSTALL_UNWRAPPED_DIR} COMPONENT ${XRT_BASE_COMPONENT})
 endif()
 
+# This application uses helper functions provided in core/common (xrt_coreutil lib)
+target_link_libraries(${XBTRACER_NAME}
+  PRIVATE
+  xrt_coreutil
+)
```
stsoe, line 46: Ahh,  do you think the idea with the launcher was that it can be used with end user application without the need of XRT?   With this change it no longer can.  So, while your XRT changes are good, maybe if intend is no XRT dependencies?   If this is the case, then maybe just include win_utils.h in launcher.cpp just as you had to begin with only from its new location?

wendyliang25: Hi @stsoe , right, sorry that i forgot about it, while focusing on passing compilation, yes, so the solution is:
* Opt1: use inline in the header
* Opt2: use static link to the xrt_coreutil (not sure if we can only statically link, will see how to do it in CMake, or make the whole `xbtrace`r static link only
* Opt3: specify utils .cpp/h as components to compile in `xbtracer`
* Opt4: as what you suggested, only include `win_utils.h` in `launcher.cpp`

Could you suggest which is the most appropriate choice ? thanks

stsoe: You already have inline in header (1)  so I say we go for (4) and avoid changing linking of the launcher to depend on XRT at runtime.

wendyliang25: Thanks @stsoe , i have updated the code to use Opt4. could you check again? thanks

PR #8927, opened 2025-05-02
CR-1227928: Fix Alveo hardware emulation segmentation fault

jvillarre:
Profiling keeps a shared pointer to an xrt::device in order to access the PL portion.  In Alveo hardware emulation, at the end of the application when our objects are being destroyed, the shared_pointer is destroyed and the xrt::device attempts to clean itself up, but this leads to a problem as the simulation has already closed and the xrt::device cannot cleanly be removed.  

The seg fault was discovered through regression testing.

This pull request makes it so profiling releases its shared pointer reference before the simulation is stopped, by catching it in the hook in hardware emulation shim destructor.

Low risk as this is only called at the end of Alveo hardware emulation.

The original failing test has been verified.

No documentation impact.


IshitaGhosh (approved):
Changes look good.

PR #8935, opened 2025-05-06
Enabling AIE Reconfig overhead test to only run in elf flow

aktondak:
This PR updates the aie-reconfig-overhead test to run in elf mode only for it to be automatically enabled on linux.
James has enabled this test in elf mode on linux already. Thus to migrate this test to linux for parity, we should enable it by default in elf mode for window<--> Linux compatibility.

https://jira.xilinx.com/browse/VITIS-14542
Opened through linux windows parity work for xrt-smi

Solved through enabling this test in elf mode only

None

Tested on xsjstrix45 linux machine :
```
Test 1 [0000:c5:00.1]     : aie-reconfig-overhead                               
    Details               : Array reconfiguration overhead: 3.9 ms
```

None

### DIFF
--- a/src/runtime_src/core/tools/common/tests/TestAIEReconfigOverhead.cpp
+++ b/src/runtime_src/core/tools/common/tests/TestAIEReconfigOverhead.cpp
@@ -32,19 +32,7 @@ TestAIEReconfigOverhead::run(std::shared_ptr<xrt_core::device> dev)
   boost::property_tree::ptree ptree = get_test_header();
   ptree.erase("xclbin");
 
-  // Check Whether Use ELF or DPU Sequence
-  auto elf = XBValidateUtils::get_elf();
-  std::string xclbin_path; 
-  
-  if (!elf) {
-    xclbin_path = XBValidateUtils::get_xclbin_path(dev, xrt_core::query::xclbin_name::type::validate, ptree);
-    if (XBU::getVerbose())
-      XBValidateUtils::logger(ptree, "Details", "Using DPU Sequence");
-  } else {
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/tools/common/tests/TestAIEReconfigOverhead.cpp
```diff
@@ -67,81 +55,35 @@ TestAIEReconfigOverhead::run(std::shared_ptr<xrt_core::device> dev)
...
-  else {
-    bo_ifm = xrt::ext::bo{working_dev, buffer_size};
-    bo_ofm = xrt::ext::bo{working_dev, buffer_size};
-    bo_inter = xrt::ext::bo{working_dev, inter_size};
-    bo_mc = xrt::ext::bo{working_dev, 16};
-  }
+  bo_ifm = xrt::ext::bo{working_dev, buffer_size};
+  bo_ofm = xrt::ext::bo{working_dev, buffer_size};
+  bo_inter = xrt::ext::bo{working_dev, inter_size};
+  bo_mc = xrt::ext::bo{working_dev, 16};
```
AShivangi, line 86: This is not needed

AShivangi (changes requested):
Can you please double check that dpu flow on windows, elf flow on windows and linux show the same number?

aktondak:
> Can you please double check that dpu flow on windows, elf flow on windows and linux show the same number?

Looking at the numbers reported by Kelly, it looks consistent :

image

AShivangi (approved):

PR #8938, opened 2025-05-08
xdna migration of userspace in submodule xrt

SravanKumarAllu-xilinx:

As part of "XDNA migration of userspace" (moving Telluride VE2 shim under XDNA repo, added changes which should be part of xrt (submodule of XDNA repo) , these are partial changes , full changes will be enabled when complete changes are raised in Telluride , XDNA repo's

N/A

We can maintain single public repo of "XDNA" for the development of Telluride set of devices rather than maintaining different repo's.

N/A

bo_sharing import-export import-export-across-process mem_ops multi-layer multi_partition multi_runs runlist simple_int8 vadd

N/A

### DIFF
--- a/src/runtime_src/core/edge/user/system_linux.cpp
+++ b/src/runtime_src/core/edge/user/system_linux.cpp
@@ -36,11 +36,9 @@
   #define MACHINE_NODE_PATH ""
 #endif
 
-#ifdef EDGE_VE2
-#include "shim/device.h"
-#endif
-
-#ifdef EDGE_VE2_XDNA
+#ifdef XDNA_VE2
+#include "shim_ve2/xdna_device.h"
+#elif defined(EDGE_VE2_XDNA)
 #include "shim/xdna_device.h"
 #endif
 
--- a/src/runtime_src/xdp/profile/device/aie_trace/ve2/aie_trace_offload_ve2.cpp
+++ b/src/runtime_src/xdp/profile/device/aie_trace/ve2/aie_trace_offload_ve2.cpp
@@ -25,8 +25,13 @@
 #include "core/common/shim/hwctx_handle.h"
 #include "core/include/xrt/xrt_hw_context.h"
 #include "core/common/api/hw_context_int.h"
+#ifdef XDNA_VE2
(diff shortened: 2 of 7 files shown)


stsoe (approved):

src/runtime_src/core/edge/user/system_linux.cpp
```diff
@@ -44,6 +44,10 @@
 #include "shim/xdna_device.h"
 #endif
 
+#ifdef XDNA_VE2 // this is temporary flag , will be removed once "shim_ve2" is moved under XDNA/src
+#include "shim_ve2/xdna_device.h"
+#endif
```
chvamshi-xilinx, line 49: As discussed offline, 
Remove EDGE_VE2 flag & add EDGE_VE2_XDNA as else of XDNA_VE2.
Once we move to public xdna, we should cleanup these flags.

maxzhen: > As discussed offline, Remove EDGE_VE2 flag & add EDGE_VE2_XDNA as else of XDNA_VE2. Once we move to public xdna, we should cleanup these flags.

Hi @chvamshi-xilinx ,

Are you talking about removing the flag but leave the code here in XRT? Or do you plan to move entire ve2 shim to xdna-driver repo later? @sonals , @stsoe and me discussed offline and it seems that xdna-driver repo would be a better place for ve2 shim code. Can we just do it now instead of leaving them in XRT?

chvamshi-xilinx (approved):

src/runtime_src/core/edge/user/system_linux.cpp
stsoe: @chvamshi-xilinx .  I don't understand the intent here?  XRT referencing shim specific code is not right.  There should not be any references to xdna, nor any inclusion of xdna repo code.

chvamshi-xilinx:
Hi @maxzhen / @stsoe ,
As discussed offline. Will reimplement the dynamic discovery logic based on plugin mechanism. Closing this PR.

PR #8940, opened 2025-05-08
fix patch() API for shim level testing

maxzhen:
The existing patch() API does not patch save and restore instruction buffer correctly. This PR fixes it.
Low risk since this API is not called in normal XRT flow.
Tested with the shim level test which is also pending to be checked in.


larry9523 (approved):

rbramand-xilinx (approved):
The approach I was discussing in the mail avoids switch case and calling get_* functions again but that said these changes looks simple and avoids updating shim tests.

chvamshi-xilinx (approved):

PR #8953, opened 2025-05-13
fix ELF test cases in shim test

maxzhen:
Fix ELF flow test in shim test on NPU on Linux.
https://github.com/Xilinx/XRT/pull/8890 introduced this issue.
The function signature is not right which force compiler to direct the call to a wrong member function.
Low, only impact shim test
Tested on NPU with shim test.
N/A

### DIFF
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -392,7 +392,7 @@ private:
 
 public:
   void
-  patch_it(uint8_t* base, uint64_t value)
+  patch_it(uint8_t* base, uint64_t value, bool first)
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -392,7 +392,7 @@ struct patcher
 
 public:
   void
-  patch_it(uint8_t* base, uint64_t value, bool first)
+  patch_it(uint8_t* base, uint64_t value, bool)
```
rbramand-xilinx, line 395: may be use bool /*first*/

rbramand-xilinx (approved):

PR #8956, opened 2025-05-14
Make control code id or kernel instance as string in ELF flow

rbramand-xilinx:
Made kernel instance or control code id as string instead of uint32_t in ELF flow.
This is done in accordance with change in ELF spec where a kernel like DPU can have instance name as "abc" instead of just having id numbers.

Low

Tested with new ELF flow on SNL flow and txn cp test case on strix linux hw

NA

### DIFF
--- a/src/runtime_src/core/include/xrt/experimental/xrt_module.h
+++ b/src/runtime_src/core/include/xrt/experimental/xrt_module.h
@@ -95,25 +95,13 @@ public:
   XRT_API_EXPORT
   module(const xrt::module& parent, const xrt::hw_context& hwctx);
 
-  /**
-   * module() - Constructor associate module with hardware context
-   *
-   * @param parent
-   *   Parent module with instruction buffer to move into hwctx
-   * @param hwctx
-   *   Hardware context to associate with module
-   * @param ctrl_code_idx
-   *   index of control code inside the parent module
-   *
-   * Copy content of existing module into an allocation associated
-   * with the specified hardware context.
-   * If module has multiple control codes, ctrl code id is used to
-   * identify the control code that needs to be run.
-   *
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/include/xrt/experimental/xrt_module.h
```diff
@@ -107,13 +107,13 @@ class module : public detail::pimpl<module_impl>
...
-   * If module has multiple control codes, index is used to identify
-   * the control code that needs to be run.
+   * If module has multiple control codes, ctrl code id is used to
+   * identify the control code that needs to be run.
    *
    * Throws if module is not compatible with hardware context
    */
   XRT_API_EXPORT
-  module(const xrt::module& parent, const xrt::hw_context& hwctx, uint32_t ctrl_code_idx);
+  module(const xrt::module& parent, const xrt::hw_context& hwctx, std::string ctrl_code_id);
```
stsoe, line 116: We are sure no binary in-the-field is using the removed ctor, right?

rbramand-xilinx: This constructor is being called from copy_module function from xrt_kernel.cpp and no application code must be using it. So I guess this change should be safe Soren.

stsoe: I see, if it is internal, can't it be in module_int ?

rbramand-xilinx: Yes Soren, removed the unused constructor from being public and gave undocumented version of construction using impl handle. Please review

stsoe (commented):
Minor suggestions

stsoe (approved):

PR #8963, opened 2025-05-15
Updating profiling to reflect the current Native XRT APIs being tracked

jvillarre:
When host level profiling is enabled, the profiling library keeps statistics on each API that is called.  In order to create a summary table at the end and correctly identify if a function is a Native XRT API, an OpenCL API, or another category, profiling keeps a table of known API names.  This pull request updates the table of known API names to reflect the current functions that are being profiled.

Issue was discovered through internal testing and inspection.

Low risk, as this will only affect the final table in the profile summary.

As we're now catching additional APIs and adding them to the profiling summary table, the current suite of internal tests will have to update their golden files in cases where the newly monitored APIs appear.

No documentation impact.


IshitaGhosh (approved):
Changes look good.

PR #8964, opened 2025-05-16
Added daemon suboption as default configure subcommand

ManojTakasi:
https://jira.xilinx.com/browse/CR-1240855 The --daemon suboption was being reported as an unrecognized command when installing the latest XRT package without a device present. This issue does not occur when a device is available.

image-2025-05-15-13-26-16-295

The issue arises because SmiDefault.cpp is used when no device is detected. This file does not include  --daemon command leading to the observed error.

The fix ensures that the --daemon suboption is recognized even when no device is present.

low

Tested by building and running xbmgmt configure --daemon command with and without device.

NA

### DIFF
--- a/src/runtime_src/core/tools/common/SmiDefault.cpp
+++ b/src/runtime_src/core/tools/common/SmiDefault.cpp
@@ -56,6 +56,7 @@ create_configure_subcommand()
   std::map<std::string, std::shared_ptr<xrt_core::smi::option>> configure_suboptions;
   configure_suboptions.emplace("device", std::make_shared<xrt_core::smi::option>("device", "d", "The Bus:Device.Function (e.g., 0000:d8:00.0) device of interest", "common", "", "string"));
   configure_suboptions.emplace("help", std::make_shared<xrt_core::smi::option>("help", "h", "Help to use this sub-command", "common", "", "none"));
+  configure_suboptions.emplace("daemon", std::make_shared<xrt_core::smi::option>("daemon", "", "Update the device daemon configuration", "hidden", "", "none"));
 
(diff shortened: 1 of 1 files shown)


aktondak:
> deamon will show up in --advanced even for RyzenAI which might cause confusion.. @ManojTakasi, what's the impact of this issue?
> 
> @aktondak daemon is an xbmgmt option, when are you planning to separate xrt-smi and xbmgmt and will this issue be fixed by that?

xrt-smi and xbmgmt are separated out for all subcommands. xbmgmt still follows the old flow of option addition. This should not go into smiDefault as it is meant only for xrt-smi options. 
@ManojTakasi Let me check why this issue is coming and connect with you.

chvamshi-xilinx:
> > deamon will show up in --advanced even for RyzenAI which might cause confusion.. @ManojTakasi, what's the impact of this issue?
> > @aktondak daemon is an xbmgmt option, when are you planning to separate xrt-smi and xbmgmt and will this issue be fixed by that?
> 
> xrt-smi and xbmgmt are separated out for all subcommands. xbmgmt still follows the old flow of option addition. This should not go into smiDefault as it is meant only for xrt-smi options. @ManojTakasi what command is being run here ?Are we sure its xbmgmt configure and not xrt-smi configure? is it a pcie based flow or edge ?

Hi @aktondak .
This must be "xbmgmt configure" coming from postinstallation script
https://github.com/Xilinx/XRT/blob/master/src/CMake/config/postinst.in#L119
@ManojTakasi , can you double check whether this issue fixed the original issue.,

ManojTakasi:
Apologies for the confusion,  this issue is present in version 2025.1, not in the master branch.

PR #8972, opened 2025-05-20
support dynamic tracing new api

xuhz:
1. when app hangs, the already collected dtrace data should still be dumped
2. change to support new dtrace lib ApIs
3. leverage new config_bo() api
1. when timeout happens, dump dtrace result first then return
2. use new APIs
low
since driver change and/or fw change are not ready, so there is no e2e test done
but I test with some stub code.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -2519,12 +2519,17 @@ public:
   [[nodiscard]] ert_cmd_state
   wait(const std::chrono::milliseconds& timeout_ms) const
   {
+    // dump dtrace buffer if ini option is enabled
+    static auto dtrace_lib_path = xrt_core::config::get_dtrace_lib_path();
+
     ert_cmd_state state {ERT_CMD_STATE_NEW}; // initial value doesn't matter
     if (timeout_ms.count()) {
       auto [ert_state, cv_status] = cmd->wait(timeout_ms);
       if (cv_status == std::cv_status::timeout) {
-        state = ERT_CMD_STATE_TIMEOUT;
-	goto done;
+        if (!dtrace_lib_path.empty()) {
+          xrt_core::module_int::dump_dtrace_buffer(m_module);
+	}
+        return ERT_CMD_STATE_TIMEOUT;
       }
 
       state = ert_state;
@@ -2538,9 +2543,6 @@ public:
     if (dump)
       xrt_core::module_int::dump_scratchpad_mem(m_module);
 
-done:  
-    // dump dtrace buffer if ini option is enabled
-    static auto dtrace_lib_path = xrt_core::config::get_dtrace_lib_path();
     if (!dtrace_lib_path.empty())
       xrt_core::module_int::dump_dtrace_buffer(m_module);
 
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -2520,13 +2520,13 @@ public:
   wait(const std::chrono::milliseconds& timeout_ms) const
   {
     // dump dtrace buffer if ini option is enabled
-    static auto dtrace_lib_path = xrt_core::config::get_dtrace_lib_path();
+    static bool dtrace = !xrt_core::config::get_dtrace_lib_path();
 
     ert_cmd_state state {ERT_CMD_STATE_NEW}; // initial value doesn't matter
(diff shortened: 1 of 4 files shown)


chvamshi-xilinx (approved):
Looks good

rbramand-xilinx (approved):
Looks Good, thanks for making changes Brian.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -2519,11 +2519,18 @@ class run_impl
   [[nodiscard]] ert_cmd_state
   wait(const std::chrono::milliseconds& timeout_ms) const
   {
+    // dump dtrace buffer if ini option is enabled
+    static auto dtrace_lib_path = xrt_core::config::get_dtrace_lib_path();
```
stsoe, line 2523: Please make it a bool. 
```suggestion
    static bool dtrace = !xrt_core::config::get_dtrace_lib_path().empty();
```

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -2142,67 +2141,75 @@ class module_sram : public module_impl
...
     // required data, buffers are synced after data is filled.
     try {
+      std::vector<uint64_t> buffers(buffers_length);
+      get_dtrace_buffer_size(buffers.data());
+
+      std::map<uint32_t, size_t> buf_sizes;
+      size_t total_size = 0;
+
+      const uint32_t MASK32 = 0xffffffff;
+      const uint32_t SHIFT32 = 32;
```
stsoe, line 2195: Never ever ALLCAPS except for macros.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es9-avoid-all_caps-names

Plus these should be compile time constexpr

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -2142,67 +2141,75 @@ class module_sram : public module_impl
...
+      for (const auto& entry : buffers)
+      {//for each entry, lower 32 is the uc index, and upper 32 is the length in word for that uc
```
stsoe, line 2197:
```suggestion
      for (const auto& entry : buffers) {
        //for each entry, lower 32 is the uc index, and upper 32 is the length in word for that uc
```

src/runtime_src/core/common/api/xrt_kernel.cpp
xuhz: sure. will change

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -2535,9 +2542,7 @@ class run_impl
     static bool dump = xrt_core::config::get_feature_toggle("Debug.dump_scratchpad_mem");
     if (dump)
       xrt_core::module_int::dump_scratchpad_mem(m_module);
-    
-    // dump dtrace buffer if ini option is enabled
-    static auto dtrace_lib_path = xrt_core::config::get_dtrace_lib_path();
+
     if (!dtrace_lib_path.empty())
```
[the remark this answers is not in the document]
xuhz: sure. will change

src/runtime_src/core/common/api/xrt_module.cpp
xuhz: sure. will change

src/runtime_src/core/common/api/xrt_module.cpp
xuhz: sure. will change

xuhz:
> initialize_dtrace_buf() function is not quite clean, it is initializing data members but that is not proper. Addtionally the function is too long.
> 
> The constructor should initialize the data members in its initializer list. It should be like this:
> 
> ```
> ctor() 
>   : m_member(blah)
>   , m_dtrace_ctrl_bo{initialize_dtrace_ctrl_bo()}
>   , m_dtrace_mem_bo{initialize_dtrace_mem_bo()}
>   , etc.
> {}
> ```
> 
> The initialize_dtrace_...() functions can of course return an default bo if dtrace is not enabled.

agree :) but this should be in this PR.
This PR just follows the existing style -- there are so many long functions in that class...

stsoe:
> agree :) but this should be in this PR.
> This PR just follows the existing style -- there are so many long functions in that class...

Right, it need not be in this PR.  It was a general comment on poorly organized code.

xuhz:
retest this please

PR #8973, opened 2025-05-21
update aiebu commit to support aie4

HimanshuChoudhary-Xilinx:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,8 +9,8 @@
 	url = https://github.com/microsoft/GSL.git
 [submodule "src/runtime_src/core/common/aiebu"]
 	path = src/runtime_src/core/common/aiebu
-	url = https://github.com/HimanshuChoudhary-Xilinx/aiebu1.git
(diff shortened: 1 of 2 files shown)


.gitmodules
```diff
@@ -9,8 +9,8 @@
 	url = https://github.com/microsoft/GSL.git
 [submodule "src/runtime_src/core/common/aiebu"]
 	path = src/runtime_src/core/common/aiebu
-	url = https://github.com/Xilinx/aiebu.git
-        branch = main-ge
+	url = https://github.com/HimanshuChoudhary-Xilinx/aiebu1.git
+        branch = xrt_intergration
```
stsoe, line 13: This must be a mistake

stsoe (changes requested):

.gitmodules
HimanshuChoudhary-Xilinx: this was for local test before i merge to aiebu .... i have updated this

PR #8978, opened 2025-05-27
Fix preemption-overhead test status when not supported

rchane:
Preemption test fails when not supported (xclbin not available)
Discovered while testing preemption test for amdxdna
Added a check to put skip token instead of fail token in not supported scenario
N/A
Testing on Linux device.
N/A


aktondak (approved):

maxzhen (approved):
@rchane , so does it also fail on Windows?

rchane:
> @rchane , so does it also fail on Windows?

Yes, this issue should also be prevalent on Windows. I will reach out to Kelly regarding this.

PR #8980, opened 2025-05-28
Fix AIE Profile for VE2 DSP flow

IshitaGhosh:
Even for DSP flow on VE2, zocl devices attach to device 0, like older HW generations. As compilation for XDP libraries for DSP flow is still separate (from ML enabled flow) for VE2, AIE Profile library is now built to identify zocl device for zocl mode only. In ML/XDNA enabled compilation, AIE profile library identifies both xdna and zocl nodes and enables aie profiling if matching "xdp_mode" is set.

Unit test on VE2


chvamshi-xilinx (approved):
looks good

PR #8996, opened 2025-06-03
Fix default ctrl code id when sub kernel name is not passed

rbramand-xilinx:
This PR fix the issue to pick proper sub kernel when it is not passed explicitly by appplication.
It will pick the sub kernel available if its the only sub kernel available. If multiple sub kernels are available and user doesnt pass any sub kernel name then an exception is thrown

While parsing ELF we store all control codes in a map with key as sub kernel name (ctrl code id) and value as control code buffer, added a module_int function that returns the first entry of this map if its the only available sub kernel, otherwise an exception is thrown while creating xrt::ext::kernel object.

Low as this doesn't affect the original flow and it only comes into picture in xclbin2elf flow

Tested using SNL baremetal flow on strix and test works as expected.

NA

### DIFF
--- a/src/runtime_src/core/common/api/module_int.h
+++ b/src/runtime_src/core/common/api/module_int.h
@@ -47,6 +47,12 @@ struct kernel_info {
 xrt::module
 create_run_module(const xrt::module& parent, const xrt::hw_context& hwctx, const std::string& ctrl_code_id);
 
+// ELF can have multiple ctrl codes and when user doesn't provide
+// id while running the kernel then by default 1st ctrl code is run
+// This function returns the 1st available ctrl code id
+std::string
+get_default_ctrl_id(const xrt::module& module);
+
 // Fill in ERT command payload in ELF flow. The payload is after extra_cu_mask
 // and before CU arguments.
 uint32_t*
--- a/src/runtime_src/core/common/api/module_int.h
+++ b/src/runtime_src/core/common/api/module_int.h
@@ -47,9 +47,9 @@ struct kernel_info {
 xrt::module
 create_run_module(const xrt::module& parent, const xrt::hw_context& hwctx, const std::string& ctrl_code_id);
 
-// ELF can have multiple ctrl codes and when user doesn't provide
-// id while running the kernel then by default 1st ctrl code is run
-// This function returns the 1st available ctrl code id
+// If the user does not specify a sub-kernel ID,
+// select it automatically only if exactly one sub-kernel exists.
+// Otherwise, throw an exception
 std::string
 get_default_ctrl_id(const xrt::module& module);
 
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -1323,7 +1323,7 @@ private:
   size_t num_cumasks = 1;              // Required number of command cu masks
   control_type protocol = control_type::none; // Default opcode
   uint32_t uid;                        // Internal unique id for debug
-  std::string m_ctrl_code_id = "";     // ID to identify which ctrl code to load from elf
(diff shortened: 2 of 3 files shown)


stsoe (commented):
> added a module_int function that returns the first entry of this unordered_map which will be the first available sub kernel in the ELF.

An unordered_map makes this function non-deterministic.   You will get different first sub kernel depending on implementation of unordered map and surely differences between linux and windows.  Have you profiled the code when selecting std::unordered_map over std::map?

rbramand-xilinx:
> > added a module_int function that returns the first entry of this unordered_map which will be the first available sub kernel in the ELF.
> 
> An unordered_map makes this function non-deterministic. You will get different first sub kernel depending on implementation of unordered map and surely differences between linux and windows. Have you profiled the code when selecting std::unordered_map over std::map?

Hi Soren, thanks for pointing it out. I will change the data structure, you are right unordered_map doesn't work in this case.

src/runtime_src/core/common/api/module_int.h
```diff
@@ -47,6 +47,12 @@ struct kernel_info {
 xrt::module
 create_run_module(const xrt::module& parent, const xrt::hw_context& hwctx, const std::string& ctrl_code_id);
 
+// ELF can have multiple ctrl codes and when user doesn't provide
+// id while running the kernel then by default 1st ctrl code is run
+// This function returns the 1st available ctrl code id
+std::string
```
larry9523, line 53: The sub kernel id is a string. And it is ambiguous to say the 1st ctrl code. What we have agreed is if there is only one sub kernel, we use that one. Otherwise, we should throw because we don't know which sub kernel's control code we should pick if there is no sub kernel id is specified.

larry9523 (changes requested):

src/runtime_src/core/common/api/module_int.h
rbramand-xilinx: oh okay Larry, this would make things easy. I will change the implementation accordingly. Thanks

stsoe (approved):

larry9523 (approved):

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1512,14 +1512,14 @@ class kernel_impl : public std::enable_shared_from_this<kernel_impl>
...
-  static std::string
-  get_ctrlcode_id(const std::string& name)
+  std::string
+  get_ctrlcode_id(const std::string& id)
   {
     // kernel name will be of format - <kernel_name>:<ctrl code index>
-    if (auto i = name.find(":"); i != std::string::npos)
-      return name.substr(i + 1);
+    if (auto i = id.find(":"); i != std::string::npos)
+      return id.substr(i + 1);
```
sayyanna, line 1520: If the sub-kernel field is empty like "id": , what will be the behavior? This is just for error handling.

rbramand-xilinx: Hi Srilatha, as Larry told in his earlier comment if id is empty we will try to pick default available sub kernel if that is only available sub kernel (i.e: only one sub kernel is present). If more than one sub kernel is available and id is not passed or empty  XRT will throw exception.

larry9523: @rbramand-xilinx I think Sri Latha's question is if we want to support the syntax like "name:"

I thought we should only support
"name"
"name:id"

But I am fine if we want to treat "name:" as "name".

rbramand-xilinx: yes Larry the present code supports either "name:id" or "name".  Existing xclbin flow also works the same way, so may be continue this way.

larry9523: what happens to "name:"? Do we throw exception?

PR #8997, opened 2025-06-03
Modified strcture data element from fixed size to flexible array.

ManojTakasi:
This PR is cherry-picked from https://github.com/Xilinx/XRT/pull/8669 and https://github.com/Xilinx/XRT/pull/8746

https://jira.xilinx.com/browse/CR-1214907 We are getting UBSAN: array-index-out-of-bounds call trace while compiling xocl xclmgmt drivers in ubuntu 24.04

Reference: https://www.kernel.org/doc/html/latest/process/deprecated.html#zero-length-and-one-element-arrays

Added macros to pick flexible arrays when comipiling kernel code.

Modifications are in common files and will affect all platforms.

### DIFF
--- a/src/runtime_src/core/common/drv/cu_hls.c
+++ b/src/runtime_src/core/common/drv/cu_hls.c
@@ -110,7 +110,7 @@ static void cu_hls_xgq_start(struct xrt_cu_hls *cu_hls, u32 *data)
 	u32 i = 0;
 
 	num_reg = (cmd->hdr.count - (sizeof(struct xgq_cmd_start_cuidx)
-				     - sizeof(cmd->hdr) - sizeof(cmd->data)))/sizeof(u32);
+				     - sizeof(cmd->hdr)))/sizeof(u32);
 	for (i = 0; i < num_reg; ++i) {
 		cu_write32(cu_hls, ARGS + i * 4, cmd->data[i]);
 	}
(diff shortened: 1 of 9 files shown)


karthdmg-xilinx (approved):

stsoe (changes requested):
I am not comfortable with ert.h changes from data[1] -> data[].  While I completely agree that we should have used data[] from inception of this code, changing it now has ripple effects, e.g. sizeof(ert_start_kernel_cmd) changes, and who knows what XRT is relying doing and relying on? 

I am all for data[] for all new structs added, but leave existing ones alone.

stsoe:
> I am not comfortable with ert.h changes from data[1] -> data[]. While I completely agree that we should have used data[] from inception of this code, changing it now has ripple effects, e.g. sizeof(ert_start_kernel_cmd) changes, and who knows what XRT is relying doing and relying on?
> 
> I am all for data[] for all new structs added, but leave existing ones alone.

I take that back given that this is a cherry pick from code that was merged to master long ago.  Please ignore my comment above.  Also, user space is unchanged, so whatever it does with sizeof is unaffected (good).  It would be good if we had a project to consolidate all code on data[], which is exactly the right to use.

stsoe (approved):

chvamshi-xilinx:
@ManojTakasi , As this is a big change. Can you please make sure edge tests also works with this change. You can try couple of basic tests.

ManojTakasi:
Hi @chvamshi-xilinx, I tested the vadd and simple_graph applications after applying these changes, and both tests working as expected.

PR #9017, opened 2025-06-09
Context health exception message on ERT_STATE_COMMAND_TIMEOUT

sayyanna:
https://jira.xilinx.com/browse/AIESW-4265
Report context health through run.wait2() exception message.
Compose exception message when calling what()
Added changes to report context health through run.wait2() exception message, composed exception message when calling what()

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -3031,7 +3031,6 @@ public:
     auto ctx_health = get_ert_ctx_health_data(run_impl->get_ert_packet());
     return {ctx_health->aie_data, ctx_health->aie_data_size};
   }
-
 };
 
 // class runlist_impl - The internals of a runlist


src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4267,6 +4301,13 @@ what() const noexcept
   return handle->m_message.c_str();
 }
 
+const char*
+run::aie_error::
+what() const noexcept
+{
+  std::string msg = handle->m_message + handle->get_exception_message();
+  return msg.c_str();
```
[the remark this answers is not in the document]
stsoe: Yes, this is a bug.   Will not work.  You have to create/update the message in the constructor of the exception and store as part of m_msg in the error_impl class.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -196,6 +196,19 @@ cmd_state_to_string(ert_cmd_state state)
     : itr->second;
 }
 
+std::string
+fatal_err_type_to_string(uint32_t fatal_error_type)
```
stsoe: This function is not obvious given that it's argument type is a uint32_t.   What is the argument?  Who would want to call this function.    I think this is very specific to error_impl, so make it a private member function of that class.

stsoe (changes requested):

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4286,6 +4327,33 @@ command_error(const xrt::run& run, ert_cmd_state state, const std::string& msg)
   : detail::pimpl<runlist::command_error_impl>(std::make_shared<runlist::command_error_impl>(run, state, msg))
 {}
 
+runlist::aie_error::
+aie_error(const xrt::run& run, ert_cmd_state state, const std::string& what)
+  : command_error(run, state, what)
+{
+  construct_exception_message(run, state, what);
+}
```
stsoe: This initializes m_message, basically overwriting "what", so do it this way instead

```suggestion
### share with run::aie_error and runlist::aie_error
static std::string
amend_aie_error_message(const ert_packet* epkt, const std::string& what)
{
   ...;
   return msg;
}

runlist::aie_error::
aie_error(const xrt::run& run, ert_cmd_state state, const std::string& what)
  : command_error(run, state, amend_aie_error_message(run.get_ert_packet(), what)
{}
```

src/runtime_src/core/include/xrt/experimental/xrt_kernel.h
```diff
@@ -90,6 +92,9 @@ class runlist : public detail::pimpl<runlist_impl>
     XRT_API_EXPORT
     span<const uint32_t>
     data() const;
+  private:
+    void
+    construct_exception_message(const xrt::run& run, ert_cmd_state state, const std::string& msg);
```
stsoe: Not needed in visible API
```suggestion
```

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4267,6 +4281,33 @@ what() const noexcept
...
+run::aie_error::
+aie_error(const xrt::run& run, const std::string& what)
+  : command_error(run, what)
+{
+  construct_exception_message(run,what);
+}
+
+void
+run::aie_error::
+construct_exception_message(const xrt::run& run, const std::string& msg)
```
stsoe: Redo as suggested below.   (1) not needed as member function, (2) shared with runlist

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4267,6 +4281,33 @@ what() const noexcept
   return handle->m_message.c_str();
 }
 
+run::aie_error::
+aie_error(const xrt::run& run, const std::string& what)
+  : command_error(run, what)
+{
+  construct_exception_message(run,what);
+}
```
stsoe: See comment for runlist::aie_error

stsoe (changes requested):

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4267,6 +4268,35 @@ what() const noexcept
   return handle->m_message.c_str();
 }
 
+static std::string
+amend_aie_error_message(const ert_packet* epkt, const std::string& msg)
+{
+  std::ostringstream oss;
```
stsoe: I wonder if we are anticipating more that ERT_CMD_STATE_TIMEOUT?  If not, then this function should simply early-exit (returning `what`) if epkt->state != ERT_CMD_STATE_TIMOUT.  It would simplify the logic.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4267,6 +4268,31 @@ what() const noexcept
...
```
stsoe: It should be like this.  early-exit, means return ASAP, it reduces indentation and make the code simpler.

if (epkt->status != ERT_CMD_STATE_TIMEOUT)
  return msg;

// Now everything stays at this level, no additional scoping.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4267,6 +4268,31 @@ what() const noexcept
...
+  const std::map<uint32_t, const char*> fatal_error_string {
```
```suggestion
  static const std::map<uint32_t, const char*> fatal_error_string {
```

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4267,6 +4268,31 @@ what() const noexcept
...
+    {0, "N/A"}
+  };
+
+  oss << msg << "\n";
+  if (epkt->state == ERT_CMD_STATE_TIMEOUT) {
+    auto ctx_health = get_ert_ctx_health_data(const_cast<ert_packet*>(epkt));
+    auto itr = fatal_error_string.find(ctx_health->fatal_error_type);
+    auto fatal_error_type = (itr == fatal_error_string.end() ? "out of range": itr->second);
+    oss<<"txn_op_idx = 0x"<< std::uppercase << std::hex << std::setfill('0') << std::setw(8) << ctx_health->txn_op_idx
```
```suggestion
    oss << "txn_op_idx = 0x" << std::uppercase << std::hex << std::setfill('0') << std::setw(8) << ctx_health->txn_op_idx
```

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4267,6 +4268,31 @@ what() const noexcept
...
+        <<"\nctx_pc = 0x"<< std::uppercase << std::hex << std::setfill('0') << std::setw(8) << ctx_health->ctx_pc
```
```suggestion
        << "\nctx_pc = 0x"<< std::uppercase << std::hex << std::setfill('0') << std::setw(8) << ctx_health->ctx_pc
```

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4267,6 +4268,31 @@ what() const noexcept
...
+        <<"\nfatal_error_type "<<fatal_error_type<<"\n";
```
```suggestion
        << "\nfatal_error_type "<< fatal_error_type << "\n";
```

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4267,6 +4268,32 @@ what() const noexcept
...
+{
+  std::ostringstream oss;
+  static const std::map<uint32_t, const char*> fatal_error_string {
+    {0, "N/A"}
+  };
+
+  if (epkt->state != ERT_CMD_STATE_TIMEOUT)
+    return msg;
+
+  oss << msg << "\n";
```
stsoe: No need to create the stringstream unless we need it.

```suggestion
  static const std::map<uint32_t, const char*> fatal_error_string {
    {0, "N/A"}
  };

  if (epkt->state != ERT_CMD_STATE_TIMEOUT)
    return msg;
    
  std::ostringstream oss;
  oss << msg << "\n";
```

stsoe (changes requested):

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4267,6 +4268,32 @@ what() const noexcept
...
+  auto ctx_health = get_ert_ctx_health_data(const_cast<ert_packet*>(epkt));
```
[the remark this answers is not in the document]
stsoe: Is this const_cast needed?

src/runtime_src/core/include/xrt/detail/ert.h
```diff
@@ -1199,7 +1199,7 @@ get_ert_regmap_size_bytes(struct ert_start_kernel_cmd* pkt)
 }
 /* ert_ctx_health_data structure is valid only if the ert opcode is START_NPU_* and cmd state is ERT_CMD_STATE_TIMEOUT*/
 static inline struct ert_ctx_health_data*
-get_ert_ctx_health_data(struct ert_packet* pkt)
+get_ert_ctx_health_data(const struct ert_packet* pkt)
```
stsoe, line 1202: This is okay,  but not entirely correct.  The function is using explicit casting, which is has to because it is C, but it is casting away the const of the health data.    If implemented correctly, this function should have cast to and returned `const struct ert_ctx_health_data*`  but let's leave it as is.

PR #9023, opened 2025-06-11
[XDP] Removing unused parameters to avoid Windows warning treated as error

jvillarre:
A few functions in the profiling library had unused parameters, which end up causing warnings on Windows builds and these warnings are treated as errors in some builds.  This pull request removes the unused parameters.

Removes unused parameters from internal utility functions.

Low risk as these functions were not called from outside of the object.

No documentation impact.


IshitaGhosh (approved):
Changes look good.

PR #9030, opened 2025-06-17
Enable XRT ELF flow for new ELF format (group elf)

rbramand-xilinx:
Added XRT core framework changes for parsing new ELF format.
Spec page - https://amd.atlassian.net/wiki/spaces/AIE/pages/779178481/IPU+Execution+Buffer+ELF+Specification#IPUExecutionBufferELFSpecification-ELFDefinition

Added ELF parsing, patching logic for new ELF format for both aie2ps and aie2p flows.
Made sure existing flows (xclbin + elf) doesn't break.

Moderate.
Have verified unit tests for all the flows but needs extensive testing as the code that is changed is in critical path.

1. aie2ps new ELF flow on HW
2. aie2ps existing test cases on HW
3. strix linux non control packet xclbin + elf flow on HW
4. strix windows concurrency tests on HW

TODO :
1. Strix linux control packet test case
2. Strix windows existing flows
3. Strix full Elf flow on linux and windows

Need to document the new Host code side changes to use the full ELF flow

### DIFF
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -46,18 +46,18 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
   void
   create_module_map(const xrt::elf& elf)
   {
-    xrt::module module{elf};
+    xrt::module module_obj{elf};
 
     // Store the module in the map against all available kernels in the ELF
     // This will be useful for module lookup when creating xrt::kernel object
-    // using kernel name    
-    const auto& kernels_info = xrt_core::module_int::get_kernels_info(module);
+    // using kernel name
+    const auto& kernels_info = xrt_core::module_int::get_kernels_info(module_obj);
     for (const auto& k_info : kernels_info) {
       auto kernel_name = k_info.props.name;
       if (m_module_map.find(kernel_name) != m_module_map.end())
         throw std::runtime_error("kernel already exists, cannot use this ELF with this hw ctx\n");
 
-      m_module_map.emplace(std::move(kernel_name), module);
+      m_module_map.emplace(std::move(kernel_name), module_obj);
     }
   }
 
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -63,7 +63,7 @@ namespace
 static constexpr size_t elf_page_size = AIE_COLUMN_PAGE_SIZE;
 static constexpr uint8_t Elf_Amd_Aie2p        = 69;
 static constexpr uint8_t Elf_Amd_Aie2ps       = 64;
-static constexpr uint8_t Elf_Amd_Aie2p_config = 70;
+static constexpr uint8_t Elf_Amd_Aie2ps_group = 70;
 
 // In aie2p max bd data words is 8 and in aie4/aie2ps its 9
 // using max bd words as 9 to cover all cases
@@ -446,6 +446,34 @@ demangle(const std::string& mangled_name)
 #endif
 }
 
+// checks if ELF has .group sections
+static bool
+is_group_elf(const ELFIO::elfio& elfio)
+{
+  constexpr uint8_t major_ver_mask = 0xF0;
+  constexpr uint8_t minor_ver_mask = 0x0F;
+  constexpr uint8_t shift = 4;
+
+  auto abi_version = elfio.get_abi_version();
+  auto os_abi = elfio.get_os_abi();
+
+  if (os_abi == Elf_Amd_Aie2p) {
+    // ELF of aie2p with version >= 1.0 uses group sections
+    constexpr uint8_t group_elf_major_version = 1;
+    if (((abi_version & major_ver_mask) >> shift) >= group_elf_major_version)
+      return true;
+
+    return false;
+  }
+  else {
+    // ELF of aie2ps/aie4 with version >= 0.3 uses group sections
+    constexpr uint8_t group_elf_minor_version = 3;
+    if ((abi_version & minor_ver_mask) >= group_elf_minor_version)
+      return true;
+      
+    return false;
+  }
+}
 } // namespace
 
 namespace xrt
@@ -507,6 +535,12 @@ public:
 
   virtual const buf&
   get_pdi(const std::string& /*pdi_name*/) const
+  {
+    throw std::runtime_error("Not supported");
+  }
+
+  virtual xrt::bo&
+  get_pdi_bo(const std::string& /*pdi_name*/)
   {
     throw std::runtime_error("Not supported");
   }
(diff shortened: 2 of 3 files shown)


src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -790,19 +907,141 @@ class module_elf : public module_impl
...
+  // classes for simplicity and avoids unnecessary boilerplate setters,
+  // getters code
+  // NOLINTBEGIN
+  xrt::elf m_elf;
+  const ELFIO::elfio& m_elfio; // we should not modify underlying elf
+  uint8_t m_os_abi = Elf_Amd_Aie2p;
+  // store arg and its corresponding patcher information for each
+  // kernel + subkernel combination
+  // key - ctrlcode id(grp idx), value - map of key string and patcher info
+  std::unordered_map<uint32_t, std::map<std::string, patcher>> m_arg2patcher;
```
stsoe: I am vary of std::unordered_map.   How many entries are you expecting 1000s ?   Do you have data that shows the need for unordered_map?   Unordered map is expensive memory wise.  A comment as to why unordered_map was chosen would help.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1303,27 +1303,27 @@ class kernel_impl : public std::enable_shared_from_this<kernel_impl>
...
+  size_t fa_num_inputs = 0;                   // Fast adapter number of inputs per meta data
+  size_t fa_num_outputs = 0;                  // Fast adapter number of outputs per meta data
+  size_t fa_input_entry_bytes = 0;            // Fast adapter input desc bytes
+  size_t fa_output_entry_bytes = 0;           // Fast adapter output desc bytes
+  size_t num_cumasks = 1;                     // Required number of command cu masks
   control_type protocol = control_type::none; // Default opcode
-  uint32_t uid;                        // Internal unique id for debug
-  std::string m_ctrl_code_id;          // ID to identify which ctrl code to load from elf
+  uint32_t uid;                               // Internal unique id for debug
+  uint32_t m_ctrl_code_id = UINT32_MAX;       // ID to identify which ctrl code to load from elf
```
stsoe: This is nitpick, but I prefer symbolic constants to default values.   I am not sure what UINT32_MAX represents, but having something like <type>::no_id (or no_ctrl_cod_id) is more telling.  Regarding where the symbolic constant should be, I don't know, but as public in some class is usually what I do.

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -790,19 +907,141 @@ class module_elf : public module_impl
...
+
+        sec_ids.push_back(sec->get_index());
+        // fill the sec_to_grp_map for lookup later
+        // for this kind of ELF we fill UINT32_MAX as there is no group section
+        m_sec_to_grp_map[sec->get_index()] = UINT32_MAX;
+      }
+
+      // fill kenerl name to id map with id as grp idx
+      m_kname_to_id_map[kname] = UINT32_MAX;
+      id_sections_map.emplace(UINT32_MAX, sec_ids);
```
stsoe:
```suggestion
      id_sections_map.emplace(UINT32_MAX, std::move(sec_ids));
```
Here is the UINT32_MAX that probably should be symbolic to communicate what it represents.

stsoe (commented):
Lots of changes.  I didn't do a very good review.  The changes in xrt_module.cpp look rather complicated, but maybe it's what it is.

src/runtime_src/core/common/api/xrt_kernel.cpp
rbramand-xilinx: yes Soren, I will make the change accordingly

src/runtime_src/core/common/api/xrt_module.cpp
rbramand-xilinx: Hi Soren, I initially chose std::unordered_map because it offers faster average case access and insertion times and in this case the order of elements is not important.
That said, I do understand its higher memory usage due to the underlying hash table structure. For smaller datasets, the constant overhead of hashing can outweigh the performance benefits, making std::map a more balanced choice.
Since the expected number of entries is relatively small and performance is not a bottleneck here, I’ll switch to std::map for better memory efficiency and maintainability. Thanks for the feedback!

stsoe (approved):

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -43,6 +43,24 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
   std::shared_ptr<xrt_core::usage_metrics::base_logger> m_usage_logger =
       xrt_core::usage_metrics::get_usage_metrics_logger();
 
+  void
+  create_module_map(const xrt::elf& elf)
+  {
+    xrt::module module{elf};
```
sonals: Please use another variable name instead of _module_ which is a keyword in C++20.

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -735,38 +776,161 @@ class module_userptr : public module_impl
 // construct patcher objects for each argument.
 class module_elf : public module_impl
 {
-protected:
-  xrt::elf m_elf;
-  const ELFIO::elfio& m_elfio; // we should not modify underlying elf
-  uint8_t m_os_abi = Elf_Amd_Aie2p;
-  std::map<std::string, patcher> m_arg2patcher;
+  std::vector<std::string>
+  split(const std::string& s, char delimiter)
```
sonals, line 780: `const`

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -482,24 +510,24 @@ class module_impl
   // single partition.  The returned vector has the control
   // code as extracted from ELF or userptr.
   virtual const std::vector<ctrlcode>&
-  get_data() const
+  get_data(uint32_t /*id*/) const
```
sonals, line 513: More for future consideration, should all the base class virtual functions with `throw` implementation be a _pure_ virtual function?

sonals (approved):

src/runtime_src/core/common/api/xrt_module.cpp
rbramand-xilinx: Hi Sonal, class module_sram inherits this class and and it doesn't implement most of the functions in module_impl class.
So if we make them pure virtual then we need to override these functions in module_sram and use throw there which I feel is redudant.
That said xrt_module.cpp needs refactoring and I will do it with next set of PRs. Thanks

src/runtime_src/core/common/api/xrt_module.cpp
rbramand-xilinx: I will make the function static as it doesn't use any member variables

chvamshi-xilinx:
@rbramand-xilinx confirmed that all basic tests are working on both client/ve2.

PR #9031, opened 2025-06-17
[XDP] Fix ExternalBuffers identification and storage

vipangul:
Observed in two stamped designs on Client
Not all shim tiles were getting configured for metric sets 
Same Port Names are mapped to different shim tiles, this was causing Key collision in internal data structure storage.

Each port name is uniquely identified and stored in associated map.

Low

Verified on Client, original Test now passes.


jvillarre (approved):

IshitaGhosh (approved):
Changes look good now.

PR #9042, opened 2025-06-24
remove unnecessary header includes

xfreid:
There are quite a few #includes that turn out be either not needed or redundant

N/A

Remove unnecessary includes;

Low

Build XRT successfully

N/A

### DIFF
--- a/src/runtime_src/tools/xclbinutil/CBOR.cxx
+++ b/src/runtime_src/tools/xclbinutil/CBOR.cxx
@@ -16,11 +16,6 @@
 
 #include "CBOR.h"
 
-#include <stdexcept>
-#include <vector>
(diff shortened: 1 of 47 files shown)


src/runtime_src/tools/xclbinutil/CBOR.cxx
```diff
@@ -16,11 +16,6 @@
 
 #include "CBOR.h"
 
-#include <stdexcept>
```
stsoe, line 19: Huh.  Line 57 throws so `<stdexcept>` header is needed. Don't  rely on indirect includes, always include-what-you-use.

stsoe (changes requested):

PR #9045, opened 2025-06-26
Providing user the option of setting cert opcode timeout value

SravanKumarAllu-xilinx:

Providing user the option of setting cert opcode timeout value

N/A

Added cert_timeout option for user to set in xrt.ini with required value

N/A

simple vadd test to check whether we are able to get the user set value in "shim" layer

N/A


src/runtime_src/core/common/config_reader.h
```diff
@@ -444,6 +444,13 @@ get_xclbin_programing()
   return value;
 }
 
+inline unsigned int
+get_cert_timeout()
+{
+  static unsigned int value = detail::get_uint_value("Runtime.cert_timeout",50); //default value for cert timeout is 50msec
```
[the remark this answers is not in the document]
SravanKumarAllu-xilinx: Added const unsigned int variable and used the same instead of magic number

src/runtime_src/core/common/config_reader.h
```diff
@@ -12,6 +12,8 @@
 
 #include <boost/property_tree/ptree_fwd.hpp>
 
+constexpr unsigned int default_cert_timeout = 50;
```
stsoe: Doesn't belong here.  The clang-tidy warning was bogus, the code is self explanatory if you change the key to `Runtime.cert_timeout_ms`
```suggestion
```

src/runtime_src/core/common/config_reader.h
```diff
@@ -444,6 +446,13 @@ get_xclbin_programing()
   return value;
 }
 
+inline unsigned int
+get_cert_timeout()
+{
+  static unsigned int value = detail::get_uint_value("Runtime.cert_timeout", default_cert_timeout); //default value for cert timeout is 50msec
```
stsoe:
```suggestion
  static unsigned int value = detail::get_uint_value("Runtime.cert_timeout_ms", 50); // NOLINT
```

src/runtime_src/core/common/config_reader.h
SravanKumarAllu-xilinx: Done

src/runtime_src/core/common/config_reader.h
SravanKumarAllu-xilinx: Done

PR #9047, opened 2025-06-27
Enable arm64 build on Windows

jhoyamd:
- remove /CETCOMPAT for ARM build
- update cmake to ignore building aiebu and python directories when building on ARM

Update cmake to support Windows on ARM64 build

New request to support Windows on ARM64 build

Update CMAKE complier/linker flag
disable python and aiebu when using ARM64 compiler

No impact to existing build configuration

Test build for on x64(release/debug) and arm64 (release/debug)

N/A


jhoyamd:
> * update cmake to ignore building aiebu and python directories when building on ARM

@stsoe, updated to " update cmake to ignore building aiebu and python directories when building on ARM"

stsoe:
The PR is still not correct, why is aiebu submodule updated?

jhoyamd:
> The PR is still not correct, why is aiebu submodule updated?

fixed

PR #9048, opened 2025-06-27
[2025.1] Fix for hardware emulation seg fault when PL profiling is enabled on Alveo

jvillarre:
Backporting pull request 8927 to 2025.1 branch, which fixes a seg fault in Alveo hardware emulation when PL profiling/trace is enabled.

The original issue is that profiling keeps a shared pointer to an xrt::device in order to access the PL portion, but in Alveo hardware emulation, at the end of the application when our objects are being destroyed, the shared_pointer is destroyed and the xrt::device attempts to clean itself up, but this leads to a problem as the simulation has already closed and the xrt::device cannot cleanly be removed.

This issue was discovered through regression testing and fixed in master.

This pull request makes it so profiling releases its shared pointer reference before the simulation is stopped, by catching it in the hook in hardware emulation shim destructor.

Low risk as this is specifically targeted to Alveo hardware emulation.

The original failing test has been verified with the fix on the 2025.1 branch.

No documentation impact.

### DIFF
--- a/src/runtime_src/xdp/profile/database/static_info_database.cpp
+++ b/src/runtime_src/xdp/profile/database/static_info_database.cpp
@@ -1,6 +1,6 @@
 /**
  * Copyright (C) 2016-2022 Xilinx, Inc
- * Copyright (C) 2022-2024 Advanced Micro Devices, Inc. - All rights reserved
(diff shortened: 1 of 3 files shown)


IshitaGhosh (approved):
Changes look good.

PR #9056, opened 2025-07-01
Fixing pyxrt based on Jeff's bug

shaurjai:
Not able to call pyxrt.ext.bo.access_mode without erroring
pyextbo bug in using access_mode and read/write enum values
Edited constructor for enum so it used ext instead of extbo, preventing shadowing of bo values by ext.bo values.
Low
Tested calls to access_mode and pyxrt.ext.bo.read
Low


fifield (approved):
Works for me, thanks.

PR #9059, opened 2025-07-04
Fix HIP kernel argument extraction

sonals:
Update the HIP kernel parameter walking/parsing code 

NA

Clarified the semantics of parameters passed in `kernelParams` as part of `hipLaunchKernel`. Every element in the array
is a pointer to the _actual_ parameter. The actual parameter may be a scalar or a buffer. In the case of former we end 
up pointer to a scalar. In the case of latter we end up with pointer to pointer since the buffer is represented as pointer.

Need to review all the existing HIP tests to ensure they are composing the `kernelParams` correctly.

Tested with model-tests testcase which was converted to HIP

NA


chvamshi-xilinx (approved):
Looks good.

PR #9060, opened 2025-07-07
AIESW-6055 [XRT-RUNNER] Enhance runner recipe with debug buffers

stsoe:
Add support for debug buffers in recipe::resources::buffers by introducing a new buffer type `debug`.

Enhance the runner API to extract raw buffer data as a span of bytes.

The `debug` type buffer is created specific to a hw context (as opposed to device) and cannot be used as kernel argument.  Its use is specific to XRT tools.

The raw data returned through runner API can be used with any resource buffer. The interpretation and validity of the data is at the discretion of the application that instantiates and uses the runner.

The recipe json is enhanced to accept debug buffers as shown here:

```
  "resources": {
    "buffers": [
      {
        "name": "tops",   // you pick a desired name
        "size": 53591040,
        "type": "debug"
      },
```

The runner API is amended with two methods to map resource data and return the data as a span.  The xrt::detail::span is because C++17 does not have support for std::span, and because at least GCC in C++20 does not have stable ABI for std::span.

```
  // map_buffer() - Get raw buffer data as a span of bytes
  // The buffer is synced from device as part of this call.
  // The returned data cannot be written to.
  XRT_API_EXPORT
  xrt::detail::span<const std::byte>
  map_buffer(const std::string& name) const;

  // map_buffer() - Get raw buffer data as a span of type
  // The buffer is synced from device as part of this call.
  // The returned data cannot be written to.
  template <typename MapType>
  xrt::detail::span<MapType>
  map_buffer(const std::string& name) const
  {
    static_assert(std::is_pointer<MapType>::value &&
                  std::is_const<std::remove_pointer_t<MapType>>::value,
                  "MapType must be a pointer and const-qualified type");

    auto span = map_buffer(name);
    return {reinterpret_cast<MapType*>(span.data()), span.size() / sizeof(MapType)};
  }
```

### DIFF
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -1530,6 +1530,15 @@ map()
   });
 }
 
+const void*
+bo::
+map() const
+{
+  return xdp::native::profiling_wrapper("xrt::bo::map", [this]{
+    return handle->get_hbuf();
+  });
+}
+
 void
 bo::
 write(const void* src, size_t size, size_t seek)
--- a/src/runtime_src/core/common/runner/runner.cpp
+++ b/src/runtime_src/core/common/runner/runner.cpp
@@ -16,12 +16,14 @@
 #include "core/common/error.h"
 #include "core/common/module_loader.h"
 #include "core/common/time.h"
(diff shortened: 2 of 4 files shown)


aktondak (approved):
Thanks for the quick implementation. Looks good !

PR #9062, opened 2025-07-09
Returning proper device index in edgecase instead of hardcoding

ManojTakasi:
In Telluride, we have two devices running simultaneously on the edge side. Currently, both devices are showing the same BDF: 0000:00:00.0.
To resolve this, we modified the BDF function to use the device ID, which now returns distinct BDFs for each device.
we updated the str2index() function to return device index properly. This change ensures compatibility with the legacy flow where devices are accessed using an index, such as in the command:
xrt-smi examine -r aie -d 0.

Added get_bdf_info() function to return the proper bdf in the edge side aswell.
Added xrt_add_subdirectory instead of add_subdirectory

low.

Tested by building and running vadd testcase on vek385

### DIFF
--- a/src/runtime_src/core/tools/common/XBUtilities.cpp
+++ b/src/runtime_src/core/tools/common/XBUtilities.cpp
@@ -181,15 +181,6 @@ XBUtilities::get_available_devices(bool inUserDomain)
   return pt;
 }
 
-/*
- * currently edge supports only one device
- */
-static uint16_t
-deviceId2index()
-{
-  return 0;
-}
-
 std::string
 XBUtilities::str_available_devs(bool _inUserDomain)
 {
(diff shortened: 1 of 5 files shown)


src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -319,8 +310,8 @@ str2index(const std::string& str, bool _inUserDomain)
 
     auto bdf = xrt_core::device_query<xrt_core::query::pcie_bdf>(device);
     // if the bdf is zero, we are dealing with an edge device
```
AShivangi: please remove this comment as well

ManojTakasi: Removed, thank you.

PR #9063, opened 2025-07-09
common: xrt_core::device: add context manager member

wendyliang25:
There is context manager required per xrt_core::device, we add it as a member to the xrt_core::device class.

The context manager is only going to be used by context management which is defined in common/api/context_mgr.cpp and it is not required to expose at all. We define a basic class in the xrt_core::device for the context manager and make the member in xrt_core::device as as shared pointer to this base class. The context_mgr.cpp defines the actual derived class, and it is the one to create create and get the context manager for the context management.

Without this change, if we hold e.g. xrt::kernel in global space, after application completes, when system is doing cleanup, it is possible to delete the static variable defined in the function before the function is called during the cleanup.


stsoe (commented):
Hey Wendy.  Let's simplify this a bit more.
The device class should have the shared_ptr data member as you have done, but only one member function:

```
std::shared_ptr<context_mgr>
get_context_mgr();
```

This member should be defined in context_mgr.cpp as
```
std::shared_ptr<context_mgr>
device::
get_context_mgr()
{
  if (m_ctx_mgr)
    return m_ctx_mgr;

 return (m_ctx_mgr = std::make_shared<device_context_mgr>());
}
```

The context_mgr.cpp::create(...) function will be
```
std::shared_ptr<device_context_mgr>
create(xrt_core::device* device)
{
   return std::static_pointer_cast<device_context_mgr>(device->get_context_mgr());
}
```

I think this would work and be simpler.

And done use `foo(void)` simply do `foo()` :-)

wendyliang25:
> Hey Wendy. Let's simplify this a bit more. The device class should have the shared_ptr data member as you have done, but only one member function:
> 
> ```
> std::shared_ptr<context_mgr>
> get_context_mgr();
> ```
> 
> This member should be defined in context_mgr.cpp as
> 
> ```
> std::shared_ptr<context_mgr>
> device::
> get_context_mgr()
> {
>   if (m_ctx_mgr)
>     return m_ctx_mgr;
> 
>  return (m_ctx_mgr = std::make_shared<device_context_mgr>());
> }
> ```
> 
> The context_mgr.cpp::create(...) function will be
> 
> ```
> std::shared_ptr<device_context_mgr>
> create(xrt_core::device* device)
> {
>    return std::static_pointer_cast<device_context_mgr>(device->get_context_mgr());
> }
> ```
> 
> I think this would work and be simpler.
> 
> And don't use `foo(void)` simply do `foo()` :-)

Thanks @stsoe , i have updated the patch based on the above comments, could you check again? thanks

src/runtime_src/core/common/device.h
```diff
@@ -45,6 +46,12 @@ class device : public ishim
 {
 public:
   template <typename T> using span = xrt_core::span<T>;
+
+  class context_mgr {
```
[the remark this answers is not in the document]
wendyliang25: based on discussion with @stsoe , we are going to ignore this comment in this case, as it is not necessary.

src/runtime_src/core/common/api/context_mgr.cpp
```diff
@@ -119,54 +119,53 @@ class device_context_mgr
...
-  static std::mutex ctx_mutex;
-  std::lock_guard<std::mutex> lk(ctx_mutex);
-  auto cmgr = d2cmgr[device].lock();
-  if (!cmgr && create)
-    // NOLINTNEXTLINE using new for weak_ptr
-    d2cmgr[device] = cmgr = std::shared_ptr<device_context_mgr>(new device_context_mgr); 
-  return cmgr;
+  if (!device) {
+    throw std::runtime_error("Empty device, failed to get device context manager.");  }
+  return std::static_pointer_cast<device_context_mgr>(device->get_context_mgr());
```
stsoe:
```suggestion
  if (!device)
    throw std::runtime_error("Empty device, failed to get device context manager.");
    
  return std::static_pointer_cast<device_context_mgr>(device->get_context_mgr());
```

stsoe (approved):
Looks good.   Thanks!
One minor comment related for formatting.

src/runtime_src/core/common/api/context_mgr.cpp
wendyliang25: thanks @stsoe , i have updated the patch, could you check again? thanks

PR #9069, opened 2025-07-16
[XDP] Added method to get PL device Interface for register Xclbin flow

vipangul:
Required for hw contex register xclbin flow.

NA

Device ID 0 is reserved for PL Device Interface.

Low
Verified on edge.


vipangul:
Closing this, will create a separate combined PR which will have creation of config, deviceID handling.

PR #9070, opened 2025-07-17
Added read() call to consume the interrput when interrput is occured

ManojTakasi:
https://jira.xilinx.com/browse/CR-1221077 wait_ip_interrupt(timeout) is not working as expected, but wait_ip_interrupt(timeout) along with wait_ip_interrupt() is working properly.

poll() is not consuming the interrupt; it only notifies when data is available. Because of this, future poll() calls behave incorrectly.
 
A read() call was added to the function when the interrupt is triggered. This consumes and clears the interrupt properly.

low
Tested by building and running Zebra application on VEK280.

NA


stsoe (approved):
LGTM

PR #9081, opened 2025-07-22
[RFC] hip: use hipMemPrefetchAsync() to sync buffer in HIP flow

wendyliang25:
This Patch sets solve two HIP XRT APIs issues:
* Some NPUs only support non-coherent memory, and thus, we need to sync data to memory before AIE kernel execution, and we will need to sync data after kernel execution to make sure CPU gets the right data. This patch set use `hipMemPrefetchAsync()` to sync data before kernel launch and after kernel execution.

Here is the example to use the API:
```
uint32_t* ibuf0, ibuf1, ibuf2, obuf;
uint32_t* ibuf0_dev, ibuf1_dev, ibuf2_dev, obuf_dev;
hipHostMalloc(&ibuf0, size, FLAG);
hipHostMalloc(&ibuf1, size, FLAG);
hipHostMalloc(&ibuf2, size, FLAG);
hipHostMalloc(&obuf, size, FLAG);
hipHostGetDevicePointer(&ibuf0_dev, ibuf0, 0);
hipHostGetDevicePointer(&ibuf1_dev, ibuf1, 0);
hipHostGetDevicePointer(&ibuf2_dev, ibuf2, 0);
hipHostGetDevicePointer(&obuf_dev, obuf, 0);
uint64_t opcode = 3;
std::array<void *, 8> args = {
      &opcode,
      nullptr,
      nullptr,
      &ibuf0_dev,
      &ibuf1_dev,
      &ibuf2_dev, nullptr, &obuf_dev};
hipMemPrefetchAsync(ibuf0_dev, size, 0, stream);
hipMemPrefetchAsync(ibuf1_dev, size, 0, stream);
hipMemPrefetchAsync(ibuf2_dev, size, 0, stream);
hipMemPrefetchAsync(obuf_dev, size, 0, stream); // sometimes, obuf, may also want to sync before kernel starts
hipModuleLaunchKernel(function, gridDimX, gridDimY, gridDimZ,
                      blockDimX, blockDimY, blockDimZ,
                      sharedMemBytes, stream,
                      args.data(), nullptr);
hipStreamSynchronize(stream);
hipMemPrefetchAsync(obuf_dev, size, 0, stream); // sync() will be completed in cmd submit()
hipStreamSynchronize(stream); // can be optional as it will clear the completed sync() command from the queue
```

Tested with `expoi-v1-fp32-256x256-1x_GroupConv__coeffs_up2_conv_conv2_conv_kernel_2_Conv` model on strix

### DIFF
--- a/src/runtime_src/hip/api/hip_memory.cpp
+++ b/src/runtime_src/hip/api/hip_memory.cpp
@@ -435,6 +435,19 @@ namespace xrt::core::hip
       std::make_shared<memory_pool_command>(hip_stream, memory_pool_command::memory_pool_command_type::alloc, pool, *dev_ptr, size));
     s_hdl->enqueue(command_cache.get(cmd_hdl));
   }
+
+  static void
+  hip_mem_prefetch_async(const void* dev_ptr, size_t count, int device, hipStream_t stream)
+  {
+    // device is not required for xrt::bo lookup, only buffer device address is required.
+    (void)device;
+    auto hip_stream = get_stream(stream);
+    throw_invalid_value_if(!hip_stream, "Invalid stream handle.");
+    auto s_hdl = hip_stream.get();
+    auto cmd_hdl = insert_in_map(command_cache,
+                                 std::make_shared<mem_prefetch_command>(hip_stream, dev_ptr, count));
+    s_hdl->enqueue(command_cache.get(cmd_hdl));
+  }
 } // xrt::core::hip
 
 template<typename F> hipError_t
@@ -621,8 +634,7 @@ hipMemPoolTrimTo(hipMemPool_t mem_pool, size_t min_bytes_to_hold)
 hipError_t
 hipMemPrefetchAsync(const void* dev_ptr, size_t count, int device, hipStream_t stream)
 {
-  // TODO: implement this once the linux driver implements HMM
-  return hipSuccess;
+  return handle_hip_memory_error([&] { xrt::core::hip::hip_mem_prefetch_async(dev_ptr, count, device, stream); });
 }
 
 // Allocates memory from a specified pool with stream ordered semantics.
--- a/src/runtime_src/hip/core/event.cpp
+++ b/src/runtime_src/hip/core/event.cpp
@@ -242,6 +242,28 @@ bool memory_pool_command::wait()
   return true;
 }
 
+bool mem_prefetch_command::submit()
+{
+  auto hip_mem_and_off = memory_database::instance().get_hip_mem_from_addr(m_dev_ptr);
+  auto hip_mem = hip_mem_and_off.first;
+  size_t hip_mem_off = hip_mem_and_off.second;
+  throw_invalid_value_if(!hip_mem, "Invalid prefetch buf address.");
+  throw_invalid_value_if((hip_mem->get_size() < (hip_mem_off + m_size)),
+                         "Invalid prefetch buf address or size.");
+
+  // The under xrt::bo::sync() behaves the same for both TO_DEVICE or FROM_DEVICE direction.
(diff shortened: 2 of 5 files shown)


src/runtime_src/hip/core/event.h
```diff
@@ -206,6 +207,21 @@ class memory_pool_command : public command
...
+public:
+  // sync() always happens in submit()
+  mem_prefetch_command(std::shared_ptr<stream> s, const void* dev_ptr, size_t size)
+    : command(command::type::mem_prefetch, std::move(s)), m_dev_ptr(dev_ptr), m_size(size)
+  {}
+  bool submit() override;
+  bool wait() override;
+
+protected:
+  const void* m_dev_ptr;
```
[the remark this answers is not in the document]
wendyliang25: changed `protected` to `private`

src/runtime_src/hip/core/memory.cpp
```diff
@@ -129,6 +129,13 @@ namespace xrt::core::hip
     m_bo.sync(direction);
   }
 
+  void
+  memory::sync(xclBOSyncDirection direction, size_t sz, size_t offset)
+  {
+    assert(m_bo);
```
stsoe: Let's not use assert, better throw and catch / convert to error code C++ / C boundary.

stsoe (changes requested):

src/runtime_src/hip/core/memory.cpp
wendyliang25: Thanks @stsoe , i have updated the PR, will send another PR to replace existing `assert` with throwing error. could you check this patch again? thanks

sonals (approved):

src/runtime_src/hip/api/hip_memory.cpp
```diff
@@ -435,6 +435,19 @@ namespace xrt::core::hip
       std::make_shared<memory_pool_command>(hip_stream, memory_pool_command::memory_pool_command_type::alloc, pool, *dev_ptr, size));
     s_hdl->enqueue(command_cache.get(cmd_hdl));
   }
+
+  static void
+  hip_mem_prefetch_async(const void* dev_ptr, size_t count, int device, hipStream_t stream)
+  {
+    // device is not required for xrt::bo lookup, only buffer device address is required.
+    (void)device;
```
stsoe, line 443: The most common way to make a parameter unused in c++ is to not name it.
```suggestion
  static void
  hip_mem_prefetch_async(const void* dev_ptr, size_t count, int, hipStream_t stream)
  {
```

src/runtime_src/hip/core/event.cpp
```diff
@@ -242,6 +242,28 @@ bool memory_pool_command::wait()
...
+  auto hip_mem_and_off = memory_database::instance().get_hip_mem_from_addr(m_dev_ptr);
+  auto hip_mem = hip_mem_and_off.first;
+  size_t hip_mem_off = hip_mem_and_off.second;
+  throw_invalid_value_if(!hip_mem, "Invalid prefetch buf address.");
+  throw_invalid_value_if((hip_mem->get_size() < (hip_mem_off + m_size)),
+                         "Invalid prefetch buf address or size.");
+
+  // The under xrt::bo::sync() behaves the same for both TO_DEVICE or FROM_DEVICE direction.
+  // we pick xclBOSyncDirection::XCL_BO_SYNC_BO_TO_DEVICE as input argument here.
+  hip_mem->sync(xclBOSyncDirection::XCL_BO_SYNC_BO_TO_DEVICE, m_size, hip_mem_off);
```
chvamshi-xilinx, line 256: We need to call async (call std::async)  function & wait for this thread to complete in wait() API.
Otherwise , this becomes a blocking call.

PR #9083, opened 2025-07-22
hip: replay assert with throw error

wendyliang25:
assert() is to identify and diagnose program bugs during development. In release build when NDEBUG macro is defined, asser macros are completely removed from the compiled code.
In HIP XRT implementation, we want to check parameters and throw errors when check fails. handle_hip_memory_error() will catch the error and return error code.


sonals (approved):
Thanks for cleaning this up.

PR #9084, opened 2025-07-22
hip: core: stream: clear m_top_event when stream completes

wendyliang25:
If a stream is set to wait for an event, all the commands enqueued to the stream, after setting the stream waiting on the event will be added to the commands chain to the event so that when an event is completed, all those dependent commands will execute. When the stream which depends on an event completed, we should reset the event on which the stream depends on, otherwise, the commands enqueued to the stream later will added to the event commands chain. Since the event has completed, it will not signal again, and thus, those commands will not run again, even we want to rerun the stream.

### DIFF
--- a/src/runtime_src/hip/core/stream.cpp
+++ b/src/runtime_src/hip/core/stream.cpp
@@ -121,6 +121,8 @@ await_completion()
       command_cache.remove(cmd.get());
     m_cmd_queue.pop_front();
   }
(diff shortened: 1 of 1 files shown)


sonals (approved):
Do we have a testcase to model this?

chvamshi-xilinx:
As Sonal suggested, Lets have unit tests for all of these cases.

PR #9100, opened 2025-07-29
hip: function: introduce runs cache in hip functions

wendyliang25:
In HIP stream execution flow, every kernel launch, we will need to get a run for the command to send to the NPU firmware to execute the kernel. In a lot of cases, between each runs, we just want to update parameters, in such case, we don't want to create new run each time.
And based on the evaluation, each run creation takes ~100us. And thus, it will significantly save kernel launch time, if we can cache previous runs.

This patch introduce a cache to cache the old runs for the function. When calling function, in current supported case, it is the XRT hip stream command, it will call the function's get_run() to allocate a run if no used run in the cache, otherwise, it will return a used run to the calling function. When the XRT hip stream command completes, that is when it is destroyed, in the kernel_start event destructor, it will call function's release_run() to return the run to the cache.

We don't set upper limit to the cache. In HIP stream execution flow, stream will clears all its events when current stream complemetes, that is whenever stream schronization is called, the commands in the stream will be destroyed. The runs used in the streams will be returned to the pool.

In HIP stream execution flow, every kernel launch, we will need to get a run for the command to send to the NPU firmware to execute the kernel. In a lot of cases, between each runs, we just want to update parameters, in such case, we don't want to create new run each time.
And based on the evaluation, each run creation takes ~100us. And thus, it will significantly save kernel launch time, if we can cache previous runs.

Repeat hipModuleLaunchKernel(), and get the average execution time for each run, compared to the raw XRT API kernel launch repeat
```
   for (unsigned int i = 0; i < iter; i++) {
        runf.set_arg(0, 3);
        runf.set_arg(3, obj0);
        runf.set_arg(4, odd(i) ? obj1 : obj2);
        runf.set_arg(5, odd(i) ? obj2 : obj1);
        runf.set_arg(7, odd(i) ? obj4 : obj4b);
    }
```
Seeing >100us difference, checked the HIP implementation, seeing in kernel_start() constructor, it calls `xrt::run` constructor, update this function, seeing the `xrt::run` constructor took this much time, and also update the raw XRT APIs repeat loop to:
```
  for (unsigned int i = 0; i < iter; i++) {
        auto runf = xrt::run(k);
        runf.set_arg(0, 3);
    }
```
seeing ~100us average for each repeat compared to:
```
auto runf = xrt::run(k);
for (unsigned int i = 0; i < iter; i++) {
      runf.set_arg(0, 3);
  }
```
Confirmed the `xrt::run()` took ~100us average for > 1000 run

To solve this issue, will need to cache the run, so that each rerun doesn't need to create a run again.
Thanks to @maxzhen' suggestion, refers to the `exec_buffer_cache`, this implementation follows the pool implementation is widely used in XRT.
However, when to release the run? As HIP XRT maintains the lifetime of a commands in a stream, and each kernel execution command has a run. This implementation returns the run to the pool when the command is destroyed when the command in the stream completes.

Tested repeating launch kernel with HIP command for 100K times.
With this change, we are seeing 100us less than the previous implementation.
Compared to raw XRT API, seeing 3 - 4us extra with HIP. (note: compared to 1K repeats, 100K repeats produce much more stable average time, only ~1us difference in 10runs)


src/runtime_src/hip/core/module.h
```diff
@@ -121,6 +127,27 @@ class function
...
+    return xrt::run(m_xrt_kernel);
+  }
+
+  void
+  release_run(xrt::run& run)
+  {
+    std::scoped_lock lock(m_runs_mutex);
+    m_runs_cache.push_back(std::move(run));
+  }
 };
```
stsoe: Best practice here is to either pass-by-value and move, or pass-by-rvalue and move
So 1) `release_run(xrt::run run)` or 2) `release_run(xrt::run&& run)`  For both you want to std::move to cache, but 
2) is better and the calling code would do `func->release_run(std::move(r)), here it is explicit for calling code that the `r` object is being moved and can no longer be used.

wendyliang25: thanks @stsoe , i have updated the code following 2). could you check again. the func->release_run(std::move(r)) is used in the `kernel_start` destructor. thanks

stsoe (approved):

PR #9110, opened 2025-07-31
tools: common: tests: AIE reconfigOverhead: add return to normal path

wendyliang25:
The test function TestAIEReconfigOverhead::run() requires return, add missing return to the normal path to fix compilation error.


AShivangi (approved):

wendyliang25:
> Yes, a recent mistake was made that was not there yesterday. But nevertheless, Coverity barfed on the non-void function not returning a value so I created a PR yesterday to address that: #9107

Thanks @stsoe , shall i close this PR ? thanks

stsoe:
See #9107

PR #9111, opened 2025-07-31
Reusing device reset function for resetting the device. It was resetting the shared pointer earlier

chvamshi-xilinx:
Reusing device reset function for resetting the device. It was resetting the shared pointer earlier. 
There is a request from a customer to provide a reset function at device level.

Not a Bug

Provided an internal reset function for ocl consumption & updated the device reset functionality fot device reset 

None

Verified hw_emu flows

Yes, need to update reset API documentation


chvamshi-xilinx:
> LGTM. The `virtual reset() {}` will be redefined next, right?

yes Soren. @saifuddin-xilinx  will add the reset code.

PR #9112, opened 2025-07-31
fixing yocto build script to generate embedded RPMs

chvamshi-xilinx:
fixing yocto build script to generate embedded RPMs

Recent change of CMAKE Refactoring caused this regression

Fixed the script

None

Verified build

None


stsoe (commented):
Should you add CMAKE_INSTALL_PREFIX=/usr also to yocto?

chvamshi-xilinx:
> CMAKE_INSTALL_PREFIX=/usr

Yes Soren. That is also required. in Yocto /usr/lib and /lib are symlinks to each other, thatswhy tests worked at my end. Let me add this cmake flag aswell for consistency.

PR #9114, opened 2025-08-01
Aligning bo size for cma allocation

bisingha-xilinx:
https://jira.xilinx.com/browse/CR-1245557
The issue was when we request Linux to allocate cma memory it used to allocate randomly from different different regions for smaller sizes. Which cause test failure for some PL tests in which PL kernel does not have connection or access to that memory region.
Problem was solved by rounding up the size to 2*page_size if the size is smaller and for bigger size, we're just rounding up with the page size.
n/a
Tested the given testcase from the CR. Also tested few edge cases:
1. Simple VADD, 
2. simple GMIO, 
3. too many async operations on a GMIO test)
n/a


chvamshi-xilinx:
Backport this change to 2025.1 also.

PR #9115, opened 2025-08-01
[XDP] Modified Stream width bytes to suppport AIE2PS aie_profiling api

predutta:
Incorrect Stream width when profiling api's were being executed in aie2ps

N/A

Earlier default stream width was considered to be 4 bytes. Problem was solved by adding an if block which checked for the hardware generation, verified if it is aie2ps and accordingly set the stream width to 8bytes.

N/A

This has been tested on the aie2ps board on hw.

N/A


vipangul (approved):

predutta:
Comments added in the code to address the issue of explaining the significance of the numbers of 4 and 8.

pgschuey (approved):

PR #9116, opened 2025-08-02
HIP: fix coverity

wendyliang25:
This patch set fixes HIP coverity reports. except this one:
```
** CID 535924:       Concurrent data access violations  (MISSING_LOCK)
/src/runtime_src/hip/core/event.h: 84           in xrt::core::hip::command::get_time()()


_____________________________________________________________________________________________
*** CID 535924:         Concurrent data access violations  (MISSING_LOCK)
/src/runtime_src/hip/core/event.h: 84             in xrt::core::hip::command::get_time()()
78         return cstate;
79       }
80     
81       std::chrono::time_point<std::chrono::system_clock>
82       get_time()
83       {
>>>     CID 535924:         Concurrent data access violations  (MISSING_LOCK)
>>>     Accessing "this->ctime" without holding lock "xrt::core::hip::event.m_mutex_rec_coms". Elsewhere, "xrt::core::hip::command.ctime" is written to with "event.m_mutex_rec_coms" held 1 out of 1 times.
84         return ctime;
85       }
86     
87       void
88       set_state(state newstate)
89       {

** CID 535923:       Uninitialized variables  (USE_AFTER_MOVE)
/src/runtime_src/core/common/aiebu/src/cpp/common/writer.h: 91           in aiebu::section_writer::add_symbols(std::vector<aiebu::symbol, std::allocator<aiebu::symbol>> &)()
```
Because the `get_time()` is implemented in `class command`, however, the `ctime` is set in `bool event::synchronize()` and `class event` inherits `class command`. and only `event` sets `ctime`, looks like it is better to move `ctime` from `class command` to `class event` ? or if we want to keep `ctime` in the event, how about make `command::get_time()` as virtual or return 0 and let child class does the actual implementation because it is better to keep set ctime and get ctime in the same class.
Hi @chvamshi-xilinx , could you comment on this? thanks

### DIFF
--- a/src/runtime_src/hip/api/hip_event.cpp
+++ b/src/runtime_src/hip/api/hip_event.cpp
@@ -30,6 +30,7 @@ static void hip_event_record(hipEvent_t eve, hipStream_t stream)
   throw_invalid_value_if(!stream, "stream passed is nullptr");
   auto hip_stream = get_stream(stream);
   auto hip_ev = std::dynamic_pointer_cast<event>(command_cache.get(eve));
+  throw_invalid_value_if(!hip_ev, "dynamic_pointer_cast failed");
   hip_ev->record(std::move(hip_stream));
 }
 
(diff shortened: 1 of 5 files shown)


stsoe (approved):

chvamshi-xilinx:
Hi @wendyliang25 ,
You are correct. hipEventElapsedTime(float *ms, hipEvent_t start, hipEvent_t stop) is the only API which requires the time. As this function takes even as arguments ,  You can move the ctime to event class from command class.

PR #9132, opened 2025-08-06
AIESW-5531 Allow mmap shim error through xrt-smi examine

rchane:
https://jira.xilinx.com/browse/AIESW-5531
Discovered while testing, noticed "No devices" when in actuality there was a memlock ulimit issue
Allowed mmap error through previously caught exception.
N/A
Tested on Linux STX:
```
> xrt-smi examine
XRT build version: 2.20.0
Build hash: 15429831f51c5b7b4cedcab7f46528aca1c90cc4
Build date: 2025-08-06 14:15:06
Git branch: HEAD
PID: 2749330
UID: ---
[Wed Aug  6 21:20:04 2025 GMT]
HOST: ---
EXE: /opt/xilinx/xrt/bin/unwrapped/xrt-smi
[xrt-smi] ERROR: mmap(addr=0x793258000000, len=67108864, prot=3, flags=8209, offset=4362100736) failed (err=-11): Resource temporarily unavailable
```
N/A

### DIFF
--- a/src/runtime_src/core/tools/common/XBUtilities.cpp
+++ b/src/runtime_src/core/tools/common/XBUtilities.cpp
@@ -365,7 +365,10 @@ XBUtilities::collect_devices( const std::set<std::string> &_deviceBDFs,
     for(xrt_core::device::id_type index = 0; index < total; ++index) {
       try {
         _deviceCollection.push_back(get_device_internal(index, _inUserDomain));
-      } catch (...) {
+      } catch (const std::runtime_error& e) {
+        if (boost::icontains(e.what(), "mmap"))
+          throw std::runtime_error(e.what());
+      }  catch (...) {
         /* If the device is not available, quietly ignore it
            Use case: when a device is being reset in parallel */
       }


AShivangi (commented):
Should we print out  xrt error or do we need to wrap it and add a comment telling user to increase the ulimit?

rchane:
> Should we print out xrt error or do we need to wrap it and add a comment telling user to increase the ulimit?

I think it's fine to not wrap it to be consistent with the previous error code, considering the user would have already seen the xrt error of "No devices found". In the ticket, Max mentioned this mmap error was specific enough to point to the ulimit issue (or even be pointed out in docs), if that sounds good to you.

maxzhen:
> > Should we print out xrt error or do we need to wrap it and add a comment telling user to increase the ulimit?
> 
> I think it's fine to not wrap it to be consistent with the previous error code, considering the user would have already seen the xrt error of "No devices found". In the ticket, Max mentioned this mmap error was specific enough to point to the ulimit issue (or even be pointed out in docs), if that sounds good to you.

Yes, the ask is "xrt-smi can show the exception message from SHIM". The ask is not about this specific ulimit case.

src/runtime_src/core/tools/common/XBUtilities.cpp
```diff
@@ -365,7 +365,10 @@ XBUtilities::collect_devices( const std::set<std::string> &_deviceBDFs,
     for(xrt_core::device::id_type index = 0; index < total; ++index) {
       try {
         _deviceCollection.push_back(get_device_internal(index, _inUserDomain));
-      } catch (...) {
+      } catch (const std::runtime_error& e) {
+        if (boost::icontains(e.what(), "mmap"))
```
aktondak, line 369: Is it a good idea to have a keyword matching here ? Shouldn't xrt-smi layer just be a passthrough and error out whatever error is thrown from driver without looking at the error itself ? 
@maxzhen

rchane: There is a comment below this for the catch where errors are normally caught and quietly ignored, I wasn't sure to what extent we wanted to preserve this so I caught the mmap first. I am open to throwing all errors if Max and @chvamshi-xilinx  agree!

maxzhen: It has to be throwing all errors by default. You can have exceptions when you know what you're doing.

rchane: Understood, let me follow up with Vamshi for more context on what errors might show up for the commented case and then open another PR. Thank you, Max.

aktondak: @rchane 
Please ensure getting reviews from @chvamshi-xilinx and @maxzhen before the next merge.

PR #9133, opened 2025-08-06
Cert xrt ini

xuhz:
add entry in xrt.ini controlling cert log buf size and file to save log buf
these entries will be used in shim
low


stsoe (approved):

chvamshi-xilinx (commented):
@rbramand-xilinx  is making similar changes. He is creating xrt::bo in hw_context constructor & and dumping the it to a file in  hw_context destructor. That way, logic to get the logs would be same for All the platforms

xuhz:
> @rbramand-xilinx is making similar changes. He is creating xrt::bo in hw_context constructor & and dumping the it to a file in hw_context destructor. That way, logic to get the logs would be same for All the platforms

ok. that would be better.
linux shim needs to change then

chvamshi-xilinx:
We can close this as other PR about log buffer is merged.

PR #9137, opened 2025-08-07
HIP changes to support full ELF flow

rbramand-xilinx:
Added changes in hip module APIs to load full ELF

Hip module load calls are modified to try to load full ELF and when exception is caught fallback to load xclbin + elf 
Ideal way is to modify hip module calls to load full ELFs but as our compilers at present doesn't generate full ELFs and we need to support xclbin + elf flow for some time the changes are made this way.

Also hipModuleLoadData API only takes void* pointer as input but to create elfio object we need data pointer and size, so added a function to calculate size of elf from elf header.

Low as it impacts only HIP flows

Tested existing model test expoi (xclbin + elf) test on strix Linux and it works as expected
Also converted expoi full elf c++ application to HIP based application and tested it on strix Linux, test passes

Need to document the usage when HIP documentation is made available

### DIFF
--- a/src/runtime_src/hip/api/hip_module.cpp
+++ b/src/runtime_src/hip/api/hip_module.cpp
@@ -138,20 +138,23 @@ static size_t
 estimate_elf_size(const void* data)
 {
   auto bytes = static_cast<const unsigned char*>(data);
-  if (bytes[0] != 0x7f || bytes[1] != 'E' || bytes[2] != 'L' || bytes[3] != 'F')
+  constexpr unsigned char ELF_HEADER_MAGIC[] = {0x7f, 'E', 'L', 'F'};
+
+  if (bytes[0] != ELF_HEADER_MAGIC[0] || bytes[1] != ELF_HEADER_MAGIC[1] ||
+      bytes[2] != ELF_HEADER_MAGIC[2] || bytes[3] != ELF_HEADER_MAGIC[3])
     throw std::runtime_error("Invalid ELF magic number");
 
   if (bytes[4] == ELFIO::ELFCLASS32) {
     // 32 bit ELF
     auto header = static_cast<const ELFIO::Elf32_Ehdr*>(data);
-    return std::max(header->e_shoff + header->e_shentsize * header->e_shnum,
-                    header->e_phoff + header->e_phentsize * header->e_phnum);
+    return std::max(header->e_shoff + static_cast<ELFIO::Elf32_Off>(header->e_shentsize) * header->e_shnum,
+                    header->e_phoff + static_cast<ELFIO::Elf32_Off>(header->e_phentsize) * header->e_phnum);
   }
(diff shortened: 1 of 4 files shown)


stsoe (approved):
LGTM.

src/runtime_src/hip/core/module.h
```diff
@@ -115,10 +153,26 @@ class function
 public:
   function() = default;
   function(module_xclbin* mod_hdl, const xrt::module& xrt_module, const std::string& name);
+  function(module_full_elf* mod_hdl, const std::string& name);
 
   module_xclbin*
+  get_xclbin_module() const
```
chvamshi-xilinx: Do we need this function?

src/runtime_src/hip/core/module.h
```diff
@@ -115,10 +153,26 @@ class function
...
+  {
+    return m_xclbin_module;
+  }
+
+  module_full_elf*
+  get_full_elf_module() const
+  {
```
chvamshi-xilinx: Ditto. Do we need this function?

src/runtime_src/hip/api/hip_module.cpp
```diff
@@ -21,12 +23,21 @@ hip_module_launch_kernel(hipFunction_t f, uint32_t /*gridDimX*/, uint32_t /*grid
   auto hip_mod = module_cache.get(static_cast<function*>(func_hdl)->get_module());
   throw_invalid_resource_if(!hip_mod, "module associated with function is unloaded");
 
-  // function store module handle of xclbin
-  auto hip_xclbin_mod = std::dynamic_pointer_cast<module_xclbin>(hip_mod);
-  throw_invalid_resource_if(!hip_xclbin_mod, "getting hip module using dynamic pointer cast failed");
+  std::shared_ptr<function> hip_func;
+  if (hip_mod->is_full_elf_module()) {
```
chvamshi-xilinx: cant we abstract this logic inside get_function API itself?

src/runtime_src/hip/api/hip_module.cpp
```diff
@@ -49,18 +60,30 @@ hip_module_get_function(hipModule_t hmod, const char* name)
   auto mod_hdl = reinterpret_cast<module_handle>(hmod);
   auto hip_mod = module_cache.get(mod_hdl);
   throw_invalid_resource_if(!hip_mod, "module not available");
-  // module handle passed should be elf module handle
-  throw_invalid_resource_if(hip_mod->is_xclbin_module(), "invalid module handle passed");
 
-  auto hip_elf_mod = std::dynamic_pointer_cast<module_elf>(hip_mod);
-  throw_invalid_resource_if(!hip_elf_mod, "getting hip module using dynamic pointer cast failed");
+  if (hip_mod->is_full_elf_module()) {
```
chvamshi-xilinx: ditto, cant we abstract this inside the class itself?

PR #9138, opened 2025-08-07
Update AIESW-5531 Allow shim error through xrt-smi examine

rchane:
https://jira.xilinx.com/browse/AIESW-5531
Update to https://github.com/Xilinx/XRT/pull/9132
Updated default behavior to allow all shim errors through as asked in previous PR.
Waiting on review to double check if there are any edge cases to be considered.
N/A
Tested on STX xdna. 
N/A


AShivangi (approved):

chvamshi-xilinx (approved):
Looks good. There are no such cases for edge as we can have only one device.
Based on the comment in the code, there may be a case for Alveo, where xbutil validate is run for device and xbutil reset is run for another device parallel.

rchane:
Thank you for the review, we'll display all errors for now and catch edge cases if they arise.

PR #9147, opened 2025-08-11
Fix hardwired /opt/xilinx/xrt paths

stsoe:
Use compile time macros for install and library path.

When running XRT applications compiled from debian upstream artifacts, XRT runtime loaded libraries were picked up from non-existent /opt/xilinx/xrt path.

### DIFF
--- a/src/runtime_src/core/common/detail/linux/xilinx_xrt.h
+++ b/src/runtime_src/core/common/detail/linux/xilinx_xrt.h
@@ -1,10 +1,15 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright (C) 2023-2024 Advanced Micro Devices, Inc. All rights reserved.
+// Copyright (C) 2023-2025 Advanced Micro Devices, Inc. All rights reserved.
 
 #include <filesystem>
 #include <stdexcept>
 #include <string>
 #include <vector>
+
+#ifndef XRT_INSTALL_PREFIX
+# error "XRT_INSTALL_PREFIX is undefined"
+#endif
+
 namespace xrt_core::detail {
 
 namespace sfs = std::filesystem;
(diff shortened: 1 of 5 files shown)


src/runtime_src/core/common/detail/linux/xilinx_xrt.h
```diff
@@ -1,28 +1,34 @@
...
 xilinx_xrt()
 {
-#if defined (__aarch64__) || defined (__arm__)
-  return sfs::path("/usr");
-#else
-  return sfs::path("/opt/xilinx/xrt");
-#endif
+  // This returns CMAKE_INSTALL_PREFIX.  The internal default cmake
+  // install path is /opt/xilinx/xrt, for upstreaming most likely /usr.
+  return sfs::path(XRT_INSTALL_PREFIX);
```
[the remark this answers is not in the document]
stsoe: compile time

src/runtime_src/core/common/module_loader.cpp
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright (C) 2016-2020 Xilinx, Inc
-// Copyright (C) 2024 Advanced Micro Devices, Inc. All rights reserved.
+// Copyright (C) 2024-2025 Advanced Micro Devices, Inc. All rights reserved.
```
maxzhen, line 3: only copyright year change in this file?

src/runtime_src/core/common/detail/linux/xilinx_xrt.h
```diff
@@ -1,28 +1,34 @@
...
 }
 
 std::vector<sfs::path>
 platform_repo_path()
 {
-  return {sfs::path("/lib/firmware/amdnpu"), sfs::path("/opt/xilinx/xrt/amdxdna")};
+  // If CMAKE_INSTALL_PREFIX is /usr, then /usr/amdxdna will be invalid,
+  // not sure if this matters or why the second path is even present?
+  // Regardless, default /opt/xilinx/xrt behavior remains unchanged.
+  return {sfs::path("/lib/firmware/amdnpu"), xilinx_xrt() / "amdxdna"};
```
maxzhen, line 31: Is this used by xrt-smi validate? It has to be removed or will be broken in upstream env.

src/runtime_src/core/common/module_loader.cpp
stsoe: Yeah, my mistake.

src/runtime_src/core/common/detail/linux/xilinx_xrt.h
stsoe: I don't know who uses it or for what.  But  you are right, if this is xrt-smi, it is completely wrong.

PR #9150, opened 2025-08-11
Adding configure support to enable/disable event-tracing and firmware…

aktondak:
… logging

This PR adds new xrt-smi configure sub-options to enable and disable event tracing and firmware logging. These sub-options will be enabled from xdna and mcdm shims when the support is there in the driver.

https://jira.xilinx.com/browse/AIESW-4326
https://jira.xilinx.com/browse/AIESW-5344

Solved by adding required handling as a new xrt-smi configure command

N/A

Will be tested when it is enabled from shim

NA

### DIFF
--- a/src/runtime_src/core/tools/xbutil2/OO_EventTrace.cpp
+++ b/src/runtime_src/core/tools/xbutil2/OO_EventTrace.cpp
@@ -1,0 +1,111 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+
+// ------ I N C L U D E   F I L E S -------------------------------------------
+// Local - Include Files
+#include "OO_EventTrace.h"
+#include "tools/common/XBUtilitiesCore.h"
+#include "tools/common/XBUtilities.h"
+#include "tools/common/XBHelpMenusCore.h"
+
+// 3rd Party Library - Include Files
+#include <boost/program_options.hpp>
+#include <boost/algorithm/string.hpp>
+namespace po = boost::program_options;
+
+// ----- C L A S S   M E T H O D S -------------------------------------------
+
+OO_EventTrace::OO_EventTrace( const std::string &_longName, bool _isHidden )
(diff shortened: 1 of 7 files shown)


src/runtime_src/core/tools/xbutil2/OO_EventTrace.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+  m_positionalOptions.
+    add("action", 1 /* max_count */)
+  ;
+}
+
+void
+OO_EventTrace::validate_args() const {
+  if(m_action.empty() && !m_help)
+    throw xrt_core::error(std::errc::operation_canceled, "Please specify a action 'enable', 'disable' or 'status'");
```
AShivangi: Didn't you want to do status through examine?

src/runtime_src/core/tools/xbutil2/OO_FirmwareLog.cpp
```diff
@@ -0,0 +1,114 @@
...
+    : OptionOptions(_longName, _isHidden, "Enable|disable firmware log")
+    , m_device("")
+    , m_action("")
+    , m_help(false)
+    , m_log_level(0)
+{
+  m_optionsDescription.add_options()
+    ("device,d", boost::program_options::value<decltype(m_device)>(&m_device), "The Bus:Device.Function (e.g., 0000:d8:00.0) device of interest")
+    ("help,h", boost::program_options::bool_switch(&m_help), "Help to use this sub-command")
+    ("log-level", boost::program_options::value<decltype(m_log_level)>(&m_log_level), "Log level ranging from 0 to 5")
```
AShivangi: log-level DRC? what if value is >5.

Also how is this passed to the driver?

aktondak: I have removed it. Today the shim is not ready for enabling disabling so this PR mostly for checking once its there.
The DRC should not exist in xrt-smi ideally and should through whatever error firmware has if log level is set to out of bounds. I have inculcated it.

rchane (approved):

PR #9156, opened 2025-08-14
xrt-smi examine implementation for context health report

aktondak:
This PR implements the Context Health Monitoring Report for XRT-SMI. This PR also implements the routines to allow a watch mode for reports which will be subsequently used for upcoming reports namely event tracing and firmware logging.

https://jira.xilinx.com/browse/AIESW-8460

[report_watch_mode]

- Generic utility class for implementing watch mode functionality across any XRT-SMI report
- Specific report implementation for hardware context health monitoring

Usage:

Standard xrt-smi report interface: **xrt-smi examine --report context-health**
Filter support: --element ctx_id=<id1>,<id1>... for specific context monitoring

Currently the the shim driver implementation is being updated to take a hardware context as parameter and query firmware only for that context. This report will be updated to take the list of hardware contexts queried once that change is in.

None. New implementation

Tested on Linux. The watch mode has been tested on windows but the report is not tested since the shim changes are being planned with [this change](https://gitenterprise.xilinx.com/XRT/XRT-MCDM/pull/1860).

Following is the report that is shown :
```
---------------------------
[0000:c5:00.1] : NPU Strix
---------------------------
Context Health Report
=====================

Total Contexts: 1

  Context Health Information:
    |Txn Op Idx  |Ctx PC      |Fatal Err Type  |Fatal Err Ex Type  |Fatal Err Ex PC  |Fatal App Module  |
    |------------|------------|----------------|-------------------|-----------------|------------------|
    |0xffffffff  |0xdeadbeef  |0               |0                  |0x0              |0                 |

(Press Ctrl+C to exit watch mode | Last update: Sat Aug 16 02:49:41 2025 GMT)
Context Health Report (Total: 0) - Sat Aug 16 02:49:42 2025 GMT
```

None


src/runtime_src/core/common/query_requests.h
```diff
@@ -1981,6 +1982,29 @@ struct aie_partition_info : request
...
+  struct data
+  {
+    uint32_t context_id;
+    uint32_t txn_op_idx;
+    uint32_t ctx_pc;
+    uint32_t fatal_error_type;
+    uint32_t fatal_error_exception_type;
+    uint32_t fatal_error_exception_pc;
+    uint32_t fatal_error_app_module;
+  };
```
stsoe: I think we need to leverage ert.h: struct ert_ctx_health_data, so data should be
`using data = struct ert_ctx_health_data;` othewise things will get out-of-sync.

src/runtime_src/core/common/query_requests.h
```diff
@@ -1981,6 +1982,29 @@ struct aie_partition_info : request
...
+
+  using result_type = std::vector<struct data>;
```
stsoe:
```suggestion
  using result_type = std::vector<data>;
```

src/runtime_src/core/tools/common/ReportWatchMode.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include "core/common/time.h"
+
+// 3rd Party Library - Include Files
+#include <chrono>
+#include <thread>
+#include <csignal>
+#include <atomic>
+#include <iomanip>
+#include <sstream>
+#include <algorithm>
```
stsoe: sort

src/runtime_src/core/tools/common/ReportWatchMode.h
```diff
@@ -0,0 +1,140 @@
...
+#pragma once
+
+// Please keep external include file dependencies to a minimum
+#include <vector>
+#include <string>
+#include <functional>
+#include <iostream>
+
+namespace xrt_core {
+  class device;
```
stsoe:
```suggestion
class device;
```
Let's not indent namespace code.

src/runtime_src/core/tools/common/ReportWatchMode.h
```diff
@@ -0,0 +1,140 @@
...
+  /**
+   * @brief Restore the original SIGINT signal handler
+   * 
+   * - Restores the signal handler that was active before setup_signal_handler()
+   * - Clears internal state flags
+   * - Safe to call multiple times or without prior setup
+   * 
+   * @note Should be called before exiting watch mode
+   */
+  static void restore_signal_handler();
```
stsoe: Could these be static functions in the unnamed namespace in ReportContextHealth.cpp?  Since they are private, they can only be used within ReportWatchMode, so having them in the API header is just noise.  What I would do in .cpp is create 
```
namespace signal_hander {
// the global variables that are related to signal_handler goes here
// make sure to declare them static
...

// signal handler APIs
static void 
setup()
{
   ...
}

static void
restore()
{
   ...
}

static void 
watch()
{ 
  ...
}
}
```

src/runtime_src/core/tools/common/reports/ReportContextHealth.cpp
```diff
@@ -0,0 +1,229 @@
...
+#include "core/common/query_requests.h"
+#include "core/common/time.h"
+#include "tools/common/ReportWatchMode.h"
+
+// 3rd Party Library - Include Files
+#include <boost/format.hpp>
+#include <vector>
+#include <iomanip>
+#include <sstream>
+#include <algorithm>
```
stsoe:
```suggestion
#include <boost/format.hpp>

#include <algorithm>
#include <iomanip>
#include <sstream>
#include <vector>
```

src/runtime_src/core/tools/common/reports/ReportContextHealth.cpp
```diff
@@ -0,0 +1,229 @@
...
+// Helper structure to hold context data for table generation
+struct ContextData {
+  uint32_t context_id;
+  uint32_t txn_op_idx;
+  uint32_t ctx_pc;
+  uint32_t fatal_error_type;
+  uint32_t fatal_error_exception_type;
+  uint32_t fatal_error_exception_pc;
+  uint32_t fatal_error_app_module;
+};
```
stsoe: Do not duplicate the struct that is already defined in ert.h

src/runtime_src/core/tools/common/reports/ReportContextHealth.cpp
```diff
@@ -0,0 +1,229 @@
...
+  ss << "    |";
+  for (const auto& col : columns) {
+    ss << std::string(col.width - 1, '-') << "|";
+  }
+  ss << "\n";
+}
+
+void
+ReportContextHealth::getPropertyTreeInternal(const xrt_core::device* dev,
+                                            bpt& pt) const
```
stsoe: My personal taste, you can ignore.  Makes scanning the code easier.
```suggestion
void
ReportContextHealth::
getPropertyTreeInternal(const xrt_core::device* dev, bpt& pt) const
```

src/runtime_src/core/tools/common/reports/ReportContextHealth.cpp
```diff
@@ -0,0 +1,229 @@
...
+      context_pt.put("ctx_pc", context.ctx_pc);
+      context_pt.put("fatal_error_type", context.fatal_error_type);
+      context_pt.put("fatal_error_exception_type", context.fatal_error_exception_type);
+      context_pt.put("fatal_error_exception_pc", context.fatal_error_exception_pc);
+      context_pt.put("fatal_error_app_module", context.fatal_error_app_module);
+      contexts_array.push_back(std::make_pair("", context_pt));
+    }
+    context_health_pt.add_child("contexts", contexts_array);
+    context_health_pt.put("context_count", context_health_data.size());
+  } catch (const std::exception& e) {
```
stsoe: Aligning try/catch makes it easier to scan the code.
```suggestion
  } 
  catch (const std::exception& e) {
```

src/runtime_src/core/tools/common/reports/ReportContextHealth.cpp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// Helper function to parse context IDs from elements_filter
+static std::vector<uint32_t> parse_context_ids(const std::vector<std::string>& elements_filter) {
+  std::vector<uint32_t> context_ids;
+  
+  for (const auto& filter : elements_filter) {
+    if (filter.find("ctx_id=") == 0) {
+      std::string ctx_ids_str = filter.substr(7); // // NOLINT(cppcoreguidelines-avoid-magic-numbers) 
+      std::stringstream ctx_stream(ctx_ids_str);
```
stsoe: I think you should define the stringstream outside the loop and reseed inside the loop using `std::stringstream::set(std::string)`

src/runtime_src/core/tools/common/reports/ReportContextHealth.cpp
```diff
@@ -0,0 +1,229 @@
...
+  }
+  ss << "\n";
+  
+  // Generate separator line
+  ss << "    |";
+  for (const auto& col : columns) {
+    ss << std::string(col.width - 1, '=') << "|";
+  }
+  ss << "\n";
+}
```
AShivangi: Doesn't 2DTable takes care of this?

src/runtime_src/core/tools/xbutil2/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ file(GLOB XBUTIL_V2_BASE_FILES
   "../common/reports/platform/*.cpp"
   "../common/reports/aie/*.cpp"
   "../common/Report.cpp"
+  "../common/ReportWatchMode.cpp"
```
AShivangi: Reports should go in `src/runtime_src/core/tools/common/reports/` so they can be automatically be picked up by L18

src/runtime_src/core/tools/common/ReportWatchMode.h
aktondak: That;s fair too. I've moved this to cpp in a named namespace

src/runtime_src/core/tools/common/reports/ReportContextHealth.cpp
aktondak: I started this report as non 2D. But yes using a existing utility makes more sense here. I have updated.

src/runtime_src/core/tools/xbutil2/CMakeLists.txt
aktondak: I have renamed the file. I'd prefer for this to stay out of report since this is intended to be a generic utility for now

src/runtime_src/core/tools/common/reports/ReportContextHealth.cpp
aktondak: Used the ert struct now

src/runtime_src/core/common/query_requests.h
aktondak: This makes sense. This makes my life easier too to maintain less code. I've updated to use @sayyanna 's ert struct for context health

src/runtime_src/core/tools/common/SmiWatchMode.cpp
```diff
@@ -0,0 +1,155 @@
...
+  // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables) - Required for signal handling
+  std::atomic<bool> g_watch_interrupted{false};
+
+  // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables) - Required for signal handling
+  void (*g_old_signal_handler)(int) = nullptr;
+
+  // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables) - Required for signal handling
+  bool g_signal_handler_set = false;
+
+  void watch_signal_handler(int signal) {
```
stsoe: When namespaced, you should use much simpler names, like just `watch`, it will be called as `signal_handler::watch(..)`

src/runtime_src/core/tools/common/SmiWatchMode.cpp
```diff
@@ -0,0 +1,155 @@
...
+   * 
+   * - Saves the current SIGINT handler for restoration later
+   * - Installs custom handler that sets atomic interrupt flag
+   * - Uses thread-safe signal handling compatible with XRT patterns
+   * - Only installs once (subsequent calls are ignored)
+   * 
+   * @note Must be paired with restore_signal_handler() call
+   * @note Uses standard signal() function for cross-platform compatibility
+   */
+  void setup_signal_handler() {
```
stsoe: Rename to `setup()`

src/runtime_src/core/tools/common/SmiWatchMode.cpp
```diff
@@ -0,0 +1,155 @@
...
+  /**
+   * @brief Restore the original SIGINT signal handler
+   * 
+   * - Restores the signal handler that was active before setup_signal_handler()
+   * - Clears internal state flags
+   * - Safe to call multiple times or without prior setup
+   * 
+   * @note Should be called before exiting watch mode
+   */
+  void restore_signal_handler() {
```
stsoe: Rename to `restore()`

src/runtime_src/core/tools/common/SmiWatchMode.cpp
```diff
@@ -0,0 +1,155 @@
...
+                               const ReportGenerator& report_generator,
+                               const std::string& report_title)
+{
+  if (!device || !report_generator) {
+    output << "Error: Invalid device or report generator provided to watch mode\n";
+    return;
+  }
+
+  // Set up signal handler for Ctrl+C
+  signal_handler::setup_signal_handler();
```
stsoe:
```suggestion
  signal_handler::setup();
```

src/runtime_src/core/tools/common/SmiWatchMode.cpp
```diff
@@ -0,0 +1,155 @@
...
+  
+  output << "Starting " << report_title << " Watch Mode (Press Ctrl+C to exit)\n";
+  output << "Update interval: 1 second\n";
+  output << "=======================================================\n\n";
+  output.flush();
+  
+  signal_handler::g_watch_interrupted = false;
```
stsoe: You might want to do `signal_hander::interrupt()` to keep this variable encapsulated.

src/runtime_src/core/tools/common/SmiWatchMode.cpp
```diff
@@ -0,0 +1,155 @@
...
+  std::string last_report;
+  
+  // Filter out watch-specific options for the report generator
+  auto filtered_elements = filter_out_watch_options(elements_filter);
+  
+  while (!signal_handler::g_watch_interrupted) {
```
stsoe: How about `while (signal_handler::active())`.   ?

src/runtime_src/core/tools/common/reports/ReportContextHealth.cpp
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  // There can only be 1 root node
+  pt.add_child("context_health", context_health_pt);
+}
+
+// Helper function to parse context IDs from elements_filter
+static std::vector<uint32_t> 
+parse_context_ids(const std::vector<std::string>& elements_filter) {
+  std::vector<uint32_t> context_ids;
```
stsoe: Consistency

```suggestion
parse_context_ids(const std::vector<std::string>& elements_filter) 
{
  std::vector<uint32_t> context_ids;
```

src/runtime_src/core/tools/common/reports/ReportContextHealth.cpp
```diff
@@ -0,0 +1,177 @@
...
+{
+  // Defer to the 20202 format.  If we ever need to update JSON data,
+  // Then update this method to do so.
+  getPropertyTree20202(dev, pt);
+}
+
+void
+ReportContextHealth::
+getPropertyTree20202(const xrt_core::device* dev,
+                                         bpt& pt) const
```
stsoe: This should fit ?
```suggestion
getPropertyTree20202(const xrt_core::device* dev, bpt& pt) const
```

src/runtime_src/core/tools/common/reports/ReportContextHealth.cpp
```diff
@@ -0,0 +1,177 @@
...
+      }
+      break;
+    }
+  }
+  
+  return context_ids;
+}
+
+static std::string
+generate_context_health_report(const xrt_core::device* dev, const std::vector<std::string>& elements_filter, bool include_timestamp = true) {
```
stsoe: This on the other hand does not fit, so break it.    Not sure why it doesn't align the the suggestion ?
```suggestion
static std::string
generate_context_health_report(const xrt_core::device* dev, 
                                                    const std::vector<std::string>& elements_filter, 
                                                    bool include_timestamp = true) 
{
```

stsoe (changes requested):
Some minors.

PR #9161, opened 2025-08-16
Add simple archive extraction utility to XRT

stsoe:
Use ARIO (part of ELFIO) as an archive extraction utility.

Idea is that xrt-smi will read validation data from an archive installed at build time.  The archive creation can be static and checked into VTD, or we can enhance build of XRT-MCDM and amdxdna to create the archive from VTD artifacts at build time.

Update ELFIO for archive support in newly added ARIO module.

### DIFF
--- a/src/runtime_src/core/common/CMakeLists.txt
+++ b/src/runtime_src/core/common/CMakeLists.txt
@@ -23,6 +23,7 @@ else()
 endif()
 
 add_library(core_common_library_objects OBJECT
+  archive.cpp
   config_reader.cpp
   debug.cpp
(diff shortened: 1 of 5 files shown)


stsoe:
> We should add a public API to enumerate all the entries in the archive (equivalent of option **t** in GNU ar tool)?

@sonals How would this API be used and by whom?   Do you see xrt-smi needing this? We already have GNU ar on Linux, and I am not adding any new tools. This PR is about extraction for individual files from an archive.

aktondak (approved):

PR #9166, opened 2025-08-18
Add changes to dump UC log buffer

rbramand-xilinx:
Added changes to dump firmware (uC) log buffer when ini option is enabled.

In hw ctx constructor, log buffer is created and initialized if ini option for firmware logging is enabled and the buffer is dumped in hw ctx destructor. We are dumping each uC log in separate file so that it is easy for user to understand the log. To identify between hwctx's its slotidx is used in file name.

Low as it doesn't impact regular flow

Tested uC log buffer dumping in Telluride by implementing required virtual functions in Telluride shim

May be we need to document about the new ini option.

### DIFF
--- a/src/runtime_src/core/common/api/bo_int.h
+++ b/src/runtime_src/core/common/api/bo_int.h
@@ -35,19 +35,21 @@ enum class use_type {
 
 // create_bo() - Create a buffer object within a device for specific use case
 //
-// Allocates a buffer object in the given device. All the public xrt::bo constructors
-// doesnt use 64 bit flags. So this function acts as an extension to create buffer
-// with specific use flag (debug/dtrace/log). This API is useful for creating buffers
-// that outlive hw context.
+// Allocates a buffer object in the given device. All the public
+// xrt::bo constructors doesnt use 64 bit flags. So this function acts
+// as an extension to create buffer with specific use flag
+// (debug/dtrace/log). This API is useful for creating buffers that
+// outlive hw context.
 XRT_CORE_COMMON_EXPORT
 xrt::bo
 create_bo(const std::shared_ptr<xrt_core::device>& m_core_device, size_t sz, use_type type);
 
 // create_bo() - Create a buffer object within a hwctx for specific use case
 //
-// Allocates a buffer object within a hwctx. All the public xrt::bo constructors
-// doesnt use 64 bit flags. So this function acts as an extension to create buffer
-// with specific use flag (debug/dtrace/log).
+// Allocates a buffer object within a hwctx. All the public xrt::bo
+// constructors doesnt use 64 bit flags. So this function acts as an
+// extension to create buffer with specific use flag
+// (debug/dtrace/log).
 XRT_CORE_COMMON_EXPORT
 xrt::bo
 create_bo(const xrt::hw_context& hwctx, size_t sz, use_type type);
@@ -55,11 +57,12 @@ create_bo(const xrt::hw_context& hwctx, size_t sz, use_type type);
 // config_bo() - Configure the buffer object for the given use case (based on
 // the flag passed)
 //
-// Configure the buffer object to be used for debug, dtrace, log, debug queue
-// purpose based on buffer type. The buffer object is tied to a slot using
-// the hw ctx passed to this call. If the hwctx passed in null then hwctx that is
-// used to create the bo is used. A map of uc or column index and
-// buffer size is used to split the buffer among the columns in the partition/slot.
+// Configure the buffer object to be used for debug, dtrace, log,
+// debug queue purpose based on buffer type. The buffer object is tied
+// to a slot using the hw ctx passed to this call. If the hwctx passed
+// in null then hwctx that is used to create the bo is used. A map of
+// uc or column index and buffer size is used to split the buffer
+// among the columns in the partition/slot.
 XRT_CORE_COMMON_EXPORT
 void
 config_bo(const xrt::bo& bo, const std::map<uint32_t, size_t>& buf_sizes,
@@ -67,11 +70,11 @@ config_bo(const xrt::bo& bo, const std::map<uint32_t, size_t>& buf_sizes,
 
 // unconfig_bo() - Unconfigure the buffer object configured earlier
 //
-// XRT Userspace can use this function to explicitly unconfigure the bo. It gives
-// more control on when to unconfigure the bo instead of relying on buffer handle
-// destruction
-// NOTE: The caller must ensure that the same buffer object and context handle
-// used in the corresponding config_bo() call are passed to this function.
+// XRT Userspace can use this function to explicitly unconfigure the
+// bo. It gives more control on when to unconfigure the bo instead of
+// relying on buffer handle destruction NOTE: The caller must ensure
+// that the same buffer object and context handle used in the
+// corresponding config_bo() call are passed to this function.
 XRT_CORE_COMMON_EXPORT
 void
 unconfig_bo(const xrt::bo& bo, const xrt_core::hwctx_handle* ctx_handle = nullptr);
--- a/src/runtime_src/core/common/api/bo_int.h
+++ b/src/runtime_src/core/common/api/bo_int.h
@@ -33,6 +33,16 @@ enum class use_type {
   uc_debug = XRT_BO_USE_UC_DEBUG // For microblaze debug data
 };
 
+// create_bo() - Create a buffer object within a device for specific use case
+//
+// Allocates a buffer object in the given device. All the public xrt::bo constructors
+// doesnt use 64 bit flags. So this function acts as an extension to create buffer
+// with specific use flag (debug/dtrace/log). This API is useful for creating buffers
+// that outlive hw context.
+XRT_CORE_COMMON_EXPORT
+xrt::bo
+create_bo(std::shared_ptr<xrt_core::device> m_core_device, size_t sz, use_type type);
+
 // create_bo() - Create a buffer object within a hwctx for specific use case
 //
 // Allocates a buffer object within a hwctx. All the public xrt::bo constructors
@@ -47,11 +57,13 @@ create_bo(const xrt::hw_context& hwctx, size_t sz, use_type type);
 //
 // Configure the buffer object to be used for debug, dtrace, log, debug queue
 // purpose based on buffer type. The buffer object is tied to a slot using
-// the hw ctx that is used to create the bo. A map of uc or column index and
+// the hw ctx passed to this call. If the hwctx passed in null then hwctx that is
+// used to create the bo is used. A map of uc or column index and
 // buffer size is used to split the buffer among the columns in the partition/slot.
 XRT_CORE_COMMON_EXPORT
 void
-config_bo(const xrt::bo& bo, const std::map<uint32_t, size_t>& buf_sizes);
+config_bo(const xrt::bo& bo, const std::map<uint32_t, size_t>& buf_sizes,
+          const xrt_core::hwctx_handle* ctx_handle = nullptr);
 
 // unconfig_bo() - Unconfigure the buffer object configured earlier
 //
@@ -60,7 +72,7 @@ config_bo(const xrt::bo& bo, const std::map<uint32_t, size_t>& buf_sizes);
 // destruction
 XRT_CORE_COMMON_EXPORT
 void
-unconfig_bo(const xrt::bo& bo);
+unconfig_bo(const xrt::bo& bo, const xrt_core::hwctx_handle* ctx_handle = nullptr);
 
 } // bo_int, xrt_core
 
--- a/src/runtime_src/core/common/api/bo_int.h
+++ b/src/runtime_src/core/common/api/bo_int.h
@@ -41,7 +41,7 @@ enum class use_type {
 // that outlive hw context.
 XRT_CORE_COMMON_EXPORT
 xrt::bo
-create_bo(std::shared_ptr<xrt_core::device> m_core_device, size_t sz, use_type type);
+create_bo(const std::shared_ptr<xrt_core::device>& m_core_device, size_t sz, use_type type);
 
 // create_bo() - Create a buffer object within a hwctx for specific use case
 //
@@ -70,6 +70,8 @@ config_bo(const xrt::bo& bo, const std::map<uint32_t, size_t>& buf_sizes,
 // XRT Userspace can use this function to explicitly unconfigure the bo. It gives
 // more control on when to unconfigure the bo instead of relying on buffer handle
 // destruction
+// NOTE: The caller must ensure that the same buffer object and context handle
+// used in the corresponding config_bo() call are passed to this function.
 XRT_CORE_COMMON_EXPORT
 void
 unconfig_bo(const xrt::bo& bo, const xrt_core::hwctx_handle* ctx_handle = nullptr);
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -1698,8 +1698,8 @@ get_offset(const xrt::bo& bo)
   return handle->get_offset();
 }
 
-xrt::bo
-create_bo(const std::shared_ptr<xrt_core::device>& device, size_t sz, use_type type)
+static xrtBufferFlags
+compose_internal_bo_flags(use_type type)
 {
   xcl_bo_flags flags {0};  // see xrt_mem.h
   flags.flags = XRT_BO_FLAGS_CACHEABLE;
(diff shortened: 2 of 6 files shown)


src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -101,8 +153,52 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
   }
 
+  // Initializes uc log buffer, configures it by splitting the buffer
+  // equally among available columns
+  // Made this API public so that it can be called from xrt::hw_context constructor
+  void
+  initialize_uc_log_buffer()
+  {
+    // Create Log buffer for uc logging if ini option is enabled
+    static auto uc_log_buf_size = xrt_core::config::get_uc_log_buffer_size();
```
xuhz: since we are going to set equal buf size to each uc, how about we specify  *per uc buffer size* in xrt.ini?
then we don't need to divide here. but we need to do multiply for create_bo.
there is one advantage here, since cert has a restriction on the buf size -- it has to be multiple of 32B, we need to do a check here and do roundup if necessary.
so per uc size would be easier

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -61,6 +69,50 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
+        return;
+
+      // sync the log buffer
+      m_uc_log_bo.sync(XCL_BO_SYNC_BO_FROM_DEVICE);
+      auto num_uc = m_hdl->get_num_uc();
+      // split buffer equally among available columns
+      auto uc_buf_size = m_uc_log_bo.size() / num_uc;
+
+      // dump the log buffer for each uc in a separate file
+      for (size_t i = 0; i < num_uc; i++) {
```
xuhz: the file is binary, so a name with ext 'txt' looks confusing.
do we need to add 'date' to filename so that when same ctx id is reused, we don't overwrite the log?

xuhz: nits Q:
do we want to save in one big file for multi-uc? i am ok with either way one big file or one file for each uc.
we need a tool to interpret the binary anyway.

src/runtime_src/core/common/api/xrt_hw_context.cpp
rbramand-xilinx: sure @xuhz thanks, will make changes accordingly

src/runtime_src/core/common/api/xrt_hw_context.cpp
rbramand-xilinx: I was using older cert which dumped readable content so used .txt extension.
I will change the extension and also include date in the file name. Thanks

rbramand-xilinx: > nits Q: do we want to save in one big file for multi-uc? i am ok with either way one big file or one file for each uc. we need a tool to interpret the binary anyway.

I thought it would easy to interpret if we dump into separate files, but I am also okay either way

src/runtime_src/core/common/shim/hwctx_handle.h
```diff
@@ -55,6 +55,13 @@ class hwctx_handle
   virtual slot_id
   get_slotidx() const = 0;
 
+  // Get number of UCs in the hardware context created
```
xuhz, line 58: do we want to this implemented in shim?
xrt (elfloader) doesn't have this info?

chvamshi-xilinx: @xuhz , hardware context should have this information. It knows how many columns in a partition (hardware context). Based on architecture, it knows number of ucs as well.

src/runtime_src/core/common/api/bo_int.h
```diff
@@ -33,6 +33,16 @@ enum class use_type {
...
 
+// create_bo() - Create a buffer object within a device for specific use case
+//
+// Allocates a buffer object in the given device. All the public xrt::bo constructors
+// doesnt use 64 bit flags. So this function acts as an extension to create buffer
+// with specific use flag (debug/dtrace/log). This API is useful for creating buffers
+// that outlive hw context.
+XRT_CORE_COMMON_EXPORT
+xrt::bo
+create_bo(std::shared_ptr<xrt_core::device> m_core_device, size_t sz, use_type type);
```
stsoe:
```suggestion
create_bo(const std::shared_ptr<xrt_core::device>& device, size_t sz, use_type type);
```

src/runtime_src/core/common/api/xrt_bo.cpp
```diff
@@ -1715,19 +1731,22 @@ create_bo(const xrt::hw_context& hwctx, size_t sz, use_type type)
...
+config_bo(const xrt::bo& bo, const std::map<uint32_t, size_t>& buf_sizes,
+          const xrt_core::hwctx_handle* ctx_handle)
 {
   auto bo_impl = bo.get_handle();
-  auto ctx = bo_impl->get_hwctx_handle();
-  bo_impl->get_handle()->config(ctx, buf_sizes);
+  // use the ctx handle passed to this call
+  // else use ctx handle used to create this buffer object
+  ctx_handle ? bo_impl->get_handle()->config(ctx_handle, buf_sizes)
+             : bo_impl->get_handle()->config(bo_impl->get_hwctx_handle(), buf_sizes);
```
stsoe: Formatting to make debugging easier
```suggestion
  ctx_handle 
    ? bo_impl->get_handle()->config(ctx_handle, buf_sizes)
    : bo_impl->get_handle()->config(bo_impl->get_hwctx_handle(), buf_sizes);
```

src/runtime_src/core/common/api/xrt_bo.cpp
```diff
@@ -1715,19 +1731,22 @@ create_bo(const xrt::hw_context& hwctx, size_t sz, use_type type)
...
 
 void
-unconfig_bo(const xrt::bo& bo)
+unconfig_bo(const xrt::bo& bo, const xrt_core::hwctx_handle* ctx_handle)
 {
   auto bo_impl = bo.get_handle();
-  auto ctx = bo_impl->get_hwctx_handle();
-  bo_impl->get_handle()->unconfig(ctx);
+  ctx_handle ? bo_impl->get_handle()->unconfig(ctx_handle)
+             : bo_impl->get_handle()->unconfig(bo_impl->get_hwctx_handle());
```
stsoe: Ditto formatting to easy debugging

src/runtime_src/core/common/api/xrt_bo.cpp
```diff
@@ -1698,6 +1698,22 @@ get_offset(const xrt::bo& bo)
   return handle->get_offset();
 }
 
+xrt::bo
+create_bo(std::shared_ptr<xrt_core::device> m_core_device, size_t sz, use_type type)
```
stsoe:
```suggestion
create_bo(const std::shared_ptr<xrt_core::device>& device, size_t sz, use_type type)
```

src/runtime_src/core/common/api/xrt_bo.cpp
```diff
@@ -1698,6 +1698,22 @@ get_offset(const xrt::bo& bo)
...
+  flags.flags = XRT_BO_FLAGS_CACHEABLE;
+  flags.access = XRT_BO_ACCESS_LOCAL;
+  flags.dir = XRT_BO_ACCESS_READ_WRITE;
+  flags.use = static_cast<uint32_t>(type);
+
+  // While the memory group should be ignored (inferred) for
+  // debug / trace / log buffers, it is still passed in as a default
+  // group 1 with no implied correlation to xclbin connectivity
+  // or memory group.
+  return xrt::bo{alloc(device_type{m_core_device}, sz, flags.all, 1)};
```
stsoe:
```suggestion
  return xrt::bo{alloc(device_type{device}, sz, flags.all, 1)};
```

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -278,6 +388,7 @@ post_alloc_hwctx(const std::shared_ptr<hw_context_impl>& handle)
   // called in XDP create a hw_context to the underlying implementation
   xrt_core::xdp::update_device(handle.get(), true);
   handle->get_usage_logger()->log_hw_ctx_info(handle.get());
+  handle->initialize_uc_log_buffer();
```
stsoe: We are sure initialize_uc_log_buffer() belongs in this function and not in constructor initializer list?  This function is specific to XDP meaning it is called only when XDP is configured specially in ini.

stsoe (changes requested):

src/runtime_src/core/common/api/xrt_hw_context.cpp
rbramand-xilinx: Hi Soren, are you talking about post_alloc_hwctx function?
The post_alloc_hwctx is a common function (to avoid code duplicacy) called with hw ctx impl pointer created from different constructors.

stsoe: got it thanks, my oversight.

chvamshi-xilinx (approved):

src/runtime_src/core/common/api/bo_int.h
```diff
@@ -47,11 +57,13 @@ create_bo(const xrt::hw_context& hwctx, size_t sz, use_type type);
 //
 // Configure the buffer object to be used for debug, dtrace, log, debug queue
 // purpose based on buffer type. The buffer object is tied to a slot using
-// the hw ctx that is used to create the bo. A map of uc or column index and
+// the hw ctx passed to this call. If the hwctx passed in null then hwctx that is
+// used to create the bo is used. A map of uc or column index and
```
stsoe: - What happens if config() is called with ctx_handle, but unconfig() is not?   
- What happens if config() is called but unconfig() is not?
- What happens if unconfig() is called, but config() was not?

It is a very error prone design.  It would be much better if config_bo() returned an RAII object that managed what needs to be manged so that calling unconfig() is not required.   In fact if any of the bullet questions result in unrecoverable error, then we must make changes to fix this.

rbramand-xilinx: Hi Soren, create_bo returns xrt::bo object. config_bo is just a bo_int function that can be called with xrt::bo argument.
config_bo doesn't return any object.
And also in xrt::bo shim handle destruction we are checking if its a special bo (log, debug, dtrace) and calling unconfig_bo.
We have provided unconfig_bo internal api so that we have more flexibility to unconfigure the address at any point instead of relying on bo destructor call.
Even though unconfig_bo is not called we unconfigure at the time of bo destruction. unconfig_bo function is for ease of use

rbramand-xilinx: Hi Soren, as discussed offline we can have a ticket to change the design to have RAII style. I have updated the comment for unconfig_bo call to use it cautiously.

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -105,8 +168,55 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
+  // Initializes uc log buffer, configures it by splitting the buffer
+  // equally among available columns
+  // Made this API public so that it can be called from xrt::hw_context constructor
+  void
+  initialize_uc_log_buffer()
+  {
+    // Create Log buffer for uc logging if ini option is enabled
+    static auto uc_log_buf_size = xrt_core::config::get_log_buffer_size_per_uc();
+    if (!uc_log_buf_size || m_uc_log_bo_initialized)
+      return;
```
stsoe: So this function will be executed always and proceed to try/catch block even if uc log buffer is not supported.  I guess it is okay, and that it is intentional that a stray xrt.ini will result in an exception on anything but aie4?

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -105,8 +168,55 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
     return shared_from_this();
   }
 
+  // Initializes uc log buffer, configures it by splitting the buffer
+  // equally among available columns
+  // Made this API public so that it can be called from xrt::hw_context constructor
+  void
+  initialize_uc_log_buffer()
```
stsoe: Here I will again argue that initialize (called during ctor) and dump (called during dtor) should be managed by an embedded struct (say `struct hw_context_impl::uc_log_buffer`) stored as a data member of `hw_context_impl`.  The embedded struct ctor would do the work if initialize_uc_log_buffer() and the dtor would call dump().

src/runtime_src/core/common/api/xrt_hw_context.cpp
rbramand-xilinx: Yes Soren

src/runtime_src/core/common/api/xrt_hw_context.cpp
rbramand-xilinx: sure Soren, I will refactor the code accordingly. Thanks

rbramand-xilinx: I have refactored the code as suggested Soren, please review.

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -37,13 +48,103 @@
...
+      xrt_core::message::send(xrt_core::message::severity_level::debug, "xrt_hw_context", ss.str());
+    }
+
+    static xrt::bo
+    init_and_get_uc_log_bo(std::shared_ptr<xrt_core::device> device,
+                           xrt_core::hwctx_handle* ctx_hdl,
+                           size_t size_per_uc,
+                           size_t num_uc)
+    {
+      auto bo = xrt_core::bo_int::create_bo(std::move(device),
```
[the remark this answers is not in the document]
stsoe: This is fair, you should use `const std::shared_ptr<xrt_core::device>&`.

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -37,13 +48,103 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
   using qos_type = cfg_param_type;
   using access_mode = xrt::hw_context::access_mode;
 
+  // Struct used for initializing and dumping firmware log buffer
+  struct uc_log_buffer
```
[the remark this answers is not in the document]
stsoe: Ignore.  The impl class is never copied, so uc_log_buffer isn't copied either.

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -37,13 +48,103 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
+  {
+    size_t m_num_uc; // number of uc's
+    uint32_t m_slot_idx; // index of slot in which these uc's are present
+    size_t m_size_per_uc;
+    xrt::bo m_uc_log_bo; // log buffer used for uc logging
+
+    static void
```
stsoe: This probably shouldn't be static.  Let this function use the data member, so no need to pass the bo.
Note static is required for auxiliary functions used in constructor initializer list, but this function is not, as far as I can tell.

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -37,13 +48,103 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
+      xrt_core::bo_int::config_bo(bo, uc_buf_map, ctx_hdl); // configure the log buffer
+
+      xrt_core::message::send(xrt_core::message::severity_level::debug, "xrt_hw_context",
+                              "uC log buffer initialized successfully");
+
+      return bo;
+    }
+
+    // may throw
+    uc_log_buffer(std::shared_ptr<xrt_core::device> device,
```
stsoe:
```suggestion
    uc_log_buffer(const std::shared_ptr<xrt_core::device>& device,
```

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -65,13 +166,39 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
+    if (!uc_log_buf_size || !ctx_hdl)
+      return nullptr;
+
+    // We get size of single uc but we create one buffer for all uc's and split it
+    // uc needs buffer that is 32 Byte aligned
+    constexpr std::size_t alignment = 32;
+    // round up size to be 32 Byte aligned
+    size_t uc_aligned_size = (uc_log_buf_size + alignment - 1) & ~(alignment - 1);
+    try {
+      return std::make_unique<uc_log_buffer>(std::move(device), ctx_hdl, uc_aligned_size);
```
stsoe: uc_log_buffer no longer contains a device, e.g. it doesn't have a ctor that takes rvalue device, so pass just the device reference here.

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -65,13 +166,39 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
     }
   }
 
+  // Initializes uc log buffer, configures it by splitting the buffer
+  // equally among available columns
+  static std::unique_ptr<uc_log_buffer>
+  init_uc_log_buf(std::shared_ptr<xrt_core::device> device, xrt_core::hwctx_handle* ctx_hdl)
```
stsoe:
```suggestion
  init_uc_log_buf(const std::shared_ptr<xrt_core::device>& device, xrt_core::hwctx_handle* ctx_hdl)
```

stsoe (approved):

chvamshi-xilinx:
Thanks a lot @stsoe  for making the final changes. This feature is very important for lot of debugs.

PR #9168, opened 2025-08-18
[XRT-SMI]Interface buffer addition for firmware debuggability reports

aktondak:
This PR adds the buffer required for log streaming between Drivers and XRT-SMI. 

None

The firmware_debug_buffer will be used as buffer type for payload streaming to xrt-smi. Both XDNA and MCDM will use this buffer to fill up trace events and firmware logs. 

None

Will be tested once firmware events are logged onto the buffer.

NA

### DIFF
--- a/src/runtime_src/core/common/query_requests.h
+++ b/src/runtime_src/core/common/query_requests.h
@@ -4112,14 +4112,11 @@ struct event_trace : request
 
   static const key_type key = key_type::event_trace;
 
-  std::any
-  get(const device*) const override = 0;
-
-  // Parameterized get method based on key_type
+  // Parameterized get method based on key_type passed via std::any
   // Returns event_trace_version if key is event_trace_version
   // Returns firmware_debug_buffer if key is event_trace
-  virtual std::any
-  get(const device*, const key_type&) const = 0;
+  std::any
+  get(const device*, const std::any& key) const override = 0;
 
   void
   put(const device*, const std::any&) const override = 0;
@@ -4143,14 +4140,11 @@ struct firmware_log : request
 
   static const key_type key = key_type::firmware_log;
 
-  std::any
-  get(const device*) const override = 0;
-
-  // Parameterized get method based on key_type
+  // Parameterized get method based on key_type passed via std::any
   // Returns firmware_log_version if key is firmware_log_version
   // Returns firmware_debug_buffer if key is firmware_log
-  virtual std::any
-  get(const device*, const key_type&) const = 0;
+  std::any
+  get(const device*, const std::any& key) const override = 0;
 
   void
   put(const device*, const std::any&) const override = 0;
--- a/src/runtime_src/core/common/query_requests.h
+++ b/src/runtime_src/core/common/query_requests.h
@@ -324,7 +324,9 @@ enum class key_type
   performance_mode,
   preemption,
   event_trace,
+  event_trace_version,
   firmware_log,
+  firmware_log_version,
   frame_boundary_preemption,
   debug_ip_layout_path,
   debug_ip_layout,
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/query_requests.h
```diff
@@ -4077,10 +4077,29 @@ struct preemption : request
...
 
+/**
+ * This structure provides a common interface for accessing firmware debug data,
+ * supporting both polling and streaming modes of operation.
+ * 
+ * Fields:
+ * - buffer_offset: Absolute index of last-read event. Updated by driver to new offset 
+ *                  till the point where data is read from buffer.
+ * - data: Pointer to data buffer to be filled by driver with firmware log/trace data
+ * - size: Size in bytes of the data buffer filled by driver
```
maxzhen: We probably want to document it clearly about the meaning when it used as input and output. When it is input, it is the total size of the buffer. When it is output, it is the total bytes filled by driver.

src/runtime_src/core/common/query_requests.h
```diff
@@ -4077,10 +4077,29 @@ struct preemption : request
 
 };
 
+/**
+ * This structure provides a common interface for accessing firmware debug data,
+ * supporting both polling and streaming modes of operation.
+ * 
+ * Fields:
+ * - buffer_offset: Absolute index of last-read event. Updated by driver to new offset
```
maxzhen: It should be the offset to begin loading the next chunk of data, not the last offset read stopped. Thus, 0 can be used to start the very first read. The driver will also update the offset, on return, to the offset to start reading next. Thus, xrt-smi can echo it back in next read.
It should be in bytes.

src/runtime_src/core/common/query_requests.h
```diff
@@ -4077,10 +4077,29 @@ struct preemption : request
...
+/**
+ * This structure provides a common interface for accessing firmware debug data,
+ * supporting both polling and streaming modes of operation.
+ * 
+ * Fields:
+ * - buffer_offset: Absolute index of last-read event. Updated by driver to new offset 
+ *                  till the point where data is read from buffer.
+ * - data: Pointer to data buffer to be filled by driver with firmware log/trace data
+ * - size: Size in bytes of the data buffer filled by driver
+ * - b_wait: Directive for driver whether to wait for new events or return immediately.
```
maxzhen: "Directive for driver whether to wait for new events or return immediately" -> Directive for driver whether to wait for new payload data or return immediately when there is nothing to read, not a single byte.

src/runtime_src/core/common/query_requests.h
```diff
@@ -4077,10 +4077,29 @@ struct preemption : request
...
+ *           True = wait for new events. Always false for non-watch mode operations.
```
maxzhen: No need to say "Always false for non-watch mode operations.". It is not part of the API definition.

src/runtime_src/core/common/query_requests.h
```diff
@@ -4077,10 +4077,29 @@ struct preemption : request
...
+ */
+struct firmware_log_buffer {
+    uint64_t buffer_offset;
```
maxzhen: It is not "buffer_offset", which sounds like offset in a buffer. It is the absolute offset for the whole log or event trace.

src/runtime_src/core/common/query_requests.h
```diff
@@ -4077,10 +4077,29 @@ struct preemption : request
...
+ * 
+ * Fields:
+ * - buffer_offset: Absolute index of last-read event. Updated by driver to new offset 
+ *                  till the point where data is read from buffer.
+ * - data: Pointer to data buffer to be filled by driver with firmware log/trace data
+ * - size: Size in bytes of the data buffer filled by driver
+ * - b_wait: Directive for driver whether to wait for new events or return immediately.
+ *           True = wait for new events. Always false for non-watch mode operations.
+ */
+struct firmware_log_buffer {
```
maxzhen: This is for both fw log and event tracing, so let's use a generic name.

aktondak:
> This data structure is common for both event tracing and fw logging. So, don't just mention "event" in your comment. Maybe say "payload".

This has been renamed.

aktondak:
> We also need to fetch the major and minor version from driver. Do you have that query defined?

This wasn't captured in the xrt-smi page. I'll add that and plan it with a separate PR

maxzhen (approved):

larry9523 (approved):

PR #9169, opened 2025-08-19
VITIS-14639: get thermal info and set thermal temp threshold

bisingha-xilinx:
https://jira.xilinx.com/browse/VITIS-14639
Its a new feature
It was solved by having two new APIs for users:
1. xrt::aie::device::get_thermal() - to get thermal information like temperature.
2. xrt::aie::device::set_thermal_threshold() - to set temperature threshold. Currently there is no sysfs entry/support from the thermal driver, so the API is implemented but functionality.

Please note:
1. added a enum class thermal in xrt::aie::device to scale it later if there comes any other requirement to read other thermal informations like type, power, voltage, etc.
2. get_thermal() function implemented in the device_linux.
n/a
Tested by enabling the thermal driver for versal and tested below:
1. reading the thermal -> successful
3. reading form invalid zone -> throwing error (expected)
4. setting thermal temperature threshold -> throwing error (expected since its not supported yet)
need to add this two new APIs for user.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_device.cpp
+++ b/src/runtime_src/core/common/api/xrt_device.cpp
@@ -532,12 +532,12 @@ write_aie_reg(pid_t pid, uint16_t context_id, uint16_t col, uint16_t row, uint32
 
 float
 device::
-get_thermal(thermal arg) const
+get_thermal(thermal_type thermal) const
 {
   return xdp::native::profiling_wrapper("xrt::device::get_thermal",
-  [this, arg] {
+  [this, thermal] {
     try {
-      return xrt_core::device_query<xrt_core::query::aie_thermal>(get_handle().get(), arg);
+      return xrt_core::device_query<xrt_core::query::aie_thermal>(get_handle().get(), thermal);
     }
     catch (const xrt_core::query::no_such_key&) {
       throw std::runtime_error("get_thermal is not supported on this platform");
@@ -547,12 +547,12 @@ get_thermal(thermal arg) const
 
 void
 device::
-set_thermal_threshold(thermal arg, uint32_t value) const
+set_thermal_threshold(thermal_type thermal, uint32_t value) const
 {
   return xdp::native::profiling_wrapper("xrt::device::set_thermal_threshold",
-  [this, arg, value] {
+  [this, thermal, value] {
     try {
-      xrt_core::device_update<xrt_core::query::aie_thermal>(get_handle().get(), arg, value);
+      xrt_core::device_update<xrt_core::query::aie_thermal>(get_handle().get(), thermal, value);
     }
     catch (const xrt_core::query::no_such_key&) {
       throw std::runtime_error("set_thermal_threshold is not supported on this platform");
--- a/src/runtime_src/core/edge/user/device_linux.cpp
+++ b/src/runtime_src/core/edge/user/device_linux.cpp
@@ -683,12 +683,12 @@ set_thermal_threshold_temperature(uint32_t val)
 struct aie_thermal
 {
   using result_type = query::aie_thermal::result_type;
-  using thermal_type = xrt::aie::device::thermal;
+  using thermal_type = xrt::aie::device::thermal_type;
 
   static result_type
-  get(const xrt_core::device* device, key_type key, const std::any& arg)
+  get(const xrt_core::device* device, key_type key, const std::any& thermal)
   {
-    auto thermal_t = std::any_cast<thermal_type>(arg);
+    auto thermal_t = std::any_cast<thermal_type>(thermal);
 
     switch(thermal_t) {
       case thermal_type::temperature:
@@ -699,9 +699,9 @@ struct aie_thermal
   }
 
   static void
-  put(const xrt_core::device* device, key_type key, const std::any& arg, const std::any& value)
+  put(const xrt_core::device* device, key_type key, const std::any& thermal, const std::any& value)
   {
-    auto thermal_t = std::any_cast<thermal_type>(arg);
+    auto thermal_t = std::any_cast<thermal_type>(thermal);
     auto val = std::any_cast<uint32_t>(value);
 
     switch(thermal_t) {
--- a/src/runtime_src/core/include/xrt/xrt_aie.h
+++ b/src/runtime_src/core/include/xrt/xrt_aie.h
@@ -59,12 +59,14 @@ public:
   enum class buffer_state { idle, running, completed };
 
   /**
-   * @enum thermal - aie thermal type
+   * @enum thermal_type - enumerates the type of thermal parameters available for the AIE device
    * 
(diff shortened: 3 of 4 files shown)


chvamshi-xilinx:
Need @stsoe approval as this PR contains a new interface to application

src/runtime_src/core/common/api/xrt_device.cpp
```diff
@@ -529,6 +529,36 @@ write_aie_reg(pid_t pid, uint16_t context_id, uint16_t col, uint16_t row, uint32
...
+    }
+    catch (const xrt_core::query::no_such_key&) {
+      throw std::runtime_error("get_thermal is not supported on this platform");
+    }
+  });
+}
+
+void
+device::
+set_thermal_threshold(const thermal& arg) const
```
chvamshi-xilinx: As discussed in the meeting, update the API to take uint32_t as threshold value

chvamshi-xilinx (changes requested):
.

src/runtime_src/core/common/api/xrt_device.cpp
bisingha-xilinx: Hi @chvamshi-xilinx updated it accordingly. Thank you for reviewing.

stsoe (changes requested):
I am not keen on adding ishim setter and getter functions, which doesn't scale too well.  I realize it's the easiest route though, and it's internal so can always be changed.   Maybe the query infratructure could be used?   

Can we think of a way to not pollute  APIs with setters and getters, which is far worse compared to internal.  The xrt::device class has a get_info method that can return any informational typed value.  We lack a set counter part maybe, but that could be implemented.

bisingha-xilinx:
> I am not keen on adding ishim setter and getter functions, which doesn't scale too well. I realize it's the easiest route though, and it's internal so can always be changed. Maybe the query infratructure could be used?
> 
> Can we think of a way to not pollute APIs with setters and getters, which is far worse compared to internal. The xrt::device class has a get_info method that can return any informational typed value. We lack a set counter part maybe, but that could be implemented.
> 
> Don't worry too much about this comment. I am not sure what is right here. Possible that the public API changes is what we want.

Hi @stsoe, thank for reviewing. Updated the pr by using query infrastructure instead of ishim setter/getter. Please review and let me know your comments.

Thanks.

chvamshi-xilinx (approved):

src/runtime_src/core/common/query_requests.h
```diff
@@ -295,6 +295,8 @@ enum class key_type
   xclbin_slots,
   aie_get_freq,
   aie_set_freq,
+  aie_get_thermal,
+  aie_set_thermal,
```
stsoe: Rename to aie_thermal, and use struct aie_thermal with `get` and `put`.   Check some of the other structs that use both.  I notice same pattern for aie_get/set_freq,

src/runtime_src/core/edge/user/device_linux.cpp
```diff
@@ -657,6 +657,71 @@ struct aie_set_freq
...
+    throw xrt_core::error(-ENODEV, "Failed to open file: " + path);
+
+  int millidc;
+  file >> millidc;
+  file.close();
+
+  return millidc / 1000.0f;
+}
+
+struct aie_get_thermal
```
stsoe: Rename to aie_thermal and have both get() and put().

src/runtime_src/core/include/xrt/xrt_aie.h
```diff
@@ -185,6 +193,34 @@ class device : public xrt::device
...
+
+  /**
+   * set_thermal_threshold() - set AIE thermal threshold based on arg
+   *
+   * @param arg
+   *  thermal enum option
+   * @param value
+   *  threshold value to set
+   *
+   * This function sets thermal threshold eg: temparature threshold of AIE
```
stsoe: What does it mean to set the thermal threshold?  What behavior does it change?

src/runtime_src/core/include/xrt/xrt_aie.h
```diff
@@ -185,6 +193,34 @@ class device : public xrt::device
...
+   * @param arg
+   *  thermal enum option
+   * @return
+   *  returns aie thermal information based on argument
+   *
+   * This function reads thermal information eg: temparature of AIE
+   */
+  XCL_DRIVER_DLLESPEC
+  float
+  get_thermal(const thermal& arg) const;
```
stsoe:
```suggestion
  get_thermal(thermal) const;
```

src/runtime_src/core/include/xrt/xrt_aie.h
```diff
@@ -58,6 +58,14 @@ class device : public xrt::device
    */
   enum class buffer_state { idle, running, completed };
 
+  /**
+   * @enum thermal - aie thermal type
+   * 
+   * @var temperature
+   *   Thermal zone temperature
```
stsoe: Maybe expand a little and explain what thermal zone temperature is.

src/runtime_src/core/include/xrt/xrt_aie.h
```diff
@@ -58,6 +58,14 @@ class device : public xrt::device
...
+   */
+  enum class thermal { temperature };
```
stsoe: I would probably use `thermal_type` here, or _category, or something better.   I have a comment about `arg` later which you could then name `thermal`.

src/runtime_src/core/common/api/xrt_device.cpp
```diff
@@ -529,6 +529,36 @@ write_aie_reg(pid_t pid, uint16_t context_id, uint16_t col, uint16_t row, uint32
       }
     });
 }
+
+float
+device::
+get_thermal(const thermal& arg) const
```
stsoe:
```suggestion
get_thermal(thermal arg) const
```

src/runtime_src/core/common/api/xrt_device.cpp
```diff
@@ -529,6 +529,36 @@ write_aie_reg(pid_t pid, uint16_t context_id, uint16_t col, uint16_t row, uint32
...
+    }
+    catch (const xrt_core::query::no_such_key&) {
+      throw std::runtime_error("get_thermal is not supported on this platform");
+    }
+  });
+}
+
+void
+device::
+set_thermal_threshold(const thermal& arg, uint32_t value) const
```
stsoe:
```suggestion
set_thermal_threshold(thermal arg, uint32_t value) const
```

bisingha-xilinx:
Hi @stsoe thank you for reviewing and for the suggestions. I've updated the pr accordingly. Please review and let me know your comments.
Thanks.

stsoe (approved):
Thanks.  I didn't  realize the getput infra was missing for what you needed to do.

PR #9175, opened 2025-08-20
Add ctrlpkt buffer pool in xrt::kernel

rbramand-xilinx:
Added buffer pool that caches pre created ctrlpkt buffers. This is done to avoid the overhead of creating them in xrt::run object.
In some test cases xrt::run object creation comes in critical path, so we need to avoid creating buffers with unpinned memory.

 During xrt::run object creation we create xrt::bo's for control code, control packet etc. Control code/Instruction buffer uses pinned memory so allocation overhead is minimal whereas with control packet its not the case. This PR adds changes to maintain buffer pool that can be used for control packet at xrt::kernel level.
Available memory for buffer pool is taken from ini option.
Cache size is limited and when all the cached buffers are in use we create buffer on the fly and use it.
The buffer cache is designed in a way to hold up only the memory that is specified with ini option.

**NOTE :** This PR only considers ctrlpkt buffer only, the next step is to design the solution that can tackle all the buffers used in module_sram . It needs lot of changes in XRT, so can be done with next set of PRs.

Low to moderate

Tested using expoi test case in model-test repo on strix linux - the test case has ctrlpkt data and is multi threaded.

Also collected data with different sized ctrlpkt ELFs on strix Linux and the performance is improved. Below is the comparison with and without these changes 

image

TODO : Get numbers on Windows setup

NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -1278,23 +1278,21 @@ public:
   { return arg.type; }
 };
 
-} // namespace
-
-namespace xrt {
 // class buffer_cache - A thread safe cache to hold pre created buffers
+//
 // This class maintains a fixed size pool of reusable buffers to
-// minimize allocation overhead.
-// get_buffer() returns a buffer from the cache if available, or creates one on demand.
-// release_buffer() returns a buffer to the cache only if the pool size limit is not exceeded
-
-
+// minimize allocation overhead.  get_buffer() returns a buffer from
+// the cache if available, or creates one on demand.  release_buffer()
+// returns a buffer to the cache only if the pool size limit is not
+// exceeded
 class buffer_cache
 {
-  std::vector<xrt::bo> m_bo_cache;
-  std::mutex m_mutex;
   xrt::hw_context m_hw_ctx;
   std::vector<uint8_t> m_buf_data;
   const std::size_t m_max_cache_size;
+
+  mutable std::mutex m_mutex;
+  mutable std::vector<xrt::bo> m_bo_cache;
 
   static void
   fill_bo_with_data(xrt::bo& bo, const std::vector<uint8_t>& data)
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1282,6 +1282,71 @@ class argument
...
+  }
+
+public:
+  buffer_cache(xrt::hw_context ctx, std::vector<uint8_t> data, size_t max_size)
+    : m_hw_ctx(std::move(ctx))
+    , m_buf_data(std::move(data))
+    , m_max_cache_size(max_size)
+  {
+    // Pre allocate the cache size
+    if (m_max_cache_size > 0) {
```
sonals: Can you do an early exit here to remove one control flow indentation?

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1842,6 +1945,24 @@ class kernel_impl : public std::enable_shared_from_this<kernel_impl>
   {
     return regmap_size;
   }
+
+  bool
+  is_ctrlpkt_cache_enabled()
```
sonals: const

sonals (changes requested):
Minor changes suggested, looks good otherwise.

sonals (approved):

sayyanna:
Data from a Strix windows machine, with and without the optimization
image

larry9523 (commented):
1. we don't have to deal with it now, but will we disable the ctlpkt cache in full elf flow?
2. if the actuall ctrlpkt size is larger than the configured, should we create just 1 ctrlpkt cache instead of none? Even with 1 ctrlpkt cache, we can still reduce the xrt::run contructor overhead, right?

rbramand-xilinx:
> 1. we don't have to deal with it now, but will we disable the ctlpkt cache in full elf flow?
> 2. if the actuall ctrlpkt size is larger than the configured, should we create just 1 ctrlpkt cache instead of none? Even with 1 ctrlpkt cache, we can still reduce the xrt::run contructor overhead, right?

Hi Larry, as discussed offline this flow works for full ELF flow as well because we are using ctrl code id to get ctrlpkt data.
Also I have made changes to have minimum pool size as 1 (when ctrlpkt size > pool memory size)

larry9523 (approved):

chvamshi-xilinx (approved):

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1988,13 +2115,19 @@ class run_impl : public std::enable_shared_from_this<run_impl>
   // If module has multiple control codes, id is used to identify
   // the control code that needs to be run.
   // By default first control code that is available is picked
-  static xrt::module
+  xrt::module
```
stsoe: Must be static, it is used during initialization and cannot depend on the object being initialized.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1282,6 +1282,72 @@ class argument
 
 namespace xrt {
 
+class buffer_cache
```
stsoe, line 1288: What is this?  Please comment carefully so I know what I am looking at.  The PR is nicely commented, but code comments are very important for future code reading.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1618,6 +1718,12 @@ class kernel_impl : public std::enable_shared_from_this<kernel_impl>
 
     // amend args with computed data based on kernel protocol
     amend_args();
+
+    if (m_module) {
+      // ELF flow, initialize ctrlpkt buffer cache
+      initialize_ctrlpkt_bo_cache();
+    }
```
stsoe: Please change initialize_ctrlpkg_bo_cache() to be a static member function called from initializer list with necessary data members.  This enforces that the function is working with initialized data only and not accidentally uses unitialized data members.  The function should return std::unique_ptr<buffer_cache>, but must not modify or use the object being initialized.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1838,10 +1944,23 @@ class kernel_impl : public std::enable_shared_from_this<kernel_impl>
...
-  get_regmap_size()
+  get_regmap_size() const
   {
     return regmap_size;
   }
+
+  xrt::bo
+  get_ctrlpkt_buffer()
+  {
+    return m_ctrlpkt_bo_cache ? m_ctrlpkt_bo_cache->get_buffer() : xrt::bo{};
```
stsoe: Debugging
```suggestion
    return m_ctrlpkt_bo_cache 
      ? m_ctrlpkt_bo_cache->get_buffer() 
      : xrt::bo{};
```

src/runtime_src/core/common/api/xrt_kernel.cpp
rbramand-xilinx: sure Soren, will make changes accordingly

PR #9183, opened 2025-08-21
CR-1246941: Fixing trace issue

bisingha-xilinx:
https://jira.xilinx.com/browse/CR-1246941
When we moved to hw context flow for edge we did not cover all the trace/profiling usecases. And pr https://github.com/Xilinx/XRT/pull/8853 disables xdp device update for hw cotext flow, it should be enabled.
Problem was solved by enabling the xdp device update for hw context flow and by mapping CUs and debug IPs.
n/a
Tested the given testcase by enabling the aie trace in xrt.ini
n/a


src/runtime_src/core/edge/user/shim.cpp
```diff
@@ -1305,8 +1305,18 @@ int shim::load_hw_axlf(xclDeviceHandle handle, const xclBin *buffer, drm_zocl_cr
   bool checkDrmFD = xrt_core::config::get_enable_flat() ? false : true;
   ZYNQ::shim *drv = ZYNQ::shim::handleCheck(handle, checkDrmFD);
 
-  if (!hw_context_enable)
```
vipangul, line 1308: As we discussed offline, please restore update_device code. For hwContext flow, hooks come from xrt_hw_context.cpp.

bisingha-xilinx: Hi @vipangul thank you for the review. will update the PR.

PR #9185, opened 2025-08-21
[XDP] Enhancement to Bytes Count XRT Profile API to accurately measure clock cycles

vipangul:
Added user events

VITIS-14440 - Improvements to profile API to measure byte count accurately.

If to tal byte count is lesser than the total transfer on the graph.port, counter would not stop counting as soon as requested bytes transfer has completed.
 
To address this, similar to how it is being supported in xrt.ini , XRT profile API has also been improved to -
- Avail use of combo3 state machine to accurately measure the byte count.
- It also aligns implementation for Profile APIs and xrt.ini profiling

Low

Verified on AIE1 & HW_EMU designs

### DIFF
--- a/src/runtime_src/core/edge/user/aie/common_layer/adf_profiling_api.cpp
+++ b/src/runtime_src/core/edge/user/aie/common_layer/adf_profiling_api.cpp
@@ -159,14 +159,50 @@ err_code profiling::profile_stream_start_to_transfer_complete_cycles(XAie_DevIns
     uint8_t streamWidthInBytes = 4;
     if (dev->DevProp.DevGen == XAIE_DEV_GEN_AIE2PS)
         streamWidthInBytes = 8;
+    // Counter1 - Measures the total beats for the specified total bytes.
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/edge/user/aie/common_layer/adf_profiling_api.cpp
```diff
@@ -159,15 +159,51 @@ err_code profiling::profile_stream_start_to_transfer_complete_cycles(XAie_DevIns
...
+    std::vector<XAie_Events>         combo_events;
+    std::vector<XAie_EventComboOps>  combo_opts;
+    std::vector<XAie_Events>         comboConfigedEvents;
+    XAie_Events newStartEvent = XAIE_EVENT_NONE_CORE;
+
+    // Set up a combo event using start & count event type
+    comboEvent3 = fal_util::s_pXAieDev->tile(shimColumn, 0).pl().comboEvent(4);
+    int rcCombo = comboEvent3->reserve();
+    if (rcCombo != XAIE_OK)
+      std::cout << "ERROR: event::start_profiling: Failed to reserve combo event resources." << std::endl;
```
vipangul: Need to remove debug logs.

jvillarre (approved):

PR #9186, opened 2025-08-21
Remove xbmgmt bash completion from setup scripts

rchane:
Recent testing when sourcing setup script on Linux would show warning about missing xbmgmt script before running xrt-smi normally.
Recent cleanup of xbutil/xbmgmt mentions in code may have caused this, discovered in personal testing
Removed the relevant xbmgmt script sourcing from the setup script.
N/A
Tested on STX xdna, sourcing works normally again without the extra message.
N/A


rchane:
> can you remove the actual script as well as it is not being used anywhere else anymore?

I looked into it further and see that it is still being installed for Alveo. Let me see if I can update the setup scripts.

PR #9189, opened 2025-08-21
Event trace report and parse changes

aktondak:
This PR adds the implementation for Event Trace report and a parser for event_trace.yaml that is shipped by the firmware in tha latest stric firmware release.
The parser implementation is based on a python parses the firmware team uses for their debugging. 

https://jira.xilinx.com/browse/AIESW-5508

The problem was solved by using an open source yaml parsing library yaml-cpp for yaml parsing and using the parsed configuration to formalize the event trace data in a readable table.

What is missing in this : 

1. This implementation misses handling of missed events since actual firmware data is not used to test this.
2. This implementation also assumes the yaml configuration will be stored in driver build and loaded at run time through a valid path retrieved from UMD. 

None

This is not tested on actual data yet since the driver has not implemented the ioctl changes to retrieve actual firmware data. This PR has been tested using a dummy data generated in the shim and using it to print a table from parsed configuration.
Dummy data used :
```
      std::vector<trace_event> dummy_events = {
        // NPU Scheduling events
        {1000000, 0x0000, 0x1},  // FRAME_START with context_id=1
        {1001000, 0x0001, 0x1},  // FRAME_DONE with context_id=1
        {1002000, 0x0002, 0x12}, // PREEMPTION_START with context_id=1, op_id=2
        {1003000, 0x0003, 0x12}, // PREEMPTION_DONE with context_id=1, op_id=2
        
        // Mailbox events
        {1004000, 0x0004, 0x10}, // PROCESS_MGMT_MSG_START with MSG_OPCODE=0x10
        {1005000, 0x0005, 0x10}, // PROCESS_MGMT_MSG_DONE with MSG_OPCODE=0x10
        {1006000, 0x0006, 0x01000110}, // CREATE_CONTEXT with priority=0x1, cols=1, start=1, ctx=1
        {1007000, 0x0007, 0x10001},     // DELETE_CONTEXT with context_id=1, status=1
        
        // Clock/Power events
        {1008000, 0x000a, 0x101}, // GATE_AIE2_CLKS with num_cols=1, start_col=1
        {1009000, 0x000b, 0x101}, // UNGATE_AIE2_CLKS with num_cols=1, start_col=1
        
        // Error events
        {1010000, 0x0010, 0x101}, // ASYNC_ERROR with num_cols=1, start_col=1
        
        // PDI Load events
        {1011000, 0x0012, 0x1000A001}, // PDI_LOAD_START with context_id=1, address=0xA000
        {1012000, 0x0013, 0x1000A001}, // PDI_LOAD_DONE with context_id=1, address=0xA000
      };
```
Report generated : 
```
<img width="1247" height="476" alt="image" src="https://github.com/user-attachments/assets/b423d0ba-3917-451c-9337-93e309680b59" />

```

### DIFF
--- a/src/runtime_src/core/tools/xbutil2/EventTraceConfig.cpp
+++ b/src/runtime_src/core/tools/xbutil2/EventTraceConfig.cpp
@@ -1,0 +1,553 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+
+#include "EventTraceConfig.h"
+
+#include <algorithm>
+#include <fstream>
+#include <iomanip>
+#include <iostream>
+#include <regex>
+#include <set>
+#include <sstream>
+#include <yaml-cpp/yaml.h>
+
+#include "core/common/query_requests.h"
+#include "core/common/device.h"
+
+event_trace_config::
+event_trace_config(const std::string& yaml_file_path) 
+  : event_bits(16), payload_bits(48), 
+    file_major(0), file_minor(0), config_valid(false) {
+
+  load_from_yaml(yaml_file_path);
+}
+
+bool
+event_trace_config::
+load_from_yaml(const std::string& yaml_file_path) {
+  try {
+    YAML::Node config = load_yaml_file(yaml_file_path);
+    
+    parse_version(config);
+    parse_data_format(config);
+    parse_lookups(config);
+    parse_categories(config);
+    parse_arg_sets(config);
+    parse_events(config);
+    
+    config_valid = true;
+    return true;
+    
+  } catch (const YAML::Exception& e) {
+    std::cerr << "YAML parsing error: " << e.what() << std::endl;
+    return false;
+  } catch (const std::exception& e) {
+    std::cerr << "Configuration error: " << e.what() << std::endl;
+    return false;
+  } 
+}
+
+YAML::Node
+event_trace_config::
+load_yaml_file(const std::string& yaml_file_path) {
+  if (yaml_file_path.empty()) {
+    throw std::runtime_error("YAML file path cannot be empty");
+  }
+  
+  std::ifstream file(yaml_file_path);
+  if (!file.is_open()) {
+    throw std::runtime_error("Cannot open YAML file: " + yaml_file_path);
+  }
+  
+  return YAML::LoadFile(yaml_file_path);
+}
+
+void
+event_trace_config::
+parse_data_format(const YAML::Node& config) {
+  if (!config["data_format"]) {
+    throw std::runtime_error("Missing required 'data_format' section in YAML");
+  }
+  
+  const auto& data_format = config["data_format"];
+  
+  if (!data_format["event_bits"]) {
+    throw std::runtime_error("Missing 'event_bits' in data_format section");
+  }
+  if (!data_format["payload_bits"]) {
+    throw std::runtime_error("Missing 'payload_bits' in data_format section");
+  }
+  
+  event_bits = data_format["event_bits"].as<uint32_t>();
+  payload_bits = data_format["payload_bits"].as<uint32_t>();
+  
+  if (event_bits == 0 || payload_bits == 0) {
+    throw std::runtime_error("Event bits and payload bits must be greater than 0");
+  }
+}
+
+void
+event_trace_config::
+parse_version(const YAML::Node& config) {
+  // Version is optional, default to 0.0 if not present
+  if (config["version"]) {
+    const auto& version = config["version"];
+    
+    if (version["major"]) {
+      file_major = version["major"].as<uint16_t>();
+    }
+    if (version["minor"]) {
+      file_minor = version["minor"].as<uint16_t>();
+    }
+  }
+}
+void
+event_trace_config::
+parse_lookups(const YAML::Node& config) {
+  if (!config["lookups"]) {
+    // Lookups are optional
+    return;
+  }
+  
+  for (const auto& lookup_pair : config["lookups"]) {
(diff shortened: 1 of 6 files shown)


src/runtime_src/core/tools/xbutil2/EventTraceConfig.cpp
```diff
@@ -0,0 +1,460 @@
...
+#include <iomanip>
+#include <regex>
+#include <set>
+#include <sstream>
+
+#include "core/common/json/nlohmann/json.hpp"
+
+namespace {
+constexpr uint32_t EVENT_BITS_DEFAULT = 16;
+constexpr uint32_t PAYLOAD_BITS_DEFAULT = 48;
```
stsoe: Please no UPPERCASE for anything but macros.

src/runtime_src/core/tools/xbutil2/EventTraceConfig.cpp
```diff
@@ -0,0 +1,460 @@
...
+    config_valid = true;
+    return true;
+  } 
+  catch (const std::exception& e) {
+    throw std::runtime_error("JSON parsing error: " + std::string(e.what()));
+  }
+}
+
+nlohmann::json 
+event_trace_config::load_json_file(const std::string& json_file_path)
```
stsoe: This looks like a general function not specific to the class.  I would move to unnamed namespace static function.

src/runtime_src/core/tools/xbutil2/EventTraceConfig.cpp
```diff
@@ -0,0 +1,460 @@
...
+  try {
+    nlohmann::json config = load_json_file(json_file_path);
+    parse_version(config);
+    parse_data_format(config);
+    parse_lookups(config);
+    parse_categories(config);
+    parse_arg_sets(config);
+    parse_events(config);
+    config_valid = true;
+    return true;
```
stsoe: This parsing is specific to the class so they all belong as member functions as you have them, but I would make the static and have them return was is being parsed, e.g. 
```
static std::pair<uint16_t, uint16_t>
event_trace_config::
parse_version(...)
{
   ...
}
```
This way all data members can be initialized in contructor initializer list and the body can remain empty.  The first data member of the class would be the config json object obtained by calling static load_json_file, the parsing function would be parsed this json or sub-elements of this json.

src/runtime_src/core/tools/xbutil2/EventTraceConfig.cpp
```diff
@@ -0,0 +1,460 @@
...
+    name_check.insert(name);
+    category_info cat_info = create_category_info(category, forced_id_categories);
+    category_map[name] = cat_info;
+  }
+  assign_category_ids(forced_id_categories);
+}
+
+event_trace_config::category_info
+event_trace_config::
+create_category_info(const nlohmann::json& category, std::set<uint32_t>& forced_id_categories)
```
stsoe: I can't tell for sure, but could this be a static member function?  Does it modify data members, and if it does, should it?

I think this comment applies to a lot of the functions.  Let me know.

src/runtime_src/core/tools/xbutil2/EventTraceConfig.h
```diff
@@ -0,0 +1,339 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+
+#pragma once
```
stsoe: I am not positive pragma once works on Linux. For consistency I think headers should be using plain old include guards.

src/runtime_src/core/tools/xbutil2/EventTraceConfig.h
```diff
@@ -0,0 +1,339 @@
...
+#include <vector>
+#include "core/common/json/nlohmann/json.hpp"
+
+// Forward declaration for device
+namespace xrt_core { class device; }
+
+/**
+ * @brief Raw event record structure
+ */
+struct event_record
```
stsoe, line 21: I don't think you want this struct in the global namespace?  Probably none of the classes should be in global namespace.

src/runtime_src/core/tools/xbutil2/ReportEventTrace.h
```diff
@@ -0,0 +1,81 @@
...
+class ReportEventTrace : public Report {
+public:
+  /**
+   * @brief Constructor for event trace report
+   * 
+   * Initializes the report with:
+   * - Report name: "event-trace"
+   * - Description: "Log to console firmware event trace information"
+   */
+  ReportEventTrace() : Report("event-trace", "Log to console firmware event trace information", true /*deviceRequired*/) { /*empty*/ };
```
stsoe:
```suggestion
  ReportEventTrace() 
    : Report("event-trace", "Log to console firmware event trace information", true /*deviceRequired*/) 
    {};
```

src/runtime_src/core/tools/xbutil2/EventTraceConfig.cpp
aktondak: That makes sense to me. I have simplified this to keep everything restricted to initialization list.

src/runtime_src/core/tools/xbutil2/EventTraceConfig.h
aktondak: I agree. I've moved all not member changing APIs to be static members

src/runtime_src/core/tools/xbutil2/EventTraceConfig.cpp
aktondak: Fixed

maxzhen (approved):

PR #9190, opened 2025-08-22
backport chnages to support Alma Linux 9.6

nindanaoto:
#### Problem solved by the commit
Driver compilation issue on Alma Linux 9.6
#### How problem was solved, alternative solutions (if any) and why they were rejected
Backporting the master's changes. 
#### Risks (if any) associated the changes in the commit
Since these changes are coming from merely modifications introduced in the latest Linux kernel APIs, no risk is present. 
#### What has been tested and how, request additional testing if necessary
xbutil examine with Alveo U280 recognize the board on Alma Linux 9.6.

PR #9191, opened 2025-08-22
backport clangtidy workflow files

nindanaoto:
Old clangtidy workflows are not working now. 
I encountered this issue when I raised #9190. But I guess  that this commit itself is not the cause of the issue. 
Backported master's workflow files. Since this requires careful checking, I believe that doing the same thing (just copy and paste) at the maintainer's side might be a better solution. 
Any changes to the workflow file can potentially introduce malicious code into the workflow file. Although I can assure that these are just copies of the master's files, trusting me (a random person at a Japanese University) is a risk.
Verifying the diff against the master's workflow files and the test run of the workflow may be necessary on the maintainer's side.


stsoe (approved):
This LGTM.  I was under the impression that release branch like 2023.2 were basically frozen, or otherwise cherry-picked from master, hence no real need for clangtidy here.  However, I do not see a problem merging this backport.  Thanks @nindanaoto .

PR #9193, opened 2025-08-22
Relative Path Enforcement for Profiling Directory Plugin 

meghnakon:
Enforcing relative path for the profiling directory plugin. No absolute paths will be allowed to ensure Vitis Analyzer will work smoothly.

User specified paths must start with './' or be the name of a folder that can be created within the current directory. 

Fix has been tested on the VCK190 to ensure that output files will not be written to an invalid directory that the user specified.


meghnakon:
.

PR #9196, opened 2025-08-25
print ctx heath exception message in non-elf flow

sayyanna:
Exception message is not getting printed in case of ERT_CMD_STATE_TIMEOUT with TXN flow
Exception message is not getting printed in case of ERT_CMD_STATE_TIMEOUT with TXN flow
Added case for TXN flow.
None
Tested with TXN binary 
No

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -2571,6 +2571,7 @@ public:
     std::string msg = "Command failed to complete successfully (" + cmd_state_to_string(state) + ")";
     
     switch (epkt->opcode) {
+    case ERT_START_CU:
     case ERT_START_NPU:
     case ERT_START_NPU_PREEMPT:
     case ERT_START_NPU_PREEMPT_ELF:
@@ -3190,6 +3191,7 @@ class runlist_impl
   {
     auto epkt = run.get_ert_packet();
     switch (epkt->opcode) {
+    case ERT_START_CU:
     case ERT_START_NPU:
     case ERT_START_NPU_PREEMPT:
     case ERT_START_NPU_PREEMPT_ELF:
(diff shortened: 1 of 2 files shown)


stsoe (changes requested):
I don't think this change is right.  It will trigger for Alveo where AIE error makes no sense.  How come ert::opcode::ERT_START_CU is used with NPU?

larry9523:
> AIE error makes no sense.

There are still quite a few models that are not using ELF (and preemption). In non-elf flow, ERT_START_CU is the opcode.

We are trying to debug some timeout cases, while we see ctx health data in KMD log but xrt exception messages are missing.

stsoe:
> > AIE error makes no sense.
> 
> There are still quite a few models that are not using ELF (and preemption). In non-elf flow, ERT_START_CU is the opcode.
> 
> We are trying to debug some timeout cases, while we see ctx health data in KMD log but xrt exception messages are missing.

But ERT_START_CU should not be used for NPU.  We need to differentiate between CU (Alveo) and NPU.   I was under the impression that before ELF it was always ERT_START_NPU.

larry9523:
> > > AIE error makes no sense.
> > 
> > 
> > There are still quite a few models that are not using ELF (and preemption). In non-elf flow, ERT_START_CU is the opcode.
> > We are trying to debug some timeout cases, while we see ctx health data in KMD log but xrt exception messages are missing.
> 
> But ERT_START_CU should not be used for NPU. We need to differentiate between CU (Alveo) and NPU. I was under the impression that before ELF it was always ERT_START_NPU.

Unfortunately, ERT_START_CU was used for the traditional non-elf cases, where NPU kernel is no difference than any other kernel/softkernel on Alveo.

stsoe:
> > > > AIE error makes no sense.
> > > 
> > > 
> > > There are still quite a few models that are not using ELF (and preemption). In non-elf flow, ERT_START_CU is the opcode.
> > > We are trying to debug some timeout cases, while we see ctx health data in KMD log but xrt exception messages are missing.
> > 
> > 
> > But ERT_START_CU should not be used for NPU. We need to differentiate between CU (Alveo) and NPU. I was under the impression that before ELF it was always ERT_START_NPU.
> 
> Unfortunately, ERT_START_CU was used for the traditional non-elf cases, where NPU kernel is no difference than any other kernel/softkernel on Alveo.

Argh, we cannot change per this PR.   We need to have a solution, where ERT_START_CU is not in the switch statement.    I can take a look my tomorrow.

PR #9203, opened 2025-08-27
Adding support for rhel10.0

manikandan-xilinx:
Adding support for rhel10.0 to address the PRO-3320


manikandan-xilinx:
This change should not board test failures. It's a workflow change

PR #9205, opened 2025-08-30
XRT API to Update shared buffer during runtime

karthdmg-xilinx:
XRT API to Update shared buffer during runtime
NA
Implement XRT API to Update shared buffer during runtime
NA
1. Host Application that updates shared buffer calling the new update API.
2. 
NA

### DIFF
--- a/src/runtime_src/core/edge/user/aie/common_layer/adf_runtime_api.cpp
+++ b/src/runtime_src/core/edge/user/aie/common_layer/adf_runtime_api.cpp
@@ -321,6 +321,90 @@ err_code graph_api::end(unsigned long long cycleTimeout)
     return err_code::ok;
 }
 
+
+
+err_code checkSharedBufferConfigForUpdate(
+                        const shared_buffer_config* pSharedBufferConfig,
+                        const graph_config* pGraphConfig, size_t numBytes)
+{
+    if (!pSharedBufferConfig)
+        return errorMsg(err_code::internal_error,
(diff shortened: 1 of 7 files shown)


src/runtime_src/core/edge/user/aie/common_layer/adf_runtime_api.cpp
```diff
@@ -321,6 +321,89 @@ err_code graph_api::end(unsigned long long cycleTimeout)
     return err_code::ok;
 }
 
+
+
```
chvamshi-xilinx, line 325: We shouldn't use camelCase in our code, but as this file comes from Compiler team, we are ok.

src/runtime_src/core/edge/user/aie/graph_object.cpp
```diff
@@ -39,6 +39,9 @@ graph_object::graph_object(ZYNQ::shim* shim, const xrt::uuid& uuid , const char*
 
   /* Initialize graph rtp metadata */
   rtps = xrt_core::edge::aie::get_rtp(device.get(), graph_config.id, m_hwctx);
+  /* Initialize graph shared buffer config */
+  shared_buffer_configs = xrt_core::edge::aie::get_shared_buffers(device.get());
```
chvamshi-xilinx: As discussed offline, These shared_buffer_configs should be per graph_id and should support hw_context also.
Please make that change

chvamshi-xilinx (changes requested):
Please make requested changes

PR #9209, opened 2025-09-03
Update debian build files

shramov:
Debian build files were outdated and some fixes are needed to build packages

Extra dependencies were introduced in #8600
Rule file names changed in the commit 2fd9d27b1d3857311d2bde47b53dc5db8c5cffa7
I have not found exact commit that changed installation prefix

Alternative to arch-specific dependencies is to drop them - aie-rt is included as submodule, they are not needed.

Built packages with these fixes for ubuntu 20.04 22.04 and 24.04, tested on 20.04

### DIFF
--- a/build/debian/control
+++ b/build/debian/control
@@ -21,7 +21,7 @@ Build-Depends: cmake,
                libssl-dev,
                libudev-dev,
                libxml2-dev,
-	       libxaiengine,
+               libxaiengine [arm64],
                libyaml-dev,
                lsb-release,
                ocl-icd-dev,
@@ -32,7 +32,7 @@ Build-Depends: cmake,
                rapidjson-dev,
                systemtap-sdt-dev,
                uuid-dev,
-	       xaiefal,
+               xaiefal [arm64],
 Standards-Version: 4.5.0
 
 Package: xrt


build/debian/control
```diff
@@ -21,7 +21,7 @@ Build-Depends: cmake,
                libssl-dev,
                libudev-dev,
                libxml2-dev,
-	       libxaiengine,
+	       libxaiengine [arm64],
```
rbramand-xilinx: please use proper indentation

build/debian/control
```diff
@@ -32,7 +32,7 @@ Build-Depends: cmake,
                rapidjson-dev,
                systemtap-sdt-dev,
                uuid-dev,
-	       xaiefal,
+	       xaiefal [arm64],
```
rbramand-xilinx: ditto

build/debian/rules
```diff
@@ -8,7 +8,7 @@ override_dh_auto_configure:
 	if [ "$$(dpkg-architecture -q DEB_TARGET_ARCH)" = arm64 ]; then \
 	    XRT_NATIVE_BUILD=no DKMS_FLOW=yes CXXFLAGS="-DXRT_ENABLE_AIE -DFAL_LINUX=on" dh_auto_configure  -- -DXRT_AIE_BUILD=ON; \
```
rbramand-xilinx, line 9: I know in this PR we are not checking for arm64 builds, but as we dont use XRT_NATIVE_BUILD flag anymore can we use this PR to change this flag to XRT_EDGE

shramov: Maybe it's better to split this change into another PR and use aie-rt from submodule?

build/debian/control
shramov: Fixed. BTW indentation was inherited from previous commit fe1a86ad65449f01e769bcc36a47b768f7253990

build/debian/control
shramov: Same here

build/debian/rules
shramov: But I can not test it so if abybody can volounteer and test that change - it would be great

rbramand-xilinx: yeah may be @ManojTakasi can fix this using separate PR

shramov:
Added Signed-of-by to make DCO checker happy

shramov:
@chvamshi-xilinx @stsoe Can you take a minute and review this? It would be pity if this pull request will become forgotten :(

stsoe:
> @chvamshi-xilinx @stsoe Can you take a minute and review this? It would be pity if this pull request will become forgotten :(

Hi @chvamshi-xilinx.  I am fine with these changes if reviewers are okay.   Since CI is not testing build/debian anyway, please go ahead and merge.

chvamshi-xilinx:
Thanks Soren. Let me merge.

shramov:
> Since CI is not testing build/debian anyway

I can provide workflow file that uses debian/rules so they will not break eventualy if you think this makes sense

stsoe:
> > Since CI is not testing build/debian anyway
> 
> I can provide workflow file that uses debian/rules so they will not break eventualy if you think this makes sense

That would be great.  Thanks!

PR #9210, opened 2025-09-03
Adding support back to rhel 10

manikandan-xilinx:


manikandan-xilinx:
merging as it would not affect the hw tests

PR #9213, opened 2025-09-04
[XRT-SMI] Archive migration 

aktondak:
This PR adds the archive migration support for xrt-smi. This also adds the test-changes to run the latency test using archived artifacts. Going forward, I'll be migrating other xrt-smi tests as well based on this POC. 

https://jira.xilinx.com/browse/AIESW-10130

Solved by adding a new query to request correct archive based on platform and extracting correct archive artifacts at test-run time.

None

Tested with latency test on linux:
```
aktondak@xsjstrix45:/proj/rdi/staff/aktondak/xdna/xdna-driver/build$ xrt-smi validate -r latency 
WARNING: User doesn't have admin permissions to set performance mode. Running validate in Default mode
Validate Device           : [0000:c5:00.1]
    Platform              : NPU Strix
    Power Mode            : Default
Test 1 [0000:c5:00.1]     : latency 
    Details               : Average latency: 47.0 us                            
```

None

### DIFF
--- a/src/runtime_src/core/tools/common/tests/TestNPULatency.cpp
+++ b/src/runtime_src/core/tools/common/tests/TestNPULatency.cpp
@@ -32,23 +32,17 @@ TestNPULatency::run(std::shared_ptr<xrt_core::device> dev, const xrt_core::archi
 {
   boost::property_tree::ptree ptree = get_test_header();
   
-  // If no archive provided, fall back to standard implementation
   if (archive == nullptr) {
     XBValidateUtils::logger(ptree, "Info", "No archive provided, using standard latency test");
     return ptree;
   }
   
   try {
-    // Extract recipe and profile as strings from archive (flattened names in ar archive)
     std::string recipe_data = archive->data("recipe_latency.json");
     std::string profile_data = archive->data("profile_latency.json"); 
     
-    // Extract binary artifacts and build artifacts repository
     xrt_core::runner::artifacts_repository artifacts_repo;
     
-    // Based on the recipe structure, extract the specific artifacts referenced:
-    // - xclbin from header.xclbin: "validate_latency_strx.xclbin"
-    // - ctrlcode (elf files) from kernels[].ctrlcode: "nop_latency_strx.elf"
     std::vector<std::string> artifact_names = {
       "validate.xclbin", 
       "nop.elf" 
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/tools/common/tests/TestNPULatency.cpp
```diff
@@ -23,24 +24,60 @@ boost::property_tree::ptree
...
+    };
+    
+    // Extract available artifacts from archive into repository
+    for (const auto& artifact_name : artifact_names) {
+      try {
+        std::string artifact_data = archive->data(artifact_name);
+        
+        // Convert string to vector<char> for artifacts_repository
+        std::vector<char> artifact_binary(artifact_data.begin(), artifact_data.end());
+        artifacts_repo[artifact_name] = std::move(artifact_binary);
```
stsoe, line 58: Ah, I see this is not cool.  Unfortunately ARIO only supports return std::string() for data but the artifacts_repo uses std::vector, so there is one too many copy here.  Using std::string in the repo doesn't quite make sense IMO, so I am not sure what to do?

aktondak: I'm not sure either since I'm restricted by 2 higher level APIs (runner and ARIO) whose signature's cannot be changed. Maybe memcpy can be a faster approach? Might be overkill?
This is not into the critical path of benchmarking NPU latency since this is running on CPU so maybe we can afford this penalty.
How does reading artifacts from a file not encounter this?

stsoe: Reading from file is indeed doing a copy in much the same way.  The difference here though is that we already, through ARIO, have the data in memory, then copying to a new memory location.   Lets punt on this for now; as you say, it is not in the critical path.

src/runtime_src/core/tools/common/tests/TestNPULatency.cpp
```diff
@@ -23,24 +24,54 @@ boost::property_tree::ptree
...
+    
+    xrt_core::runner::artifacts_repository artifacts_repo;
+    
+    std::vector<std::string> artifact_names = {
+      "validate.xclbin", 
+      "nop.elf" 
+    };
+    
+    // Extract available artifacts from archive into repository
+    for (const auto& artifact_name : artifact_names) {
```
stsoe, line 52: Maybe do this:

```suggestion 
    // Extract available artifacts from archive into repository
    for (auto artifact_name : {"validate.xclbin", "nop.elf"}) {
```

aktondak: I plan to change to this a bit in coming PRs when I migrate other tests too. This code will be common to all tests so I'll create a common API

PR #9214, opened 2025-09-04
Changing the building of XDP modules to shared libraries for versioning

jvillarre:
The XDP plugin modules are currently built as modules, which do not incorporate the XRT major version in their name.  CMake explicitly does not add versioning to modules, but it is a necessary aspect of the XRT code base.  This changes the modules to shared libraries with the major number as part of the name and changes the loading to recognize the new name.

Additionally, this pull request adds the capability to load a few plugins from an SDK location separate from the XRT install on Client Windows machines.

Low risk as the implementation remains the same, just the packaging and naming of the built .so files.

The loading of the newly created .so files has been tested on Edge and the loading of .dlls from the SDK install has been tested on Client Windows.

No documentation impact.

### DIFF
--- a/src/runtime_src/core/common/module_loader.cpp
+++ b/src/runtime_src/core/common/module_loader.cpp
@@ -173,15 +173,26 @@ module_path(const std::string& module)
   path /= module + ".dll";
 #else
   path /= XRT_LIB_DIR;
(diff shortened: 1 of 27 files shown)


IshitaGhosh (approved):
Changes look good

PR #9215, opened 2025-09-05
[XRT-SMI] Context health report update

aktondak:
We decided to also show context id in the context health report. This change alters the result type for context_health structure and adds the context id to field to the report

https://jira.xilinx.com/browse/AIESW-10899

Added the field. Will update drivers to fill the ctx_id field appropriately

N/A

Tested on linux : 
image

NA

### DIFF
--- a/src/runtime_src/core/common/query_requests.h
+++ b/src/runtime_src/core/common/query_requests.h
@@ -2001,7 +2001,11 @@ struct aie_partition_info : request
 //   - std::vector<std::pair<uint32_t, uint32_t>>: Returns contexts matching (context_id, pid) pairs (Linux)
 struct context_health_info : request
 {
-  using result_type = std::vector<ert_ctx_health_data>;
+  struct smi_context_health {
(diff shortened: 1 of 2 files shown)


NishadSaraf (commented):
For Linux, Context is unique to a process. It's better to add a PID along with the context ID.

maxzhen (approved):

maxzhen:
> For Linux, context ID is unique to a process. It's better to add a PID along with the context ID.

Or you can group ctxs by pid

aktondak:
> > For Linux, context ID is unique to a process. It's better to add a PID along with the context ID.
> 
> Or you can group ctxs by pid

That's a good idea. I have updated the logic to make a PID sprific table. Here is how the print looks like with 2 instanced of xrt-smi validate running, each spawning multiple contexts :
image

NishadSaraf (approved):

stsoe (approved):

PR #9230, opened 2025-09-12
Fix minor bug in log buffer dumping

rbramand-xilinx:
Fixed a minor bug where shim hwctx handle is destroyed before log buffer is destroyed, this happens because we call explicit destruction of shim hwctx handle in hw_context_impl destructor.
Also creating host only buffer instead of cacheable buffer for all these special buffers.

It was caught when testing the feature on aie4

Calling destruction of log buffer before shim hw ctx handle destruction in hw_context_impl class

Low

Tested the feature on aie4 on both Linux and windows and things work as expected

NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -1702,7 +1702,7 @@ static xrtBufferFlags
 compose_internal_bo_flags(use_type type)
 {
   xcl_bo_flags flags {0};  // see xrt_mem.h
-  flags.flags = XRT_BO_FLAGS_CACHEABLE;
+  flags.flags = XRT_BO_FLAGS_HOST_ONLY;
   flags.access = XRT_BO_ACCESS_LOCAL;
   flags.dir = XRT_BO_ACCESS_READ_WRITE;
   flags.use = static_cast<uint32_t>(type);
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -1702,7 +1702,14 @@ static xrtBufferFlags
 compose_internal_bo_flags(use_type type)
 {
   xcl_bo_flags flags {0};  // see xrt_mem.h
-  flags.flags = XRT_BO_FLAGS_HOST_ONLY;
+
+  if (type == use_type::debug) {
+    // client use case, create buffer in sram
+    flags.flags = XRT_BO_FLAGS_CACHEABLE;
+  }
+  else
+    flags.flags = XRT_BO_FLAGS_HOST_ONLY;
+
   flags.access = XRT_BO_ACCESS_LOCAL;
   flags.dir = XRT_BO_ACCESS_READ_WRITE;
   flags.use = static_cast<uint32_t>(type);
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -1703,12 +1703,22 @@ compose_internal_bo_flags(use_type type)
 {
   xcl_bo_flags flags {0};  // see xrt_mem.h
 
-  if (type == use_type::debug) {
+  // This function is used to create internal buffers
+  // for debug/trace/log use cases.
+  // Sanity check the use_type
+  switch (type) {
+  case use_type::debug :
     // client use case, create buffer in sram
     flags.flags = XRT_BO_FLAGS_CACHEABLE;
-  }
-  else
+    break;
+  case use_type::dtrace :
+  case use_type::uc_debug :
+  case use_type::log :
     flags.flags = XRT_BO_FLAGS_HOST_ONLY;
+    break;
+  default:
+    throw std::runtime_error("create_bo is called with invalid buffer type\n");
+  }
 
   flags.access = XRT_BO_ACCESS_LOCAL;
   flags.dir = XRT_BO_ACCESS_READ_WRITE;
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/common/api/xrt_bo.cpp
```diff
@@ -1702,7 +1702,7 @@ static xrtBufferFlags
 compose_internal_bo_flags(use_type type)
 {
   xcl_bo_flags flags {0};  // see xrt_mem.h
-  flags.flags = XRT_BO_FLAGS_CACHEABLE;
+  flags.flags = XRT_BO_FLAGS_HOST_ONLY;
```
stsoe: Are you sure this is correct?   I was under the impression that for example XRT_BO_USE_DEBUG required to be allocated in SRAM, which is controlled by CACHEABLE?

rbramand-xilinx: Yeah if strix needs it we need to create cacheable buffer but I checked with Vamshi and he confirmed that strix doesnt use this flow for debug buffer and aie2ps, aie4 can all work well with host only buffer.

chvamshi-xilinx: Yes Soren. This is correct. in newer class of devices, UC is logging this information which has full access to DDR.
In Strix, that isn't the case.
I am 100% sure, but I will let @larry9523 also to confirm :-).

stsoe: But isn't this function used on older class devices also, aka Strix?   How are debug buffers created on Strix, if not by this function?

chvamshi-xilinx: I am not sure @stsoe , 
@jvillarre , Can you please check and confirm.

chvamshi-xilinx: @stsoe / @jvillarre ,
Checked the code just now. Looks like Strix is also using same function with type::debug.
@rbramand-xilinx  is going to add a special handling for Strix here.

rbramand-xilinx: I have made changes accordingly, please review

src/runtime_src/core/common/api/xrt_bo.cpp
```diff
@@ -1702,7 +1702,14 @@ static xrtBufferFlags
...
 {
   xcl_bo_flags flags {0};  // see xrt_mem.h
-  flags.flags = XRT_BO_FLAGS_CACHEABLE;
+
+  if (type == use_type::debug) {
+    // client use case, create buffer in sram
+    flags.flags = XRT_BO_FLAGS_CACHEABLE;
+  }
+  else
+    flags.flags = XRT_BO_FLAGS_HOST_ONLY;
```
stsoe:
```suggestion
  if (type == use_type::debug)
    // client use case, create buffer in sram
    flags.flags = XRT_BO_FLAGS_CACHEABLE;
  else
    flags.flags = XRT_BO_FLAGS_HOST_ONLY;
```

stsoe (commented):
I am not comfortable with these non-obvious changes.   Why is ml_timeline.cpp switched to uc_debug?  How does this work on Strix?   

Also, if aie4, then, even use_type::debug should be host, shouldn't it?

larry9523:
> I am not comfortable with these non-obvious changes. Why is ml_timeline.cpp switched to uc_debug? How does this work on Strix?
> 
> Also, if aie4, then, even use_type::debug should be host, shouldn't it?

AIE4 should use type::uc_debug and config the buffer. The AIE4 specific debug buf implementation has not been checked in yet. @IshitaGhosh and @jvillarre

chvamshi-xilinx:
> I am not comfortable with these non-obvious changes. Why is ml_timeline.cpp switched to uc_debug? How does this work on Strix?
> 
> Also, if aie4, then, even use_type::debug should be host, shouldn't it?

Hi @stsoe ,
ml_timeline plugin is different for each type of device. This change is in ve2(telluride) ml_timeline.cpp. Telluride and AIE4 devices need to use uc_debug as uc has direct access to host buffer.

stsoe:
> > I am not comfortable with these non-obvious changes. Why is ml_timeline.cpp switched to uc_debug? How does this work on Strix?
> > Also, if aie4, then, even use_type::debug should be host, shouldn't it?
> 
> Hi @stsoe , ml_timeline plugin is different for each type of device. This change is in ve2(telluride) ml_timeline.cpp. Telluride and AIE4 devices need to use uc_debug as uc has direct access to host buffer.

Why do we need both use_type::debug and use_type::uc_debug?  So the burden is on the caller to know what platform is used.  It doesn't make sense.

chvamshi-xilinx:
> > > I am not comfortable with these non-obvious changes. Why is ml_timeline.cpp switched to uc_debug? How does this work on Strix?
> > > Also, if aie4, then, even use_type::debug should be host, shouldn't it?
> > 
> > 
> > Hi @stsoe , ml_timeline plugin is different for each type of device. This change is in ve2(telluride) ml_timeline.cpp. Telluride and AIE4 devices need to use uc_debug as uc has direct access to host buffer.
> 
> Why do we need both use_type::debug and use_type::uc_debug? So the burden is on the caller to know what platform is used. It doesn't make sense.

We thought of providing only one internal API to create debug buffer/log buffer/trace buffer etc, XDP has separate plugins for different platforms, they know which platform they are working in their plugins, but XRT coreutil doesn't know. Thatswhy, provided two options (one for older platforms and another one for newer platforms) sothat we can differentiate to create cacheable vs host only buffers.

rbramand-xilinx:
@stsoe as discussed offline I have added switch case that does sanity as well. Please review

stsoe (approved):

PR #9231, opened 2025-09-14
Add minimal version info header to help compiler cache

shramov:
xrt/detail/abi.h includes autogenerated version.h that contains timestamp but use only major/minor version macros. Using smaller header enables use of compiler caches (like ccache).

Make it possible to use ccache (or other compiler cache) at buildhosts

Slim version header is added that contains only necessary info. Old version.h is left unchanged but it is possible to include slim one there to reduce code duplication

Why this solution is better then -DXRT_CCACHE?

XRT_CCACHE uses global define -DDISABLE_ABI_CHECK so technicaly compilation result is different and some errors can not be detected (redefinition of variables/macros from version.h for example).

This patch fixes root cause why -DDISABLE_ABI_CHECK was introduced and same code base with same flags can be compiled with or without ccache.

Code that uses version macros without including version.h will break

### DIFF
--- a/src/CMake/config/version-slim.h.in
+++ b/src/CMake/config/version-slim.h.in
@@ -1,0 +1,13 @@
+#ifndef _XRT_VERSION_SLIM_H_
+#define _XRT_VERSION_SLIM_H_
+
+/*
+ * This header contains minimal version info needed for xrt/detail/abi.h header
+ */
+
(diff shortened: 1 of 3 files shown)


stsoe:
Hi @shramov .  Thank you very much for these changes.    I have built on your changes in f #9238.

PR #9232, opened 2025-09-14
Add workflow that builds deb package using debian/rules

shramov:
Debian build rules are not used in regular builds and break from time to time.

Workflow uses dependencies list extracted from debian/control with simple perl script to make sure they are valid.

This is draft until some issues are clarified and resolved:

- [ ] It looks like packaging is gradually shifting from /opt/xilinx/xrt to /usr (see f47d3cd5 for example). Is it true?
- [ ] What is minimal amount of tests that should be performed on installed package? Calling xclbinutil and compiling xrt/00_hello tests is enought?
- [ ] This pull request uses ccache (that greatly reduces compilation time) that depends on #9231 (if it is not merged then using ccache does not make sense).
- [ ] Also there are files installed (but not packaged) in /usr/src/xrt-aws-2.20.0 (dkms files reports that this is xrt-awsmgmt driver). Should it be packaged into xrt-aws-dkms package?

### DIFF
--- a/.github/workflows/debian-validate.yml
+++ b/.github/workflows/debian-validate.yml
@@ -1,0 +1,52 @@
+name: Validate Debian build
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: [ubuntu-24.04]
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+        submodules: recursive
+    - uses: actions/cache@v4
+      with:
+        path: |
+          /home/runner/.ccache
+          /home/runner/.cache/ccache
+        key: debian-validate:${{ matrix.os }}:${{ github.sha }}
+        restore-keys: 'debian-validate:${{ matrix.os }}:'
+    - name: disable man-db trigger # See https://github.com/actions/runner-images/issues/10977
+      run: |
+        echo "set man-db/auto-update false" | sudo debconf-communicate
+        sudo dpkg-reconfigure man-db
+    - name: install deps
+      run: |
+        ln -s build/debian debian
+        sudo apt update
+        sudo apt remove --yes libpython3*-dev # By default github images have python3-dev packages installed, drop them
+        sudo apt install --yes build-essential ccache
+        sudo apt satisfy --yes --no-install-recommends "`perl ./debian/print-build-deps.pl`"
+    - name: build
+      env:
+        CMAKE_CXX_COMPILER_LAUNCHER: ccache
+        CMAKE_C_COMPILER_LAUNCHER: ccache
+      run: dpkg-buildpackage --no-sign -b -j4
+    - name: test
+      continue-on-error: true
+      run: |
+        set -x
+        test ! -f /opt/xilinix/xrt/bin/xclbinutil
+        if pkg-config --exists xrt; then echo "XRT pkg-config file installed: `pkg-config --modversion xrt`"; false; else true; fi
+        sudo dpkg -i ../xrt_*.deb
+        test -f /opt/xilinx/xrt/bin/xclbinutil
+        pkg-config --exists xrt
+        mkdir build-test
+        cd build-test
+        c++ -ohello `pkg-config --cflags --libs xrt` -Wall ../tests/xrt/00_hello/main.cpp
+
+# vim: sts=2 sw=2 et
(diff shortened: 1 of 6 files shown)


stsoe:
> It looks like packaging is gradually shifting from /opt/xilinx/xrt to /usr (see https://github.com/Xilinx/XRT/commit/f47d3cd549dff89f7bcaa48b83254c40f4598503 for example). Is it true?

The CMake files were changed to obey and never set CMAKE_INSTALL_PREFIX.  So it's entirely up to the build to specify the install for packages.

>  What is minimal amount of tests that should be performed on installed package? Calling xclbinutil and compiling xrt/00_hello tests is enough?

Sound sufficient to me.

> This pull request uses ccache (that greatly reduces compilation time) that depends on https://github.com/Xilinx/XRT/pull/9231 (if it is not merged then using ccache does not make sense).

I will look at #9231.   However, ccache is already supported with build.sh.   It sets -DXRT_CCACHE=1 during configuration and disables inclusion of version.h. Do we need anything additional?  I do like the fact that the generated version file is in the same relative location as the installed header.   

> Also there are files installed (but not packaged) in /usr/src/xrt-aws-2.20.0 (dkms files reports that this is xrt-awsmgmt driver). Should it be packaged into xrt-aws-dkms package?

@ManojTakasi , @chvamshi-xilinx ?

shramov:
> The CMake files were changed to obey and never set CMAKE_INSTALL_PREFIX. So it's entirely up to the build to specify the install for packages.

But what is right way in your opinion? Use /opt/xilinx/xrt or /usr for packaging? Official packages (last available at least) use /opt/xilinx/xrt.

> I will look at #9231. However, ccache is already supported with build.sh. It sets -DXRT_CCACHE=1 during configuration and disables inclusion of version.h. Do we need anything additional? I do like the fact that the generated version file is in the same relative location as the installed header.

Oh, i've missed XRT_CCACHE define, but it is not needed if datetime field is not included into whole build tree. Please take a look at #9231, maybe you'll find it usefull.

stsoe:
> But what is right way in your opinion? Use /opt/xilinx/xrt or /usr for packaging? Official packages (last available at least) use /opt/xilinx/xrt.

I am under the impression that upstream is in control of CMAKE_INSTALL_PREFIX.    I am working on a Debian upstream for XRT for NPU.   It will use /usr as the install root.    I am not familiar with what build/debian is used for or by whom?  But I would not recommend changing the install location as this time; let's see if ultimately this build/debian can be merged with what I am working on.

> Please take a look at https://github.com/Xilinx/XRT/pull/9231, maybe you'll find it usefull.

Yes, very useful.  Thank you.

shramov:
> I am not familiar with what build/debian is used for or by whom?

We were using build/debian (with minor tweaks) for 3+ years for Alveo cards but it looks like I've picked up wrong base for packaging.

At first glance 'official' debianization that can be found at https://salsa.debian.org/xilinx-packages-team/xrt lacks xrt-xocl-dkms. So at least for now build/debian is pretty useful.

I'll try to use package from salsa and report back.

stsoe:
> > I am not familiar with what build/debian is used for or by whom?
> 
> We were using build/debian (with minor tweaks) for 3+ years for Alveo cards but it looks like I've picked up wrong base for packaging.
> 
> At first glance 'official' debianization that can be found at https://salsa.debian.org/xilinx-packages-team/xrt lacks xrt-xocl-dkms. So at least for now build/debian is pretty useful.
> 
> I'll try to use package from salsa and report back.

The salsa repo XRT packages is what I am working on updating.  But this upstreaming will change to support NPU only.  It will not be an edge build and there will be no xocl-dkms.  It sounds like buid/debian serves a different purpose.

shramov:
> The salsa repo XRT packages is what I am working on updating. But this upstreaming will change to support NPU only. It will not be an edge build and there will be no xocl-dkms. It sounds like buid/debian serves a different purpose.

Is it true that XRT_NPU and XRT_ALVEO (or XRT_XRT) modes of compilation are in conflict? It looks like that enabling both (NPU and ALVEO or NPU and XRT) is possible. 

XRT_NPU mostly affect CMake component names that are not used in packaging, at least one test is disabled (xclbinutil file-check) and src/runtime_src/core/common/xdp/profile.cpp compiled with -DXDP_CLIENT_BUILD.

stsoe:
This PR appears stale in the context of salsa, where both NPU and Alveo are built together.

PR #9234, opened 2025-09-15
Fix for c compiler warning. 

svup-xilinx:
Fix for c compiler warning in xrt_device.h.
c compiler is throwing warning for including xrt_exception.h in xrt_device.h.
 Removing the warning from xrt_exception.h.
Low
Compiled simple application just including xrt_device.h using gcc compiler in windows. 
NA


stsoe (commented):
Let's not change this file.  Instead remove the warning from xrt_exception.h, it's too noisy.

PR #9235, opened 2025-09-15
Create scratchpad buffer per context instead of per run

rbramand-xilinx:
Changes are made to create scratchpad memory per hw context instead of per run. As scratchpad memory is used to dump content of memory during preemption of a context, creating it per context makes sense. This also reduces the overhead of destruction of xrt::run object as now xrt::hw_context holds the scratch pad buffer object.

It was discovered by @wendyliang25 during analysis of HIP performance.

xrt::hw_context now has member for scratchpad memory which is created on demand.
We get size per column required as argument as the size of scratchpad memory is specific to a device. Shims that use this feature thus needs to implement virtual function get_num_cols which is used to calculate overall size of scratchpad memory of that context.

Low to moderate

Tested with preemption test case on Strix Linux
Also compared the xrt::run destruction time taken on Strix Linux and it improved
Time taken for ~xrt::run::run without this change for sentence_transformers_paraphrase_multilingual_MiniLM_L12_v2 test case is around 488us and with this change its around 100us.

NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -9,6 +9,7 @@
 
 #include "core/common/config_reader.h"
 #include "core/common/message.h"
+#include "core/common/time.h"
 #include "core/common/utils.h"
 #include "core/include/xrt/xrt_hw_context.h"
 #include "core/include/xrt/experimental/xrt_ext.h"
@@ -48,18 +49,6 @@ dump_bo(const char* buf_map, const std::string& filename, size_t size, size_t of
 
   const char* buf = buf_map + offset;
   ofs.write(buf, static_cast<std::streamsize>(size));
-}
-
-// Returns the current timestamp in string format
-static std::string
-get_timestamp_str()
-{
-  auto current_time = std::chrono::system_clock::now();
-  std::time_t time = std::chrono::system_clock::to_time_t(current_time);
-  std::stringstream time_stamp;
-  time_stamp << std::put_time(std::localtime(&time), "%Y-%m-%d_%H-%M-%S");
-
-  return time_stamp.str();
 }
 }
 
@@ -127,7 +116,7 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
         // Use timestamp, slot index in file name
         for (size_t i = 0; i < m_num_uc; i++) {
           auto file_name = "uc_log_" + std::to_string(xrt_core::utils::get_pid()) + "_"
-              + get_timestamp_str() + "_" + std::to_string(m_slot_idx) + "_" + std::to_string(i) + ".bin";
+              + xrt_core::get_timestamp_for_filename() + "_" + std::to_string(m_slot_idx) + "_" + std::to_string(i) + ".bin";
           dump_bo(m_uc_log_bo.map<char*>(), file_name, m_size_per_uc, (i * m_size_per_uc));
         }
       }
@@ -148,6 +137,7 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
   std::unique_ptr<uc_log_buffer> m_uc_log_buf;
   // During preemption, when a hardware context is interrupted L2 memory contents
   // are saved to a scratchpad memory allocated specifically for that context.
+  std::once_flag m_scratchpad_init_flag; // used for thread safe lazy init of scratchpad
   xrt::bo m_scratchpad_buf;
   std::shared_ptr<xrt_core::usage_metrics::base_logger> m_usage_logger =
       xrt_core::usage_metrics::get_usage_metrics_logger();
@@ -401,18 +391,18 @@ public:
 
   // Creates scratchpad memory buffer on demand
   // If buffer is already created, returns the existing one
+  // std::call_once is used to ensure thread safe lazy initialization
   const xrt::bo&
   get_scratchpad_mem_buf(size_t size_per_col)
   {
-    if (m_scratchpad_buf)
-      return m_scratchpad_buf;
-
-    try {
-      // create scratchpad memory buffer using this context
-      m_scratchpad_buf = xrt::ext::bo{xrt::hw_context(get_shared_ptr()),
-                                      size_per_col * m_hdl->get_num_cols()};
-    }
-    catch (...) { /*do nothing*/ }
+    std::call_once(m_scratchpad_init_flag, [this, size_per_col] () {
+      try {
+        // create scratchpad memory buffer using this context
+        m_scratchpad_buf = xrt::ext::bo{xrt::hw_context(get_shared_ptr()),
+                                        size_per_col * m_hdl->get_num_cols()};
+      }
+      catch (...) { /*do nothing*/ }
+    });
 
     return m_scratchpad_buf;
   }
@@ -431,7 +421,7 @@ public:
 
     std::string dump_file_name =
         "preemption_scratchpad_mem_" + std::to_string(m_hdl->get_slotidx()) + "_" +
-        get_timestamp_str() + ".bin";
+        xrt_core::get_timestamp_for_filename() + ".bin";
     dump_bo(m_scratchpad_buf.map<char*>(), dump_file_name, m_scratchpad_buf.size());
 
     std::string msg {"Dumped scratchpad buffer into file : "};
--- a/src/runtime_src/core/common/api/xrt_module.cpp
+++ b/src/runtime_src/core/common/api/xrt_module.cpp
@@ -79,6 +79,8 @@ static const char* const Control_Code_Symbol = "control-code";
 static constexpr uint8_t mangled_prefix_length = 2;
 static constexpr uint8_t decimal_base = 10;
 
+constexpr size_t operator"" _kb(unsigned long long v)  { return 1024u * v; } //NOLINT
+
 struct buf
 {
   std::vector<uint8_t> m_data;
@@ -1503,7 +1505,7 @@ public:
   size_t
   get_scratch_pad_mem_size() const override
   {
-    constexpr size_t scratchpad_mem_size = 512 * 1024; // 512 KB
+    constexpr size_t scratchpad_mem_size = 512_kb;
     return scratchpad_mem_size;
   }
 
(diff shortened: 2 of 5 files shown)


src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1509,7 +1503,8 @@ class module_elf_aie2p : public module_elf
   size_t
   get_scratch_pad_mem_size() const override
   {
-    return m_scratch_pad_mem_size;
+    constexpr size_t scratchpad_mem_size = 512 * 1024; // 512 KB
```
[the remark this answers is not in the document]
stsoe: It may be better to use operator _kb?
```
constexpr size_t operator"" _kb(unsigned long long v)  { return 1024u * v; } //NOLINT
```
Then write
```constexpr auto scratchpad_mem_size = 512_kb```

**Review by stsoe — APPROVED** [review]

File: src/runtime_src/core/common/api/xrt_module.cpp (hunk 1 above)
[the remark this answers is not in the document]
Comment (rbramand-xilinx) [reply src/runtime_src/core/common/api/xrt_module.cpp]: Sure Soren, I will make changes accordingly

**Comment by larry9523** [comment]
The changes look good. Can you please double check that the preemption test cases we used have L1 preemption point, where we really have L2 mem dependency between each layer?

File: src/runtime_src/core/common/api/xrt_hw_context.cpp [hunk 2]
```diff
@@ -36,6 +37,30 @@
...
+}
+
+// Returns the current timestamp in string format
+static std::string
+get_timestamp_str()
+{
+  auto current_time = std::chrono::system_clock::now();
+  std::time_t time = std::chrono::system_clock::to_time_t(current_time);
+  std::stringstream time_stamp;
+  time_stamp << std::put_time(std::localtime(&time), "%Y-%m-%d_%H-%M-%S");
```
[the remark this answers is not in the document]
wendyliang25: can we move this function to https://github.com/Xilinx/XRT/blob/master/src/runtime_src/core/common/debug.h ?

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -383,6 +388,45 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
+      return m_scratchpad_buf;
+
+    try {
+      // create scratchpad memory buffer using this context
+      m_scratchpad_buf = xrt::ext::bo{xrt::hw_context(get_shared_ptr()),
+                                      size_per_col * m_hdl->get_num_cols()};
+    }
+    catch (...) { /*do nothing*/ }
+
+    return m_scratchpad_buf;
```
wendyliang25, line 407: additional empty line, and needs 2 spaces for indentation?

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -383,6 +388,45 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
     }
   }
 
+  // Creates scratchpad memory buffer on demand
+  // If buffer is already created, returns the existing one
+  const xrt::bo&
+  get_scratchpad_mem_buf(size_t size_per_col)
+  {
+    if (m_scratchpad_buf)
```
wendyliang25: do we need lock to protect `m_scratchpad_buf` ? assuming multi app threads can share the same hw context? cc @stsoe ?

src/runtime_src/core/common/api/xrt_hw_context.cpp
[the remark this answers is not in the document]
rbramand-xilinx: moving to debug.h doesn't make sense, may be utils.h would be correct place.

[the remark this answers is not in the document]
wendyliang25: Hi @rbramand-xilinx , thanks, could you move it to `utils.h` ?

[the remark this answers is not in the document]
stsoe: In that case, it needs to be made thread safe. std::localtime is not thread safe.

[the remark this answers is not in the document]
wendyliang25: > In that case, it needs to be made thread safe.

from google: for the threadsafe get local time for POSIX
```
void print_local_time(time_t raw_time) {
    struct tm timeinfo; // Thread-local storage for timeinfo
    localtime_r(&raw_time, &timeinfo); // Thread-safe version

    char buffer[80];
    strftime(buffer, sizeof(buffer), "%Y-%m-%d %H:%M:%S", &timeinfo);
    std::cout << "Thread ID: " << std::this_thread::get_id() << " - Local time: " << buffer << std::endl;
}
```
for windows, it will be `localtime_s`

rbramand-xilinx:
> The changes look good. Can you please double check that the preemption test cases we used have L1 preemption point, where we really have L2 mem dependency between each layer?

Yes Larry, sentence_transformers_paraphrase_multilingual_MiniLM_L12_v2  test case uses L1 preemption point. I have verified this test and it works as expected. But thats said I will try to run this test using multiple threads and see.

src/runtime_src/core/common/api/xrt_hw_context.cpp
rbramand-xilinx: This is in accordance with our coding guidelines.

src/runtime_src/core/common/api/xrt_hw_context.cpp
rbramand-xilinx: Yes I will use std::call_once to ensure thread safe way of creating buffer only once and reusing it for any further calls. Thanks for pointing it out

rbramand-xilinx:
> > The changes look good. Can you please double check that the preemption test cases we used have L1 preemption point, where we really have L2 mem dependency between each layer?
> 
> Yes Larry, sentence_transformers_paraphrase_multilingual_MiniLM_L12_v2 test case uses L1 preemption point. I have verified this test and it works as expected. But thats said I will try to run this test using multiple threads and see.

@larry9523, @chvamshi-xilinx  Tested sentence_transformers test with multiple threads and each thread having its own context and runs the kernel multiple times and I have noticed scratchpad memory is shared across runs in same context and all the iterations passed.
Do we need to do any additional testing? or are we good now

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -386,6 +381,45 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
     }
   }
 
+  // Creates scratchpad memory buffer on demand
+  // If buffer is already created, returns the existing one
+  // std::call_once is used to ensure thread safe lazy initialization
+  const xrt::bo&
+  get_scratchpad_mem_buf(size_t size_per_col)
+  {
+    std::call_once(m_scratchpad_init_flag, [this, size_per_col] () {
```
stsoe, line 398:
```suggestion
    static std::once_flag flag;
    std::call_once(flag, [this, size_per_col] () {
```

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -144,6 +135,10 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
   access_mode m_mode;
   std::unique_ptr<xrt_core::hwctx_handle> m_hdl;
   std::unique_ptr<uc_log_buffer> m_uc_log_buf;
+  // During preemption, when a hardware context is interrupted L2 memory contents
+  // are saved to a scratchpad memory allocated specifically for that context.
+  std::once_flag m_scratchpad_init_flag; // used for thread safe lazy init of scratchpad
```
stsoe, line 140: Why a data member?

stsoe (changes requested):

src/runtime_src/core/common/api/xrt_hw_context.cpp
wendyliang25: > This is in accordance with our coding guidelines.
i overlooked the code, i had thought the return line was part of catch statement.

larry9523:
> > > The changes look good. Can you please double check that the preemption test cases we used have L1 preemption point, where we really have L2 mem dependency between each layer?
> > 
> > 
> > Yes Larry, sentence_transformers_paraphrase_multilingual_MiniLM_L12_v2 test case uses L1 preemption point. I have verified this test and it works as expected. But thats said I will try to run this test using multiple threads and see.
> 
> @larry9523, @chvamshi-xilinx Tested sentence_transformers test with multiple threads and each thread having its own context and runs the kernel multiple times and I have noticed scratchpad memory is shared across runs in same context and all the iterations passed. Do we need to do any additional testing? or are we good now

Thanks @rbramand-xilinx . That should be good enough for now.

larry9523 (approved):

src/runtime_src/core/common/api/xrt_hw_context.cpp
rbramand-xilinx: Hi @stsoe, if I make it static it will be shared across all instances of the class right.
The intended behavior is to initialize scratchpad buffer once per hw ctx object.

stsoe: Oh, yes, you are right.

src/runtime_src/core/common/api/xrt_hw_context.cpp
stsoe: Got it now.   Thanks!

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -401,18 +391,18 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
-    }
-    catch (...) { /*do nothing*/ }
+    std::call_once(m_scratchpad_init_flag, [this, size_per_col] () {
+      try {
+        // create scratchpad memory buffer using this context
+        m_scratchpad_buf = xrt::ext::bo{xrt::hw_context(get_shared_ptr()),
+                                        size_per_col * m_hdl->get_num_cols()};
+      }
+      catch (...) { /*do nothing*/ }
+    });
```
wendyliang25, line 405: Hi @rbramand-xilinx , sorry i forgot to ask, if m_scratchpad_buf creation failed, how does the caller know? and what the caller should do since we are ignoring the failure here. thanks

wendyliang25 (approved):

PR #9239, opened 2025-09-16
Slight change to runner and xrt_hw_context to support smi elf flow

hlaccabu:
Runner could not execute smi validate tests on npu3 with elf flow. 

#### How problem was solved
By looking at runner.cpp, I saw that xrt::kernel was being called for elf flow instead of correct xrt::ext::kernel.

Tested smi validate with elf flow on npu3, runner operated as intended.

### DIFF
--- a/src/runtime_src/core/common/api/hw_context_int.h
+++ b/src/runtime_src/core/common/api/hw_context_int.h
@@ -56,6 +56,12 @@ get_module(const xrt::hw_context& hwctx, const std::string& kname);
 size_t
 get_partition_size(const xrt::hw_context&);
 
+// get_elf_flow() - Returns true if hwctx was created with elf file/flow
+// Returns false everywhere else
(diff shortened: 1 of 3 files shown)


rchane (approved):

stsoe (dismissed):
Ah no, not quite, but close :-) This has to be an internal API. 
Declare get_elf_flow() in core/common/api/hw_context_int.h and define in xrt_hw_context.cpp in the xrt_core::hw_context_int namespace.  No changes to public APIs.

hlaccabu:
> Ah no, not quite, but close :-) This has to be an internal API. Declare get_elf_flow() in core/common/api/hw_context_int.h and define in xrt_hw_context.cpp in the xrt_core::hw_context_int namespace. No changes to public APIs.

Ah, I see it now looking at some of the examples, thanks.

PR #9247, opened 2025-09-18
Check version before printing ctx health exception message

sayyanna:
Added version check before printing ctx health exception message. At present ctx health data structure only supports version 0.


src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4423,6 +4423,8 @@ amend_aie_error_message(const ert_packet* epkt, const std::string& msg)
   std::ostringstream oss;
   oss << msg << "\n";
   auto ctx_health = get_ert_ctx_health_data(epkt);
+  if (ctx_health->version != 0) return oss.str();
```
```suggestion
  if (ctx_health->version != 0) 
    return oss.str();
```

stsoe (changes requested):

stsoe (approved):

PR #9257, opened 2025-09-20
Implement HIP error APIs to get XRT exception information and hardware async error

wendyliang25:

Implement HIP error APIs to get XRT exception information and hardware async error
XRT has rich error information using c++ exception handling, however, HIP APIs are C APIs, this patch set enables users to use `hipGetErrorString()` to get the details of the error.

NPU hardware can raise asynchronous errors which can be related or not related to the last HIP API function call. This patch set implements `hipGetLastError()` and `hipPeekAtLastError()` to get the last async errors, if there was async error happens before, the
API returns `hipErrorUnknown`, otherwise it returns `hipSuccess`. That is in XRT HIP,   `hipGetLastError()` and `hipPeekAtLastError()` are only used to get asynchronous errors from hardware.

When hardware async error is signaled to the kernel driver from hardware, Linux or Windows kernel driver will then cache the last async error. It may or may not be connected to the context of the application. The kernel driver doesn't have enough information to connect a launch. The kernel driver encodes the last async error and caches it.

HIP maps last async error per device local to each thread, when user call get last async error, it gets enquire the last async error from kernel driver through XRT error query API, it compared to the one saved locally with timestamp, if timestamps of the two are the same, it means no new async error, if they are different, it means there is new async error since the last get last error call.

* hipPeekAtLastError() - returns last saved error, if there was async error detected last time. If there was no async error detected last time, it enquire async error from XRT to see if there is new async error has happened. if there is new async error has happened, save it and return error.

* hipGegLastError() - returns last saved error, if there was async error detected last time, and clears the last saved error. If there was no async error detected last time, it enquires async error from XRT to see if there is a new one has happened. If yes, return it.

Test with https://gitenterprise.xilinx.com/XRT/testcases-v2/blob/main/unittests/hip/src/test-error.cpp
* basic error test which gets return value from a failing HIP API call, and use get error string to get the error
* runs get errors in multi threads environment
* create stream with good and bad runs, test to get stream synchronization failure error
* create run to trigger asynchronous error, and use get last error API to get asynchronous error.
Here is the test implementation: 
[hip-test-error.cpp.txt](https://github.com/user-attachments/files/22437751/hip-test-error.cpp.txt)

### DIFF
--- a/src/runtime_src/hip/core/stream.cpp
+++ b/src/runtime_src/hip/core/stream.cpp
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright (C) 2024 Advanced Micro Devices, Inc. All rights reserved.
 
+#include <typeinfo>
+
 #include "hip/config.h"
 #include "hip/hip_runtime_api.h"
 
@@ -112,17 +114,37 @@ stream::
 await_completion()
 {
   std::lock_guard<std::mutex> lk(m_cmd_lock);
+  uint32_t cmd_id = 0;
+  bool has_failure = false;
   while(!m_cmd_queue.empty()) {
     auto cmd = m_cmd_queue.front();
-    cmd->wait();
+    try {
+      cmd->wait();
+      if (cmd->get_state() != command::state::completed)
+        throw std::runtime_error("execution failed.");
+    }
+    catch (const std::exception &ex) {
+      std::string err_str = "CMD[" + std::to_string(cmd_id) + "]:" + get_unmangled_type_name(*cmd) + ":" + ex.what();
+      xrt::core::hip::error::instance().record_local_error(hipErrorLaunchFailure, err_str);
+      has_failure = true;
+    }
+    catch (...) {
+      std::string err_str = "CMD[" + std::to_string(cmd_id) + "]:" + get_unmangled_type_name(*cmd) + ":unknown exception";
+      xrt::core::hip::error::instance().record_local_error(hipErrorLaunchFailure, err_str);
+      has_failure = true;
+    }
     // kernel_start and copy_buffer cmds needs to be explicitly removed from cache
     // there is no destroy call for them
     if (cmd->get_type() != command::type::event)
       command_cache.remove(cmd.get());
     m_cmd_queue.pop_front();
+    cmd_id++;
   }
   // reset m_top_event as stream completed
   m_top_event = nullptr;
+  // throw launch failure error if there is any command failed to execute.
+  if (has_failure)
+    throw_hip_error(hipErrorLaunchFailure, "Stream execution failed.");
 }
 
 void


src/runtime_src/hip/core/common.h
```diff
@@ -8,11 +8,21 @@
...
+
 #include "context.h"
 #include "device.h"
 
 #include <stack>
 #include <thread>
+#include <string>
+#include <typeinfo>
+#include <memory>
+#include <cstdlib>
```
stsoe: sort within group

src/runtime_src/hip/core/common.h
```diff
@@ -8,11 +8,21 @@
...
+
+#if defined(__GNUC__) || defined(__clang__)
+#include <cxxabi.h>
```
stsoe:
```suggestion
# include <cxxabi.h>
```

src/runtime_src/hip/core/common.h
```diff
@@ -42,32 +52,68 @@ insert_in_map(map& m, value&& v)
   m.add(handle, std::move(v));
   return handle;
 }
+
+template <class T>
+std::string
+get_untangled_type_name(const T& obj)
```
stsoe:
```suggestion
get_unmangled_type_name(const T& obj)
```

src/runtime_src/hip/core/common.h
```diff
@@ -42,32 +52,68 @@ insert_in_map(map& m, value&& v)
...
+
 template<typename F> hipError_t
 handle_hip_func_error(const char* func_name, hipError_t default_err, F && f)
 {
+  xrt::core::hip::error::instance().reset_local_errors();
   try {
     std::forward<F>(f)();
     return hipSuccess;
   }
+  catch (const xrt::core::hip::hip_exception &ex) {
```
stsoe:
```suggestion
  catch (const xrt::core::hip::hip_exception& ex) {
```
Nit, most commonly used.

src/runtime_src/hip/core/error.cpp
```diff
@@ -4,140 +4,274 @@
...
+  {
+    return m_what.c_str();
+  }
+
+  // TODO: there can be extra checking to return better HIP error code based on the
+  // value of exception. However, there will be work needs to do in the platform dependent
+  // driver, e.g. today, on windows, it only throws std::runtime_error if it fails to
+  // allocate buffer with no specific error code, once, it provides error code, we can
+  // extend this implementation to provide proper HIP error code.
+  const static std::map<int, hipError_t> sys_to_hip_error_map =
```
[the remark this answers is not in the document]
stsoe: It generally is better when static complex data structures are kept as static local variables within a single function that uses the map.  I am guilty of making static globals also, but when certainly if a global is only used within a single function, it should be defined statically within this function.

src/runtime_src/hip/core/error.h
```diff
@@ -7,42 +7,166 @@
...
 
 #include "hip/config.h"
+#include "hip/core/device.h"
 #include "hip/hip_runtime_api.h"
 
+#include "xrt/experimental/xrt_error.h"
+
 namespace xrt::core::hip
 {
+  /**
```
stsoe, line 24: Do not indent within a namespace.

```suggestion
namespace xrt::core::hip {
/**
```

src/runtime_src/hip/core/error.h
```diff
@@ -7,42 +7,166 @@
...
 
-  protected:
-    error();
+    /**
+     * @brief Get the thread-local error string for a given HIP error code.
+     * @param err HIP error code.
+     * @return Error string.
+     */
+    const
+    char*
```
stsoe: Probably a bit easier to read the return type on a single line
```suggestion
    const char*
```

src/runtime_src/hip/core/error.h
```diff
@@ -7,42 +7,166 @@
...
+    /**
+     * @brief Thread-local map of HIP error codes to error strings.
+     */
+    std::map<hipError_t, std::string> m_local_errors;
+
+    /**
+     * @brief get last error without lock
+     * @return Last device HIP error code
+     */
+    hipError_t get_last_error_nolock();
```
stsoe:
```suggestion
    hipError_t 
    get_last_error_nolock();
```

src/runtime_src/hip/core/error.cpp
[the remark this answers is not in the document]
wendyliang25: Thanks @stsoe , I moved it as const to a function.

src/runtime_src/hip/core/common.h
wendyliang25: updated

src/runtime_src/hip/core/error.h
wendyliang25: I have removed the indentation within the namespace

src/runtime_src/hip/core/error.h
wendyliang25: sorry that I left the unimplemented function, i have removed this function.

PR #9259, opened 2025-09-22
Revert "hip: function: introduce runs cache in hip functions (#9100)"

wendyliang25:
This reverts commit b2f0c97545ff20e0fb7d0d51e6b136dfb3461b52. Different resources pooling for an XRT run are handled within XRT, it is not required to have XRT runs pooling in HIP.


src/runtime_src/hip/core/event.cpp
```diff
@@ -132,7 +132,7 @@ kernel_start::kernel_start(std::shared_ptr<stream> s, std::shared_ptr<function>
    */
 
   // create run object and set args
-  r = func->get_run();
+  r = xrt::run(k);
```
chvamshi-xilinx, line 137: Do we still need definitions of get_run and release_run?

wendyliang25: Hi @chvamshi-xilinx , good catch, i have removed the `get_run()` and `release_run()`, could you check again? thanks

chvamshi-xilinx (approved):

PR #9260, opened 2025-09-22
[XDP] Add support to configure trace/profile for a specific partition in a multipartition scenario

snigdha-gupta:
In a multipartition scenario, there was no way to configure trace and profile for a specific partition. This commit adds support for configuring trace and profile for a specific partition. Trace and profile settings are independent of each other, allowing flexibility to users.

N/A - this is a feature enhancement

Implemented first-match logic: the first matching partition is configured. Trace and profile can be configured independently per partition (e.g., partition 1 can have profile enabled but trace disabled)

Since first-match is used, the order of partition configurations matters

Tested on VCK190 for the following cases:
- First-match logic with overlapping partitions
- Independent trace and profile settings per partition
- Edge cases: no matching partition, all partitions match

The first-match behavior and independence of trace and profile settings should be documented for ease of users


IshitaGhosh (approved):
Changes look good for single threaded host application.
@snigdha-gupta  If host application creates HW Contexts using multiple threads, this new implementation might end up configuring AIE Profile/Trace for multiple contexts. However, this is a complicated scenario and can be revisited later.

PR #9261, opened 2025-09-23
Drop redundant assignment operator for xrt::uuid

shramov:
Code that resets uuid variable using value assigment `uuid_variable = {};` is broken after #8622 

Drop redundant assignment operator.
xrt::uuid already has non-explicit constructor from xuid_t that enables assignment from xuid_t.

Removed operator is choosen by compiler in value assignment `uuid_variable = {};` that leads to call with null pointer and segmentation fault.

Minimal example that demonstrates bug: https://godbolt.org/z/GvExv98j8 (line 52)
Minimal code that uses xrt/xrt_uuid.h header (compile with `g++ -ouuid uuid.cc -luuid`):

```
#include <xrt/xrt_uuid.h>

int main()
{
	xrt::uuid uuid = {};
	uuid = {};
}
```

### DIFF
--- a/src/runtime_src/core/include/xrt/xrt_uuid.h
+++ b/src/runtime_src/core/include/xrt/xrt_uuid.h
@@ -1,19 +1,6 @@
-/**
- * Copyright (C) 2016-2021 Xilinx, Inc
- *
- * Licensed under the Apache License, Version 2.0 (the "License"). You may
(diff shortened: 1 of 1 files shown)


stsoe:
CI is  picky about when it can validate a PR.  I am merging this PR without CI validation. @manikandan-xilinx

PR #9262, opened 2025-09-23
Get partition size from xclbin instead of shim

rbramand-xilinx:
Added logic to get partition size from xclbin at the time of hw ctx creation instead of calling a virtual function on shim handle
Also fixed a small bug when setting elf flow

This change is added for EOU which eliminates adding overriding functions in all required shims as information can be parsed from xclbin at userspace

Low

Tested by running preemption scratchpad test on Strix and the test works as expected

NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -295,6 +295,7 @@ public:
       m_hdl = m_core_device->create_hw_context(elf, m_cfg_param, m_mode);
       m_partition_size = part_size;
       create_module_map(elf);
+      m_elf_flow = true; // ELF flow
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -187,10 +189,32 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
+        return 0;
+
+      // try to get partition size from xclbin AIE_METADATA section
+      auto aie_part = xrt_core::xclbin::get_aie_partition(axlf);
+      return aie_part.ncol;
+    }
+    catch (...) {
+      return 0;
+    }
+  }
```
stsoe, line 211: I think this function and aie_paritition_obj from xclbin_parser.* all belongs in xrt_xclbin.*   But that's more work, and maybe silly if we are obsoleting xclbins.

PR #9266, opened 2025-09-24
Prepare --transform-pdi to support windows

xfreid:
Enable aie-pdi-transform windows build, this PR is only to prepare --transform-pdi to support windows, --transform-pdi switch itself is still disabled for windows 

Currently aie-pdi-transform build is disabled on windows

Update the source code in aie-pdi-transform by replacing posix specific code with standard C code; update xclbinutil CMakeLists.txt to link against the static library produced by aie-pdi-transform on Windows

No

Unittest, manual test

N/A

### DIFF
--- a/src/runtime_src/tools/xclbinutil/CMakeLists.txt
+++ b/src/runtime_src/tools/xclbinutil/CMakeLists.txt
@@ -65,10 +65,7 @@ endif()
 
 # ==-- x c l b i n u t i l --==================================================
 
-# aie-pdi-transform only supports linux build currently
-# if(NOT WIN32)
-  add_subdirectory(aie-pdi-transform)
-# endif()
+add_subdirectory(aie-pdi-transform)
 
 set(XCLBINUTIL_NAME "xclbinutil")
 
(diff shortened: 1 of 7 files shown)


stsoe (commented):
These changes are little bit beyond me. 
Coding LGTM, but I can't review the effect of the changes.

It does look like the PR description doesn't match the changes.  Is Windows disabled or enabled?

xfreid:
> These changes are little bit beyond me. Coding LGTM, but I can't review the effect of the changes.
> It does look like the PR description doesn't match the changes. Is Windows disabled or enabled?

The Windows build is enabled. Without this change, aie-pdi-transform is only built and linked on Linux; with this change, aie-pdi-transform is built and linked on both Linux and Windows. 
The pdi transformation functionality is controlled by --transform-pdi, which is ONLY enabled on Linux. This doesn't change. I will try to enable this switch for Windows in the next PR.
This PR mainly lays down the foundation for the next PR.
The changes in this PR are mainly to replace the posix specific code (for example, <fcntl.h>, <unistd.h>,  open(), read()) with standard C (for example, fopen(), fread()), so that aie-pdi-transform can be built for both Linux and Windows.
Please let me know if this clarifies your questions.

stsoe (approved):

PR #9267, opened 2025-09-24
Update gcc-8 check in build.sh for ARM64 platforms

hlaccabu:
Silent exit on arm64 build if gcc-8 isn't present (even if gcc version > 8)

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -21,18 +21,22 @@ if [[ $CMAKE_MAJOR_VERSION != 3 ]]; then
     fi
 fi
 
-
 if [[ $CPU == "aarch64" ]] && [[ $OSDIST == "ubuntu" ]]; then
-    # On ARM64 Ubuntu use GCC version 8 if available since default
-    # (GCC version 7) has random Internal Compiler Issues compiling XRT
-    # C++14 code
-    gcc-8 --version > /dev/null 2>&1
-    status1=$?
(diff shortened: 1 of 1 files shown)


build/build.sh
```diff
@@ -21,18 +21,19 @@ if [[ $CMAKE_MAJOR_VERSION != 3 ]]; then
...
+
+    gcc_version=$(gcc --version 2> /dev/null | head -n1 | awk '{print $3}' | cut -d. -f1)
+    gpp_version=$(g++ --version 2> /dev/null | head -n1 | awk '{print $3}' | cut -d. -f1)
+
+    if [[ "$gcc_version" -ge 8 && "$gpp_version" -ge 8 ]]; then
+	export CC=gcc
+	export CXX=g++
+    else
+	echo "Installed gcc/g++ version is < 8, please run xrt_deps.sh"
+	exit 1
```
stsoe, line 39: Sorry didn't see this earlier.   Now on Ubunut20.04 where gcc version is 7, how will this new logic switch to gcc-8 installed via xrt-deps?   I think we still need to set CC and CXX to gcc-8 in this case.

hlaccabu: I see, I can add an else if gcc-8 is present, export CC=gcc-8

hlaccabu:
> Yeah, looks like it will work. Did you try it?

Yes, I was able to compile on arm64 without issue :)

PR #9268, opened 2025-09-26
Change xbmgmt2 executable name to xbmgmt

shramov:
When installing to system path wrapper script is not used and 'xbmgmt' command is available under 'xbmgmt2' name.

Use same name as wrapper script (and same as win32 build).

Alternative is to rename binary on install but that is more complex solution.

Scripts that rely on unwrapped name will be broken.

Utility is already named as xbmgmt (https://xilinx.github.io/XRT/master/html/xbmgmt.html) so nothing is changed


chvamshi-xilinx:
ok to test

chvamshi-xilinx (approved):

PR #9270, opened 2025-09-26
hip: CMake: propagate XRT include header directory to user

wendyliang25:
XRT HIP extension header will be installed in XRT include header directory, propagate the XRT include header directory to user, such that when user depends on XRT::xrt_hip, the XRT header directory will be added by CMake to the include system directories path.


stsoe (approved):
LGTM

PR #9271, opened 2025-09-27
Do not fail when LINUX_VERSION variable is empty

shramov:
Running build/build.sh on debain/testing fail due to missing version tag.

Debian testing and unstable do not have version number so LINUX_VERSION variable is empty. CMake is not happy with unquoted use of empty variables.

Use quotes around ${LINUX_VERSION}. Alternative is to get VERSION_CODENAME if ID is missing.

Also looks like there is code duplication in src/CMake/nativeLnx.cmake:
```
if (${LINUX_FLAVOR} MATCHES "^centos")
  execute_process(
    COMMAND awk "{print $4}" /etc/redhat-release
    COMMAND tr -d "\""
    OUTPUT_VARIABLE LINUX_VERSION
    OUTPUT_STRIP_TRAILING_WHITESPACE
  )
else()
  execute_process(
    COMMAND awk -F= "$1==\"VERSION_ID\" {print $2}" /etc/os-release
    COMMAND tr -d "\""
    OUTPUT_VARIABLE LINUX_VERSION
    OUTPUT_STRIP_TRAILING_WHITESPACE
)
endif()
```
and src/CMake/cpackLin.cmake:
```
if (DEFINED CROSS_COMPILE)
  set(CPACK_REL_VER ${LINUX_VERSION})
elseif (${LINUX_FLAVOR} MATCHES "^centos")
  execute_process(
    COMMAND awk "{print $4}" /etc/redhat-release
    COMMAND tr -d "\""
    OUTPUT_VARIABLE CPACK_REL_VER
    OUTPUT_STRIP_TRAILING_WHITESPACE
  )
else()
  execute_process(
      COMMAND awk -F= "$1==\"VERSION_ID\" {print $2}" /etc/os-release
      COMMAND tr -d "\""
      OUTPUT_VARIABLE CPACK_REL_VER
      OUTPUT_STRIP_TRAILING_WHITESPACE
)
endif()
```
Maybe it can be replaced by simple `set(CPACK_REL_VER "${LINUX_VERSION}")`?

### DIFF
--- a/src/CMake/cpackLin.cmake
+++ b/src/CMake/cpackLin.cmake
@@ -58,7 +58,7 @@ set(CPACK_COMPONENT_XRT-DEV_DEPENDS xrt)
 SET(CPACK_RPM_SPEC_MORE_DEFINE "%define __python python3")
 
 if (DEFINED CROSS_COMPILE)
-  set(CPACK_REL_VER ${LINUX_VERSION})
+  set(CPACK_REL_VER "${LINUX_VERSION}")
 elseif (${LINUX_FLAVOR} MATCHES "^centos")
   execute_process(
     COMMAND awk "{print $4}" /etc/redhat-release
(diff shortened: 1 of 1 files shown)


stsoe:
Hi @shramov .  Changes are good, but does debian/testing have to include cpackLin in the first place?  I ran into the same problem with XRT on salsa, but I excluded cpackLin.cmake by defining `-DXRT_UPSTREAM_DEBIAN=1` when configuring (see [here](https://github.com/stsoe/XRT-DEBIAN/blob/main/debian/rules)).   There are other CMake configs guarded by XRT_UPSTREAM.

shramov:
This does not affect packaging but when running build.sh directly. I was checking that my other changes do not break normal builds.

PR #9283, opened 2025-09-30
Trace/Profile support using runlist

rbramand-xilinx:
Added trace/profile support using runlist.
Earlier XDP hooks were in xrt::hw_context constructor and destructor but now because of recent optimizations we have to move them to xrt::runlist.
Spec for this change is captured in confluence page - https://amd.atlassian.net/wiki/spaces/~rbramand/pages/1117815036/Trace+Profile+flow+proposal+using+runlist

While the solution has its own limitations it can act as workaround to collect XDP data

Low as the feature works only when ini options related to xdp are enabled

Tested using runlist test case on strix linux and xdp init and exit runs are properly added and removed as expected

May be we need to update about limitations on when Profile/Trace data can be captured.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -12,6 +12,7 @@
 #include "core/common/time.h"
 #include "core/common/utils.h"
 #include "core/include/xrt/xrt_hw_context.h"
+#include "core/include/xrt/xrt_kernel.h"
 #include "core/include/xrt/experimental/xrt_ext.h"
 #include "core/include/xrt/experimental/xrt_module.h"
 #include "bo_int.h"
@@ -141,6 +142,11 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
   // are saved to a scratchpad memory allocated specifically for that context.
   std::once_flag m_scratchpad_init_flag; // used for thread safe lazy init of scratchpad
   xrt::bo m_scratchpad_buf;
+  // Vector of XDP runs to be added at beginning and end of runlist
+  // These runs initializes/configures AI array and collects the profile/trace data
+  std::vector<xrt::run> m_xdp_init_runs;
+  std::vector<xrt::run> m_xdp_exit_runs;
+
   std::shared_ptr<xrt_core::usage_metrics::base_logger> m_usage_logger =
       xrt_core::usage_metrics::get_usage_metrics_logger();
   bool m_elf_flow = false;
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -141,6 +142,11 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
   // are saved to a scratchpad memory allocated specifically for that context.
   std::once_flag m_scratchpad_init_flag; // used for thread safe lazy init of scratchpad
   xrt::bo m_scratchpad_buf;
+  // Vector of XDP runs to be added at beginning and end of runlist
+  // These runs initializes/configures AI array and collects the profile/trace data
+  std::vector<xrt::run> m_xdp_init_runs;
+  std::vector<xrt::run> m_xdp_exit_runs;
+
```
stsoe: I would prefer to see these managed by xdp itself.  Is that possible?

I imagine xdp will be calling the register functions, alas it can keep the data in xdp and we don't need register and unregister here in hwctx.

The runlist can make a callback into xdp to get the list of runs to prepend and append. 

This would be cleaner and more general; I think.

IshitaGhosh: Hi @stsoe, XDP will create separate "xrt::run" objects for its different plugins (say AIE Trace, AIE Profile). Currently XDP does not have a container to store the different configuration and exit xrt::run objects for the different plugins.

We can create one, but I wanted to understand the benefits of XDP maintaining it as opposed to xrt.
If XRT does not store the run objects from XDP, then XDP will need to be informed about when to send the run objects before XRT calls execute on runlist.
I had a little chat with Rahul. He explained XRT can add new callbacks before runlist execute and XDP can send the vector of custom run objects. In that solution, there would be two new callbacks into XDP (one for init run objects and another for exit run objects). That sounds a bit more complex, as opposed to two new vectors 
"  std::vector<xrt::run> m_xdp_init_runs;
  std::vector<xrt::run> m_xdp_exit_runs;"
 in xrt. If xrt maintains the vector, then XDP can register the run objects during hw context creation (like it does now) and be done.

stsoe: Basically, I don't want XDP to leave bread crumbs in various XRT objects.    We had that a long while back when we supported only OpenCL and Alveo.  It took a long time to root out XDP from all the places.

It make sense that XDP wants to append and prepend a runlist, and it makes sense for xrt::runlist to support pre and append.  But it doesn't make sense that the code is wired to store these pre and append run objects outside XDP and for the runlist know about init and exit objects that some some reason are stored in the hwctx implementation.   Just too strange, IMO.

I really think XDP should manage this without storing objects in XRT coreutil.  The hook should be entirely within the implementation of xrt::runlist.  Right before submission, a callback into XDP (or anyone else registered) with the runlist, XDP then prepends the init run objects and appends the exit run objects.  Is that not possible?

IshitaGhosh: Thanks for explaining in details @stsoe . We discussed with Vamshi and Rahul. XDP will store the run objects it needs to prepend and append to runlist and two callbacks from XRT to XDP will be provided to pass the vector of run objects to XRT.
These callbacks will be executed for all devices that support runlist. However, XDP plans to create custom run objects for some special cases only.

rbramand-xilinx:
Hi @stsoe and @IshitaGhosh,
I have made changes as suggested by Soren and I have added placeholder functions in profile.h.
Please review.

stsoe (commented):
Can we nuke the changes to xrt_hw_context* ?

I think the xrt_kernel.cpp changes are still too xdp specific.  Maybe I am asking too much, but how about implementing xrt::runlist::append(xrt::run) and prepend(xrt::run).  Then have one callback into any registered functions at execution time (we already appending in execute()) with the runlist?   Let xdp pre- and append what it needs. 

Hopefully the callback would not be registered in the normal case, hence be a no-op.

chvamshi-xilinx:
Closing this PR as firmware team provided a mechanism to disable the new mode

PR #9285, opened 2025-09-30
Fix Dtrace bug

rbramand-xilinx:
This PR fixes bug in dtrace code.
Dtrace library provided in ini option is opened using dlopen and is used to create, fill dtrace buffer and is later closed. When dlclose is called on the library all the variables stored in the library gets destroyed but while dumping the dtrace buffer we need some data in dtrace library which is lost. Added changes to extend the lifetime of library handle created by adding it as a member in module object.

This PR also adds changes to initialize log buffer initial 8 bytes that will be used for metadata to all zeros before configuring

https://github.com/Xilinx/XRT/pull/8721 introduced the bug. It was discovered while testing the feature on windows

Extended the scope of library handle created using dlopen to the scope of module object by making it a member in it.

Low

Tested dtrace function on Telluride and it works as expected.
Need to test on aie4 windows where the issue is reported
 
NA


chvamshi-xilinx (approved):

pranjalv-xilinx:
@stsoe can we merge this change?

PR #9288, opened 2025-09-30
HIP memory APIs fixes

wendyliang25:
This patchset contains the following fixes to HIP APIs:
* check if the host memory has been registered in host memory registration API to avoid double registration
* according to HIP memory API documentation, does nothing and returns 0 if freeing null pointer
* according to HIP memory API documentation, allows get device memory pointer for memory registered with `hipHostRegister()` with `hipHostRegisterMapped`


wendyliang25:
Hi @stsoe , @chvamshi-xilinx , ping, could you review this PR? thanks

PR #9291, opened 2025-10-01
Handle exception in xbtop when no xclbin is loaded into card (and some cleanups)

shramov:

xbtop fails to display dynamic region page for cards after reset (when there is no xclbin loaded).

Also xbtop is ignoring Ctrl-C due to strict 'raw' mode used for TTY.

Last patch removes bashisms so wrapper script is compatible with sh (dash or ash).

Feel free to drop any commits you don't like from this pull request

By the way, same cleanups can be applied to xrt-smi and xbmgmt wrappers - no arguments, cleaner win32 detection and exec, launcher script also uses args copy and thus 'eval' for launch

I can provide cleanup if you think this can be useful

It would be great if someone can check that calling `exec "$0".bat "$@"` is ok, maybe there is some windows magic that is incompatible with exec. However there is no .bat file for xbtop...

### DIFF
--- a/src/runtime_src/core/tools/xbtop/xbtop
+++ b/src/runtime_src/core/tools/xbtop/xbtop
@@ -1,26 +1,6 @@
-#!/bin/bash
-
-# -- Detect a Windows environment and automatically switch to the .bat file
-if [[ "`uname`" == windows32* ]] || [[ "`uname`" == CYGWIN* ]] ; then
-  trap "" INT
-  "$0.bat" "$@"
-  exit $?
-fi
+#!/bin/sh
 
 # Working variables
 XRT_PROG="`dirname \"$0\"`/../python/xbtop.py"
 
-# -- Examine the options 
-XRTWARP_PROG_ARGS_size=0
-XRTWRAP_PROG_ARGS=()
-while [ $# -gt 0 ]; do
-  case "$1" in
-    *)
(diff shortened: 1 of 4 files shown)


shramov:
By the way, same cleanups can be applied to xrt-smi and xbmgmt wrappers - no arguments, cleaner win32 detection and exec

launcher script also uses args copy

I can provide cleanup if you think this can be useful

src/runtime_src/core/tools/xbtop/xbtop
```diff
@@ -1,26 +1,11 @@
-#!/bin/bash
+#!/bin/sh
 
 # -- Detect a Windows environment and automatically switch to the .bat file
-if [[ "`uname`" == windows32* ]] || [[ "`uname`" == CYGWIN* ]] ; then
-  trap "" INT
-  "$0.bat" "$@"
-  exit $?
-fi
+case "`uname`" in
```
AShivangi: actually, since we don't have bat, does this even make sense? xbtop is not supported on windows today

shramov: Yep, there is no bat, but I've tried to preserve functionality (and as example how to do this check without grep or bash-specific syntax).

Removed this code in new version

AShivangi (approved):

PR #9293, opened 2025-10-02
Fix xrt::runner to preserve legacy throughput/latency calculations

stsoe:
Make sure legacy mode uses number of runs in recipe for latency and throughput calculations

Throughput number regressed compared to before #9226 

In legacy mode, the number of recipe::runs is used for throughput and latency calculations.  In non-legacy mode, the recipe::runs is always considered as one runlist and calculations are based on the how many times (depth) the runlist is duplicated.

### DIFF
--- a/src/runtime_src/core/common/runner/runner.cpp
+++ b/src/runtime_src/core/common/runner/runner.cpp
@@ -2055,6 +2055,7 @@ class profile
     std::string m_name;
     mode m_mode = mode::none;
     size_t m_depth = 1;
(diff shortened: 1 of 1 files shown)


aktondak (approved):
Did you run the throughput test with these changes? 
What is the op/s number you see?

aktondak:
> Did you run the throughput test with these changes? What is the op/s number you see?

I see you've posted in description. Looks good, Thanks !

PR #9298, opened 2025-10-06
CR-1248332: Display vss reset IP info under the --info option

xvijaysri:
Some kernel instances (vss reset) are not available in build_metadata section but available in ip_layout section because they are created internally during v++ link.
Info like base address of vss rest IPs are still useful and should be displayed under xclbinutil --info option. Therefore xclbinutil --info should display info about all kernels in ip_layout section.

### DIFF
--- a/src/runtime_src/tools/xclbinutil/FormattedOutput.cxx
+++ b/src/runtime_src/tools/xclbinutil/FormattedOutput.cxx
@@ -913,11 +913,13 @@ reportKernels(std::ostream& _ostream,
   }
   // Print all remaining IPs and the base address
   for (auto& ptIPData : ipLayout) {
+    // m_name is of the form '<kernel name>:<instance name>'	  
     auto sKernelInstance = ptIPData.get<std::string>("m_name");
     auto pos = sKernelInstance.find(':');
     if (pos != std::string::npos){
       auto sKernel = sKernelInstance.substr(0,pos);
       auto sInstance = sKernelInstance.substr(pos+1);
+      // If the kernel name matches one of the kernels in build metadata, do nothing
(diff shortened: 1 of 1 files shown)


src/runtime_src/tools/xclbinutil/FormattedOutput.cxx
```diff
@@ -909,6 +911,22 @@ reportKernels(std::ostream& _ostream,
       }
     }
   }
+  // Print all remaining IPs and the base address
+  for (auto& ptIPData : ipLayout) {
+    auto sKernelInstance = ptIPData.get<std::string>("m_name");
+    auto pos = sKernelInstance.find(':');
```
xfreid, line 918: Add a short comment to explain the format of the m_name

src/runtime_src/tools/xclbinutil/FormattedOutput.cxx
```diff
@@ -909,6 +911,22 @@ reportKernels(std::ostream& _ostream,
...
+      auto sKernel = sKernelInstance.substr(0,pos);
+      auto sInstance = sKernelInstance.substr(pos+1);
+      if(std::find(kernel_names.begin(), kernel_names.end(), sKernel) == kernel_names.end()) {
+        _ostream << boost::format("%s %s\n") % "Kernel:" % sKernel;
+        _ostream << "--------------------------\n";
+        _ostream << boost::format("%-16s %s\n") % "Instance:" % sInstance;
+        auto sBaseAddress = ptIPData.get<std::string>("m_base_address");
+        _ostream << boost::format("   %-13s %s\n") % "Base Address:" % sBaseAddress;
+      }
+    }
```
xfreid, line 930: What happens when sKernel is not found in kernel_names?
If it needs to be handled, add code
if it can be ignored, add comment

xfreid (approved):

PR #9304, opened 2025-10-08
Patch nativeLnx.cmake to use pkg-config for libdrm not found

jgmelber:

I was running into errors in the XRT build step towards building xdna-driver. I could not find the package libdrm even though it was installed on the system. This patch helps cmake look in the right place. 

Package not found libdrm 

Use pkg-config from system to avoid version compatibility 

This was tested on a system with Ubuntu 24.04 and Linux 6.14 building [xdna-driver@1.6](https://github.com/amd/xdna-driver/tree/1.6) and [xrt@0212043](https://github.com/Xilinx/XRT/tree/021204355eeaa034ff69aae407ace2265adf047a)

### DIFF
--- a/src/CMake/nativeLnx.cmake
+++ b/src/CMake/nativeLnx.cmake
@@ -10,6 +10,8 @@
 
 
 # --- PkgConfig ---
+# Force use of system pkg-config to avoid version compatibility issues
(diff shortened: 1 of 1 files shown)


jgmelber:
> Looks like an environment issue, where an old local pkg-config is in the path before /usr/bin. I would prefer not to merge this proposed change.

Thanks for the comments. If anything this can serve as documentation in case anyone else runs into this as well.

PR #9317, opened 2025-10-13
Added extra gaurd to run bash completion scripts in the non posix bash

ManojTakasi:
https://jira.xilinx.com/browse/CR-1251600 Sourcing setup.sh triggers bash-completion loading under bash running in POSIX (sh) mode. In that mode function names containing hyphens are rejected, so /etc/bash_completion.d/apport_completion fails with: /etc/bash_completion.d/apport_completion: line 107: `_apport-bug': not a valid identifier

Added extra gaurd to run bash completion scripts in the non posix bash

low

Tested on ubuntu 22.04 on both bash and dash

NA


chvamshi-xilinx:
Please backport to 2025.2 and mark the CR fixed

PR #9319, opened 2025-10-13
Fix for CR-1251818

snigdha-gupta:
[CR-1251818](https://jira.xilinx.com/browse/CR-1251818)

The XAIE API used in getCounterPayload expects a relative column (relCol) rather than the absolute column (col). Passing the absolute column resulted in accessing gated tiles during profiling setup and produced repeated XAie_LinuxIO_Read32 gated tile errors

Use the relative column (relcol) when calling getCounterPayload

N/A

Tested on VCK190

N/A

### DIFF
--- a/src/runtime_src/xdp/profile/plugin/aie_profile/util/aie_profile_config.cpp
+++ b/src/runtime_src/xdp/profile/plugin/aie_profile/util/aie_profile_config.cpp
@@ -7,6 +7,7 @@
 #include "xdp/profile/plugin/aie_profile/util/aie_profile_util.h"
 #include "xdp/profile/plugin/aie_base/aie_base_util.h"
 #include "xdp/profile/database/static_info/aie_util.h"
+#include "xdp/profile/database/database.h"
 
 #include <cmath>
 #include <cstring>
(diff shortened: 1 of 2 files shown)


src/runtime_src/xdp/profile/plugin/aie_profile/edge/aie_profile.cpp
```diff
@@ -440,7 +440,7 @@ namespace xdp {
           uint16_t phyEndEvent   = physicalEventIds.second;
 
           // Get payload for reporting purposes
-          uint64_t payload = getCounterPayload(aieDevInst, tileMetric.first, type, col, row, 
+          uint64_t payload = getCounterPayload(aieDevInst, tileMetric.first, type, relCol, row,
```
IshitaGhosh: Have we verified load_xclbin style applications too?

jvillarre (approved):

vipangul (approved):
We should keep note of below for 2026.1:
-  Profile APIs support for multi-partition designs would need to be verified using xrt.ini settings.

PR #9335, opened 2025-10-16
hip: remove memory pool APIs

wendyliang25:
Remove memory pool related APIs.

Remove HIP memory pool APIs due to:
* API returns virtual device memory, which we cannot make any use of it for NPU
* there is no connection between the virtual device memory and the underlying allocated XRT buffer object
* the allocated virtual device memory is determined by size only.


wendyliang25:
ping, hi @stsoe , @sonals , @chvamshi-xilinx , any comments? thanks

PR #9340, opened 2025-10-17
hip: initial implementation of HIP Graph APIs.

svup-xilinx:
Initial implementation of HIP Graph APIs:
- Added graph_node, graph, graph_exec classes to support for HIP Graph implementation.
- Added kernel_list_start which is inherited from command class to support xrt::run_list.
- graph_node is generic node to contain command of different types (memcpy, memset, kernel etc...)
- graph_node also contains the information about connection to other nodes.
- graph class is to contain all the nodes in a list and maintain handle to node map.
- graph class provides a function to return the flattened graph in a list after resolving all the dependencies.
- graph_exec class to resolve all nodes and make an optimized executable list.
- Removed cstream from command class as graph APIs does not provide stream when adding a graph node, cstream is not used and not required to provide in constructor of command.

None
Low
Tested on Windows Strix machine with 1, 2, 4, 16, 32 kernel nodes.
NA

### DIFF
--- a/src/runtime_src/hip/api/hip_graph.cpp
+++ b/src/runtime_src/hip/api/hip_graph.cpp
@@ -49,9 +49,12 @@ hip_graph_add_kernel_node(hipGraph_t g, const hipGraphNode_t *pDependencies,
   return node_hdl;
 }
 
+// TODO: Implement error reporting and logging for graph instantiation:
+//   - pErrorNode: Pointer to error node if error occured during graph instantiation
+//   - pLogBuffer, bufferSize: log messages about instantiation to the buffer
 static command_handle
-hip_graph_instantiate(hipGraph_t g, hipGraphNode_t *pErrorNode,
-                      char *pLogBuffer, size_t bufferSize)
+hip_graph_instantiate(hipGraph_t g, hipGraphNode_t* /*pErrorNode*/,
+                      char* /*pLogBuffer*/, size_t /*bufferSize*/)
 {
   throw_invalid_resource_if(!g, "graph is nullptr");
 
--- a/src/runtime_src/hip/core/event.cpp
+++ b/src/runtime_src/hip/core/event.cpp
@@ -409,6 +409,9 @@ submit()
   if (kernel_list_start_state == state::running)
     return true;
 
+  if (kernel_list_start_state == state::completed)
+    return true;
+
   if (kernel_list_start_state != state::init)
     return false;
 
--- a/src/runtime_src/hip/core/graph.cpp
+++ b/src/runtime_src/hip/core/graph.cpp
@@ -34,7 +34,7 @@ get_ordered_nodes() const
 
   // Compute indegree and enqueue root nodes
   for (const auto& node : m_node_list) {
-    indegree[node.get()] = node->get_deps_size();
+    indegree[node.get()] = node->get_deps_list().size();
     if (indegree[node.get()] == 0)
       q.push(node);
   }
--- a/src/runtime_src/hip/core/graph.h
+++ b/src/runtime_src/hip/core/graph.h
@@ -22,7 +22,6 @@ private:
   std::shared_ptr<command> m_cmd;
   std::vector<std::shared_ptr<graph_node>> m_deps_list; // parents
   std::vector<std::weak_ptr<graph_node>> m_children;    // children as weak_ptr to avoid cycles
-  size_t m_deps_size = 0;
 
 public:
   graph_node() = default;
@@ -34,12 +33,6 @@ public:
   get_cmd() const
   {
     return m_cmd;
-  }
-
-  size_t
-  get_deps_size() const
-  {
-    return m_deps_size;
   }
 
   const std::vector<std::shared_ptr<graph_node>>&
@@ -71,7 +64,6 @@ public:
   {
     parent->m_children.push_back(weak_from_this());
     m_deps_list.push_back(std::move(parent));
-    m_deps_size++;
   }
 };
 


stsoe (commented):
Hi @svup-xilinx .  Thanks for the pull request, I will wait reviewing until you have addressed all clang-tidy recommendations.

svup-xilinx:
> Hi @svup-xilinx . Thanks for the pull request, I will wait reviewing until you have addressed all clang-tidy recommendations.

Hi @stsoe all clang-tidy recommendations are addressed.

src/runtime_src/hip/core/graph.h
```diff
@@ -0,0 +1,124 @@
...
+  std::vector<std::weak_ptr<graph_node>> m_children;    // children as weak_ptr to avoid cycles
+  size_t m_deps_size = 0;
+
+public:
+  graph_node() = default;
+  explicit graph_node(std::shared_ptr<command> cmd)
+    : m_cmd(std::move(cmd))
+  {}
+
+  const std::shared_ptr<command>
```
stsoe: Return types by value should not be declared const.  It prevents certain compiler optimizations.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-return-const

Same thing goes with by-value function arguments.

src/runtime_src/hip/core/graph.h
```diff
@@ -0,0 +1,124 @@
...
+      if (auto sptr = wptr.lock()) {
+        result.push_back(sptr);
+      }
+    }
+    return result;
+  }
+
+  // Adds a dependency (parent) to this node and updates the parent's children list.
+  // This establishes a direct connection from parent to child in the graph.
+  // The child is stored as a weak_ptr in the parent's m_children vector to avoid strong reference cycle dependencies.
```
stsoe: break line

src/runtime_src/hip/core/graph.h
```diff
@@ -0,0 +1,124 @@
...
+  void
+  add_dep_node(std::shared_ptr<graph_node> parent)
+  {
+    m_deps_list.push_back(parent);
+    parent->m_children.push_back(weak_from_this());
```
stsoe: swap and move parent in m_deps_list.push_back ??

src/runtime_src/hip/core/graph.cpp
```diff
@@ -0,0 +1,154 @@
...
+std::vector<std::shared_ptr<command>>
+init(std::shared_ptr<graph> graph)
+{
+  std::vector<std::shared_ptr<command>> cmd_list;
+  for (const auto& node : graph->get_ordered_nodes()) {
+    // Skip nodes that are null or have no command to execute
+    if (!node || !node->get_cmd())
+      continue;
+
+    auto cmd_ptr = node->get_cmd();
```
stsoe:
```suggestion
    if (!node)
      continue;

    auto cmd_ptr = node->get_cmd();
    if (!cmd_ptr)
      continue;
```

src/runtime_src/hip/core/graph.cpp
```diff
@@ -0,0 +1,154 @@
...
+    if (cmd_ptr->get_type() != command::type::kernel_start) {
+      cmd_list.push_back(cmd_ptr);
+      continue;
+    }
+
+    // For kernel_start commands, group by hardware context if possible
+    auto cmd = std::dynamic_pointer_cast<kernel_start>(cmd_ptr);
+    if (!cmd)
+      continue;
+    auto hw_ctx = cmd->get_function()->get_hw_ctx();
```
stsoe:
```suggestion
      continue;
      
    auto hw_ctx = cmd->get_function()->get_hw_ctx();
```

src/runtime_src/hip/core/graph.cpp
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  return cmd_list;
+}
+
+graph_exec::
+graph_exec(std::shared_ptr<graph> graph)
+  : command(type::graph_exec)
+  , m_cmd_exec_list(init(graph))
+{}
```
stsoe, line 138: Thank you for following the initializer list idiom with empty constructor body!  Your code is very easy to read.

src/runtime_src/hip/core/graph.cpp
```diff
@@ -0,0 +1,154 @@
...
+    for (auto cmd : m_cmd_exec_list) {
+      cmd->wait();
+    }
+    set_state(state::completed);
+    return true;
+  }
+  else if (graph_exec_state == state::completed)
+    return true;
+
+  return false;
```
stsoe: I think this is the same logic with one less indentation.

```suggestion
  auto graph_exec_state = get_state();
  if (graph_exec_state == state::completed)
    return true;
    
  if (graph_exec_state != state::running)
    return false;

  for (auto cmd : m_cmd_exec_list)
    cmd->wait();

  set_state(state::completed);
  return true;
```

src/runtime_src/hip/core/graph.cpp
```diff
@@ -0,0 +1,154 @@
...
+    for (auto cmd : m_cmd_exec_list) {
+      cmd->submit();
+    }
+    set_state(state::running);
+    return true;
+  }
+  else if (graph_exec_state == state::running)
+    return true;
+
+  return false;
```
stsoe: I think this can be reduced one level also, same as wait().

src/runtime_src/hip/core/graph.h
```diff
@@ -0,0 +1,124 @@
...
+#define xrthip_graph_h
+
+#include "event.h"
+#include "module.h"
+
+// node_handle - opaque graph node handle
+using node_handle = void*;
+
+// graph_handle - opaque graph handle
+using graph_handle = void*;
```
stsoe, line 16: These should not be in the global namespace.

stsoe (changes requested):
Some mostly style comments.  Your code is clean and very easy to read.

src/runtime_src/hip/core/event.cpp
```diff
@@ -15,15 +15,15 @@ event::event()
...
 void event::record(std::shared_ptr<stream> s)
 {
-  cstream = std::move(s);
+  m_recorded_stream = std::move(s);
   auto ev = std::dynamic_pointer_cast<event>(command_cache.get(static_cast<command_handle>(this)));
   throw_invalid_handle_if(!ev, "event passed is invalid");
   if (is_recorded()) {
     // already recorded
-    cstream->erase_cmd(ev);
+    m_recorded_stream->erase_cmd(ev);
```
wendyliang25: There is a bug in this function which mentioned in https://github.com/Xilinx/XRT/pull/9328

src/runtime_src/hip/core/graph.cpp
```diff
@@ -0,0 +1,154 @@
...
+}
+
+static
+std::vector<std::shared_ptr<command>>
+init(std::shared_ptr<graph> graph)
+{
+  std::vector<std::shared_ptr<command>> cmd_list;
+  for (const auto& node : graph->get_ordered_nodes()) {
+    // Skip nodes that are null or have no command to execute
+    if (!node || !node->get_cmd())
```
wendyliang25: in which case we will have empty node or a node don't have command to execute? thanks

src/runtime_src/hip/core/graph.cpp
```diff
@@ -0,0 +1,154 @@
...
+      continue;
+
+    auto cmd_ptr = node->get_cmd();
+
+  // Add non-kernel_start commands to cmd_list
+    if (cmd_ptr->get_type() != command::type::kernel_start) {
+      cmd_list.push_back(cmd_ptr);
+      continue;
+    }
```
wendyliang25, line 76: do we want to have non-kernel_start node in this implementation?
how about the dependency between the non kernel node and the next kernel node?

src/runtime_src/hip/core/graph.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+    // For kernel_start commands, group by hardware context if possible
+    auto cmd = std::dynamic_pointer_cast<kernel_start>(cmd_ptr);
+    if (!cmd)
```
wendyliang25: why we throw error here? as it previously check if it is non kernel node already, looks like this will be some runtime error.
But again, do we want to support non-kernel nodes in this implementation? or we assume all the non kernel node will complete in submit, i think the memory commands today such as memcpy calling std::async() in submit() which looks like will break the dependencies here. maybe need to those commands, either not using std::async() all the time or add a mode not to do std::async() in submit()

src/runtime_src/hip/core/graph.cpp
```diff
@@ -0,0 +1,154 @@
...
+      continue;
+    auto hw_ctx = cmd->get_function()->get_hw_ctx();
+    if (!hw_ctx)
+      continue;
```
wendyliang25: is it possible to find the hardware context later? or shall we error here? thanks

src/runtime_src/hip/core/graph.cpp
```diff
@@ -0,0 +1,154 @@
...
+        last_cmd->add_run(cmd->get_run());
+        continue;
+      }
+    }
+
+    // If the last command is not a kernel_list_start or with different hardware context,
+    // Create a new kernel_list_start command for this hardware context
+    auto kl_cmd = std::make_shared<kernel_list_start>(hw_ctx);
+    kl_cmd->add_run(cmd->get_run());
+    cmd_list.push_back(std::move(kl_cmd));
```
wendyliang25: just wonder if the command is on a different context to the previous command in the graph, does it suggest there is dependency between the two? do we also need to insert dependencies between them? or the start sequence will be handled by IPU firmware in FIFO fashion, and the next runlist will not start until the previous one completes? thanks.
Or for the first implementation, we don't try to solve this issue first.
Or you can chain the list of commands together, that is add dependencies of kernel_list_start command, or add the kernel_list_start command to the chain of the previous kernel_list_start command and to the non previous non kernel commands.

src/runtime_src/hip/core/graph.cpp
svup-xilinx: we shall error out here. I will update the code.

src/runtime_src/hip/core/graph.cpp
svup-xilinx: Addressed dependency issue between the nodes.

src/runtime_src/hip/core/event.cpp
```diff
@@ -406,6 +406,38 @@ bool memcpy_command::wait()
...
+bool
+kernel_list_start::
+submit()
+{
+  auto kernel_list_start_state = get_state();
+  if (kernel_list_start_state == state::running)
+    return true;
+
+  if (kernel_list_start_state != state::init)
+    return false;
```
wendyliang25, line 416: how about if it is complete? in this case, shall we return true?

src/runtime_src/hip/api/hip_graph.cpp
```diff
@@ -0,0 +1,154 @@
...
+  auto hip_func = hip_mod->get_function(func_hdl);
+  throw_invalid_resource_if(!hip_func, "invalid function passed");
+
+  // Create a kernel_start command and wrap it in a graph_node
+  auto hip_cmd = std::make_shared<kernel_start>(hip_func, pNodeParams->kernelParams);
+  auto node_hdl = hip_graph->add_node(std::make_shared<graph_node>(hip_cmd));
+
+  // Add dependencies if provided
+  if (pDependencies && numDependencies) {
+    auto node = hip_graph->get_node(node_hdl);
```
wendyliang25, line 44: this line can be move out of the loop?

src/runtime_src/hip/api/hip_graph.cpp
```diff
@@ -0,0 +1,154 @@
...
+static command_handle
+hip_graph_instantiate(hipGraph_t g, hipGraphNode_t *pErrorNode,
+                      char *pLogBuffer, size_t bufferSize)
+{
+  throw_invalid_resource_if(!g, "graph is nullptr");
+
+  auto hip_graph = graph_cache.get_or_error(g);
+  throw_invalid_resource_if(!hip_graph, "invalid graph passed");
+
+  return insert_in_map(command_cache, std::make_shared<graph_exec>(hip_graph));
```
wendyliang25, line 64: we are not returning error Node, and not returning log Buffer. Is there any plan to implement that?
please add comments to this function and maybe don't put the name of those unused input args. e.g. change it to:
`hip_graph_instantiate(hipGraph_t g, hipGraphNode_t* /* pErrorNode */, char* /*pLogBuffer*/, size_t /*bufferSize*/)` and add comments to explain. thanks

src/runtime_src/hip/core/graph.h
```diff
@@ -0,0 +1,128 @@
...
+  // Adds a dependency (parent) to this node and updates the parent's children list.
+  // This establishes a direct connection from parent to child in the graph.
+  // The child is stored as a weak_ptr in the parent's m_children vector to avoid
+  // strong reference cycle dependencies.
+  void
+  add_dep_node(std::shared_ptr<graph_node> parent)
+  {
+    parent->m_children.push_back(weak_from_this());
+    m_deps_list.push_back(std::move(parent));
+    m_deps_size++;
```
wendyliang25: do we need this variable? can we get the size from `m_deps_list` instead?

src/runtime_src/hip/core/graph.cpp
```diff
@@ -0,0 +1,191 @@
...
+    if (indegree[node.get()] == 0)
+      q.push(node);
+  }
+
+  // enqueue nodes only when all their dependencies are added to result
+  while (!q.empty()) {
+    auto node = q.front();
+    q.pop();
+    result.push_back(node);
+    for (const auto& child : node->get_children()) {
```
wendyliang25, line 47: from previous lines:
```
if (indegree[node.get()] == 0)
      q.push(node);
  }
```
does that mean only if the node has no child, it will be opushed to `q`, and this while loop, only look for nodes in `q`, then this `for` loop will never run?
Actually, is there any reason to have `q`, why not just check again `m_node_list`, and doesn't looks like this is required, as we can get the number of dependencies from `node->get_children()`?
the implementation looks like will end up as this:
graph: 
```
   A1 -> A
   B1 -> B
   A - > B   C
```
the above implementation will result in : `C`, and the function will throw error in the end.

Should it end up as : runlist `A1 - A - B1 - B - C` or runlist `A1 - B1 - A - B - C`?

Thanks

wendyliang25:
Hi @svup-xilinx , thanks for your patch, could you add test cases to https://gitenterprise.xilinx.com/XRT/testcases-v2/tree/main/unittests/hip:
Do you have the following test cases:
* all kernel nodes:
   * graph1: A_hwctx_0, B_hwctx_0, C_hwctx_0
   * graph2: A_hwctx_0 -> B_hwctx_0 -> C_hwctx_0
   * graph3: A_hwctx_0 -> B_hwctx_0, C_hwctx_0 -> B_hwctx_0, B_hwctx_0 -> D_hwctx_0
   * graph4: A_hwctx_0, B_hwctx_0, C_hwctx_1
   * graph4: A_hwctx_0 -> B_hwctx_0 -> C_hwctx_1
   * graph4: A_hwctx_0 -> B_hwctx_1 -> C_hwctx_0
   * graph4: A_hwctx_0 -> B_hwctx_0, C_hwctx_1 -> B_hwctx_0, B_hwctx_0 -> D_hwctx_1
* kernel nodes + non-kernel nodes:
   * graph1: A_memset -> A_hwctx_0, B_hwctx_0, C_hwctx_0
   * graph2: A_memset -> A_hwctx_0, B_memst->B_hwctx_0,  A_hwctx_0 -> B_hwctx_0, B_hwctx_0 -> C_hwctx_1

src/runtime_src/hip/api/hip_graph.cpp
svup-xilinx: It was intentionally kept inside the if statement to limit its scope, as obtaining the node isn’t necessary when there are no dependency nodes

src/runtime_src/hip/core/graph.cpp
svup-xilinx: indegree represents the **parents** to the current node. 
from the example graph
   **A1 -> A
   B1 -> B
   A - > B   C**

The runlist will be: `A1 - B1 - A - B - C` 

- **A1** and **B1** have no dependencies (in-degree = 0), so they are enqueued in q first.
- node **A** has in-degree 1 (one parent). A is enqueued after A1 is dequeued and processed, which reduces A’s in-degree to zero.
- node **B** has in-degree 2 (two parents). B is enqueued only after both B1 and A have been dequeued and processed, reducing B’s in-degree to zero.
- node **C** has in-degree 1 (one parent). C is enqueued only after A has been dequeued and processed, reducing C’s in-degree to zero.
- The `q` will be empty when all nodes are processed and pushed to `result` vector

wendyliang25: i c, thanks @svup-xilinx , i had actually misunderstood the child nodes, ok, m_deps is to record dependencies, and children nodes represent dependents.

src/runtime_src/hip/api/hip_graph.cpp
wendyliang25: ok

src/runtime_src/hip/core/event.cpp
svup-xilinx: yes. I will update the code.

src/runtime_src/hip/core/graph.h
svup-xilinx: removed m_deps_size. using  m_deps_list  to get the size.

PR #9342, opened 2025-10-17
Limit maximum number of jobs in xrt-runner queue

stsoe:
Scripted execution can specify any number of jobs, which can lead to out-of-memory when creating and adding the xrt::runner jobs to the queue.

Extend profile execution::mode with validate to allow easy filtering of profile executions to any one of the supported modes (validate, latency, throughput).

This PR adds a new switch `--max-queue-size <num>` which causes `job_queue::add(...)` to wait if current number of jobs in the queue exceeds the specified max.  The option implies `--asap` so that jobs are processed as soon as possible rather than all created up-front.

### DIFF
--- a/src/runtime_src/core/common/runner/main.cpp
+++ b/src/runtime_src/core/common/runner/main.cpp
@@ -120,12 +120,19 @@ filter_mode(json& profile, const std::string& mode)
   if (mode == "all")
     return;
 
+  // Legacy profile nothing to filter
+  if (!profile.contains("executions"))
+    return;
+
   auto& execs = profile["executions"];
   execs.erase(std::remove_if(execs.begin(), execs.end(),
                              [mode](const json& exec) {
-                               return exec["mode"] != mode;
+                               return (!exec.contains("mode") || exec["mode"] != mode);
                              }),
               execs.end());
+
+  if (execs.empty())
+    throw std::runtime_error("No execution profile with mode '" + mode + "'");
 }
 
 // Touch up profile(s) with specified iterations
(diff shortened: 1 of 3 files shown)


maxzhen:
Driver is already doing back pressure by blocking job submission thread, if the number of commands (jobs) submitted to driver exceeds the limit (currently 4 on both Linux and Windows). Why do we need this in runner?

stsoe:
> Driver is already doing back pressure by blocking job submission thread, if the number of commands (jobs) submitted to driver exceeds the limit (currently 4 on both Linux and Windows). Why do we need this in runner?

Because we otherwise run out of resources when for example a script contains many models.   The runner creates a (job) xrt::runner objects for each model in the script file, the resources used by a job are not released until the job has finished execution.   We quickly run out of memory if we do not block job creation after some number of jobs are active while waiting for some jobs to complete.    I agree the switch is a bit magical and probably should be based on memory consumption instead, e.g. block while current memory use is > some limit.

maxzhen:
> > Driver is already doing back pressure by blocking job submission thread, if the number of commands (jobs) submitted to driver exceeds the limit (currently 4 on both Linux and Windows). Why do we need this in runner?
> 
> Because we otherwise run out of resources when for example a script contains many models. The runner creates a (job) xrt::runner objects for each model in the script file, the resources used by a job are not released until the job has finished execution. We quickly run out of memory if we do not block job creation after some number of jobs are active while waiting for some jobs to complete. I agree the switch is a bit magical and probably should be based on memory consumption instead, e.g. block while current memory use is > some limit.

If you, first, create all jobs, then submit them, driver's back pressure can't help. But, should you create and submit job right away? Submission is async, so why wait?

PR #9360, opened 2025-10-22
AIESW-15957 : Fix circular dependency bug b/w xrt::module and xrt::hw_context

rbramand-xilinx:
Fixed circular dependency b/w xrt::hw_context and xrt::module.

xrt::hw_context was holding a map of xrt::module's in full ELF flow and xrt::module was holding pdi xrt::bo's which were creating using hw ctx that caused circular dependency. Because of this both xrt::module and xrt::hw_context objects were not getting destroyed at application exit. This PR fixes this issue.

The bug was introduced by PR - https://github.com/Xilinx/XRT/pull/9030 which moved holding pdi xrt::bo's from module_sram to xrt::module that added xrt::hw_context dependency indirectly (as xrt::bo holds xrt::hw_context)

It was discovered in local testing of @wendyliang25. 

Storing Elf's map in xrt::hw_context instead of module's map and creating module on the fly when requested from xrt:;kernel.
xrt::module is still created while storing xrt:;elf because the parsing logic to get kernel's info is still in xrt::module and moving that logic to xrt:;elf needs lot of refactoring. Added a TODO in the code for that part.

Low as only full ELF flow in strix is impacted (which is fixed)

Tested full ELF flow on strix linux and confirmed that hw ctx is destroyed properly.

NA

Thanks to @wendyliang25 and @stsoe for identifying and debugging the issue.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -147,7 +147,13 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
 
   std::shared_ptr<xrt_core::device> m_core_device;
   xrt::xclbin m_xclbin;
-  std::map<std::string, xrt::module> m_module_map; // map b/w kernel name and module
+
+  // map b/w kernel name and Elf
+  // Stores the Elf corresponding to each kernel name it contains.
+  // This map is used for lookup when creating xrt::kernel object
+  // using kernel name.
+  std::map<std::string, xrt::elf> m_elf_map;
+
   // No. of cols in the AIE partition managed by this hw ctx
   // Devices with no AIE will have partition size as 0
   uint32_t m_partition_size = 0;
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -397,8 +407,8 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
   xrt::module
   get_module(const std::string& kname) const
   {
-    if (auto itr = m_module_map.find(kname); itr != m_module_map.end())
-      return itr->second;
+    if (auto itr = m_elf_map.find(kname); itr != m_elf_map.end())
+      return xrt::module{itr->second}; // create module from the ELF that has this kernel
```
stsoe, line 411: Isn't it a problem that multiple xrt::modules are now created for potentially the same elf?  The modules are pretty heavy-weight.

stsoe: Cleanup to follow. 

> module creation for second time can be eliminated if we move parsing ELF logic to get kernels info from xrt::module to xrt::elf

PR #9364, opened 2025-10-22
AIESW-16275 Support tracking of internal BO creation

stsoe:
XRT should allow KMD to track all internal buffers created as part of running an application.  In particular, KMD should accumulate the size of all internal buffers.

In order to support this, XRT must channel all internal BO creation through a single internal API that supports tracking of different use types of buffers.  For example, if XRT creates a preemption buffer, then the use type should reflect this, if XRT creates an instruction buffer, then the use type should reflect this.

It is up to KMD to determine the granularity of how it collects internal BO usage, for example, it could bucket all internal BOs into one and report only one number, or it could be more granular and provide a number of each use type of a buffer.

The BO flags passed from XRT to UMD (and for Windows to KMD) encodes the use type in the flags argument.  If the use type is different from 0, it implies the buffer is an internal XRT created buffer.

This PR may require changes to SHIM, but at least on Linux the changes have been validated through model tests, albeit, the use type is not tracked by KMD at this point.

The PR definitely breaks binary compatibility between xrt_coreutil and xrt_core.

### DIFF
--- a/src/runtime_src/core/common/api/bo_int.h
+++ b/src/runtime_src/core/common/api/bo_int.h
@@ -30,7 +30,7 @@ enum class use_type {
   log         = XRT_BO_USE_LOG,         // logging info
   debug_queue = XRT_BO_USE_DEBUG_QUEUE, // debug queue data
   uc_debug    = XRT_BO_USE_UC_DEBUG,    // microblaze debug data
-  host_only   = XRT_BO_USE_HOST_ONLY,   // instruction (instructon buffer)
+  host_only   = XRT_BO_USE_HOST_ONLY,   // system memory buffer
   instruction = XRT_BO_USE_INSTRUCTION, // instruction (instructon buffer)
   preemption  = XRT_BO_USE_PREEMPTION,  // preemption data
   scratch_pad = XRT_BO_USE_SCRATCH_PAD, // scratch pad data


maxzhen (approved):

src/runtime_src/core/common/api/bo_int.h
```diff
@@ -24,13 +24,18 @@ get_offset(const xrt::bo& bo);
...
-  log = XRT_BO_USE_LOG, // For logging info
-  debug_queue = XRT_BO_USE_DEBUG_QUEUE, // For debug queue data
-  uc_debug = XRT_BO_USE_UC_DEBUG // For microblaze debug data
+  debug       = XRT_BO_USE_DEBUG,       // debug data
+  kmd         = XRT_BO_USE_KMD,         // shared with kernel mode driver
+  dtrace      = XRT_BO_USE_DTRACE,      // dynamic trace data
+  log         = XRT_BO_USE_LOG,         // logging info
+  debug_queue = XRT_BO_USE_DEBUG_QUEUE, // debug queue data
+  uc_debug    = XRT_BO_USE_UC_DEBUG,    // microblaze debug data
+  host_only   = XRT_BO_USE_HOST_ONLY,   // instruction (instructon buffer)
```
rbramand-xilinx: comment for host_only flag seems incorrect

PR #9365, opened 2025-10-23
hip: event: correct event completion timestamping

wendyliang25:
Before, event always timestamp before it launches its chain of commands, which causes some issues:
* whenevery event synchronization is called, the event will timestamp even it has completed before.

This patch also make the following change:
* reorganizes the event sync() and submit() due to they almost the same except that submit() will not wait for dependencies to complete. Doing this change is because, the timestamp will be updated in these two functions
* event query() function checks state only instead of checking each depedencies as event completion will be check in sync() and submit().

### DIFF
--- a/src/runtime_src/hip/core/event.cpp
+++ b/src/runtime_src/hip/core/event.cpp
@@ -75,26 +75,40 @@ bool event::is_recorded()
 bool event::query()
 {
   //This function will return true if all commands in the appropriate stream which specified to hipEventRecord() have completed.
-  std::lock_guard lock(m_recorded_cmds_lock);
-  for (auto& rec_com : m_recorded_commands){
-    state command_state = rec_com->get_state();
-    if (command_state != state::completed){
-      return false;
-    }
-  }
-  return true;
+  std::lock_guard lock(m_state_lock);
+  return (get_state() == state::completed);
 }
 
-// check if all dependencies are completed and update event state accordingly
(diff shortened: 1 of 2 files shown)


src/runtime_src/hip/core/event.cpp
```diff
@@ -75,26 +75,40 @@ bool event::is_recorded()
...
-  std::lock_guard lock(m_recorded_cmds_lock);
-  for (auto& rec_com : m_recorded_commands){
-    state command_state = rec_com->get_state();
-    if (command_state != state::completed){
-      return false;
-    }
-  }
-  return true;
+  std::lock_guard lock(m_state_lock);
+  return (get_state() == state::completed);
```
stsoe, line 79: Shouldn't the mutex lock be inside get_state()?

wendyliang25: Hi @stsoe , the current implementation have the m_state_lock to event instead of the basic class command, and the get_state() is the basic class function. I think the basic class doesn't have any lock, so that the child class can control when to do the lock, here in case of event, trying to lock the state from getting the state, do something and then set the state.
In this particular function, it only get the state.

wendyliang25: Discussed with @stsoe , since base class `cmd` doesn't have a lock, and it is up to child classes to implement lock if required, will keep the locking inside this event `query()` function instead of moving it to the basic class().

PR #9368, opened 2025-10-24
Add aie coredump support using hw ctx

rbramand-xilinx:
Added support to get aie coredump of all the tiles within a context
spec for same can be found at - https://amd.atlassian.net/wiki/spaces/AIE/pages/1098471914/AIE+Coredump

We want to get coredump from both API level and using xrt-smi so implemented the feature using device query mechanism by using ctx id to identify which ctx dump is needed.

Low

Tested aie coredump on Telluride and the feature works as expected

NA


chvamshi-xilinx (approved):

stsoe (approved):
LGTM

PR #9373, opened 2025-10-24
Fix bug to remove circular dependency

rbramand-xilinx:
Removed circular dependency b/w xrt::hw_context and xrt::bo (scratchpad mem buf).
Also made changes to use common API (xrt_core::bo_int::create_bo) for internal bo creations

Scratchpad buffer is needed one per hw ctx, so moved scratchpad xrt::bo to xrt::hw_context level which added circular dependency as bo stores hw ctx that is used to create it

scratchpad buffer is now created with device which removes circular dependency and as this bo is member of ctx its lifetime is tied to ctx.

Low

Tested with testcase that has L2 preemption (test that use scratchpad memory) and the test works as expected. Hw ctx also destroyed properly

NA

### DIFF
--- a/src/runtime_src/core/common/api/bo_int.h
+++ b/src/runtime_src/core/common/api/bo_int.h
@@ -24,18 +24,19 @@ get_offset(const xrt::bo& bo);
 // enum for different buffer use flags
 // This is for internal use only
 enum class use_type {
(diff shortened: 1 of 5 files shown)


stsoe (approved):
Awesome.  Thank you!

chvamshi-xilinx (approved):

PR #9374, opened 2025-10-24
[XDP] Fix for CR-1254001 (for 2025.2 branch)

snigdha-gupta:
CR-1254001

This commit fixes an issue where AIE trace behavior was inconsistent when running multipartition designs consecutively without rebooting the board. During the first run, trace data would be collected successfully, but on the second run, the warning: "[XRT] WARNING: Neither PLIO nor GMIO trace infrastructure is found in the given design. So, AIE event trace will not be available." would be thrown.
The root cause was traced to incorrect UUID retrieval logic that used `.back()` on the xclbin_slots query result. In multipartition designs where the first xclbin contains PL components and the second contains AIE components, the `.back()` method would incorrectly reference the AIE xclbin. This caused the warning to be thrown as AIE xclbin does not have PLIO information.

The problem was solved by using the hw_ctx_impl to retrieve UUID for hw_ctx style runs. This approach ensures that we obtain the correct xclbin regardless of the partition ordering. 

N/A

- VCK 190 (both hw_ctx and non hw_ctx flow)
- Client
- VE2

N/A


jvillarre (approved):

IshitaGhosh (approved):
Changes look good

PR #9387, opened 2025-10-29
AIESW-16751 XRT umd and user mode library to support windows versioning

stsoe:
Add versioninfo resource to DLLs and APPs in XRT.

Configure a `version-dll.rc` and `version-app.rc` at CMake configure time. The config file is copied from ticket where it is pre-confgured with required fields.

The generated file is include in CMake targets.  For now only a small subset of targets (those that are distributed) include the versioninfo.

Subject to change.

image

### DIFF
--- a/build/build-win22.sh
+++ b/build/build-win22.sh
@@ -175,7 +175,7 @@ if [ $dbg == 1 ]; then
         echo "${cmake_flags[@]}"
         "$CMAKE" -G "Visual Studio 17 2022" -B $BUILDDIR_DOS/WDebug $cmake_flags $BUILDDIR_DOS/../src
     fi
-    "$CMAKE" --build $BUILDDIR_DOS/WDebug --verbose --config Debug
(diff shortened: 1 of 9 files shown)


arpitxilinx:
@stsoe Is this version # used just for the identifying the version of dll or the intent is also to check for compatibility between XRT and UMD?
The reason I ask is similar thing has to be done for windbg extension as well.  When the relevant data structs in driver changes then that change has to be reflected in extension.

stsoe:
> @stsoe Is this version # used just for the identifying the version of dll or the intent is also to check for compatibility between XRT and UMD? The reason I ask is similar thing has to be done for windbg extension as well. When the relevant data structs in driver changes then that change has to be reflected in extension.

I imagine both, but probably just for visual inspection.  I don't know how this info is used?  When I propagate the change to XRT-MCDM, I will update the dlls there including yours.

PR #9391, opened 2025-10-31
Fix for CR-1254841 (2025.2 branch)

snigdha-gupta:
CR-1254841

PLIO trace files are incorrectly mapped in hw_context designs containing multiple independent partitions with different trace configurations (PLIO in first partition, GMIO in second partition)

Prevent configuration of any upcoming PLIO partitions after the first PLIO partition is configured. This ensures trace file mappings remain consistent across partitions

N/A

Tested on VCK190

N/A


IshitaGhosh (approved):
Changes look good

PR #9392, opened 2025-10-31
Hip: graph: Fix Coverity warnings

svup-xilinx:
Fix coverity warnings:
- performance inefficiency in passing shared pointer
None
Low
NA

### DIFF
--- a/src/runtime_src/hip/core/graph.cpp
+++ b/src/runtime_src/hip/core/graph.cpp
@@ -59,7 +59,7 @@ get_ordered_nodes() const
 
 static
 std::vector<std::shared_ptr<graph_node>>
-init(std::shared_ptr<graph> graph)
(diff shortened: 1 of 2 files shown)


src/runtime_src/hip/core/graph.cpp
```diff
@@ -134,7 +134,7 @@ init(std::shared_ptr<graph> graph)
 graph_exec::
 graph_exec(std::shared_ptr<graph> graph)
   : command(type::graph_exec)
-  , m_node_exec_list(init(graph))
+  , m_node_exec_list(init(std::move(graph)))
```
stsoe: I don't see the shared ownership of graph.   The graph is not stored anywhere here, so the signature of this function should be either `const std::shared_ptr<graph>&` or `graph* graph`, ditto for the init function.

PR #9397, opened 2025-10-31
AIESW-14372  xrt-smi] Convert preemption benchmark to Runner

AShivangi:
[AIESW-14372](https://jira.xilinx.com/browse/AIESW-14372) [xrt-smi] Convert preemption benchmark to Runner

New change

Converted the preemption testcase to utilize the new archive architecture. VTD is being changed by https://github.com/Xilinx/VTD/pull/28 

N/A

Tested on windows/MCDM

N/A

### DIFF
--- a/src/runtime_src/core/tools/common/tests/TestPreeemptionOverhead.cpp
+++ b/src/runtime_src/core/tools/common/tests/TestPreeemptionOverhead.cpp
@@ -22,35 +22,32 @@ using json = nlohmann::json;
 static constexpr uint32_t num_of_preemptions = 500;
 
 // ----- S T A T I C   M E T H O D S -------------------------------------------
+
+
+
 static double
 measure_preemption_overhead(const std::shared_ptr<xrt_core::device>& dev,
                             const std::string& recipe_data,
                             const std::string& profile_data,
                             const xrt_core::runner::artifacts_repository& artifacts_repo)
 {
-  // Run without preemption
-  xrt_core::device_update<xq::preemption>(dev.get(), static_cast<uint32_t>(0));
-  
-  double baseline_exec_time = 0;
-  {
+  // Run the test and return the execution time
+  auto measure_exec_time = [&]() -> double {
     xrt_core::runner runner(xrt::device(dev), recipe_data, profile_data, artifacts_repo);
     runner.execute();
     runner.wait();
     auto report = json::parse(runner.get_report());
-    baseline_exec_time = report["cpu"]["elapsed"].get<double>()/report["iterations"].get<int>();
-  }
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/query_requests.h
```diff
@@ -641,7 +639,13 @@ struct runner : request
     tct_one_column_path,
     tct_all_column_recipe,
     tct_all_column_profile,
-    tct_all_column_path
+    tct_all_column_path,
```
aktondak: This is not required anymore. You can go ahead and remove the whole runner structure itself if you like.

src/runtime_src/core/tools/common/tests/TestPreeemptionOverhead.cpp
```diff
@@ -7,123 +7,72 @@
...
-  catch (const std::exception& )
+  double baseline_exec_time = 0;
   {
-    throw xrt_core::error("Not enough columns available. Please make sure no other workload is running on the device.");
+    xrt_core::runner runner(xrt::device(dev), recipe_data, profile_data, artifacts_repo);
+    runner.execute();
+    runner.wait();
+    auto report = json::parse(runner.get_report());
+    baseline_exec_time = report["cpu"]["elapsed"].get<double>()/report["iterations"].get<int>();
   }
```
aktondak: Can you post the error you see without adding this scope here. @stsoe can help us explain the runner behavior in that case.

AShivangi: Yes,
So when I remove the scope from the runners, I see "Failed to escape (0x80000011): The requested resource is in use." error when the host code tries to run L44 (device_update). I think the device is not getting release after the runner is done because I reuse it after device_update.

stsoe: What scope are you talking about?

AShivangi: I mean placing the runner within the scope of L35–L41 so it releases the device before it’s used again in L44.

stsoe: Ah, I missed that.   Yeah, that scope and error makes no sense to me.
But the code should anyways be refactored so put the scope into its own function rather than duplicating the exact same code L47-L53

aktondak:
Let's also add the error handling for when archive pointer is null, similar to this PR : https://github.com/Xilinx/XRT/pull/9388.
Otherwise it seg faults for cases when archive is not present.

aktondak (approved):

PR #9409, opened 2025-11-05
Deterministic build for xrt-smi and xrt libaries

hyunjiki-amd:
xrt-smi.exe is reported to be not deterministic builds at link time.

To fix this issue, we need to add a linker flag to configure for deterministic builds.

Add /experimental:deterministic linker flag if MSVC. It will normalize variables such as order of input files, timestamps, et cetera.

None

Verified on Strix device by running fc /b <xrt-smi-1.exe> <xrt-smi-2.exe> command and returned no differences encountered

None

### DIFF
--- a/src/runtime_src/CMakeLists.txt
+++ b/src/runtime_src/CMakeLists.txt
@@ -70,7 +70,6 @@ else()
       string(CONCAT MP_OPTION "/MP" ${MSVC_PARALLEL_JOBS})
       add_compile_options(${MP_OPTION})
     endif()
-    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /experimental:deterministic")
   else()
     # lots of warnings and all warnings as errors
     # add_compile_options(-Wall -Wextra -pedantic -Werror)


hyunjiki-amd:
Verified the changes on Strix device by running `fc /b <xrt-smi-1.exe> <xrt-smi-2.exe>` and confirmed that they are identical.
But `fc /b <xrt_coreutil_1.dll> <xrt_coreutil_2.dll>` are not identical even after setting the linker flag - request for additional investigation.

Are there any other external dependencies that might have caused this non-deterministic behaviour? According to the Excel sheet attached in [EDGEML-11662](https://amd.atlassian.net/browse/EDGEML-11662), wbemuuid.lib (wbemcli_i.obj) is out of policy, which is from [XRT/src/runtime_src/core/common/detail/windows/sysinfo.h#L16](https://github.com/Xilinx/XRT/blob/808069aaf81d8fc45bbf3c80034c2a6e0e3f4ca2/src/runtime_src/core/common/detail/windows/sysinfo.h#L16).

src/runtime_src/CMakeLists.txt
```diff
@@ -70,6 +70,7 @@ else()
       string(CONCAT MP_OPTION "/MP" ${MSVC_PARALLEL_JOBS})
       add_compile_options(${MP_OPTION})
     endif()
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /experimental:deterministic")
```
stsoe: This make file is not specific to xrt-smi.  General windows options should be added under src/CMake/nativeWin.cmake.   But please explain what `experimenal:deterministic` is doing and why it is desired?

hyunjiki-amd: Hi @stsoe, thank you for the feedback, will move the linker flag to src/CMake/nativeWin.cmake.
This flag enables the deterministic builds (doesn't modify macros like timestamps or order of input files)

PR #9420, opened 2025-11-08
build.sh: Don't use -Werror as default.

lubosz:
While being a good practice to use during development, enabling `-Werror` by default for deployment and user builds is a guarantee for failed builds.

It is not feasible for development to keep up with all warnings in all compiler versions, especially ones from the future.

To avoid the build being broken, this patch sets `-Werror` being disabled as default behavior. This mimics the behavior of CMake being invoked directly without the build.sh script.

While a zero warning policy is good, and `-Werror` should also be enabled by default on the CI, it is a bad practice for deployment.

https://embeddedartistry.com/blog/2017/05/22/werror-is-not-your-friend/

Project fails building on newer compiler versions.

Providing a saner default.

The project will more likely build on newer compiler versions.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -57,7 +57,7 @@ usage()
     echo "[-opt]                      Build optimized library only (default)"
     echo "[-edge]                     Build edge of x64.  Turns off opt and dbg"
     echo "[-hip]                      Enable hip bindings"
-    echo "[-disable-werror]           Disable compilation with warnings as error"
+    echo "[-enable-werror]            Enable compilation with warnings as error"
     echo "[-nocmake]                  Skip CMake call"
     echo "[-noert]                    Do not treat missing ERT FW as a build error"
     echo "[-noinit]                   Do not initialize Git submodules"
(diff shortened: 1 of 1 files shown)


stsoe:
@lubosz build.sh is for development, developers are required to build with errors as warning.  While porting to new compilers, developers can disable `-Werror` as needed.   Why do we need to change this?

lubosz:
@stsoe Thank you for your response.

This is just a recommendation for the reasons listed above.

> build.sh is for development

According to the documentation, `build.sh` is the primary way of building XRT.
https://xilinx.github.io/XRT/master/html/build.html#building-the-xrt-runtime

This means that everyone following this documentation, may it be developer, distributor or advanced user, will face `-Werror` if they use a compiler version more recent than the version the XRT developers used at the time of the code release.

It's just a frustrating experience for on-boarding new developers.

In the end this MR is just to stimulate thoughts, I did not expect it to be merged. Maybe pointing the fact that `-Werror` is enabled by default and how to disable it in the documentation is another valid way to improve this.

I did my duty and started my contribution history to this project by fixing a warning that was treated as error, a review would be very appreciated:
https://github.com/Xilinx/XRT/pull/9419

stsoe:
Disabling `-Werror` by default is not desired.

PR #9421, opened 2025-11-09
Add checks while creating xrt::kernel to use proper ELF

rbramand-xilinx:
With the full ELF flow now enabled in XRT, users may inadvertently use incorrect APIs when creating xrt::kernel objects. This PRintroduces validation checks to detect and report improper API usage during xrt::kernel creation.

This is an enhancement aimed at improving Ease of Use (EOU). 

The correct usage of xrt::kernel creation APIs is as follows:
1. xrt::kernel(ctx, module, "kernel_name") — for xclbin + ELF flow
2. xrt::kernel(ctx, "kernel_name") — for full ELF flow

However, users might mistakenly pass an xrt::module created via the full ELF flow into the first API, or vice versa. This commit adds runtime checks to detect such misuse and throws appropriate exceptions to guide developers.

Low

Did negative testing to verify that exceptions are correctly thrown when APIs are misused.

NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -1657,6 +1657,56 @@ private:
     return std::make_unique<buffer_cache>(ctx, std::move(ctrlpkt_data), max_pool_size, xbi::use_type::ctrlpkt);
   }
 
+  // Function that checks if hw ctx is created using xcbin/ELF
+  // If is_elf_flow is true, then it checks if hw ctx is created using ELF
+  // and returns the hw ctx
+  // If is_elf_flow is false, then it checks if hw ctx is created using XCLBIN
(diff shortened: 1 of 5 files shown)


chvamshi-xilinx (approved):

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1657,6 +1657,56 @@ class kernel_impl : public std::enable_shared_from_this<kernel_impl>
     return std::make_unique<buffer_cache>(ctx, std::move(ctrlpkt_data), max_pool_size, xbi::use_type::ctrlpkt);
   }
 
+  // Function that checks if hw ctx is created using xcbin/ELF
```
larry9523, line 1660: xclbin

larry9523 (approved):

PR #9423, opened 2025-11-10
[XRT-SMI] Move event trace and firmware log features from report to stand-alone options

aktondak:
This PR moves the firmware log and event trace features away from being a report to being stand alone options with their own further sub-options. Reasons for this change : 
1. Both features do not fit the criteria for being stand-alone reports as they are dumping the changing contents of the DRAM buffer or running in watch mode. Reports are meant for a single snapshot of the state of hardware.
2. Both features have multiple sub-options which were being catered through --element tag but it is not scalable with more sub-options coming along (watch, raw, status etc.)
3. Help printing was not intuitive and elaborate and there were persistent queries from internal users on the usage of commands. This refactor simplifies the command and makes it more intuitive with native help printing support.
4. Now there is no need for unintuitive filler commands like "--element", "--action" or even "--run".

None

Handled via introducing OptionOptions to xrt-smi examine and moving watch, status and raw options to it.

None

Tested on Linux : 
Following updated commands works as intended : 
```
xrt-smi configure --advanced --firmware-log --enable --log-level 4
xrt-smi examine --advanced --firmware-log --watch
```

None

### DIFF
--- a/src/runtime_src/core/tools/xbutil2/OO_EventTraceExamine.h
+++ b/src/runtime_src/core/tools/xbutil2/OO_EventTraceExamine.h
@@ -23,7 +23,7 @@ private:
   bool m_watch;
   bool m_status;
   bool m_raw;
-  mutable uint32_t m_watch_mode_offset;
+  mutable uint64_t m_watch_mode_offset;
 
   // Log generation methods for examine functionality
   std::string 
--- a/src/runtime_src/core/tools/xbutil2/OO_FirmwareLogExamine.cpp
+++ b/src/runtime_src/core/tools/xbutil2/OO_FirmwareLogExamine.cpp
@@ -117,7 +117,7 @@ OO_FirmwareLogExamine::generate_raw_logs(const xrt_core::device* dev,
     size_t buf_size = data_buf.size;
 
     // Simply print the raw payload data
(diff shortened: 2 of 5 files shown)


src/runtime_src/core/tools/xbutil2/OO_EventTrace.cpp
```diff
@@ -18,32 +18,26 @@ namespace po = boost::program_options;
...
-    , m_action("")
+    , m_enable(false)
+    , m_disable(false)
     , m_help(false)
 {
   m_optionsDescription.add_options()
     ("device,d", boost::program_options::value<decltype(m_device)>(&m_device), "The Bus:Device.Function (e.g., 0000:d8:00.0) device of interest")
     ("help", boost::program_options::bool_switch(&m_help), "Help to use this sub-command")
+    ("enable", boost::program_options::bool_switch(&m_enable), "Enable event tracing")
+    ("disable", boost::program_options::bool_switch(&m_disable), "Disable event tracing")
```
AShivangi, line 29: Do you want to turn them into "action" like we do in force-preemption?

aktondak: No, I think --enable itself is more intuitive. --action just adds another keyword to the mix which is not required

PR #9443, opened 2025-11-18
Secure register/memory read and write

HimanshuChoudhary-Xilinx:
1. Passing PID and context-id to firmware for secure read and wite of register /memory.
2. AIE memory and register access operations in the XRT API layer (xrt_device.cpp) were directly calling ishim layer functions. PR adds query infrastructure. (aie_read/aie_write)
3. read_aie_mem, write_aie_mem, read_aie_reg, write_aie_reg fallback to older ishim layer call if Platform doesnt implement query aie_read or aie_write 
4. added PID as query arg in aie-coredump 

NA

Refactored AIE memory and register access to use the XRT query infrastructure

No public API signature changes
Fallback to old implementation ensures backward compatibility

calling device level apis read_aie_mem, write_aie_mem, read_aie_reg, write_aie_reg  from host application
NA

### DIFF
--- a/src/runtime_src/core/common/query_requests.h
+++ b/src/runtime_src/core/common/query_requests.h
@@ -4289,7 +4289,7 @@ struct aie_read : request
   using result_type = std::vector<char>;
   static const key_type key = key_type::aie_read;
 
-  virtual std::any
+  std::any
   get(const device*, const std::any&) const override = 0;
 };
 
@@ -4307,7 +4307,7 @@ struct aie_write : request
   using result_type = size_t;  // returns bytes written
   static const key_type key = key_type::aie_write;
 
-  virtual std::any
+  std::any
   get(const device*, const std::any&) const override = 0;
 };
 


src/runtime_src/core/common/api/xrt_device.cpp
```diff
@@ -466,10 +466,18 @@ read_aie_mem(pid_t pid, uint16_t context_id, uint16_t col, uint16_t row, uint32_
...
+        args.pid = static_cast<uint64_t>(pid);
+        args.context_id = context_id;
+        args.col = col;
+        args.row = row;
+        args.offset = offset;
+        args.size = size;
+
+        return xrt_core::device_query<xrt_core::query::aie_read>(get_handle().get(), args);
       }
       catch (const xrt_core::query::no_such_key&) {
```
chvamshi-xilinx, line 489: You have to have older implementation in catch path.

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -502,7 +502,8 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
   {
     try {
       return xrt_core::device_query<xrt_core::query::aie_coredump>(m_core_device,
-                                                                   m_hdl->get_slotidx());
+                                                                   m_hdl->get_slotidx(),
+                                                                   static_cast<uint64_t>(xrt_core::utils::get_pid()));
```
chvamshi-xilinx, line 506: fix the indentation

src/runtime_src/core/common/query_requests.h
```diff
@@ -348,6 +350,7 @@ enum class key_type
   xocl_errors_ex,
   xocl_ex_error_code2string,
 
+  aie_read,
```
chvamshi-xilinx, line 351: implement aie_write also

HimanshuChoudhary-Xilinx:
@stsoe  Please review this pr

src/runtime_src/core/common/query_requests.h
```diff
@@ -4273,14 +4275,50 @@ struct read_trace_data : request
...
+// Used for reading AIE memory of a specific tile within a ctx
+struct aie_read : request
+{
+  struct args {
+    uint64_t  pid;
+    uint16_t  context_id;
+    uint16_t  col;
+    uint16_t  row;
+    uint32_t  offset;
+    uint32_t  size;
```
stsoe, line 4287: initialize to default values (`= 0`)

src/runtime_src/core/common/query_requests.h
```diff
@@ -4273,14 +4275,50 @@ struct read_trace_data : request
...
+
+// Used for writing AIE memory/register of a specific tile within a ctx
+struct aie_write : request
+{
+  struct args {
+    uint64_t  pid;
+    uint16_t  context_id;
+    uint16_t  col;
+    uint16_t  row;
+    uint32_t  offset;
```
stsoe, line 4304: initialize to default values (`= 0`)

src/runtime_src/core/common/query_requests.h
```diff
@@ -4273,14 +4275,50 @@ struct read_trace_data : request
...
+
 // Used for getting AIE coredump of all tiles within a ctx
 struct aie_coredump : request
 {
   using result_type = std::vector<char>;
   static const key_type key = key_type::aie_coredump;
 
   std::any
-  get(const device*, const std::any&) const override = 0;
+  get(const device*, const std::any&, const std::any&) const override = 0;
```
diukirde, line 4321: @stsoe can we have both types of getters? windows does not need pid

chvamshi-xilinx: coreutil is always going to pass the PID, you can ignore it in the shim if you don't use it. As coreutil is always going to send query with PID, there is no point of having another getter.

diukirde: can we keep one parameter only? like pass it as a pair<ctx_id, pid> or ctx_id only

chvamshi-xilinx: Any issues with current interface? Current implementation & your proposal are same. I don't know if this is causing any issue in Windows :-) .
cc: @HimanshuChoudhary-Xilinx / @rbramand-xilinx ,

diukirde: Windows SHIM currently supports only single param for all device queries. With this implementation, we have to start supporting more than one param. I do not see the need for Windows to start supporting since we will only be using ctx_id.

chvamshi-xilinx: Ok. @HimanshuChoudhary-Xilinx , lets use structure similar to read/write mem/register instead of two arguments.

PR #9445, opened 2025-11-19
reporting correct bo size

bisingha-xilinx:
reporting correct bo size
The bug was if driver rounds up the bo size, userspace used to report the cached original requested bo size when host application requests with xrt::bo::size()
Problem was solved by updating the cached bo size with the actual bo size. Extra call is being made to get the actual size.
n/a
Tested by allocating a bo with less size and driver rounds up it 8K and checked if xrt::bo::size() reports the updated bo size. with this pr xrt::bo:;size() reports correct bo size.
n/a

### DIFF
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -1075,8 +1075,10 @@ alloc_kbuf(const device_type& device, size_t sz, xrtBufferFlags flags, xrtMemory
 {
   XRT_TRACE_POINT_SCOPE(xrt_bo_alloc_kbuf);
   auto handle = alloc_bo(device, sz, flags, grp);
-  auto boh = std::make_shared<xrt::buffer_kbuf>(device, std::move(handle), sz);
-  boh->get_usage_logger()->log_buffer_info_construct(device->get_device_id(), sz, device.get_hwctx_handle());
+  // Query actual allocated size from driver (bo size may be rounded up)
+  auto actual_sz = handle->get_properties().size;
+  auto boh = std::make_shared<xrt::buffer_kbuf>(device, std::move(handle), actual_sz);
+  boh->get_usage_logger()->log_buffer_info_construct(device->get_device_id(), actual_sz, device.get_hwctx_handle());
   return boh;
 }
 
--- a/src/runtime_src/core/edge/drm/zocl/common/zocl_bo.c
+++ b/src/runtime_src/core/edge/drm/zocl/common/zocl_bo.c
@@ -666,6 +666,7 @@ zocl_create_bo_ioctl(struct drm_device *dev, void *data, struct drm_file *filp)
 	 *       actual size allocated.
 	 */
 	zocl_update_mem_stat(zdev, bo->gem_base.size, 1, bo->mem_index);
+	args->size = bo->size;
 
 	return ret;
 }
--- a/src/runtime_src/core/edge/user/shim.cpp
+++ b/src/runtime_src/core/edge/user/shim.cpp
@@ -293,7 +293,7 @@ xclAllocBO(size_t size, unsigned flags, xrt_core::hwctx_handle* hwctx_hdl)
   if (result)
     throw std::bad_alloc();
 
-  xclLog(XRT_DEBUG, "%s: size %ld, flags 0x%x", __func__, size, flags);
+  xclLog(XRT_DEBUG, "%s: size %lld, flags 0x%x", __func__, info.size, flags);
   xclLog(XRT_INFO, "%s: ioctl return %d, bo handle %d", __func__, result, info.handle);
 
   return std::make_unique<buffer_object>(this, info.handle, hwctx_hdl);


src/runtime_src/core/common/api/xrt_bo.cpp
```diff
@@ -1075,8 +1075,10 @@ alloc_kbuf(const device_type& device, size_t sz, xrtBufferFlags flags, xrtMemory
 {
   XRT_TRACE_POINT_SCOPE(xrt_bo_alloc_kbuf);
   auto handle = alloc_bo(device, sz, flags, grp);
-  auto boh = std::make_shared<xrt::buffer_kbuf>(device, std::move(handle), sz);
-  boh->get_usage_logger()->log_buffer_info_construct(device->get_device_id(), sz, device.get_hwctx_handle());
+  // Query actual allocated size from driver (bo size may be rounded up)
+  auto actual_sz = handle->get_properties().size;
+  auto boh = std::make_shared<xrt::buffer_kbuf>(device, std::move(handle), actual_sz);
+  boh->get_usage_logger()->log_buffer_info_construct(device->get_device_id(), actual_sz, device.get_hwctx_handle());
```
stsoe, line 1081: I am not comfortable with this change.   The BO created by the user must reflect the size requested by the user, not the rounded up size.  The rounded up size is an implementation detail. 

Please revert.

chvamshi-xilinx:
Hi @stsoe ,
There is some confusion to the users. This has come up in below use case.
1) Application1 created a buffer with 2k. We are allocating 4k. bo.size returns 2k.
2) Application2 imported the same buffer using fd. This buffer size is returned as 4k as this under hood uses gem object created by Application1.

I understand your concern, but am not sure how can we fix above issue.
@maxzhen Any thoughts.

stsoe:
> Hi @stsoe , There is some confusion to the users. This has come up in below use case.
> 
> 1. Application1 created a buffer with 2k. We are allocating 4k. bo.size returns 2k.
> 2. Application2 imported the same buffer using fd. This buffer size is returned as 4k as this under hood uses gem object created by Application1.
> 
> I understand your concern, but am not sure how can we fix above issue. @maxzhen Any thoughts.

The imported buffer should reflect the size the user created the original buffer with.   This size must be stored in KMD apparently.   We cannot expose the internal buffer size, it would be even more confusing.

chvamshi-xilinx:
> > Hi @stsoe , There is some confusion to the users. This has come up in below use case.
> > 
> > 1. Application1 created a buffer with 2k. We are allocating 4k. bo.size returns 2k.
> > 2. Application2 imported the same buffer using fd. This buffer size is returned as 4k as this under hood uses gem object created by Application1.
> > 
> > I understand your concern, but am not sure how can we fix above issue. @maxzhen Any thoughts.
> 
> The imported buffer should reflect the size the user created the original buffer with. This size must be stored in KMD apparently. We cannot expose the internal buffer size, it would be even more confusing.

KMD storing doesn't solve the problem as this buffer can be imported from different driver. for example: zocl is exporting a buffer, and xdna driver is importing. While xdna is importing, it can only check the gem handles which has the updated size.

stsoe:
> > > Hi @stsoe , There is some confusion to the users. This has come up in below use case.
> > > 
> > > 1. Application1 created a buffer with 2k. We are allocating 4k. bo.size returns 2k.
> > > 2. Application2 imported the same buffer using fd. This buffer size is returned as 4k as this under hood uses gem object created by Application1.
> > > 
> > > I understand your concern, but am not sure how can we fix above issue. @maxzhen Any thoughts.
> > 
> > 
> > The imported buffer should reflect the size the user created the original buffer with. This size must be stored in KMD apparently. We cannot expose the internal buffer size, it would be even more confusing.
> 
> KMD storing doesn't solve the problem as this buffer can be imported from different driver. for example: zocl is exporting a buffer, and xdna driver is importing. While xdna is importing, it can only check the gem handles which has the updated size.

Got it.  I don't know then, but I am not good with the UMD change as proposed.

chvamshi-xilinx:
> > > > Hi @stsoe , There is some confusion to the users. This has come up in below use case.
> > > > 
> > > > 1. Application1 created a buffer with 2k. We are allocating 4k. bo.size returns 2k.
> > > > 2. Application2 imported the same buffer using fd. This buffer size is returned as 4k as this under hood uses gem object created by Application1.
> > > > 
> > > > I understand your concern, but am not sure how can we fix above issue. @maxzhen Any thoughts.
> > > 
> > > 
> > > The imported buffer should reflect the size the user created the original buffer with. This size must be stored in KMD apparently. We cannot expose the internal buffer size, it would be even more confusing.
> > 
> > 
> > KMD storing doesn't solve the problem as this buffer can be imported from different driver. for example: zocl is exporting a buffer, and xdna driver is importing. While xdna is importing, it can only check the gem handles which has the updated size.
> 
> Got it. I don't know then, but I am not good with the UMD change as proposed.

Thanks Soren. Let me check with filer. I will ask them whether there is any functional issues they are facing or just the confusion.
If thre are no functional issues, we can add a documentation over imported bo's

maxzhen:
> > > > > Hi @stsoe , There is some confusion to the users. This has come up in below use case.
> > > > > 
> > > > > 1. Application1 created a buffer with 2k. We are allocating 4k. bo.size returns 2k.
> > > > > 2. Application2 imported the same buffer using fd. This buffer size is returned as 4k as this under hood uses gem object created by Application1.
> > > > > 
> > > > > I understand your concern, but am not sure how can we fix above issue. @maxzhen Any thoughts.
> > > > 
> > > > 
> > > > The imported buffer should reflect the size the user created the original buffer with. This size must be stored in KMD apparently. We cannot expose the internal buffer size, it would be even more confusing.
> > > 
> > > 
> > > KMD storing doesn't solve the problem as this buffer can be imported from different driver. for example: zocl is exporting a buffer, and xdna driver is importing. While xdna is importing, it can only check the gem handles which has the updated size.
> > 
> > 
> > Got it. I don't know then, but I am not good with the UMD change as proposed.
> 
> Thanks Soren. Let me check with filer. I will ask them whether there is any functional issues they are facing or just the confusion. If thre are no functional issues, we can add a documentation over imported bo's

Yeah, the code that is importing the buffer don't see the very original size from the exporter. It just importing pages.

chvamshi-xilinx:
Closing for now. We need some other solution if filer is not convinced..

PR #9451, opened 2025-11-19
Add checks while creating xrt::kernel to use proper ELF

rbramand-xilinx:
Problem solved by the commit
With the full ELF flow now enabled in XRT, users may inadvertently use incorrect APIs when creating xrt::kernel objects. This PRintroduces validation checks to detect and report improper API usage during xrt::kernel creation.

This is an enhancement aimed at improving Ease of Use (EOU).

The correct usage of xrt::kernel creation APIs is as follows:

xrt::kernel(ctx, module, "kernel_name") — for xclbin + ELF flow
xrt::kernel(ctx, "kernel_name") — for full ELF flow
However, users might mistakenly pass an xrt::module created via the full ELF flow into the first API, or vice versa. This commit adds runtime checks to detect such misuse and throws appropriate exceptions to guide developers.

Low

Did negative testing to verify that exceptions are correctly thrown when APIs are misused.

Documentation impact (if any)
NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -1699,6 +1699,12 @@ private:
                                "xrt::module passed is not created using full ELF");
     }
     else {
+      // In xclbin only flow, module can be null
+      // return empty module in that case as other kernel_impl functions
+      // handle null module
+      if (!mod)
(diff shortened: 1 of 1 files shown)


rbramand-xilinx:
Opened new PR with same commit as earlier PR was landing on same machine for pcie hw tests.

PR #9442, opened 2025-11-16
Add Arch Linux support

kashif:
- Add Arch Linux package list to `xrtdeps.sh`
- Fix `Boost.System` compatibility for Boost >= 1.69 (header-only)
- Add pacman installation support
- Add pybind11 and HIP runtime installation for Arch Linux

#### Problem solved by the commit

This commit adds support for building XRT on Arch Linux, which was previously unsupported. The build system only supported Debian-based, RHEL-based, Fedora, SUSE, and Mariner distributions.

Additionally, it fixes Boost compatibility issues with newer Boost versions (>= 1.69) where `Boost.System` became header-only but the build system was still attempting to find it as a compiled component.

#### How problem was solved, alternative solutions (if any) and why they were rejected
1. Added `arch_package_list()` to `xrtdeps.sh` with Arch-specific package names
  2. Modified `boostUtil.cmake` to conditionally include system component only for Boost < 1.69 when using CMake > 3.29
  3. Added Arch-specific pybind11 installation using pacman instead of pip
  4. Added HIP runtime installation for Arch via pacman
  
#### What has been tested and how, request additional testing if necessary

  Tested successfully:
  - Platform: Arch Linux (kernel 6.17.8-arch1-1)
  - Hardware: AMD Ryzen AI 9 HX 370 w/ NPU Strix (Radeon 890M)
  - Boost version: 1.89.0
  - CMake version: 4.1.2
  - GCC version: 15.2.1

PR #9461, opened 2025-11-24
cmake/boostUtil: Don't look for boost system stub library.

lubosz:
Boost system has been header only since 1.69 (from 2018) and the stub library has been dropped in 1.89, resulting in CMake errors when being looked for.

This patch drops the redundant find_package component to boost_system, which was only a stub library since boost version 1.69 from 2018.

See https://github.com/boostorg/system/blob/1061db7dcd534bf93ea10558165115382fe3b87c/doc/system/changes.adoc#changes-in-boost-189

This fixes the following CMake error when bulding with boost 1.89 or later:
```
CMake Error at /usr/lib/cmake/Boost-1.89.0/BoostConfig.cmake:141 (find_package):
  Could not find a package configuration file provided by "boost_system"
  (requested version 1.89.0) with any of the following names:

    boost_systemConfig.cmake
    boost_system-config.cmake

  Add the installation prefix of "boost_system" to CMAKE_PREFIX_PATH or set
  "boost_system_DIR" to a directory containing one of the above files.  If
  "boost_system" provides a separate development package or SDK, be sure it
  has been installed.
```

Redundancy in the build system and build breakage in the current boost release.

Following boost recommendation: Removing the redundant library.

This assumes that XRT is only built with versions 1.69 of boost or higher. If versions below 1.69 must be supported, the find command will need to list system as optional dependency (As seen in the official documentation).

Tested building with boost 1.89. It should still build with all boost version since 2018, but the CI should maybe test it.

### DIFF
--- a/src/CMake/boostUtil.cmake
+++ b/src/CMake/boostUtil.cmake
@@ -5,29 +5,17 @@
 # In particular the script runtime_src/tools/script/boost.sh downloads
 # and builds static Boost libraries compiled with fPIC so that they
 # can be used to resolve symbols in XRT dynamic libraries.
-
-# Note: Boost.System became header-only in Boost 1.69, so we only need
-# filesystem and program_options as compiled components for newer Boost versions
-set(BOOST_REQUIRED_COMPONENTS filesystem program_options)
-
 if (DEFINED ENV{XRT_BOOST_INSTALL})
   set(XRT_BOOST_INSTALL $ENV{XRT_BOOST_INSTALL})
   set(Boost_USE_STATIC_LIBS ON)
   if(CMAKE_VERSION VERSION_GREATER "3.29")
-    # First try to find Boost to get version
-    find_package(Boost CONFIG
-      HINTS $ENV{XRT_BOOST_INSTALL})
-    # For older Boost (< 1.69), also include system component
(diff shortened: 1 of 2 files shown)


stsoe:
@lubosz, please sign-off the commits.  Thanks for fixing this.

lubosz:
> @lubosz, please sign-off the commits. Thanks for fixing this.

Done. Sorry I previously I only set `commit.gpgsign=true`, e.g. "verified" status. I know also did the `--signoff` on `git commit`.

lubosz:
I see this was already partially addressed in a bit less elegant way by #9442

lubosz:
I have updated my patch, rebasing on top of #9442.

Following improvements are made in addition:
* My patch also addresses this issue in `src/runtime_src/core/tools/xbflash2/CMakeLists.txt`, which I needed to do for a successful build.
* My patch also drops linking the stub library on CMake versions below 3.29
* I am proposing my previous solution to the problem by dropping support for boost versions below 1.69, which were released in 2018. This simplifies the build code by a lot. If you prefer keeping support for old boost versions I can implement the [suggested solution](https://github.com/boostorg/system/blob/1061db7dcd534bf93ea10558165115382fe3b87c/doc/system/changes.adoc#changes-in-boost-189) by the boost project and add `system` to `OPTIONAL_COMPONENTS`. But I would consider this as unnecessary.

PR #9465, opened 2025-11-25
AIESW-14034 - Dump log buffer periodically

rbramand-xilinx:
This PR adds new buffer dumper that parses metadata, dumps the buffer periodically
This is done according to spec in page - https://amd.atlassian.net/wiki/spaces/AIE/pages/1126711923/MDS+SWV+Debug+Features+for+Bringup#CERT-log-improvement

Its an enhancement over existing log buffer dumping

Added a new class buffer_dumps that takes in metadata and dumps data periodically, asynchronously (by launching a thread)
At present the data is dumped to file only.
Dumping to syslog and console is not added yet because the data in the buffer is encoded and without decoding the data, dumping to syslog/console doesn't make sense.
TODO : Add support to dump to syslog and console after adding the decoder logic
Also tweak the dumper interval and chunk size as per experiments with different model test cases.

Low

Tested by dumping uC log on Telluride and the feature works as expected

NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -83,7 +83,7 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
       // parameters for uc log buffer dumper
       // tweak dump interval, size_per_uc based on experiments
       constexpr size_t size_per_uc = 2_mb;
-      constexpr size_t dump_interval_ms = 3;
+      constexpr size_t dump_interval_ms = 50;
       constexpr size_t metadata_size = 32;
       constexpr size_t count_offset = 0;
       constexpr size_t count_size = 8;
@@ -280,13 +280,13 @@ public:
     m_uc_log_buf.reset();
 
     try {
-      // finish_flush_device should only be called when the underlying 
+      // finish_flush_device should only be called when the underlying
       // hw_context_impl is destroyed. The xdp::update_device cannot exist
       // in the hw_context_impl constructor because an existing
       // shared pointer must already exist to call get_shared_ptr(),
       // which is not true at that time.
       xrt_core::xdp::finish_flush_device(this);
-      
+
       // Reset within scope of dtor for trace point to measure time to reset
       m_hdl.reset();
     }
@@ -531,7 +531,7 @@ xrt::hw_context
 create_hw_context_from_implementation(void* hwctx_impl)
 {
   if (!hwctx_impl)
-    throw std::runtime_error("Invalid hardware context implementation."); 
+    throw std::runtime_error("Invalid hardware context implementation.");
 
   auto impl_ptr = static_cast<xrt::hw_context_impl*>(hwctx_impl);
   return xrt::hw_context(impl_ptr->get_shared_ptr());
@@ -585,7 +585,7 @@ post_alloc_hwctx(const std::shared_ptr<hw_context_impl>& handle)
 {
   // Update device is called with a raw pointer to dyanamically
   // link to callbacks that exist in XDP via a C-style interface
-  // The create_hw_context_from_implementation function is then 
+  // The create_hw_context_from_implementation function is then
   // called in XDP create a hw_context to the underlying implementation
   xrt_core::xdp::update_device(handle.get(), true);
   handle->get_usage_logger()->log_hw_ctx_info(handle.get());
--- a/src/runtime_src/core/common/buffer_dumper.cpp
+++ b/src/runtime_src/core/common/buffer_dumper.cpp
@@ -124,17 +124,15 @@ dump_chunk_data(size_t chunk_index, size_t start, size_t length, uint8_t* chunk)
 
 void
 buffer_dumper::
-process_chunks()
+process_chunks_no_lock()
 {
-  std::lock_guard lock(m_dump_mutex);
-
   // Map buffer once for all chunks
-  auto* base_ptr = m_config.dump_buffer.map<uint8_t*>();
+  auto base_ptr = m_config.dump_buffer.map<uint8_t*>();
 
   for (size_t i = 0; i < m_config.num_chunks; i++)
   {
     const size_t chunk_offset = i * m_config.chunk_size;
-    uint8_t* chunk = base_ptr + chunk_offset;
+    auto chunk = base_ptr + chunk_offset;
 
     // sync only the metadata for the current chunk to read the logged count
     m_config.dump_buffer.sync(XCL_BO_SYNC_BO_FROM_DEVICE, m_config.metadata_size, chunk_offset);
@@ -157,7 +155,8 @@ process_chunks()
       if (to_dump <= bytes_to_end) {
         // Data doesn't wrap, sync contiguous range
         m_config.dump_buffer.sync(XCL_BO_SYNC_BO_FROM_DEVICE, to_dump, chunk_offset + start_offset);
-      } else {
+      }
+      else {
         // Data wraps around, sync two ranges
         m_config.dump_buffer.sync(XCL_BO_SYNC_BO_FROM_DEVICE, bytes_to_end, chunk_offset + start_offset);
         m_config.dump_buffer.sync(XCL_BO_SYNC_BO_FROM_DEVICE, to_dump - bytes_to_end, chunk_offset + m_config.metadata_size);
@@ -171,18 +170,24 @@ process_chunks()
 
 void
 buffer_dumper::
+process_chunks()
+{
+  std::lock_guard lock(m_dump_mutex);
+  process_chunks_no_lock();
+}
+
+void
+buffer_dumper::
 dumping_loop()
 {
   while (!m_stop_thread) {
     try {
-      {
-        std::unique_lock lock(m_dump_mutex);
-        m_cv.wait_for(lock, std::chrono::milliseconds(m_config.dump_interval_ms),
-                      [this] { return m_stop_thread.load(); });
-      }
+      std::unique_lock lock(m_dump_mutex);
+      m_cv.wait_for(lock, std::chrono::milliseconds(m_config.dump_interval_ms),
+                    [this] { return m_stop_thread.load(); });
 
       if (!m_stop_thread)
-        process_chunks();
+        process_chunks_no_lock();
     }
     catch (const std::exception& e) {
       // Log error but keep thread running
--- a/src/runtime_src/core/common/buffer_dumper.h
+++ b/src/runtime_src/core/common/buffer_dumper.h
@@ -58,22 +58,33 @@ private:
   std::condition_variable m_cv;
   std::vector<std::ofstream> m_file_streams;
 
+  // Read the logged count from chunk metadata
   size_t
   read_logged_count(uint8_t* chunk);
 
+  // Write chunk data to file (metadata header + new data payload)
   void
   dump_chunk_data(size_t chunk_index, size_t start, size_t length, uint8_t* chunk);
 
+  // Process all chunks without acquiring lock (caller must hold m_dump_mutex)
+  void
+  process_chunks_no_lock();
+
+  // Process all chunks with lock acquisition
   void
   process_chunks();
 
+  // Background thread function that periodically checks for new data
   void
   dumping_loop();
 
 public:
+  // Reads configuration and opens files for dumping
+  // Starts background dumping thread
   explicit
   buffer_dumper(config cfg);
 
+  // Stop background thread and flush remaining data
   ~buffer_dumper();
 
   // Delete copy and move operations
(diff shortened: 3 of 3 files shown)


src/runtime_src/core/common/buffer_dumper.h
```diff
@@ -0,0 +1,88 @@
...
+{
+  size_t chunk_size = 0;
+  size_t metadata_size = 0;
+  size_t count_offset = 0;
+  size_t count_size = 0;
+  size_t num_chunks = 0;
+  size_t dump_interval_ms = 0;
+  std::string dump_file_prefix;
+  xrt::bo dump_buffer;
+};
```
stsoe: Embed in class buffer_dumper and explain the fields.

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+void
+buffer_dumper::
+flush()
+{
+  // process chunks to dump the remaining data
+  process_chunks();
+}
+
+} // xrt_core
```
stsoe: Missing newline

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -0,0 +1,219 @@
...
+}
+
+void
+buffer_dumper::
+dumping_loop()
+{
+  while (!m_stop_thread) {
+    try {
+      {
+        std::unique_lock<std::mutex> lock(m_dump_mutex);
```
stsoe:
```suggestion
        std::unique_lock lock(m_dump_mutex);
```

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -0,0 +1,219 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+#define XCL_DRIVER_DLL_EXPORT  // in same dll as exported xrt apis
+#define XRT_API_SOURCE         // in same dll as coreutil
+#define XRT_CORE_COMMON_SOURCE // in same dll as coreutil
```
stsoe: Not needed, nothing exported

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+#include "buffer_dumper.h"
+
+#include "core/common/message.h"
+#include "core/common/time.h"
+#include "core/common/utils.h"
+#include "core/include/xrt/deprecated/xrt.h"
```
stsoe: Why this one?

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -0,0 +1,219 @@
...
+  catch (const std::exception& e) {
+    xrt_core::message::send(xrt_core::message::severity_level::debug, "buffer_dumper",
+        std::string{"Failed to flush dump buffer: "} + e.what());
+  }
+
+  for (auto& fs : m_file_streams) {
+    if (fs.is_open()) {
+      fs.flush();
+      fs.close();
+    }
```
stsoe: I doubt this is necessary.  The streams are automatic and flushed and closed when going out of scope.

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -0,0 +1,219 @@
...
+  m_dump_thread = std::thread(&buffer_dumper::dumping_loop, this);
+}
+
+buffer_dumper::
+~buffer_dumper()
+{
+  m_stop_thread = true;
+  m_cv.notify_one();
+
+  if (m_dump_thread.joinable()) {
```
stsoe: What makes the threads non joinable?

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -0,0 +1,219 @@
...
+    m_dump_thread.join();
+  }
+
+  // Flush the remaining data
+  // catch exceptions to avoid throwing in destructor
```
stsoe, line 42: join() can also throw, move try/catch out of this block

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -68,22 +70,26 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
+    std::unique_ptr<xrt_core::buffer_dumper> m_uc_log_bo_dumper;
+
+    static std::unique_ptr<xrt_core::buffer_dumper>
+    init_uc_log_bo_dumper(const std::shared_ptr<xrt_core::device>& device,
+                          xrt_core::hwctx_handle* ctx_hdl,
+                          size_t num_uc)
     {
+      // parameters for uc log buffer dumper
+      // tweak dump interval, size_per_uc based on experiments
+      constexpr size_t size_per_uc = 2 * 1024 * 1024;
```
stsoe: I'd prefer `2_mb` with 
```
constexpr std::size_t 
operator""_mb(unsigned long long v)  { return 1024u * 1024u * v; }
```

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -68,22 +70,26 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
+      constexpr size_t dump_interval_ms = 3;
```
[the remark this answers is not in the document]
rbramand-xilinx: 32 bytes metadata size is proper, out of which 8 bytes are used at present and rest is for future use.

rbramand-xilinx:
> Please move class buffer_dumper into the unnamed namespace on xrt_hw_context.cpp; it need not be separated into a .h and .cpp file.

Hi Soren, buffer_dumper class at present is used only with uc_log_buffer but in future it will be used with other kind of buffers as well. So created separate .h and .cpp files for this class.

src/runtime_src/core/common/buffer_dumper.cpp
rbramand-xilinx: yes Soren, they are redundant. I will remove these lines. Thanks

src/runtime_src/core/common/buffer_dumper.cpp
rbramand-xilinx: yes this is also redundant, I will remove it

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+    // sync only the metadata for the current chunk to read the logged count
+    m_config.dump_buffer.sync(XCL_BO_SYNC_BO_FROM_DEVICE, m_config.metadata_size, chunk_offset);
+
+    size_t logged_count = read_logged_count(chunk);
+    size_t& dumped_count = m_dumped_counts[i];
+    size_t logged_wrap = logged_count / m_data_size;
+    size_t dumped_wrap = dumped_count / m_data_size;
+
+    if (logged_count > dumped_count && logged_wrap > dumped_wrap)
```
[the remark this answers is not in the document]
rbramand-xilinx: No we need both the checks because if logged_count is less than dumped_count and logged_wrap > dumped_wrap that is not overwriting, we still have valid data

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -96,53 +105,35 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
+
+      // create buffer dumper object to dump the log buffer contents
+      xrt_core::buffer_dumper::config config;
+      config.chunk_size = size_per_uc;
+      config.metadata_size = metadata_size;
+      config.count_offset = count_offset;
+      config.count_size = count_size;
+      config.num_chunks = num_uc;
+      config.dump_interval_ms = dump_interval_ms;
+      config.dump_file_prefix = "uc_log_" + std::to_string(ctx_hdl->get_slotidx());
```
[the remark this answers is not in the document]
rbramand-xilinx: It is expected because we can only pass prefix and adding timestamp in prefix doesn't makes sense

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -68,22 +73,26 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
+
+    static std::unique_ptr<xrt_core::buffer_dumper>
+    init_uc_log_bo_dumper(const std::shared_ptr<xrt_core::device>& device,
+                          xrt_core::hwctx_handle* ctx_hdl,
+                          size_t num_uc)
     {
+      // parameters for uc log buffer dumper
+      // tweak dump interval, size_per_uc based on experiments
+      constexpr size_t size_per_uc = 2_mb;
+      constexpr size_t dump_interval_ms = 3;
```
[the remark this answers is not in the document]
rbramand-xilinx: we dont expect user to pass these values and these values can be changed for each buffer_dumper object during construction, so this should be okay. But that said we need to do experiments to decide on correct set of values

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -0,0 +1,203 @@
...
+  while (!m_stop_thread) {
+    try {
+      {
+        std::unique_lock lock(m_dump_mutex);
+        m_cv.wait_for(lock, std::chrono::milliseconds(m_config.dump_interval_ms),
+                      [this] { return m_stop_thread.load(); });
+      }
+
+      if (!m_stop_thread)
+        process_chunks();
```
stsoe: I see that `process_chunks()` acquires `m_dump_mutex`, I would remove the scope around the unique_lock here, rename private `process_chunks()` to `procecss_chunks_no_lock()`, and remove the lock_guard from `process_chunks_no_lock()`.   I would then change `flush()` to acquire the lock and call `process_chunks_no_lock()`, or I would add `process_chunks()` to acquire the lock and call `process_chunks_no_lock()`.

src/runtime_src/core/common/buffer_dumper.h
```diff
@@ -0,0 +1,91 @@
...
+
+  void
+  process_chunks();
+
+  void
+  dumping_loop();
+
+public:
+  explicit
+  buffer_dumper(config cfg);
```
stsoe, line 85: comment that public API.  The ctor starts the dumping loop thread, opens some files ....  The dtor stops the thread and joins with it.
flush() does ...

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -0,0 +1,203 @@
...
+}
+
+void
+buffer_dumper::
+process_chunks()
+{
+  std::lock_guard lock(m_dump_mutex);
+
+  // Map buffer once for all chunks
+  auto* base_ptr = m_config.dump_buffer.map<uint8_t*>();
```
stsoe:
```suggestion
  auto base_ptr = m_config.dump_buffer.map<uint8_t*>();
```

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+  for (size_t i = 0; i < m_config.num_chunks; i++)
+  {
+    const size_t chunk_offset = i * m_config.chunk_size;
+    uint8_t* chunk = base_ptr + chunk_offset;
```
stsoe:
```suggestion
    auto chunk = base_ptr + chunk_offset;
```

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -0,0 +1,203 @@
...
+    if (dumped_count != logged_count) {
+      size_t to_dump = logged_count - dumped_count;
+      const size_t start_offset = (dumped_count % m_data_size) + m_config.metadata_size;
+      const size_t bytes_to_end = m_config.chunk_size - start_offset;
+
+      // Sync only the data range we need to dump
+      if (to_dump <= bytes_to_end) {
+        // Data doesn't wrap, sync contiguous range
+        m_config.dump_buffer.sync(XCL_BO_SYNC_BO_FROM_DEVICE, to_dump, chunk_offset + start_offset);
+      } else {
```
stsoe:
```suggestion
      } 
      else {
```

src/runtime_src/core/common/buffer_dumper.cpp
rbramand-xilinx: refactored accordingly Soren, Thanks

src/runtime_src/core/common/api/xrt_hw_context.cpp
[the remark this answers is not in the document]
rbramand-xilinx: Changing this to 50 ms

chvamshi-xilinx (approved):

stsoe (approved):

PR #9477, opened 2025-11-28
fix Linux kernel version check condition

firas-aschkar:
1. set Linux kernel version check condition to greater or equal, instead of greater only, thus fixing compile error when building against Linux-6.1.0, and
2. remove trailing space


maxzhen:
ok to test

stsoe:
@firas-aschkar  I had to rebase your PR to get past CI problems. 
Please sign-off the commit (git commit --signoff)

PR #9492, opened 2025-12-06
fix: Remove boost system dependencies 

X-Ryl669:
Since it's a deprecated empty shim that was removed in newer boost packaging.

Current master doesn't build on Arch linux and latest kernel because of the Boost System dependency package in CMake.
Boost system is an empty package, there's no sources or headers in it and it's only present for a small period so it doesn't break legacy code depending on it. Many rolling release distributions have removed it so it doesn't build if you rely on it.  

I haven't created an issue for it, but will do if required.

I've remove the dependency since it was useless anyway.

On very very old system, where Boost System was not empty, this could break. However, it's unlikely such system would build the current driver due to their aging kernel.

It fixes a build issue, no actual code was modified, so it's unlikely to break the driver features


stsoe:
Multiple other PRs related to Arch has be merged since this PR was created.

PR #9540, opened 2026-01-13
AIESW-21287 add new qos parameter sleep_to_limit_fps

mkorhone:

Add new qos parameter `sleep_to_limit_fps` to profile.json that can be used to simulate a model running at a fixed frame rate

N/A

Alternative solution is to use the existing `sleep` parameter in the `iteration` node in `executions`.  This sleep duration is fixed, but since the sleep amount varies based on what else is running, it seems less brittle to automatically compute the sleep duration.

None unless this new parameter is used.  Otherwise, we should consider this parameter experimental.

Ran full model tests both with and without this new parameter

Potentially could document this new parameter

### DIFF
--- a/src/runtime_src/core/common/runner/profile.md
+++ b/src/runtime_src/core/common/runner/profile.md
@@ -38,13 +38,29 @@ how many times along with what should be done in each iteration.
 This section is optional.
 
 Simple key/value pairs designating configuration parameters for
-hardware context creation. 
+hardware context creation and execution control.
 ```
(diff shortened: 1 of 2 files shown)


mkorhone:
This PR should be considered experimental and useful to start a discussion on the value of this feature.  There are a few TODOs documented in the code changes that should be addressed prior to considering merging this.  In the meantime, this PR can be used internally as needed by developers that benefit from this new feature.

PR #9587, opened 2026-02-05
[XRT-SMI] Cmd-chain test rename

aktondak:
This test renames the command chain latency test to runlist latency and throughput since that more accurately describes what is being tested

None

Solved via renaming. This will require name changes in the driver code as well when the xrt submodule is updated

None

None

None


rchane (approved):

sonals:
Let's do a full analysis on impact to backwards compatibility. Will the current scripts for parsing console and JSON reports work without change?

PR #9643, opened 2026-03-04
Enable strong volatile ordering

hyunjiki-amd:
Enable strong volatile ordering

`/volatile:ms` switch enables Microsoft extended volatile semantics that guarantee strong ordering

Could add as AdditionalOptions in vcxproj files

None

Confirmed that the compile option is successfully set by building locally

None


yanyu-amd:
Hi @hyunjiki-amd 

Let's try turn on warning-c4746 to see if we have any compiler volatile switch dependent volatile usage in the code.

https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-c4746?view=msvc-170

stsoe (commented):
It's my understand this option is not recommended.  Can you explain the reasoning behind adding this option?

PR #9658, opened 2026-03-10
Improve xrt-smi bash and csh completion scripts

AngryLoki:

Modern bash versions use autoloading for completion scripts, so "Autocomplete enabled for the xrt-smi command" in the middle of completion did not look well and was removed.

Other changes:
- Update command set to configure/examine/validate (remove program/reset)
- Add device BDF completion by reading /sys/class/accel/ symlinks
- Restructure bash completion to handle global options before commands
- Add terminal option handling (--version, --help) to prevent further completions

N/A / discovered in Gentoo [ebuild](https://github.com/gentoo/guru/blob/dev/dev-util/xrt/xrt-999999.ebuild#L155), which installs bash completion file to a proper place.

Rewrite bash completion and update csh competition to match bash functionality.

Low

Tested with csh-6.24.16 and bash-5.3.9

N/A


sonals:
ok to test

AngryLoki:
(rebased and added add "advanced" options)

PR #9717, opened 2026-04-08
Add hwctx ctors for default access mode

stsoe:
Per request, default access mode to shared for elf ctors just like was done for xclbin ctors.


rbramand-xilinx (approved):
Looks good.

PR #9719, opened 2026-04-08
Ensure is_done is set for unmanaged commands

richard-byron:
- Fixes hitting assert in run_impl destructor when unmanaged have actually finished but their state has not been updated
- Resolves potential deadlock by removing lock from is_done.  The lock here should not be necessary, and is_done is called from set_dtrace_control_file while it already holds the mutex
- Fixes potential race conditions with m_callbacks being accessed outside of locks, which could theoretically results in callbacks being called multiple times, or not at all.
  
Introduced in https://github.com/Xilinx/XRT/pull/9649

There are simpler solutions.  Just calling get_state() before is_done() works, as long as you can convince the compiler it actually needs to update m_done.  That would leave the other issues unresolved, but I'm not aware of them actually causing problems.

Might change behavior with multiple threads waiting on runs to complete.  

Tested on several designs, including ones that were hitting the assert and ones that weren't.  More testing would definitely be desirable.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -694,14 +694,9 @@ class kernel_command : public xrt_core::command
 public:
   using execbuf_type = xrt_core::bo_cache::cmd_bo<ert_start_kernel_cmd>;
   using callback_function_type = std::function<void(ert_cmd_state)>;
+  using callback_list = std::vector<callback_function_type>;
 
 private:
-  struct callback_properties {
-    callback_function_type  fn;
-    mutable bool            called;
-  };
-  using callback_list = std::vector<callback_properties>;
-
   // Return state of underlying exec buffer packet This is an
   // asynchronous call, the command object may not be in the same
   // state as reflected by the return value.
@@ -720,7 +715,7 @@ public:
     , m_hwqueue(std::move(hwqueue))
     , m_hwctx(std::move(hwctx))
     , m_execbuf(m_device->create_exec_buf<ert_start_kernel_cmd>())
-    , m_done(true)
+    , m_run_state(run_state::ready)
   {
     static unsigned int count = 0;
     m_uid = count++;
@@ -763,10 +758,7 @@ public:
   bool
   is_done() const
   {
-    if (!m_done && !m_managed) {
-      m_done = get_state() >= ERT_CMD_STATE_COMPLETED;
-    }
-    return m_done;
+    return (m_run_state != run_state::running) || (!m_managed && get_state() >= ERT_CMD_STATE_COMPLETED);
   }
 
   // Return state of command object.  The underlying packet
@@ -779,7 +771,9 @@ public:
     // is a no-op on platforms where command state is live.
     m_hwqueue.poll(this);
 
-    return get_state_raw();
+    auto state = get_state_raw();
+    notify(state);  // update command state accordingly
+    return state;
   }
 
   // Return kernel return code from command object for PS kernels
@@ -817,27 +811,52 @@ public:
     ert_cmd_state state = ERT_CMD_STATE_MAX;
     {
       std::lock_guard<std::mutex> lk(m_mutex);
-      if (!m_managed && !m_done)
+      if (!m_managed && m_run_state != run_state::ready)
         throw xrt_core::error(ENOTSUP, "Cannot add callback to running unmanaged command");
+      m_callbacks.emplace_back(std::move(fcn));
       auto pkt = get_ert_packet();
       state = static_cast<ert_cmd_state>(pkt->state);
-      complete = m_done && state >= ERT_CMD_STATE_COMPLETED;
-      m_callbacks.emplace_back(callback_properties({std::move(fcn), complete}));
+      complete = (m_run_state == run_state::ready) && (state >= ERT_CMD_STATE_COMPLETED);
     }
 
     // lock must not be held while calling callback function
     if (complete)
-      fcn(state);
+      m_callbacks.back()(state);
   }
 
   // Remove last added callback
   void
   pop_callback()
   {
-    std::lock_guard<std::mutex> lk(m_mutex);
-    if (!m_callbacks.empty()) {
+    if (!m_callbacks.empty())
       m_callbacks.pop_back();
+  }
+
+  // Run registered callbacks.
+  void
+  run_callbacks(ert_cmd_state state) const
+  {
+    std::vector<const callback_function_type*> copy;
+
+    {
+      std::lock_guard<std::mutex> lk(m_mutex);
+
+      if (m_callbacks.empty())
+        return;
+    
+
+      // cannot lock mutex while calling the callbacks
+      // so copy address of callbacks while holding the lock
+      // then execute callbacks without lock
+      copy.reserve(m_callbacks.size());
+
+      std::transform(m_callbacks.begin(),m_callbacks.end()
+                     ,std::back_inserter(copy)
+                     ,[](const callback_function_type& cb) { return &cb; });
     }
+
+    for (auto cb : copy)
+      (*cb)(state);
   }
 
   // Submit the command for execution.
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -758,7 +763,9 @@ class kernel_command : public xrt_core::command
   bool
   is_done() const
   {
-    std::lock_guard<std::mutex> lk(m_mutex);
+    if (!m_done && !m_managed) {
+      m_done = get_state() >= ERT_CMD_STATE_COMPLETED;
```
stsoe: I am not comfortable updating `m_done` outside the lock.   I could be in process of being update by `run()`

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -772,9 +779,7 @@ class kernel_command : public xrt_core::command
     // is a no-op on platforms where command state is live.
     m_hwqueue.poll(this);
 
-    auto state = get_state_raw();
-    notify(state);  // update command state accordingly
-    return state;
+    return get_state_raw();
```
stsoe: This is behavior change, if get_state() reflects completed, `m_done` can still be false.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -814,52 +819,25 @@ class kernel_command : public xrt_core::command
...
       state = static_cast<ert_cmd_state>(pkt->state);
       complete = m_done && state >= ERT_CMD_STATE_COMPLETED;
+      m_callbacks.emplace_back(callback_properties({std::move(fcn), complete}));
     }
 
-    // lock must not be helt while calling callback function
+    // lock must not be held while calling callback function
     if (complete)
-      m_callbacks.get()->back()(state);
+      fcn(state);
```
stsoe: Calling moved fcn ?

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -694,9 +694,14 @@ class kernel_command : public xrt_core::command
 public:
   using execbuf_type = xrt_core::bo_cache::cmd_bo<ert_start_kernel_cmd>;
   using callback_function_type = std::function<void(ert_cmd_state)>;
-  using callback_list = std::vector<callback_function_type>;
 
 private:
+  struct callback_properties {
+    callback_function_type  fn;
+    mutable bool            called;
```
stsoe: I am not following why `called` is necessary?

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -954,27 +942,36 @@ class kernel_command : public xrt_core::command
...
+      // cannot lock mutex while calling the callbacks
+      // so copy address of callbacks while holding the lock
+      // then execute callbacks without lock
+      if (!m_callbacks.empty()) {
+        copy.reserve(m_callbacks.size());
+        for (auto& cb : m_callbacks) {
+          if (!cb.called) {
+            cb.called = true;
+            copy.emplace_back(cb.fn);
+          }
```
stsoe: I don't see what was wrong with earlier logic, nor do I see the need for `called`?

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -998,7 +995,7 @@ class kernel_command : public xrt_core::command
   mutable std::mutex m_mutex;
   mutable std::condition_variable m_exec_done;
 
-  std::unique_ptr<callback_list> m_callbacks;
+  callback_list m_callbacks; // don't see any reason this was a pointer?
```
stsoe: Agree !

stsoe (changes requested):
Hi @richard-byron.  Thanks for the suggested changes. 

If run_impl dtor complains, then that is because wait2() was called as expected.  The execution model requires wait to be called, but the dtor could potentially call existing get_state() to update m_done before complaining or aborting. 

I see the deadlock you mention, thanks!, but that should be fixed without the changes in this PR.

I also see the potential race in m_callbacks, but that I think is only one place missing a lock, e.g. pop_callback()?

I am not exited about these changes because I can think in particular get_state() results in potential problems when it no longer calls notify().  I added few specific comments.  

I would like if we could fix just the deadlock, and the unnecessary pointer to callback vector.  I do not want to work-around missing calls to functions that update m_done, (wait2() or get_state()).

richard-byron:
I think the minimal working change is:
 - remove the lock in is_done() (without the other change here.  The read of m_done is atomic anyway, so this is not required and avoids the deadlock)
 - set m_done in wait() when m_managed is false (as in the PR currently)

I wasn't aware wait was required to be called, or I probably would have just done that

You could just add get_state() in the destructor instead, but that was getting optimized out when I tried it, unless I printed the return value.

stsoe:
> I think the minimal working change is:
> 
> * remove the lock in is_done() (without the other change here.  The read of m_done is atomic anyway, so this is not required and avoids the deadlock)

`m_done` is not atomic, maybe it should be.  As is, it should not be updated outside the lock.

> * set m_done in wait() when m_managed is false (as in the PR currently)

You mean as the PR does in `is_done()` ?  With m_done changed to atomic and with the PR change in is_done(), the call to wait prior to run_impl  dtor becomes optional.  This sounds okay to me.

richard-byron:
What I mean when I say m_done is atomic is that the read from memory should already be atomic (unless it was using 64 bits for a bool on a 32 bit processor, or something crazy like that).  The lock doesn't actually do anything.
I don't think m_done needs to be atomic, but it probably makes things clearer if it is.

is_done() could then become:
`m_done.compare_exchange_strong(false, get_state_raw() >= ERT_CMD_STATE_COMPLETED); return m_done;`
but my question with that is, do we need to call m_hwqueue.poll() first?  In that case get_state() seems to be the better call, but we probably don't want is_done() triggering a call to notify() and potentially firing off callbacks?

I did also think about `is_done(bool updateState = false)` and having the destructor call `is_done(true);` but wasn't sure I liked that either.

As long as it's ok to require wait to be called, I think the cleanest solution is just to add:
`m_done.compare_exchange_strong(false, get_state_raw() >= ERT_CMD_STATE_COMPLETED);`

to the end of both kernel_command wait() functions (and remove the lock from is_done).

stsoe:
Without lock (or atomic) how does one guarantee memory ordering or that writers change become visible to readers?

yeah, poll and notify need to be called.  I don't want this over engineered.  Let's change m_done to atomic.  Let is_done call get_state() with no changes, such that notify is called also without any changes.  Something like this (and atomic) being the only change.   

```
  bool
  is_done() const
  {
    if (!m_done && !m_managed)
      return (get_state() >= ERT_CMD_STATE_COMPLETED);

    return m_done;
  }
```

richard-byron (commented):
The problem with calling get_state() inside is_done() is that we might end up trying to take the lock twice.  Avoiding that requires making additional changes, but I think this implementation avoids all threading issues without changing behavior.

stsoe (commented):
The execution model requires that applications call xrt::run::wait2() to ensure completion is reflected in the state of the run implementation.  I feel there are too many unrelated changes in this PR, maybe they work, but I cannot easily understand why they are necessary.

There is a fix required to resolve the dtrace logging deadlock you brought to our attention.  We will ticket this and resolve differently.

richard-byron:
Calling wait2 does not update m_done, so the assert in the destructor can still fire spuriously.  If calling wait2 is a requirement though, then you're correct that there is a simpler fix - just update m_done in wait/wait2.

stsoe:
> Calling wait2 does not update m_done, so the assert in the destructor can still fire spuriously. If calling wait2 is a requirement though, then you're correct that there is a simpler fix - just update m_done in wait/wait2.

It's rather convoluted, I agree.   But wait2() definitely ends up reflecting completion with m_done set to true.  If a command is "managed" then notify() is called with state change when command monitor observes the command completing. If command is "unmanaged" then m_hwqueue.wait() calls notify().  In both cases m_done reflects completion.

richard-byron:
I'm fairly certain m_done was not being set for "unmanaged" commands that had actually completed, even after waiting.  The graph I was testing certainly had issues though - so possibly there was some error causing that.  Happy to just close this in any case.

PR #9721, opened 2026-04-10
Run constructor Run wait and run start hook for temporal sharing usecase

garimadhaked:
Fix : Added XDP dtrace plugin support for temporal sharing usecase.
- Added run constructor, run wait and run start hooks
- updated xdp submodule
- added dtrace plugin related settings in xrt.ini
- Use of aiebu api to get info from elf for dtrace

Testing: Verified few examples with and without XDP dtrace plugin. Verified temporal sharing example as well.

### DIFF
--- a/src/runtime_src/core/common/api/CMakeLists.txt
+++ b/src/runtime_src/core/common/api/CMakeLists.txt
@@ -29,11 +29,3 @@ target_include_directories(core_common_api_library_objects
   ${XRT_SOURCE_DIR}/runtime_src
   ${XRT_SOURCE_DIR}/runtime_src/core/common/elf
   )
-
-# Compile definitions for VE2 XDNA builds to enable run hooks
-if (XDP_VE2_BUILD_CMAKE STREQUAL "yes")
-  target_compile_definitions(core_common_api_library_objects
-    PRIVATE
-    XDP_VE2_BUILD=1
-    )
-endif()
--- a/src/runtime_src/core/common/api/CMakeLists.txt
+++ b/src/runtime_src/core/common/api/CMakeLists.txt
@@ -29,3 +29,11 @@ target_include_directories(core_common_api_library_objects
   ${XRT_SOURCE_DIR}/runtime_src
   ${XRT_SOURCE_DIR}/runtime_src/core/common/elf
   )
+
+# Compile definitions for VE2 XDNA builds to enable run hooks
+if (XDP_VE2_BUILD_CMAKE STREQUAL "yes")
+  target_compile_definitions(core_common_api_library_objects
+    PRIVATE
+    XDP_VE2_BUILD=1
+    )
+endif()
--- a/src/runtime_src/core/common/api/kernel_int.h
+++ b/src/runtime_src/core/common/api/kernel_int.h
@@ -24,6 +24,10 @@
 #include <string>
 #include <vector>
 
+namespace xrt_core::xdp {
+struct xrt_kernel_data;
+}
+
 namespace xrt_core { namespace kernel_int {
 
 // Provide access to kdma command based BO copy Used by xrt::bo::copy.
@@ -97,26 +101,10 @@ get_hw_ctx(const xrt::kernel& kernel);
 xrt::kernel
 create_kernel_from_implementation(const xrt::kernel_impl* kernel_impl);
 
-// Run-level accessors for XDP dtrace hooks
+// Fill XDP kernel data from a run object for profiling hooks
 XRT_CORE_COMMON_EXPORT
-uint32_t
-get_run_uid(const xrt::run& run);
-
-XRT_CORE_COMMON_EXPORT
-xrt::hw_context
-get_run_hwctx(const xrt::run& run);
-
-XRT_CORE_COMMON_EXPORT
-std::string
-get_run_kernel_name(const xrt::run& run);
-
-XRT_CORE_COMMON_EXPORT
-xrt::module
-get_run_module(const xrt::run& run);
-
-XRT_CORE_COMMON_EXPORT
-ert_cmd_state
-get_run_state(const xrt::run& run);
+void
+get_xdp_kernel_data(const xrt::run& run, xrt_core::xdp::xrt_kernel_data* data);
 
 }} // kernel_int, xrt_core
 
--- a/src/runtime_src/core/common/api/kernel_int.h
+++ b/src/runtime_src/core/common/api/kernel_int.h
@@ -106,6 +106,12 @@ XRT_CORE_COMMON_EXPORT
 void
 get_xdp_kernel_data(const xrt::run& run, xrt_core::xdp::xrt_kernel_data* data);
 
+// Set dtrace control file on a run_impl handle
+// This is used by XDP profiling to set the CT file without requiring xrt::run
+XRT_CORE_COMMON_EXPORT
+void
+set_dtrace_control_file(xrt::run_impl* run_impl, const std::string& path);
+
 }} // kernel_int, xrt_core
 
 #endif
--- a/src/runtime_src/core/common/api/kernel_int.h
+++ b/src/runtime_src/core/common/api/kernel_int.h
@@ -9,15 +9,11 @@
 
 // This file defines implementation extensions to the XRT Kernel APIs.
 #include "core/include/xrt/experimental/xrt_kernel.h"
-#include "core/include/xrt/experimental/xrt_module.h"
 #include "core/include/xrt/experimental/xrt_xclbin.h"
-#include "core/include/xrt/xrt_hw_context.h"
 
 #include "core/common/config.h"
 #include "core/common/xclbin_parser.h"
 #include "core/common/shim/buffer_handle.h"
-
-#include "ert.h"
 
 #include <bitset>
 #include <cstdint>
--- a/src/runtime_src/core/common/api/kernel_int.h
+++ b/src/runtime_src/core/common/api/kernel_int.h
@@ -101,10 +101,10 @@ get_hw_ctx(const xrt::kernel& kernel);
 xrt::kernel
 create_kernel_from_implementation(const xrt::kernel_impl* kernel_impl);
 
-// Fill XDP kernel data from a run object for profiling hooks
+// Fill XDP kernel data from a run_impl pointer for profiling hooks
 XRT_CORE_COMMON_EXPORT
 void
-get_xdp_kernel_data(const xrt::run& run, xrt_core::xdp::xrt_kernel_data* data);
+get_xdp_kernel_data(const xrt::run_impl* run_impl, xrt_core::xdp::xrt_kernel_data* data);
 
 // Set dtrace control file on a run_impl handle
 // This is used by XDP profiling to set the CT file without requiring xrt::run
--- a/src/runtime_src/core/common/api/kernel_int.h
+++ b/src/runtime_src/core/common/api/kernel_int.h
@@ -9,14 +9,19 @@
 
 // This file defines implementation extensions to the XRT Kernel APIs.
 #include "core/include/xrt/experimental/xrt_kernel.h"
+#include "core/include/xrt/experimental/xrt_module.h"
 #include "core/include/xrt/experimental/xrt_xclbin.h"
+#include "core/include/xrt/xrt_hw_context.h"
 
 #include "core/common/config.h"
 #include "core/common/xclbin_parser.h"
 #include "core/common/shim/buffer_handle.h"
 
+#include "ert.h"
+
 #include <bitset>
 #include <cstdint>
+#include <string>
 #include <vector>
 
 namespace xrt_core { namespace kernel_int {
@@ -91,6 +96,27 @@ get_hw_ctx(const xrt::kernel& kernel);
 // This is used for logging usage mertrics
 xrt::kernel
 create_kernel_from_implementation(const xrt::kernel_impl* kernel_impl);
+
+// Run-level accessors for XDP dtrace hooks
+XRT_CORE_COMMON_EXPORT
+uint32_t
+get_run_uid(const xrt::run& run);
+
+XRT_CORE_COMMON_EXPORT
+xrt::hw_context
+get_run_hwctx(const xrt::run& run);
+
+XRT_CORE_COMMON_EXPORT
+std::string
+get_run_kernel_name(const xrt::run& run);
+
+XRT_CORE_COMMON_EXPORT
+xrt::module
+get_run_module(const xrt::run& run);
+
+XRT_CORE_COMMON_EXPORT
+ert_cmd_state
+get_run_state(const xrt::run& run);
 
 }} // kernel_int, xrt_core
 
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -2480,7 +2480,7 @@ public:
     XRT_DEBUGF("run_impl::run_impl(%d)\n" , uid);
 
     // XDP run_constructor hook
-    xrt_core::xdp::run_constructor(this, get_xdp_kernel_data());
+    xrt_core::xdp::run_constructor(this);
   }
 
   // Clones a run impl, so that the clone can be executed concurrently
@@ -2794,7 +2794,7 @@ public:
     prep_start();
 
     // XDP profiling hook - called immediately before run is submitted
-    xrt_core::xdp::run_start(get_xdp_kernel_data());
+    xrt_core::xdp::run_start(this);
 
     // log kernel start info
     // This is in critical path, we need to reduce log overhead
@@ -2898,9 +2898,7 @@ public:
     }
 
     // XDP profiling hook - called after wait completes
-    auto xdp_data = get_xdp_kernel_data();
-    xdp_data.ert_state = static_cast<int>(state);
-    xrt_core::xdp::run_wait(xdp_data);
+    xrt_core::xdp::run_wait(this);
 
     dump_logs(true); // dump required logs
 
@@ -2955,9 +2953,7 @@ public:
     }
 
     // XDP profiling hook - called after wait completes
-    auto xdp_data = get_xdp_kernel_data();
-    xdp_data.ert_state = static_cast<int>(state);
-    xrt_core::xdp::run_wait(xdp_data);
+    xrt_core::xdp::run_wait(this);
 
     // dump required logs
     dump_logs(true);
@@ -4238,19 +4234,20 @@ create_kernel_from_implementation(const xrt::kernel_impl* kernel_impl)
 }
 
 void
-get_xdp_kernel_data(const xrt::run& run, xrt_core::xdp::xrt_kernel_data* data)
+get_xdp_kernel_data(const xrt::run_impl* run_impl, xrt_core::xdp::xrt_kernel_data* data)
 {
-  auto run_handle = run.get_handle();
-  auto kernel = run_handle->get_kernel();
-  data->uid = run_handle->get_uid();
+  auto kernel = run_impl->get_kernel();
+  data->uid = run_impl->get_uid();
   data->name = kernel->get_name();
   data->hwctx = kernel->get_hw_context();
   data->mod = kernel->get_module();
+  data->ert_state = static_cast<int>(run_impl->state());
 }
 
 void
 set_dtrace_control_file(xrt::run_impl* run_impl, const std::string& path)
 {
+  XRT_TRACE_POINT_SCOPE(set_dtrace_control_file); 
   run_impl->set_dtrace_control_file(path);
 }
 
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -47,6 +47,7 @@
 #include "core/common/trace.h"
 #include "core/common/usage_metrics.h"
 #include "core/common/xclbin_parser.h"
+#include "core/common/xdp/profile.h"
 
 #include <boost/format.hpp>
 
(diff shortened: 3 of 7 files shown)


src/runtime_src/core/common/xdp/profile.cpp
```diff
@@ -145,6 +145,82 @@ end_poll(void* handle)
 
 } // end namespace xrt_core::xdp::aie::profile
 
+namespace xrt_core::xdp::aie::dtrace {
+
+std::function<void(void*, bool)> update_device_cb;
```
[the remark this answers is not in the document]
garimadhaked: This is std::function object that gets assigned at runtime via register_callback, so it cannot be made const because it needs to be modified after initialization.

src/runtime_src/core/common/api/CMakeLists.txt
```diff
@@ -29,3 +29,11 @@ target_include_directories(core_common_api_library_objects
...
   ${XRT_SOURCE_DIR}/runtime_src/core/common/elf
   )
+
+# Compile definitions for VE2 XDNA builds to enable run hooks
+if (XDP_VE2_BUILD_CMAKE STREQUAL "yes")
+  target_compile_definitions(core_common_api_library_objects
+    PRIVATE
+    XDP_VE2_BUILD=1
+    )
+endif()
```
stsoe: Isn't this a define for just one TU?   Scope it to where needed, please.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4222,6 +4217,23 @@ create_kernel_from_implementation(const xrt::kernel_impl* kernel_impl)
...
+make_run_info(void* run, void* hwctx_handle, uint32_t run_uid,
+              const char* kernel_name, void* elf_handle = nullptr, int ert_cmd_state = 0)
+{
+  xrt_core::xdp::run_info info;
+  info.run = run;
+  info.hwctx_handle = hwctx_handle;
+  info.run_uid = run_uid;
+  info.kernel_name = kernel_name;
+  info.elf_handle = elf_handle;
+  info.ert_cmd_state = ert_cmd_state;
```
stsoe: Explain what this is.   What, why, how.  This function doesn't appear to belong in xrt_kernel.cpp, but I don't know, because I am not sure what the intent is?

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4232,7 +4244,16 @@ run::
...
            ("xrt::run::run", alloc_run, krnl.get_handle()))
-{}
+{
+  if (xrt_core::config::get_aie_dtrace()) {
+    auto hwctx = krnl.get_handle()->get_hw_context();
+    const auto& mod = krnl.get_handle()->get_module();
+    auto elf_hdl = mod ? xrt_core::module_int::get_elf_handle(mod) : nullptr;
+    auto info = make_run_info(this, hwctx.get_handle().get(), handle->get_uid(),
+                              krnl.get_handle()->get_name().c_str(), elf_hdl.get());
+    xrt_core::xdp::run_constructor(info);
```
stsoe: This code does not belong here in the ctor.  Maybe in alloc_run, but again I have no idea what is done here or why?

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4246,6 +4267,12 @@ run::
 start()
 {
   XRT_TRACE_POINT_SCOPE(xrt_run_start);
+  if (xrt_core::config::get_aie_dtrace()) {
+    auto hwctx = handle->get_kernel()->get_hw_context();
+    auto info = make_run_info(this, hwctx.get_handle().get(), handle->get_uid(),
+                              handle->get_kernel()->get_name().c_str());
+    xrt_core::xdp::run_start(info);
+  }
```
stsoe: Can this code go into xdp rather than be here?
It would be best if we can move XDP specifics to xdp.   Sure, you can make a conditional  call into xdp form any function, but you should probably move what is done here into xdp?

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4278,21 +4305,37 @@ run::
...
+  auto state = xdp::native::profiling_wrapper("xrt::run::wait",
     [this, &timeout_ms] {
       return handle->wait(timeout_ms);
     });
+  if (xrt_core::config::get_aie_dtrace()) {
+    auto hwctx = handle->get_kernel()->get_hw_context();
+    auto info = make_run_info(const_cast<xrt::run*>(this), hwctx.get_handle().get(),
+                              handle->get_uid(), handle->get_kernel()->get_name().c_str(),
+                              nullptr, static_cast<int>(state));
+    xrt_core::xdp::run_wait(info);
```
stsoe: Ditto here.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4278,21 +4305,37 @@ run::
...
+  auto cvst = xdp::native::profiling_wrapper("xrt::run::wait",
     [this, &timeout_ms] {
       return handle->wait_throw_on_error(timeout_ms);
     });
+  if (xrt_core::config::get_aie_dtrace()) {
+    auto hwctx = handle->get_kernel()->get_hw_context();
+    auto info = make_run_info(const_cast<xrt::run*>(this), hwctx.get_handle().get(),
+                              handle->get_uid(), handle->get_kernel()->get_name().c_str(),
+                              nullptr, static_cast<int>(handle->state()));
+    xrt_core::xdp::run_wait(info);
```
stsoe: I see the same code over and over again, this does not look right.

stsoe (changes requested):

src/runtime_src/core/common/api/xrt_kernel.cpp
garimadhaked: I have added hooks in run constructor, run start and run wait so that xdp can do configuration. 
Today xdp does configuration during hw context constructor but that is not valid for temporal sharing usecase.
So added these hooks at run start/wait and run constructor.

src/runtime_src/core/common/api/xrt_kernel.cpp
garimadhaked: Earlier added xrt_core::xdp::run_constructor(this, hwctx.get_handle().get(), handle->get_uid(),
                              krnl.get_handle()->get_name().c_str(), elf_hdl.get())

Vamshi suggested to have a separate structure for the arguments and just pass the structure to these xdp hooks so moved this to xdp/profile.h

src/runtime_src/core/common/api/CMakeLists.txt
garimadhaked: I missed removing the compile time control from cmake. Removed it.

src/runtime_src/core/common/api/xrt_kernel.cpp
garimadhaked: We have to add XDP hooks for run constructor start and wait to support temporal sharing for aie trace/profile and ddr bandwidth controlled with dynamic trace. so added these hooks everywhere.

src/runtime_src/core/common/api/xrt_kernel.cpp
garimadhaked: moved make_run_info inside xdp.

stsoe: Not clean enough.  Can we not move all changes to xdp and do what you like there. E.g. xrt_core::xdp::run_start(*this).

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4278,21 +4287,37 @@ run::
...
+  auto state = xdp::native::profiling_wrapper("xrt::run::wait",
     [this, &timeout_ms] {
       return handle->wait(timeout_ms);
     });
+  if (xrt_core::config::get_aie_dtrace()) {
+    auto hwctx = handle->get_kernel()->get_hw_context();
+    auto info = xrt_core::xdp::make_run_info(const_cast<xrt::run*>(this), hwctx.get_handle().get(),
+                                             handle->get_uid(), handle->get_kernel()->get_name().c_str(),
+                                             nullptr, static_cast<int>(state));
+    xrt_core::xdp::run_wait(info);
```
stsoe: Ditto.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4278,21 +4287,37 @@ run::
...
+  auto cvst = xdp::native::profiling_wrapper("xrt::run::wait",
     [this, &timeout_ms] {
       return handle->wait_throw_on_error(timeout_ms);
     });
+  if (xrt_core::config::get_aie_dtrace()) {
+    auto hwctx = handle->get_kernel()->get_hw_context();
+    auto info = xrt_core::xdp::make_run_info(const_cast<xrt::run*>(this), hwctx.get_handle().get(),
+                                             handle->get_uid(), handle->get_kernel()->get_name().c_str(),
+                                             nullptr, static_cast<int>(handle->state()));
+    xrt_core::xdp::run_wait(info);
```
stsoe: Ditto

src/runtime_src/core/common/api/xrt_kernel.cpp
stsoe: Not clean enough. Can we not move all changes to xdp and do what you like there. E.g. xrt_core::xdp::run_ctor(*this).

stsoe (changes requested):

src/runtime_src/core/common/api/xrt_kernel.cpp
garimadhaked: I have made the changes but run_impl and kernel_impl are defined in xrt_kernel.cpp, not in a header. So I can't directly access them from xdp/profile.cpp.

So I added accessor functions to kernel_int.h for the run-level data needed by dtrace plugin.

stsoe (changes requested):
Thanks, looks better.  But please don't change the function bodies of xrt::run APIs, please move your code to the implementation (alloc_run, run_imp::start(), and so forth.).

stsoe (commented):
I am still suggesting changes here.
I would like `xdp/profile.h`, which is included in `xrt_kernel.cpp`, to define a struct like this
```
namespace <what-ever-your-namespace-is> {
struct xrt_kernel_data
{
   uint32_t uid;
   std::string name;
   xrt::hw_context hwctx;
   xrt::module mod;
};
}
```
This struct is visible to `xrt_kernel.cpp` because it includes `profile.h`

In `kernel_int.h`, forward declare this struct and declare a function to fill it:
```
namespace <what-ever-your-namespace-is> {
struct xrt_kernel_data; // forward decl
}

namespace xrt_core::kernel_int {
...
void
get_xdp_kernel_data(xrt_kernel_data*);
} 
```

In xrt_kernel.cpp, define the function:
```
namespace xrt_cpre::kernel_int {
...
void
get_xdp_kernel_data(xrt_kernel_data* data)
{
   data->uid = ...;
   data->name = ...;
   data->hwctx = ...;
   data->mod = ...;
}
```

This accomplishes the same as you have now, but creates only one new internal function, which by its name is a little more specific.   I realize that this may be very close to the make_info() function you had earlier, but this data exchange is simpler and much cleaner. 

We still need the API level functions cleaned up and moved appropriately into impl functions.

garimadhaked (commented):
I have modified the code as per suggestion did move the hooks to implementation. But for run constructor hook I could not move because the set_dtrace_control_file() is xrt::run obejct method.

src/runtime_src/core/common/xdp/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ add_library(core_common_xdp_profile_objects OBJECT
 target_include_directories(core_common_xdp_profile_objects
   PRIVATE
   ${XRT_SOURCE_DIR}/runtime_src
+  ${XRT_SOURCE_DIR}/runtime_src/core/common/elf
```
stsoe, line 10: Why?

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4232,7 +4256,10 @@ run::
 run(const kernel& krnl)
   : handle(xdp::native::profiling_wrapper
            ("xrt::run::run", alloc_run, krnl.get_handle()))
-{}
+{
+  // XDP run_constructor hook called here where xrt::run object is available
+  xrt_core::xdp::run_constructor(*this);
```
stsoe: I would prefer no constructor body here.  I think the xdp hook should be moved to the run_impl constructor because you definitely don't care about mailbox impl. 

Also, why can't profile.cpp operate at the handle level like xrt::xrt_hw_context_impl does?  The run_impl class is derived from enable_shared_from_this, so you can easily go from a raw run_impl* (this) to std::share_ptr<run_impl> if you need to participate in ownership, or even an xrt::run if you really had to.  

It sounds like you need to call  run::set_dtrace_control_file(), which dispatches to run_impl::set_dtrace_control_file(), but this latter function you can call from a run_impl* provided you add a kernel_int internal function. 

I am looking for some consistency here.  I think calling xrt_core::xdp with raw handles or typed raw handles is better than requiring xrt::run objects.

src/runtime_src/core/common/xdp/CMakeLists.txt
garimadhaked: profile.cpp includes core/common/api/module_int.h
module_int.h includes elf_int.h
elf_int.h includes <elfio/elfio.hpp>
The ELFIO library headers are located at core/common/elf/elfio/

module_int.h is included specifically for get_elf_handle(). Which is used by xdp to get op_location from the elf.

src/runtime_src/core/common/api/xrt_kernel.cpp
garimadhaked: I have done the required changes. Please do not merge the changes, With this change I had to update xdp code as well. Until XDP PR is merged this PR can not be merged.

    - Move run_constructor XDP hook from xrt::run constructor to run_impl constructor
    - Change run_constructor signature to take run_impl* and xrt_kernel_data
      instead of xrt::run&
    - Add kernel_int::set_dtrace_control_file() internal function to allow
      XDP plugins to set dtrace control file using run_impl* handle
    
    This change enables the XDP plugin to call set_dtrace_control_file via
    kernel_int internal API without needing to cast to xrt::run.

garimadhaked:
Updated xdp submodule to pick latest changes. PR can be merged now.
Verified the feature.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -2471,6 +2478,9 @@ class run_impl : public std::enable_shared_from_this<run_impl>
     , uid(create_uid())
   {
     XRT_DEBUGF("run_impl::run_impl(%d)\n" , uid);
+
+    // XDP run_constructor hook
+    xrt_core::xdp::run_constructor(this, get_xdp_kernel_data());
```
stsoe, line 2483: With my suggestion above, this becomes:
`xrt_core::xdp::run_constructor(this);`

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -2783,6 +2793,9 @@ class run_impl : public std::enable_shared_from_this<run_impl>
 
     prep_start();
 
+    // XDP profiling hook - called immediately before run is submitted
+    xrt_core::xdp::run_start(get_xdp_kernel_data());
```
stsoe, line 2797:
```suggestion
    xrt_core::xdp::run_start(this);
```

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -2884,6 +2897,11 @@ class run_impl : public std::enable_shared_from_this<run_impl>
       state = cmd->wait();
     }
 
+    // XDP profiling hook - called after wait completes
+    auto xdp_data = get_xdp_kernel_data();
+    xdp_data.ert_state = static_cast<int>(state);
+    xrt_core::xdp::run_wait(xdp_data);
```
stsoe, line 2903:
```suggestion
    xrt_core::xdp::run_wait(this);
```

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -2936,6 +2954,11 @@ class run_impl : public std::enable_shared_from_this<run_impl>
...
```
stsoe, line 2960:
```suggestion
    xrt_core::xdp::run_wait(this);
```

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4220,8 +4237,24 @@ create_kernel_from_implementation(const xrt::kernel_impl* kernel_impl)
...
+void
+get_xdp_kernel_data(const xrt::run& run, xrt_core::xdp::xrt_kernel_data* data)
+{
+  auto run_handle = run.get_handle();
+  auto kernel = run_handle->get_kernel();
+  data->uid = run_handle->get_uid();
+  data->name = kernel->get_name();
+  data->hwctx = kernel->get_hw_context();
+  data->mod = kernel->get_module();
+}
```
stsoe, line 4249:
```suggestion
void
get_xdp_kernel_data(const run_impl* rimpl, xrt_core::xdp::xrt_kernel_data* data)
{
  auto kernel = run_impl->get_kernel();
  data->uid = run_impl->get_uid();
  data->name = kernel->get_name();
  data->hwctx = kernel->get_hw_context();
  data->mod = kernel->get_module();
  data->ert_state = run_impl->state();
}
```

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4246,6 +4279,7 @@ run::
 start()
 {
   XRT_TRACE_POINT_SCOPE(xrt_run_start);
+  // XDP run_start hook is called in run_impl::start()
```
stsoe, line 4282:
```suggestion
```

src/runtime_src/core/common/xdp/profile.cpp
```diff
@@ -145,6 +149,81 @@ end_poll(void* handle)
...
+  if (end_poll_cb)
+    end_poll_cb(handle);
+}
+
+void
+run_constructor(void* run, void* hwctx_handle, uint32_t run_uid, const char* kernel_name, void* elf_handle)
+{
+  if (run_constructor_cb)
+    run_constructor_cb(run, hwctx_handle, run_uid, kernel_name, elf_handle);
+}
```
stsoe, line 209:
```suggestion
void
run_constructor(const run_impl* rimpl)
{
  if (run_constructor_cb) {
    xrt_kernel_data xkd {};
    get_xdp_kernel_data(rimpl, &xdk);
    run_constructor_cb(run, xkd.what_you_need, xdk.what_you_need2, ...);
  }   
}
```

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -2401,6 +2402,12 @@ class run_impl : public std::enable_shared_from_this<run_impl>
   // Run-level dtrace result file postfix
   std::string m_dtrace_result_file_postfix;
 
+  xrt_core::xdp::xrt_kernel_data
+  get_xdp_kernel_data() const
+  {
+    return {uid, kernel->get_name(), kernel->get_hw_context(), m_module};
+  }
+
```
stsoe, line 2410: This function is not free, it creates a string, adds reference count under spin lock, etc. There is no point calling this function when the xdp callback flow is not enabled.

Please change all your callbacks to conditionally (only when _cb functions are indeed initialized) call xrt_core::kernel_int::get_xdp_kernel_data(...).  

Change the signature of that function to take a run_impl* rather than the void*, then it can easily call get_xdp_kernel_data(rimpl, &kdata).    I was not earlier implying going straight to void* in your callbacks, in fact it is better to stay typed as long as it makes sense.  What I was implying earlier was that you wanted to pass the handle pointer from the impl classes rather than passing the API exposed xrt::run., this you have done, only lost the type :-).

The point here is also that none of the XRT code should extract data on behalf of XDP requirements, it must be the other way around where XDP extracts from XRT when and only when XDP is enabled.

src/runtime_src/core/common/api/kernel_int.h
```diff
@@ -87,15 +96,22 @@ get_regmap_size(const xrt::kernel& kernel);
...
-
 // Allows the creation of the kernel object from a kernel_impl pointer
 // This is used for logging usage mertrics
 xrt::kernel
 create_kernel_from_implementation(const xrt::kernel_impl* kernel_impl);
 
+// Fill XDP kernel data from a run object for profiling hooks
+XRT_CORE_COMMON_EXPORT
+void
+get_xdp_kernel_data(const xrt::run& run, xrt_core::xdp::xrt_kernel_data* data);
```
stsoe, line 107: Review this change in context of other comments.

```suggestion
XRT_CORE_COMMON_EXPORT
void
get_xdp_kernel_data(const xrt::run_impl*, xrt_core::xdp::xrt_kernel_data*);
```

stsoe (changes requested):
Still not quite there, some relative minor adjustments.

jvillarre (approved):

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4220,8 +4227,25 @@ create_kernel_from_implementation(const xrt::kernel_impl* kernel_impl)
...
+  data->name = kernel->get_name();
+  data->hwctx = kernel->get_hw_context();
+  data->mod = kernel->get_module();
+  data->ert_state = static_cast<int>(run_impl->state());
+}
+
+void
+set_dtrace_control_file(xrt::run_impl* run_impl, const std::string& path)
+{
+  XRT_TRACE_POINT_SCOPE(set_dtrace_control_file);
```
```suggestion
```
Let's not trace internal APIs and certainly doesn't make sense for XDP.

stsoe (approved):
Looks great now.  Thanks!
One nit, please remove the TRACE call from kernel_int::set_dtrace...().

PR #9723, opened 2026-04-13
AIESW-26391:Fix eff_net hip testcase failure.

svup-xilinx:
HIP kernels using pointer arithmetic for sub-buffer addresses (e.g., address + offset) fail because extracting the buffer object returns only the base address, losing the offset calculation. All computed addresses collapse to the same base value, causing kernels to access wrong memory.
[AIESW-26391](https://jira.xilinx.com/browse/AIESW-26391)
HIP global arguments using pointer arithmetic (base + offset) incorrectly receive only the base address, causing kernels to access wrong memory locations.
Use get_hip_mem_from_addr() to extract both the base buffer object and offset. When offset is non-zero (pointer arithmetic detected), pass the raw pointer to preserve the exact computed address. When offset is zero (base buffer), pass the buffer object with proper binding for memory management.
Low
tested with test case added in
https://gitenterprise.xilinx.com/XRT/Telluride/pull/743
NA

### DIFF
--- a/src/runtime_src/hip/core/event.cpp
+++ b/src/runtime_src/hip/core/event.cpp
@@ -255,14 +255,29 @@ kernel_start::kernel_start(std::shared_ptr<function> f, void** args)
         break;
 
       case karg::argtype::global: {
+        // Handle both base buffer addresses and computed addresses (base + offset).
+        // If offset is non-zero, carve out a sub-buffer and pass it as BO arg.
         void **bufptr = static_cast<void **>(args[idx]);
-        auto hip_mem = memory_database::instance().get_hip_mem_from_addr(*bufptr).first;
+        auto hip_mem_info = memory_database::instance().get_hip_mem_from_addr(*bufptr);
+        auto hip_mem = hip_mem_info.first;
         if (!hip_mem) {
-            std::string err_msg = "failed to get memory from arg at index - " + std::to_string(idx);
-	    throw_hip_error(hipErrorInvalidValue, err_msg.c_str());
-	}
+          std::string err_msg = "failed to get memory from arg at index - " + std::to_string(idx);
+          throw_hip_error(hipErrorInvalidValue, err_msg.c_str());
+        }
 
-        r.set_arg(arg->index, hip_mem->get_xrt_bo());
+        if (!hip_mem_info.second) {
+          r.set_arg(static_cast<int>(arg->index), hip_mem->get_xrt_bo());
+        }
+        else {
+          auto parent_size = hip_mem->get_size();
+          auto sub_offset = hip_mem_info.second;
+          throw_invalid_value_if(sub_offset >= parent_size,
+                                 "kernel arg pointer offset out of range.");
+
+          auto sub_size = parent_size - sub_offset;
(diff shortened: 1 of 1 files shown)


src/runtime_src/hip/core/event.cpp
```diff
@@ -255,14 +255,10 @@ kernel_start::kernel_start(std::shared_ptr<function> f, void** args)
...
+        // address + offset use cases.
         void **bufptr = static_cast<void **>(args[idx]);
-        auto hip_mem = memory_database::instance().get_hip_mem_from_addr(*bufptr).first;
-        if (!hip_mem) {
-            std::string err_msg = "failed to get memory from arg at index - " + std::to_string(idx);
-	    throw_hip_error(hipErrorInvalidValue, err_msg.c_str());
-	}
-
-        r.set_arg(arg->index, hip_mem->get_xrt_bo());
+        r.set_arg(arg->index, *bufptr);
```
stsoe: I don't think this is safe.   (1)  Are we not losing the test to ensure that there is a backing xrt::bo associated with the base address of the bufptr?  (2) sizeof(void*) is platform specific, but 8 bytes on 64bit system, so safe in this case.  (3) more serious, xrt::bos are bound to the cmd associated with the xrt::run object on which the bo is set as argument (see xrt_kernel.cpp:2613), this is missed here.  At some point we want to create xrt::bo as unpinned memory and only pin at execution time, pinning will only happen for xrt::bo objects that are bound to the command.

svup-xilinx: Made below changes
1. Keeping the memory check to ensure there is a backing xrt::bo
2. Passing xrt::bo when there is no offset present in provided address to bind to cmd.
3. Passing raw pointer when there is an offset.

src/runtime_src/hip/core/event.cpp
```diff
@@ -255,14 +255,22 @@ kernel_start::kernel_start(std::shared_ptr<function> f, void** args)
...
         if (!hip_mem) {
-            std::string err_msg = "failed to get memory from arg at index - " + std::to_string(idx);
-	    throw_hip_error(hipErrorInvalidValue, err_msg.c_str());
-	}
+          std::string err_msg = "failed to get memory from arg at index - " + std::to_string(idx);
+          throw_hip_error(hipErrorInvalidValue, err_msg.c_str());
+        }
+
+        if (hip_mem_info.second)
+          r.set_arg(static_cast<int>(arg->index), *bufptr);
```
stsoe: This will not work because *bufptr is treated as a scalar and not bound to the command buffer as I indicated earlier.    I understand the offset, but you need to go the xrt::bo route with a sub-buffer.  Best I can tell, you don't need to worry about managing the scope of the sub-buffer, all that really matters is that set_arg is called with a bo that is a sub-buffer of the parent bo, whose lifetime you are managing.

So, at line 270, carve out an xrt::bo{hip_men->get_xrt_bo(), size, offset} where you should be able to compute size and offset from the hip memory?   Then pass the bo to set_arg.

svup-xilinx: updated code to use sub buffer when there is a offset.

chvamshi-xilinx (approved):

stsoe (approved):

PR #9734, opened 2026-04-15
Add trace points in xrt classes

rbramand-xilinx:
Added new trace points in xrt classes and its member functions

Its an EOU/Enhancement change

Added trace points in xrt objects alloc functions with _ctor in name and also in impl destructors with _dtor in name

Low

Tested with perf tool on ve2 and used a python script to pretty print the statistics and it works as expected

NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -232,6 +232,7 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
   static std::unique_ptr<uc_log_buffer>
   init_uc_log_buf(const std::shared_ptr<xrt_core::device>& device, xrt_core::hwctx_handle* ctx_hdl)
   {
+    XRT_TRACE_POINT_SCOPE(xrt_hw_context_init_uc_log_buf);
     // If uc log buffer is not supported then this function returns nullptr
     static auto uc_log_enabled = xrt_core::config::get_uc_log();
     if (!ctx_hdl || !uc_log_enabled)
@@ -347,6 +348,7 @@ public:
   void
   add_config(const xrt::elf& elf)
   {
+    XRT_TRACE_POINT_SCOPE(xrt_hw_context_add_config);
     auto part_size = elf.get_partition_size();
 
     // create hw ctx handle if not already created
@@ -409,6 +411,7 @@ public:
   void
   dump_uc_log_buffer()
   {
+    XRT_TRACE_POINT_SCOPE(xrt_hw_context_dump_uc_log_buffer);
     if (!m_uc_log_buf)
       return;
 
@@ -537,6 +540,7 @@ public:
   void
   dump_scratchpad_mem()
   {
+    XRT_TRACE_POINT_SCOPE(xrt_hw_context_dump_scratchpad_mem);
     if (m_scratchpad_buf.size() == 0) {
       xrt_core::message::send(xrt_core::message::severity_level::debug, "xrt_hw_context",
                               "preemption scratchpad memory is not available");
@@ -728,7 +732,7 @@ template<typename Cfg>
 static std::shared_ptr<hw_context_impl>
 alloc_hwctx_from_cfg(const xrt::device& device, const xrt::uuid& xclbin_id, Cfg&& cfg)
 {
-  XRT_TRACE_POINT_SCOPE(xrt_hw_context_cfg_ctor);
+  XRT_TRACE_POINT_SCOPE(xrt_hw_context_alloc_from_cfg);
   return post_alloc_hwctx(std::make_shared<hw_context_impl>(
       device.get_handle(), xclbin_id, hwctx_cfg_storage{std::forward<Cfg>(cfg)}));
 }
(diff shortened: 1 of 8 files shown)


stsoe (changes requested):
Why all these tracepoints?   How about adding tracepoints strategically  where we want to measure bottlenecks?   Tracepoints are not meant to show function calls.  The dtor tracepoints are completely useless, they measure nothing at all.

rbramand-xilinx:
> Why all these tracepoints? How about adding tracepoints strategically where we want to measure bottlenecks? Tracepoints are not meant to show function calls. The dtor tracepoints are completely useless, they measure nothing at all.

okay Soren, I will make changes accordingly and have tracepoints at needed places only. Thanks

chvamshi-xilinx:
> Why all these tracepoints? How about adding tracepoints strategically where we want to measure bottlenecks? Tracepoints are not meant to show function calls. The dtor tracepoints are completely useless, they measure nothing at all.

Hi @stsoe ,
We would like to enable trace log in destructors sothat we would get to know that destructors are being called etc.
In all destructor, Rahul added a TRACEPOINT_LOG sothat we can get the log.
In other places, he enabled TRACEPOINT_SCOPE.
As these are anyway no-ops if trace is not enabled, do you still have concern?

stsoe:
> > Why all these tracepoints? How about adding tracepoints strategically where we want to measure bottlenecks? Tracepoints are not meant to show function calls. The dtor tracepoints are completely useless, they measure nothing at all.
> 
> Hi @stsoe , We would like to enable trace log in destructors sothat we would get to know that destructors are being called etc. In all destructor, Rahul added a TRACEPOINT_LOG sothat we can get the log. In other places, he enabled TRACEPOINT_SCOPE. As these are anyway no-ops if trace is not enabled, do you still have concern?

Thanks @chvamshi-xilinx .  It sounds like you adding trace points for debugging function calls,  I don't think that is intended purpose?   My understanding is that tracepoints are to locate bottlenecks.   Yes, they are no-ops (at least on Linux), but my concern is the noise it generates by adding uninteresting tracepoints to the tracelog.

chvamshi-xilinx:
> > > Why all these tracepoints? How about adding tracepoints strategically where we want to measure bottlenecks? Tracepoints are not meant to show function calls. The dtor tracepoints are completely useless, they measure nothing at all.
> > 
> > 
> > Hi @stsoe , We would like to enable trace log in destructors sothat we would get to know that destructors are being called etc. In all destructor, Rahul added a TRACEPOINT_LOG sothat we can get the log. In other places, he enabled TRACEPOINT_SCOPE. As these are anyway no-ops if trace is not enabled, do you still have concern?
> 
> Thanks @chvamshi-xilinx . It sounds like you adding trace points for debugging function calls, I don't think that is intended purpose? My understanding is that tracepoints are to locate bottlenecks. Yes, they are no-ops (at least on Linux), but my concern is the noise it generates by adding uninteresting tracepoints to the tracelog.

Got it. Thanks @stsoe 
@rbramand-xilinx , please make changes suggested by Soren.
For identifying whether destructors are being called properly or not etc, we can use some other mechanism (verbosity level etc)

rbramand-xilinx:
Hi @stsoe, added trace points only at places where we need to identify bottlenecks as suggested. Please review. Thanks

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -409,6 +411,7 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
   void
   dump_uc_log_buffer()
   {
+    XRT_TRACE_POINT_SCOPE(xrt_hw_context_dump_uc_log_buffer);
```
stsoe, line 414: Isn't it the dump() function you want to trace?

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4512,9 +4522,16 @@ set_read_range(const xrt::kernel& kernel, uint32_t start, uint32_t size)
   ip->m_readrange = {start, size};
 }
 
+static std::shared_ptr<runlist_impl>
+alloc_runlist(xrt::hw_context hwctx)
+{
+  XRT_TRACE_POINT_SCOPE(xrt_runlist_alloc);
```
stsoe, line 4528: What are we measuring?  The constructor is trivial with 0 overhead.  I understand why alloc_runlist was created, it was to ensure that you can measure the time it takes to create runlist_impl, which does all initialization in its initializer list.  This makes sense if we truly want to measure the construction time, but do we?

The function is kind of like an observer of hwctx, where the sink is really runlist_impl ctor.  Shouldn't alloc_runlist take hwctx by const ref and leave the copying to the runlist_impl ctor?

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4527,6 +4544,7 @@ void
 runlist::
 add(const xrt::run& run)
 {
+  XRT_TRACE_POINT_SCOPE(xrt_runlist_add);
```
stsoe, line 4547: Why are we measuring here and not in `&&` method.  If we want to trace anything here, then it should be runlist_impl::add().

src/runtime_src/core/common/api/xrt_kernel.cpp
stsoe, line 4569: This wasn't added n this PR, but probably shouldn't even be here, it probably should be in runlist_imp::execute().

stsoe (commented):
When I look at all these  tracepoints, I think they are added without much thought about what we want to measure.  There are tracepoints in Alveo obsolete code, .e.g update arg functions, why were they added?

I am dubious about the value of many of these tracepoints.  Profiling of user space code is not through tracepoints, but with a profiler, e.g. valgrind.   

In the past we have added tracepoints to show bottlenecks through kernel module code paths.  It make a lot of sense to trace API overhead in UMD code paths that involve critical KMD code paths.   Is this the thinking behind all these tracepoints added here?  Otherwise just use valgrind or some other profiling tool to optimize the code if we think there are bottlenecks.

PR #9744, opened 2026-04-17
Deprecate messages added for kernel, IP and buffer constructors

karthdmg-xilinx:
Mark legacy XRT entry points as deprecated in favor of hw_context-based APIs where a replacement exists (xrt::kernel, xrt::ip, xrt::aie::buffer, xrt::aie::profiling, 
and the xclDeviceHandle kernel overload).
old APIs are not getting depricate warnings.
Mark legacy XRT entry points as deprecated in favor of hw_context-based APIs where a replacement exists (xrt::kernel, xrt::ip, xrt::aie::buffer, xrt::aie::profiling, 
and the xclDeviceHandle kernel overload).
None 
build a sample application that uses these ctors and verify the warning is logged
None

### DIFF
--- a/src/runtime_src/core/include/xrt/xrt_kernel.h
+++ b/src/runtime_src/core/include/xrt/xrt_kernel.h
@@ -795,6 +795,7 @@ public:
    * other kernels and other process will have shared access to same
    * compute units.
    */
+  [[deprecated("deprecated, please use kernel(hw_context, name) instead")]]
   XRT_API_EXPORT
   kernel(const xrt::device& device, const xrt::uuid& xclbin_id, const std::string& name,
          cu_access_mode mode = cu_access_mode::shared);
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/include/xrt/xrt_kernel.h
```diff
@@ -807,14 +808,13 @@ class kernel
   /// @endcond
 
   /// @cond
-  /// Deprecated construtor for exclusive access
-  kernel(const xrt::device& device, const xrt::uuid& xclbin_id, const std::string& name, bool ex)
-    : kernel(device, xclbin_id, name, ex ? cu_access_mode::exclusive : cu_access_mode::shared)
-  {}
+  /// Deprecated constructor for exclusive access
+  kernel(const xrt::device& device, const xrt::uuid& xclbin_id, const std::string& name, bool ex);
```
stsoe, line 812: Any reason for this change,  Shouldn't this ctor be deprecated as well, e.g. calling the already deprecated ctor?

karthdmg-xilinx: added deprication warning for this. replaced an inline function with declaration. moved the implementation to cpp file. we are adding messages for declarations only to skip xrt build failures due to g++ build catching the exception.

stsoe: I see, so if this ctor was left as it was, then compilation including xrt_kernel.h would always issue warning regardless of whether the ctor was used or not.   I can see that, it makes sense.  Thanks!

stsoe (approved):

chvamshi-xilinx (approved):

stsoe:
@karthdmg-xilinx @chvamshi-xilinx 
This PR breaks pyxrt compilation.  See AIESW-31422.

chvamshi-xilinx:
> @karthdmg-xilinx @chvamshi-xilinx This PR breaks pyxrt compilation. See AIESW-31422.

@stsoe , Assigned it to @svup-xilinx . He will fix it.

PR #9752, opened 2026-04-23
AIESW-29024:CERT log buffer improvements

svup-xilinx:
- Enable uC log collection by default. No overhead on success case.
- Add configurable per-uC log buffer size via xrt.ini (Debug.uc_log_size_per_uc_kb). Default size 16kb.
- Make background dumper thread optional and configurable (Debug.uc_log_dumper_thread).
- Implement safe thread creation/cleanup when dumper thread is disabled.
[AIESW-29024](https://jira.xilinx.com/browse/AIESW-29024)
Low
tested on telluride board with and without dumping thread. 
NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -42,10 +42,11 @@
 #endif
 
 namespace {
-static constexpr double hz_per_mhz = 1'000'000.0;
-
 constexpr std::size_t
-operator""_mb(unsigned long long v) { return 1024u * 1024u * v; }
+operator""_kb(unsigned long long v) { return 1024U * v; } // NOLINT(cppcoreguidelines-avoid-magic-numbers)
+
+constexpr double
+operator""_mhz(long double v) { return static_cast<double>(v) * 1'000'000.0; } // NOLINT(cppcoreguidelines-avoid-magic-numbers)
 
 // Dumps the content into a file with given size from given offset
 static void
@@ -123,12 +124,13 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
                           size_t num_uc)
     {
       // parameters for uc log buffer dumper
-      // tweak dump interval, size_per_uc based on experiments
-      constexpr size_t size_per_uc = 2_mb;
       constexpr size_t dump_interval_ms = 50;
       constexpr size_t metadata_size = sizeof(xrt_core::buffer_dumper::log_entry);
       constexpr size_t count_offset = 0;
       constexpr size_t count_size = 8;
+
+      // Get per-uC cert log size from xrt.ini.
+      const auto size_per_uc = xrt_core::config::get_uc_log_size_per_uc_kb() * 1_kb;
 
       auto bo = xrt_core::bo_int::create_bo(
           ctx_hdl, device, (size_per_uc * num_uc), xrt_core::bo_int::use_type::log);
@@ -161,6 +163,7 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
       config.dump_file_prefix = "uc_log_" + std::to_string(ctx_hdl->get_slotidx());
       config.dump_buffer = std::move(bo);
       config.dump_bin_format = xrt_core::config::get_uc_log_bin_format();
+      config.enable_dumper_thread = xrt_core::config::get_uc_log_dumper_thread();
 
       return std::make_unique<xrt_core::buffer_dumper>(std::move(config));
     }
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -43,9 +43,10 @@
 
 namespace {
 static constexpr double hz_per_mhz = 1'000'000.0;
+static constexpr std::size_t bytes_per_kb = 1024U;
 
 constexpr std::size_t
-operator""_mb(unsigned long long v) { return 1024u * 1024u * v; }
+operator""_kb(unsigned long long v) { return bytes_per_kb * v; }
```
stsoe: Please just use user-defined literal here and add // NOLINT to prevent clang-tidy raising an issue.  No need for static constexpr here, but do make sure the operators are constexpr.

Also please defined literal for _mhz as well.

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -123,13 +124,15 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
       // parameters for uc log buffer dumper
-      // tweak dump interval, size_per_uc based on experiments
-      constexpr size_t size_per_uc = 2_mb;
       constexpr size_t dump_interval_ms = 50;
       constexpr size_t metadata_size = sizeof(xrt_core::buffer_dumper::log_entry);
       constexpr size_t count_offset = 0;
       constexpr size_t count_size = 8;
 
+      // Get per-uC cert log size from xrt.ini.
+      const auto size_per_uc = static_cast<std::size_t>(
```
stsoe: Is this static_cast really necessary when _kb returns std::size_t and argument unsigned int converts to _kb argument type?

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -3006,11 +3012,11 @@ class run_impl : public std::enable_shared_from_this<run_impl>
     if (dump)
       xrt_core::hw_context_int::dump_scratchpad_mem(kernel->get_hw_context());
 
-    // dump uc log in error cases
-    if (!success)
-      xrt_core::hw_context_int::dump_uc_log_buffer(kernel->get_hw_context());
-    else
-      // update usage logger in success cases
+    // dump uc log buffer in both success and error cases
+    xrt_core::hw_context_int::dump_uc_log_buffer(kernel->get_hw_context());
```
stsoe: There is more and more code executed in dump_logs unconditionally, I would like to see that dump_logs is a no-op in success case.

Dumping of logs does not appear well designed. There are too many dump functions and it is not clear which one dumps what and why.

src/runtime_src/core/common/config_reader.h
```diff
@@ -1185,7 +1185,7 @@ get_run_buffer_pool_max_size()
 inline bool
 get_uc_log()
 {
-  static bool value = detail::get_bool_value("Debug.uc_log", false);
+  static bool value = detail::get_bool_value("Debug.uc_log", true);
```
stsoe, line 1190: What are the implications of make this default to true?

stsoe (changes requested):

src/runtime_src/core/common/api/xrt_hw_context.cpp
svup-xilinx: Updated code to use // NOLINT and added literal for _mhz.

src/runtime_src/core/common/api/xrt_hw_context.cpp
svup-xilinx: removed static_cast as it is not nessasary.

src/runtime_src/core/common/api/xrt_kernel.cpp
svup-xilinx: Reverted to dump uc logs in error cases only.

src/runtime_src/core/common/config_reader.h
svup-xilinx: As discussed offline updated comments in the code.

chvamshi-xilinx (approved):

stsoe (approved):

stsoe:
@chvamshi-xilinx .  The change of config Debug.uc_log to default true, means that we will create a uc log buffer even on Strix.  This does not seem right.   Can we revert the change or otherwise guard non AIE4 platforms?   I am not recalling why uc_log buffer has to be created by default?
cc: @svup-xilinx  @maxzhen

chvamshi-xilinx:
> @chvamshi-xilinx . The change of config Debug.uc_log to default true, means that we will create a uc log buffer even on Strix. This does not seem right. Can we revert the change or otherwise guard non AIE4 platforms? I am not recalling why uc_log buffer has to be created by default? cc: @svup-xilinx @maxzhen

Hi @stsoe , uc_log is going to be empty in all the success cases. uc log is going to be useful in failure case, uc log size would be very small even in failure case.
There is an email from @xuhz  with all the explanation and requirement. This PR is raised based on that.

Regarding Strix, @svup-xilinx , Can you please fix the issue.

PR #9755, opened 2026-04-27
AIESW-29276 Update xrt-smi help menu

rchane:
https://jira.xilinx.com/browse/AIESW-29276
Discovered during internal bash testing
Updated help wording, separated show-hidden from advanced help, and specified -d requires BDF in error message
May require Alveo users to specify BDF with -d/--device if they left -d/--device empty previously
Tested on KRK. Updated `device` description:
```
-d, --device <BDF> - Specify a BDF. If the option is omitted and there is only 1 device, that
                       device is used
```
N/A

### DIFF
--- a/src/runtime_src/core/tools/common/XBHelpMenusCore.cpp
+++ b/src/runtime_src/core/tools/common/XBHelpMenusCore.cpp
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright (C) 2020-2022 Xilinx, Inc
-// Copyright (C) 2022-2025 Advanced Micro Devices, Inc. All rights reserved.
+// Copyright (C) 2022-2026 Advanced Micro Devices, Inc. All rights reserved.
 // ------ I N C L U D E   F I L E S -------------------------------------------
 // Local - Include Files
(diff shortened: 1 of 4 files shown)


src/runtime_src/core/tools/common/XBHelpMenusCore.cpp
```diff
@@ -373,8 +373,15 @@ XBUtilities::report_commands_help( const std::string &_executable,
 
   report_option_help("OPTIONS", _optionDescription);
 
-  if (XBU::getAdvance())
-    report_option_help(std::string("OPTIONS ") + sHidden, _optionHidden);
+  if (XBU::getAdvance()) {
```
aktondak: I'm not clear on what this code is doing. Doesn't seem right to be looping over all _optionHidden

rchane: This is creating a new options_description from the one that was handed into the function without the subCmd description, as there seems to be no method to erase/remove an option once added. I think we can make this cleaner if and when we eventually move away from boost handling.

aktondak (changes requested):

aktondak (approved):

PR #9758, opened 2026-04-28
EDGEML-13495: Fix Binskim issues in xrt

yvonyuan:
https://amd.atlassian.net/browse/EDGEML-13495 and https://amd.atlassian.net/browse/EDGEML-13496

Fixed Binskim BA6002: eliminate duplicate strings.
Fixed Binskim BA2007: build at warning level 3

### DIFF
--- a/src/CMake/nativeWin.cmake
+++ b/src/CMake/nativeWin.cmake
@@ -44,10 +44,10 @@ if (MSVC)
     /Qspectre     # compile with the Spectre mitigations switch
     /ZH:SHA_256   # enable secure source code hashing
     /guard:cf     # enable compiler control guard feature (CFG) to prevent attackers from redirecting execution to unsafe locations
+    /GF           # eliminate duplicate strings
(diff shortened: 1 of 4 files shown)


stsoe (changes requested):
Please check https://github.com/Xilinx/XRT/pull/9500
There is a reason the conditional options were added.

stsoe (commented):
Please confirm build are clean, I think you will get warnings at link time if not a compile time.
And sign-off, please.

yvonyuan:
> Please check #9500 There is a reason the conditional options were added.

Make sense to me. Moved /OPT:REF back to non-Debug only. I will re-run the Binskim check and if the issue is still there, we might consider treat Debug BA6005 as a BinSkim exception

yvonyuan:
Closed this PR as after removing xclbinutil from the artifacts for SWV npu driver, issue BA2007 and BA6002 goes away.

PR #9759, opened 2026-04-28
Minor re-factor tweaks

stsoe:
Some small re-factor changes to xrt_hw_context.cpp with few more comments.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -6,25 +6,27 @@
 #define XRT_API_SOURCE         // exporting xrt_hwcontext.h
 #define XCL_DRIVER_DLL_EXPORT  // exporting xrt_xclbin.h
 #define XRT_CORE_COMMON_SOURCE // in same dll as coreutil
+#include "core/include/xrt/xrt_hw_context.h"
+#include "hw_context_int.h"
+
+#include "core/include/xrt/detail/span.h"
+#include "core/include/xrt/experimental/xrt_ext.h"
 
 #include "core/common/buffer_dumper.h"
 #include "core/common/config_reader.h"
+#include "core/common/device.h"
 #include "core/common/message.h"
 #include "core/common/query_requests.h"
 #include "core/common/time.h"
+#include "core/common/trace.h"
+#include "core/common/usage_metrics.h"
 #include "core/common/utils.h"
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -205,30 +214,54 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
+  // specifically for that context.
   std::once_flag m_scratchpad_init_flag; // used for thread safe lazy init of scratchpad
   xrt::bo m_scratchpad_buf;
+
   std::shared_ptr<xrt_core::usage_metrics::base_logger> m_usage_logger =
       xrt_core::usage_metrics::get_usage_metrics_logger();
+
   bool m_elf_flow = false;
 
+  // update_elf_map() - Insert all kernels in elf
```
rbramand-xilinx: Minor fix
Mismatch b/w function name and the name in comment

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -351,27 +384,16 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
-
-    // add ELF only if partition size matches existing configuration
-    if (m_partition_size != part_size)
+    if (m_hdl && m_partition_size != part_size)
       throw std::runtime_error("can not add config to ctx with different configuration\n");
 
-    // Add kernels available in ELF to elf map
-    // This function throws if kernel with same name is already present
-    create_elf_map(elf);
+    std::lock_guard lk(m_mutex);
```
rbramand-xilinx: init_from_elf() is currently invoked while holding the mutex and performs a driver call to create the HW context. This can potentially extend the lock hold time due to kernel latency.
So I was thinking whether this needs to be under the lock or if it could be done outside the critical section.
Ideally this should be okay as hw ctx creation is not in critical path but just a thought.

rbramand-xilinx (approved):

src/runtime_src/core/common/api/xrt_hw_context.cpp
stsoe: @rbramand-xilinx .  Thanks for the comments.  My concern was that add_config() could theoretically be called from different threads, and without locking m_hdl creation could be created under race (multiple handles created).   I don't think we have choice but to lock as I did?

PR #9761, opened 2026-04-30
AIESW-31422:Fix use of deprecated kernel apis.

svup-xilinx:
The deprecated constructors kernel (device, uuid, name) and kernel (device, uuid, name, mode) were being called directly from Python bindings, causing compilation failures
[AIESW-31422](https://jira.xilinx.com/browse/AIESW-31422)
Added compatibility shim for both kernel constructor overloads using modern hw_context API internally to eliminate C++ deprecation warnings while maintaining Python backward compatibility.
- kernel(device, uuid, name) → uses hw_context with shared mode
- kernel(device, uuid, name, mode) → uses hw_context with specified mode
- Issues Python DeprecationWarning at runtime
- Fixes the issue properly instead of suppressing warnings on Windows.
Python.
Tested on windows with custom made test on Strix machine.
None. Runtime warnings guide users to modern API.

### DIFF
--- a/src/python/pybind11/src/pyxrt.cpp
+++ b/src/python/pybind11/src/pyxrt.cpp
@@ -33,10 +33,6 @@
 #include <stdexcept>
 #include <string>
 
-#ifdef _MSC_VER
-#pragma warning(disable: 4996)  // C4996: deprecated function; Python binding delegates to C++
-#endif
-
 namespace py = pybind11;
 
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx (approved):

stsoe (approved):
I am not able to review Python binding code to the level where I understand what's going on.  I assume the changes are okay.

PR #9764, opened 2026-04-30
Apply Fedora xbmgmt link patch

stsoe:
Remove boost_filesystem from xbmgmt link.  It is not used.


stsoe:
Argh, old compilers kill this patch.

PR #9770, opened 2026-05-05
Add Debug.profiling_runtime_config xrt.ini option for XDP

jyothees99:
Add a cached string getter xrt_core::config::get_profiling_runtime_config() in core/common/config_reader.h that reads an inline JSON blob from the new Debug.profiling_runtime_config xrt.ini key. The blob carries XDP runtime configuration; today only its control_instrumentation section is consumed (by the aie_dtrace plugin in the XDP submodule).

Add a lightweight cached probe in core/common/xdp/profile.cpp that looks for the control_instrumentation subtree inside that blob without linking against xdp_core (which is loaded dynamically). Route all seven get_aie_dtrace() call sites through a new aie_dtrace_enabled() helper that returns true when either Debug.aie_dtrace is set in xrt.ini or the blob carries control_instrumentation. This lets the plugin auto-load when only the runtime config is supplied.

aie_trace, aie_profile, ml_timeline and all other plugin gates are intentionally untouched; they continue to use their existing xrt.ini settings only.

Made-with: Cursor

The VAIML provider/runtime is moving toward passing XDP profiling configuration as a single JSON blob (vaiml_config.profiling_runtime_config in the model's vitisai_config.json) instead of asking the user to set individual AIE_dtrace_settings.* xrt.ini keys (AIESW-31158). FlexmlRT forwards this blob verbatim to XRT via xrt::ini::set, and XDP owns the schema. XRT needs to:

Accept this blob through xrt.ini and expose it to XDP via a cached getter.
Auto-enable the aie_dtrace plugin when the blob carries a control_instrumentation section, so the user does not also have to set Debug.aie_dtrace=true in addition to providing the JSON.
Without (1) the blob has nowhere to live; without (2) users must configure the same intent in two places, which defeats the centralization goal.

Not a bug fix. New feature plumbing for the VAIML profiling-config design (AIESW-31158). No prior PR introduced anything that needs reverting.
Solution (this commit, in xrt_coreutil):

Added Debug.profiling_runtime_config xrt.ini key, with a cached xrt_core::config::get_profiling_runtime_config() string getter in core/common/config_reader.h.
Added a small Boost.PropertyTree probe in core/common/xdp/profile.cpp that detects whether the blob contains a control_instrumentation subtree with any of aie_tile, mem_tile, or interface_tile. The probe is intentionally minimal so it does not require linking against xdp_core, which is loaded dynamically.
Routed all seven xrt_core::config::get_aie_dtrace() call sites through a single new aie_dtrace_enabled() helper that returns true if either the legacy Debug.aie_dtrace ini key is set or the blob's control_instrumentation is present.
The full JSON parser/consumer (xdp::profiling_runtime_config::*) lives on the matching XDP-submodule branch — kept on the XDP side because the schema is XDP-owned and is expected to evolve faster than xrt_coreutil.

Alternatives considered and rejected:

Put the full JSON parser in xrt_coreutil. Rejected — ties the schema to a stable layer. The schema belongs to XDP and needs to iterate. Probe vs full parser is the minimum-coupling split that still lets the load-time gate work.
Require users to also set Debug.aie_dtrace=true alongside the new key. Rejected — duplicates the same intent in two places, surprises users whose JSON is silently ignored, and contradicts the centralization goal of the new design.
Have FlexmlRT translate the JSON into individual AIE_dtrace_settings.* keys. Rejected per the design doc — VAIML runtime should not need to know XDP's per-key schema; XDP owns it.
No
VAIML design bevformer_tiny_resnet50_batch6_without_split

### DIFF
--- a/src/runtime_src/core/common/config_reader.h
+++ b/src/runtime_src/core/common/config_reader.h
@@ -181,6 +181,18 @@ get_aie_dtrace()
   return value;
 }
 
+// Inline JSON blob carrying XDP profiling runtime configuration.
+// When non-empty, XDP parses this and uses it to drive per-plugin settings
+// (e.g. control_instrumentation for aie_dtrace) in preference to the
+// legacy AIE_*_settings.* xrt.ini sections. See
+// xdp/profile/plugin/vp_base/profiling_runtime_config.h for the consumer.
+inline std::string
+get_profiling_runtime_config()
+{
+  static std::string value = detail::get_string_value("Debug.profiling_runtime_config", "");
+  return value;
+}
+
 inline bool
 get_aie_debug()
 {
--- a/src/runtime_src/core/common/xdp/profile.cpp
+++ b/src/runtime_src/core/common/xdp/profile.cpp
@@ -11,6 +11,9 @@
 #include "core/common/message.h"
 #include "core/include/xrt/xrt_kernel.h"
 
+#include <boost/property_tree/json_parser.hpp>
+#include <boost/property_tree/ptree.hpp>
+
 #include <cstring>
 #include <functional>
 #include <sstream>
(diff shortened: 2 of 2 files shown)


src/runtime_src/core/common/xdp/profile.cpp
```diff
@@ -68,6 +71,84 @@ namespace {
     return hwem;
   }
 #endif
+
```
IshitaGhosh: @jyothees99 According to the spec for "profiling_runtime_config", all the XDP Plugin Level Flags/Features, will be set using "Debug.aie_trace"/"Debug.aie_dtrace" and "Debug.xdp_mode". Only the detailed settings for control_instrumentation/trace can be done via "Debug.profiling_runtime_config". The detailed settings are equivalent to AIE_Trace_settings etc..
So, we dont need to parse "profiling_runtime_config"  during XDP plugin load.
It will be parsed in XDP for detailed settings.

src/runtime_src/core/common/config_reader.h
```diff
@@ -181,6 +181,18 @@ get_aie_dtrace()
   return value;
 }
 
+// Inline JSON blob carrying XDP profiling runtime configuration.
+// When non-empty, XDP parses this and uses it to drive per-plugin settings
+// (e.g. control_instrumentation for aie_dtrace) in preference to the
+// legacy AIE_*_settings.* xrt.ini sections. See
+// xdp/profile/plugin/vp_base/profiling_runtime_config.h for the consumer.
+inline std::string
+get_profiling_runtime_config()
```
IshitaGhosh, line 190: The new API get_profiling_runtime_config looks good.

IshitaGhosh (changes requested):

IshitaGhosh (approved):
New changes look good.

garimadhaked (approved):

garimadhaked (approved):

PR #9772, opened 2026-05-05
Print ELF filename, UUID and kernel name:instance in context health report

sayyanna:
Print ELF filename/UUID (if file name available) and kernel name:instance in context health report
Report the ctrlcode ELF and context which was executing (currently it is not obvious which ELF met with error). We should report the ELF .note.xrt.UID which is stored inside the ELF itself.

Kernel Instance: DPU:fa8c35f430e346a18e9a2464bf578d96_subgraph_32
ELF UUID:        ff2870dc-66a5-49ac-f114-f9731ce20992
ctx_state = 0x00000001 (CTX_STATUS_ERROR)
ctx_error_type = 0x00000003 (NPU_ASYNC_EVENT_CTX_ERR_PREEMPTION_TIMEOUT)
number of uC reported = 6
uc_info[0]:
uc_idx=0x00000000
uc_idle_status=0x00000000 (NONE)
misc_status=0x00000000 (NONE)
fw_state=0x0000000C (FW_STATE_CHECK_NEWPAGE)
page_idx=0x00000043
offset=0x000000DC
restore_page=0x00000000
restore_offset=0x00000000
uc_ear=0x00000000
uc_esr=0x00000000


src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -1502,7 +1502,9 @@ valid_or_error(const std::shared_ptr<elf_impl>& handle)
 elf::
 elf(const std::string& fnm)
   : detail::pimpl<elf_impl>{create_elf_impl(load_elfio(fnm))}
-{}
+{
+  handle->set_filename(fnm);
+}
```
stsoe: No constructor body here.   elf_impl construction needs to be refactored.
@rbramand-xilinx.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1402,6 +1402,7 @@ class kernel_impl : public std::enable_shared_from_this<kernel_impl>
 
 private:
   std::string name;                           // kernel name
+  std::string m_kernel_instance;             // full "kernel:instance" string, empty if no instance
```
```suggestion
  std::string m_instance_name;             // full "kernel:instance" string, empty if no instance
```

src/runtime_src/core/common/api/CMakeLists.txt
```diff
@@ -28,4 +28,5 @@ target_include_directories(core_common_api_library_objects
   PRIVATE
   ${XRT_SOURCE_DIR}/runtime_src
   ${XRT_SOURCE_DIR}/runtime_src/core/common/elf
+  ${XRT_SOURCE_DIR}/runtime_src/core/common/aiebu/src/cpp/include
```
stsoe: Why?

stsoe (changes requested):
I will review after Rahul has looked it over.
Please fix the few comments I added, and please make sure not chmod on any files, e.g. your workspace files should be 644, it appears they are all changed to 755?

src/runtime_src/core/common/api/elf_int.h
```diff
@@ -289,6 +289,9 @@ class elf_impl
   static constexpr uint32_t schema_mask = ~addend_mask;
   // NOLINTEND
 
+  // Filename this ELF was loaded from (empty if loaded from buffer/stream)
+  std::string m_filename;
```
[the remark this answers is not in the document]
rbramand-xilinx: @sayyanna please move this member variable above NOLINTEND to supress clang-tidy warning

src/runtime_src/core/common/api/xrt_elf.cpp
rbramand-xilinx: yeah I agree Soren we can directly pass filename or buffer/stream to elf_impl constructor and create elfio object there. This way we can eliminate set_filename function
@sayyanna will you refactor it accordingly? if not I can make those changes. Please do let me know. Thanks

src/runtime_src/core/common/api/elf_int.h
```diff
@@ -360,6 +363,20 @@ class elf_impl
...
+  // Get the filename this ELF was loaded from (empty if loaded from buffer/stream)
+  const std::string&
+  get_filename() const
+  {
+    return m_filename;
+  }
+
+  // Set the filename (called by xrt::elf constructors that load from a path)
+  void
+  set_filename(const std::string& filename)
```
rbramand-xilinx: This function is not needed if we do refactoring as mentioned in below comment

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1755,6 +1756,7 @@ class kernel_impl : public std::enable_shared_from_this<kernel_impl>
   // construction and shared ownership must be tied to the kernel_impl
   kernel_impl(std::shared_ptr<device_type> dev, xrt::hw_context ctx, xrt::module mod, const std::string& nm)
     : name(nm.substr(0,nm.find(":")))                          // filter instance names
+    , m_instance_name(nm.find(':') != std::string::npos ? nm : "") // full "kernel:instance" string
```
rbramand-xilinx: m_instance_name doesnt make sense here as you are storing full canonical name, may be use something like m_full_name or m_canonical_name

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1896,6 +1899,12 @@ class kernel_impl : public std::enable_shared_from_this<kernel_impl>
     return name;
   }
 
+  const std::string&
+  get_instance_name() const
```
rbramand-xilinx: same here, change function name after changing member variable name

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4764,9 +4795,42 @@ amend_aie_error_message(const ert_packet* epkt, const std::string& msg)
   return oss.str();
 }
 
+// get_elf_identity_from_run() - Extract ELF filename, kernel instance, and UUID from a run object.
+// Returns empty values if the run has no associated ELF (non-ELF flow).
+// UUID (from .note.xrt.UID) serves as a fallback identifier when the filename is unavailable.
+static std::tuple<std::string, std::string, std::string>
+get_elf_identity_from_run(const xrt::run& run)
```
rbramand-xilinx: please check if this function can be refactored by clubbing try blocks

rbramand-xilinx (changes requested):

src/runtime_src/core/common/api/xrt_elf.cpp
sayyanna: > yeah I agree Soren we can directly pass filename or buffer/stream to elf_impl constructor and create elfio object there. This way we can eliminate set_filename function @sayyanna will you refactor it accordingly? if not I can make those changes. Please do let me know. Thanks

@rbramand-xilinx, Please take care of the refactoring, may be that can be handled as a separate task outside of this PR?

rbramand-xilinx: Please add the changes I have shared offline which will eliminate set_filename function

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -201,6 +201,164 @@ cmd_state_to_string(ert_cmd_state state)
...
+
+  auto itr = fw_state_string.find(fw_status);
+  return itr == fw_state_string.end()
+    ? "out of range"
+    : itr->second;
+}
+std::string
+ctx_error_type_to_string(uint32_t ctx_error_type)
+{
+  static const std::map<uint32_t, const char*> ctx_error_type_string {
```
rbramand-xilinx: can we use switch case instead, it avoids static map and default case should handle "Out of range" cleanly
Something like below

std::string
uc_fwstate_to_string(uint32_t fw_status)
{
  switch (fw_status) {
  case 0x1001: return "FW_STATE_LEADER_HOST_QUEUE_POP";
  case 0x1002: return "FW_STATE_LEADER_DISTRIBUTE_WORK";
  // ...
  case 0x0FFF: return "FW_STATE_EXIT";
  default:
    std::ostringstream oss;
    oss << "unknown (0x" << std::hex << std::uppercase << fw_status << ")";
    return oss.str();
  }
}

rbramand-xilinx (changes requested):

src/runtime_src/core/common/api/xrt_kernel.cpp
sayyanna: I personally think table-driven map is a reasonable fit here, adding big switch/case (45 cases) is not needed. Also it is in sync with existing code like cmd_state_to_string(ert_cmd_state state). Please note that this executed only at the end on the event of ERT_CMD_STATE_TIMEOUT
@stsoe, please give your suggestion as well.

rbramand-xilinx: oh I thought there are only 6-7 fields here, if thats the case it is okay. Thanks

rbramand-xilinx (approved):
Please fix clang-tidy warnings as well. Thanks

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -201,6 +201,164 @@ cmd_state_to_string(ert_cmd_state state)
...
+    {2, "CTX_STATUS_IDLE"},
+    {3, "CTX_STATUS_RUNNABLE"},
+    {4, "CTX_STATUS_RUNNING"},
+    {5, "CTX_STATUS_PREEMPTING"},
+  };
+
+  auto itr = ctx_status_string.find(ctx_status);
+  return itr == ctx_status_string.end()
+    ? "UNKNOWN"
+     : itr->second;
```
stsoe: Why not use a constexpr std::array?   The map is zero indexed anyway.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -201,6 +201,164 @@ cmd_state_to_string(ert_cmd_state state)
     : itr->second;
 }
 
+std::string
```
```suggestion
static std::string
```

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -201,6 +201,164 @@ cmd_state_to_string(ert_cmd_state state)
...
+    {5, "CTX_STATUS_PREEMPTING"},
+  };
+
+  auto itr = ctx_status_string.find(ctx_status);
+  return itr == ctx_status_string.end()
+    ? "UNKNOWN"
+     : itr->second;
+}
+
+std::string
```
```suggestion
static std::string
```

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4764,9 +4952,37 @@ amend_aie_error_message(const ert_packet* epkt, const std::string& msg)
...
+  auto impl = run.get_handle();
+  if (!impl)
+    return {};
+  try {
+    const auto& mod = impl->get_module();
+    if (!mod)
+      return {};
+    auto elf_handle = xrt_core::module_int::get_elf_handle(mod);
+    std::string uuid_str;
+    uuid_str = elf_handle->get_cfg_uuid().to_string();
```
stsoe: Fold into return

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4764,9 +4952,37 @@ amend_aie_error_message(const ert_packet* epkt, const std::string& msg)
...
+  }
+}
+
 run::aie_error::
 aie_error(const xrt::run& run, const std::string& what)
-  : command_error(run, amend_aie_error_message(run.get_ert_packet(), what))
+  : command_error(run, [&] {
+      auto [filename, kernel_instance, elf_uuid] = get_elf_identity_from_run(run);
+      return amend_aie_error_message(run.get_ert_packet(), what, filename, kernel_instance, elf_uuid);
+    }())
```
stsoe: To minimize code duplication, let's pass the run to `amend_aie_error_message`

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -4783,7 +4999,10 @@ command_error(const xrt::run& run, ert_cmd_state state, const std::string& msg)
 
 runlist::aie_error::
 aie_error(const xrt::run& run, ert_cmd_state state, const std::string& what)
-  : command_error(run, state, amend_aie_error_message(run.get_ert_packet(), what))
+  : command_error(run, state, [&] {
+      auto [filename, kernel_instance, elf_uuid] = get_elf_identity_from_run(run);
+      return amend_aie_error_message(run.get_ert_packet(), what, filename, kernel_instance, elf_uuid);
+    }())
```
stsoe: Avoid this code duplication.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -201,6 +201,164 @@ cmd_state_to_string(ert_cmd_state state)
...
+    {0x0022, "FW_STATE_SYNC_MM2DM_HANG"},
+    {0x0FFF, "FW_STATE_EXIT"},
+  };
+
+  auto itr = fw_state_string.find(fw_status);
+  return itr == fw_state_string.end()
+    ? "UNKNOWN"
+    : itr->second;
+}
+std::string
```
```suggestion
static std::string
```

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -201,6 +201,164 @@ cmd_state_to_string(ert_cmd_state state)
...
+ctx_error_type_to_string(uint32_t ctx_error_type)
+{
+  static const std::map<uint32_t, const char*> ctx_error_type_string {
+    {0, "NPU_ASYNC_EVENT_CTX_ERR_HWSCH_FAILURE"},
+    {1, "NPU_ASYNC_EVENT_CTX_ERR_STOP_FAILURE"},
+    {2, "NPU_ASYNC_EVENT_CTX_ERR_AIE_FAILURE"},
+    {3, "NPU_ASYNC_EVENT_CTX_ERR_PREEMPTION_TIMEOUT"},
+    {4, "NPU_ASYNC_EVENT_CTX_ERR_NEW_PROCESS_FAILURE"},
+    {5, "NPU_ASYNC_EVENT_CTX_ERR_UC_CRITICAL_ERROR"},
+    {6, "NPU_ASYNC_EVENT_CTX_ERR_UC_COMPLETION_TIMEOUT"},
```
stsoe: std::array

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -201,6 +201,164 @@ cmd_state_to_string(ert_cmd_state state)
...
+    result += (boost::format("UNKNOWN(0x%x)") % unknown).str();
+  }
+  return result;
+}
+
+// uc_idle_status is a bit mask (ert_uc_health_info / hsa_lite_status).
+std::string
+uc_idle_status_flags_to_string(uint32_t idle_status)
+{
+  static const std::array<std::pair<uint32_t, const char*>, 3> idle_status_flags {{
```
```suggestion
  static constexpr std::array<std::pair<uint32_t, const char*>, 3> idle_status_flags {{
```

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -201,6 +201,164 @@ cmd_state_to_string(ert_cmd_state state)
...
+
+  std::string result;
+  uint32_t unknown = idle_status;
+  for (const auto& entry : idle_status_flags) {
+    if (idle_status & entry.first) {
+      if (!result.empty())
+        result += " | ";
+      result += entry.second;
+      unknown &= ~entry.first;
+    }
```
stsoe: Flip condition for early continue.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1896,6 +2057,12 @@ class kernel_impl : public std::enable_shared_from_this<kernel_impl>
     return name;
   }
 
+  const std::string&
+  get_full_name() const
+  {
+    return m_full_name;
+  }
+
```
stsoe: See comment about m_full_name.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -1402,6 +1560,7 @@ class kernel_impl : public std::enable_shared_from_this<kernel_impl>
 
 private:
   std::string name;                           // kernel name
+  std::string m_full_name;                    // full "kernel:instance" string, empty if no instance
```
stsoe, line 1555: I understand the code mechanics, but I don't understand what m_full_name gives you.  

It is set to `nm` if `nm` contains `:` otherwise empty string.  Yet in `get_elf_identity_from_run()` you use this m_full_name which could potentially be empty.   Why are you not simply reporting m_name regardless of whether it was truncated from `foo:bar` to `foo`.  The name `foo` is unique within the ELF you are getting in  `get_elf_identity_from_run()`.

src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -913,8 +914,8 @@ class elf_aie_gen2 : public elf_impl
   }
 
 public:
-  elf_aie_gen2(ELFIO::elfio&& elfio, elf::platform platform)
-    : elf_impl{std::move(elfio), platform}
+  elf_aie_gen2(ELFIO::elfio&& elfio, elf::platform platform, std::string path = {})
```
stsoe: Why default path?

src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -1352,8 +1353,8 @@ class elf_aie_gen2_plus : public elf_impl
   }
 
 public:
-  elf_aie_gen2_plus(ELFIO::elfio&& elfio, elf::platform platform)
-    : elf_impl{std::move(elfio), platform}
+  elf_aie_gen2_plus(ELFIO::elfio&& elfio, elf::platform platform, std::string path = {})
```
stsoe: Why default path?

src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -1463,19 +1464,19 @@ namespace {
 
 // Factory function - creates correct derived type based on platform
 static std::shared_ptr<xrt::elf_impl>
-create_elf_impl(ELFIO::elfio&& elfio)
+create_elf_impl(ELFIO::elfio&& elfio, std::string path = {})
```
stsoe: Why default path?

src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -1463,19 +1464,19 @@ namespace {
...
 {
   auto os_abi = elfio.get_os_abi();
 
   switch (os_abi) {
   case static_cast<uint8_t>(xrt::elf::platform::aie2p):
-    return std::make_shared<xrt::elf_aie_gen2>(std::move(elfio), xrt::elf::platform::aie2p);
+    return std::make_shared<xrt::elf_aie_gen2>(std::move(elfio), xrt::elf::platform::aie2p, std::move(path));
```
stsoe: You are always passing the path regardless of it being default or not, so don't make it default in the `elf_aie_gen2(_plus)` ctor.

src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -1463,19 +1464,19 @@ namespace {
...
   case static_cast<uint8_t>(xrt::elf::platform::aie2ps):
   case static_cast<uint8_t>(xrt::elf::platform::aie2ps_group):
   case static_cast<uint8_t>(xrt::elf::platform::aie4):
   case static_cast<uint8_t>(xrt::elf::platform::aie4a):
   case static_cast<uint8_t>(xrt::elf::platform::aie4z):
-    return std::make_shared<xrt::elf_aie_gen2_plus>(std::move(elfio), static_cast<xrt::elf::platform>(os_abi));
+    return std::make_shared<xrt::elf_aie_gen2_plus>(std::move(elfio), static_cast<xrt::elf::platform>(os_abi), std::move(path));
```
stsoe: ditto

src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -1501,7 +1502,7 @@ valid_or_error(const std::shared_ptr<elf_impl>& handle)
 
 elf::
 elf(const std::string& fnm)
-  : detail::pimpl<elf_impl>{create_elf_impl(load_elfio(fnm))}
+  : detail::pimpl<elf_impl>{create_elf_impl(load_elfio(fnm), fnm)}
```
stsoe, line 1511: Again, you always passing the fnm to create_elf_impl, so why make path a default argument in create_elf_impl?

src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -1630,6 +1631,12 @@ get_kernel_properties_and_args(std::shared_ptr<xrt::elf_impl> elf_impl,
   throw std::runtime_error("Kernel not found: " + kernel_name);
 }
 
+const std::string&
+get_filename(const std::shared_ptr<xrt::elf_impl>& elf_impl)
+{
+  return elf_impl->get_filename();
+}
```
stsoe:
```suggestion
const std::string&
get_filename(const xrt::elf_impl* elf_impl)
{
  return elf_impl->get_filename();
}
```

src/runtime_src/core/common/api/elf_int.h
```diff
@@ -287,10 +288,11 @@ class elf_impl
   static constexpr uint32_t addend_shift = 4;
   static constexpr uint32_t addend_mask = ~((uint32_t)0) << addend_shift;
   static constexpr uint32_t schema_mask = ~addend_mask;
+
   // NOLINTEND
 
-  // Protected constructor - takes already-loaded ELFIO and platform
-  elf_impl(ELFIO::elfio&& elfio, elf::platform platform);
+  // Protected constructor - takes already-loaded ELFIO, platform, and optional file path
+  elf_impl(ELFIO::elfio&& elfio, elf::platform platform, std::string path = {});
```
stsoe: Remove default, see other comments.

src/runtime_src/core/common/api/elf_int.h
```diff
@@ -468,6 +477,10 @@ std::pair<xrt_core::xclbin::kernel_properties, std::vector<xrt::xarg>>
 get_kernel_properties_and_args(std::shared_ptr<xrt::elf_impl> elf_impl,
                                const std::string& kernel_name);
 
+// Return the filename this ELF was loaded from (empty if loaded from buffer/stream).
+const std::string&
+get_filename(const std::shared_ptr<xrt::elf_impl>& elf_impl);
```
stsoe: Pass only shared ptr if you need to participate in ownership.

```suggestion
get_filename(const xrt::elf_imp* elf_impl);
```

stsoe (changes requested):
I feel error reporting is getting unwieldy with lots of AIE specific code in xrt_kernel.cpp.  But put that aside for now.  I added some general comments and suggestions.

src/runtime_src/core/common/api/xrt_elf.cpp
sayyanna: When an ELF is loaded from an in-memory buffer rather than the filesystem, its path is empty.

src/runtime_src/core/common/api/xrt_elf.cpp
sayyanna: When an ELF is loaded from an in-memory buffer rather than the filesystem, the path will be 
empty.

src/runtime_src/core/common/api/xrt_kernel.cpp
sayyanna: In full ELF flow, the ELF can contains multiple sub-kernels. Recently we noticed a PSU timeout which has many sub-kernels. Using the changes in this PR, we can find which kernel instance got timedout. Kernel Instance: DPU:**fa8c35f430e346a18e9a2464bf578d96_subgraph_32**

sayyanna:
> I feel error reporting is getting unwieldy with lots of AIE specific code in xrt_kernel.cpp. But put that aside for now. I added some general comments and suggestions.

Sure, I will refactor this code later to move AIE specific code to a separate file

PR #9773, opened 2026-05-05
AIESW-32028 Automatic core dump collection on xrt.ini key

stsoe:
Add `Runtime.aie_coredump_file` xrt.ini key to enable AIE coredump.

If specified in xrt.ini, then the hwctx is configured to enable aie coredump on command error (`ERT_CMD_STATE_TIMEOUT`).  The coredump, if succesfully generated, is dumped to specified file and process is aborted.

If UMD fails to generate a coredump (under xrt.ini condition) then error message is produced but standard error path is taken to throw an exception without aborting.

To configure the hwctx for coredump, XRT internally sets a configuration parameter (`aie_coredump = 1`) which is passed to UMD and then KMD.


rbramand-xilinx (approved):
Looks good

PR #9743, opened 2026-04-17
Add changes to support ELF files with merged sections

sayyanna:
#### Problem solved by the commit

https://amd.atlassian.net/wiki/spaces/AIE/pages/1582276490/Single+Section+Per-Column+ELF+Format#New:-Merged-Format

#### How problem was solved, alternative solutions (if any) and why they were rejected
Merge all pages for a given (column, [instance]) into a single contiguous ELF section
#### Risks (if any) associated the changes in the commit
ELF version is bumped up to support new changes. 
#### What has been tested and how, request additional testing if necessary
Tested on windows: 
ELFs with : os_abi 0x4b and abi 0x20 - full/config ELF (having .target in asm) 
                  os_abi 0x4b and abi 0x21 - full/config ELF (having .target in asm), with merged sections
                  os_abi 0x46, and abi 0x3 - full/config ELF 
                  os_abi 0x46, and abi 0x2 (xclbin +elf)

PR #9774, opened 2026-05-05
Fix get_column_and_page for partial elf control packet case

sayyanna:
get_column_and_page fails for partial ELF control packet case. 
https://github.com/Xilinx/XRT/pull/9743
Added proper logic to check merged and per-page sections
None
model tests test cases
partial ELF testcase


rbramand-xilinx (approved):
Looks good.
As discussed offline let us verify ctrlpkt partial ELF test case before merging. Thanks

PR #9778, opened 2026-05-06
dynamic tracing compiler json result support

AdvaitNaik:

Pull in https://github.com/Xilinx/aiebu/pull/278 and https://github.com/Xilinx/aiebu/pull/284 - dynamic tracing compiler refactor for json output support
Added `dtrace_output_json_format` (bool, default: false) to control the output format of tracing results.
When set to false (default), results are written to a Python file. When set to true, results are written in JSON format.
Updated the `dtrace_log_level` default from 1 to 0 and narrow the effective range to 0–2
Updated uc log schema to the latest version

Added flexibility to dump tracing result in json format while tracing/profiling

This is a feature enhancement over existing dynamic tracing

Low, this applies only when dynamic tracing is enabled

Tested by dumping tracing json result on windows medusa board and the feature works as expected

NA


stsoe (approved):
Looks much simpler now.  Thank you!

xuhz (approved):

PR #9782, opened 2026-05-08
AIESW-32262: skip connectivity check for imported buffer objects

bisingha-xilinx:
Imported BOs have pre-allocated physical backing from another device and cannot be re-mapped to a different memory bank. Skip the connectivity validation in validate_bo_at_index() for imported BOs to allow cross-device zero-copy buffer sharing (e.g., xDNA to PL) without triggering a spurious bank mismatch clone

Cross-device imported BOs (e.g., xDNA to PL) report bank index 0 because dma-buf carries no bank metadata. This triggers a spurious connectivity mismatch and buffer clone, defeating zero-copy.
AIESW-32262
Skip connectivity validation for imported BOs since their physical backing is fixed.
No risks

None.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -2290,8 +2290,10 @@ class run_impl : public std::enable_shared_from_this<run_impl>
   xrt::bo
   validate_bo_at_index(size_t index, const xrt::bo& bo)
   {
-    // ELF flow doesn't have arg connectivity, so skip validation
-    if (!kernel->get_xclbin())
+    // ELF flow doesn't have arg connectivity, so skip validation.
+    // Imported BOs have fixed physical backing from another device
+    // and bank re-mapping is not applicable.
+    if (!kernel->get_xclbin() || xrt_core::bo::is_imported(bo))
       return bo;
 
     // Check if connectivity validation should be skipped via INI option


chvamshi-xilinx (approved):
Please club multiple checks to a single check. Looks like we already have two checks where we are returning bo without checks.

stsoe (commented):
Couldn't this fail and error check in Alveo xclbin flow?  If a CU is connected to a specific memory bank, then the imported BO better be in that bank.   In Alveo flow, the driver is storing the bank where the BO is allocated and the exported BO should inherit these flags.

I understand the motivation here, and that this is not Alveo, but I think the check to disregard the bank check should be different somehow.  Not sure how?

chvamshi-xilinx:
> Couldn't this fail and error check in Alveo xclbin flow? If a CU is connected to a specific memory bank, then the imported BO better be in that bank. In Alveo flow, the driver is storing the bank where the BO is allocated and the exported BO should inherit these flags.
> 
> I understand the motivation here, and that this is not Alveo, but I think the check to disregard the bank check should be different somehow. Not sure how?

Yes Soren, But this check is causing a problem even if the imported buffer is coming from same bank. XRT/Driver doesn't know the bank information of imported_bo. Driver is returning "0" blindly(default bank) today due to which many tests passing, but if the corresponding argument is connected to some other bank, this check fails for imported_bo.

stsoe:
> > Couldn't this fail and error check in Alveo xclbin flow? If a CU is connected to a specific memory bank, then the imported BO better be in that bank. In Alveo flow, the driver is storing the bank where the BO is allocated and the exported BO should inherit these flags.
> > I understand the motivation here, and that this is not Alveo, but I think the check to disregard the bank check should be different somehow. Not sure how?
> 
> Yes Soren, But this check is causing a problem even if the imported buffer is coming from same bank. XRT/Driver doesn't know the bank information of imported_bo. Driver is returning "0" blindly(default bank) today due to which many tests passing, but if the corresponding argument is connected to some other bank, this check fails for imported_bo.

Which driver are we talking about?  AFAIK xocl is able to pass the bank along and now we are removing that check, right?

chvamshi-xilinx:
> > > Couldn't this fail and error check in Alveo xclbin flow? If a CU is connected to a specific memory bank, then the imported BO better be in that bank. In Alveo flow, the driver is storing the bank where the BO is allocated and the exported BO should inherit these flags.
> > > I understand the motivation here, and that this is not Alveo, but I think the check to disregard the bank check should be different somehow. Not sure how?
> > 
> > 
> > Yes Soren, But this check is causing a problem even if the imported buffer is coming from same bank. XRT/Driver doesn't know the bank information of imported_bo. Driver is returning "0" blindly(default bank) today due to which many tests passing, but if the corresponding argument is connected to some other bank, this check fails for imported_bo.
> 
> Which driver are we talking about? AFAIK xocl is able to pass the bank along and now we are removing that check, right?

I am talking about zocl, But even xocl should have returned some default bank as xocl also doesnt know the bank ID for the imported BOs.
As Driver is just getting fd and converting fd to dma_buf pointer, it doesn't know the bank information. Thatswhy we are removing this check in XRT core.

chvamshi-xilinx:
Merging this based on offline discussions with Soren.

PR #9784, opened 2026-05-08
[XRT-SMI] exporting get api for windows build

aktondak:
This PR exports the get_config api which is a requirement for windows MCDM shim to build 

None


rchane (approved):

src/runtime_src/core/common/smi/smi_ryzen.cpp
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
```
stsoe, line 2:
```suggestion
// Copyright (C) 2025-2026Advanced Micro Devices, Inc. All rights reserved.
```

PR #9792, opened 2026-05-12
CR-1268451 - Fix Segfault in libxrt_coreutil.so during dynamic xclbin swap on U55C/U50

karthdmg-xilinx:
caching hw contexts using std:map resulting in hw context alive even after coming out of scope in application.
this leads to failure when loading multiple xclbins from application.

replace with std:optional, pair so old hw context is destroyed when loading a new xclbin.

destruct old hw context if present wen loading a new hw context.

caching hw contexts using std:map resulting in hw context alive even after coming out of scope in application.
this leads to failure when loading multiple xclbins from application.

replace with std:optional, pair so old hw context is destroyed when loading a new xclbin.

None
Tested regression suite
NA


src/runtime_src/xrt/device/hal2.h
```diff
@@ -201,7 +202,9 @@ class device : public xrt_xocl::hal::device
   xrt::hw_context
   get_xrt_hwctx(const uuid& uuid) const override
   {
-    return m_hw_contexts.at(uuid);
+    if (!m_hw_ctx.has_value() || m_hw_ctx->first != uuid)
+      throw std::out_of_range("hal2::device::get_xrt_hwctx");
```
rbramand-xilinx: Can we have proper error message to understand what's wrong

src/runtime_src/xrt/device/hal2.cpp
```diff
@@ -276,7 +276,9 @@ loadXclBin(const xclBin* xclbin)
 {
   auto xclbin_obj = xrt::xclbin(reinterpret_cast<const axlf*>(xclbin));
   auto uuid = m_handle.register_xclbin(xclbin_obj);
-  m_hw_contexts[uuid] = xrt::hw_context(m_handle, uuid);
+  m_hw_ctx.emplace(std::piecewise_construct,
```
rbramand-xilinx: This can be simplified as below :

m_hw_ctx.reset();

m_hw_ctx.emplace(
    std::in_place,
    uuid,
    xrt::hw_context(m_handle, uuid)
);

src/runtime_src/xrt/device/hal2.h
```diff
@@ -70,11 +73,9 @@ class device : public xrt_xocl::hal::device
...
   mutable boost::optional<hal2::device_info> m_devinfo;
 
-  // In hwctx flow, we create hw contexts from xclbins on demand
-  // instead of explicitly loading the xclbins.  The hwctx are cached
-  // here and accessors are provided so that xclbin references can be
-  // converted into the hwctx that has loaded the xclbin.
-  std::map<xrt::uuid, xrt::hw_context> m_hw_contexts;
+  // One active xrt::hw_context per HAL session. emplace replaces the prior
+  // context so the driver slot is freed before a second xclbin load (e.g. kernel_swap).
+  std::optional<std::pair<xrt::uuid, xrt::hw_context>> m_hw_ctx;
```
stsoe: I don't understand what optional does in this case?   The change in this PR seems to allow just one hwctx so why optional over must xrt::hw_context?

src/runtime_src/xrt/device/hal2.h
```diff
@@ -201,7 +202,9 @@ class device : public xrt_xocl::hal::device
   xrt::hw_context
   get_xrt_hwctx(const uuid& uuid) const override
   {
-    return m_hw_contexts.at(uuid);
+    if (!m_hw_ctx.has_value() || m_hw_ctx->first != uuid)
+      throw std::out_of_range("hal2::device::get_xrt_hwctx");
+    return m_hw_ctx->second;
```
stsoe: Could this be just 
```
if (!m_hwctx)
  error ...;

if (m_hwctx.get_xclbin_uuid() != uuid)
  error ...;

return m_hwctx;
```

src/runtime_src/xrt/device/hal2.cpp
stsoe: How about just `m_hwctx = xrt::hw_context{...}` ?
Key is that this PR allows only on hwctx, which I don't if it is right or wrong?

src/runtime_src/xrt/device/hal2.cpp
```diff
@@ -276,7 +276,8 @@ loadXclBin(const xclBin* xclbin)
 {
   auto xclbin_obj = xrt::xclbin(reinterpret_cast<const axlf*>(xclbin));
   auto uuid = m_handle.register_xclbin(xclbin_obj);
-  m_hw_contexts[uuid] = xrt::hw_context(m_handle, uuid);
+  m_hw_ctx = {}; // clear active hw context
+  m_hw_ctx = xrt::hw_context(m_handle, uuid);
```
chvamshi-xilinx, line 280: If you just call **_m_hw_ctx =  xrt::hw_context(m_handle, uuid)_**; Is destructor not getting called before the constructor of new hw_context?
Is that the reason you are clearing and creating separately?

karthdmg-xilinx: the call to xrt::hw_context(m_handle, uuid) creating another hw_ctx before the old hw_ctx being destructed (from = op). this is resulting in failure of loading the second xclbin. so added m_hw_ctx = {} to destruct old before we cosntruct new one.

stsoe (approved):
LGTM, under the assumption that supporting one hwctx only is correct.

chvamshi-xilinx (approved):

PR #9793, opened 2026-05-12
[CR-1241855] graph.wait(cycles) doesn't trigger any exception

ManojTakasi:
[CR-1241855] graph.wait(cycles) doesn't catch the timeout exception but will force the graph to end
Add throw at the right function
low
Tested on both vck190 hw and hw_emu

### DIFF
--- a/src/runtime_src/core/edge/user/aie/graph_object.cpp
+++ b/src/runtime_src/core/edge/user/aie/graph_object.cpp
@@ -214,6 +214,7 @@ graph_object::wait_graph(uint64_t cycle)
 
     graph_api_obj->wait(cycle);
     state = graph_state::suspend;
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/edge/user/aie/graph_object.cpp
```diff
@@ -214,6 +214,7 @@ graph_object::wait_graph(uint64_t cycle)
 
     graph_api_obj->wait(cycle);
     state = graph_state::suspend;
+    throw xrt_core::error(-ETIME, "Wait graph '" + name + "' timeout");
```
chvamshi-xilinx, line 217: As discussed offline, this looks incorrect. Please check the call inside 215 line call

chvamshi-xilinx:
Closing as this is not proper fix.

PR #9794, opened 2026-05-12
AIESW-28041 Use env variable for is_advanced API

rchane:
https://jira.xilinx.com/browse/AIESW-28041
Add env variable check for both Windows and Linux
Initially wanted to do a sudo check on Linux, decided against during review. To be more platform agnostic, decided to make both Windows and Linux do an environment variable check for is_advanced and let further validation be done at the driver level.
Will need to update Linux xdna pipeline, Windows still accepts old flow but will eventually need to update as well.
Tested on Linux and Windows, works as desired.
Will update xrt-smi documentation.


aktondak (changes requested):
Do we need to check env in both linux and windows APIs separately?
Can we not add the env check upfront in xrt_core::sysinfo's is_advance() which is common to both platforms. This way we can use c++ std::getenv() instead of platform specific ones

aktondak (approved):

PR #9811, opened 2026-05-14
AIESW-32864: Fix temperature missing from platform JSON output

sandilya-xilinx:
AIESW-32864: Fix temperature missing from platform JSON output

Problem:
xrt-smi examine --report platform --format JSON produces a JSON file
with no temperature field, while the console output shows temperature
correctly.

Bug history:
Thermal data was never added to the platform property tree in
add_platform_info() for the ryzen case. The JSON output gap existed
since the initial implementation — temperature was only read inside
writeReport() for console display and was never included in the
property tree that gets serialized to JSON.

Root cause:
writeReport() called read_thermals() independently (console-only path).
The property tree built by platform_info() -> add_platform_info() had
no thermal data, so JSON serialization produced no temperature field.
Electrical power was already handled correctly via add_electrical_info()
in the ryzen case of add_platform_info().

Fix:
- info_platform.cpp: Add add_thermal_info() following the same pattern
  as add_electrical_info(). Call it in the ryzen case of
  add_platform_info() so thermal.temp_C is part of the canonical
  property tree used for both JSON and console output.
  Temperature decimal formatting is preserved as it applies inside
  sensor.cpp read_data_driven_thermals().

- ReportRyzenPlatform.cpp: Remove the direct read_thermals() call and
  sensor.h include. writeReport() now reads thermal.temp_C from the
  passed-in property tree, consistent with electrical.power_consumption_watts.

Alternative considered:
Inject thermal into the ptree inside getPropertyTree20202() in
ReportRyzenPlatform.cpp after calling device.get_info<platform>().
Rejected because info_platform.cpp is the architecturally correct
location — it is where all other platform data (electrical, clocks,
board info) is assembled, and fixes there automatically benefit all
callers of platform_info().

Risks:
Low. add_thermal_info() adds one additional read_thermals() driver
query per platform info call. If thermal sensors are unavailable,
xq::exception is caught and silently ignored — thermal is absent from
the output, same behavior as electrical when unavailable. No change
to existing data fields; thermal.temp_C is a new additive field.

Testing:
Validated on MDS hardware:
  xrt-smi examine --report platform --format JSON
  JSON output now includes "thermal": { "temp_C": "33.83" }
  Decimal formatting confirmed correct through new code path.

Documentation: None.

### DIFF
--- a/src/runtime_src/core/common/info_platform.cpp
+++ b/src/runtime_src/core/common/info_platform.cpp
@@ -350,12 +350,11 @@ add_thermal_info(const xrt_core::device* device, ptree_type& pt)
   try {
     ptree_type thermals_pt = xrt_core::sensor::read_thermals(device);
     const ptree_type& thermals = thermals_pt.get_child("thermals");
-    for (const auto& kv : thermals) {
-      const ptree_type& pt_temp = kv.second;
-      if (!pt_temp.get<bool>("is_present", false))
+    for (const auto& [key, child] : thermals) {
+      if (!child.get<bool>("is_present", false))
         continue;
       ptree_type thermal;
-      thermal.put("temp_C", pt_temp.get<std::string>("temp_C", "N/A"));
+      thermal.put("temp_C", child.get<std::string>("temp_C", "N/A"));
       pt.put_child("thermal", thermal);
       break;
     }
--- a/src/runtime_src/core/common/info_platform.cpp
+++ b/src/runtime_src/core/common/info_platform.cpp
@@ -345,6 +345,27 @@ add_electrical_info(const xrt_core::device* device, ptree_type& pt)
 }
 
 void
+add_thermal_info(const xrt_core::device* device, ptree_type& pt)
+{
+  try {
+    ptree_type thermals_pt = xrt_core::sensor::read_thermals(device);
+    const ptree_type& thermals = thermals_pt.get_child("thermals");
+    for (const auto& kv : thermals) {
+      const ptree_type& pt_temp = kv.second;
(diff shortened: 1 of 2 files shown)


src/runtime_src/core/common/info_platform.cpp
```diff
@@ -344,6 +344,27 @@ add_electrical_info(const xrt_core::device* device, ptree_type& pt)
...
+  try {
+    ptree_type thermals_pt = xrt_core::sensor::read_thermals(device);
+    const ptree_type& thermals = thermals_pt.get_child("thermals");
+    for (const auto& kv : thermals) {
+      const ptree_type& pt_temp = kv.second;
+      if (!pt_temp.get<bool>("is_present", false))
+        continue;
+      ptree_type thermal;
+      thermal.put("temp_C", pt_temp.get<std::string>("temp_C", "N/A"));
+      pt.put_child("thermal", thermal);
```
stsoe: Even as you don't use kv.first it is sometimes cleaner to use structured bindings for the for-loop.  Feel free to ignore.

```suggestion
    for (const auto& [key, child] : thermals) {
      if (!child.get<bool>("is_present", false))
        continue;
      ptree_type thermal;
      thermal.put("temp_C", child.get<std::string>("temp_C", "N/A"));
      pt.put_child("thermal", thermal);
```

stsoe (approved):

src/runtime_src/core/common/info_platform.cpp
sandilya-xilinx: Thank you @stsoe for the suggestion, adopted that change.

PR #9816, opened 2026-05-14
AIESW-30957 Limit certain xrt-smi examine queries to proper reports

rchane:
https://jira.xilinx.com/browse/AIESW-30957
Discovered during testing
get_available_devices had these metadata queries included, such as FW version/CERT version, that were being called with every time. This change limits those queries to only be done for host report and --version, where they are needed.
N/A
Tested on GPT
N/A


aktondak (changes requested):
Instead of introducing a default parameter which might get difficult to maintain. 
Let's instead introduce a new API on XBUtilities (maybe get_devices) which only does the getting devices part and skips filling up additional metadata. Then use that API wherever we want to get devices without filling any additional metadat.
We'll change less of existing code this way and will only make incremental changes

aktondak (approved):

aktondak:
Let's confirm with @diukirde that the current minimal implementation is not as expensive as now we only request for device class from driver.

diukirde:
> Let's confirm with @diukirde that the current minimal implementation is not as expensive as now we only request for device class from driver.

This looks good to me. I will verify this once the PR is merged.

PR #9820, opened 2026-05-18
AIESW-29594 Long process names cause formatting issues in the AIE partition report

rchane:
https://jira.xilinx.com/browse/AIESW-29594
Too long process name was right-shifting table output, discovered during internal testing. This problem was due to some hardcoded header lengths that were introduced when first changing aie-partitions report table output.
Introduce stacked functionality to Table2D, which uses some pre-existing Table2D methods to pre-calculate column width (and expand length to the longest element) before printing to output
N/A, does not affect pre-existing Table2D construction
```
AIE Partitions
  Total Memory Usage: 1088 MB
  Partition Index   : 0
    Columns: [0, 1, 2, 3, 4, 5, 6, 7]
    HW Contexts:
      |PID                             |Ctx ID    |Submissions  |Migrations   |Frame Evts  |Err  |Priority  |
      |Process Name                    |Status    |Completions  |Suspensions  |Layer Evts  |     |GOPS      |
      |Memory Usage                    |Instr BO  |             |             |            |     |FPS       |
      |                                |          |             |             |            |     |Latency   |
      |--------------------------------|----------|-------------|-------------|------------|-----|----------|
      |10800                           |1         |26           |0            |0           |0    |Normal    |
      |xrt-smi-testing-validation.exe  |Active    |25           |0            |0           |     |N/A       |
      |1088 MB                         |64 KB     |             |             |            |     |N/A       |
      |                                |          |             |             |            |     |N/A       |
      |--------------------------------|----------|-------------|-------------|------------|-----|----------|
```
N/A

### DIFF
--- a/src/runtime_src/core/tools/common/Table2D.cpp
+++ b/src/runtime_src/core/tools/common/Table2D.cpp
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright (C) 2022 Xilinx, Inc
-// Copyright (C) 2022-2024 Advanced Micro Devices, Inc. All rights reserved.
+// Copyright (C) 2022-2026 Advanced Micro Devices, Inc. All rights reserved.
 
 // ------ I N C L U D E   F I L E S -------------------------------------------
 // Local - Include Files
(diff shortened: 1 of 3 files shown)


src/runtime_src/core/tools/common/Table2D.cpp
```diff
@@ -17,6 +17,154 @@ Table2D::Table2D(const std::vector<HeaderData>& headers) : m_inter_entry_padding
...
+  for (size_t i = 0; i < column_count; ++i) {
+    HeaderData header;
+    header.name = "";
+    header.justification = justification;
+    addHeader(header);
+  }
+}
+
+void
+Table2D::updateStackedColumnWidth(size_t column_index, const std::string& data)
```
aktondak: nit.
But lets use snake case wherever we can in new code so we can align better with the XRT way of formatting code.

aktondak (approved):

PR #9830, opened 2026-05-25
xbmgmt2: use XBMGMT2_NAME in XRT_EDGE target_compile_definitions

### LINKED ISSUE #9829: xbmgmt2: configure fails with -DXRT_EDGE=ON due to wrong target variable
## Summary

`src/runtime_src/core/tools/xbmgmt2/CMakeLists.txt` references the wrong CMake variable in its `XRT_EDGE` branch, causing the configure step to fail for any build that sets `-DXRT_EDGE=ON`. The non-edge / native PCIe build path is unaffected. Reproduced against current `master` as well as the `202610.2.23.0_Canonical` tag.

## Reproduction

Configure XRT with `XRT_EDGE` enabled — anything that exercises the edge build path triggers it. Minimal repro on a native arm64 host (also reproduces via cross-compile or by manually forcing `-DXRT_EDGE=ON`):

```
$ cmake -S src -B build -DXRT_EDGE=ON -DXRT_AIE_BUILD=ON
...
-- add_subdirectory(xbmgmt2)
CMake Error at xrt/XRT/src/runtime_src/core/tools/xbmgmt2/CMakeLists.txt:68 (target_compile_definitions):
  Cannot specify compile definitions for target "PRIVATE" which is not built
  by this project.
```

## Root cause

`src/runtime_src/core/tools/xbmgmt2/CMakeLists.txt:68`:

```cmake
if (NOT XRT_EDGE)
  target_compile_definitions(${XBMGMT2_NAME} PRIVATE ENABLE_NATIVE_SUBCMDS_AND_REPORTS)
else()
  target_compile_definitions(${XBUTIL2_NAME} PRIVATE ENABLE_DEFAULT_ONE_DEVICE_OPTION)
endif()
```

`XBUTIL2_NAME` is set in the sibling `src/runtime_src/core/tools/xbutil2/CMakeLists.txt:63` (`set(XBUTIL2_NAME "xrt-smi")`) — that's a directory-local scope and is not visible inside `xbmgmt2/`. In this scope the variable expands to an empty string, and CMake parses the call as `target_compile_definitions(PRIVATE ENABLE_DEFAULT_ONE_DEVICE_OPTION)`, which fails because there's no target name.

The `if (NOT XRT_EDGE)` branch on line 66 correctly uses `${XBMGMT2_NAME}`, so this looks like a copy/paste from `xbutil2/CMakeLists.txt:78` where the variable rename was missed.

## Impact

- Any configure with `-DXRT_EDGE=ON` fails before generation completes — the build doesn't produce `xbmgmt`, `xrt-smi`, or any other target in `runtime_src`.
- Native PCIe builds (`-DXRT_EDGE` unset / `OFF`) are not affected: the else-branch isn't entered.
- Affects current `master` and the `202610.2.23.0_Canonical` release tag (and presumably every earlier release where the `XRT_EDGE` branch existed in this file).

## Suggested fix

```diff
--- a/src/runtime_src/core/tools/xbmgmt2/CMakeLists.txt
+++ b/src/runtime_src/core/tools/xbmgmt2/CMakeLists.txt
@@ -65,7 +65,7 @@ add_executable(${XBMGMT2_NAME} ${XBMGMT_V2_SRCS})
 if (NOT XRT_EDGE)
   target_compile_definitions(${XBMGMT2_NAME} PRIVATE ENABLE_NATIVE_SUBCMDS_AND_REPORTS)
 else()
-  target_compile_definitions(${XBUTIL2_NAME} PRIVATE ENABLE_DEFAULT_ONE_DEVICE_OPTION)
+  target_compile_definitions(${XBMGMT2_NAME} PRIVATE ENABLE_DEFAULT_ONE_DEVICE_OPTION)
 endif()
```

## Environment

- Distro: Ubuntu 26.04 LTS (arm64)
- CMake: 4.1.2
- XRT tag: `202610.2.23.0_Canonical` (also reproduced on `master` HEAD)
- Reproducer: `cmake -DXRT_EDGE=ON -DXRT_AIE_BUILD=ON` (sbuild / native build, no cross-compile required to trigger the configure-time error)

Reported while packaging XRT for Ubuntu (source package `xrt`).

talhaHavadar:
Fixes #9829.

## Summary

`src/runtime_src/core/tools/xbmgmt2/CMakeLists.txt` references `${XBUTIL2_NAME}` in its `XRT_EDGE` branch when calling `target_compile_definitions`. `XBUTIL2_NAME` is defined in the sibling `xbutil2/CMakeLists.txt` and is not visible in this directory scope, so it expands to an empty string. CMake then parses the call as `target_compile_definitions(PRIVATE ENABLE_DEFAULT_ONE_DEVICE_OPTION)` and aborts configure with:

```
CMake Error at src/runtime_src/core/tools/xbmgmt2/CMakeLists.txt:68 (target_compile_definitions):
  Cannot specify compile definitions for target "PRIVATE" which is not built
  by this project.
```

The non-edge branch above already uses the correct `${XBMGMT2_NAME}`. This change makes the edge branch consistent.

## Reproduction (before this PR)

```
cmake -S src -B build -DXRT_EDGE=ON -DXRT_AIE_BUILD=ON
```

Configure fails as above. With this patch, configure proceeds past xbmgmt2.

## Impact

- Configure fails for any build that sets `-DXRT_EDGE=ON` (arm64 / edge / Zynq).
- Native PCIe builds (`-DXRT_EDGE` unset or `OFF`) are unaffected — the else-branch is never entered.

## Notes

- One-line cmake fix, no behavioural change for the non-edge path.
- DCO sign-off added in commit.

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/CMakeLists.txt
+++ b/src/runtime_src/core/tools/xbmgmt2/CMakeLists.txt
@@ -65,7 +65,7 @@ add_executable(${XBMGMT2_NAME} ${XBMGMT_V2_SRCS})
 if (NOT XRT_EDGE)
   target_compile_definitions(${XBMGMT2_NAME} PRIVATE ENABLE_NATIVE_SUBCMDS_AND_REPORTS)
 else()
-  target_compile_definitions(${XBUTIL2_NAME} PRIVATE ENABLE_DEFAULT_ONE_DEVICE_OPTION)
+  target_compile_definitions(${XBMGMT2_NAME} PRIVATE ENABLE_DEFAULT_ONE_DEVICE_OPTION)
 endif()
 
 # Static build is a Linux / Ubuntu option only


chvamshi-xilinx (changes requested):
xbmgmt is not compiled for EDGE flows. I think, we should remove the else part completely.

src/runtime_src/core/tools/xbmgmt2/CMakeLists.txt
```diff
@@ -65,7 +65,7 @@ add_executable(${XBMGMT2_NAME} ${XBMGMT_V2_SRCS})
 if (NOT XRT_EDGE)
   target_compile_definitions(${XBMGMT2_NAME} PRIVATE ENABLE_NATIVE_SUBCMDS_AND_REPORTS)
 else()
```
ManojTakasi, line 67: Agree with removing else() entirely. Edge handling already lives in xbutil2/CMakeLists.txt. xbmgmt2 is only built for XRT_XRT/XRT_ALVEO, and no xbmgmt2 source uses that macro. so, we can drop this else()

PR #9834, opened 2026-05-25
Replace strcpy with memcpy

hyunjiki-amd:
To meet the driver compliance, banned APIs must be replaced according to https://learn.microsoft.com/en-us/windows-hardware/drivers/devtest/28719-banned-api-usage-use-updated-function-replacement

Issue fixed: driver compliance

Replace deprecated APIs in core/common (e.g., strncpy to memcpy)

None

Confirmed that the build is successful with the changes locally

None

Signed-off-by: Hyunji Kim [Hyunji.Kim@amd.com](mailto:Hyunji.Kim@amd.com)

### DIFF
--- a/src/runtime_src/core/edge/common_em/config.cxx
+++ b/src/runtime_src/core/edge/common_em/config.cxx
@@ -622,7 +622,7 @@ namespace xclemulation{
           if(mName.empty() == false)
           {
             if(strlen(mName.c_str()) < 256)//info.mName is static array of size 256
-              std::strcpy(info.mName, mName.c_str());
+              std::memcpy(info.mName, mName.c_str(), strlen(mName.c_str()) + 1);
           }
         }
         else if(prop.first == "HalMajorVersion")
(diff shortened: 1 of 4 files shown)


stsoe (commented):
Don't change Linux driver code, please

src/runtime_src/core/edge/common_em/config.cxx
```diff
@@ -622,7 +622,7 @@ namespace xclemulation{
           if(mName.empty() == false)
           {
             if(strlen(mName.c_str()) < 256)//info.mName is static array of size 256
-              std::strcpy(info.mName, mName.c_str());
+              std::memcpy(info.mName, mName.c_str(), strlen(mName.c_str()) + 1);
```
stsoe: This does not look the same.    I think you should replace with
```suggestion
              const auto len = std::min(mName.size(), sizeof(info.mName) - 1);
              std::memcpy(info.mName, mName.data(), len);
              info.mName[len] = '\0';
```
too bad it's so verbose, but so be it.

stsoe (changes requested):

PR #9835, opened 2026-05-25
Replace strncpy with memcpy

hyunjiki-amd:
To meet the driver compliance, banned APIs must be replaced according to https://learn.microsoft.com/en-us/windows-hardware/drivers/devtest/28719-banned-api-usage-use-updated-function-replacement

Issue fixed: driver compliance

Replace deprecated APIs (e.g., strncpy to memcpy)

None

Confirmed that the build is successful with the changes locally

None

Signed-off-by: Hyunji Kim [Hyunji.Kim@amd.com](mailto:Hyunji.Kim@amd.com)

### DIFF
--- a/src/runtime_src/hip/api/hip_device.cpp
+++ b/src/runtime_src/hip/api/hip_device.cpp
@@ -129,9 +129,7 @@ hip_get_device_properties(hipDeviceProp_t* props, hipDevice_t device)
   // Query device name using rom_vbnv
   // Copy device name to device_props.name, ensuring no buffer overflow
   auto name_str = (xrt_core::device_query<xrt_core::query::rom_vbnv>(device_handle));
-  size_t len = name_str.length();
-  if (len >= sizeof(device_props.name))
-    len = sizeof(device_props.name) - 1;
+  auto len = std::min(name_str.size(), sizeof(device_props.name) - 1);
   std::memcpy(device_props.name, name_str.c_str(), len);
   device_props.name[len] = '\0';
   // Extract and assign PCI domain, bus, and device IDs from the queried PCIe BDF tuple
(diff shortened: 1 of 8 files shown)


hyunjiki-amd:
Will remove src/runtime_src/core/edge/drm changes as they're Linux drivers https://github.com/Xilinx/XRT/blob/master/src/runtime_src/core/edge/drm/README

stsoe (commented):
Hi @hyunjiki-amd .  You cannot change Linux driver code.  So most of this PR is not valid.

src/runtime_src/hip/api/hip_device.cpp
```diff
@@ -134,8 +129,11 @@ hip_get_device_properties(hipDeviceProp_t* props, hipDevice_t device)
   // Query device name using rom_vbnv
   // Copy device name to device_props.name, ensuring no buffer overflow
   auto name_str = (xrt_core::device_query<xrt_core::query::rom_vbnv>(device_handle));
-  std::strncpy(device_props.name, name_str.c_str(), sizeof(device_props.name) - 1);
-  device_props.name[sizeof(device_props.name) - 1] = '\0';
+  size_t len = name_str.length();
+  if (len >= sizeof(device_props.name))
+    len = sizeof(device_props.name) - 1;
+  std::memcpy(device_props.name, name_str.c_str(), len);
+  device_props.name[len] = '\0';
```
stsoe:
```suggestion
auto len = std::min(name_str.size(), sizeof(device_props.name) - 1);
std::memcpy(device_props.name, name_str.data(), len);
device_props.name[len] = '\0';
```

stsoe (commented):
Please follow this pattern instead of if conditions
```
auto len = std::min(name_str.size(), sizeof(device_props.name) - 1);
std::memcpy(device_props.name, name_str.data(), len);
device_props.name[len] = '\0';
```

src/runtime_src/hip/api/hip_device.cpp
stsoe: Do not forget to sign all commits.  `git commit -s ...`

PR #9840, opened 2026-05-27
[XRT-SMI] Add resnet50 real model for strx and npu3 devices

aktondak:
This PR adds the resnet50 model as a runnable test in xrt-smi for both strx and npu3 devices. Currently these are added as advanced tests.

https://jira.xilinx.com/browse/AIESW-31781

None

Tested on strx and npu3 devices. 

None

### DIFF
--- a/src/runtime_src/core/common/smi/smi_ryzen.cpp
+++ b/src/runtime_src/core/common/smi/smi_ryzen.cpp
@@ -31,7 +31,7 @@ config_gen_ryzen()
     {"runlist-throughput", "Run end-to-end throughput test using runlist", "hidden"},
     {"df-bw", "Run bandwidth test on data fabric", "hidden"},
     {"gemm", "Measure the TOPS value of GEMM INT8operations", "common"},
-    {"resnet50", "Run ResNet-50 model and report latency and throughput", "common"},
+    {"resnet50", "Run ResNet-50 model and report latency and throughput", "hidden"},
     {"latency", "Run end-to-end latency test", "common"},
     {"quick", "Run a subset of four tests: \n1. latency \n2. throughput \n3. runlist-latency \n4. runlist-throughput", "hidden"},
     {"tct-all-col", "Measure average TCT processing time for all columns", "hidden"},
@@ -145,7 +145,7 @@ config_gen_npu3()
     {"tct-one-col", "Measure average TCT processing time for one column", "hidden"},
     {"tct-all-col", "Measure average TCT processing time for all columns", "hidden"},
     {"gemm", "Measure the TOPS value of GEMM INT8operations", "common"},
-    {"resnet50", "Run ResNet-50 model and report latency and throughput", "common"},
+    {"resnet50", "Run ResNet-50 model and report latency and throughput", "hidden"},
     {"preemption-overhead", "Measure preemption overhead at noop and memtile levels", "hidden"}
   };
 }
--- a/src/runtime_src/core/common/smi/smi_ryzen.cpp
+++ b/src/runtime_src/core/common/smi/smi_ryzen.cpp
@@ -31,6 +31,7 @@ config_gen_ryzen()
     {"runlist-throughput", "Run end-to-end throughput test using runlist", "hidden"},
     {"df-bw", "Run bandwidth test on data fabric", "hidden"},
     {"gemm", "Measure the TOPS value of GEMM INT8operations", "common"},
+    {"resnet50", "Run ResNet-50 model and report latency and throughput", "common"},
     {"latency", "Run end-to-end latency test", "common"},
     {"quick", "Run a subset of four tests: \n1. latency \n2. throughput \n3. runlist-latency \n4. runlist-throughput", "hidden"},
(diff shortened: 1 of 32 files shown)


src/runtime_src/core/common/smi/smi_ryzen.cpp
```diff
@@ -31,6 +31,7 @@ config_gen_ryzen()
     {"runlist-throughput", "Run end-to-end throughput test using runlist", "hidden"},
     {"df-bw", "Run bandwidth test on data fabric", "hidden"},
     {"gemm", "Measure the TOPS value of GEMM INT8operations", "common"},
+    {"resnet50", "Run ResNet-50 model and report latency and throughput", "hidden"},
```
sonals: Please do not use model name as command name. Name it **sanity**, **smoke**, or **basic**.

src/runtime_src/core/tools/common/tests/TestResnet50.cpp
```diff
@@ -0,0 +1,144 @@
...
+    "ifm_resnet50_npu3.bin",
+    "wts_resnet50_npu3.bin",
+    "ctrl_pkt0_resnet50_npu3.bin",
+    "ofm_resnet50_npu3.bin",
+  });
+  run_resnet50(dev, recipe_data, profile_data, artifacts_repo, ptree);
+}
+
+TestResnet50::TestResnet50()
+  : TestRunner("resnet50", "Run ResNet-50 model and report latency and throughput")
```
sonals: See comment above about command name

src/runtime_src/core/tools/common/tests/TestSanity.cpp
```diff
@@ -0,0 +1,144 @@
...
+                           const std::string& recipe_data,
+                           const std::string& profile_data,
+                           const xrt_core::runner::artifacts_repository& artifacts_repo,
+                           boost::property_tree::ptree& ptree) const
+{
+  xrt_core::runner runner(xrt::device(dev), recipe_data, profile_data, artifacts_repo);
+  runner.execute();
+  runner.wait();
+  log_runner_report(ptree, runner.get_report());
+  ptree.put("status", XBValidateUtils::test_token_passed);
```
sonals, line 25: We are blindly setting the status as success here. Are we relying on runner throwing an exception if the test fails?

aktondak: We're relying on runner's validation with validation set to true in profile.json
In case validation fails, runner will throw a validation_error and fail the xrt-smi test.

src/runtime_src/core/tools/common/tests/TestResnet50.cpp
```diff
@@ -0,0 +1,144 @@
...
+    return;
+
+  const auto& cpu = exec_report["cpu"];
+  if (cpu.contains("throughput"))
+    XBValidateUtils::logger(ptree, "Details",
+      "Average throughput: " + std::to_string(cpu["throughput"].get<double>()) + " op/s");
+  if (cpu.contains("latency"))
+    XBValidateUtils::logger(ptree, "Details",
+      "Average latency: " + std::to_string(cpu["latency"].get<double>()) + " us");
+}
```
sonals: We just need passed failed for this test. There is no need to report metrics.

src/runtime_src/core/common/smi/smi_ryzen.cpp
aktondak: renamed to sanity

src/runtime_src/core/tools/common/tests/TestSanity.cpp
```diff
@@ -0,0 +1,102 @@
...
+  });
+  run_sanity(dev, recipe_data, profile_data, artifacts_repo, ptree);
+}
+
+TestSanity::TestSanity()
+  : TestRunner("sanity", "Run a small model and validate sanity of the device")
+{}
+
+boost::property_tree::ptree
+TestSanity::run(const std::shared_ptr<xrt_core::device>&)
```
sonals: What is the unnamed argument for? Can we update the interface to drop it altogether?

aktondak: The run API with only device ptr is a pure virtual function in the base class. 
I introduced another base class API "run" while doing runner migration with device and archive ptr arguments for all tests using runner and archive flow. 
The unnamed argument path is an error path meaning some tests (like this) cannot be run without an archive pointer. I can remove this by changing the pure virtual in base to not being pure virtual and having a default implementation.
But i dont want to change the base class.

sonals: Please file a JIRA to resolve this in a future PR.

sonals (approved):

src/runtime_src/core/tools/common/tests/TestSanity.cpp
aktondak: I have fixed it in this PR by making the base class non pure virtual and implementing a default API.
Also removed all subsequent APIs which are not required anymore

rchane (approved):

PR #9842, opened 2026-06-02
AIESW-35018 Fix failure token for xrt-smi preemption overhead test

rchane:
https://jira.xilinx.com/browse/AIESW-35018
Discovered in internal testing
Test was always passing even after failures due to test_token_passed being set after cleanup. Now, we check for any failures before setting to passed after cleanup.
N/A
Testing on GPT.
N/A

### DIFF
--- a/src/runtime_src/core/tools/common/tests/TestPreeemptionOverhead.cpp
+++ b/src/runtime_src/core/tools/common/tests/TestPreeemptionOverhead.cpp
@@ -89,13 +89,16 @@ run_strix(const std::shared_ptr<xrt_core::device>& dev, const xrt_core::archive*
   catch(const std::exception& e) {
     XBValidateUtils::logger(ptree, "Error", e.what());
     ptree.put("status", XBValidateUtils::test_token_failed);
+
+    // Restore the original preemption state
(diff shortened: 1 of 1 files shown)


aktondak (changes requested):
Why cant we do an early exit at line 118 when the test fails? This if check is unnecessary. In case the test fails, we should disable preemption, log error and return.

rchane:
> Why cant we do an early exit at line 118 when the test fails? This if check is unnecessary. In case the test fails, we should disable preemption, log error and return.

We need to restore the original preemption state regardless of failure, so there would be repetitive code in that case (restoration both in and out of the catch). In the case there is an early failure before the restoration, we still need to restore after.

aktondak:
> > Why cant we do an early exit at line 118 when the test fails? This if check is unnecessary. In case the test fails, we should disable preemption, log error and return.
> 
> We need to restore the original preemption state regardless of failure, so there would be repetitive code in that case (restoration both in and out of the catch).

That is fine. Its still a single execution at runtime (either the catch code path kicks in or the one following it). String based if checks are expensive at runtime and difficult to reason about too later. We should always optimize for runtime

aktondak (approved):

PR #9843, opened 2026-06-02
[XRT-SMI] Update the xrt-smi test to run in pure throughput mode

aktondak:
This PR updates the xrt-smi throughput test to run in pure throughput mode and changes the code to query throughput from runner.
This PR also removes the advance subcommand from ryzen specific xrt-smi as it is not applicable anymore

https://jira.xilinx.com/browse/AIESW-34896
https://jira.xilinx.com/browse/AIESW-35013

None

Tested on strx and npu3 boards : 
```
C:\Users\aktondak>xrt-smi validate --advanced -r throughput
-------------------------------------------------------------------------
                    DISCLAIMER  (xrt-smi --advanced)
You are running a developer command that may change system configuration.
                Continue only if you understand the risks.
-------------------------------------------------------------------------
    Platform              : NPU Medusa
    Power Mode            : performance
Test 1 [00c5:00:01.1]     : throughput
    Details               : Average throughput: 77703.0 op/s
Validation completed
```

None


src/runtime_src/core/common/smi/smi_ryzen.cpp
```diff
@@ -193,7 +180,6 @@ populate_smi_instance(xrt_core::smi::smi* smi_instance, const xrt_core::device*
     smi_instance->add_subcommand("validate",  generator->create_validate_subcommand());
     smi_instance->add_subcommand("examine",  generator->create_examine_subcommand());
     smi_instance->add_subcommand("configure",  generator->create_configure_subcommand());
-    smi_instance->add_subcommand("advanced",  generator->create_advanced_subcommand());
```
sonals, line 196: Could this break user's existing scripts?

src/runtime_src/core/tools/common/tests/TestNPUThroughput.cpp
```diff
@@ -14,6 +14,18 @@
 using json = nlohmann::json;
 #include <filesystem>
 
+namespace {
+
+double
+get_throughput_ops(const json& report)
+{
+  if (report.contains("executions") && !report["executions"].empty())
+    return report["executions"][0]["cpu"]["throughput"].get<double>();
```
sonals: We are assuming the index 0 is throughput run. Should we add some sanity checks here to ensure it is indeed the case?

src/runtime_src/core/common/smi/smi_ryzen.cpp
aktondak: No this will not break it since I was made aware that this is redundant and not being used in ryzen specific usage of xrt-smi. 
This is only relevant to alveo which this code change doesn't alter.

src/runtime_src/core/tools/common/tests/TestNPUThroughput.cpp
aktondak: I have added a throw in-case the profile.json contains more than 1 execution runs.

src/runtime_src/core/common/smi/smi_ryzen.cpp
sonals: Can we leave it there with a comment in the code and if used by a user, generate a warning that this feature has been deprecated? That way we will prevent accidental breakage of any user scripts/automation.

rchane (approved):

src/runtime_src/core/common/smi/smi_ryzen.cpp
aktondak: I believe that's not ideal because the Subcommand code is common between ryzen and alveo. If we add deprecated warning in the common code, it will break "xrt-smi advanced" for alveo. 
"xrt-smi advanced" was never meant for ryzen devices and above code change materializes it. There are no further subcommands like "xrt-smi advanced -r xxxx" which are applicable for ryzen. 
Referring to @rchane and @AShivangi in case I missed any assumption as this sub option predates me.

sonals (approved):

AShivangi (approved):

sonals (approved):

PR #9845, opened 2026-06-03
AIESW-27330 : Add different sinks for uc log buffer

rbramand-xilinx:
The uC log buffer dumper previously only supported writing parsed log entries to files. There was no way to route uC logs to the OS system log (syslog on Linux, Windows Event Log on Windows) or to console, making it difficult to correlate uC log output with other system-level logs in production environments or automated test pipelines where file-based output is inconvenient.

 This is a new feature extending the existing buffer_dumper infrastructure to dump to new sinks (syslog or console)

A new **Debug.uc_log_dump** key is added to xrt.ini supporting four values:

file or "" (default) — existing behavior, writes timestamped per-chunk files
syslog — routes parsed log lines through the OS system log: Linux syslog/journald via openlog("XRT", ...) / syslog(), Windows Application Event Log via RegisterEventSourceA("AMD_XRT") / ReportEventA()
console — writes to stderr
null  — silently discards all uC log output
A dedicated send_uc_log() function is added to message.cpp/message.h that creates its own dispatcher from the uc_log_dump ini value, completely independent of runtime_log.
Parsed log entries are split line-by-line before dispatch so each line becomes a discrete syslog/Event Log entry.

The file stream in dump_chunk_data() is now opened lazily only when the sink requires it (file or binary format), so no files are created when syslog, console, or null is configured.

TODO : At present log from uC is not tagged with message type like info, warning or error, we send all the uC log entries with level info to syslog. Need to get the correct tag from uC logs so that we can send the entries to syslog with proper severity level

Low as it can be enabled only with ini option and default behavior is not changed.

Tested uC log dumping on Linux and windows and things work as expected

Added comments where ever necessary to explain different sinks

### DIFF
--- a/src/runtime_src/core/common/buffer_dumper.cpp
+++ b/src/runtime_src/core/common/buffer_dumper.cpp
@@ -93,20 +93,30 @@ void
 buffer_dumper::
 dump_chunk_data(size_t chunk_index, size_t start, size_t length, uint8_t* chunk)
 {
-  std::ofstream& fs = get_or_open_stream(chunk_index);
+  // Determine if we need to open a file stream based on the sink configuration
+  // Binary dump always writes to file
+  // Default behavior is to write parsed logs to file
+  const bool use_file_sink = m_config.dump_bin_format
+                             || m_config.uc_log_dump == "file"
+                             || m_config.uc_log_dump.empty();
 
-  // Check if stream is in good state before writing
-  if (!fs.good()) {
-    throw std::runtime_error("File stream for chunk " + std::to_string(chunk_index) +
-                             " is in bad state before write");
+  // Open a file stream only when the sink requires it
+  std::ofstream* fs_ptr = nullptr;
+  if (use_file_sink) {
+    std::ofstream& fs = get_or_open_stream(chunk_index);
+    if (!fs.good())
+      throw std::runtime_error("File stream for chunk " + std::to_string(chunk_index) +
+                               " is in bad state before write");
+    fs_ptr = &fs;
   }
 
   // Calculate start offset and bytes to end for circular buffer wrapping
   const size_t start_offset = (start % m_data_size) + m_config.metadata_size;
   const size_t bytes_to_end = m_config.chunk_size - start_offset;
 
-  // UC log dump in binary format
+  // uC log dump in binary format (always uses file regardless of uc_log_dump setting)
   if (m_config.dump_bin_format) {
+    std::ofstream& fs = *fs_ptr;
     // Always write/update metadata since this function is called when there's new data
     // and metadata(count) gets updated when there is new data
     // Seek to beginning and write/update metadata
@@ -222,12 +232,29 @@ dump_chunk_data(size_t chunk_index, size_t start, size_t length, uint8_t* chunk)
     // log entries may have been lost at the boundary before the next read.
     parsed_stream << "[0.000000000] [CERT] [Dumper]--------------[Separator]--------------\n";
 
-    // Append parsed output to file
-    fs.seekp(0, std::ios::end);
-    fs << parsed_stream.str();
-
-    if (!fs)
-      throw std::runtime_error("Failed to write parsed UC log for chunk " + std::to_string(chunk_index));
+    // Dispatch parsed output to the configured sink
+    const std::string& sink = m_config.uc_log_dump;
+    if (sink == "file" || sink.empty()) {
+      // Write to file
+      fs_ptr->seekp(0, std::ios::end);
+      *fs_ptr << parsed_stream.str();
+      if (!*fs_ptr)
+        throw std::runtime_error("Failed to write parsed UC log for chunk " + std::to_string(chunk_index));
+    }
+    else if (sink != "null") {
+      // syslog or console: send each line as a separate message so the sink can
+      // format/route it correctly
+      std::istringstream lines(parsed_stream.str());
+      std::string line;
+      while (std::getline(lines, line)) {
+        if (!line.empty())
+          // TODO: Pass decoded severity from uC log entry. For now all entries
+          // are sent with info severity. Also may be pass uC index in tag for better traceability.
+          xrt_core::message::send_uc_log(sink, xrt_core::message::severity_level::info,
+                                         "xrt_uc_log", line.c_str());
+      }
+    }
+    // sink == "null": discard
   }
   catch (const std::exception& e) {
     // Log parsing error, log warning and continue
@@ -235,9 +262,11 @@ dump_chunk_data(size_t chunk_index, size_t start, size_t length, uint8_t* chunk)
                             std::string{"UC log parsing failed: "} + e.what());
   }
 
-  fs.flush();
-  if (!fs)
-    throw std::runtime_error("Failed to flush chunk " + std::to_string(chunk_index));
+  if (fs_ptr) {
+    fs_ptr->flush();
+    if (!*fs_ptr)
+      throw std::runtime_error("Failed to flush chunk " + std::to_string(chunk_index));
+  }
 }
 
 void
--- a/src/runtime_src/core/common/buffer_dumper.cpp
+++ b/src/runtime_src/core/common/buffer_dumper.cpp
@@ -91,72 +91,146 @@ read_logged_count(uint8_t* chunk)
 
 void
 buffer_dumper::
-dump_chunk_data(size_t chunk_index, size_t start, size_t length, uint8_t* chunk)
+dump_chunk_data_binary(size_t chunk_index, size_t start_offset, size_t bytes_to_end,
+                       size_t length, uint8_t* chunk)
 {
-  // Determine if we need to open a file stream based on the sink configuration
-  // Binary dump always writes to file
-  // Default behavior is to write parsed logs to file
-  const bool use_file_sink = m_config.dump_bin_format
-                             || m_config.uc_log_dump == "file"
-                             || m_config.uc_log_dump.empty();
+  std::ofstream& fs = get_or_open_stream(chunk_index);
+  if (!fs.good())
+    throw std::runtime_error("File stream for chunk " + std::to_string(chunk_index) +
+                             " is in bad state before write");
 
-  // Open a file stream only when the sink requires it
-  std::ofstream* fs_ptr = nullptr;
-  if (use_file_sink) {
+  // Rewrite metadata at the start on every call: the count field in the metadata
+  // is updated by the device each time new data is written, so we must keep it
+  // in sync with the appended payload.
+  fs.seekp(0);
+  fs.write(reinterpret_cast<const char*>(chunk),
+           static_cast<std::streamsize>(m_config.metadata_size));
+  if (!fs)
+    throw std::runtime_error("Failed to write metadata for chunk " + std::to_string(chunk_index));
+
+  // Append payload after existing file content
+  fs.seekp(0, std::ios::end);
+
+  if (length <= bytes_to_end) {
+    // Payload fits before the circular buffer wrap point; write in one shot
+    fs.write(reinterpret_cast<const char*>(chunk + start_offset),
+             static_cast<std::streamsize>(length));
+    if (!fs)
+      throw std::runtime_error("Failed to write " + std::to_string(length) +
+                               " bytes to chunk " + std::to_string(chunk_index));
+  }
+  else {
+    // Payload straddles the wrap point; write the two contiguous segments separately
+    fs.write(reinterpret_cast<const char*>(chunk + start_offset),
+             static_cast<std::streamsize>(bytes_to_end));
+    if (!fs)
+      throw std::runtime_error("Failed to write first part (" + std::to_string(bytes_to_end) +
+                               " bytes) to chunk " + std::to_string(chunk_index));
+
+    fs.write(reinterpret_cast<const char*>(chunk + m_config.metadata_size),
+             static_cast<std::streamsize>(length - bytes_to_end));
+    if (!fs)
+      throw std::runtime_error("Failed to write wrapped part (" +
+                               std::to_string(length - bytes_to_end) +
+                               " bytes) to chunk " + std::to_string(chunk_index));
+  }
+
+  fs.flush();
+  if (!fs)
+    throw std::runtime_error("Failed to flush chunk " + std::to_string(chunk_index));
+}
+
+std::string
+buffer_dumper::
+parse_log_entries(size_t start_offset, size_t bytes_to_end, size_t length,
+                  uint8_t* chunk)
+{
+  std::ostringstream out;
+  size_t parsed_bytes = 0;
+
+  while (parsed_bytes < length) {
+    // Resolve the byte offset of the current entry, accounting for circular buffer wrap
+    size_t entry_offset = (parsed_bytes < bytes_to_end)
+      ? start_offset + parsed_bytes
+      : m_config.metadata_size + (parsed_bytes - bytes_to_end);
+
+    log_entry log;
+    std::memcpy(&log, chunk + entry_offset, sizeof(log_entry));
+
+    // Look up format string in the schema; fall back to a generic placeholder
+    // when the log_id is unrecognized. The fallback is chosen by argument count
+    // (derived from entry length) so the output still shows available data.
+    constexpr std::array<const char*, 3> default_formats = {
+      "unknown !\n",
+      "unknown %d !!\n",
+      "unknown %d unknown %d !!!\n"
+    };
+
+    auto log_schema_it = uc_log_schema.logs.find(log.log_id);
+    const char* log_format = (log_schema_it != uc_log_schema.logs.end())
+      ? log_schema_it->second.c_str()
+      : default_formats[std::min(
+          static_cast<std::size_t>(
+              log.length - (offsetof(log_entry, argument1) / sizeof(uint32_t))),
+          default_formats.size() - 1)];
+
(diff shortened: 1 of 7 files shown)


src/runtime_src/core/common/message.cpp
```diff
@@ -223,4 +223,13 @@ sendv(severity_level l, const char* tag, const char* format, va_list args)
...
 }
 
+void
+send_uc_log(const std::string& sink, severity_level l, const char* tag, const char* msg)
+{
+  // Dedicated dispatcher for uC log output, driven by Debug.uc_log_dump (not runtime_log).
+  // No verbosity filtering — all uC log lines are forwarded unconditionally.
+  static auto dispatcher = message_dispatch::make_dispatcher(sink);
+  dispatcher->send(l, tag, msg);
+}
```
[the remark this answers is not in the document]
rbramand-xilinx: This is the expected behavior as ini option remains same throughout the lifetime of the program

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -93,20 +93,30 @@ void
...
-  // Check if stream is in good state before writing
-  if (!fs.good()) {
-    throw std::runtime_error("File stream for chunk " + std::to_string(chunk_index) +
-                             " is in bad state before write");
+  // Determine if we need to open a file stream based on the sink configuration
+  // Binary dump always writes to file
+  // Default behavior is to write parsed logs to file
+  const bool use_file_sink = m_config.dump_bin_format
+                             || m_config.uc_log_dump == "file"
+                             || m_config.uc_log_dump.empty();
```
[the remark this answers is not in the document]
rbramand-xilinx: This is also the expected behavior as uC log dumping is enabled using uc_log option and uc_log_dump option only provides details about sink and redirecting the log to file when the option is not set or empty maintains backward compatibilty.

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -222,22 +232,41 @@ dump_chunk_data(size_t chunk_index, size_t start, size_t length, uint8_t* chunk)
...
+      std::string line;
+      while (std::getline(lines, line)) {
+        if (!line.empty())
+          // TODO: Pass decoded severity from uC log entry. For now all entries
+          // are sent with info severity. Also may be pass uC index in tag for better traceability.
+          xrt_core::message::send_uc_log(sink, xrt_core::message::severity_level::info,
+                                         "xrt_uc_log", line.c_str());
+      }
+    }
+    // sink == "null": discard
```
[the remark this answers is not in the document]
rbramand-xilinx: As mentioned in earlier comment this is expected to maintain backward compatibility

src/runtime_src/core/common/config_reader.h
```diff
@@ -1242,6 +1242,19 @@ get_uc_log_size_per_uc_kb()
...
+inline std::string
+get_uc_log_dump()
+{
+  // Sink for parsed uC log output: "file" (default), "syslog", "console", or "null"
+  // "file"    -> write to timestamped per-chunk files
+  // "syslog"  -> route to OS system log (Linux syslog / Windows Event Log under AMD_XRT)
+  // "console" -> write to stderr
+  // "null"    -> discard
+  // Note: uc_log_bin_format=true always writes raw binary to 'file' regardless of this setting.
+  static std::string value = detail::get_string_value("Debug.uc_log_dump", "file");
```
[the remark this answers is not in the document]
rbramand-xilinx: No this is intentional as to maintain backward compatibility

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -222,22 +232,41 @@ dump_chunk_data(size_t chunk_index, size_t start, size_t length, uint8_t* chunk)
...
+      if (!*fs_ptr)
+        throw std::runtime_error("Failed to write parsed UC log for chunk " + std::to_string(chunk_index));
+    }
+    else if (sink != "null") {
+      // syslog or console: send each line as a separate message so the sink can
+      // format/route it correctly
+      std::istringstream lines(parsed_stream.str());
+      std::string line;
+      while (std::getline(lines, line)) {
+        if (!line.empty())
```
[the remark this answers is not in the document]
rbramand-xilinx: This is in accordance with existing codebase where single line in conditional statement wont have braces

src/runtime_src/core/common/detail/linux/syslog.h
```diff
@@ -6,6 +6,7 @@
 #include "core/common/message.h"
```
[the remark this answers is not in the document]
rbramand-xilinx: CMake file has proper include_directories added and build is passing. So this is not an issue as relative path is included.

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -93,20 +93,30 @@ void
...
   }
 
   // Calculate start offset and bytes to end for circular buffer wrapping
   const size_t start_offset = (start % m_data_size) + m_config.metadata_size;
   const size_t bytes_to_end = m_config.chunk_size - start_offset;
 
-  // UC log dump in binary format
+  // uC log dump in binary format (always uses file regardless of uc_log_dump setting)
   if (m_config.dump_bin_format) {
+    std::ofstream& fs = *fs_ptr;
```
stsoe: Isn't this a nullptr if use_sink_file is false?

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -93,20 +93,30 @@ void
 buffer_dumper::
 dump_chunk_data(size_t chunk_index, size_t start, size_t length, uint8_t* chunk)
```
stsoe: Feels like a function that does to much.  Not sure if it can be refactored a bit?

src/runtime_src/core/common/buffer_dumper.cpp
stsoe: Ah, no it's not, but it underscores the complexity of this function.

src/runtime_src/core/common/buffer_dumper.cpp
rbramand-xilinx: Yes Soren, I have refactored the function and split it. Please review. Thanks

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -91,153 +91,190 @@ read_logged_count(uint8_t* chunk)
...
-      }
-
-      parsed_bytes += m_config.metadata_size;
-    }
+  // syslog or console: each line is dispatched as a discrete message so the
+  // sink can apply its own formatting and routing per entry
+  std::istringstream lines(text);
+  std::string line;
+  while (std::getline(lines, line)) {
+    if (!line.empty())
```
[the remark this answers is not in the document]
rbramand-xilinx: Its as expected and follows coding style of this repo. Ignore

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -91,153 +91,191 @@ read_logged_count(uint8_t* chunk)
...
```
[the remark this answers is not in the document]
rbramand-xilinx: Ignore as code is as per guidelines of this repo

src/runtime_src/core/common/buffer_dumper.cpp
rbramand-xilinx: I have split the function now to make things clear Soren, please review

stsoe (approved):

src/runtime_src/core/common/message.cpp
chvamshi-xilinx, line 233: The static dispatcher here means the sink is determined once (on first call) and cached for the process lifetime. This is fine given the ini value is static, but worth noting that `make_dispatcher` returns a raw `message_dispatch*` (allocated with `new`). Since this is a `static` local, it's intentionally leaked (process-lifetime singleton). Consider wrapping in `std::unique_ptr` for clarity of ownership, or add a comment clarifying this is a deliberate process-lifetime allocation.

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -91,153 +91,191 @@ read_logged_count(uint8_t* chunk)
 
 void
 buffer_dumper::
-dump_chunk_data(size_t chunk_index, size_t start, size_t length, uint8_t* chunk)
+dump_chunk_data_binary(size_t chunk_index, size_t start_offset, size_t bytes_to_end,
```
chvamshi-xilinx, line 94: The refactored `dump_chunk_data` is much cleaner now — the separation into `dump_chunk_data_binary`, `parse_log_entries`, and `dispatch_parsed_log` makes the flow easy to follow. Good refactoring per Soren's feedback.

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -91,153 +91,191 @@ read_logged_count(uint8_t* chunk)
...
+void
+buffer_dumper::
+dispatch_parsed_log(size_t chunk_index, const std::string& text)
+{
+  const std::string& sink = m_config.uc_log_dump;
+
+  // "null" sink: silently discard
+  if (sink == "null")
+    return;
+
```
chvamshi-xilinx, line 222: In `dispatch_parsed_log`, the file sink path opens a file stream even for syslog/console/null sinks because `get_or_open_stream` is only guarded by the `if (sink == "file" || sink.empty())` check. This is correct as written.

However, if `m_file_streams` vector is pre-allocated in the constructor (which it is — `m_file_streams(m_config.num_chunks)`), when the sink is "syslog" or "console", those streams remain default-constructed (not opened) throughout the lifetime, which is a minor wasted allocation. Not a bug, but consider conditionally allocating the file streams vector only when the sink requires files (i.e., when `uc_log_dump == "file"` or `dump_bin_format == true`).

src/runtime_src/core/common/buffer_dumper.cpp
chvamshi-xilinx, line 247: The line-by-line dispatch via `std::getline` + `send_uc_log` for syslog/console is the right approach — each syslog entry should be a discrete message.

Minor concern: if `text` contains a large number of lines (e.g., a large dump), this loop will call `send_uc_log` many times in quick succession. Since the `send_uc_log` dispatcher is a static singleton (no per-call locking internally for syslog), this should be fine for syslog. But for the console dispatcher, each call acquires a mutex lock (see `console_dispatch::send`). If performance under high log volume becomes a concern, consider batching console output. Not blocking, just something to keep in mind for the future.

src/runtime_src/core/common/detail/linux/syslog.h
```diff
@@ -14,7 +15,7 @@ class syslog_dispatch : public message_dispatch
 {
 public:
   syslog_dispatch()
-  { openlog("sdaccel", LOG_PID|LOG_CONS, LOG_USER); }
+  { openlog("XRT", LOG_PID|LOG_CONS, LOG_USER); }
```
chvamshi-xilinx, line 17: Good change updating the syslog ident from the legacy "sdaccel" to "XRT". This makes log filtering via `journalctl -t XRT` straightforward.

The tag inclusion in the message (`[tag] : msg`) is a nice addition for distinguishing different XRT subsystems in syslog output. One consideration: the syslog format string `"%s"` prevents format-string injection, which is correct.

src/runtime_src/core/common/detail/linux/syslog.h
```diff
@@ -26,7 +27,10 @@ class syslog_dispatch : public message_dispatch
 
   void
   send(severity_level l, const char* tag, const char* msg) override
-  { syslog(m_severity_map[l], "%s", msg); }
+  {
```
chvamshi-xilinx, line 29: The `std::string` concatenation here (`std::string("[") + tag + "] : " + msg`) creates temporary allocations on every syslog call. For a hot path (high-frequency uC log entries), consider using a pre-sized `std::string` with `reserve()` or a stack-allocated buffer:

```cpp
char buf[512];
std::snprintf(buf, sizeof(buf), "[%s] : %s", tag, msg);
syslog(m_severity_map[l], "%s", buf);
```

This avoids heap allocations in the logging path. Not critical since uC log dumps aren't typically on the hot path, but good practice for logging infrastructure.

src/runtime_src/core/common/config_reader.h
```diff
@@ -1242,6 +1242,19 @@ get_uc_log_size_per_uc_kb()
...
 
+inline std::string
+get_uc_log_dump()
+{
+  // Sink for parsed uC log output: "file" (default), "syslog", "console", or "null"
+  // "file"    -> write to timestamped per-chunk files
+  // "syslog"  -> route to OS system log (Linux syslog / Windows Event Log under AMD_XRT)
+  // "console" -> write to stderr
+  // "null"    -> discard
+  // Note: uc_log_bin_format=true always writes raw binary to 'file' regardless of this setting.
```
chvamshi-xilinx, line 1267: The config documentation is clear. One suggestion: since the PR description mentions "null or empty" discards, but the code in `dispatch_parsed_log` treats empty as "file" (for backward compatibility), the comment here should clarify that empty means "file" (not null/discard) to match the actual implementation:

```cpp
// "file" or "" (default) -> write to timestamped per-chunk files
```

This avoids confusion between the PR description and the code behavior.

src/runtime_src/core/common/message.h
```diff
@@ -38,6 +38,15 @@ XRT_CORE_COMMON_EXPORT
 void
 send(severity_level l, const char* tag, const char* msg);
 
+// Send a uC log line directly to the given sink ("syslog" or "console"),
+// bypassing runtime_log and verbosity filtering.
+// Used by buffer_dumper for uc_log_dump routing.
```
chvamshi-xilinx, line 43: The new `send_uc_log` declaration introduces a `std::string` by const-reference for the `sink` parameter. Since `send_uc_log` uses a static dispatcher (initialized once), the `sink` parameter is only meaningfully used on the first call. On subsequent calls it's passed but ignored. This is fine functionally but slightly misleading to callers.

Consider whether the sink should be configured at a higher level (e.g., during `buffer_dumper` construction) rather than passed on every call. Alternatively, add a brief doc note that the sink is latched on first call.

chvamshi-xilinx (commented):
Overall this is a well-structured PR that cleanly extends the buffer_dumper infrastructure. The refactoring into separate functions (`dump_chunk_data_binary`, `parse_log_entries`, `dispatch_parsed_log`) significantly improves readability over the original monolithic function.

**What looks good:**
- Clean separation of concerns between binary dump and parsed text dispatch
- Lazy file creation avoids unnecessary I/O when syslog/console/null is configured
- Backward compatibility is preserved (empty/default = file)
- The `send_uc_log` dedicated path avoids conflating uC log routing with the general `runtime_log` verbosity filtering
- Updating syslog ident from legacy "sdaccel" to "XRT" is a welcome modernization

**Suggestions (non-blocking):**
1. The `send_uc_log` sink parameter is latched on first call via static init — consider documenting this or restructuring so the sink isn't passed on every call
2. Minor performance: `std::string` concatenation in `syslog_dispatch::send()` could use a stack buffer to avoid heap allocations in the logging path
3. The comment in `config_reader.h` mentions "null" as discard, but empty string actually maps to "file" — align the documentation with the code
4. Consider conditionally allocating `m_file_streams` only when the sink needs files

LGTM with minor suggestions. Nice work on the refactoring iteration.

chvamshi-xilinx:
Hi @rbramand-xilinx ,
These are cursor generated comments on your PR. Please review and fix if you think anything is critical.  I was just doing some trials and used your PR for the trial :-).

src/runtime_src/core/common/message.cpp
rbramand-xilinx: make_dipatcher already uses make_unique and returns unique_ptr

src/runtime_src/core/common/buffer_dumper.cpp
rbramand-xilinx: Yeah but for released version of firmware the messages are very limited so this should not be a concerm.
For debug version performance drop is expected.

src/runtime_src/core/common/detail/linux/syslog.h
rbramand-xilinx: syslog accepts printf style formatting, so will use that to avoid creating temporary string

src/runtime_src/core/common/buffer_dumper.cpp
rbramand-xilinx: Added changes to not allocate vector if file sink is not configured

src/runtime_src/core/common/message.h
rbramand-xilinx: Added comment in code explaining same

chvamshi-xilinx (approved):

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -91,153 +91,192 @@ read_logged_count(uint8_t* chunk)
...
-      }
-
-      parsed_bytes += m_config.metadata_size;
-    }
+  // syslog or console: each line is dispatched as a discrete message so the
+  // sink can apply its own formatting and routing per entry
+  std::istringstream lines(text);
+  std::string line;
+  while (std::getline(lines, line)) {
+    if (!line.empty())
```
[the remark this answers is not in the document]
rbramand-xilinx: ignore

PR #9856, opened 2026-06-06
AIESW-35158:Dump data before updating count on overflow

svup-xilinx:
When the UC log ring buffer overflowed, the code updated the dumped counter before actually saving any data, causing the dump check to become false and all logs to be lost. The fix syncs and dumps the buffer contents before updating the counter, so overflow now saves the latest available logs instead of discarding everything.
[AIESW-35158](https://jira.xilinx.com/browse/AIESW-35158)
On overflow detection, sync the data buffer from device and call dump_chunk_data() to save the contents before updating dumped_count. This preserves the latest logs in the buffer.
Low
Tested on Windows medusa board with debug cert firmware using xrt-smi validate. Able to reproduce the issue with xrt-smi validate and verified with the fix. 
NA

### DIFF
--- a/src/runtime_src/core/common/buffer_dumper.cpp
+++ b/src/runtime_src/core/common/buffer_dumper.cpp
@@ -299,13 +299,13 @@ process_chunks_no_lock()
     size_t logged_wrap = logged_count / m_data_size;
     size_t dumped_wrap = dumped_count / m_data_size;
 
-    // Overwrite detected: dump entire buffer and catch up and resume dumping from current position.
+    // Overflow detected: reposition to dump the most recent full buffer's worth of data.
     if (logged_count > dumped_count && logged_wrap > dumped_wrap) {
-      m_config.dump_buffer.sync(XCL_BO_SYNC_BO_FROM_DEVICE, m_data_size,
-                                chunk_offset + m_config.metadata_size);
-      dump_chunk_data(i, logged_count - m_data_size, m_data_size, chunk);
-      dumped_count = logged_count;
-      continue;
+      xrt_core::message::send(xrt_core::message::severity_level::warning, "buffer_dumper",
+                              "UC log overrun on chunk " + std::to_string(i) + ": " +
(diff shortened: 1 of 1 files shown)


chvamshi-xilinx (approved):

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -299,9 +299,14 @@ process_chunks_no_lock()
     size_t logged_wrap = logged_count / m_data_size;
     size_t dumped_wrap = dumped_count / m_data_size;
 
-    // Overwrite detected; catch up and resume dumping from current position.
-    if (logged_count > dumped_count && logged_wrap > dumped_wrap)
+    // Overwrite detected: dump entire buffer and catch up and resume dumping from current position.
+    if (logged_count > dumped_count && logged_wrap > dumped_wrap) {
```
rbramand-xilinx, line 303: I feel we can make this simple :

    if (logged_count > dumped_count && logged_wrap > dumped_wrap) 
      dumped_count = logged_count - m_data_size;
 
This way below code automatically handles what needs to be synced and dumped.

Additionally what I was thinking is to log a message saying that there was overflow.

svup-xilinx: Updated the code

rbramand-xilinx: Looks good Srikanth but may be we can print number of lost  bytes also

if (logged_count > dumped_count && logged_wrap > dumped_wrap) {
      xrt_core::message::send(xrt_core::message::severity_level::warning, "**xrt_buffer_dumper**",
          "Overwrite detected in chunk " + std::to_string(i) + ": " +
          **std::to_string(logged_count - dumped_count - m_data_size) +
          " bytes lost**. Recovering from oldest surviving position.");
      dumped_count = logged_count - m_data_size;
    }

rbramand-xilinx (changes requested):

rbramand-xilinx (approved):

PR #9862, opened 2026-06-08
smi: interrupt watch mode on Ctrl+C via sigaction without SA_RESTART

NishadSaraf:
Watch mode (e.g. xrt-smi examine --firmware-log --watch) blocks in a kernel wait_event_interruptible() that returns -ERESTARTSYS on Ctrl+C. The SIGINT handler was installed with signal(), which carries SA_RESTART semantics by default on glibc, so the kernel silently auto-restarts the ioctl. Control never returns to the watch loop to check its interrupt flag, and Ctrl+C appears to hang.

Internal testing. 

Install the handler via sigaction() with sa_flags = 0 (no SA_RESTART) on POSIX so the interrupted ioctl returns EINTR and the watch loop unwinds. Windows has no SA_RESTART concept and keeps the portable signal() path.

None

Verified the change on Strix

None

### DIFF
--- a/src/runtime_src/core/tools/common/SmiWatchMode.cpp
+++ b/src/runtime_src/core/tools/common/SmiWatchMode.cpp
@@ -22,8 +22,13 @@ namespace signal_handler {
   // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables) - Required for signal handling
   std::atomic<bool> watch_interrupted{false};
 
+#ifdef _WIN32
   // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables) - Required for signal handling
   void (*old_signal_handler)(int) = nullptr;
+#else
+  // NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables) - Required for signal handling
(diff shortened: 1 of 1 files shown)


maxzhen (approved):

aktondak (approved):
Looks good. Strange we dont have a cross platform standard API behavior from C++. Maybe this should be standardized within XRT itself so utilities like xrt-smi dont keep adding platform specific code.
Will let @stsoe comment further.
Thanks for making the change.

stsoe:
> Looks good. Strange we dont have a cross platform standard API behavior from C++. Maybe this should be standardized within XRT itself so utilities like xrt-smi dont keep adding platform specific code. Will let @stsoe comment further. Thanks for making the change.

There will be no signal handling code within XRT itself.  No XRT library code should accidentally add a signal handler.  It's fine for xrt-smi of course.

stsoe (approved):
This probably should be abstracted more cleanly behind a platform independent API.  But as I mentioned in another comment, this abstraction should be at the tools layer, not in XRT library code.   I am fine with this PR if owners of xrt-smi are fine.

PR #9872, opened 2026-06-16
AIESW-13862: Fix group_id crash for scalar kernel arguments

svup-xilinx:
Calling xrt::kernel::group_id() on a scalar argument crashes
with std::out_of_range because connectivity::default_connection is only
populated for arguments with memory connections. When a scalar argument
appears after all pointer arguments, its index falls outside the vector
bounds causing the crash.
[AIESW-13862](https://jira.xilinx.com/browse/AIESW-13862)
Added a bounds check in connectivity::get_arg_memidx() to return
no_memidx when argidx falls outside default_connection.
Low
Tested on telluride board.
NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -696,6 +696,13 @@ class ip_context
     int32_t
     get_arg_memidx(size_t argidx) const
     {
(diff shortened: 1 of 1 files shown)


stsoe (changes requested):
Your PR description is very good.   Please add that as comment in the code for the function you have changed.

stsoe (approved):

chvamshi-xilinx (approved):

PR #9880, opened 2026-06-22
[XRT-SMI] miniz de-compression addition for binaries used by xrt-smi validate

aktondak:
xrt_smi_*.a VTD archives are growing larger largely because of larger binaries used for real model testing (e.g. Strix wts32_hw_sanity_strix.bin ~51 MiB). This change reduces distributed archive size by zlib-compressing selected binaries at pack time while keeping the same logical bytes after load.
This PR adds the miniz compression library which is a distributable third party library released as a pair of .h/.c. xrt-smi and xbmgmt are linked to this .c at compile time and used for decompressing the compressed binary. 
This decompression can be used for any other requirement going forward. 
Observed on Strix: full archive ~92 MiB → ~80 MiB (~14% overall); the main weights blob ~24% smaller (~13 MiB saved). Decompression is in-process in xrt-smi/xbmgmt (no gzip/unzip subprocesses), Windows and Linux friendly, without requiring system zlib as an end-user install.

Not a bug fix — this is a size / distribution enhancement. No regression from a prior PR.

Pack (VTD) — VTD/archive/build_archives.py:

Stage files into a temp tree before ar.
For selective binaries with size > 1 MiB, apply Python zlib.compress.
Unpack (XRT tools) — XBUtilities::extract_artifacts_from_archive:

New optional parameter is_compressed (default false).
When true, each listed member is passed through miniz tinfl_decompress_mem_to_heap(..., TINFL_FLAG_PARSE_ZLIB_HEADER); on failure, original bytes are kept (supports mixed compressed + uncompressed members in one list).
TestSanity.cpp passes is_compressed=true for Strix and NPU3 sanity artifact lists only.

None

Tested via compressing the weights binary for sanity test in strix and decompressing through miniz while xrt-smi run time. The test works as expected. 
The compression saves about 15MBs on the merged archive file

xdna-driver/xrt/.../third_party/miniz/README.md: upstream version (3.1.1) to link to miniz release.

### DIFF
--- a/src/runtime_src/core/tools/common/XBUtilities.cpp
+++ b/src/runtime_src/core/tools/common/XBUtilities.cpp
@@ -32,6 +32,8 @@
 #include <regex>
 #include <stdexcept>
 
+#include "tools/common/third_party/miniz/miniz.h"
+
 
 #ifdef _WIN32
 # pragma comment(lib, "Ws2_32.lib")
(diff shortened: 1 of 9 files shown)


stsoe (commented):
Is there a ticket for this?  I am not overly impressed with with saving a few 10s MB diskspace at the expense of more complicated code and adding 10k lines of code.

aktondak:
> Is there a ticket for this? I am not overly impressed with with saving a few 10s MB diskspace at the expense of more complicated code and adding 10k lines of code.

@stsoe 
There is no ticket. This is just something we pro-actively wanted to address before the size of generated archive grows any further. We can choose to delay this until something is reported on the size.

PR #9886, opened 2026-06-29
trace: make sys/sdt.h include optional

mhgerlach30:
sys/sdt.h (SystemTap USDT) is not present in all Linux build environments, notably cross-compilation toolchains and embedded distributions that do not ship systemtap-sdt-dev.  The unconditional include causes a build failure in those environments.

Use __has_include to make the inclusion conditional.  When sys/sdt.h is unavailable, define no-op stubs for the four USDT macros used by XRT (STAP_PROBEV, DTRACE_PROBE, DTRACE_PROBE1, DTRACE_PROBE2) so that the tracing call sites compile without change.

Compilation errors in environments without sys/sdt.h

Define no-op stubs for the four USDT macros used by XRT (STAP_PROBEV, DTRACE_PROBE, DTRACE_PROBE1, DTRACE_PROBE2).

Successfully compiled on Ubuntu 24.04 with and without systemtap-sdt-dev installed.

### DIFF
--- a/src/runtime_src/CMakeLists.txt
+++ b/src/runtime_src/CMakeLists.txt
@@ -12,6 +12,7 @@ endif()
 
 option(XOCL_VERBOSE "Enable xocl verbosity" OFF)
 option(XRT_VERBOSE "Enable xrt verbosity" OFF)
+option(XRT_ENABLE_USDT "Enable SystemTap USDT tracepoints" ON)
 
 if (XOCL_VERBOSE)
   add_compile_options("-DXOCL_VERBOSE")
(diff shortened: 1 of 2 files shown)


stsoe:
Couldn't this silently ignore cases where some build infrastructure failed to install systemtap-sdt-dev?  That's the point I realize, but silently stubbing out doesn't seem right to me.

mhgerlach30:
> Couldn't this silently ignore cases where some build infrastructure failed to install systemtap-sdt-dev? That's the point I realize, but silently stubbing out doesn't seem right to me.

Good point. The build will get a clear error message if systemtap-sdt-dev in not installed when XRT_ENABLE_USDT=ON.

stsoe (approved):
Thank you.  Looks good.
CI is broken currently so merge will be delayed.

stsoe (approved):

mhgerlach30:
> Thank you. Looks good. CI is broken currently so merge will be delayed.

Thank you for speedy responses.

Is https://github.com/Xilinx/aiebu/pull/314 something you can look at too?

PR #9890, opened 2026-07-01
support pl ip patching schema pl_ddr_64 

HimanshuChoudhary-Xilinx:
support pl ip patching schema "pl_ddr_64"

https://amd.atlassian.net/wiki/spaces/AIE/pages/1744115100/Weight+Prefetch+via+PL+IP+and+PLIO+Feature+Specification
XRT patches words 8+9 of wts_params with the physical weight buffer address via APPLY_OFFSET_PL at BO bind time.

### DIFF
--- a/src/runtime_src/core/common/api/elf_patcher.cpp
+++ b/src/runtime_src/core/common/api/elf_patcher.cpp
@@ -157,6 +157,17 @@ patch_ctrl57_aie4(uint32_t* bd_data_ptr, uint64_t patch)
 
 void
 symbol_patcher::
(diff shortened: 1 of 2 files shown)


stsoe (approved):
Lots of a magic numbers and accessing bd_data_ptr[] without a clue as to whether this is safe or not.   Anyways, your change builds on top of existing and LGTM in the context of existing code.   

I think all this patching code should be refactored and moved to AIEBU, but that's separate from this PR.

chvamshi-xilinx (approved):

PR #9894, opened 2026-07-03
SWSPLAT-30710 elf_patcher validate patch offset against BO size before write

stsoe:
For each patch_config built from .rela.dyn entries in an attacker-supplied AIE ELF, offset_to_patch_buffer originates from the ELF relocation r_offset field.  In patch_symbol() the expression:

  auto bd_data_ptr = reinterpret_cast<uint32_t*>(base + offset);

was computed with no check that offset + required_patch_bytes <= bo.size(). patch57() then reads/writes 9 words (36 bytes) at that address.  A crafted ELF with r_offset = bo_size - 4 writes 32 bytes past the end of the mapped instruction BO into adjacent heap memory (CWE-787, CVSS 7.1).

The same unchecked pattern existed in patch_symbol_raw() which takes a raw pointer with no accompanying size, and in the secondary sync offset computed for symbol_type::pl_ddr_64.

SWSPLAT-30710. Identified by Mythos AI security audit (MYTHOS-RyzenAI-XRT-H3). Parent finding: SWSPLAT-30707.

1. Added required_patch_bytes(symbol_type) — a pure helper that returns the minimum byte span past bd_data_ptr required by each patching scheme, derived from the highest word index each patch*() function accesses: - shim_dma_base_addr (patch57):  words [1],[2],[8] → span [0..8] = 9 words
     - pl_ddr_64 (patch_pl_ddr64):   words [8],[9]     → span [0..9] = 10 words
     - all others: 1–4 words
   Magic numbers are suppressed with NOLINT(readability-magic-numbers); each
   case has a comment stating the word indices accessed and the index span
   required.

2. patch_symbol(): capture bo.size() once, then for every config entry check offset > bo_size || patch_bytes > bo_size - offset (overflow-safe form) and throw xrt_core::error(-EINVAL) before computing bd_data_ptr.

3. The sync lambda is simplified to take no size argument: it reuses the patch_bytes value already computed for the bounds check, eliminating the per-case hardcoded byte counts that previously duplicated the information in required_patch_bytes(). pl_ddr_64 retains an explicit sync call because its sync starts 8 words into the block rather than at offset, but the bounds check (10 words) already covers that sub-range.

4. patch_symbol_raw(): added buf_size parameter (callers already know the buffer size) and apply the identical guard.  Updated the one call site in xrt_module.cpp to pass sz, which is already validated against inst->size() at that point.

5. Updated elf_patcher.h declaration to match the new signature.

Low. All changes are early-exit guards on attacker-controlled offsets. Valid ELF files with offsets that fit inside the BO are unaffected. The buf_size parameter addition is an ABI change to patch_symbol_raw(), but that function is internal (not exported in the public API).

Code review against the SWSPLAT-30710 finding and manual inspection of all patch*() word-index accesses to verify required_patch_bytes() covers each case.  AFL++/LibFuzzer harness from the ticket is recommended to confirm the OOB write is closed.

### DIFF
--- a/src/runtime_src/core/common/api/elf_patcher.cpp
+++ b/src/runtime_src/core/common/api/elf_patcher.cpp
@@ -4,12 +4,49 @@
 #define XCL_DRIVER_DLL_EXPORT  // in same dll as xrt_api
 #define XRT_CORE_COMMON_SOURCE // in same dll as core_common
 #include "elf_patcher.h"
+#include "core/common/error.h"
 
 #include <cstdlib>
 #include <cstring>
 #include <stdexcept>
 
 namespace xrt_core::elf_patcher {
+
+// Return the minimum number of bytes required past bd_data_ptr for a given symbol type.
+// Used to bounds-check the patch offset before any pointer arithmetic.
+static size_t
+required_patch_bytes(symbol_type type)
+{
+  switch (type) {
+  case symbol_type::address_64:
+    // words [0],[1]
+    return 2 * sizeof(uint32_t);   // NOLINT(readability-magic-numbers)
+  case symbol_type::scalar_32bit_kind:
+    // word [0]
+    return 1 * sizeof(uint32_t);   // NOLINT(readability-magic-numbers)
+  case symbol_type::shim_dma_base_addr_symbol_kind:
+    // words [1],[2],[8] — buffer must cover indices [0..8]
+    return 9 * sizeof(uint32_t);   // NOLINT(readability-magic-numbers)
+  case symbol_type::shim_dma_aie4_base_addr_symbol_kind:
+    // words [0],[1]
+    return 2 * sizeof(uint32_t);   // NOLINT(readability-magic-numbers)
+  case symbol_type::control_packet_57:
+  case symbol_type::control_packet_48:
+    // words [2],[3] — buffer must cover indices [0..3]
+    return 4 * sizeof(uint32_t);   // NOLINT(readability-magic-numbers)
+  case symbol_type::shim_dma_48:
+    // words [1],[2] — buffer must cover indices [0..2]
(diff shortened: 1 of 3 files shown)


larry9523 (commented):
This is a legitimate finding and the fix looks good to me.

But there are two other issues that needs a follow up fix.
1) When we try to cache/resume the patching block (max_bd_words) to deal with RMW patching for multiple set_arg in a single xrt::run object, we blindly cache/resotre max_bd_words (9 dwords and 36 bytes). For scalar_32bit_kind, the malicious ELF can set scalar type and offset to bo_size - 4. So it can pass the check introduced by this PR but we are still read/write 36 bytes which is 32 bytes out of the BO boundary.
2) For pl_ddr_64, it has 10 dwords but we only cache/restore 9 dwords. We will have problem of word[9] because we do not cache/restore it.

I suggest we can merge this PR but fix the above problems by a follow up PR.

rbramand-xilinx:
> This is a legitimate finding and the fix looks good to me.
> 
> But there are two other issues that needs a follow up fix.
> 
> 1. When we try to cache/resume the patching block (max_bd_words) to deal with RMW patching for multiple set_arg in a single xrt::run object, we blindly cache/resotre max_bd_words (9 dwords and 36 bytes). For scalar_32bit_kind, the malicious ELF can set scalar type and offset to bo_size - 4. So it can pass the check introduced by this PR but we are still read/write 36 bytes which is 32 bytes out of the BO boundary.
> 2. For pl_ddr_64, it has 10 dwords but we only cache/restore 9 dwords. We will have problem of word[9] because we do not cache/restore it.
> 
> I suggest we can merge this PR but fix the above problems by a follow up PR.

Hi @larry9523, yes your observations are right and I would send a follow up PR addressing these issues. Thanks for pointing them out.

PR #9902, opened 2026-07-06
restoring patching schema "unknown" as 10 and adding pl_ddr_64  as 11, which inline with aiebu

HimanshuChoudhary-Xilinx:
restoring patching schema "unknown" as 10 and adding pl_ddr_64  as 11, which inline with aiebu

### DIFF
--- a/src/runtime_src/core/common/api/elf_patcher.h
+++ b/src/runtime_src/core/common/api/elf_patcher.h
@@ -82,8 +82,8 @@ enum class symbol_type {
   control_packet_57 = 7,                   // patching scheme needed by firmware to patch control packet for aie2ps
   address_64 = 8,                          // patching scheme needed to patch pdi address
   control_packet_57_aie4 = 9,              // patching scheme needed by firmware to patch control packet for aie4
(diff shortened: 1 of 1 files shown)


stsoe (changes requested):
If the numbers need to align between AIEBU and this file, then it must be done symbolically through an exported enum from AIEBU.    No one will have any clue about this hidden dependency. 

Better yet, move all this patching madness inside AIEBU, where it makes more sense.

chvamshi-xilinx:
Merging this as this is important. Filed https://jira.xilinx.com/browse/AIESW-38336 to track moving patching logic to aiebu

larry9523:
> If the numbers need to align between AIEBU and this file, then it must be done symbolically through an exported enum from AIEBU. No one will have any clue about this hidden dependency.
> 
> Better yet, move all this patching madness inside AIEBU, where it makes more sense.

I think the wording here is not to align with AIEBU. The key here is to align with ELF spec where
10 is unknown 
11 is PL addr

Both AIEBU and XRT (elf parser) should be conform to ELF spec.

PR #9907, opened 2026-07-07
xrt-smi: treat unsupported telemetry ioctl as optional in the caller

Scottcjn:
Follow-up to amd/xdna-driver#1449.

That PR skipped optional telemetry in the amdxdna shim. The review there asked for the shim to faithfully return the ioctl error and let the caller decide whether the query is optional, so I am withdrawing the shim change and moving the decision here.

aie2_telemetry_info and aie2_preemption_info now treat only EOPNOTSUPP and ENOTTY as telemetry not available, which is what aie2_get_info returns from its default case on an older amdxdna driver (or if the get_aie_info op is absent). Any other error including EINVAL is still surfaced so real failures are not hidden. aie2_preemption_info uses the same rtos_telemetry query, so it gets the same handling and no longer throws on an old driver.

No shim change is needed. ReportTelemetry already reads every field through get_child with an empty fallback, so an empty or partial telemetry tree renders fine.

### DIFF
--- a/src/runtime_src/core/common/info_telemetry.cpp
+++ b/src/runtime_src/core/common/info_telemetry.cpp
@@ -2,17 +2,31 @@
 // Copyright (C) 2024-2025 Advanced Micro Devices, Inc. All rights reserved.
 #define XRT_CORE_COMMON_SOURCE
 #include "info_telemetry.h"
+#include "error.h"
 #include "query_requests.h"
 
 #include <boost/format.hpp>
+#include <cerrno>
 #include <vector>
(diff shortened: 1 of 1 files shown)


Scottcjn:
All checks except apu-package-build are passing and that one looks blocked on the bot authorization step rather than the change itself. Could an admin add the build label when you get a chance? Happy to fix anything the full pipeline turns up.

stsoe (approved):

AShivangi (approved):

PR #9927, opened 2026-07-17
[XRT-SMI] Update TCT tests to follow different logic from strix and report throughput

aktondak:
The TCT validate tests did not handle MDS (NPU3) correctly: they assumed fixed Strix token counts when computing latency and throughput. This change adds an MDS path that uses runner elapsed time and iteration count, keeps Strix behavior unchanged, and shares a common workload runner between both platforms. Separately, xrt-smi watch mode now refreshes the terminal on each step for snapshot-style reports (examine, context-health), while firmware-log and event-trace streaming watch modes are unchanged. The shim also maps firmware log configure failures to a clear “Invalid firmware log level” message.

None

TCT tests use one run_tct_workload() path and branch only in calculate_tct_metrics() (Strix uses fixed samples × iterations; MDS uses iterations only). We rejected duplicating full Strix/MDS run functions and using the runner’s precomputed latency/throughput fields, which do not match the MDS workload model. Watch mode defaults refresh_terminal to true, with firmware-log and event-trace explicitly opting out.

The reported TCT numbers would now change

Tested on npu3 platform. This TCT throughput seems close to no-op throughput so numbers make more sense now : 
```
C:\Users\aktondak>xrt-smi validate --advanced -r tct-all-col
-------------------------------------------------------------------------
                    DISCLAIMER  (xrt-smi --advanced)
You are running a developer command that may change system configuration.
                Continue only if you understand the risks.
-------------------------------------------------------------------------
    Platform              : NPU Medusa
    Power Mode            : performance
Test 1 [00c5:00:01.1]     : tct-all-col
    Details               : Average TCT throughput (all columns): 10273.5 TCT/s
```

None

### DIFF
--- a/src/runtime_src/core/tools/common/SmiWatchMode.h
+++ b/src/runtime_src/core/tools/common/SmiWatchMode.h
@@ -70,7 +70,7 @@ public:
  * by any XRT-SMI report. It handles:
  * - Element filter parsing for watch mode options
  * - Signal handling (Ctrl+C interruption) with graceful cleanup
- * - Optional terminal clear between updates (examine --watch only)
+ * - Terminal clear between updates (disabled for streaming firmware-log/event-trace watch)
  * - Optional refresh interval (seconds; 0 = no delay between updates)
  * - Cross-platform compatibility (Windows/POSIX)
  * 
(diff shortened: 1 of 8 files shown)


AShivangi (approved):
Thanks for fixing this!
Isn't this number too low? on strix it is about 800 TCT/s

rchane (approved):

PR #9929, opened 2026-07-20
Add parameter to fix checkout

amd-gevadlam:
The actions/checkout@v3 tag was updated upstream to include a new security guardrail that blocks checking out fork PR code in pull_request_target workflows by default. This is to prevent "pwn request" vulnerabilities where untrusted fork code runs with access to the base repo's secrets and GITHUB_TOKEN.

PRs before July 18 (e.g. #9926) were unaffected because they ran against the older checkout SHA. PRs after (e.g. #9927) all fail at checkout before the build even starts.

**Fix**
Add allow-unsafe-pr-checkout: true to the three "Checkout PR" steps in xrt_ci.yml (in the build, windows-build, and apu-package-build jobs).

This is safe in our workflow because the authorize job already gates all builds behind an allowlist check — untrusted fork authors cannot trigger builds without prior approval.

### DIFF
--- a/.github/workflows/xrt_ci.yml
+++ b/.github/workflows/xrt_ci.yml
@@ -67,7 +67,8 @@ jobs:
           ref: "${{ github.event.pull_request.head.ref }}"
           fetch-depth: 0
           path: ${{ github.workspace }}/${{ github.run_number }}  
-          submodules: recursive   
+          submodules: recursive
+          allow-unsafe-pr-checkout: true
 
       - name: Checkout private repository      
         uses: actions/checkout@v3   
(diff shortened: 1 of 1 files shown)


stsoe (changes requested):
For history sake, it would be helpful to explain the change in the PR description.

amd-gevadlam:
> For history sake, it would be helpful to explain the change in the PR description.

Added, Could you please review? We neeed to get this merged in order to reflect in the CI builds. The PR build would still fail because it takes only the master's CI code.

stsoe (approved):
LGTM.  Thanks for adding the description.

PR #9930, opened 2026-07-20
AIESW-3443 Initial ABI support for xrt-smi

rchane:
https://jira.xilinx.com/browse/AIESW-3443
Internal testing concerns over --format options, discussions on how to protect backward compatibility for JSON schemas
This PR adds JSON ABI versioning for xrt-smi examine and validate, so report output can evolve without breaking existing parsers. A central schema registry and ReportSchemaProjector project each report’s canonical property tree onto a selected ABI before both CLI rendering and JSON file output. 
Alveo and AIE2 platforms continue to use --format (JSON, JSON-2020.2); beyond uses --json (default). xrt-smi will resolve the active ABI from explicit flags or from device type. Today’s ABIs are identity projections; numbered versions are reserved for parser-breaking property-tree changes.
Will break --format for non Alveo/AIE2.
Tested on STX and npu3. Testcases edit PR is merged.
Need to document in user guide and announce to stakeholders.


src/runtime_src/core/tools/common/ReportSchemaProjector.cpp
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2026 Advanced Micro Devices, Inc. All rights reserved.
+
+#include "ReportSchemaProjector.h"
+
+#include "core/common/time.h"
+
+bool
+ReportSchemaProjector::emitsJsonDocument(Report::SchemaVersion schemaVersion)
```
aktondak: Can we rename this to a better name. This API doesn't emit any document. 
Also, lets use snake case for all new code added.

src/runtime_src/core/tools/common/ReportSchemaProjector.cpp
```diff
@@ -0,0 +1,55 @@
...
+                                             bool useJsonVersionNaming)
+{
+  boost::property_tree::ptree node;
+  node.put("schema", Report::getSchemaOutputLabel(schemaVersion, useJsonVersionNaming));
+  node.put("creation_date", xrt_core::timestamp());
+  return node;
+}
+
+boost::property_tree::ptree
+ReportSchemaProjector::projectLatest(const boost::property_tree::ptree& canonicalSuperset)
```
aktondak: Why is this API returning its own argument ?

rchane: The "project" functions are meant to edit existing ptrees. If there are edits, they can be added, but for now, there is nothing to edit.

src/runtime_src/core/tools/common/ReportSchemaProjector.cpp
rchane: Sure, let me update that.

src/runtime_src/core/common/smi/smi_alveo.cpp
```diff
@@ -41,7 +41,7 @@ create_validate_subcommand()
   validate_suboptions.emplace("device", std::make_shared<xrt_core::smi::option>("device", "d", "The Bus:Device.Function (e.g., 0000:d8:00.0) device of interest", "common", "", "string"));
   validate_suboptions.emplace("format", std::make_shared<xrt_core::smi::option>("format", "f", "Report output format. Valid values are:\n"
                                 "\tJSON        - Latest JSON schema\n"
-                                "\tJSON-2020.2 - JSON 2020.2 schema", "common", "JSON", "string"));
+                                "\tJSON-2020.2 - JSON 2020.2 schema (legacy)", "common", "JSON", "string"));
```
AShivangi, line 44: do we need to add legacy for alveo as well?

src/runtime_src/core/tools/common/Report.h
```diff
@@ -16,20 +16,21 @@
...
   };
  
-  // Helper mapping between string and enum
   struct SchemaDescription {
     SchemaVersion schemaVersion;
-    bool isVisable;
-    std::string optionName;
+    bool isVisable;              // Listed in --format help when optionName is set
+    std::string optionName;      // --format value (e.g. JSON, JSON-2020.2)
+    std::string jsonVersionName; // --json value (e.g. default)
```
aktondak: Do we need this jsonVersionName new variable addition. I think we can reuse the optionName itself for --json option as well. This change is causing a lot of downstream edits.

src/runtime_src/core/tools/common/Report.cpp
```diff
@@ -17,46 +19,98 @@
...
+    if (!entry.jsonVersionName.empty()
+        && boost::iequals(entry.jsonVersionName, jsonVersion))
+      return entry.schemaVersion;
+  }
+
+  return SchemaVersion::unknown;
+}
+
+Report::SchemaVersion
+Report::resolveSchemaFromFormatName(const std::string& formatName)
```
AShivangi: does this needs to be exposed? can it be an internal static function or maybe just a lambda func?

src/runtime_src/core/tools/common/ReportSchemaProjector.h
```diff
@@ -0,0 +1,27 @@
...
+// Projects the canonical xrt-smi property-tree superset onto a frozen JSON ABI.
+class ReportSchemaProjector {
+ public:
+  static bool emitsJsonDocument(Report::SchemaVersion schemaVersion);
+
+  static boost::property_tree::ptree makeSchemaVersionNode(Report::SchemaVersion schemaVersion,
+                                                           bool useJsonVersionNaming);
+
+  static boost::property_tree::ptree project(Report::SchemaVersion target,
+                                             const boost::property_tree::ptree& canonicalSuperset);
```
AShivangi: Please add documentation for all the exposed APIs

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -417,10 +426,13 @@ SubCmdValidate::handle_errors_and_validate_tests(const boost::program_options::v
...
     throw xrt_core::error("Power mode not specified");
 
-  // When json is specified, make sure an accompanying output file is also specified
+  // When json output is requested, make sure an accompanying output file is also specified
    if (vm.count("format") && options.m_output.empty())
     throw xrt_core::error("Please specify an output file to redirect the json to");
 
+   if (vm.count("json") && options.m_output.empty())
+    throw xrt_core::error("Please specify an output file to redirect the json to");
+
```
AShivangi:
```suggestion
// Either JSON output selector requires an accompanying output file
if ((vm.count("format") || vm.count("json")) && options.m_output.empty())
  throw xrt_core::error("Please specify an output file to redirect the json to");

```

src/runtime_src/core/tools/xbutil2/SubCmdExamine.cpp
```diff
@@ -226,20 +230,24 @@ SubCmdExamine::execute(const SubCmdOptions& _options) const
...
+                             % (usesJsonAbiOption ? options.m_json : options.m_format)).str());
 
-    // DRC check
-    // When json is specified, make sure an accompanying output file is also specified
     if (vm.count("format") && options.m_output.empty())
       throw xrt_core::error("Please specify an output file to redirect the json to");
 
+    if (vm.count("json") && options.m_output.empty())
+      throw xrt_core::error("Please specify an output file to redirect the json to");
+
```
AShivangi:
```suggestion
// Either JSON output selector requires an accompanying output file
if ((vm.count("format") || vm.count("json")) && options.m_output.empty())
  throw xrt_core::error("Please specify an output file to redirect the json to");

```

src/runtime_src/core/tools/common/Report.cpp
```diff
@@ -17,46 +19,98 @@
...
   for (const auto & entry : m_schemaVersionVector) {
-    if (boost::iequals(entry.optionName, _schemaVersionName)) 
+    if (!entry.optionName.empty() && boost::iequals(entry.optionName, _schemaVersionName))
+      return entry;
+    if (!entry.jsonVersionName.empty() && boost::iequals(entry.jsonVersionName, _schemaVersionName))
       return entry;
   }
 
-  // Return back the unknown entry
   return getSchemaDescription("");
```
AShivangi:
```suggestion
  return getSchemaDescription(SchemaVersion::unknown);
```

The fallback here recurses infinitely. The `unknown` entry's `optionName` and `jsonVersionName` are now empty strings, and the loop above guards every comparison with `!entry.optionName.empty() / !entry.jsonVersionName.empty()`. So when no name matches, `getSchemaDescription("")` also matches nothing — it just calls `getSchemaDescription("")` again, recursing until the stack overflows.

AShivangi (changes requested):

src/runtime_src/core/tools/common/Report.h
rchane: Removed jsonVersionName.

src/runtime_src/core/common/smi/smi_alveo.cpp
rchane: Yes, we're marking the entire option as legacy

src/runtime_src/core/tools/common/Report.cpp
rchane: Good catch; thanks! Fixed.

src/runtime_src/core/tools/common/Report.cpp
```diff
@@ -17,46 +19,124 @@
...
+  { SchemaVersion::json_20202,    true,  "JSON-2020.2", "",        "JSON 2020.2 schema (legacy)"},
 };
 
+namespace {
+
+// Maps a CLI version string to SchemaVersion.
+// for_json=true matches json_version_name (--json); false matches optionName (--format).
+// Each path rejects names that belong to the other flag.
+Report::SchemaVersion
+lookup_abi(const std::string& name, bool for_json)
```
aktondak: This lookup is getting too complicated because we're dealing with --json as something different than --format. 
I think the optionName should repurposed for --json too and that should simplify a lot of extra lookups we're doing for --json.
Earlier the permissible names were JSON, JSON-2020.2, JSON-internal
Lets just augment this to also include "default", "2.21" etc. whatever we want to add in future. The "for_json" flag keeping is causing complication which can be avoided.

src/runtime_src/core/tools/common/Report.cpp
```diff
@@ -17,46 +19,124 @@
...
 const Report::SchemaDescription &
-Report::getSchemaDescription(const std::string & _schemaVersionName) 
+Report::getSchemaDescription(const std::string & name)
 {
-  // Look for a match
   for (const auto & entry : m_schemaVersionVector) {
-    if (boost::iequals(entry.optionName, _schemaVersionName)) 
+    if (!entry.optionName.empty() && boost::iequals(entry.optionName, name))
+      return entry;
+    if (!entry.json_version_name.empty() && boost::iequals(entry.json_version_name, name))
```
aktondak: We're moving away from boost. strings can directly be compared with == operator

src/runtime_src/core/tools/common/Report.h
```diff
@@ -40,6 +39,31 @@ class Report : public JSONConfigurable {
   static const Report::SchemaDescription & getSchemaDescription(SchemaVersion _schemaVersion);
   static const SchemaDescriptionVector & getSchemaDescriptionVector() { return m_schemaVersionVector; };
 
+  /** Resolved ABI and if the --json registry name is used. */
+  struct JsonAbiChoice {
```
aktondak: I dont think we need this struct. Written json file doesn't need to know if the command came from --format or --json. The SchemaVersion already correctly adds json_latest to the list and that's all that should be required

aktondak (approved):
This is much cleaner now. Thanks for making this change :)

AShivangi (approved):

PR #9934, opened 2026-07-22
dynamic tracing compiler coalesce result buffer support

AdvaitNaik:

Pull in 
https://github.com/Xilinx/aiebu/pull/295
- add/change API, and provide a buffer to save results of multiple runs (https://jira.xilinx.com/browse/AIESW-30744)

https://github.com/Xilinx/aiebu/pull/305; https://github.com/Xilinx/aiebu/pull/325
- avoid creating empty dtrace results when probe don't fire (https://jira.xilinx.com/browse/AIESW-37535)
- duplicate filename in dtrace parser, performance optimization and bug fixes (https://jira.xilinx.com/browse/AIESW-36216)

https://github.com/Xilinx/aiebu/pull/321
- action mask poll register (https://jira.xilinx.com/browse/AIESW-39037)

This is a performance enhancement with dtrace JSON output support; eliminating per-run file I/O when opted in.
Every XRT `run.dump_dtrace_buffer()` performs file I/O to dump dtrace results — once per iteration. In multi-run scenarios (runlists or latency benchmarks with 1000+ iterations on the same run), this creates N files with N file system round-trips.
```
Mode                 | Latency                   | [Debug]
                     | run.dump_dtrace_buffer()  | dtrace_control_file_path=test.ct
---------------------|---------------------------|------------------------------------------
Python file I/O      | 3,048-3,389 us            | 
JSON file I/O        | 3,045-3,389 us            | dtrace_output_json_format=true
JSON buffered        | 607-718 us                | dtrace_output_json_format=true
                     |                           | dtrace_coalesce_result=true
```

```
End-to-end Performance Numbers
Mode                 | Latency         | Throughput
---------------------|-----------------|----------------------
Python file I/O      | 3,790 us        | 263 op/s
JSON file I/O        | 3,421 us        | 292 op/s
JSON buffered        | 1,992 us        | 501 op/s
```

Accumulate dtrace JSON results in an in-memory nlohmann::ordered_json buffer. Each run.dump_dtrace_buffer() appends to this buffer. A single file is dumped when the hw context is destroyed.
Coalesce buffer remains opt-in via Debug.dtrace_coalesce_result=true and Debug.dtrace_output_json_format=true. Per-run file I/O (Python/JSON direct dump) is unchanged when coalesce is off.
Coalesce buffer holds up to `Debug.dtrace_coalesce_result_memory_mb` of JSON per hw context in host memory until flush or spill.

Low, this applies only when dynamic tracing is enabled

Tested by dumping tracing json result on windows medusa board and the feature works as expected

NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -212,6 +212,33 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
     }
   };  // struct uc_log_buffer
 
+  // struct dtrace_result_buffer - coalesce per-run dtrace JSON result for this hwctx
+  struct dtrace_result_buffer
+  {
+    std::unique_ptr<xrt_core::dtrace_buffer_dumper> m_dtrace_dumper;
+
+    static std::unique_ptr<xrt_core::dtrace_buffer_dumper>
+    init_dtrace_buffer_dumper(xrt_core::hwctx_handle* ctx_hdl)
+    {
+      // Configure coalesce buffer size for this hw context slot
+      xrt_core::dtrace_buffer_dumper::config config {};
+      config.slot_idx = ctx_hdl->get_slotidx();
+      config.max_bytes = static_cast<size_t>(xrt_core::config::get_dtrace_coalesce_result_memory_mb()) * 1024 * 1024;
+      return std::make_unique<xrt_core::dtrace_buffer_dumper>(std::move(config));
+    }
+
+    explicit
+    dtrace_result_buffer(xrt_core::hwctx_handle* ctx_hdl)
+      : m_dtrace_dumper(init_dtrace_buffer_dumper(ctx_hdl))
+    {}
+
+    void
+    append(const std::string& key, const std::string& result_json)
+    {
+      m_dtrace_dumper->append(key, result_json);
+    }
+  };  // struct dtrace_result_buffer
+
   std::shared_ptr<xrt_core::device> m_core_device;
   xrt::xclbin m_xclbin;
 
@@ -235,6 +262,7 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
   access_mode m_mode;
   std::unique_ptr<xrt_core::hwctx_handle> m_hdl;
   std::unique_ptr<uc_log_buffer> m_uc_log_buf;
+  std::unique_ptr<dtrace_result_buffer> m_dtrace_result_buf;
 
   // During preemption, when a hardware context is interrupted
   // L2 memory contents are saved to a scratchpad memory allocated
@@ -406,10 +434,26 @@ public:
     return shared_from_this();
   }
 
+  void
+  append_dtrace_result(const std::string& key, const std::string& result_json)
+  {
+    // Lazily create the coalesce buffer on first append for this hw context
+    if (!m_hdl)
+      return;
+
+    if (!m_dtrace_result_buf)
+      m_dtrace_result_buf = std::make_unique<dtrace_result_buffer>(m_hdl.get());
+
+    m_dtrace_result_buf->append(key, result_json);
+  }
+
   ~hw_context_impl()
   {
     // This trace point measures the time to tear down a hw context on the device
     XRT_TRACE_POINT_SCOPE(xrt_hw_context_dtor);
+
+    // dump dtrace buffered results before shim hwctx handle is destroyed
+    m_dtrace_result_buf.reset();
 
     // dump uC log buffer before shim hwctx handle is destroyed
     m_uc_log_buf.reset();
@@ -801,6 +845,14 @@ xrt::hw_context::cfg_type
 get_cfg_map(const xrt::hw_context& hwctx)
 {
   return hwctx.get_handle()->get_cfg_map();
+}
+
+void
+append_dtrace_result(const xrt::hw_context& hwctx,
+                     const std::string& key, const std::string& result_json)
+{
+  // Append a per-run dtrace JSON result to the hw context coalesce buffer
+  hwctx.get_handle()->append_dtrace_result(key, result_json);
 }
 
 } // xrt_core::hw_context_int
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -41,13 +41,16 @@
 
 namespace {
 
-// NOLINTBEGIN(cppcoreguidelines-avoid-magic-numbers)
+// NOLINTBEGIN(cppcoreguidelines-avoid-magic-numbers, bugprone-implicit-widening-of-multiplication-result)
 constexpr std::size_t
 operator""_kb(unsigned long long v) { return 1024U * v; }
 
+constexpr std::size_t
+operator""_mb(unsigned long long v) { return 1024U * 1024U * v; }
+
 constexpr double
 operator""_mhz(long double v) { return static_cast<double>(v) * 1'000'000.0; }
-// NOLINTEND(cppcoreguidelines-avoid-magic-numbers)
+// NOLINTEND(cppcoreguidelines-avoid-magic-numbers, bugprone-implicit-widening-of-multiplication-result)
 
 template <typename T>
 using span = xrt::detail::span<T>;
@@ -223,7 +226,7 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
       // Configure coalesce buffer size for this hw context slot
       xrt_core::dtrace_buffer_dumper::config config {};
       config.slot_idx = ctx_hdl->get_slotidx();
-      config.max_bytes = static_cast<size_t>(xrt_core::config::get_dtrace_coalesce_result_memory_mb()) * 1024 * 1024;
+      config.max_bytes = static_cast<size_t>(xrt_core::config::get_dtrace_coalesce_result_memory_mb()) * 1_mb;
       return std::make_unique<xrt_core::dtrace_buffer_dumper>(std::move(config));
     }
 
@@ -437,6 +440,10 @@ public:
   void
   append_dtrace_result(const std::string& key, const std::string& result_json)
   {
+    // Concurrent run clones can dump dtrace on the same hw context;
+    // protect lazy init and append with m_mutex
+    std::lock_guard lk(m_mutex);
+
     // Lazily create the coalesce buffer on first append for this hw context
     if (!m_hdl)
       return;
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -215,33 +215,6 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
     }
   };  // struct uc_log_buffer
 
-  // struct dtrace_result_buffer - coalesce per-run dtrace JSON result for this hwctx
-  struct dtrace_result_buffer
-  {
-    std::unique_ptr<xrt_core::dtrace_buffer_dumper> m_dtrace_dumper;
-
-    static std::unique_ptr<xrt_core::dtrace_buffer_dumper>
-    init_dtrace_buffer_dumper(xrt_core::hwctx_handle* ctx_hdl)
-    {
-      // Configure coalesce buffer size for this hw context slot
-      xrt_core::dtrace_buffer_dumper::config config {};
-      config.slot_idx = ctx_hdl->get_slotidx();
-      config.max_bytes = static_cast<size_t>(xrt_core::config::get_dtrace_coalesce_result_memory_mb()) * 1_mb;
-      return std::make_unique<xrt_core::dtrace_buffer_dumper>(std::move(config));
-    }
-
-    explicit
-    dtrace_result_buffer(xrt_core::hwctx_handle* ctx_hdl)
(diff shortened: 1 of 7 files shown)


src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -369,4 +371,102 @@ flush()
...
+
+      m_spilled = true;
+      const auto max_mb = m_config.max_bytes / (1024 * 1024);
+      xrt_core::message::send(xrt_core::message::severity_level::warning, "dtrace_buffer_dumper",
+                              std::string{"[dtrace] : coalesce buffer limit ("} + std::to_string(max_mb) + " MB) reached. "
+                              + "Further dtrace results will not be buffered for hardware context "
+                              + std::to_string(m_config.slot_idx));
+
+      return;
+    }
```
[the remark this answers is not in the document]
AdvaitNaik: Intentional: 
On cap overflow we spill only previously buffered runs, discard the run that would exceed the limit, then disable further coalescing for this hw context.

We use dump(4) for the cap check because the flushed output is also written with dump(4) in write_to_file(). The limit is meant to bound the serialized coalesce payload we eventually write, not the compact JSON byte count. Measuring with dump() would under-estimate size vs what is actually written and could allow the spill/flush to exceed max_bytes.

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -406,11 +434,27 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
+  append_dtrace_result(const std::string& key, const std::string& result_json)
+  {
+    // Lazily create the coalesce buffer on first append for this hw context
+    if (!m_hdl)
+      return;
+
+    if (!m_dtrace_result_buf)
+      m_dtrace_result_buf = std::make_unique<dtrace_result_buffer>(m_hdl.get());
+
+    m_dtrace_result_buf->append(key, result_json);
```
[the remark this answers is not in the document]
AdvaitNaik: fixed

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -212,6 +212,33 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
+  {
+    std::unique_ptr<xrt_core::dtrace_buffer_dumper> m_dtrace_dumper;
+
+    static std::unique_ptr<xrt_core::dtrace_buffer_dumper>
+    init_dtrace_buffer_dumper(xrt_core::hwctx_handle* ctx_hdl)
+    {
+      // Configure coalesce buffer size for this hw context slot
+      xrt_core::dtrace_buffer_dumper::config config {};
+      config.slot_idx = ctx_hdl->get_slotidx();
+      config.max_bytes = static_cast<size_t>(xrt_core::config::get_dtrace_coalesce_result_memory_mb()) * 1024 * 1024;
```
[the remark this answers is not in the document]
AdvaitNaik: fixed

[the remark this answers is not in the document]
AdvaitNaik: fixed

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -369,4 +371,102 @@ flush()
...
+
+    if (m_results.dump(4).size() > m_config.max_bytes) {
+      m_results.erase(key);
+
+      // Flush buffered runs, warn that the cap was reached, and disable further appends
+      if (!m_results.empty())
+        write_to_file();
+
+      m_spilled = true;
+      const auto max_mb = m_config.max_bytes / (1024 * 1024);
```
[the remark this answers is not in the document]
AdvaitNaik: fixed

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -212,6 +215,33 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
     }
   };  // struct uc_log_buffer
 
+  // struct dtrace_result_buffer - coalesce per-run dtrace JSON result for this hwctx
+  struct dtrace_result_buffer
```
stsoe: Why this class?  There is no state or invariants. 
Can't m_dtrace_result_buf on line 268 simply be `std::unique_ptr<xrt_core::dtrace_buffer_dumper>`  with `init_dtrace_buffer_dumper ()` being scoped with `hw_context_impl`?  Maybe you are trying to mimic what `struct uc_logic_buffer` does, but at least it has state, albeit completely useless state that is not used for anything but initialization, so that class also serves no real purpose.

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -235,6 +265,7 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
   access_mode m_mode;
   std::unique_ptr<xrt_core::hwctx_handle> m_hdl;
   std::unique_ptr<uc_log_buffer> m_uc_log_buf;
+  std::unique_ptr<dtrace_result_buffer> m_dtrace_result_buf;
```
stsoe: Maybe just `std::unique_ptr<xrt_core::dtrace_buffer_dumper>` ?

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -369,4 +371,102 @@ flush()
...
+{
+  // No-op once the coalesce buffer has spilled or been disabled
+  if (m_spilled)
+    return;
+
+  try {
+    auto entry = nlohmann::ordered_json::parse(result_json);
+    m_results[key] = std::move(entry);
+
+    if (m_results.dump(4).size() > m_config.max_bytes) {
```
stsoe: Very expensive dump at every append.  Track accumulated bytes incremented by result_json.size(), decrement on erase. 

Also, cap semantics is not clear.   What is the purpose of this if-block?

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1142,19 +1151,30 @@ class module_run_aie_gen2_plus : public module_run
...
-                                   + "/dtrace_dump"
-                                   + postfix
-                                   + (xrt_core::config::get_dtrace_output_json_format() ? ".json" : ".py");
+      if (is_dtrace_buffer_result()) {
+        // dtrace output is serialized and buffered into result buffer
+        std::string result_key = "dtrace_dump" + postfix;
 
-      get_dtrace_result_file(m_dtrace.dtrace_handle.get(), result_file_path.c_str());
+        // Get result as JSON string from aiebu
+        std::string result_json = get_dtrace_result_buffer(m_dtrace.dtrace_handle.get());
```
stsoe, line 1159: Per `get_dtrace_result_buffer()`, the returned value can be "null" when error is encountered,  this code is not handling the error.

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -369,4 +371,102 @@ flush()
...
+                                     + "/dtrace_dump"
+                                     + "_ctx_" + std::to_string(m_config.slot_idx)
+                                     + "_" + xrt_core::get_timestamp_for_filename()
+                                     + ".json";
+
+  std::ofstream json_file(result_file_path);
+  if (!json_file)
+    throw std::runtime_error("[dtrace] : failed to open file for dumping dtrace buffer result");
+
+  json_file << m_results.dump(4) << "\n";
```
stsoe, line 408: Should there be a check for the write failing?  You could annotate the stream with exceptions and have it automatically throw on error (`json_file.exceptions(std::ios::failbit | std::ios::badbit)`)   but resetting m_results if the results were not written, is probably not the right logic?

stsoe (changes requested):

src/runtime_src/core/common/api/xrt_hw_context.cpp
AdvaitNaik: Agreed - the wrapper was mirroring `uc_log_buffer` without adding value. Removed `dtrace_result_buffer`; `m_dtrace_result_buf` is now `std::unique_ptr<xrt_core::dtrace_buffer_dumper>`, with `init_dtrace_buffer_dumper()` scoped on `hw_context_impl`.

src/runtime_src/core/common/buffer_dumper.cpp
AdvaitNaik: Thank you for catching on the cost - calling dump(4) on every append was unnecessary. Updated to track each append with `entry.dump(4).size()` to match the pretty-printed format written by `write_to_file()`

The overflow block is used when the cap is exceeded, we discard the run that triggered it, spill any previously buffered runs to disk, log a warning, and set m_spilled so no further appends are accepted for that hw context. Added comments to make that explicit.

src/runtime_src/core/common/api/xrt_module.cpp
AdvaitNaik: Added a check for the "null" return

src/runtime_src/core/common/buffer_dumper.cpp
AdvaitNaik: Added post-write/flush checks and only clear m_results after both succeed

stsoe (approved):
LGTM.  Thanks!

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -369,4 +371,115 @@ flush()
...
+append(const std::string& key, const std::string& result_json)
+{
+  // No-op once the coalesce buffer has spilled or been disabled
+  if (m_spilled)
+    return;
+
+  try {
+    auto entry = nlohmann::ordered_json::parse(result_json);
+    const auto entry_size = entry.dump(4).size();
+    m_results[key] = std::move(entry);
```
xuhz: why we blindly append then check and delete the appended  if spill, instead of check before append?

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -369,4 +371,115 @@ flush()
...
+  json_file.flush();
+  if (!json_file)
+    throw std::runtime_error("[dtrace] : failed to flush dtrace buffer result file");
+
+  xrt_core::message::send(xrt_core::message::severity_level::debug, "dtrace_buffer_dumper",
+                          std::string{"[dtrace] : dtrace buffer dumped successfully to - "}
+                          + result_file_path);
+
+  // Clear after the write completes without error
+  m_results = nlohmann::ordered_json::object();
```
xuhz, line 421: if once spilled, we discard all remaining, do we still need to clear after write completes?
or alternatively, once spill, we save to file and clear also the spill flag, so we can create multiple files without discard?
probably the latter is better, we don't want to increase the memory usage, but increase disk usage should be ok

src/runtime_src/core/common/api/xrt_module.cpp
```diff
@@ -1142,19 +1151,35 @@ class module_run_aie_gen2_plus : public module_run
...
-                                   + "/dtrace_dump"
-                                   + postfix
-                                   + (xrt_core::config::get_dtrace_output_json_format() ? ".json" : ".py");
+      if (is_dtrace_buffer_result()) {
+        // dtrace output is serialized and buffered into result buffer
+        std::string result_key = "dtrace_dump" + postfix;
+
+        // Get result as JSON string from aiebu
+        std::string result_json = get_dtrace_result_buffer(m_dtrace.dtrace_handle.get());
+        if (result_json == "null") {
```
xuhz, line 1160: for a run without probe firing, e.g we only have tracepoint for hang or npi interrupt, is this result_json null?
if so, it is a pretty common case, we don't want a warning msg here, otherwise, it will be flood of this msg,
if not, we need to make sure we don't append anything

src/runtime_src/core/common/buffer_dumper.cpp
AdvaitNaik: fixed

src/runtime_src/core/common/api/xrt_module.cpp
AdvaitNaik: added check for empty json buffer when no probe fire

src/runtime_src/core/common/buffer_dumper.cpp
AdvaitNaik: adopted the latter approach for overflow, spill and save to file, and create multiple files without discard

xuhz (approved):

src/runtime_src/core/common/buffer_dumper.cpp
```diff
@@ -369,4 +371,112 @@ flush()
...
+{
+  try {
+    auto entry = nlohmann::ordered_json::parse(result_json);
+    // No probes fired
+    if (entry.empty())
+      return;
+
+    // Spill to disk when the next append would exceed the in-memory cap, then
+    // continue buffering into a fresh in-memory batch.
+    const auto entry_size = entry.dump(4).size();
```
stsoe: I really don't understand why the size limit has anything to do with how the json is formatted (`dump(_n_)`)?

stsoe (commented):
One comment / question?

AdvaitNaik:
> One comment / question?

Updated - the in-memory cap should not depend on pretty-print formatting. Updated to track `result_json.size()` per append; `dump(n)` is only used when writing the spill file to disk

stsoe (approved):
LGTM

rbramand-xilinx (approved):
Looks good

AdvaitNaik:
> LGTM

@stsoe can you please help merge

PR #9935, opened 2026-07-23
AIESW-39741 xrt-smi validate pmode warning not showing up

AShivangi:
[AIESW-39741](https://jira.xilinx.com/browse/AIESW-39741) xrt-smi validate pmode warning not showing up

Driver behavior changed a couple weeks back, that's when it was discovered.

remove searching for "permission denied" string as it is not platform agnostic and is unreliable

N/A. 

Tested on windows

```
>xrt-smi validate -r latency
WARNING: User doesn't have admin permissions to set performance mode. Running validate in default mode
    Platform              : NPU Medusa B0
    Power Mode            : default
Test 1 [00c5:00:01.1]     : latency
    Details               : Average latency: 73.0 us
Validation completed

>echo %errorlevel%
0

```

N/A

### DIFF
--- a/src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
+++ b/src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
@@ -590,6 +590,8 @@ SubCmdValidate::execute(const SubCmdOptions& _options) const
   //get current performance mode
   const auto og_pmode = xrt_core::device_query_default<xq::performance_mode>(device, 0);
   const auto parsed_og_pmode = xq::performance_mode::parse_status(og_pmode);
+  // Tracks whether we successfully changed the pmode; only then should we reset it.
+  bool pmode_changed = false;
   //--pmode
   try {
     if (!options.m_pmode.empty()) {
@@ -613,10 +615,12 @@ SubCmdValidate::execute(const SubCmdOptions& _options) const
       else {
         throw xrt_core::error(boost::str(boost::format("Invalid pmode value: '%s'\n") % options.m_pmode));
       }
+      pmode_changed = true;
       XBU::verbose(boost::str(boost::format("Setting power mode to `%s` \n") % options.m_pmode));
     }
     else if(!boost::iequals(parsed_og_pmode, "PERFORMANCE")) {
       xrt_core::device_update<xq::performance_mode>(device.get(), xq::performance_mode::power_type::performance);
+      pmode_changed = true;
       XBU::verbose("Setting power mode to `performance`\n");
     } 
   } catch (const xq::no_such_key&) {
@@ -625,9 +629,8 @@ SubCmdValidate::execute(const SubCmdOptions& _options) const
     std::cerr << boost::format("\nERROR: %s\n") % e.what();
     printHelp();
     throw xrt_core::error(std::errc::operation_canceled);
-  } catch (const std::exception & ex) { //check if permission was denied, i.e., no sudo access
-    if(boost::icontains(ex.what(), "Permission denied"))
-      std::cout << boost::format("WARNING: User doesn't have admin permissions to set performance mode. Running validate in %s mode") 
+  } catch (const std::exception&) {
+    std::cout << boost::format("WARNING: User doesn't have admin permissions to set performance mode. Running validate in %s mode") 
                                     % parsed_og_pmode << std::endl;
   }
   // -- Run the tests --------------------------------------------------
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -625,25 +629,23 @@ SubCmdValidate::execute(const SubCmdOptions& _options) const
     std::cerr << boost::format("\nERROR: %s\n") % e.what();
     printHelp();
     throw xrt_core::error(std::errc::operation_canceled);
-  } catch (const std::exception & ex) { //check if permission was denied, i.e., no sudo access
-    if(boost::icontains(ex.what(), "Permission denied"))
-      std::cout << boost::format("WARNING: User doesn't have admin permissions to set performance mode. Running validate in %s mode") 
+  } catch (const std::exception&) {
+    std::cout << boost::format("WARNING: User doesn't have admin permissions to set performance mode. Running validate in %s mode")
```
aktondak: Can we use XBUtil::logger or std::cerr here ? 
Also, we should catch the exception and include it in the warning. Maybe something like : 
"Failed to update pmode: *exception reason here*. Running validate in %s mode"

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -590,6 +590,8 @@ SubCmdValidate::execute(const SubCmdOptions& _options) const
   //get current performance mode
   const auto og_pmode = xrt_core::device_query_default<xq::performance_mode>(device, 0);
   const auto parsed_og_pmode = xq::performance_mode::parse_status(og_pmode);
+  // Tracks whether we successfully changed the pmode; only then should we reset it.
+  bool pmode_changed = false;
```
aktondak, line 593: This also feels untidy. We're getting into corner cases where the application is doing the handling of when and when not to make the pmode setting call. 
IMO, xrt-smi should always try to set pmode to performance irrespective of what original pmode. If it was already performance, driver/firmware can choose to ignore it, Its a driver optimization and driver decision. 
Lets get this behavior reviewed by others too.

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
AShivangi: Logger is specifically for per test details, errors. This is out of scope for it. I can rediect the message to cerr and include the error msg that comes from the driver

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
AShivangi: If we decide the driver should ignore the pmode setting when it's already in the requested state, then I'm fine removing this check. But if that's not possible, I can't remove it, xrt-smi validate has to remain a non-privileged command

aktondak: There is no guarantee of whether "pmode_changed" captures the true state of changed pmode anyway. A different process (say another spawned xrt-smi configure) can come while one is running and change the pmode in between. So we cannot rely on it. 
"xrt-smi validate" should be cleanly defined to be a privileged command and we should print a warning in case the user is not root irrespective of what pmode is set. 
Driver will ignore the pmode setting but it'll check the privileges first and we dont control that.

src/runtime_src/core/tools/xbutil2/SubCmdValidate.cpp
```diff
@@ -625,25 +629,23 @@ SubCmdValidate::execute(const SubCmdOptions& _options) const
...
+      xrt_core::device_update<xq::performance_mode>(device.get(), static_cast<xq::performance_mode::power_type>(og_pmode));
   } catch (const xq::no_such_key&) {
     // Do nothing, as performance mode setting is not supported
-  } catch (const std::exception & ex) { //check if permission was denied, i.e., no sudo access
-    if(!boost::icontains(ex.what(), "Permission denied")) {
-      std::cerr << boost::format("\nERROR: %s\n") % ex.what();
-      throw xrt_core::error(std::errc::operation_canceled);
-    }
+  } catch (const std::exception & ex) {
+    std::cerr << boost::format("\nERROR: %s\n") % ex.what();
```
aktondak: This should also be a warning as the above handling when trying to set. We should create a common API for this and call the same at both sites.

AShivangi: I think this should be an error. Here we fail as we are not able to "reset" to the original pmode. Think of it as cleanup failing

aktondak: This error is hitting run environments and causing issues. IMO this should be a warning 
https://jira.xilinx.com/browse/AIESW-40231

aktondak:
Thanks for the update.
We should put all of setting and resetting device query in one single try block and catch it with warning. It will be much cleaner and we'll not have to deal with empty catches and maintaining "pmode_changed".

aktondak:
> Thanks for the update. We should put all of setting and resetting device query in one single try block and catch it with warning. It will be much cleaner and we'll not have to deal with empty catches and maintaining "pmode_changed".

Actually we can't do that because we are running the tests even if setting pmode failed to maintain backward compatibility.
Why have we removed the original warning for unable to set pmode ? This will change the current behavior on linux. That's a genuine warning and should show it. There shouldn't be a reason for warning to show for resetting and not for setting

AShivangi:
> > Thanks for the update. We should put all of setting and resetting device query in one single try block and catch it with warning. It will be much cleaner and we'll not have to deal with empty catches and maintaining "pmode_changed".
> 
> Actually we can't do that because we are running the tests even if setting pmode failed to maintain backward compatibility. Why have we removed the original warning for unable to set pmode ? This will change the current behavior on linux. That's a genuine warning and should show it. There shouldn't be a reason for warning to show for resetting and not for setting

Do we show a warning on linux today? On windows we don't which is why I completely removed it. If you go through the email again you'll see what my concerns are about the warning being shown for performance mode. From my past experience, that kind of wording will give us a huge blow back and I don't want to deal with that. 

Setting doesn't have a warning since we _don't change_ anything and we already mention it in the validate header what pmode we're running in.
Resetting has a warning since we _changed_ the state of the device and now we need to inform the user that we ahve left the device not in its og state. 

I will try to clean up the try/catches

aktondak:
Yes the warning shows up on linux and I think its a fair and valid warning and should show up on windows too. But I'll let you/other reviewers decide:
aktondak@xsjstrix45:/proj/xsjhdstaff3/aktondak/xdna$ xrt-smi validate 
**WARNING: User doesn't have admin permissions to set performance mode. Running validate in default mode**
Validate Device           : [0000:c5:00.1]
    Platform              : NPU Strix
    Power Mode            : default

Test 3 [0000:c5:00.1]     : throughput 
    Details               : Average throughput: 50677.0 op/s                    

Validation completed

AShivangi:
Closing this PR as it is not needed for now

PR #9938, opened 2026-07-25
fix bad any_cast in `xrt::info::device::kdma`

simeonschaub:
Was only triggered when an exception got thrown, but I think it's good to fix nonetheless.

Discovered by Claude


stsoe (approved):
Change is good

stsoe:
+ @amd-gevadlam .  Why does XRT-CI not process this PR?

PR #9939, opened 2026-07-25
build: various mingw fixes

simeonschaub:
These were needed in order to cross-compile for windows in JuliaPackaging/Yggdrasil#14225. There were some additional patches in that PR, but I excluded the more controversial/hacky fixes for now. Some of the patches were written with help from Claude, but verified and reviewed one-by-one.

Signed-off-by: Simeon David Schaub <simeon@schaub.rocks>

### DIFF
--- a/src/runtime_src/core/common/detail/windows/utils.h
+++ b/src/runtime_src/core/common/detail/windows/utils.h
@@ -52,6 +52,7 @@ sys_dep_get_last_err_msg()
 inline std::string
 getenv(const char* name)
 {
+#ifdef _MSC_VER
   char* value = nullptr;
   size_t len = 0;
 
@@ -62,6 +63,10 @@ getenv(const char* name)
   // Use unique_ptr to ensure memory is freed even if string constructor throws
   std::unique_ptr<char, decltype(&std::free)> guard(value, &std::free);
   return std::string(guard.get());
+#else
+  const char* value = std::getenv(name);
+  return value ? std::string(value) : std::string{};
+#endif
 }
 
 inline std::string
--- a/src/runtime_src/core/common/smi/smi.h
+++ b/src/runtime_src/core/common/smi/smi.h
@@ -30,7 +30,7 @@ struct basic_option {
   std::string m_type;
 };
 
-class option : public basic_option {
+class XRT_CORE_COMMON_EXPORT option : public basic_option {
   std::string m_alias;
   std::string m_default_value;
   std::string m_value_type;
@@ -76,7 +76,7 @@ public:
 // This class is used to represent an option with a multiline description.
 // For example, --run can have multiple test names as its description.
 // These subnames are also queried using the generic API get_list
-class listable_description_option : public option {
+class XRT_CORE_COMMON_EXPORT listable_description_option : public option {
   std::vector<basic_option> m_description_array;
 public:
   listable_description_option(std::string name, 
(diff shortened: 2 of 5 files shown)


stsoe (changes requested):
Trace logging conditionals do not seem correct.  Explain how common/trace.cpp compiles?
Please separate trace changes into its own PR.
Thanks!

simeonschaub:
> Trace logging conditionals do not seem correct. Explain how common/trace.cpp compiles?

That particular patch came from a previous Ygddrasil build, so I must admit I am not familiar with the exact details behind it. I did confirm that the cross build fails without that patch, though it did include a very similar change to `src/runtime_src/core/common/detail/windows/trace_init.h`, I just noticed I didn't include here. I have dropped it for now from this PR.

src/runtime_src/core/common/smi/smi.h
```diff
@@ -30,7 +30,7 @@ struct basic_option {
   std::string m_type;
 };
 
-class option : public basic_option {
+class XRT_CORE_COMMON_EXPORT option : public basic_option {
```
stsoe: Class level exporting is not a good practice, why was this added?  What symbol is being referenced that is not already exported?

src/runtime_src/core/common/smi/smi.h
```diff
@@ -76,7 +76,7 @@ class option : public basic_option {
 // This class is used to represent an option with a multiline description.
 // For example, --run can have multiple test names as its description.
 // These subnames are also queried using the generic API get_list
-class listable_description_option : public option {
+class XRT_CORE_COMMON_EXPORT listable_description_option : public option {
```
stsoe: Ditto.

stsoe (changes requested):

src/runtime_src/core/common/smi/smi.h
simeonschaub: IIRC, it was due to interplay with us having to patch other headers due to mingw not liking certain `__declspec(dllimport)` annotations. I can also remove these changes from this PR

src/runtime_src/tools/xclbinutil/aie-pdi-transform/src/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright (C) 2018-2022 Xilinx, Inc. All rights Reserved.
 # Copyright (C) 2022-2026 Advanced Micro Devices, Inc. All rights Reserved.
-if(NOT WIN32)
+if(NOT MSVC)
   add_compile_options(-Wextra -fvisibility=default)
 else()
   add_compile_options(/wd4244 /wd4267 /wd4477 /wd4245)
```
stsoe, line 7: Reverse, /wd* are MSVC specific

src/runtime_src/core/common/detail/windows/utils.h
```diff
@@ -62,6 +63,10 @@ getenv(const char* name)
   // Use unique_ptr to ensure memory is freed even if string constructor throws
   std::unique_ptr<char, decltype(&std::free)> guard(value, &std::free);
   return std::string(guard.get());
+#else
+  const char* value = std::getenv(name);
+  return value ? std::string(value) : std::string{};
+#endif
```
stsoe, line 69: It's not clear that these changes check for MinGW, please adjust, e.g. maybe `#ifdef __MINGW32__ / #else `.
Also changes seem a little haphazardly, for example, I don't believe strerror_s works with MinGW, so why wasn't this fixed?     Is this PR even needed, it is a concerted effort to build XRT with MinGW, or is just change a little here and there?

PR #9941, opened 2026-07-26
[Windows] Fix AIEBU linker crash from IPO

thomthehound:

The x64 Windows Release build fails while linking `aiebu-asm`. MSVC `link.exe` exits with `0xC00000FD` (`STATUS_STACK_OVERFLOW`) during LTCG code generation.

The Windows build enables `/GL` and `/LTCG` through directory-wide options, so AIEBU inherits IPO even though its standalone Windows build does not enable it.

Technically, PR #9500 introduced the failure by enabling `/GL` and `/LTCG` for non-Debug targets under `nativeWin.cmake`, but this was not a problem until recent AIEBU revisions.

The issue was found during a full x64 Windows Release build with Visual Studio 18 Community and MSVC 14.51.36231.

Replace the manual `/GL`, `/LTCG`, and `/LTCG:OFF` pairing with CMake's `INTERPROCEDURAL_OPTIMIZATION`.

IPO is still enabled for non-Debug targets. It is only disabled while AIEBU targets are created, then restored for subsequent targets.

CastGuard follows each target's IPO property because `/guard:cast` requires `/GL`. This removes MSVC warning D9007 from AIEBU targets while preserving CastGuard where IPO remains enabled.

Alternatives considered:

* Disabling IPO for all Windows targets was rejected because only AIEBU triggers the linker failure.
* Disabling IPO only for `aiebu_asm_objects` was tested and still produced the same linker stack overflow.
* Removing and restoring literal `/GL` and `/LTCG` flags around AIEBU was rejected because it depends on their exact spelling and placement.

Low.

Non-Debug targets retain IPO and CastGuard. AIEBU retains its normal Release optimization settings, including `/O2` and `/Ot`. It simply no longer inherits IPO or CastGuard. The impact of this should be almost unnoticeable. 

**The change is limited to MSVC builds.**

* Completed a full x64 Windows Release `ALL_BUILD` with Visual Studio 18 Community and MSVC 14.51.36231.
* Build result: 0 errors and 0 warnings.
* Confirmed that AIEBU targets do not receive `/GL`, `/LTCG`, `/LTCGOUT`, `/guard:cast`, or `/d2CastGuardFailureMode:fastfail`.
* Confirmed that subsequent Release targets retain IPO and CastGuard.

None.

### DIFF
--- a/src/CMake/nativeWin.cmake
+++ b/src/CMake/nativeWin.cmake
@@ -36,6 +36,16 @@ if (MSVC)
   # Static linking with the CRT
   set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
 
+  # Use target-initialized IPO so compiler and linker settings stay paired
+  # when a subdirectory overrides the policy.
+  set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ON)
+  set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_DEBUG OFF)
+
+  # CastGuard requires /GL; key it to each target's IPO property rather than
+  # the build configuration alone.
+  set(_xrt_msvc_cast_guard_enabled
+    "$<AND:$<NOT:$<CONFIG:Debug>>,$<BOOL:$<TARGET_PROPERTY:INTERPROCEDURAL_OPTIMIZATION>>>")
+
   add_compile_options(
     /MT$<$<CONFIG:Debug>:d>  # static linking with the CRT
     /Zc:__cplusplus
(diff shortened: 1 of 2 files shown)


stsoe (commented):
Hi @thomthehound .  I too observed failures when building AIEBU VS 2025.  Please confirm this is what you saw:
https://github.com/Xilinx/XRT/commit/73611ef185f155eaee22c90ba18663b34cc646a5
I didn't have the energy to get to the bottom of it back then.

I a little vary making changes here because these options have been pushed on us based on MSFT requirements.   We to make sure all release builds continue to use the required options, and if that is not possible, then escalate.    I guess what I would like to understand is why this is specific to AIEBU,  is it too aggressive on the stack at link time and why?

thomthehound:
> Hi @thomthehound . I too observed failures when building AIEBU VS 2025. Please confirm this is what you saw: [73611ef](https://github.com/Xilinx/XRT/commit/73611ef185f155eaee22c90ba18663b34cc646a5) I didn't have the energy to get to the bottom of it back then.
> 
> I a little vary making changes here because these options have been pushed on us based on MSFT requirements. We to make sure all release builds continue to use the required options, and if that is not possible, then escalate. I guess what I would like to understand is why this is specific to AIEBU, is it too aggressive on the stack at link time and why?

Hi, Soren! It's always a pleasure!

Yes, that is *exactly* what I saw.

This was the smallest scope I found during testing that still builds properly while preserving the required options everywhere else. But I understand the concern, and I'll investigate the AIEBU failure further. As it stands, though, I don't see a narrower fix that works here.

stsoe:
Hi @sonals.  This is due to the same error I encountered months back.  We probably should get to the bottom as to why AIEBU fails building with required options with VS 2025 when it builds with VS 2022.

thomthehound:
right. `window2022`. I hadn't accounted for that. This can't work with such and old version of CMake. We'd have to lose CastGuard for all of AIEBU, which is unacceptable.

Well, I guess that means looking at AIEBU then. Sorry about wasting your time.

thomthehound:
Closing this due to the source-level fix in [AIEBU #334](https://github.com/Xilinx/aiebu/pull/334).

PR #9942, opened 2026-07-26
[SWDEV-577798] ASAN: scope MSVC CRT flags for /fsanitize=address builds

jhoyamd:
Problem solved by the commit

  -asan (/fsanitize=address) builds of XRT failed on Windows/MSVC for both x64 and ARM64 due to C runtime (CRT) and
  ASan-runtime library mismatches. AddressSanitizer requires the dynamic CRT (/MD), but the XRT Windows build hard-codes
   the static CRT (/MT) for static-Boost tool targets, producing LNK2038/LNK1319 runtime-library-mismatch link failures.
   This commit makes ENABLE_ASAN builds link and run cleanly while leaving the default (non-ASAN) build completely
  unchanged.

  Not a regression fix — no prior PR introduced a defect. This is enablement work: the existing Windows CMake hardening
  (static /MT CRT, Hybrid-CRT ucrt swap) was simply never reconciled with /fsanitize=address, which mandates the dynamic
   CRT. Discovered when attempting to produce an ASAN-instrumented XRT build (Jira SWDEV-577798): the VS generator maps
  an explicit /MT into per-object <RuntimeLibrary>, overriding the per-target MSVC_RUNTIME_LIBRARY property, so
  instrumented /MD objects collided with /MT at link time.

  All changes are strictly scoped behind ENABLE_ASAN and are no-ops on the default build:
  - nativeWin.cmake: on the ASAN path select the dynamic CRT via CMAKE_MSVC_RUNTIME_LIBRARY=MultiThreadedDLL, and emit
  the explicit /MT only when NOT ENABLE_ASAN. Skip the Hybrid-CRT ucrt swap under ASAN so the static ASan malloc thunk
  can intercept the allocator cleanly. Derive the ASan lib dir from the compiler path (lib/x64 or lib/arm64), add
  /fsanitize=address and /INCREMENTAL:NO.
  - xclbinutil / xbutil2 (xrt-smi) CMakeLists: keep these /MT static-Boost tool targets building without ASAN
  instrumentation, so they are not dropped from build output — the driver INF requires xclbinutil.exe / xrt-smi.exe.
  - aiebu submodule bump for the matching ASAN CRT scoping.

  Alternatives rejected: (1) Instrumenting xclbinutil/xrt-smi too — rejected because they link prebuilt non-instrumented
   static Boost, forcing /MD there yields container-overflow annotation mismatches (LNK2038
  annotate_string/annotate_vector). (2) Forcing /MD globally via an explicit compile flag — rejected because the VS
  generator overrides the per-target property, re-breaking the /MT tool targets. Using CMAKE_MSVC_RUNTIME_LIBRARY +
  guarding explicit flags is the only combination that lets per-target overrides survive.

  Low. Every change is gated on ENABLE_ASAN (default OFF), so the shipping/default build path is bit-for-bit unaffected.
   Residual risk is confined to the opt-in ASAN configuration: the ASan lib-dir derivation assumes the standard MSVC
  bin/Hostx64/<arch> → lib/<arch> layout, and ARM64 ASAN requires MSVC v145 (VS 2026).

  Both x64 and ARM64 -asan builds were run to completion — GREEN. Verified xclbinutil.exe and xrt-smi.exe are present in
   the build output (INF requirement). Default (non-ASAN) build path unchanged by construction (all edits behind NOT
  ENABLE_ASAN / ENABLE_ASAN guards). Requested additional testing: runtime execution of an ASAN-instrumented XRT load to
   confirm the interceptor initializes and no CRT double-allocation occurs at process start.

  Documentation impact (if any)

  None. No public API, CLI, or user-facing behavior change; ENABLE_ASAN is an internal opt-in developer build flag. A
  short note in the Windows build README on enabling ASAN (and the ARM64 VS 2026 / v145 requirement) would be a
  nice-to-have but is not required by this commit.

### DIFF
--- a/src/CMake/nativeWin.cmake
+++ b/src/CMake/nativeWin.cmake
@@ -32,12 +32,24 @@ add_compile_definitions("BOOST_BIND_GLOBAL_PLACEHOLDERS")
 # the warning is bogus.  Remove defintion when fixed
 add_compile_definitions("_SILENCE_CXX17_ALLOCATOR_VOID_DEPRECATION_WARNING")
 
+option(ENABLE_ASAN "Build with AddressSanitizer (/fsanitize=address). Requires MSVC v145 (VS 2026) for ARM64." OFF)
+
 if (MSVC)
-  # Static linking with the CRT
-  set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+  if (ENABLE_ASAN)
+    # Always use /MD (not /MDd): ASAN has its own heap tracking and does not
+    # integrate with the debug CRT. Avoids CMake 4.x try_compile rejection of
+    # MultiThreadedDLLDebug for the ARM64 cross-compiler.
+    set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreadedDLL")
+  else()
+    set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+  endif()
 
   add_compile_options(
-    /MT$<$<CONFIG:Debug>:d>  # static linking with the CRT
+    # Explicit /MT only on the non-ASAN path. On the ASAN path CRT selection is
+    # driven by CMAKE_MSVC_RUNTIME_LIBRARY / the per-target MSVC_RUNTIME_LIBRARY
+    # property: an explicit /MD here would be mapped into <RuntimeLibrary> by the
+    # VS generator and override the per-target /MT that xclbinutil/xrt-smi need.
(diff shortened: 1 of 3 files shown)


stsoe (changes requested):
The detailed PR description (written by AI, presumably) misses the "why" this is needed and "what" problem it solves beyond the mechanics to make the build work.

jhoyamd:
> The detailed PR description (written by AI, presumably) misses the "why" this is needed and "what" problem it solves beyond the mechanics to make the build work.

Commit message updated with why and how.

stsoe (approved):
LGTM.   XRT's CMake is inflexible and should change so options like this one is easier to accommodate.  I think XRT should honor CMAKE_MSVC_RUNTIME_LIBRARY set by parents with appropriate defaults.

PR #9947, opened 2026-07-31
SWSPLAT-30708 fix bo_cache bo_size regression from 9f49fda (#9895)

stsoe:
9f49fda (PR #9895) renamed m_bo_size to bo_size and accidentally introduced a * 1024u multiplier, inflating the exec BO allocation from 4 KB to 4 MB.

Remove the multiplier; BoSize has always been in bytes.


stsoe:
Fixed in #9946

PR #9955, opened 2026-08-04
Fix xilinx_xrt() root discovery on Debian multiarch installs

wendyliang25:
xilinx_xrt() derived the XRT root by taking two parent_path() levels up from the libxrt_coreutil.so path reported by dladdr(). This assumes coreutil lives directly in <root>/lib, which only holds when CMAKE_INSTALL_LIBDIR is a single path component (lib, lib64). On a Debian multiarch install CMAKE_INSTALL_LIBDIR is lib/x86_64-linux-gnu, so coreutil is one level deeper and the two-level strip returns <root>/lib instead of <root>. shim_path() then re-appends XRT_LIB_DIR, producing a doubled path such as
'/opt/amdgpu/lib/lib/x86_64-linux-gnu/libxrt_core.so.2' and the load fails with "No such library".

Strip exactly as many trailing components as XRT_LIB_DIR contains instead of a hardcoded two levels, so the root is recovered correctly for lib, lib64, and lib/x86_64-linux-gnu layouts alike. Single-component layouts (standard, lib64, single-arch Yocto) are unaffected. Also default XRT_LIB_DIR to "lib" so the header stays self-contained; the loop additionally tolerates an empty value.

Regression from 11169ba50 (SWSPLAT-24084) which switched XRT root discovery from XILINX_XRT to dladdr().

### DIFF
--- a/src/runtime_src/core/common/detail/linux/xilinx_xrt.h
+++ b/src/runtime_src/core/common/detail/linux/xilinx_xrt.h
@@ -24,9 +24,21 @@
 # error "XRT_VERSION_STRING is undefined"
 #endif
 
+// XRT_LIB_DIR (CMAKE_INSTALL_LIBDIR) is supplied as a compile definition on
+// module_loader.cpp, the only translation unit that includes this header.
+// This header is not meant to be compiled stand-alone, so error out if it is
+// missing rather than guessing a default.
+#ifndef XRT_LIB_DIR
+# error "XRT_LIB_DIR is undefined"
+#endif
+
 namespace xrt_core::detail {
 
 namespace sfs = std::filesystem;
+
+// The code below also tolerates an empty value (coreutil then sits directly in
+// the XRT root).
+constexpr const char* xrt_lib_dir = XRT_LIB_DIR;
 
 // Get XRT install path from DSO location or compile time constant.
 //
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/detail/linux/xilinx_xrt.h
```diff
@@ -28,6 +28,20 @@ namespace xrt_core::detail {
...
+// if it is ever compiled without that definition, but do not #define the macro:
+// that would defeat the #ifndef XRT_LIB_DIR / #error guard in module_loader.cpp
+// (included before that guard runs) and silently mask a missing build setting.
+// The code below also tolerates an empty value (coreutil then sits directly in
+// the XRT root).
+#ifdef XRT_LIB_DIR
+constexpr const char* xrt_lib_dir = XRT_LIB_DIR;
+#else
+constexpr const char* xrt_lib_dir = "lib";
+#endif
```
stsoe: The header file is not supposed to be used stand-alone or by compilation units where XRT_LIB_DIR is not defined.  Can we error if not defined?

src/runtime_src/core/common/detail/linux/xilinx_xrt.h
```diff
@@ -58,9 +72,19 @@ xilinx_xrt()
...
+    // XRT_LIB_DIR suffix from the directory dladdr() reported to recover
+    // the XRT root that callers expect (e.g. /opt/xilinx/xrt). XRT_LIB_DIR
+    // may be one or more path components ("lib", "lib64", or
+    // "lib/x86_64-linux-gnu"), so remove one component per component in it
+    // rather than hardcoding a fixed number of levels.
+    sfs::path xrt_root = sfs::path(info.dli_fname).parent_path();
+    for (const auto& comp : sfs::path(xrt_lib_dir)) {
+      (void) comp;
+      xrt_root = xrt_root.parent_path();
+    }
```
stsoe, line 88: Technically CMAKE_INSTALL_LIBDIR could be an absolute path.  If set to /usr/lib, then we'd over strip, I think.
```suggestion
    for (const auto& comp : sfs::path(xrt_lib_dir)) {
      if (comp.empty() || comp == comp.root_directory())
        continue;
        
      xrt_root = xrt_root.parent_path();
    }
```

stsoe (changes requested):
Yeah, true, the code was wrong.  Thanks for fixing. 
A few suggestions, one being a corner case.

src/runtime_src/core/common/detail/linux/xilinx_xrt.h
wendyliang25: fixed, skipping the root-directory and  empty components

stsoe (approved):

PR #9961, opened 2026-08-05
CodeQL Fix OOB reads in SectionAIEResourcesBin

stsoe:
Four CWE-125 heap OOB reads in copyBufferUpdateMetadata() where aie_resources_bin mpo_* fields (mpo_name, mpo_version, m_start_column, m_num_columns) were used as raw pointer offsets without bounds checking when computing sDefault fallback strings, bypassing the bounded_mpo_cstr() guard added in bb1791f6.

Two additional CWE-125 OOB reads where m_image_offset and m_image_size were used without bounds checking in both copyBufferUpdateMetadata() and writeObjImage() before reading image data.

Jira tickets:
- AIESW-41124
- AIESW-41125
- AIESW-41126
- AIESW-41127
- AIESW-41128
- AIESW-41129

CodeQL alerts #135-#140 (amd-psirt/xclbin-parser-oob, HIGH severity). The sDefault reads were missed by bb1791f6 (SWSPLAT-30717), which protected the TRACE() and writeMetadata() paths but not the fallback default value paths in copyBufferUpdateMetadata(). The image offset/size checks were never added to writeObjImage().

Replaced the four raw sDefault pointer calculations with bounded_mpo_cstr(pHdr, pHdr->field, _origSectionSize). This also fixes a pre-existing double-offset bug in the original code: the sDefault expressions incorrectly added sizeof(aie_resources_bin) on top of the mpo_* field value, but mpo_* fields already store absolute offsets from pHdr (sizeof is baked in at write time), so the fallback pointer was pointing past the actual string.

Added explicit uint64_t overflow-safe bounds checks for m_image_offset + m_image_size against section size in both copyBufferUpdateMetadata() and writeObjImage().

Low. The sDefault path is only taken when the JSON metadata stream omits a key; the double-offset bug meant the fallback was producing wrong values anyway. The image bounds checks only reject malformed xclbins.

Built xclbinutil successfully. Recommend testing xclbinutil --input with a crafted xclbin containing OOB mpo offsets and verifying it throws rather than reading past the buffer.

None

### DIFF
--- a/src/runtime_src/tools/xclbinutil/SectionAIEResourcesBin.cxx
+++ b/src/runtime_src/tools/xclbinutil/SectionAIEResourcesBin.cxx
@@ -200,7 +200,7 @@ SectionAIEResourcesBin::copyBufferUpdateMetadata(const char* _pOrigDataSection,
 
   // mpo_name
   {
-    auto sDefault = reinterpret_cast<const char*>(pHdr) + sizeof(aie_resources_bin) + pHdr->mpo_name;
+    auto sDefault = XUtil::bounded_mpo_cstr(pHdr, pHdr->mpo_name, _origSectionSize);
     auto sValue = ptSK.get<std::string>("name", sDefault);
 
     if (sValue.compare(getSectionIndexName()) != 0) {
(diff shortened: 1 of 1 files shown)


xfreid (approved):

stsoe:
Replaced by https://github.com/Xilinx/XRT/pull/9966

PR #9962, opened 2026-08-05
Fix uint64_t overflow in SectionBMC bounds check

stsoe:
copyBufferUpdateMetadata() computed `pHdr->m_offset + pHdr->m_size` as a uint64_t addition before comparing against _origSectionSize. If both fields are near UINT64_MAX the addition wraps silently, causing the bounds check to pass on a malformed section.

#### How problem was solved, alternative solutions (if any) and why they were rejected Replaced the addition with an overflow-safe two-part check:
  m_size > _origSectionSize || m_offset > _origSectionSize - m_size

None


xfreid (approved):

stsoe:
Replaced by https://github.com/Xilinx/XRT/pull/9966

PR #9967, opened 2026-08-06
[pyxrt] Expose controller transaction-result BO allocation

thomthehound:

Controller transactions can return ordered register-read results through a hardware-context-owned result BO, but C++ and Python clients have no supported way to allocate that BO. This blocks host runtimes from consuming results such as AIE DMA Finish-on-TLAST counts.

This is a missing API rather than a regression. It was found while adding host-visible Finish-on-TLAST completion counts to MLIR-AIE.

Add the experimental C++ API `xrt::ext::transaction_result_bo()`, declared in `xrt_ext.h` and implemented in `xrt_coreutil`. The implementation calls `xrt_core::bo_int::create_bo(..., use_type::debug)` internally. PyXRT exposes the same operation as `pyxrt.ext.transaction_result_bo()` by binding the public C++ API.

An earlier revision declared the internal allocator in the public header so a header-inline wrapper could remain compatible with the currently shipped Windows runtime. That approach was rejected because `bo_int` is private and subject to change. The revised implementation keeps the internal API entirely inside `xrt_coreutil`.

This adds one experimental `xrt_coreutil` export. Existing API behavior and existing client compatibility are unchanged, but clients using the new function require an XRT runtime containing that export. Current Windows driver distributions do not contain it.

The public C++ header and a direct call to the new API compile against current XRT master. The modified `xrt_bo.cpp` compiles to an object, `pyxrt.cpp` passes syntax compilation, and a linked audit shared object exports `xrt::ext::transaction_result_bo()`. The Python type stub compiles.

The underlying result-BO allocation and Finish-on-TLAST path were exercised on Windows through the earlier compatibility prototype. This corrected version only changes how they are exposed.

Adds API documentation in `xrt_ext.h` and the corresponding Python type stub.

### DIFF
--- a/src/python/pybind11/pyxrt.pyi
+++ b/src/python/pybind11/pyxrt.pyi
@@ -1174,7 +1174,7 @@ class ext:
 
     @staticmethod
     def transaction_result_bo(hwctx: hw_context, size: SupportsInt) -> bo:
-        """Allocate controller transaction-result storage for one submission."""
+        """Allocate controller transaction-result storage for a hardware context."""
         ...
     
     # Type aliases for the ext submodule
--- a/src/python/pybind11/pyxrt.pyi
+++ b/src/python/pybind11/pyxrt.pyi
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: MIT
-# Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (C) 2025-2026 Advanced Micro Devices, Inc. All rights reserved.
 from __future__ import annotations
 import sys
 from enum import IntEnum
@@ -1171,6 +1171,11 @@ class _ext_kernel(kernel):
 
 class ext:
     """Extended XRT functionality."""
+
+    @staticmethod
+    def transaction_result_bo(hwctx: hw_context, size: SupportsInt) -> bo:
+        """Allocate controller transaction-result storage for one submission."""
+        ...
     
     # Type aliases for the ext submodule
     access_mode = _ext_access_mode
--- a/src/python/pybind11/src/pyxrt.cpp
+++ b/src/python/pybind11/src/pyxrt.cpp
@@ -710,7 +710,7 @@ PYBIND11_MODULE(pyxrt, m) {
             return xrt::ext::transaction_result_bo(hwctx, size);
         },
         py::arg("hwctx"), py::arg("size"),
-        "Allocate controller transaction-result storage for one submission.");
+        "Allocate controller transaction-result storage for a hardware context.");
     
     py::class_<xrt::ext::kernel, xrt::kernel> pyextkernel(ext, "kernel", "Extended kernel object for module-backed and shared workflows.");
 
--- a/src/python/pybind11/src/pyxrt.cpp
+++ b/src/python/pybind11/src/pyxrt.cpp
@@ -703,6 +703,14 @@ PYBIND11_MODULE(pyxrt, m) {
         }),
         py::arg("hwctx"), py::arg("pid"), py::arg("export_handle"),
         "Import an extended buffer into a hardware context from another process using an exported handle.");
+
+    ext.def("transaction_result_bo",
+        [](const xrt::hw_context& hwctx, size_t size) {
+            py::gil_scoped_release release;
+            return xrt::ext::transaction_result_bo(hwctx, size);
+        },
+        py::arg("hwctx"), py::arg("size"),
+        "Allocate controller transaction-result storage for one submission.");
     
     py::class_<xrt::ext::kernel, xrt::kernel> pyextkernel(ext, "kernel", "Extended kernel object for module-backed and shared workflows.");
 
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -1731,6 +1731,12 @@ bo(const xrt::hw_context& hwctx, pid_type pid, xrt::bo::export_handle ehdl)
   : xrt::bo::bo{alloc_import_from_pid(device_type{hwctx}, pid, ehdl)}
 {}
 
+xrt::bo
+transaction_result_bo(const xrt::hw_context& hwctx, size_t size)
+{
+  return xrt_core::bo_int::create_bo(hwctx, size, xrt_core::bo_int::use_type::debug);
+}
+
 } // xrt::ext
 
 ////////////////////////////////////////////////////////////////
--- a/src/runtime_src/core/include/xrt/experimental/xrt_ext.h
+++ b/src/runtime_src/core/include/xrt/experimental/xrt_ext.h
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright (C) 2023-2025 Advanced Micro Devices, Inc. All rights reserved.
+// Copyright (C) 2023-2026 Advanced Micro Devices, Inc. All rights reserved.
 #ifndef XRT_EXT_H_
 #define XRT_EXT_H_
 
@@ -18,6 +18,19 @@
 #endif
 
 #ifdef __cplusplus
+namespace xrt_core::bo_int {
+
+/// @cond
+// Existing coreutil entry point used by the header-only experimental wrapper.
+enum class use_type : int;
+
+XRT_API_EXPORT
+xrt::bo
+create_bo(const xrt::hw_context& hwctx, size_t size, use_type type);
+/// @endcond
+
+} // namespace xrt_core::bo_int
+
 namespace xrt::ext {
 
 /*!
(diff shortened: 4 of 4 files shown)


stsoe (commented):
Hi @thomthehound . I do not want external APIs changed for the sake of supporting pyxrt.cpp. create_bo is an internal API, not of interest to external code, and it is subject to change any time, thus cannot be exposed.

thomthehound:
> Hi @thomthehound . I do not want external APIs changed for the sake of supporting pyxrt.cpp. create_bo is an internal API, not of interest to external code, and it is subject to change any time, thus cannot be exposed.

Understood and agreed. Before I revise or close this PR, may I ask whether an exported `xrt::ext::transaction_result_bo()`, *declared* in `xrt_ext.h` but *implemented* in `xrt_coreutil`, would be acceptable?

The alternative I had in mind would keep `bo_int::create_bo` entirely internal. `xrt::ext::transaction_result_bo()` could then call it inside `xrt_coreutil`, and `pyxrt.cpp` would just bind that public C++ API. The capability is not intended to be Python-only; Python is just the easiest path for me to test at the moment.

Of course, this would add a new `xrt_coreutil` export and therefore would not work with the current Windows drivers. It would, however, provide the path needed to consume Finish-on-TLAST results without exposing private APIs.

Thank you for your time, Soren!

thomthehound:
Actually, in just a little bit I can update this PR to show you what I mean instead of telling you about it. I had already implemented this earlier, before I realized I had no good means to properly test it. But now that I know it will work, I feel more confident about it. I just need to dig up some older files and make sure it still compiles.

stsoe:
> > Hi @thomthehound . I do not want external APIs changed for the sake of supporting pyxrt.cpp. create_bo is an internal API, not of interest to external code, and it is subject to change any time, thus cannot be exposed.
> 
> Understood and agreed. Before I revise or close this PR, may I ask whether an exported `xrt::ext::transaction_result_bo()`, _declared_ in `xrt_ext.h` but _implemented_ in `xrt_coreutil`, would be acceptable?
> 
> The alternative I had in mind would keep `bo_int::create_bo` entirely internal. `xrt::ext::transaction_result_bo()` could then call it inside `xrt_coreutil`, and `pyxrt.cpp` would just bind that public C++ API. The capability is not intended to be Python-only; Python is just the easiest path for me to test at the moment.
> 
> Of course, this would add a new `xrt_coreutil` export and therefore would not work with the current Windows drivers. It would, however, provide the path needed to consume Finish-on-TLAST results without exposing private APIs.
> 
> Thank you for your time, Soren!

A new API, as you suggest, implemented entirely within coreutil is fine, but first I need to understand the use case.  The comment (controller) for transaction_result_bo was not clear to me.

thomthehound:
> > > Hi @thomthehound . I do not want external APIs changed for the sake of supporting pyxrt.cpp. create_bo is an internal API, not of interest to external code, and it is subject to change any time, thus cannot be exposed.
> > 
> > 
> > Understood and agreed. Before I revise or close this PR, may I ask whether an exported `xrt::ext::transaction_result_bo()`, _declared_ in `xrt_ext.h` but _implemented_ in `xrt_coreutil`, would be acceptable?
> > The alternative I had in mind would keep `bo_int::create_bo` entirely internal. `xrt::ext::transaction_result_bo()` could then call it inside `xrt_coreutil`, and `pyxrt.cpp` would just bind that public C++ API. The capability is not intended to be Python-only; Python is just the easiest path for me to test at the moment.
> > Of course, this would add a new `xrt_coreutil` export and therefore would not work with the current Windows drivers. It would, however, provide the path needed to consume Finish-on-TLAST results without exposing private APIs.
> > Thank you for your time, Soren!
> 
> A new API, as you suggest, implemented entirely within coreutil is fine, but first I need to understand the use case. The comment (controller) for transaction_result_bo was not clear to me.

Right. Sorry about that; I suppose I should probably explain why this is needed for Finish-on-TLAST.

FoT stops the DMA transfer when the actual data ends, which matters when the output size isn't known ahead of time. But the host still needs to know how much data was actually written. The hardware puts that count in the transaction-result buffer, but there is currently no reasonable way for the host to access it. So FoT technically "works", but the host cannot tell where the valid output ends... making the feature essentially unusable.

The immediate concern that brought me down this rabbit-hole is spilling compressed data out through the shim tile. Because the data is compressed, its exact size isn't known until it has been written. We need FoT to stop the transfer at the right point, but then we also need the returned count to know how much compressed data was actually written to the BO.

So, basically: compressed output needs Finish-on-TLAST, and Finish-on-TLAST needs its result count to be visible to the host. That is what this API is for.

stsoe:
> > > > Hi @thomthehound . I do not want external APIs changed for the sake of supporting pyxrt.cpp. create_bo is an internal API, not of interest to external code, and it is subject to change any time, thus cannot be exposed.
> > > 
> > > 
> > > Understood and agreed. Before I revise or close this PR, may I ask whether an exported `xrt::ext::transaction_result_bo()`, _declared_ in `xrt_ext.h` but _implemented_ in `xrt_coreutil`, would be acceptable?
> > > The alternative I had in mind would keep `bo_int::create_bo` entirely internal. `xrt::ext::transaction_result_bo()` could then call it inside `xrt_coreutil`, and `pyxrt.cpp` would just bind that public C++ API. The capability is not intended to be Python-only; Python is just the easiest path for me to test at the moment.
> > > Of course, this would add a new `xrt_coreutil` export and therefore would not work with the current Windows drivers. It would, however, provide the path needed to consume Finish-on-TLAST results without exposing private APIs.
> > > Thank you for your time, Soren!
> > 
> > 
> > A new API, as you suggest, implemented entirely within coreutil is fine, but first I need to understand the use case. The comment (controller) for transaction_result_bo was not clear to me.
> 
> Right. Sorry about that; I suppose I should probably explain why this is needed for Finish-on-TLAST.
> 
> FoT stops the DMA transfer when the actual data ends, which matters when the output size isn't known ahead of time. But the host still needs to know how much data was actually written. The hardware puts that count in the transaction-result buffer, but there is currently no reasonable way for the host to access it. So FoT technically "works", but the host cannot tell where the valid output ends... making the feature essentially unusable.
> 
> The immediate concern that brought me down this rabbit-hole is spilling compressed data out through the shim tile. Because the data is compressed, its exact size isn't known until it has been written. We need FoT to stop the transfer at the right point, but then we also need the returned count to know how much compressed data was actually written to the BO.
> 
> So, basically: compressed output needs Finish-on-TLAST, and Finish-on-TLAST needs its result count to be visible to the host. That is what this API is for.

I thought we had debug (XDP) support for this?  @jvillarre ?

thomthehound:
Looking through the XDP code was actually how I originally discovered this result-buffer existed. Before that, I thought what I was trying to do with FoT and compressed-data spill was impossible. It was just by chance that somebody suggested that I look at the tracing and debugging code -- for an unrelated purpose, no less -- when I discovered it. But XDP only uses the buffer internally, and only for its own diagnostic collection. I couldn't find a feasible way to get the transaction result back to the submitting program (and the headers say they shouldn't be providing user-facing APIs anyway).

stsoe:
> The hardware puts that count in the transaction-result buffer

How does it do that?  What prompts it to use this "special"  buffer?  It sounds like some side-band data used during kernel execution.

maxzhen:
> Looking through the XDP code was actually how I originally discovered this result-buffer existed. Before that, I thought what I was trying to do with FoT and compressed-data spill was impossible. It was just by chance that somebody suggested that I look at the tracing and debugging code -- for an unrelated purpose, no less -- when I discovered it. But XDP only uses the buffer internally, and only for its own diagnostic collection. I couldn't find a feasible way to get the transaction result back to the submitting program (and the headers say they shouldn't be providing user-facing APIs anyway).

The debug BO is not some sort of "result" buffer. It is not part of the argument list of the kernel. It is meant for debugging purpose and can be used to retrieve some debugging info through side band communication channel which is only known to the debugger. The data it contains is not useful during normal/production operation.

Looking at your use case, it sounds like the count should be part of the official argument list. As you pointed out, user can't use the compressed data without knowing the size. The data and size combined is the result. It seems that this result should be communicated back to host via normal BO, no?

jvillarre:
The debug buffer is a special internal buffer used and managed internally by several debug features and was not intended to be directly accessible from the host code.  The values in the debug buffer will not be consistent to the host code based on what debug features are enabled and may change at any time, so I do not think it should be exposed via a host callable API.

thomthehound:
Thank you all for taking the time to look at this. I better understand the concern now, and I agree. Trying to expose a debug BO as part of the interface is not the right thing to do.

I think I was still explaining this badly earlier. Finish-on-TLAST itself does not directly write the count into the debug BO. That is only a side-effect. It just puts a record in the S2MM count FIFO. When the transaction *reads* that FIFO register, though, the controller writes the returned value into the debug BO for that hardware context.

The problem is that the register-read operation only specifies *which* registers to read. It does not provide a destination BO or address. As far as I could tell, the debug BO was the only place the controller interface provides for those results. That is why I ended up using it, and why I ended up treating it as a transaction register buffer.

I agree that returning the count through a normal BO would be much better. Perhaps I just got so caught up in something I thought I saw that I missed the obvious answer. Please accept my apologies.

PR #9980, opened 2026-08-13
AIESW-41600 Consolidate throughput and latency tests for npu3

rchane:
https://jira.xilinx.com/browse/AIESW-41600
Consolidating throughput and latency xrt-smi validate tests for npu3, skip TCT tests as well
Dropped runlist-latency and runlist-throughput (return N/A on npu3, shows up as "skipped" in output) and moved logic of runlist-throughput to default throughput test, skipped TCT tests on npu3
Test deprecation, can break pipelines if numbers are expected when parsing output for the tests
Testing on STX/npu3
Will need to update xrt-smi user guide

### DIFF
--- a/src/runtime_src/core/tools/common/tests/TestNPUThroughput.cpp
+++ b/src/runtime_src/core/tools/common/tests/TestNPUThroughput.cpp
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright (C) 2024-2025 Advanced Micro Devices, Inc. All rights reserved.
+// Copyright (C) 2024-2026 Advanced Micro Devices, Inc. All rights reserved.
 
 // ------ I N C L U D E   F I L E S -------------------------------------------
 // Local - Include Files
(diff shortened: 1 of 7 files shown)


aktondak:
Looks good to me.
Can you post the result with latest VTD and try to merge this soon ?

AShivangi (approved):

aktondak (approved):

src/runtime_src/core/tools/common/tests/TestNPUThroughput.cpp
```diff
@@ -35,6 +37,19 @@ get_throughput_from_report(const json& report) const
...
+{
+  const auto runlist_throughput = get_throughput_from_report(report);
+  if (use_runlist) {
+    const auto recipe_runs = report.at("resources").at("runs").get<double>();
+    return runlist_throughput * recipe_runs;
+  } else {
+    return runlist_throughput;
+  }
+}
+
```
stsoe, line 51: Reduce scoping, exit early
```suggestion
TestNPUThroughput::
get_ops_throughput_from_report(const json& report, bool use_runlist) const
{
  const auto runlist_throughput = get_throughput_from_report(report);
  if (!use_runlist)
    return runlist_throughput;
    
  const auto recipe_runs = report.at("resources").at("runs").get<double>();
  return runlist_throughput * recipe_runs;
}

```

stsoe (commented):
Nit

src/runtime_src/core/tools/common/tests/TestNPUThroughput.cpp
rchane: Sure, updated! Thank you, Soren.

PR #9985, opened 2026-08-19
CR-1210115:Add non-blocking RTP update/read APIs.

svup-xilinx:
Blocking xrt::graph::update() and read() wait on AIE tile locks. HIL host code needs to poll RTP ports without blocking. This commit adds update_nb() and read_nb(); they return EAGAIN when the lock is held by the kernel.
No bug fix. New feature for [CR-1210115](https://jira.xilinx.com/browse/CR-1210115) / [VITIS-16628](https://jira.xilinx.com/browse/VITIS-16628).
Added update_nb() / read_nb() through the existing graph layers down to adf_runtime_api, using non-blocking lock acquire (LOCK_TIMEOUT_NB) and undoLockAcquire() on partial failure.
Low. Callers must retry on EAGAIN. Blocking RTP APIs are unchanged.
Tested on hardware (vck190): update_nb() and read_nb() return 0 when the lock is free and EAGAIN when the kernel holds it.
API comments added in xrt_graph.h for update_nb() and read_nb(). No other docs changed.

### DIFF
--- a/src/runtime_src/core/common/api/aie/xrt_graph.cpp
+++ b/src/runtime_src/core/common/api/aie/xrt_graph.cpp
@@ -119,6 +119,18 @@ public:
     m_graphHandle->read_graph_rtp(port, buffer, size);
   }
 
+  int
+  update_rtp_nb(const char* port, const char* buffer, size_t size)
+  {
+    return m_graphHandle->update_graph_rtp_nb(port, buffer, size);
+  }
+
+  int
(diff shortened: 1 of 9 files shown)


stsoe (commented):
LGTM.  Please update graph_handle.h to be stand-alone compilable.  I am not sure I ever saw this file, it is lacking comments, albeit just follows the standard shim dispatch mechanism.   At least other shim handles are documented a bit.

chvamshi-xilinx (approved):

PR #9989, opened 2026-08-20
Correctly enumerate PF device through xrt-smi when no VF present

hlaccabu:
xrt-smi examine failed with "No device handle" on PF device without VFs present.
Xrt-smi found the PF but opened it through the mgmt path. Xrt-smi runs in the user domain, with no VFs the user list is empty so index 0 had no user device. 
Keep PF in the mgmt device list, but open it through the user/shim path (get_userpf_device()) instead of the mgmt-only path (get_mgmtpf_device()).

### DIFF
--- a/src/runtime_src/core/pcie/linux/system_linux.cpp
+++ b/src/runtime_src/core/pcie/linux/system_linux.cpp
@@ -229,7 +229,9 @@ std::shared_ptr<device>
 system_linux::
 get_userpf_device(device::id_type id) const
 {
-  auto pdev = get_pcidev(id, true);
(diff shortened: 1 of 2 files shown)


AShivangi (approved):

aktondak:
@hlaccabu 
The device enumeration is not the issue here. The PF is enumerated correctly and is also a part of the list of devices. The problem is that the userhandle check should not happen on PF devices since there is no device handle for PF devices in the shim. 
This PR should be the right fix for this issue. You may close this one : 
https://github.com/amd/xdna-driver/pull/1621

PR #9993, opened 2026-08-24
zocl: probe CU IRQs from fpga_accelerator DT overlay

ManojTakasi:
As part of [VITIS-16571](https://jira.xilinx.com/browse/VITIS-16571) / [VITIS-16885](https://jira.xilinx.com/browse/VITIS-16885), CU interrupt routing is moving out of the static zyxclmm_drm node into a driver-agnostic fpga_accelerator node delivered with the RM .dtbo. zocl previously always obtained CU IRQs via platform_get_irq() on zyxclmm_drm at driver probe and never re-probed after a DT overlay update. This commit adds the new probe path, keeps backward compatibility, and refreshes the CU interrupt controller when an xclbin carries a PARTITION_METADATA dtbo overlay.

Created SH ticket: https://jira.xilinx.com/browse/SH-3057

Read interrupts-extended from the fpga_accelerator overlay node when present, with legacy platform_get_irq fallback, and refresh CU INTC after PARTITION_METADATA overlay apply so DFX xclbin loads pick up updated interrupt routing.

1. zocl_cu_intc_setup() If fpga_accelerator exists and has interrupts-extended, use of_irq_get(fpga_np, index); otherwise fall back to platform_get_irq(pdev, index) on zyxclmm_drm.
2. zocl_cu_intc_refresh() / zocl_cu_intc_fini() Tear down and recreate CU INTC using current live DT state.
3. zocl_edge_xclbin.c After successful PARTITION_METADATA section load, call zocl_cu_intc_refresh() before CU creation so DFX xclbin loads pick up updated overlay IRQ routing.
4. zert_cu_intc_init() / zert_cu_intc_refresh() Same fpga_accelerator probe path for ERT HW (ert_hw) platforms; delegate refresh from zocl when ert_hw is present.

low

Tested on VEK385 rev b by modifying the dtbo(included fpga_accelerator node with interrupts-extended region)

### DIFF
--- a/build/yocto.build
+++ b/build/yocto.build
@@ -1,4 +1,4 @@
 REPO_URL=https://gitenterprise.xilinx.com/ssw-devops/yocto-tagged-manifest.git
 BRANCH=master
-MANIFEST_PATH=/proj/yocto/edf/2026.2/stable/Yocto_edf_2026.2_06121253
-MANIFEST_FILE=default_06121253.xml
+MANIFEST_PATH=/proj/yocto/edf/2026.2/stable/Yocto_edf_2026.2_08182207
+MANIFEST_FILE=default_08182207.xml
--- a/src/runtime_src/core/edge/drm/zocl/common/zocl_drv.c
+++ b/src/runtime_src/core/edge/drm/zocl/common/zocl_drv.c
@@ -24,6 +24,7 @@
 #include <linux/pagemap.h>
 #include <linux/io.h>
 #include <linux/kernel.h>
+#include <linux/of.h>
 #include <linux/of_address.h>
 #include <linux/of_irq.h>
 #include <linux/poll.h>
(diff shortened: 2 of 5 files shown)


stsoe (commented):
This PR will have to be rebased after #9994 has been merged.  Or you can ignore the public github windows runner failures and merge as is.

chvamshi-xilinx (changes requested):
Please make sure below four cases are covered
	1) Zocl node is in base DT and overlay DT has "fpga-accelerator" node -> User programming overlay DT using fpgautil
	2) Zocl is part of the base DT and xclbin contains DT-Overlay which contains fpga-accelerator, XRT programs using LibDFx 
	3) Both zocl & fpga-accelerator is part of the overlay - VEK385
	4) Should be backward compatible
        5) If two xclbins are having a differrent interrupt related information

ManojTakasi:
Hi Vamshi,

I've updated the code to support case 1 as well. When zocl loads an xclbin, we now check whether an fpga_accelerator node with interrupts-extended is already present in the live device tree (e.g. applied earlier via fpgautil). If so, we call zocl_cu_intc_refresh() in addition to the existing refresh path when the xclbin carries PARTITION_METADATA. This allows zocl to pick up the latest CU IRQ routing even when loading a plain xclbin that does not contain a DT overlay section.

chvamshi-xilinx (approved):

PR #9994, opened 2026-08-24
xclbinutil: replace boost::factory with lambdas for Boost 1.92 compatibility

stsoe:

`boost/functional/factory.hpp` and `boost::factory<T*>()` were removed in Boost 1.92. GitHub Actions Windows runners upgraded to Boost 1.92, causing the Windows build of `xclbinutil` to fail with C1083 (cannot open include file) errors across all 37 `Section*.cxx` files.

Discovered via CI failures on Windows public GitHub runners after the Boost version was bumped from 1.91 to 1.92.

Replaced `boost::factory<T*>()` with an equivalent C++ lambda `[]{ return new T(); }` in all 37 `Section*.cxx` files, and removed the now-unnecessary `#include <boost/functional/factory.hpp>`.

The `Section_factory` type alias is already `std::function<Section*()>`, so a zero-capture lambda is a direct drop-in with no behavior change. No other alternatives were considered — this is the canonical replacement for the removed Boost utility.

None. The lambda is semantically identical to `boost::factory<T*>()` for this usage pattern (default-constructing a heap-allocated object).

CI on Windows with Boost 1.92 (the failing configuration). Please verify Linux build as well.

None.

### DIFF
--- a/src/runtime_src/tools/xclbinutil/SectionAIEMetadata.cxx
+++ b/src/runtime_src/tools/xclbinutil/SectionAIEMetadata.cxx
@@ -17,7 +17,6 @@
 #include "SectionAIEMetadata.h"
 
 #include "XclBinUtilities.h"
-#include <boost/functional/factory.hpp>
 #include <boost/property_tree/json_parser.hpp>
 
 namespace XUtil = XclBinUtilities;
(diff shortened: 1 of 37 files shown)


stsoe:
> clang-tidy made some suggestions
> 
> There were too many comments to post at once. Showing the first 25 out of 37. Check the log or trigger a new build to see more.

The warnings are legitimate: Section_factory is std::function<Section*()> (a raw owning pointer return), and this existed with `boost::factory<T*>()` too. The lambda is semantically identical, so we haven't made anything worse, but the underlying design is the real issue.

The proper fix is a separate refactor: change Section_factory to `std::function<std::unique_ptr<Section>()>` and update Section.cxx's call site and the SectionInfo constructor. That touches Section.h, Section.cxx, and all 37 registration sites — worth doing, but it's its own PR with its own risk surface.

For this PR we are ignoring all the warnings.

PR #9997, opened 2026-08-24
Add get_aie_coredump_elf() API for generating coredump (ET_CORE) ELF

sayyanna:
Add get_aie_coredump_elf() API for generating coredump (ET_CORE) ELF
Add get_aie_coredump_elf() API for generating coredump (ET_CORE) ELF. Currently AIE coredump is generated as a raw bin file, but this PR enables generated of coredump in ELF format.   The existing get_aie_coredump() raw blob API will be deprecated.
Added new API that package a raw AIE  coredump blob into an ET_CORE ELF using the AIEBU coredump ELF writer:
  std::vector<char>
  get_aie_coredump_elf(std::optional<xrt::aie::coredump_meta> meta = std::nullopt) const;

The AIE architecture is derived from the ELF loaded in this context.  Metadata (timestamp, versions, device info, uuid) is always embedded. If @meta is provided it is used as-is; otherwise metadata is built internally from the hw_context and device.
 The coredump ELF packaging logic lives in xrt_elf.cpp and is exposed internally via xrt_core::elf_int::make_aie_coredump_elf, keeping xrt_hw_context.cpp free of any direct aiebu dependency.

 Packaging the dump as an ET_CORE ELF solves:
  - Self-describing: the ELF header encodes the AIE architecture via the OS/ABI byte, so debuggers and analysis tools can identify the target without extra context
  - Metadata-rich: a dedicated ELF note section carries timestamp, firmware version, device info, ELF UUID, and context state at capture time — critical for correlating a dump with the exact firmware image and system state
  - Integration: callers like XDP can call get_aie_coredump_elf() directly and hand off a self-contained artifact without needing to know anything about the underlying architecture

None
Added below setting in xrt.ini
[Runtime]
aie_coredump_file = aie_coredump.elf

Enabled coredump: xrt-smi configure --advanced --auto-coredump --enable
then triggered  ERT_COM_STATE_TIMEOUT, coredump is ELF generated.


src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -1661,6 +1664,49 @@ get_filename(const xrt::elf_impl* elf_impl)
...
+aiebu::aiebu_assembler::buffer_type
+osabi_to_coredump_type(uint8_t os_abi)
+{
+  switch (os_abi) {
+  case aiebu::osabi_aie2p:  return aiebu::aiebu_assembler::buffer_type::coredump_aie2p;
+  case aiebu::osabi_aie2ps: return aiebu::aiebu_assembler::buffer_type::coredump_aie2ps;
+  case aiebu::osabi_aie4:   return aiebu::aiebu_assembler::buffer_type::coredump_aie4;
+  case aiebu::osabi_aie4a:  return aiebu::aiebu_assembler::buffer_type::coredump_aie4a;
+  case aiebu::osabi_aie4z:  return aiebu::aiebu_assembler::buffer_type::coredump_aie4z;
+  default: throw std::runtime_error("AIE coredump not supported for this ELF architecture");
```
stsoe: Impossible to debug, please new line after case

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -718,6 +718,32 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
   }
 
+  std::vector<char>
+  get_aie_coredump_elf() const
+  {
+    xrt::elf elf = [this] {
+      std::lock_guard lk(m_mutex);
+      if (m_elf_map.empty())
+        throw std::runtime_error("AIE coredump ELF not available: no ELF loaded in this context");
+      // All ELFs in one hw_context target the same AIE partition — any entry yields correct OS/ABI
```
stsoe: new line

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -718,6 +718,32 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
+  }
+
+  std::vector<char>
+  get_aie_coredump_elf(const xrt::aie::coredump_meta& meta) const
+  {
+    xrt::elf elf = [this] {
+      std::lock_guard lk(m_mutex);
+      if (m_elf_map.empty())
+        throw std::runtime_error("AIE coredump ELF not available: no ELF loaded in this context");
+      // All ELFs in one hw_context target the same AIE partition — any entry yields correct OS/ABI
```
stsoe: new line

src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -1661,6 +1664,49 @@ get_filename(const xrt::elf_impl* elf_impl)
...
+}
+
+} // namespace
+
+std::vector<char>
+make_aie_coredump_elf(const xrt::elf& elf, const std::vector<char>& blob)
+{
+  auto buf_type = osabi_to_coredump_type(elf.get_handle()->get_os_abi());
+  aiebu::aiebu_assembler a(buf_type, blob, aiebu::aiebu_assembler::no_meta);
+  return a.get_elf();
```
stsoe: What makes this XRT code?  Should it be in aiebu.  Comments may help explain why it is here.

src/runtime_src/core/include/xrt/xrt_hw_context.h
```diff
@@ -13,7 +13,31 @@
 
 #ifdef __cplusplus
 
+#include <cstdint>
 #include <map>
+#include <string>
+#include <vector>
+
+namespace xrt::aie {
+
```
stsoe: This is a public API, these structs need extensive comments. 
I am not sure about the aie namespace within this file.  I need to think about this.

src/runtime_src/core/include/xrt/xrt_hw_context.h
```diff
@@ -357,6 +381,32 @@ class hw_context : public detail::pimpl<hw_context_impl>
   std::vector<char>
   get_aie_coredump() const;
 
+  /**
+   * get_aie_coredump_elf() - Returns the coredump of AIE Array as an ET_CORE ELF.
```
stsoe: What's an ET_CORE?

stsoe (changes requested):
Please comment the code, structs, etc.

stsoe:
One more question, who is the consumer of these new APIs?  If it is XDP, I would argue that they should add this code themselves.  E.g. first get the raw coredump from XRT, then re-package it using aiebu.

src/runtime_src/core/include/xrt/xrt_hw_context.h
larry9523: It is a flag in ELF header so all GNU tools know this is a coredump file

larry9523 (changes requested):
A few high level comments
1. please add some detail in the PR description about why are we doing this.
2. I would suggest we only provide one public API get_aie_coredump_elf() internally it will create an ELF coredump with metadata. It does not hurt to add metadata into ELF. (Thanks @stsoe for the input)
3. I think there is no consumer of xrt.ini path to create aie coredump (I confirm this with @IshitaGhosh and @jvillarre). So while we keep our xrt::hw_context::get_aie_coredump() API, we should generate ELF format coredump with metadata for the xrt,ini path.

src/runtime_src/core/include/xrt/xrt_hw_context.h
```diff
@@ -357,6 +361,25 @@ class hw_context : public detail::pimpl<hw_context_impl>
...
+   *
+   * @param meta
+   *  Optional metadata to embed. When omitted, metadata is populated internally.
+   *
+   * @return
+   *  ELF bytes of the ET_CORE coredump ELF
+   */
+  XRT_API_EXPORT
+  std::vector<char>
+  get_aie_coredump_elf(std::optional<xrt::aie::coredump_meta> meta = std::nullopt) const;
```
stsoe: Why is this optional?   Wy not always embed it?
Also,  my preference is two APIs and no use of std::optional.

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -1015,6 +1054,13 @@ get_aie_coredump() const
   return get_handle()->get_aie_coredump();
 }
 
+std::vector<char>
+hw_context::
+get_aie_coredump_elf(std::optional<xrt::aie::coredump_meta> meta) const
+{
+  return get_handle()->get_aie_coredump_elf(std::move(meta));
+}
+
```
stsoe, line 1049: As I mentioned earlier, we should use std::optional the public API.  Either we always include all meta data, or we provide two APIs, where one implementation creates the default meta struct before calling handle->get_aie_coredump_elf().  My question is why meta data is an option?

src/runtime_src/core/include/xrt/experimental/xrt_aie_coredump.h
```diff
@@ -0,0 +1,122 @@
...
+// Copyright (C) 2025 Advanced Micro Devices, Inc. All rights reserved.
+#ifndef XRT_AIE_COREDUMP_H_
+#define XRT_AIE_COREDUMP_H_
+
+#ifdef __cplusplus
+
+#include <cstdint>
+#include <string>
+
+namespace xrt::aie {
```
stsoe: Are these additions needed?

src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -1661,6 +1668,103 @@ get_filename(const xrt::elf_impl* elf_impl)
...
+    std::string drv_str;
+    if (const auto drivers = pt_xrt.get_child_optional("drivers")) {
+      for (const auto& kv : *drivers) {
+        const auto& drv = kv.second;
+        auto name = drv.get<std::string>("name", "");
+        auto ver  = drv.get<std::string>("version", "");
+        if (!name.empty() && !ver.empty() && ver != "unknown") {
+          if (!drv_str.empty())
+            drv_str += "; ";
+          drv_str += name + " " + ver;
```
stsoe, line 1714:
```suggestion
          if (!drv_str.empty())
            drv_str += "; ";
            
          drv_str += name + " " + ver;
```

src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -1661,6 +1668,103 @@ get_filename(const xrt::elf_impl* elf_impl)
...
+  try {
+    auto data = xrt_core::device_query<xrt_core::query::aie_partition_info>(device);
+    auto islot = static_cast<int>(slot);
+    for (const auto& entry : data) {
+      if (std::stoi(entry.metadata.id) == islot) {
+        meta.context_status = entry.is_suspended
+            ? aiebu::aie_context_status::idle
+            : aiebu::aie_context_status::running;
+        break;
+      }
```
stsoe: One less scope with early-continue

```suggestion
    for (const auto& entry : data) {
      if (std::stoi(entry.metadata.id) != islot)
        continue;
        
      meta.context_status = entry.is_suspended
          ? aiebu::aie_context_status::idle
          : aiebu::aie_context_status::running;
      break;
    }
```

src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -1661,6 +1668,103 @@ get_filename(const xrt::elf_impl* elf_impl)
...
+  case aiebu::osabi_aie4a:
+    return aiebu::aiebu_assembler::buffer_type::coredump_aie4a;
+  case aiebu::osabi_aie4z:
+    return aiebu::aiebu_assembler::buffer_type::coredump_aie4z;
+  default:
+    throw std::runtime_error("AIE coredump not supported for this ELF architecture");
+  }
+}
+
+} // namespace
```
stsoe: Don't scope the anonymous namespace inside a named namespace, move it out, or make this a static function within current namespace to keep the proximity to where it is used.

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -718,6 +718,22 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
     }
   }
 
+  std::vector<char>
+  get_aie_coredump_elf() const
```
stsoe, line 722: This function retrieves the first elf and comments that it is good enough because they all share the OS/ABI, so this makes it sound as if the elf is arbitrary and good enough as long as it is within m_elf_map.  But make_aie_coredump() retrieves the elf uuid, so that is not arbitrary?

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -3167,13 +3169,12 @@ class run_impl : public std::enable_shared_from_this<run_impl>
...
     try {
       auto hwctx = kernel->get_hw_context();
-      auto core = hwctx.get_aie_coredump();  // may throw
+      auto core = hwctx.get_aie_coredump_elf();  // may throw
 
       std::ofstream ostr{file, std::ios::binary};
       if (!ostr)
         throw std::runtime_error("Could not open '" + file + "' for writing");
       ostr.write(core.data(), static_cast<std::streamsize>(core.size()));
-      std::abort();
```
stsoe, line 3176: I think there was a mention that we should not abort on coredump, but the original spec said we should.  I want to make sure remove std::abort() is indeed correct.

larry9523: We should keep the std::abort() to make sure driver clean up hw_context after timeout. We don't have to get the exception message and coredump() at one run. When usesrs set xrt.ini to save aie coredump, they already see the timeout and exception message.

src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -1661,6 +1668,102 @@ get_filename(const xrt::elf_impl* elf_impl)
     : "";
 }
 
+aiebu::aiebu_assembler::buffer_type
```
stsoe: It was previously in a nested unnamed namespace, I suggested either moving the namespace or making the function static,
```suggestion
static aiebu::aiebu_assembler::buffer_type
```

src/runtime_src/core/common/api/xrt_elf.cpp
```diff
@@ -1661,6 +1668,102 @@ get_filename(const xrt::elf_impl* elf_impl)
...
+          std::chrono::system_clock::now().time_since_epoch()).count());
+  meta.uuid = uuid;
+
+  try {
+    boost::property_tree::ptree pt_xrt;
+    xrt_core::get_driver_info(pt_xrt);
+    std::string drv_str;
+    if (const auto drivers = pt_xrt.get_child_optional("drivers")) {
+      for (const auto& kv : *drivers) {
+        const auto& drv = kv.second;
```
stsoe: It is generally preferred to use structured bindings for anything return tuples.

```suggestion
      for (const auto& [dummy, drv] : *drivers) {
```

src/runtime_src/core/common/api/xrt_hw_context.cpp
```diff
@@ -718,6 +718,22 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
...
+  {
+    xrt::elf elf = [this] {
+      std::lock_guard lk(m_mutex);
+      if (m_elf_map.empty())
+        throw std::runtime_error("AIE coredump ELF not available: no ELF loaded in this context");
+      return m_elf_map.begin()->second;
+    }();
+
+    return xrt_core::elf_int::make_aie_coredump_elf(
+        elf, get_aie_coredump(), m_core_device.get(), m_hdl->get_slotidx(), "");
```
stsoe: You defaulted the uuid string in make_aie_coredump_elf, which I think is fine, but then this function should not mention anything about UUID and not call with the empty string.

stsoe (commented):
Looks good.  Minor corrections.

stsoe (approved):

src/runtime_src/core/include/xrt/xrt_hw_context.h
```diff
@@ -357,6 +359,21 @@ class hw_context : public detail::pimpl<hw_context_impl>
   std::vector<char>
   get_aie_coredump() const;
 
+  /**
+   * get_aie_coredump_elf() - Returns the coredump of AIE Array as an ET_CORE ELF.
+   * Fetches the raw AIE dump blob and packages it into a coredump ELF.
+   * The AIE architecture is derived from the ELF loaded in this context.
+   * Metadata (timestamp, firmware version, device info, ELF UUID) is always
```
larry9523: UUID is not always present.

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -3154,7 +3154,10 @@ class run_impl : public std::enable_shared_from_this<run_impl>
   }
 
   // abort_coredump_or_noop() - AIE coredump if enabled
-  // Dump core and abort, or do nothing.
+  // Write AIE coredump ELF to the configured file then abort.
+  // std::abort() ensures the OS reclaims driver resources (hw_context, hardware
+  // state) after a timeout. Users who set xrt.ini to capture AIE coredump
+  // have already observed the timeout exception on a prior run.
```
larry9523: nit: "Users who..." is a regular working flow but not the code itself. You can make it clear that users would see either coredump w/ xrt.ini or exception message w/o xrt.ini

src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -3167,12 +3170,22 @@ class run_impl : public std::enable_shared_from_this<run_impl>
 
     try {
       auto hwctx = kernel->get_hw_context();
-      auto core = hwctx.get_aie_coredump();  // may throw
+      // Use the ELF from this run's own module
+      // Each ELF has its own UUID; using the run's module ensures the
+      // correct UUID is embedded in the coredump metadata.
+      if (!m_module)
+        throw std::runtime_error("AIE coredump ELF not available: no ELF associated with this run");
```
larry9523, line 3179: So the xrt.ini only applies to full elf flow? If so, you may want to make it clear in the comments

src/runtime_src/core/include/xrt/xrt_hw_context.h
```diff
@@ -357,6 +359,21 @@ class hw_context : public detail::pimpl<hw_context_impl>
   std::vector<char>
   get_aie_coredump() const;
 
+  /**
+   * get_aie_coredump_elf() - Returns the coredump of AIE Array as an ET_CORE ELF.
```
larry9523:
```suggestion
   * get_aie_coredump_elf() - Returns the coredump of AIE Array as an ELF (e_type = ET_CORE)
```

larry9523 (changes requested):
Looks good. Some minor suggestions.

larry9523 (approved):

PR #9999, opened 2026-08-25
Fix self-deadlock on hw_context m_mutex when XDP re-enters add_config().

sawanm-xilinx:
Full-ELF designs hang during hardware-context setup when XDP profiling is enabled.

The bug came in with the full-ELF XDP hook: init_from_elf() calls xrt_core::xdp::update_device(), and its only caller add_config() holds m_mutex across that call. Discovered on VEK385 while validating XDP profiling on full-ELF designs — runs with aie_trace/aie_profile hung before inference and left the AIE partition claimed, requiring a board re-flash. 

m_mutex is now held only around the mutation. add_config() records whether the call performed the initialization, releases the lock, and then calls xdp::update_device(); the call is removed from init_from_elf(), which documents that the caller owns it.

Low. The lock still covers m_hdl creation and all data-structure updates, so no new race is introduced; only the XDP notification moved. The notification remains one-shot per context via the initialized latch, so XDP's re-entrant add_config() calls take the update_from_elf() branch and do not recurse. 

Verified on VEK385 (VE2) hardware with two unrelated full-ELF designs, across all combinations of aie_trace, aie_profile, and ml_timeline. No hangs occurred, and instrumentation output is produced in each case.

### DIFF
--- a/src/runtime_src/core/common/api/xrt_hw_context.cpp
+++ b/src/runtime_src/core/common/api/xrt_hw_context.cpp
@@ -273,6 +273,10 @@ class hw_context_impl : public std::enable_shared_from_this<hw_context_impl>
   // init_from_elf() - Initialize this hwctx from an ELF
   // This function is called if and only if the hwctx has no
   // underlying representation (no m_hdl).
+  //
+  // The caller (add_config()) is responsible for calling
+  // xdp::update_device() once this function has returned and the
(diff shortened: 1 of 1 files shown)


stsoe (approved):

stsoe:
Please sign off.

chvamshi-xilinx (approved):

PR #10014, opened 2026-08-27
xbmgmt: validate mem_topology size before ECC reset iteration

aktondak:
Reject undersized or empty mem_topology sysfs blobs before iterating m_mem_data entries in reset_ecc(), preventing out-of-bounds reads. 

https://jira.xilinx.com/browse/AIESW-42781

Resolved via accepting the ticket's suggestion

NA

### DIFF
--- a/src/runtime_src/core/tools/xbmgmt2/SubCmdReset.cpp
+++ b/src/runtime_src/core/tools/xbmgmt2/SubCmdReset.cpp
@@ -37,13 +37,15 @@ static void
 reset_ecc(const std::shared_ptr<xrt_core::device>& dev, xrt_core::query::reset_type& reset)
 {
   auto raw_mem = xrt_core::device_query<xrt_core::query::mem_topology_raw>(dev);
-  const mem_topology *map = (mem_topology *)raw_mem.data();
-    if(raw_mem.empty() || map->m_count == 0) {
-      std::cout << "WARNING: 'mem_topology' not found, "
+  const auto *map = reinterpret_cast<const mem_topology *>(raw_mem.data());
+  if (map->m_count <= 0 || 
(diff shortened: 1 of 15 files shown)


AShivangi (approved):

stsoe:
signoff please

src/runtime_src/core/tools/xbmgmt2/SubCmdReset.cpp
stsoe: This is part of the ticket AIESW-42781 but it is not fixed.
I highly recommend feeding the ticket to Claude and ask Claude to fix the ticket, then review.

stsoe (changes requested):

src/runtime_src/core/tools/xbmgmt2/SubCmdReset.cpp
```diff
@@ -37,18 +37,21 @@ static void
...
-    }
+    return;
+  }
 
     for(int32_t i = 0; i < map->m_count; i++) {
       if(!map->m_mem_data[i].m_used)
         continue;
-      reset.set_subdev(reinterpret_cast<const char *>(map->m_mem_data[i].m_tag));
+      reset.set_subdev(std::string(reinterpret_cast<const char *>(map->m_mem_data[i].m_tag),
+                                   sizeof(map->m_mem_data[i].m_tag)));
```
stsoe: sizeof(m_tab) is 16,what if strlen of m_tag is less?  You get a std::string padded with 0's, which is not the same as null termination.

```suggestion
    // m_tag is 16 bytes and may not be NUL-terminated in externally-produced xclbins.
    const auto& tag = map->m_mem_data[i].m_tag;
    reset.set_subdev(std::string(reinterpret_cast<const char *>(tag),
                                 strnlen(reinterpret_cast<const char *>(tag), sizeof(tag))));
```

PR #10028, opened 2026-09-01
CR-1272937: xrt::kernel::group_id() throw on missing memory connectivity

svup-xilinx:
group_id() returned 65535 when a kernel argument had no memory bank. Passing that to xrt::bo() caused a later std::out_of_range crash. The commit throws a clear error instead.
[CR-1272937](https://jira.xilinx.com/browse/CR-1272937)
group_id() now throws EINVAL if the argument index is invalid or no memory group is assigned.
Low
Tested on vck190 board
NA

### DIFF
--- a/src/runtime_src/core/common/api/xrt_kernel.cpp
+++ b/src/runtime_src/core/common/api/xrt_kernel.cpp
@@ -635,6 +635,7 @@ private:
 // then behavior is undefined.
 class ip_context
 {
+public:
   // class connectivy - Represents argument connectiviy to memory banks
   //
   // The argument connectivity is represented using a compressed bitset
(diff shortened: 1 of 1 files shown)


src/runtime_src/core/common/api/xrt_kernel.cpp
```diff
@@ -2149,11 +2149,20 @@ class kernel_impl : public std::enable_shared_from_this<kernel_impl>
...
 
+    if (argno < 0 || static_cast<size_t>(argno) >= args.size())
+      throw xrt_core::error(EINVAL, "No such kernel argument at index " + std::to_string(argno)
+                            + " for kernel '" + name + "'");
+
     // Last (for group id) connection of first ip in this kernel
     // The group id can change if cus are trimmed based on argument
     auto& ip = ipctxs.front();  // guaranteed to be non empty
+    auto memidx = ip->arg_memidx(argno);
+    if (memidx < 0)
```
stsoe: Please change class connectivity to make the compile time constant no_memidx public, then use that here:

```suggestion
    if (memidx == ip_context::connectivity::memidx)
```
If that does quite work because connectivity is a private embedded class, then it's fine to make class connectivity public within class ip_context,

stsoe (changes requested):

stsoe (approved):

PR #10037, opened 2026-09-02
Update aiebu to fix arm64 cross compilation

stsoe:
Pull in https://github.com/Xilinx/aiebu/pull/370

Amend #10032 which fails in cross compilation


stsoe:
Merging despite failing pcie_hw_tests, which is infa
bash-4.2$ more 1.stderr
None/skel/g_cshrc: No such file or directory.
lsid: ls_getentitlementinfo() failed: Communication time out

PR #10038, opened 2026-09-03
[AIESW-28579] : Sub Buffer export-import handling

rbramand-xilinx:
Fix sub buffer export import handling.

It was discovered in local testing by VART team

class buffer_sub inherits class bo_impl which is base impl class for xrt::bo and export_buffer() function is not virtual so buffer_sub uses same function which is incorrect.
Made export_buffer() as virtual and implemented it in sub buffer which throws exception. Users can export parent buffer and send fd, offset, size to importer and it can import parent and recreate sub buffer using offset, size.

Fix can also come from driver but it is tedious so following this approach

Low

Tested on Telluride hw and things work as expected 
Checked in the test case to Telluride repo - https://gitenterprise.xilinx.com/XRT/Telluride/pull/808

Added inline comments and doxygen comments

### DIFF
--- a/src/runtime_src/core/common/api/xrt_bo.cpp
+++ b/src/runtime_src/core/common/api/xrt_bo.cpp
@@ -369,13 +369,21 @@ public:
     throw xrt_core::error("buffer is not mapped");
   }
 
-  export_handle
+  virtual export_handle
   export_buffer() const
   {
     if (!shared_handle)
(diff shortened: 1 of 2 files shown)


stsoe (approved):
Perfect.  Thanks!

chvamshi-xilinx (approved):

